PR #6, opened 2017-03-01
Update logreg.md

jonathanBieler:
Small typo in example


MikeInnes:
Thanks for the patch. It's useful to hold on to `m` separately actually, for example to save the model, so I think it'd be better to change line 43 to `model = tf(m)`.

MikeInnes:
:+1: thanks for Flux's first PR :)

PR #12, opened 2017-03-15
Add semicolon after readstring

mkborregaard:
BTW this code is slightly different from the version in the examples folder (e.g in that there is a specification to look in the home folder)


mkborregaard:
Sorry I sent two - this is perhaps the important one (for the docs), but I got lazi and did the change directly on github, so it made two different PRs.

MikeInnes:
Ah cool, I thought I had a duplicate email, but this works. Thanks!

PR #18, opened 2017-04-09
Support batched training

ylxdzsw:
This PR add batched training by:
- adding a `Batched` type, which wraps an Iterator and output in batch
- add keyword argument `all` for `update!` of MXNet backend, which copy the final params to all execs
- make `mse!` supports batched arguments
- alter the training process of TensorFlow backend
- fix a little bug of `CatMat`

### DIFF
--- a/src/dims/catmat.jl
+++ b/src/dims/catmat.jl
@@ -15,7 +15,7 @@ size(b::CatMat) = (size(b.data, 1),)
 
 getindex(b::CatMat, i)::eltype(b) = slicedim(b.data, 1, i)
 
-setindex!(b::CatMat, v, i) = b[i, :] = v
+setindex!(b::CatMat, v, i) = b.data[i, :] = v
 
 allequal(xs) = all(x -> x == first(xs), xs)
 


src/backend/tensorflow/model.jl
```diff
@@ -57,11 +57,11 @@ function Flux.train!(m::Model, train, test=[]; epoch = 1, η = 0.1,
...
   for e in 1:epoch
     info("Epoch $e\n")
-    @progress for (x, y) in train
+    @progress for (x, y) in Batched(train, batch)
       y, cur_loss, _ = run(m.session, vcat(m.output, Loss, minimize_op),
-                           Dict(m.inputs[1] => batchone(convertel(Float32, x)),
-                                Y => batchone(convertel(Float32, y))))
-      if i % 5000 == 0
+                           Dict(m.inputs[1] => rebatch(convertel(Float32, rawbatch(x))),
+                                Y => rebatch(convertel(Float32, rawbatch(y)))))
```
MikeInnes: `convertel` should already work on batches without needing to convert to an array and back

src/cost.jl
```diff
@@ -1,7 +1,7 @@
 export mse, mse!
 
 function mse!(Δ, pred, target)
-  map!(-, Δ, pred, target)
+  map!(-, Δ, rawbatch(pred), rawbatch(target))
```
MikeInnes: It seems wrong that `mse!` needs to know about batching; in general, we should be able to write functions over generic arrays and have batching work for free. In this case, perhaps we can implement `broadcast!` for `Batch` so that `mse!` works as is? If it's complicated to do this it doesn't have to block the PR.

src/dims/batching.jl
```diff
@@ -59,3 +62,71 @@ runrawbatched(f, xs...) =
...
+mutable struct Batched{T,S}
+  batch::Int
+  iter::T
+  "`Batched` always read a batch in advance, and store it in `buf`"
+  buf::S
+  i
+end
+
+function Batched(iter::T, batch::Integer) where T
+  batch >= 1 || throw(ArgumentError("batch size must >= 1"))
```
MikeInnes: Is there a particular reason to disable batch size 1?

src/dims/batching.jl
```diff
@@ -59,3 +62,71 @@ runrawbatched(f, xs...) =
...
+
+next(x::Batched, ::Bool) = x.buf, false
+
+# will be less hacky if https://github.com/JuliaLang/julia/issues/18823
+function done(x::Batched, fresh)
+  fresh && return false
+
+  for ibatch in 1:x.batch
+    if done(x.iter, x.i)
+      warn("cannot perfectly divide data by batch size, remainder will be discarded")
```
MikeInnes: In principle, we could just return a smaller final batch. I'm not sure if we should or not, but it might be a nicer default.

src/utils.jl
```diff
@@ -6,16 +6,20 @@ initn(dims...) = randn(dims...)/100
 
 function train!(m, train, test = []; epoch = 1, batch = 10, η = 0.1)
     i = 0
-    Δ = zeros(length(train[1][2]))
+    Δ = zeros(batch, length(train[1][2]))
     for _ in 1:epoch
-      @progress for (x, y) in train
+      @progress for (x, y) in Batched(train, batch)
```
MikeInnes: I think perhaps `train!` should stay generic and continue to accept an iterator of either raw arrays or batches, rather than trying to handle batching entirely itself. That gives you the control to batch things ahead of time, for example.

src/utils.jl
```diff
@@ -6,16 +6,20 @@ initn(dims...) = randn(dims...)/100
...
         pred = m(x)
-        any(isnan, pred) && error("NaN")
+        any(any(isnan, p) for p in pred) && error("NaN")
         err = mse!(Δ, pred, y)
         back!(m, Δ, x)
-        i % batch == 0 && update!(m, η)
-        i % 1000 == 0 && @show accuracy(m, test)
+        if i % 100 == 0
+          update!(m, η, all=true)
+          @show accuracy(m, test)
```
MikeInnes: When I last tried this I had problems with `accuracy` not taking batches. Are the `mse!` changes enough to fix that?

src/backend/mxnet/model.jl
```diff
@@ -127,7 +127,17 @@ function Flux.back!(m::Model, Δ, xs...)
   end
 end
 
-Flux.update!(m::Model, η) = (update!(m.last, η); m)
+Flux.update!(m::Model, η; all=false) = begin
```
ylxdzsw: Yes, I saw your note in the file. I will try to implement that way. Just another idea: is it possible to use exactly the same `NDArray` among `Exec`s? I think move the `params` local varaiable in `mxparams` function to the `Model` or `Graph` object and do not reinitialize it every time should do the trick. This should completely get rid of this problem and in addition, save precious GPU memory.

src/backend/tensorflow/model.jl
ylxdzsw: Yes, it does. I didn't notice that.

src/cost.jl
ylxdzsw: I'm afraid that overloading `broadcast!` is not enough, as this is the `map!` in julia 0.6:

```
function map!{F}(f::F, dest::AbstractArray, A::AbstractArray, B::AbstractArray)
    for (i, j, k) in zip(eachindex(dest), eachindex(A), eachindex(B))
        dest[i] = f(A[j], B[k])
    end
    return dest
end
```

We treat `Batch` as an `Vector` of `Array`s, so the dimensions of those `eachindex`s are not match. What about just specialize `mse!` for `Batch`s?

src/dims/batching.jl
ylxdzsw: No, it doesn't. batch == 1 is allowed and I tested.

src/dims/batching.jl
ylxdzsw: AFAIK most frameworks simply discard the remainder, but yes, it worth being implemented as at least an option for user to choose.

src/utils.jl
ylxdzsw: How do we know the elements of an iterator is a `Batch` or raw `Array`? For example, if the iterator yields `Array`s of size (5, 28, 28, 3), is it a `Batch` of (28, 28, 3) `Array`s or just an element of 4-dimension array? Or should we export `rebatch` function and tell users to pack `Batch`es in the first case?

ylxdzsw: Here's the Keras way: Keras has two methods, `fit` and `fit_generator`, the `fit` accept a big array (can be treat as an entire batch of all data) and split it by `batch_size`; while `fit_generator` accepts iterator that generates batchs and do not have `batch_size` argument. I think we can do the same thing: `train!` still accept `batch` and pack the elements of iterator accroding to it, and add a new `train_batched!` allow user to yield the batch and we just map `rebatch` on the iterator.

src/utils.jl
ylxdzsw: umm, In fact I didn't touch it and still evaluate the model one sample at a time. It should not be hard to support batch though.

src/dims/batching.jl
MikeInnes: Ok, I'm happy to follow precedent here.

src/dims/batching.jl
MikeInnes: Obviously – reviewing PRs too late :)

src/utils.jl
MikeInnes: Whoops, my bad, I was thinking of when I tried `train!` with an iterator of batches; which almost worked aside from this line. Of course, if you expect `test` to be samples then this will work. We can discuss that assumption elsewhere.

src/backend/mxnet/model.jl
MikeInnes: I don't know that it *won't* work, so I guess it's worth a shot :)

src/cost.jl
MikeInnes: We can just use `broadcast!` instead of `map!` here.

src/utils.jl
MikeInnes: Yes, `Batch` is supposed to be user facing. The idea is that we then don't ever have to worry about the question of "is this a batch?", making things less error prone. We should also be able to write functions that work in a generic way for batches and samples, avoiding duplication.

The model interface will accept either a sample or a batch, so in theory `train!` need not know about batching at all. We can then call either `train!(m, xs)` or `train!(m, batched(xs, 10))`. (This doesn't prevent `train!` from having a `batch_size` argument for convenience / familiarity of course.)

MikeInnes: #21 will invalidate some of what I said above. So let's just get this working in a simple way, as you've done it, and worry about making things more general later.

ylxdzsw:
Rebased, plus all execs in the same mx model now share the same set of args and grads - we don't need to worry about `last` anymore.

ylxdzsw:
Rebased with latest master again. Will this get merged in the near future? Keeping syncing with master is somewhat error prone.

MikeInnes:
Err, whoops. I tried to push a change to your branch but it ended up deleting the commits.

Rest assured I have these locally and I'm rebasing/merging them right now.

ylxdzsw:
Yeah I have a local backup too. I'm a little surprised that GitHub allows pushing to forked repos directly.

MikeInnes:
Maintainers can push to branches used for PRs, which is pretty handy for patching things up before a merge without going back and forth. Of course, as soon as the PR closed itself that no longer applied, which meant I couldn't fix my mistake :P

PR #20, opened 2017-04-13
Add a graphviz backend

ylxdzsw:
This adds a "fake" backend that doesn't perform actual training/prediction. Instead, it output a DOT string witch can be rendered as an image. For example:

```
using Flux

m = Chain(
         Input(784),
         Affine(784, 128), relu,
         Affine(128, 64), relu,
         Affine(64, 10), softmax)

dot, n_params = graphviz(m)(Array{Float32}(4, 784))

open("some.dot", "w") do f write(f, dot) end
```

then `dot -Tjpeg some.dot > some.jpeg`, or use `Graphviz.jl`(which unfortunatly broken on 0.6) to get this:

image

It also counts the elements of all tensors, which can be used to estimate the memory cost and can help users to select a batch size.

### DIFF
--- a/src/backend/backend.jl
+++ b/src/backend/backend.jl
@@ -20,8 +20,6 @@ function mxnet(m)
   eval(:(MX.mxnet($m)))
 end
 
-include(joinpath(dirname(@__FILE__), "graphviz/graphviz.jl"))
+include("graphviz/graphviz.jl")
 
-function graphviz(m)
-  GV.graphviz(m)
-end
+const graphviz = GV.graphviz
--- a/src/backend/backend.jl
+++ b/src/backend/backend.jl
@@ -1,4 +1,4 @@
-export tf, mxnet
+export tf, mxnet, graphviz
 
 function loadtf()
   isdefined(Flux, :TF) && return
@@ -18,4 +18,10 @@ end
 function mxnet(m)
   loadmx()
   eval(:(MX.mxnet($m)))
+end
+
+include(joinpath(dirname(@__FILE__), "graphviz/graphviz.jl"))
+
+function graphviz(m)
+  GV.graphviz(m)
 end
--- a/src/backend/graphviz/graph.jl
+++ b/src/backend/graphviz/graph.jl
@@ -1,0 +1,156 @@
+import Base: size, <<
+
+struct Tensor
+  name
+  size
+end
+
+size(t::Tensor) = t.size
+
+uid = ((id)->()->id+=1)(0)
+
+function <<(ctx::Context, str)
+  println(ctx[:buffer], "  " ^ ctx[:depth], str...)
+  ctx
+end
+
+function <<(ctx::Context, t::Tensor)
+  ctx[:memory] += isempty(t.size) ? 0 : *(t.size...)
+  ctx
+end
+
+function scope(f, ctx, name)
+  ctx << "subgraph cluster_$(uid()) {" << "  label = \"$name\""
+  ctx[:depth] += 1
+  result = f()
+  ctx[:depth] -= 1
+  ctx << "}"
+  result
+end
+
+function interp(ctx, c::DataFlow.Constant)
+  interp(ctx, c.value)
+end
+
+function interp(ctx, c::Flux.Param)
+  t = Tensor(uid(), size(c.x)) # TODO: is it possible to get variable name?
+  ctx << t
+  ctx << """$(t.name) [shape=box, label="param::$(t.size)"]"""
+  t
+end
+
+function interp(ctx, a::AArray)
+  t = Tensor(uid(), size(a))
+  ctx << t
+  ctx << """$(t.name) [shape=hexagon, label="input::$(t.size)"]"""
+  t
+end
+
+interp(ctx, t::Tensor) = t
+
+function interp(ctx, m::Flux.Model, xs...)
+  scope(ctx, typeof(m)) do
+    g = Flux.graph(m)
+    g == nothing && error("graphviz backend doesn't support $m")
+    DataFlow.iscyclic(g) && error("This model has a cycle; try unrolling it first.")
+    interpret(ctx, g, xs...)
+  end
+end
+
+interp(ctx, i::Input, t) = t
+
+function interp(ctx, c::Conv2D, x)
+  size_intered = Flux.infer(c, x.size)
+  node = replace("""
+    Conv2D(
+      kernel: $(size(c.filter.x, 4))
+      filter: $(join(size(c.filter.x)[1:2], 'x'))
+      stride: $(join(c.stride, 'x'))
+      border: valid
+    )""", '\n', "\\n")
+  t = Tensor(uid(), size_intered == nothing ? () : size_intered)
+  ctx << t
+  ctx << """$(t.name) [shape=oval, label="$node::$(t.size)"]"""
+  ctx << "$(x.name) -> $(t.name)"
+  t
+end
+
+function interp(ctx, p::MaxPool, x)
+  size = Flux.infer(p, x.size)
+  node = replace("""
+    Pool(
+      filter: $(join(p.size, 'x'))
+      stride: $(join(p.stride, 'x'))
+      function: max
+    )""", '\n', "\\n")
+  t = Tensor(uid(), size == nothing ? () : size)
+  ctx << t
+  ctx << """$(t.name) [shape=oval, label="$node::$(t.size)"]"""
+  ctx << "$(x.name) -> $(t.name)"
+  t
+end
+
+function interp(ctx, r::Reshape, x)
+  t = Tensor(uid(), *(x.size[2:end]...) == *(r.dims...) ? (x.size[1], r.dims...) : ())
+  ctx << """$(t.name) [shape=oval, label="Reshape::$(t.size)"]"""
+  ctx << "$(x.name) -> $(t.name)"
+  t
+end
+
+function interp(ctx, m::Flux.Stateful)
+  error("graphviz backend doesn't support recurrent model yet")
+end
+
+interp(ctx, m::Flux.SeqModel) = interp(ctx, m.model)
+
+for op in (:.+, :.-, :.*, :./)
+  op_str = string(op)
+  @eval function interp(ctx, ::typeof($op), xs...)
+    t = Tensor(uid(), try
+      any(isempty, map(size, xs)) ? () :
+      Base.Broadcast.broadcast_shape(map(size, xs)...)
+    catch e
+      e isa DimensionMismatch ? () : rethrow()
+    end)
+    ctx << t
+    ctx << """$(t.name) [shape=oval, label="$($op_str)::$(t.size)"]"""
+    for x in xs
+      ctx << "$(x.name) -> $(t.name)"
+    end
+    t
+  end
+end
+
+for op in (:tanh, :σ, :relu, :softmax)
+  op_str = string(op)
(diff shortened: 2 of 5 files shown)


src/backend/backend.jl
```diff
@@ -19,3 +19,9 @@ function mxnet(m)
   loadmx()
   eval(:(MX.mxnet($m)))
 end
+
+include(joinpath(dirname(@__FILE__), "graphviz/graphviz.jl"))
```
MikeInnes: This is equivalent to `include("graphviz/graphviz.jl")`

src/backend/graphviz/graph.jl
```diff
@@ -0,0 +1,156 @@
...
+  ctx << "subgraph cluster_$(uid()) {" << "  label = \"$name\""
+  ctx[:depth] += 1
+  result = f()
+  ctx[:depth] -= 1
+  ctx << "}"
+  result
+end
+
+function interp(ctx, c::DataFlow.Constant)
+  interp(ctx, c.value)
```
MikeInnes: This is technically a little sketchy as a constant could theoretically be a dataflow graph itself; this will cause it to be interpreted when it should be left as a value.

src/backend/graphviz/graph.jl
```diff
@@ -0,0 +1,156 @@
...
+end
+
+function interp(ctx, c::Flux.Param)
+  t = Tensor(uid(), size(c.x)) # TODO: is it possible to get variable name?
```
MikeInnes: We don't currently insert variable names into the graph in any way, although that may be a useful feature eventually.

src/backend/graphviz/graph.jl
```diff
@@ -0,0 +1,156 @@
...
+
+uid = ((id)->()->id+=1)(0)
+
+function <<(ctx::Context, str)
+  println(ctx[:buffer], "  " ^ ctx[:depth], str...)
+  ctx
+end
+
+function <<(ctx::Context, t::Tensor)
+  ctx[:memory] += isempty(t.size) ? 0 : *(t.size...)
```
MikeInnes: `*(xs...) == prod(xs)`

src/backend/graphviz/graph.jl
```diff
@@ -0,0 +1,156 @@
+import Base: size, <<
```
MikeInnes: Overloading `<<` creates a globally accessible function on `Context` that really only makes sense within this module, so I suggest getting rid of the import.

src/backend/graphviz/graph.jl
```diff
@@ -0,0 +1,156 @@
...
+  t
+end
+
+interp(ctx, t::Tensor) = t
+
+function interp(ctx, m::Flux.Model, xs...)
+  scope(ctx, typeof(m)) do
+    g = Flux.graph(m)
+    g == nothing && error("graphviz backend doesn't support $m")
+    DataFlow.iscyclic(g) && error("This model has a cycle; try unrolling it first.")
```
MikeInnes: I believe GraphViz can handle cycles in principle, so it might be nice to support this; but it would probably require a more custom approach, like `DataFlow.syntax` uses.

src/backend/graphviz/graph.jl
```diff
@@ -0,0 +1,156 @@
...
+end
+
+interp(ctx, m::Flux.SeqModel) = interp(ctx, m.model)
+
+for op in (:.+, :.-, :.*, :./)
+  op_str = string(op)
+  @eval function interp(ctx, ::typeof($op), xs...)
+    t = Tensor(uid(), try
+      any(isempty, map(size, xs)) ? () :
+      Base.Broadcast.broadcast_shape(map(size, xs)...)
```
MikeInnes: Isn't this replicating parts of Flux's inference that you can reuse?

src/backend/graphviz/model.jl
```diff
@@ -0,0 +1,14 @@
+mutable struct Model <: Flux.Model
+  graph
+end
+
+graphviz(m::Flux.Model) = Model(Flux.graph(m))
+
+function (m::Model)(xs...)
```
MikeInnes: I don't think this indirection is necessary here, we can just have `graphviz(m, xs...)`.

src/backend/backend.jl
ylxdzsw: Right, I first copied above code and then modified it as it do not require dynamic loading. This line is unfortunatly forgotten.

src/backend/graphviz/graph.jl
ylxdzsw: Yes, I will delete the importation.

src/backend/graphviz/graph.jl
ylxdzsw: I once saw this function but cannot remember the name anymore... I think this may imply that such a function is not as intuitive as simply `*(xs...)` in this case. Further more, `prod(())` returns 1 but in fact we want 0 here. `*(())` will throw error so no bugs will be hidden.

src/backend/graphviz/graph.jl
ylxdzsw: Actually I don't fully understand what Constant does... Can you give some hints about how to do it properly? `iconst` leave `Param` as is and cannot interp it any more, and directly overload `interp` for `interp(ctc, ::Constant{Param})` seems rather hacky.

src/backend/graphviz/graph.jl
ylxdzsw: We may have an optional field in `Param` and automatically set it in `@net`

src/backend/graphviz/graph.jl
ylxdzsw: Yes I tried, the main problem is the `iargs` middleware which do the so called eager evaluation and cause infinity recursion. To support cycle we need to set `ctx.cache` before arguments were interpreted, which is totally possible except for the size inference: for example, if we have `x = W * x`, the size of x is actually uninferable unless we introduce a very complicated inference system that can represent size with variable and try find a solution of all relations in those sizes, like the type system in Haskell or Ocaml.

src/backend/graphviz/graph.jl
ylxdzsw: I havn't found a way to integrate the shape.jl staff, but yes I can reuse at least the `infer` function.

src/backend/graphviz/model.jl
ylxdzsw: Right, I just (unecessarily) keep the same interface as other backends, but now I realize that it is better to use the same interface as `@shapes`

ylxdzsw:
Nesting is fine, you just got nested boxes like the Affine in the example.

I added the support of specifing depth, now this example:

```
using Flux
m = Chain(Chain(Input(2), Affine(2,2)), Affine(2,2), softmax)
open("some.dot", "w") do f; write(f, graphviz(m, [1 2; 3 4], depth=1)[1]) end
```

produce this image:

image

and this is how it looks when `depth = 0`:

image

I tried `@shapes` and it works well, but a little mess when play with nested model, for example, first line of the above model looks like this:

```
dugong = magpie((((Flux.Input{1}((2,)))((((DataFlow.Input())[1]::(2, 2),)::((2, 2),))[1]::(2, 2)),)::(nothing,))[1] * _::(2, 2), _::(1, 2))
```

like reading a crazy lisp one-liner :)

The misplace of the `Input` and activations node is a strange behaviour of graphviz. I carfully declare those nodes and their outgoing edges outside the following `subgraph`, but it still be place inside the box in the output image. Need somebody who knows graphviz better to help fixing this.

src/backend/backend.jl
```diff
@@ -20,8 +20,6 @@ function mxnet(m)
   eval(:(MX.mxnet($m)))
 end
 
-include(joinpath(dirname(@__FILE__), "graphviz/graphviz.jl"))
+include("graphviz/graphviz.jl")
 
-function graphviz(m)
-  GV.graphviz(m)
-end
+const graphviz = GV.graphviz
```
MikeInnes: `import GV: graphviz` will have the same effect

src/backend/graphviz/graph.jl
MikeInnes: Normally when we embed a value `f` in an `IVertex`, that value gets interpreted as a function call; so `f` with no args would be `f()`. `Constant` allows us to embed values in themselves, without interpreting them as calls.

You could fix this easily enough by renaming set of functions `interp(ctx, ::Union{Array,Param,...})` so that they don't overlap with `interp`.

src/backend/graphviz/graph.jl
MikeInnes: I don't think it necessarily has to be as complicated as that, as the cycle is always broken by an initial value, and realistically the shape of `x` will stabilise after one iteration (though I guess this isn't technically guaranteed). Still, you'd get the same effect by unrolling once, so it's not worth worrying about this for now.

src/backend/graphviz/graph.jl
MikeInnes: Yeah, I think it'd make sense to just call `infer` on any unknown functions; otherwise we end up with multiple versions of the same thing.

MikeInnes:
> like reading a crazy lisp one-liner :)

The best kind, no? :) Ok but yeah, that's kinda ugly – `shapes` definitely needs some improvements. I'd like to get rid of the redundant hints as a starting point.

Just thinking out loud, but I think the ideal design for this viz stuff would be to have generic graph-visualising tool as part of DataFlow.jl. Then you can plot the raw network with cycles and such, or plot the output of `@shapes` for shape inference, or unroll, etc, and everything is decoupled.

That said I don't think we should really highly prioritise this feature, so what's here is more than good enough for present purposes if it's useful to you.

ranjanan:
I think it is also worth separating out input shapes from models. Right now we have an `Input(N)` which comes as part of our model description. Perhaps it would be useful to have an input argument asking for a tuple of input sizes, like `graphviz(m, input_shape = (x,y)`. Since our model building at the moment seems to involve the user specifying fixed sizes for every layer, perhaps this could be implemented by generating a different closure that wraps Chain and takes in input sizes,  and plots the architecture given those input sizes.

MikeInnes:
We only really require it for models that need to know the size in order to initialise weights, e.g. `Affine`. You can happily write a stack of `Conv` layers without sizes being known immediately.

Not sure what alternatives there may be; can you elaborate on the closure idea?

ranjanan:
Oh I simply meant something like: 

```julia
wrap(N) = Chain(
  Input(N),
  Affine(N/2), relu,
  Affine( N/4), relu,
  Affine( 10), softmax)
```
and then `graphviz(wrap, input_shape = 784)`

ToucheSir:
This is très cool, but hails from a bygone era :slightly_smiling_face:. A modern re-imagining would probably be an external package that makes use of functors and `outputsize`.

darsnack:
Yeah I think there are various options. I know Javis.jl was being considered at one point. I would say close this for inactivity and open another PR with a newer approach if wanted.

DhairyaLGandhi:
I would keep this open so it's on our radar.

darsnack:
Maybe open a visualization issue then?

darsnack:
Unless someone wants to rebase this PR

ToucheSir:
This would need a complete rewrite instead of a rebase. Do we want to take a dependency on graphviz? If the only reason to keep this open is to have a tracking issue around visualization, a slightly renamed https://github.com/FluxML/Flux.jl/issues/1104 would be a better fit.

CarloLucibello:
I added a comment about this in #1104

PR #21, opened 2017-04-18
Breaking: Remove fake batching semantics

MikeInnes:
A design goal of mine is to be able to write expressions like `W*x+b` and have them work both for individual samples and batches. Doing this fully requires some kind of support in the backends which is a reasonably large project, so as a short-term hack to get the interface I wrote things in the batched form and added wrappers to all Flux's models that convert samples to batches automatically.

As it stands this hack is really giving us the worst of both worlds. The extra layer of magic in `@net` makes things harder to explain to beginners rather than easier, and involves an extra, unfamiliar concept for users of other frameworks. For those reasons I'm removing it completely, reverting to the approach used in other libs (write and run everything in batched form always).

The main downsides are (a) the inconvenience of having to `unsqueeze` or `batchone` samples when predicting, and (b) that re-introducing this may be another breaking change if people come to rely on untyped batching. Hopefully we can mitigate both of these by adding conveniences for typed batches and encouraging them as much as possible.

### DIFF
--- a/src/backend/mxnet/model.jl
+++ b/src/backend/mxnet/model.jl
@@ -1,5 +1,3 @@
-using Flux: runrawbatched
-
 struct AlterParam
   param
   load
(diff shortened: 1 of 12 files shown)


jekbradbury:
To me, Flux seems like the first framework that has the potential to fulfill what I see as the holy grail of deep learning frontends: it could eventually be one that lets you write code in a fully dynamic way that operates on individual samples, but compiles it to an optimized graph (with control flow ops if necessary) that runs on batches. This would be an even larger project than simply supporting both unbatched and batched math (Torch7 did that fairly well, although PyTorch is reverting to the conventional always-batched approach), but I think it's ultimately doable and would dramatically simplify the process of designing and implementing complex models.

[In a graph-based framework like MXNet or TensorFlow, a majority of code is boilerplate around the graph definition language. In a fully-imperative framework like Chainer or PyTorch (what I mostly code in, and essentially what Knet wants to be the Julia equivalent of), what you write is much closer to what you get--but if you're doing complicated things with NLP/text a huge chunk of code is now dedicated to wrapping the leaky abstraction of batches (with variable-length sequence objects etc.)]

I'm curious if you share this vision, because I think I want to work on a proof of concept in Flux sometime this summer. Maybe I'll use tf.contrib.fold.Loom as the backend depending on how hard it is to build something like it from scratch in Julia.

jekbradbury:
Update: I spoke too soon, and Chris Dyer/Graham Neubig/Yoav Goldberg just released [a version of DyNet](https://github.com/clab/dynet/blob/master/examples/python/tutorials/Autobatching.ipynb) which does what I was looking for...

MikeInnes:
Yes, that's definitely a core aim. Between Flux's programming model and metaprogramming-y approaches, I think there's very little boilerplate or tedious transformation that we can't abstract over. First-class recurrence is a start there, but we can go much further in RL and NLP. Would be especially cool if these things can be done as libraries over Flux, which I think is possible.

If you're interested in putting together a proof of concept then I'm keen to support that – just let me know what you need from my side.

PR #24, opened 2017-05-02
Don't extend base functions on base types

tkelman:
Please don't monkey-patch Base, that has global side effects

### DIFF
--- a/test/recurrent.jl
+++ b/test/recurrent.jl
@@ -13,5 +13,5 @@ end
   _, ys = apply(unroll1(r).model, xs, (r.y.x,))
   @test ys[1] == tanh(xs[1] * r.Wxy.x .+ r.y.x * r.Wyy.x .+ r.by.x)
   ru = unroll(r, 3)
-  ru(batchone(Seq(squeeze.(xs))))[1] == squeeze.(ys)
+  ru(batchone(Seq(squeeze.(xs, 1))))[1] == squeeze.(ys, 1)
 end
--- a/test/recurrent.jl
+++ b/test/recurrent.jl
@@ -13,5 +13,5 @@ end
   _, ys = apply(unroll1(r).model, xs, (r.y.x,))
   @test ys[1] == tanh(xs[1] * r.Wxy.x .+ r.y.x * r.Wyy.x .+ r.by.x)
   ru = unroll(r, 3)
-  ru(batchone(Seq(squeeze.(xs, 1))))[1] == squeeze.(ys, 1)
+  ru(batchone(Seq(squeeze.(xs))))[1] == squeeze.(ys)
 end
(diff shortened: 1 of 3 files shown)


tkelman:
@MikeInnes hello?

test/recurrent.jl
```diff
@@ -13,5 +13,5 @@ end
   _, ys = apply(unroll1(r).model, xs, (r.y.x,))
   @test ys[1] == tanh(xs[1] * r.Wxy.x .+ r.y.x * r.Wyy.x .+ r.by.x)
   ru = unroll(r, 3)
-  ru(batchone(Seq(squeeze.(xs))))[1] == squeeze.(ys)
+  ru(batchone(Seq((x->squeeze(x,1)).(xs))))[1] == (x->squeeze(x,1)).(ys)
```
MikeInnes: This should be `squeeze.(xs, 1)`

tkelman: of course, not sure why I didn't use that first

tkelman: done

MikeInnes:
I think a nicer solution would be to add this method to Base, to be honest.

tkelman:
That should be done via a PR to base (then backported via Compat once merged), not potentially concealing bugs in unrelated code in some package. And I don't think it makes much sense given the docstring is
```
  squeeze(A, dims)

  Remove the dimensions specified by dims from array A. Elements of dims must be unique and within the range
  1:ndims(A). size(A,i) must equal 1 for all i in dims.
```

Unless there are many of these dimension arguments that default to 1 in other functions that I'm not thinking of.

tkelman:
Could give this a local `squeeze1` name for the time being, or `Flux.squeeze` and not import it, to avoid side effects. Whatever you prefer that can avoid the type piracy.

tkelman:
Can we move on this? Changing the behavior of totally unrelated code depending on whether or not this package is imported is not okay.

tkelman:
Bump again, this would be incompatible with the proposed meaning in https://github.com/JuliaLang/julia/pull/22000, all the more reason it shouldn't be extended here.

MikeInnes:
Perhaps it would be better for Flux to have its own, unexported `squeeze` with these semantics so it can be imported by users.

tkelman:
Sure, that works too, done.

staticfloat:
@MikeInnes fwiw, a bunch of the discussion in the above linked issue basically boiled down to me realizing that for my use cases, what I really wanted was to just use `vec()`, which takes in things like Nx1 matrices and spits out N length vectors.  It's a little weird in that it can also take in NxN matrices and spit out (N^2) vectors, but in your use case that might not be a problem.

MikeInnes:
In this case it comes up because the first dim is a batch – but often you just want to work with one (N-d) sample, so it's convenient to wrap and unwrap that in a batch of size 1. So `batch` and `unbatch` do map to `[un]squeeze(xs, 1)` and there's no real way to get around that.

Of course, it'd be much nicer to have batching be first class in the type system rather than representing it implicitly like this, but that's another project.

PR #27, opened 2017-05-08
add more ops for tf backend

alha02:


MikeInnes:
Good stuff. It would be good to keep compatibility with Julia's `sum(xs)`; I think you can do this by defaulting `dim` to `nothing` but please check that.

Not essential, but it'd be a bonus to have the equivalent MXNet ops, if you have time.

The broadcasted `tanh` etc ops bother me a bit; we're going to need a more future-proof approach to that eventually. (But that's not something to worry about in this PR.)

alha02:
Ok great, I'll look into your comments and get back to you.

alha02:
Ok, so TensorFlow.jl did indeed handle the nothing-argument as expected not only for sum but for all the reduction ops, so I added them all.

About the MXNet-ops, I have a bunch of other ops for TF (which will probably not be available for MXNet) which I would like to add. Perhaps we can save the MXNet-ones for another PR? Let me know your thoughts.

MikeInnes:
Sure thing. I'm happy with this. Thanks for the patch!

PR #28, opened 2017-05-10
add more tf ops

alha02:
- [x] reshape
- [x] transpose
- [x] permutedims
- [x] tile
- [x] chol
- [x] size
- [x] cat
- [x] fill
- [x] slice
- [x] length
- [x] pad
- [x] cast
- [x] randu
- [x] randn
- [x] solve
- [x] triangular solve
- [x] inv
- [x] det
- [x] diag
- [x] diagm
- [x] svd
- [x] gather
- [x] expand_dims
- [ ] gather_nd
- [ ] segment_sum
- [ ] segment_prod
- [ ] segment_min
- [ ] segment_max
- [ ] segment_mean
- [ ] unsorted_segment_sum
- [ ] unsorted_segment_max
- [ ] select
- [ ] sort
- [ ] stack
- [ ] unstack
- [ ] squeeze
- [ ] range
- [ ] indmin
- [ ] indmax
- [ ] .==, .!=, .>, .<, .≥, .≤
- [ ] .&, .|
- [ ] round
- [ ] boolean_mask
- [ ] find
- [ ] log10
- [ ] unique
- [ ] squared_difference
- [ ] min
- [ ] max


alha02:
Hi Mike, I am trying to add the op pad but I can't seem to get it to work. I added it here:

```julia
for op in (*, .*, .+, .^, log, exp, ceil, floor, sqrt, abs, cos,
           sin, tan, atan, asin, acos, tanh, lgamma, erf, erfc, real, imag, conj,
           transpose, permutedims, size, pad)
  @eval graph(::typeof($op), args...) = $op(args...)
end
```

but when I try to use it I get this:
```julia
julia> @net f(x,y) = pad(x,y)
ERROR: UndefVarError: pad not defined
```

I tried exporting pad but it did not work either. Do you know what I am doing wrong?

MikeInnes:
Add a normal Julia version of `pad` to Flux itself and export it. Then it should be available in `Flux.TF`.

alha02:
Thanks, I got it working now. I wasn't sure where to put the Julia-versions of the ops, so let me know if you think it sits better somewhere else.

oxinabox:
I see you just added `cat` as mapping to `concat`
I think perhaps `cat` should map to `cat`.

`concat` follows the TensorFlow signature of `concat(values, axis)`.
`cat` follows the julia signature of `cat(axis, values...)`

 Once https://github.com/malmaud/TensorFlow.jl/pull/209
is merged it will also [follow the julia convention of adding dimensions til all inputs have same rank](https://github.com/malmaud/TensorFlow.jl/pull/209/files#diff-8e1fe1bc3944f9b6680ef47c61d672faR253)

On the other hand, Flux is no-doubt already handling different frameworks disagreeing on implementation of operations, so perhaps it is better to be tying to the lower level `concat` operation

alha02:
Good point. I agree with you that it should be mapping to cat, since I think we should aim to replicate Base-behaviour whenever we have the same name, and otherwise use a different name (tile vs repeat e.g.). Changing that.

alha02:
@MikeInnes I am running into trouble when using size. It works fine when using one argument:
```julia
julia> using Flux
julia> x = randn(4,3)
4×3 Array{Float64,2}:
  1.82025    0.237459   0.0788443
 -0.415831   0.446628   0.820253 
  1.86834    0.243848   1.26545  
  0.547149  -0.955323  -0.991229 
julia> @net f(x) = size(x)
julia> m = tf(f)
Flux.TF.Model(Flux.Capacitor(IVertex((DataFlow.Line("REPL[25]", 1))((size)((DataFlow.Input())[1])))), #undef)
julia> m(x)
2-element Array{Int32,1}:
 4
 3
```

But when trying to use two arguments it runs for a good amount of time and then I get a stackoverflow error:
```julia
julia> @net f(x,y) = size(x,y)
julia> m = tf(f)
Flux.TF.Model(Flux.Capacitor(IVertex((DataFlow.Line("REPL[28]", 1))((size)((DataFlow.Input())[1], (DataFlow.Input())[2])))), #undef)
julia> m(x,1)
ERROR: StackOverflowError:
Stacktrace:
 [1] getindex(::TensorFlow.Tensor{Int32}, ::TensorFlow.Tensor{Float32}) at /Users/alha02/.julia/v0.6/TensorFlow/src/ops/indexing.jl:0
 [2] getindex(::TensorFlow.Tensor{Int32}, ::TensorFlow.Tensor{Float32}) at /Users/alha02/.julia/v0.6/TensorFlow/src/ops/indexing.jl:139 (repeats 79999 times)
```

Do you know how I can fix that?

oxinabox:
Are you aware of the situation with `size`, `shape`, `get_shape` (and `length`, `endof`) in TensorFlow.jl?
You are probably correct that `size` is the operation you want.
`size` is an alias for `shape`, the  function that inserts a computational node the gives the dimensions at runtime (dynamic).
`size` also has a `size(X,dim)` wrapper, to make indexing syntax work

But in the C++ (and python) TensorFlow API, the `Size` operation is the number of elements. We call this operation `length` to use julia terminology. Which has `endof` as an alias for indexing syntax purposes.
But if you go back to older versions on TensorFlow.jl then this length operation is called `size`. I think (but am not sure) that it is bound to a nonexported `TensorFlow.size`  not to `Base.size`.
Sill not to worry about alpha versions of the package.


On the other hand, there is also `get_shape`. This is the static equivalent of `shape` (and `size`).
It can be found without evaluating the computational graph.
But some (up to all) of its fields may be unknown.
It is used to generate some graph parts statically eg `rnn`.
Though now with `TensorFlow..while` we can generate the dynamically eg `dynamic_rnn`.
It is also very useful for debugging purposes.

`get_shape` does not exist in the tensorflow C++ API.
It is implemented in Tensorflow.jl and I believe similarly in the python client.

Well that was a nice short essay. :-P

 - https://github.com/malmaud/TensorFlow.jl/blob/d575097ebb5253f2979066a19db7f1008d8292ab/src/ops/indexing.jl

 - https://github.com/malmaud/TensorFlow.jl/blob/d575097ebb5253f2979066a19db7f1008d8292ab/src/core.jl#L157-L180
 - https://github.com/malmaud/TensorFlow.jl/blob/d575097ebb5253f2979066a19db7f1008d8292ab/src/shape_inference.jl

oxinabox:
>@MikeInnes I am running into trouble when using size. It works fine when using one argument:
> ...
> But when trying to use two arguments it runs for a good amount of time and then I get a stackoverflow 
> ...

I just checked that this can **not** be reproduced in plain TensorFlow.jl
At least not on [oxinabox/master](https://github.com/oxinabox/TensorFlow.jl)
and I can't think that there is anything in my unstable master that is relevant that is not also in @malmaud's master.

```julia
julia> using TensorFlow

julia> sess = Session(Graph())
...

julia> x=rand(4,3)
4×3 Array{Float64,2}:
 0.473319  0.744055  0.818539
 0.103022  0.831621  0.811382
 0.530068  0.751047  0.58908
 0.276317  0.59295   0.643587


julia> run(sess, size(constant(x)))
2-element Array{Int32,1}:
 4
 3


julia> run(sess, size(constant(x),1)) #A julia constant type Int64
4


julia> run(sess, size(constant(x),constant(1))) #A tensorflow constant, type Tensor{Int64}
4

julia> d = Variable(1) #Variable is the other subclass of AbstractTensor
TensorFlow.Variable(<Operation 'node'>,<Operation 'node/Assign'>)

julia> run(sess, global_variables_initializer()

julia> run(sess, size(constant(x), d)) #A TensorFlow Variable
4
```


I am trying to think what is happening:
It is at 
https://github.com/malmaud/TensorFlow.jl/blob/d575097ebb5253f2979066a19db7f1008d8292ab/src/ops/indexing.jl

Now in particular:
```julia
function Base.getindex(params::AbstractTensor, inds::Vararg{AbstractTensor})
    getindex(params, map(Tensor, inds)...)
end
```
ends up calling itself.
That is the Fallback definition -- for if there are some nonTensor subclasses of `AbstractTensor` in the arguements.
The only such class is `Variable`.
Anyway:
So after running `Tensor` on all the `inds` it is still not finding a better `getindex` to hit.

oxinabox:
hah, got it, that was really obvious,
I should have checked the stack overflow signature.

```julia

julia> run(sess, size(constant(x),constant(1f0)))
ERROR: StackOverflowError:
 in getindex(::TensorFlow.Tensor{Int32}, ::TensorFlow.Tensor{Float32}) at /home/ubuntu/.julia/v0.5/TensorFlow/src/ops/indexing.jl:139 (repeats 80000 times)
```

So this is causes by https://github.com/MikeInnes/Flux.jl/blob/d131e312fc1e568811c303168a03eac091bed403/src/backend/tensorflow/model.jl#L38

Where all inputs to a TensorFlow model are converted to `Float32`
That is a problem for Flux to resolve I think.

It might be worth doing/possible to do something that causes a `MethorError` rather than a `StackOverflow` in TensorFlow.jl.
I'll have a think about it.

alha02:
Thanks! Nice spotting :) I'll look into if I can change the binding in some way to get around that, but I have a feeling that I might need some help on this one.

Anyway, about the `size` vs `get_shape:` For the way I would like to use the `size` op it needs to be bound to the dynamic one, but if you guys think it should be bound to the static one I can bind to something else. What do you think?

oxinabox:
yes, 
I think it should be bound to the dynamic one -- that is *why* `size=shape` not `size=get_shape`.

I just wanted to clarify those different ops;
Possibly that explanation needs adding to the Tensorflow.jl doc

src/backend/tensorflow/graph.jl
```diff
@@ -33,7 +33,7 @@ graph(::typeof(chol), args...) = TensorFlow.transpose(TensorFlow.cholesky(args..
 graph(::typeof(Flux.tile), args...) = TensorFlow.tile(args...)
 graph(::typeof(Flux.slice), args...) = TensorFlow.slice(args...)
 graph(::typeof(Flux.pad), args...) = TensorFlow.pad(args...)
-graph(::typeof(size), x, dim) = TensorFlow.size(x,TensorFlow.cast(dim,Int32))
+graph(::typeof(size), x, dim) = TensorFlow.size(x,TensorFlow.cast(TensorFlow.Tensor(dim),Int32))
```
oxinabox: We have `Base.convert` overloaded.
So `convert(Tensor{Int32}, dim)`
should work

alha02: Ah, ok. Thanks

MikeInnes:
This is looking great and I'll look over it in more detail soon. Just want to give a heads up that I've made some changes to dataflow.jl such that `f.(x)` reliably lowers to `graph(broadcast, f, x)`, so that will be the right way to handle anything that requires broadcasting in normal Julia code.

alha02:
Perfect, thanks. I'll try to find some time this weekend to finish up the tests and look into the segmented ops as well.

alha02:
Hi @MikeInnes, I want to add `cond` and `while_loop` as well, but I'm not sure how to make it work. Do you have any tips?

oxinabox:
>Hi @MikeInnes, I want to add cond and while_loop as well, but I'm not sure how to make it work. Do you have any tips?

Don't.
Get that merged,
then make `cond` and `while_loop` another PR.

They are going to require a lot more thought than anything else.
TensorFlow is primarily a Static graph generation framework.
Those ops are the ones that make for Dynamic graph generation.
They are primarily used for defining dynamic RNNs,
and Flux has its own way of defining RNNs, that I am not familar with, but that I assume is static?

alha02:
I see. Well my specific use case does not involve RNNs at all, I just need to add some control flow to my Flux models. But sure, it might be a good idea to separate them into a different PR.

alha02:
Hmm, thinking about this some more I might be able to express my problem in terms of a recurrent network, which removes the need for the while loop. But since the number of unrolls is fixed the `cond`-part seems harder to get around for me. In any case, it is definitely a problem for later.

oxinabox:
In tensorflow, we use `select` to effectively limit the number of unrolls
https://github.com/malmaud/TensorFlow.jl/blob/eb13ba9c4dffae3dbcbffadd8964b7ab1d91977f/src/ops/nn.jl#L70

alha02:
Cool, thanks. Would you say that `select` is addable here without too much of a hassle?

oxinabox:
yes

src/backend/tensorflow/graph.jl
```diff
@@ -42,7 +42,7 @@ graph(::typeof(randn), x) = TensorFlow.random_normal(convert(Tensor{Int32},x);dt
 graph(::typeof(solve), A, b) = TensorFlow.matrix_solve(A, b)
 graph(::typeof(triangular_solve), A, b) = TensorFlow.matrix_triangular_solve(A, b; lower=false)
 graph(::typeof(svd), x) = Ops.svd(x)
-graph(::typeof(Flux.expand_dims), x, dim) = TensorFlow.expand_dims(x,convert(Tensor{Int32},dim))
+graph(::typeof(Flux.expand_dims), x, dim) = Ops.expand_dims(x,convert(Tensor{Int32},dim))
```
oxinabox: Does `TensorFlow.expand_dims` not work anymore?
You should avoid using functions from the `Ops` module where possible.
Ones exported (or reexported) from the main namespace will be more idiomatic julia code and generally better supported.

alha02: Ah ok, changing back to TensorFlow. Thanks.

oxinabox:
Do you plan on merging this anytime soon?
You know you are allowed more than 1 PR.

alha02:
Haha fair enough, I'll do another PR for the remaining ones so we can get this merged.

MikeInnes:
As this has gotten fairly large, maybe the right thing is to carry on treating this as a "working" branch, but split out some of the stuff that's ready into smaller PRs that are ready for review. 

For example, there are a few additions which are just one line `graph` methods – like `svd`. These are easy to review and merge so it might be good to split those out first.

alha02:
Right. Ok, so what is the easiest way to split a PR?

oxinabox:
>Right. Ok, so what is the easiest way to split a PR?

git cherry-picking.
Basically lets you select commits from one branch and add them to another.

MikeInnes:
Right, so you can check out a new branch from master and run `git cherry-pick 2dfea4` (or whatever commit hash). The other thing that's hugely useful is `git rebase -i master` which will let you reorder and edit commits ([more info](https://git-scm.com/book/en/v2/Git-Branching-Rebasing)). If this is too much of a hassle don't worry about it, I can pick some commits out myself if need be, but this would speed things up and is a massively useful thing to get used to anyway.

MikeInnes:
I think we pulled most of this out into separate PRs, but feel free to open again if I missed anything. Any new TF stuff will need to be against the release-0.2 branch.

PR #29, opened 2017-05-11
Batch iterator

MikeInnes:
@ylxdzsw's work from #18.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -19,6 +19,7 @@ include("model.jl")
 include("dims/catmat.jl")
 include("dims/batching.jl")
 include("dims/seq.jl")
+include("dims/iter.jl")
 
 include("compiler/code.jl")
 include("compiler/loops.jl")
--- a/src/dims/catmat.jl
+++ b/src/dims/catmat.jl
@@ -15,7 +15,13 @@ size(b::CatMat) = (size(b.data, 1),)
 
 getindex(b::CatMat, i)::eltype(b) = slicedim(b.data, 1, i)
 
-setindex!(b::CatMat, v, i) = b.data[i, :] = v
+setindex!(b::CatMat, v, i::Integer) = b.data[i, :] = v
+
+function setindex!(b::CatMat, xs, ::Colon)
+  for (i, x) in enumerate(xs)
+    b[i] = x
+  end
+end
 
 allequal(xs) = all(x -> x == first(xs), xs)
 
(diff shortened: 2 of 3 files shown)


MikeInnes:
I had a go at refactoring this a bit; I think that factoring out the stateful iteration logic makes this a bit cleaner, as well as more generic in future (e.g. so we can use it for `Seq`s as well as `Batch`s; or indeed any `CatMat` type).

For now I haven't tried to make this work with tuples; before, it worked, but only with tuples nested to one level.

I've also removed the warnings; I think it's better to just document this behaviour (same as `partition` in base).

@ylxdzsw let me know if this can be improved, or if I've missed anything important that was there before.

ylxdzsw:
Yes, it's much cleaner now. Thanks.

One of my purpose of introducing the `buf` field in the `Batched` rather than construct a new one on each iteration is to avoid temporary arrays, which harms the performance a lot in current GC implementation. The `taken!(s::StatefulIter, n::Integer)` seems spoil it. However, these time may be ignored anyway compared with the NNs, so it's just fine.

Tuple input could still be implemented with `mapt`. I can try it with another PR later.

Sorry I said I would split these PRs myself last weekend but I got a little busy on my graduation project (and will last one or two more weeks)

oxinabox:
Not to disparage this work.
But I do wonder if building up from @Evizero 's [JuliaML/MLDataPattern](https://github.com/JuliaML/MLDataPattern.jl) would lead to a more featureful solution, easier.

The main reason I can imagine not to is if there are special requirements to allow making Batching (etc)
not happen in julia code, but instead happen on the backend (and thus benefit from doing things like loading each batch directly into GPU RAM).
Though I thinking maybe that can be done by dispatching on the MLDataPattern types.

Thoughts?
MLDataPattern is really featureful, and has a very nice interface.

MikeInnes:
@ylxdzsw No problem, consider this me making up for the rebases I put you through :) GSoC hasn't even started yet so your graduation project should certainly be the priority.

I think that the small intermediate arrays here are much less likely to have a performance impact compared to allocating batches. But we can always test that and patch it as necessary.

@oxinabox Thanks for the ref, that'll be worth looking into. One of the main reasons to do things in a custom way is to take advantage of things like the `Batch` types; but maybe we can do that with MLDataUtils or get them on board.

The other thing is that APIs and things are still fluctuating a lot, so it's nice to be able to experiment quickly; as things settle down I certainly see us splitting more out and encouraging more integration across packages.

MikeInnes:
@ylxdzsw thanks for the patch! It would be an interesting challenge to make this work generically across CatMat types.

datnamer:
@MikeInnes What's the way to train from a generator function in the current version of flux?

PR #38, opened 2017-05-26
Fix activation functions

staticfloat:
Your activation function backprops were oh so close


MikeInnes:
Well, there's a rookie mistake if there ever was one. Thank you, wise master.

PR #41, opened 2017-06-08
Add tf ops: svd, inv, det

alha02:
This PR adds the following TF ops:

- [X] svd
- [X] inv
- [X] det

### DIFF
--- a/test/backend/tensorflow.jl
+++ b/test/backend/tensorflow.jl
@@ -25,28 +25,26 @@ end
 
 @testset "Ops" begin
 
-error_margin = 1e-4
+  @testset "svd" begin
+    A = convert(Array{Float32},randn(5,5))
+    @net f(x) = svd(x)
+    m = tf(f)
+    u,s,v = m(A)
+    @test A ≈ u*diagm(s)*transpose(v)
+  end
 
-#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-# svd
-A = convert(Array{Float32},randn(5,5))
-@net f(x) = svd(x)
-m = tf(f)
-u,s,v = m(A)
-@test maximum(abs.(u*diagm(s)*transpose(v) - A)) < error_margin
+  @testset "inv" begin
+    @net f(x) = inv(x)
+    m = tf(f)
+    @test m(A) ≈ inv(A)
+  end
 
-#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-# inv
-@net f(x) = inv(x)
-m = tf(f)
-@test maximum(abs.(m(A)-inv(A))) < error_margin
+  @testset "det" begin
+    @net f(x) = det(x)
+    m = tf(f)
+    @test m(A) ≈ det(A)
+  end
 
-#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-# det
-@net f(x) = det(x)
-m = tf(f)
-@test maximum(abs.(m(A)-det(A))) < error_margin
-
-#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+end
 
 end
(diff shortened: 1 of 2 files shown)


test/backend/tensorflow.jl
```diff
@@ -23,4 +23,30 @@ test_anon(tf)
   @test run(sess, Y, Dict(X=>xs)) ≈ d(xs)
 end
 
+@testset "Ops" begin
+
```
oxinabox: You are missing the indent for the `@testset` / `begin` block.
Also you are allowed to nest testsets
If you want -- it will even generate good outputs during test execution/failure.
idk what @MikeInnes's opinion is, I personally think it is nice and self documenting.

So this code would become

```julia
@testset "Ops" begin
    error_margin = 1e4

    @testset "svd" begin
        A = convert(Arra #...
        #...
     end

    @testset "Inv" #...
    #...
end
```

test/backend/tensorflow.jl
```diff
@@ -23,4 +23,30 @@ test_anon(tf)
...
+error_margin = 1e-4
+
+#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+# svd
+A = convert(Array{Float32},randn(5,5))
+@net f(x) = svd(x)
+m = tf(f)
+u,s,v = m(A)
+@test maximum(abs.(u*diagm(s)*transpose(v) - A)) < error_margin
```
oxinabox: Just thinking,
I think this could be written with  `@test all(isapprox.(A, u*diagm(s)*transpose(v)); atol=error_margin)`
But I think your way using `maximum(abs.(...`
is nicer.

I wonder if a method for `all`ing over elementwise approximate matrix value testing should be added to `Base`.

test/backend/tensorflow.jl
alha02: It definitely seems like a good idea to me. I'll add that for now and we can change it if Mike has a different opinion.

test/backend/tensorflow.jl
alha02: Sure, I didn't know of the function `isapprox`. I like that it has defaults based on the dataype, and it will also make things look nice and clean.

MikeInnes: There's also `@test a ≈ b`, if that's useful

MikeInnes: Oh, you already got that though :)

MikeInnes:
Thanks a lot for splitting this out! FWIW, I don't think we should worry *too* much about checking for errors that would be TensorFlow's responsibility (like numerical accuracy) – the priority on our side is only to make sure we're calling it correctly. These are fine but just noting in general.

test/backend/tensorflow.jl
```diff
@@ -23,4 +23,28 @@ test_anon(tf)
...
+    @net f(x) = svd(x)
+    m = tf(f)
+    u,s,v = m(A)
+    @test A ≈ u*diagm(s)*transpose(v)
+  end
+
+  @testset "inv" begin
+    @net f(x) = inv(x)
+    m = tf(f)
+    @test m(A) ≈ inv(A)
```
MikeInnes: If you're going to reuse `A` outside of the `svd` testset, best to define it outside of it too.

I'd also shorten these to `@test tf(@net x -> inv(A))(A) ≈ inv(A)`. Nested test sets are good but perhaps don't have as much benefit if there's only one `@test`.

alha02: Thanks for the input. I removed the nested test sets and shortened down the tests as well.

test/backend/tensorflow.jl
```diff
@@ -24,27 +24,11 @@ test_anon(tf)
...
-    @test m(A) ≈ inv(A)
-  end
-
-  @testset "det" begin
-    @net f(x) = det(x)
-    m = tf(f)
-    @test m(A) ≈ det(A)
-  end
-
+  A = convert(Array{Float32},randn(5,5))
```
oxinabox: For future reference: this can be written: `A = randn(Float32, (5,5))`

alha02: Thanks, I'll update that as well :)

MikeInnes:
Thanks for putting up with the little issues – this is flawless now. Thanks!

alha02:
It is no problem at all. I'm learning stuff, so I am perfectly happy :)

Get Outlook for iOS<https://aka.ms/o0ukef>
_____________________________

PR #44, opened 2017-06-12
training julia models

ylxdzsw:
This PR fixes bugs of `back!` in core.jl (the fallback method) and `Chain`, and add the support of `Affine`. With these we can run the training process of some basic linear models like LogisticRegression on our pure Julia implementation. This allow us to test training process and optimizers before we can run them on backends.


src/layers/affine.jl
```diff
@@ -9,3 +9,16 @@ Affine(in::Integer, out::Integer; init = initn) =
...
+  W.Δx[:] = x' * Δ
+  b.Δx[:] = sum(Δ, 1)
+  Δ * W.x'
+end
+
+function update!(m::Affine, η)
+  update!(m.W, η)
+  update!(m.b, η)
+  m
+end
```
MikeInnes: would be good to have the newline here

src/layers/control.jl
```diff
@@ -7,9 +7,25 @@ end
...
-back!(s::Chain, Δ) = foldr((m, Δ) -> back!(m, Δ), Δ, s.layers)
 update!(s::Chain, η) = foreach(l -> update!(l, η), s.layers)
 
+function back!(s::Chain, Δ, xs...)
+  crumbs = Tuple[xs]
+  N = length(s.layers)
+
+  for i in 1:N-1
+    xs = s.layers[i](xs...)
+    xs isa Tuple || (xs = (xs, ))
```
MikeInnes: we can assume that all layers in a chain are single input/output. Then it should be easy to do this as `fold` agains as well.

src/layers/affine.jl
ylxdzsw: Yes, usually I use [editorconfig](http://editorconfig.org/) for trimming trailling white spaces and insert new line automatically. It's simple and works on almost all editors and platforms. Would you mind me adding one for Flux?

src/layers/control.jl
ylxdzsw: OK, that would make things easier.

src/layers/affine.jl
MikeInnes: Better to just keep editor config locally, I think.

MikeInnes:
Great, can you add a test?

MikeInnes:
Thanks!

src/layers/affine.jl
MikeInnes: By the way, this method is already output by `@net`, but may be incorrect. Could you fix that and remove this version?

PR #48, opened 2017-06-23
Make Flux.onehot accessible in the MNIST example

peterjdolan:
Alternatively, line 4 could be rewritten to

data = [(trainfeatures(i), Flux.onehot(trainlabel(i), 0:9)) for i = 1:60_000]


MikeInnes:
Awesome, thanks :+1:

PR #50, opened 2017-07-08
Add more tf ops

alha02:
- reshape
- tile
- fill
- cast
- solve
- triangular_solve
- randu
- randn
- expand_dims

### DIFF
--- a/src/ops.jl
+++ b/src/ops.jl
@@ -1,6 +1,6 @@
-export reshape, tile, fill, cast, solve, triangular_solve, randu
+export reshape, tile, fill, cast, solve, triangular_solve, randu, randn
 
-import Base: reshape, fill
+import Base: reshape, fill, randn
 
 reshape(x::AbstractArray, dims::AbstractArray) = reshape(x,tuple(dims...))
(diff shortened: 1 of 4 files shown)


MikeInnes:
Thanks!

src/ops.jl
```diff
@@ -0,0 +1,18 @@
+export reshape, tile, fill, cast, solve, triangular_solve, randu, randn,
+       expand_dims
+
+import Base: reshape, fill, randn
+
+reshape(x::AbstractArray, dims::AbstractArray) = reshape(x,tuple(dims...))
```
tkelman: can this use a package-local name? it's encouraged not to extend Base functions on Base types, as that has global side effects and can change the behavior of unrelated code

tkelman: oh you got rid of this in https://github.com/MikeInnes/Flux.jl/commit/bfe1f8eeeece5349cdfcb632b2414d187dc7bb79 but that isn't included in your latest tag?

MikeInnes: Yeah, this isn't ideal but it's just needed as a stopgap. I'm doing a big refactor that'll get rid of the patching in the next version.

tkelman: k fair enough

PR #52, opened 2017-07-21
implement the training callback with `throttle` function

ylxdzsw:


MikeInnes:
Great, can you add a couple of tests for the implemented options? Can just be for `throttle` rather than the training process.

Given how you're using `later`, could that just be a bool?

ylxdzsw:
`later` also holds args and kwargs, but actually we don't pass anything. Should we just assume f has no arguments?

MikeInnes:
When you call `later`, `args` and `kwargs` are in scope, so you can just use them.

ylxdzsw:
When trailling option is enabled and a successive call happens more frequently than timeout with different args, there will be only one Task created at the first call in line 50, and the args catched by the closure is the args of the first call, but it should be the ones of the last call.

MikeInnes:
Ah, I see what you mean. That's the sort of thing that's great to have a test for so I don't break it in future :)

PR #55, opened 2017-08-28
add a few gradients and layers

ylxdzsw:
This adds a few recurrent layers and gradient calculation of `cat`.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -43,3 +43,19 @@ function Base.show(io::IO, l::Linear)
   l.σ == identity || print(io, ", ", l.σ)
   print(io, ")")
 end
+
+# Embedding
+
+struct Embedding
+    mat
+end
+
+Embedding(a::Integer, b::Integer; init=initn) =
+    Embedding(track(init(a, b)))
+
+Optimise.children(m::Embedding) = m.mat,
+
+(m::Embedding)(x) = m.mat[x, :]
+
+Base.show(io::IO, m::LSTM) =
+  print(io, "Embedding(", size(m.W, 1), ", ", size(m.W, 2), ')')
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -57,5 +57,5 @@ Optimise.children(m::Embedding) = m.mat,
 
 (m::Embedding)(x) = m.mat[x, :]
 
-Base.show(io::IO, m::LSTM) =
+Base.show(io::IO, m::Embedding) =
   print(io, "Embedding(", size(m.W, 1), ", ", size(m.W, 2), ')')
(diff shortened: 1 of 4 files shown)


MikeInnes:
Looks good. Can you add a grad check for cat?

src/layers/basic.jl
```diff
@@ -43,3 +43,19 @@ function Base.show(io::IO, l::Linear)
...
+# Embedding
+
+struct Embedding
+    mat
+end
+
+Embedding(a::Integer, b::Integer; init=initn) =
+    Embedding(track(init(a, b)))
+
+Optimise.children(m::Embedding) = m.mat,
```
MikeInnes: Does this parse correctly? I'm generally not that convinced of the need for Embedding given that we can just do `n -> w[n,:]` (or better, define a OneHot vector type and overload matmul to do this).

src/layers/recurrent.jl
```diff
@@ -0,0 +1,86 @@
...
+  W::S
+  b::T
+end
+
+RNN(in::Integer, out::Integer; init = initn) =
+  RNN(track(init(in+out, out)), track(zeros(1, out)))
+
+Optimise.children(m::RNN) = (m.W, m.b)
+
+(m::RNN)(x, h) = σ.([x h] * m.W .+ m.b)
```
MikeInnes: We should consistently use batch-last

src/layers/recurrent.jl
```diff
@@ -0,0 +1,86 @@
...
+init_state(m::RNN, batchsize::Integer, init=zeros) =
+  init(batchsize, size(m.W, 2))
+
+# GRU
+
+struct GRU{TWi, TWh, Tbi, Tbh}
+  Wi::TWi
+  Wh::TWh
+  bi::Tbi
+  bh::Tbh
```
MikeInnes: Share types where you can. I used two parameters in `Linear` only because there was a matrix and vector, but these are all matrices and can be the same type.

src/tracker/lib.jl
```diff
@@ -39,6 +39,19 @@ function back!(::typeof(*), Δ, a::AbstractMatrix, b::AbstractVecOrMat)
...
+tryfind(e, c, ft, ff) = (i = findfirst(c, e); i != 0 ? ft(i) : ff(i); i)
+
+function back!(::typeof(cat), Δ, dims, xs...)
+  acc = map(zero, dims)
+  for (i, x) in enumerate(xs)
+    slice = map(d->size(x, d), dims)
+    @back!(x, view(Δ, [tryfind(d, dims, i->1+acc[i]:slice[i], i->:) for d in 1:ndims(x)]...))
+    acc = acc .+ slice
+  end
+end
```
MikeInnes: Where do we catch the forward pass for this?

MikeInnes:
This is riddled with typos and other issues, and it doesn't seem like you've tried to test it or get it working at all – will have a go myself.

PR #60, opened 2017-09-11
Fix link to basics in docs

mkborregaard:
It's nice if the link on the front page of the docs works when you announce the new release on discourse :-)
Looking forward to digging into the new version!

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -11,4 +11,4 @@ Pkg.add("Flux")
 Pkg.test("Flux") # Check things installed correctly
 ```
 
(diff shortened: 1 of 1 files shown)


MikeInnes:
Indeed it would -- thanks!

mkborregaard:
Gotta say, I'm making some weighty PRs on this package.
BTW, the link to "training" on the "basics" page is broken as well - should be "training/training.html", not "models/training.html".

MikeInnes:
Documenter is doing some shitty link rewriting, so a bunch of correct links were broken. They should be fixed now.

PR #62, opened 2017-09-12
Minor RNNCell docs fix

JobJob:
This is what you meant right? It matches the preceding example.

Really liking the look of this new version btw!

### DIFF
--- a/docs/src/models/recurrence.md
+++ b/docs/src/models/recurrence.md
@@ -50,7 +50,7 @@ We sometimes refer to functions like `rnn` above, which explicitly manage state,
 ```julia
 using Flux
 
-m = Flux.RNNCell(10, 5)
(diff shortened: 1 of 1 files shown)


JobJob:
Oh actually, that gives a `invalid redefinition of constant rnn` - since the rnn function is defined above.

Maybe it should be `rnn2` or `rnn_cell` or `rnncell`? or change the bottom function to `m`? Decisions are the worst.

MikeInnes:
Yeah, that was why I changed it from `rnn`. I'd be fine with `rnn2`, or just changing the function call to `m`.

JobJob:
I chose rnn2 - to be most similar to the prev example, squash and merge? 
If you need me to rebase and/or squash myself, lemme know.

MikeInnes:
Thanks!

PR #63, opened 2017-09-12
Enables passing an activation function to RNN/RNNCell

JobJob:
Also, fixes it not using the `init` function provided


MikeInnes:
Thanks! This should definitely be using `tanh` by default as well, so I'll add that in.

PR #68, opened 2017-09-21
Fixed optimisers not applying gradient to parameters

GenaBitu:
Hi,
the more advanced optimisers (all except SGD) weren't applying the computed gradient to the parameters. Such behaviour seems rather confusing to me, so I changed it so that all optimisers behave in the same way.

Thank you,
GenaBitu

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -1,4 +1,4 @@
-function updateParam(p::Param)
+function applyGradient(p::Param)
     p.x .-= p.Δ
     p.Δ .= 0
 end
@@ -6,7 +6,7 @@ end
 function descent(p::Param, η::Real)
   function ()
 	p.Δ .= p.Δ .* η
-	updateParam(p)
+	applyGradient(p)
   end
 end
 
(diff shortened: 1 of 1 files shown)


MikeInnes:
Thanks a lot for the patch. The original idea here was that the functions in `optimisers.jl` would serve as building blocks, which you can combine together to form a real optimiser in `interface.jl`. For example you might define ADAM like this (following [Keras' optimisers](https://keras.io/optimizers/)):

```julia
ADAM(ps, η = 0.001, β1 = 0.9, β2 = 0.999, ϵ = 1e-08, decay = 0.0) =
  optimiser(ps, p -> adam(p; ...), p -> invdecay(p, decay), p -> descent(p, 1))
```

So as a user you'd construct `ADAM(ps)` and the `descent` call would actually apply the gradient.

Do you think you could try working things into this form? Let me know if that's not clear at all.

pevnak:
The construction as Mike has suggested works nicely. I would stick with it and add the needed API to it.

MikeInnes:
Thanks a lot for giving this a go -- but for now I'll close this in favour of #70.

GenaBitu:
Agreed, that's a better solution given the original intent. I am just wondering whether `descent` is the right name. That's really minor though...

PR #70, opened 2017-09-26
Pull request/c91af25c

pevnak:
This exposes the optimisers as Mike has suggested. 
It also improves handling multiplication of sparse and dense matrices, as was suggested by Sacha.

### DIFF
--- a/src/tracker/Tracker.jl
+++ b/src/tracker/Tracker.jl
@@ -69,6 +69,7 @@ end
 
 include("back.jl")
 include("lib.jl")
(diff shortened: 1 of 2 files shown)


MikeInnes:
This is great, thanks! I took the optimiser changes in https://github.com/FluxML/Flux.jl/commit/bfcc1ac25d11c5c3e04ddc1ea6c59b7b8efd968d.

As for the sparse changes – it seems like most sparse matmul code could benefit from those changes, would it make sense to submit a PR to Base for those?

PR #71, opened 2017-09-27
Fixed the cross entropy in stateless.jl

pevnak:


MikeInnes:
Perhaps `logloss` should just be called `crossentropy`, and this thing `logcrossentropy`?

As a general thought, I'm not huge on the idea of having loads of `log*` functions; perhaps there's a way to express this at the type level?

pevnak:
Yes,
I totally understand you point, though I have no idea how to do this.

MikeInnes:
Ok, I merged this in https://github.com/FluxML/Flux.jl/commit/4aa7741ba9b291d9846d55dc520b6982a6dd1aae. I took out the onehot stuff since I think that can be explicit and I'd like this to work on vectors; let me know if I missed anything though.

PR #74, opened 2017-10-03
actually use init argument in LSTMCell

### LINKED ISSUE #73: `init` argument in `LSTMCell` unused
I was browsing the code and I noticed that while the `LSTMCell` constructor takes an `init` argument, all of the layers are initialized with `initn()` instead of `init()`: https://github.com/FluxML/Flux.jl/blob/a60a754d68df79aa6e19f7f7b3992164d47703b0/src/layers/recurrent.jl#L65-L68

kleinschmidt:
fixes #73


MikeInnes:
:+1: Thanks!

PR #76, opened 2017-10-06
Added an AbstractLayer type

GenaBitu:
Recently I needed to operate on layers , so I added an abstract superclass for all of them. I am not particularly sure about the recurrent layers though.

Don't know how much of a fan of interfaces you are, though :)


MikeInnes:
Can you elaborate on your use case for this a bit?

I used to do something like this via `Flux.Model`. It's nice in some ways, but unfortunately it's only useful if *all* layers (including user-built ones) buy into it. I'm generally wary on having subtyping as part of a user interface, both because understanding of the type system is relatively advanced and because it makes it harder to have a single type implement multiple interfaces.

GenaBitu:
I am trying to implement embedded-space multi-instance learning in Flux. I've implemented that with a `MILlayer` which takes in a bag of feature vectors, applies some function to each, aggregates them and returns a single feature vector representing the whole bag. The "some function" part is another neural net which means that my `MILlayer` takes a nearal net as a parameter

I could make it take a `Chain` but then you'd have to make a `Chain` from a single layer if you wanted just one. I could make it take a `Union` of `Chain` and all the other possible layers, but that seems to me like a road to hell. So the only options I see is to implement it with no type checking whatsoever or the `AbstractLayer` (or `Flux.Model`...). I think each has its pros and cons. i come from C++ background so I like to have my types clear. I agree that it imposes some burden (although I don't think the type system is that advanced :) ) on the user of the library but IMHO the benefits (safer code) outweigh that.

I think this is the ever-present "to type or not to type" debate, I have my opinion, but the decision is (obivously) yours.

Thank you either way.

MikeInnes:
Ok, I see. Unfortunately I think the only consistent option here, given that Julia can't reason about arrow types, is to have no typing at all. If you have any typing then you are forced to reject perfectly valid models (for example, `identity` or `Dense(5, 2) ∘ Dense(10, 5)`).

On the other hand, I don't think the typing would add a huge amount in this case. Instead of throwing a method error it will just throw a "not callable" error as soon as you try to invoke the model, which ends up roughly equivalent.

GenaBitu:
Yeah, that makes sense... In that case, not doing any type checking is probably the right call.
Closing the PR.

MikeInnes:
Thanks for the contribution and discussion nonetheless!

PR #78, opened 2017-10-10
adding layer normalization

skariel:


src/layers/stateless.jl
```diff
@@ -5,3 +5,10 @@ mse(ŷ, y) = sum((ŷ .- y).^2)/length(y)
 
 logloss(ŷ, y) = -sum(y .* log.(ŷ)) / size(y, 2)
 # back!(::typeof(logloss), Δ, ŷ, y) = 0 .- Δ .* y ./ ŷ
+
+# normalization
+
+# layer normaliztion as in https://arxiv.org/abs/1607.06450
+_std(y) = sqrt.(sum((y.-mean(y)).^2) ./ (length(y)-1))
+layernormalization(y) = (y .- mean(y)) ./ _std(y)
```
ilkerkesen: Is it OK to call mean and broadcast kernel twice? I'd rather call it once, then use the value twice.

jekbradbury: definitely try to call it just once

skariel:
ready for 2nd review

src/layers/stateless.jl
skariel: fixed in latest commit

src/layers/stateless.jl
```diff
@@ -5,3 +5,14 @@ mse(ŷ, y) = sum((ŷ .- y).^2)/length(y)
 
 logloss(ŷ, y) = -sum(y .* log.(ŷ)) / size(y, 2)
 # back!(::typeof(logloss), Δ, ŷ, y) = 0 .- Δ .* y ./ ŷ
+
+# normalization
+
+# layer normaliztion as in https://arxiv.org/abs/1607.06450
```
CarloLucibello: typo normaliztion

CarloLucibello:
This is missing the rescaling parameters \alpha and \beta, as in equations (17) and (17) of the paper
https://arxiv.org/abs/1607.06450

So the proper way to do this is to define the type LayerNormalization (or LayerNorm maybe) with params \alpha and \beta, along the line of the other layers defined in Flux

skariel:
hi @CarloLucibello,

I wonder about how to implement this -- in eq. 16 alpha and beta are learnable vectors and the operations are element-wise. This would make the layernormalization state-full. Also it introduces many cases like for e.g. I use alpha=constant unity and beta=constant zero.

On the other hand, this could be implemented as a learnable linear activation layer, then one could combine the activation layer with it and get the same result. And I could just use the layer as it is, without further combinations. And if someone wants to use a learnable sigmoid, he could use it instead. Seems to me this would give better performance and flexibility, while keeping the interface complexity at a minimum

What do you think?



The same effect could be implemented as a separate statefull "Linear" layer. Keras and

CarloLucibello:
The composition you are suggesting is not the same as applying a fully connected layer after your layernormalization function, it should be done by a more strict (diagonal) affine transform. On a side note maybe it is worth to have such an elementwise affine Layer. 

If you don't want to get involved with that, maybe you can merge this pr as it is (provided @MikeInnes agrees with it), and open an issue for the creation of a normalization layer with learnable alpha and beta but also some nice interface for setting them to consts.

I was planning to implement a batch normalization layer in the next days: since they are similar maybe I can take care of both if I manage to. 

Can you add a docstring to `layernormalization`?

skariel:
So this is what I think of doing now:

1) add a constant (non-learnabe) alpha and beta to the current layer normalization, keeping it stateless (This should not affect performance since the linear transformation is done only on the mean and std parameters once)

2) add an affine (element-wise) learnable linear layer

This should cover all practical use-cases

skariel:
done, ready for review!

src/utils.jl
```diff
@@ -1,6 +1,8 @@
 # Arrays
 
 initn(dims...) = randn(dims...)/100
+initones(dims...) = ones(dims...)
+initzeros(dims...) = zeros(dims...)
```
CarloLucibello: do we need these? Why not use `ones` and `zeros`? 
I would also get rid of `ìnitn` and it's arbitrary chosen constant 100, in favour of somethin like the xavier method of knet (not in this PR though)

CarloLucibello: do we need this? Why not use directly zeros and ones?

skariel: yep, I agree with you. Basically I just wanted to be consistent with `initn`. Changing now...

skariel:
removed `initzeros` and `initones`

MikeInnes:
I couldn't edit this PR so I made some tweaks in #95. Thanks a lot @skariel for the patch and others for reviewing, would appreciate a look over my changes to make sure I haven't removed anything useful by accident.

PR #82, opened 2017-10-12
construct TrackedScalar with params(num)

CarloLucibello:


MikeInnes:
The risk of this is that it makes it easier to accidentally construct an `Array` in a model that uses `CuArray`, for example. It's probably worth it given that it's common to construct stuff with `Array` and then convert.

To answer your question, we definitely do want that behaviour, because you can't store a derivative in an integer array – without this conversion you get errors on the backward pass.

CarloLucibello:
>The risk of this is that it makes it easier to accidentally construct an Array in a model that uses CuArray, for example. It's probably worth it given that it's common to construct stuff with Array and then convert.

what about having a TrackedScalar type that is not implemented as a zero-dimensional array?

MikeInnes:
Yeah, that may well make sense eventually. It's just a bit of extra complexity so I'm waiting to see if this approach causes issues.

PR #83, opened 2017-10-17
utils: update `accuracy` for classification task

iblis17:


iblis17:
BTW, can we import model-zoo as a submodule? then we can freeze the correspond version of model-zoo

src/utils.jl
```diff
@@ -28,15 +28,31 @@ end
...
-  n = 0
-  correct = 0
-  for (x, y) in data
-    x, y = tobatch.((x, y))
-    n += size(x, 1)
-    correct += sum(argmax(m(x)) .== argmax(y))
-  end
-  return correct/n
+"""
+    accuracy(m, pred, label)
```
CarloLucibello: outclass missing

MikeInnes:
So I think the original idea was this was for it to work with a dataset exactly like you'd give `train!`, to make things convenient in that case.

That said I'm wondering if Flux should really include this at all... it might just be confusing to people if it doesn't apply to their specific dataset / setup.

iblis17:
I'm waiting for [MLMetrics](https://github.com/JuliaML/MLMetrics.jl) getting released.
And I'm comfortable with my MNIST model now: https://github.com/FluxML/model-zoo/pull/7#issuecomment-339611574

PR #84, opened 2017-10-17
layer: implement BatchNorm layer

iblis17:
Please review it carefully, no test cases :p 

I just plug this layer to the MNIST example, and set the learning rate of SGD to a crazy number: `10`.
It performs quite fine and better than the original model.

---
Ref: [Batch Normalization: Accelerating Deep Network Training by Reducing Internal Covariate Shift](https://arxiv.org/pdf/1502.03167.pdf)

### DIFF
--- a/src/layers/normalisation.jl
+++ b/src/layers/normalisation.jl
@@ -43,3 +43,72 @@ function (a::Dropout)(x)
 end
 
 _testmode!(a::Dropout, test) = (a.active = !test)
+
+"""
+    BatchNorm(dims...; λ = identity,
+              initβ = zeros, initγ = ones, ϵ = 1e-8, momentum = .1)
+
+Batch Normalization Layer
+
+See [Batch Normalization: Accelerating Deep Network Training by Reducing
+     Internal Covariate Shift](https://arxiv.org/pdf/1502.03167.pdf)
+
+In the example of MNIST,
+in order to normalize the input of other layer,
+put the `BatchNorm` layer before activation function.
+
+```julia
+julia> m = Chain(
+  Dense(28^2, 64),
+  BatchNorm(64, λ = relu),
+  Dense(64, 10),
+  BatchNorm(10),
+  softmax)
+Chain(Dense(784, 64), BatchNorm(64, λ = NNlib.relu), Dense(64, 10), BatchNorm(10), NNlib.softmax)
+```
+"""
+mutable struct BatchNorm{F,V}
+  λ::F  # activation function
+  β::V  # bias
+  γ::V  # scale
+  μ     # moving mean
+  σ     # moving std
+  ϵ::Float64
+  momentum::Float64
+  active::Bool
+end
+
+BatchNorm(dims::Integer...; λ = identity,
+          initβ = zeros, initγ = ones, ϵ = 1e-8, momentum = .1) =
+  BatchNorm(λ, param(initβ(dims)), param(initγ(dims)), 0., 1., momentum, ϵ, true)
+
+function (BN::BatchNorm)(x)
+  if !BN.active
+    μ = BN.μ
+    σ = BN.σ
+  else
+    m = size(x, 2)  # batch size
+    μ = sum(x, 2) ./ m
+    σ = sqrt.(sum((x .- μ).^2, 2) ./ (m - 1) .+ BN.ϵ)
+
+    # update moving mean/std
+    mtm = BN.momentum
+    BN.μ = mtm .* μ.data .+ (1 - mtm) .* BN.μ
+    BN.σ = mtm .* σ.data .+ (1 - mtm) .* BN.σ
+  end
+
+  BN.λ.(BN.γ .* ((x .- μ) ./ σ) .+ BN.β)
+end
+
+children(BN::BatchNorm) =
+  (BN.λ, BN.β, BN.γ, BN.μ, BN.σ, BN.momentum, BN.ϵ, BN.active)
+mapchildren(f, BN::BatchNorm) =
+  BatchNorm(λ, f(BN.β), f(BN.γ), BN.μ, BN.σ, BN.momentum, BN.ϵ, BN.active)
+
+_testmode!(BN::BatchNorm, test) = (BN.active = !test)
+
+function Base.show(io::IO, l::BatchNorm)
+  print(io, "BatchNorm($(join(size(l.β), ", "))")
+  (l.λ == identity) || print(io, ", λ = $(l.λ)")
+  print(io, ")")
+end
--- a/src/layers/normalisation.jl
+++ b/src/layers/normalisation.jl
@@ -109,7 +109,7 @@ children(BN::BatchNorm) =
   (BN.λ, BN.β, BN.γ, BN.μ, BN.σ, BN.momentum, BN.ϵ, BN.active)
 
 mapchildren(f, BN::BatchNorm) =  # e.g. mapchildren(cu, BN)
-  BatchNorm(λ, f(BN.β), f(BN.γ), BN.μ, BN.σ, BN.momentum, BN.ϵ, BN.active)
+  BatchNorm(BN.λ, f(BN.β), f(BN.γ), BN.μ, BN.σ, BN.momentum, BN.ϵ, BN.active)
 
 _testmode!(BN::BatchNorm, test) = (BN.active = !test)
 
--- a/src/layers/normalisation.jl
+++ b/src/layers/normalisation.jl
@@ -95,7 +95,7 @@ function (BN::BatchNorm)(x)
 
     ϵ = T(BN.ϵ)
     m = size(x, 2)  # batch size
-    μ = sum(x, 2) ./ m
+    μ = mean(x, 2)
     σ = sqrt.(sum((x .- μ).^2, 2) ./ m .+ ϵ)
 
     # update moving mean/std
--- a/src/layers/normalisation.jl
+++ b/src/layers/normalisation.jl
@@ -66,7 +66,7 @@ julia> m = Chain(
   softmax)
 Chain(Dense(784, 64), BatchNorm(64, λ = NNlib.relu), Dense(64, 10), BatchNorm(10), NNlib.softmax)
 
-julia> opt = SGD(params(m), 10)  # a crazy learning rate
+julia> opt = SGD(params(m), 10, decay = .1)  # a crazy learning rate
 ```
 """
 mutable struct BatchNorm{F,V,N}
@@ -85,6 +85,8 @@ BatchNorm(dims::Integer...; λ = identity,
   BatchNorm(λ, param(initβ(dims)), param(initγ(dims)), 0., 1., ϵ, momentum, true)
 
 function (BN::BatchNorm)(x)
+  λ, γ, β = BN.λ, BN.γ, BN.β
+
   if !BN.active
     μ = BN.μ
     σ = BN.σ
@@ -102,7 +104,7 @@ function (BN::BatchNorm)(x)
     BN.σ = (1 - mtm) .* BN.σ .+ mtm .* σ.data .* m ./ (m - 1)
   end
 
-  BN.λ.(BN.γ .* ((x .- μ) ./ σ) .+ BN.β)
+  λ.(γ .* ((x .- μ) ./ σ) .+ β)
 end
 
 children(BN::BatchNorm) =
--- a/src/layers/normalisation.jl
+++ b/src/layers/normalisation.jl
@@ -92,12 +92,12 @@ function (BN::BatchNorm)(x)
     ϵ = T(BN.ϵ)
     m = size(x, 2)  # batch size
     μ = sum(x, 2) ./ m
-    σ = sqrt.(sum((x .- μ).^2, 2) ./ (m - 1) .+ ϵ)
+    σ = sqrt.(sum((x .- μ).^2, 2) ./ m .+ ϵ)
 
     # update moving mean/std
     mtm = T(BN.momentum)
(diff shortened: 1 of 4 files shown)


jekbradbury:
probably shouldn't call it Normalize, since layer normalization and weight normalization are almost as common as batch normalization these days?

MikeInnes:
Yeah, "BatchNorm" seems to be the, uh, norm for this layer.

I find this docstring a bit confusing in terms of what the example is trying to tell me. I guess the message is that this is normally used before non-linearities other than softmax, but it'd be good to just say that if so.

iblis17:
🆗 ,I rename it to `BatchNorm`.

> I find this docstring a bit confusing in terms of what the example is trying to tell me. I guess the message is that this is normally used before non-linearities other than softmax, but it'd be good to just say that if so.

In the section 3.2 of paper, it put the BatchNorm before the activation function. And for the output softmax layer ... I think we can put it, also. I changed the docstring.

src/layers/basic.jl
```diff
@@ -66,3 +66,44 @@ function Base.show(io::IO, l::Dense)
...
+function (N::BatchNorm)(x)
+  μ  = sum(x, 2) ./ size(x, 2)
+  σ² = sum((x .- μ).^2, 2) ./ size(x, 2)
+  N.γ .* ((x .- μ) ./ sqrt.(σ² .+ 1e-8)) .+ N.β
+end
+
+treelike(BatchNorm)
+
+Base.show(io::IO, l::BatchNorm) =
+  print(io, "Normalize(γ = ", l.γ.data[], ", β = ", l.β.data[], ")")
```
CarloLucibello: BatchNorm here

CarloLucibello:
I wouldn't know what the best interface for this in Flux would be, but the BatchNorm should be keeping a running mean and variance to be updated in train phase and to be used in test phase. Here a Knet example https://github.com/denizyuret/Knet.jl/pull/140/files#diff-8f43844167d26280cce2ebdd6170d842R39 I wrote some time ago.

In PyTorch you can explicitely differentiate train phase and test phase with model.train() and model.eval()

iblis17:
> I wouldn't know what the best interface for this in Flux would be, but the BatchNorm should be keeping a running mean and variance to be updated in train phase and to be used in test phase.

That's why I use `treelike` in this PR
```
treelike(BatchNorm)
```

and in the model
```julia
m = Chain(..., BatchNorm(), ...)
SGD(params(m))  # optimizer will update it
```

iblis17:
hmm, I read the Algo2 in paper again, I think I misunderstood the dimension of gamma and beta.

CarloLucibello:
I was talking about `mu` and `sigma^2`, you should accumulate them somewhere

iblis17:
@CarloLucibello oh, I got your point, but do we have batching functionality in Flux?
(I do not do batching in the MNIST example

src/layers/basic.jl
```diff
@@ -66,3 +66,44 @@ function Base.show(io::IO, l::Dense)
...
+
+```julia
+julia> m = Chain(
+  Dense(28^2, 32, identity),
+  BatchNorm(),
+  relu,
+  Dense(32, 10, identity),
+  BatchNorm(),
+  softmax)
+Chain(Dense(784, 32), Normalize(γ = 1.0, β = 0.0), NNlib.relu, Dense(32, 10), Normalize(γ = 1.0, β = 0.0), NNlib.softmax)
```
CarloLucibello: other Normalize left here

CarloLucibello:
I normally handcode batching, as in Knet examples, or there is some batching functionality in some of the JuliaML packages

CarloLucibello:
In any case, in test phase you want to be able to make a prediction also on a single image, and I think this won't give you a good result with the code as it is (the variance becomes zero).

iblis17:
I will wait for the design of training/testing phase being introducing to Flux

MikeInnes:
I just merged #92 which should have everything you need for the train/test distinction.

iblis17:
@CarloLucibello please review the algo :)
and I found that BachNorm layer ate up my memory quickly with the MNIST example. :/

---
TODO
- [ ] add test cases

CarloLucibello:
Two things are missing here:

- the BatchNorm layer should work also in conjunction with convolutional layers. 
- you should keep an array *of fixed size* of moments, one for each batch in the training set, or you can keep just one set of moments and update it with a decaying average. I think we should go with the latter way, which is easier to implement and less memory consuming. 

You can look for inspiration to my PR here https://github.com/denizyuret/Knet.jl/pull/140 (the decaying update is also there but commented out).

iblis17:
More question about moments: this function looks weird 
```julia
# function Base.push!(b::BatchMoments, μ, σ)
#     b.μ = b.momentum .* b.μ .+ (1 - b.momentum) .* b.μ
#     b.σ = b.momentum .* b.σ .+ (1 - b.momentum) .* b.σ
# end
```
Should it be
```julia
b.μ = b.momentum .* μ .+ (1 - b.momentum) .* b.μ
b.σ = b.momentum .* σ .+ (1 - b.momentum) .* b.σ

```
?

CarloLucibello:
Ah, right, stupid me

Il ven 27 ott 2017, 12:36 Iblis Lin <notifications@github.com> ha scritto:

CarloLucibello:
A good default value for the momentum of the running average could be `momentum=0.1`
http://pytorch.org/docs/master/nn.html

iblis17:
We don't have convolution layer yet. Maybe add it later and in another PR ?

src/layers/normalisation.jl
```diff
@@ -43,3 +43,72 @@ function (a::Dropout)(x)
...
+    σ = BN.σ
+  else
+    m = size(x, 2)  # batch size
+    μ = sum(x, 2) ./ m
+    σ = sqrt.(sum((x .- μ).^2, 2) ./ (m - 1) .+ BN.ϵ)
+
+    # update moving mean/std
+    mtm = BN.momentum
+    BN.μ = mtm .* μ.data .+ (1 - mtm) .* BN.μ
+    BN.σ = mtm .* σ.data .+ (1 - mtm) .* BN.σ
```
CarloLucibello: momentum and epsilong being Float64 makes the output Array{Float64} inven if the input is Array{Float32}. A
solution could be to define 
```julia
T = eltype(x)
 ϵ = T(BN.ϵ)
mtm = T(BN.momentum)

```

src/layers/normalisation.jl
```diff
@@ -43,3 +43,72 @@ function (a::Dropout)(x)
...
+```
+"""
+mutable struct BatchNorm{F,V}
+  λ::F  # activation function
+  β::V  # bias
+  γ::V  # scale
+  μ     # moving mean
+  σ     # moving std
+  ϵ::Float64
+  momentum::Float64
```
CarloLucibello: don't have to be neccesarily Float64 here, we could parametrix or leave untyped

iblis17:
test case added

iblis17:
currently, it doesn't work with CuArrays.
```julia
ERROR: LoadError: don't know how to handle argument of type Flux.##83#88{Flux.BatchNorm{NNlib.#relu,TrackedArray{…,CuArray{Float32,1}},Float64}}
Stacktrace:
 [1] cudaconvert(::Flux.##83#88{Flux.BatchNorm{NNlib.#relu,TrackedArray{…,CuArray{Float32,1}},Float64}}) at /home/iblis/.julia/v0.6/CUDAnative/src/execution.jl:20
 [2] map at ./tuple.jl:161 [inlined] (repeats 2 times)
 [3] broadcast at ./broadcast.jl:17 [inlined]
 [4] _broadcast! at /home/iblis/.julia/v0.6/CuArrays/src/broadcast.jl:22 [inlined]
 [5] broadcast_t at /home/iblis/.julia/v0.6/CuArrays/src/broadcast.jl:37 [inlined]
 [6] broadcast_c at /home/iblis/.julia/v0.6/CuArrays/src/broadcast.jl:58 [inlined]
 [7] broadcast at ./broadcast.jl:434 [inlined]
 [8] tracked_broadcast(::Function, ::TrackedArray{…,CuArray{Float32,1}}, ::TrackedArray{…,CuArray{Float32,2}}, ::TrackedArray{…,CuArray{Float32,2}}, ::TrackedArray{…,CuArray{Float32,2}}, ::TrackedArray{…,CuArray{Float32,1}}) at /home/iblis/.julia/v0.6/Flux/src/tracker/lib.jl:100
 [9] (::Flux.BatchNorm{NNlib.#relu,TrackedArray{…,CuArray{Float32,1}},Float64})(::TrackedArray{…,CuArray{Float32,2}}) at /home/iblis/.julia/v0.6/Flux/src/layers/normalisation.jl:105
 [10] mapfoldl_impl(::Base.#identity, ::Flux.##57#58, ::CuArray{Float32,2}, ::Array{Any,1}, ::Int64) at ./reduce.jl:46
 [11] (::Flux.Chain)(::CuArray{Float32,2}) at /home/iblis/.julia/v0.6/Flux/src/layers/basic.jl:30
 [12] loss(::CuArray{Float32,2}, ::Flux.OneHotMatrix{CuArray{Flux.OneHotVector,1}}) at /home/iblis/git/Flux/model-zoo/mnist/mnist.jl:24
 [13] macro expansion at /home/iblis/.julia/v0.6/Flux/src/optimise/train.jl:19 [inlined]
 [14] macro expansion at /home/iblis/.julia/v0.6/Juno/src/progress.jl:128 [inlined]
 [15] #train!#112(::Flux.#throttled#14, ::Function, ::Function, ::Base.Iterators.Take{Base.Iterators.Repeated{Tuple{CuArray{Float32,2},Flux.OneHotMatrix{CuArray{Flux.OneHotVector,1}}}}}, ::Flux.Optimise.##58#62) at /home/iblis/.julia/v0.6/Flux/src/optimise/train.jl:18
 [16] (::Flux.Optimise.#kw##train!)(::Array{Any,1}, ::Flux.Optimise.#train!, ::Function, ::Base.Iterators.Take{Base.Iterators.Repeated{Tuple{CuArray{Float32,2},Flux.OneHotMatrix{CuArray{Flux.OneHotVector,1}}}}}, ::Function) at ./<missing>:0
 [17] include_from_node1(::String) at ./loading.jl:576
 [18] include(::String) at ./sysimg.jl:14
 [19] process_options(::Base.JLOptions) at ./client.jl:305
 [20] _start() at ./client.jl:371
while loading /home/iblis/git/Flux/model-zoo/mnist/mnist.jl, in expression starting on line 38
```

Any idea?

MikeInnes:
That's due to broadcast calls like `BN.a .* xs`. If you unpack first (like `a = BN.a; a .* xs`) it should work.

iblis17:
Ok, it finally works with CuArrays.

src/layers/normalisation.jl
```diff
@@ -43,3 +43,80 @@ function (a::Dropout)(x)
...
+
+    ϵ = T(BN.ϵ)
+    m = size(x, 2)  # batch size
+    μ = mean(x, 2)
+    σ = sqrt.(sum((x .- μ).^2, 2) ./ m .+ ϵ)
+
+    # update moving mean/std
+    mtm = T(BN.momentum)
+    BN.μ = (1 - mtm) .* BN.μ .+ mtm .* μ.data
+    BN.σ = (1 - mtm) .* BN.σ .+ mtm .* σ.data .* m ./ (m - 1)
```
CarloLucibello: I think this should be
BN.μ = mtm .* BN.μ .+ (1-mtm ).* μ.data

iblis17: Given this torch implementation: https://github.com/pytorch/pytorch/blob/502aaf39cf4a878f9e4f849e5f409573aa598aa9/aten/src/THNN/generic/BatchNormalization.c#L46

and I use the `momentum = .1` as default value which is same as pytorch.

So... your version (` BN.μ = mtm .* BN.μ .+ (1-mtm ).* μ.data`) should be paired with `mtm = 0.9` as default value.

CarloLucibello: right, but maybe we should go with the usual meaning for momentum, as used in optimizers, and other BN implementations e.g.
http://gluon.mxnet.io/chapter04_convolutional-neural-networks/cnn-batch-norm-scratch.html
and default to 0.9
Anyway, given there is a precedent in pytorch, any choiche is allright

iblis17: > maybe we should go with the usual meaning for momentum, as used in optimizers, and other BN implementations

So... here we have two options here

1. `mtm * new_vector + ...`
2. `mtm * old + (1-mtm) * new`

In my current code is option (1), right?

Let's dig into Flux's momentum:
https://github.com/FluxML/Flux.jl/blob/dc1f08a709d1dc9e90fba84b0dc446a4c74664de/src/optimise/optimisers.jl#L8-L11

So I think the Flux's optimizer choose option (1).

Next, Let's see the implementation of Gluon. In the doc you show `momentum`'s default value is `0.9` and
```python
    # store the moving statistics by their scope_names, inplace
    if scope_name not in _BN_MOVING_MEANS:
        _BN_MOVING_MEANS[scope_name] = mean
    else:
        _BN_MOVING_MEANS[scope_name] = _BN_MOVING_MEANS[scope_name] * momentum + mean * (1.0 - momentum)
    if scope_name not in _BN_MOVING_VARS:
        _BN_MOVING_VARS[scope_name] = variance
    else:
        _BN_MOVING_VARS[scope_name] = _BN_MOVING_VARS[scope_name] * momentum + variance * (1.0 - momentum)
```
And I think they choose option (2).

I will vote for being consistent with Flux's source code, which is option (1), with default value as `0.1`.
I can add more note in the docsting if you want

MikeInnes:
This is an excellent PR. Thanks a lot @iblis17!

PR #85, opened 2017-10-17
Lazy Softmax

MikeInnes:
`softmax` looks just like it did before:

```julia
julia> ps = softmax([1,2,3])
3-element Flux.Softmax{Float64,1,Array{Float64,1}}:
 0.0900306
 0.244728
 0.665241
```

But under the hood it's now lazy, and actually stores the original logits:

```julia
julia> ps.logits
3-element Array{Float64,1}:
 1.0
 2.0
 3.0
```

What this means is that functions like `crossentropy` can avoid computing the `softmax` entirely and be much faster; but this is completely transparent and models benefit from it for free. With our sparse matrix types this gets us fast code without memorising a bunch of horrific functions like `tf.nn.sparse_softmax_cross_entropy_with_logits`. `crossentropy(softmax(ŷ), y)` is identical.

Needs tests, gradient checks and some GPU support hacks.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -23,6 +23,7 @@ include("utils.jl")
 include("onehot.jl")
 include("tree.jl")
 
+include("layers/softmax.jl")
 include("layers/stateless.jl")
 include("layers/basic.jl")
 include("layers/recurrent.jl")
--- a/src/layers/softmax.jl
+++ b/src/layers/softmax.jl
@@ -1,19 +1,19 @@
-mutable struct Softmax{T,N,A,B} <: AbstractArray{T,N}
+mutable struct Softmax{T,N,A} <: AbstractArray{T,N}
   logits::A
-  probs::B
-  Softmax{T,N,A,B}(logits::A) where {T,N,A,B} = new(logits)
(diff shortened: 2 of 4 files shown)


jekbradbury:
I'm curious how you feel about this API now—I'm still kind of into it; I think it's orthogonal/complementary to having a good actual softmax implementation and could be a key part of the magic of Flux?

MikeInnes:
I agree completely. I'm holding off on it mostly because I neither want to introduce something fragile nor to dive into a rabbit whole of dispatch ambiguities (which I think this may cause with TrackedArrays, just like OneHotMatrix does; oh and probably with that too). Compiler AD will help a lot with these kinds of things, as will better support for wrapper types, if we can figure that out.

For my part I'll probably pick this up again once we are benchmarking properly :)

MikeInnes:
Yes, I think I might move this over to Zygote and we can start prototyping properly with it there.

ToucheSir:
Assuming this can't (theoretically) be handled by the new AD system, the PR can probably be closed because of how tied in it is with Tracker.

PR #94, opened 2017-10-23
improve optimizers

CarloLucibello:
More coherence in the interface + insert missing learning rates + some tests

### DIFF
--- a/src/optimise/interface.jl
+++ b/src/optimise/interface.jl
@@ -1,5 +1,7 @@
 call(f, xs...) = f(xs...)
 
+# note for optimisers: set to zero
+# p.Δ at the end of the weigths update
 function optimiser(ps, fs...)
   ps = [Param(p) for p in ps]
   fs = map(ps) do p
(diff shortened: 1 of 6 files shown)


src/optimise/interface.jl
```diff
@@ -47,27 +49,27 @@ RMSProp(ps, η = 0.001; ρ = 0.9, ϵ = 1e-8, decay = 0) =
...
 [ADAGrad](http://www.jmlr.org/papers/volume12/duchi11a/duchi11a.pdf) optimiser.
 Parameters don't need tuning.
 """
-ADAGrad(ps; η = 0.01, ϵ = 1e-8, decay = 0) =
+ADAGrad(ps, η = 0.01; ϵ = 1e-8, decay = 0) =
   optimiser(ps, p -> adagrad(p; η = η, ϵ = ϵ), p -> invdecay(p, decay), p -> descent(p, 1))
 
 """
-    ADADelta(params; η = 0.01, ρ = 0.95, ϵ = 1e-8, decay = 0)
+    ADADelta(params η = 0.01; ρ = 0.95, ϵ = 1e-8, decay = 0)
```
iblis17: well, missed a comma

src/optimise/optimisers.jl
```diff
@@ -1,74 +1,88 @@
 function descent(p::Param, η::Real)
-  function ()
-    p.x .-= p.Δ .* η
-    p.Δ .= 0
-  end
+    function ()
+        @. p.x -= η * p.Δ
+        @. p.Δ = 0
+    end
```
iblis17: hmm, I think the whole casebase is using 2-space indentation?

CarloLucibello: argh, unfortunately yes

MikeInnes:
This is a good set of changes, thanks! One question though, is there a specific reason for putting the `descent` step straight into the momentum optimisers, rather than keeping them separate?

CarloLucibello:
> This is a good set of changes, thanks! One question though, is there a specific reason for putting the descent step straight into the momentum optimisers, rather than keeping them separate?

Not to my knoweldge, it causes only confusion. I removed `descent` from `Nesterov` and `Momentum` in previous commits and I'm taking care of the others.

CarloLucibello:
> This is a good set of changes, thanks! One question though, is there a specific reason for putting the descent step straight into the momentum optimisers, rather than keeping them separate?

Not to my knoweldge, it causes only confusion. I removed `descent` from `Nesterov` and `Momentum` in previous commits and I'm taking care of the others.

CarloLucibello:
I had to change this line
```julia
# validword(s) = ismatch(r"^[\w-\.]+$", s)
validword(s) = ismatch(r"^\[\w-\.\]+$", s)
```
because I was having this error
```julia
julia> include("src/Flux.jl")
ERROR: LoadError: LoadError: LoadError: PCRE compilation error: invalid range in character class at offset 4
Stacktrace:
 [1] include_from_node1(::String) at ./loading.jl:576 (repeats 2 times)
 [2] macro expansion at /home/carlo/.julia/v0.6/Revise/src/Revise.jl:775 [inlined]
 [3] (::Revise.##17#18{Base.REPL.REPLBackend})() at ./event.jl:73
while loading /home/carlo/Git/Flux.jl/src/data/cmudict.jl, in expression starting on line 36
while loading /home/carlo/Git/Flux.jl/src/data/Data.jl, in expression starting on line 11
while loading /home/carlo/Git/Flux.jl/src/Flux.jl, in expression starting on line 32
```
I don't know if I did the right thing

CarloLucibello:
> This is a good set of changes, thanks! One question though, is there a specific reason for putting the descent step straight into the momentum optimisers, rather than keeping them separate?

Sorry, I misread. Also, I thought it would be better to have the descent  step in the momentum, to have the whole update in a single function and do less mental gym, but now I realize that having descent separate is necessary to correctly apply the decay on the learning rate. So, I'll change things back.

CarloLucibello:
Ok, this is ready for review and merge. Aside from the problem with `validword(s)`, it is also puzzling the fact that RMSprop doesn't pass the simple test I added, unless some decaying for the learning rate is set (the default is zero).

CarloLucibello:
I chose the learning rate to be taken as a positional argument by each optimizer, but maybe we want also that to be a keword argument

CarloLucibello:
bump

MikeInnes:
Ok, not sure why `validword` never caused an issue for me, but it was indeed incorrect. I think RMSprop was also fixed recently.

Some of the optimises were struggling so I tried adding something more "realistic", and closer to regression; hopefully that will work just as well. It seems fishy that ADAgrad behaves so poorly without an increased learning rate, given that your not supposed to tweak it, any ideas about that?

Anyway, thanks again!

PR #96, opened 2017-10-24
Make crossentropy symmetric

ChrisRackauckas:
The current loss function always goes to zero if yhat -> constant 1. It needs the extra term so it's penalized when not getting yhat correct, instead of only being penalized not getting yhat correct when y is 1.

@oxinabox had this idea.

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -3,7 +3,7 @@
 mse(ŷ, y) = sum((ŷ .- y).^2)/length(y)
 
 crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat) =
-  -sum(y .* log.(ŷ)) / size(y, 2)
+  -sum(y .* log.(ŷ) + (1-y).*log.(1-̂y)) / size(y, 2)
 
 @deprecate logloss(x, y) crossentropy(x, y)
 


MikeInnes:
Not sure this is correct; crossentropy if for comparing probability distributions, and distribution with a single category only has one correct value (1).

CarloLucibello:
also, you already have the kind of penalization you are suggesting, because of the denominator in the softmax layer typically used in combination with the crossentropy loss

ChrisRackauckas:
I was dumb.

oxinabox:
>also, you already have the kind of penalization you are suggesting, because of the denominator in the softmax layer typically used in combination with the crossentropy loss

The target in this case was putting it on top of a sigmoid output, rather than softmax, for binary classification (I still may have screwed up the logic though).

CarloLucibello:
>The target in this case was putting it on top of a sigmoid output, rather than softmax, for binary classification (I still may have screwed up the logic though).

well in the binary (2-class) classification case y and yhat would have one component and  `-sum(y .* log.(ŷ) + (1-y).*log.(1-̂y)) / size(y, 2)` would be the correct cross entropy. For general K-class classification using a K component output the thing is trickier though, since the sigmoid would define a (factorized) probability on  2^K possible outcomes. In this case `-sum(y .* log.(ŷ) + (1-y).*log.(1-̂y)) / size(y, 2)` would still work (I know because I used it), but cannot be called a cross-entropy

oxinabox:
Caffe calls [this one SigmoidCrossEntropy](http://caffe.berkeleyvision.org/doxygen/classcaffe_1_1SigmoidCrossEntropyLossLayer.html)**

CarloLucibello:
ah! But... yhat components do not generally sum to 1, so they cannot be interpreted as probabilities...

CarloLucibello:
Tensor flow has this interpretation https://www.tensorflow.org/api_docs/python/tf/nn
"Measures the probability error in discrete classification tasks in which each class is independent and not mutually exclusive. For instance, one could perform multilabel classification where a picture can contain both an elephant and a dog at the same time."
So that would be a sum of K binary cross entropies

PR #104, opened 2017-11-04
Allow array of optimisers to train!

baggepinnen:
This allows an array of optimisers to be sent to `train!`, nice if you want different learning rates etc. for your parameters.


MikeInnes:
Thanks.

I partly wonder if we should just have a single callback argument here – they are completely redundant, but it just seems clearer to pass the optimiser explicitly.

PR #107, opened 2017-11-14
Fix bug in rmsprop and adadelta

baggepinnen:
`@. p.Δ = η * p.Δ / √acc` parses correctly while `@. p.Δ /= √acc*η` seems to parse like `@. p.Δ /= (√acc*η)`, hence the step size was de facto interpreted as `1/η`


MikeInnes:
:+1: Thanks!

PR #112, opened 2017-11-24
Implement Gated Recurrent Unit

### LINKED ISSUE #90: GRU recurrent cell
Along with other recurrent layers we're missing. See the [current implementations](https://github.com/FluxML/Flux.jl/blob/master/src/layers/recurrent.jl) of LSTM and RNN for reference.

baggepinnen:
Fixes #90  :)
Tests are coming

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -7,7 +7,7 @@ module Flux
 using Juno, Requires
 using Lazy: @forward
 
-export Chain, Dense, RNN, LSTM, Dropout, LayerNorm,
+export Chain, Dense, RNN, LSTM, GRU, Dropout, LayerNorm,
   SGD, ADAM, Momentum, Nesterov,
   param, params, mapleaves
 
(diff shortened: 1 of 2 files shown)


baggepinnen:
Actually, I can not find tests on other recurrent layers. Any suggestions on reasonable ways of testing these?

MikeInnes:
For a test, how about learning to predict a trivial sequence (perhaps alternating `[1, 0]`, `[0, 1]`)? All RNNs should be capable of that, I think.

baggepinnen:
I had an idea along those lines initially, but was hesitating to implement it since I thought the time required to test the package would become very large if several layers had tests that required a train loop..

MikeInnes:
If it's simple enough it should train pretty quickly. We already do this with the optimisers as well.

MikeInnes:
Merging this for now and we'll figure out the tests. Thanks @baggepinnen!

PR #119, opened 2017-12-04
Amsgrad

baggepinnen:
This PR implements `AMSGrad` optimiser
https://openreview.net/forum?id=ryQu7f-RZ
It also fixes a small bug in `ADAM` where a square root was applied where it shouldn't be according to the original paper.


MikeInnes:
Awesome. Thanks a lot!

PR #122, opened 2017-12-05
Add `weighted_crossentropy` for imbalanced classification problems

staticfloat:
This is useful if, for example, you want to train with 75% of your data being of class 0, and 25% of your data being of class 1.

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -4,15 +4,9 @@ using NNlib: log_fast
 
 mse(ŷ, y) = sum((ŷ .- y).^2)/length(y)
 
-function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat)
-  return -sum(y .* log_fast.(ŷ)) / size(y, 2)
+function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight = 1)
+  return -sum(y .* log_fast.(ŷ) .* weight) / size(y, 2)
 end
-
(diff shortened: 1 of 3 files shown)


src/layers/stateless.jl
```diff
@@ -4,8 +4,16 @@ using NNlib: log_fast
...
 
-crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat) =
-  -sum(y .* log_fast.(ŷ)) / size(y, 2)
+function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat)
+  return -sum(y .* log_fast.(ŷ)) / size(y, 2)
+end
+
+function weighted_crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat, w::AbstractVecOrMat)
+  # Here we directly weight `y` so that this reduces to `-sum( y .* log(ŷ) .* w)`
+  return crossentropy(ŷ, y .* w)
```
vchuravy: It is probably worth it to inline `crossentropy` manually here so that broadcast fusion can remove that extra allocation.

staticfloat: Ah, didn't think of that.  Good call.

MikeInnes:
Would it make sense to have this as a keyword argument? With `weight = 1` by default it wouldn't add any overhead.

MikeInnes:
I went ahead and tried that.

Thanks @staticfloat!

PR #123, opened 2017-12-07
Added vcat for multiple TrackedVectors

GenaBitu:
Note: the previous line to this addition must be left as is to prevent ambiguity when concatenating 2 vectors


GenaBitu:
back not yet working, do not merge, will try to fix it...

GenaBitu:
Ok, back should now work as well

MikeInnes:
Cool, thanks. Can you add a gradient check for this as well?

Would be great to have a more general `cat` as well.

GenaBitu:
Do you mean this check?

MikeInnes:
Thanks a lot for the patch!

Issue #101, opened 2017-10-31
Jacobian and grad after getindex


baggepinnen:
I am missing calculation of jacobians
```julia
julia> using Flux

julia> using Flux: back!

julia> x = randn(2)
2-element Array{Float64,1}:
 -0.488323
  0.143252

julia> y = Dense(2,2)(x)

Tracked 2-element Array{Float64,1}:
 -0.0087318
  0.0192173

julia> back!(y)
ERROR: MethodError: no method matching back!(::TrackedArray{…,Array{Float64,1}})
Closest candidates are:
  back!(::Flux.Tracker.TrackedArray, ::Any) at /home/fredrikb/.julia/v0.6/Flux/src/tracker/back.jl:39
  back!(::Flux.Tracker.TrackedArray{T,0,A} where A where T) at /home/fredrikb/.julia/v0.6/Flux/src/tracker/back.jl:43
```

Furthermore, the following confuses me
```julia
julia> x = param(randn(2))
Tracked 2-element Array{Float64,1}:
  0.567788
 -0.32716 

julia> y = Dense(2,2)(x)
Tracked 2-element Array{Float64,1}:
  0.00648598
 -0.00119976

julia> back!(y[1])

julia> x.grad
2-element Array{Float64,1}:
 -0.00194926 
 -0.000263109
```

while the following returns an error
```julia
julia> x[1].grad
ERROR: UndefRefError: access to undefined reference
```
It seems the gradient information is lost when accessing an element

MikeInnes:
For `back!`, if the object isn't a scalar, you need to provide a gradient. e.g. `back!(y, [1, 2])`. We need a much better error message here though.

To get a gradient of an array element the right way would be `x.grad[1]`. We can't carry gradient information forward when you operate on a TrackedArray.

baggepinnen:
I'm trying to implement a simple `jacobian` function but really can't figure out how the gradient is calculated and accessed. The example below tries to model a time series `x[k+1] = Ax[k]`. I want to recover the `A` from the neural network by calculating the jacobian of the vector-valued output with respect to the vector-valued input. How do I calculate the gradient of a TrackedArray wrt an `Array/TrackedArray`? My attempt below does not work, but might highlight my misunderstanding.

```julia
    using Flux
    using Flux: back!, truncate!, treelike, train!, mse
    N      = 300
    n      = 2

    function generate_data()
        A      = [0.999 1; 0 0.9]
        x0     = randn(n)
[...]

PR #124, opened 2017-12-08
Add jacobian function

baggepinnen:
Implements `jacobian` to calculate the output jacobian with respect to an input vector
Discussed in #101


MikeInnes:
:+1: Thanks!

PR #125, opened 2017-12-08
Define dot operator for TrackedArray

### LINKED ISSUE #114: define `dot`  for tracked arrays
I think it would be sensible if 

```
param([1,2]) ⋅ param([2,2]) 
```
worked
rather than:
```
MethodError: no method matching dot(::TrackedArray{…,Array{Float64,0}}, ::TrackedArray{…,Array{Float64,0}})
```

Right now I have in my source:

```
LinAlg.dot(x::Flux.TrackedArray, y::Flux.TrackedArray) = sum(x.*y)
```

Probably that is all it really needs.

americast:
Extends `Base` to add support for the `⋅` operator for `TrackedArray`s.
Fixes #114.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -13,7 +13,7 @@ export Chain, Dense, RNN, LSTM,
   param, params, mapleaves
 
 using NNlib
-export σ, sigmoid, relu, leakyrelu, elu, swish, softmax
+export σ, sigmoid, relu, leakyrelu, elu, swish, softmax, ⋅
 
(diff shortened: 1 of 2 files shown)


americast:
@MikeInnes Is this required anymore?

MikeInnes:
Ah whoops, didn't see that you had this PR. No, not required any more.

Unfortunately this wouldn't have worked anyway as it doesn't carry the derivatives through.

PR #130, opened 2017-12-17
Vecdot for matrices

oschr:
Adds dispatched vecdot for TrackedArrays. Solves #114 for matrices too.


src/tracker/lib.jl
```diff
@@ -1,178 +1,188 @@
...
+LinAlg.dot(xs::TrackedVector, ys::TrackedVector) = TrackedArray(Call(dot, xs, ys), toarray(xs.data, dot(data(xs), data(ys))))
+LinAlg.dot(xs::AbstractVector, ys::TrackedVector) = TrackedArray(Call(dot, xs, ys), toarray(xs.data, dot(data(xs), data(ys))))
+LinAlg.dot(xs::TrackedVector, ys::AbstractVector) = TrackedArray(Call(dot, xs, ys), toarray(xs.data, dot(data(xs), data(ys))))
+
+function back(::typeof(dot), Δ, xs, ys)
+  @back(xs, Δ.*ys)
+  @back(ys, Δ.*xs)
+end
+
+LinAlg.vecdot(xs::TrackedArray, ys::TrackedArray) = TrackedArray(Call(vecdot, xs, ys), toarray(xs.data, vecdot(data(xs), data(ys))))
```
oschr: Changed part

src/tracker/lib.jl
```diff
@@ -1,178 +1,188 @@
...
+LinAlg.vecdot(xs::AbstractArray, ys::TrackedArray) = TrackedArray(Call(vecdot, xs, ys), toarray(xs.data, vecdot(data(xs), data(ys))))
+LinAlg.vecdot(xs::TrackedArray, ys::AbstractArray) = TrackedArray(Call(vecdot, xs, ys), toarray(xs.data, vecdot(data(xs), data(ys))))
+
+function back(::typeof(vecdot), Δ, xs, ys)
+  @back(xs, Δ.*ys)
+  @back(ys, Δ.*xs)
+end
+
```
oschr: This is the end of part I've changed (can't figure why did the comparison failed, there are 10 changed lines in total)

MikeInnes:
If I had to guess, you are using Windows and an editor that doesn't respect line endings? If so, I can probably figure this out but it might be worth using something like Atom which will do the right thing for you.

bhvieira:
If any of this functionalities are needed in Tracker perhaps it's better to move it there, as Flux does not depend on Tracker anymore.

PR #131, opened 2017-12-18
make stopping work with multiple CBs

oxinabox:
This needs tests still

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -5,7 +5,7 @@ runall(f) = f
 runall(fs::AbstractVector) = () -> foreach(call, fs)
 
 runcheck(f) = () -> f() == :stop
-runcheck(f) = () -> any(call.(fs) .== :stop)
+runcheck(fs::AbstractVector) = () -> any(call.(fs) .== :stop)
 
 
 """
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -4,6 +4,10 @@ using Flux.Tracker: back!, value
 runall(f) = f
 runall(fs::AbstractVector) = () -> foreach(call, fs)
 
+runcheck(f) = :stop == f
+runcheck(f) = () -> any(:stop .== call.(fs))
+
+
 """
     train!(loss, data, opt)
 
@@ -21,9 +25,10 @@ Flux.train!(loss, data, opt,
 The callback can return `:stop` to interrupt the training loop.
 
 Multiple optimisers and callbacks can be passed to `opt` and `cb` as arrays.
+If any of the `cb` callbacks return `:stop` then training will end.
 """
 function train!(loss, data, opt; cb = () -> ())
-  cb = runall(cb)
+  cb_stop = runcheck(cb)
   opt = runall(opt)
   @progress for d in data
     l = loss(d...)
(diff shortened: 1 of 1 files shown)


src/optimise/train.jl
```diff
@@ -4,6 +4,10 @@ using Flux.Tracker: back!, value
 runall(f) = f
 runall(fs::AbstractVector) = () -> foreach(call, fs)
 
+runcheck(f) = :stop == f
```
iblis17: :/ cannot understand this line

oxinabox: fixed.

oxinabox:
The usecase for this BTW is that one likely wants to use one callback for stopping, and one for logging.
at different throttle-rates

src/optimise/train.jl
iblis17: but I think the next line override this declaration

oxinabox: I need to stop trying to "make a quick simple PR via the webinterface". It never goes well for me.

oxinabox:
@MikeInnes  I am still thinking about #132 
but til then I think you should take a look at this, since it is a bug fix to the current implementation.

MikeInnes:
Would it be cleaner to make `runall` just return the result of the last callback? I kind of like that the current approach is composable (e.g. we can call `runall` recursively and pass `[[cb1, cb2], cb3]`).

oxinabox:
No, I don't think it would.
That would mean you were only allowed a single stopping condition callback.
I think it is quiet reasonable to have a whole bunch of stopping conditions(possibly flux should predefine some).
And to write them as separate functions -- probably as functions that return functions.

Eg:
`early_stopping` to check how well doing on a dev-set, `atol_on_improvement`, `rtol_on_improvement` etc.

If we were going to restrict it such that the callback that is a stopping condition, had to be in the final position, then we might as well do #132

MikeInnes:
I agree completely with the sentiment. Here are my considerations:

* The function list syntax is convenient, but doesn't (and shouldn't) provide something fundamentally new; you can always just define your own callback that does the same. So this issue is about convenience rather than features per se.
* Convenience of multiple stopping conditions (and predefined ones) is important, but we don't actually know what that looks like or have concrete examples yet. When it comes up there will be multiple possible approaches to solving it, this being just one option.
* I'm a bit wary of adding features to the callback syntax; it risks becoming a non-obvious new notation of its own. It's also much easier to add features like this later, if we need them, than to change or revert them. It's also easy to get the same effect explicitly (e.g. `cb = either(stop1, stop2)`).

So for those reasons I'd rather punt on this until we have a very clear idea of cases where it'll help. But I'd be OK with the "return last result" change, as it feels more obvious and will be forwards-compatible if we decide to do this later.

oxinabox:
> So for those reasons I'd rather punt on this until we have a very clear idea of cases where it'll help

Yeah, happy to keep thinking on this,
it is in the issue tracker (well PR tracker), so it isn't going to be forgotten.

I kind of like the `throw(StopTraining())` idea BTW.
Seems like it will be clear and flexible.
(E.g. to writing a single large callback that does everything from printing to early stopping, or to having many small callbacks doing different things, multiple but not all of which are causing stopping.)

MikeInnes:
Yeah, and you could expose it as `Flux.stop()` or something. That does seem like a good approach.

ToucheSir:
`Flux.stop()` has been around for a while, so this can be closed.

Issue #127, opened 2017-12-13
Stopping condition callbacks


ChrisRackauckas:
It would be nice if the callback could handle stopping conditions. For example, I could see it taking in the current weights and the current loss, and splitting back a boolean for whether or not to continue. Then it would be like:

```julia
evalcb = (weights,cur_loss) -> (@show(loss(xinv, yinv)); cur_loss < 5e-6)
```

would let the training halt on a condition.

MikeInnes:
Good idea. I've implemented this as an extension of callbacks; you can just return `:stop` and it'll do so.

I don't think we need to pass information about the model to the callback, because you can just close over it.

PR #132, opened 2017-12-18
Restructure `train!` callbacks

oxinabox:
Rather than #131 
this is a more complete reworking of the callbacks.

I like the idea of #127,
but I think that it deserves its own callback, currently called `stopping_citeria`, returning `true` when it is time to stop.
Rather than having `cb` return a magic symbol.

I renamed `cb` to `log_cb`.

Also I made both callbacks take the iteration number, and the current loss as parameters.
Because they are very often wanted in these things.
This is breaking.

A deprecation could be added by doing something like:

```
if cb!=nothing
    warn("cb deprecated...")
    push!(log_cb, (i,l)->cb())
    push!(stopping_criteria, (i,l)->cb()==:stop)
end
```

if that is desired.

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -2,7 +2,7 @@ using Juno
 using Flux.Tracker: back!, value
 
 runall(f) = f
-runall(fs::AbstractVector) = () -> foreach(call, fs)
+runall(fs::AbstractVector) = (args...) -> map(f->f(args...), fs)
 
 """
     train!(loss, data, opt)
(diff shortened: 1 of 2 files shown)


MikeInnes:
Computer science is like an early human language – the only numbers we use are 0, 1, and many.

I don't think that people will want callbacks for only these two reasons, so we need a solution that doesn't involve adding arbitrarily many `*_cb` keyword args, like the current `cb = [throttle(f, 1), throttle(g, 2)]`. (technically, `cb` itself is redundant as you can just do `opt = [ADAM(), cb]`, but that might change with cassette.)

Same goes for the loss and iteration data – will people really only ever want that, or will we have to break this every time we add something new?

This seems like a good use case for dynamic binding; then you could look up `Flux.trainstate()` if you needed it without any impact when you don't, and this is easy to extend in future. We could also change the stopping mechanism to be a call to `Flux.stop()` if that's a bit less "magic-value"-y.

oxinabox:
Yes, this posits two types (of which,  am sure you already realise you can have multiple of each, just like with current `cb`).
Normal callbacks. i.e. what I am calling `log_cb`, and callbacks that can cause termination based on their value (what I am calling `stopping_criteria`).

I think it is more semantic.  I guess really that is the key point.
One looks at it and says "What is the stopping_criteria?"
And then one looks at that list of functions
Continuing down that direction would be to warp the dataset allways in an infinite loop,
and then use a stopping_criteria callback to terminate it after `n` loops (rather than wrapping it in `Iterators.repeat(dataset, n)`)

Of course, yes this is just one option.

A call to `flux.stop()` would be I think be cleaner yes, that returning `:stop`.
Making all callbacks able to return `false` to terminate to terminate is another.
Throwing a `StopTrainingException` is another.


>Same goes for the loss and iteration data – will people really only ever want that, or will we have to break this every time we add something new?
>
> This seems like a good use case for dynamic binding; then you could look up Flux.trainstate() if you needed it without any impact when you don't

That sounds like a good idea.
Though I am not sure what you mean by a dynamic binding.
Is that one of the uses of a `let` block? (those scare me sometimes)

MikeInnes:
If one callback type is strictly more powerful than the other, why not just stick to that? I could buy it if it was a big help to clarity, but I think it's just going to mean a much more complex explanation of `train!` in the docs.

Yeah, so by dynamic binding I mean something like a let that acts through function calls. e.g.

```
global i = 0
function train!(f)
  dynamic let i = 1
    f()
  end
end
```

Normally `f` would see `i == 0` but with dynamic binding you can make it see the new value. It's easy enough to simulate with a global variable in the basic case.

davidavdav:
I don't know if this is the place to discuss this, but I don't understand the design that the stopping criterion is in a callback. 

In fact, I don't understand from reading the code of `train!()` why the char-rnn example eventually stops (as it does for me).

MikeInnes:
> I don't understand the design that the stopping criterion is in a callback.

Can you clarify that? Having a callback seems like the only way to let users choose any stopping condition.

> I don't understand from reading the code of train!() why the char-rnn example eventually stops

The `for` loop only goes over the `data` iterator once (there's no infinite `while true` for example), so it'll stop as soon as that's done. The stopping condition is only an optional early stop. You can make the loop do multiple (or infinite) epochs by modifying `data`. Does that clarify things?

davidavdav:
OK, sorry, it was not clear from the documentation that the training was over a single epoch.   And I must have not interpreted the `train!` code correctly.   So, in order to do another epoch for training, I should probably 
 - randomize the order of the batches
 - call `train!` again
 - decide on validation data if training is good enough for me

Thanks

MikeInnes:
Yeah that sounds about right. Please feel free to add things to the docs as I'm sure this could be clearer there.

oxinabox:
Closing this because now can just write custom training loops

Issue #139, opened 2018-01-04
Is `crossentropy()` correct?


SimonEnsemble:
As implemented:
```
function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight = 1)
  return -sum(y .* log_fast.(ŷ) .* weight) / size(y, 2)
end
```

I took a look at the code b/c I was getting `Inf`; I think in practice ppl add a small number to the argument of the `log` to avoid this.

But I think [binary] cross entropy should be:
```
const eps = 1e-7

function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat)
    return -sum(y .* log.(ŷ + eps)) - sum((1.0 - y) .* log.(1.0 - ŷ + eps))
end
```

See [here](https://en.wikipedia.org/wiki/Cross_entropy#Cross-entropy_error_function_and_logistic_regression).

CarloLucibello:
Crossentropy and binary crossentropy are two slightly different things. Crossentropy is meant to by used on the output of a softmax layer which encodes a probability distributions over K classes. The binary crossentropy you wrote should be applied on the output of a logistic or a  tanh layer, encoding a factorized probability over 2^K binary configurations. So, both methods you wrote express cross entropies, but they are meant to be applied to different probabilistic spaces. 

In my experience (with Knet, but I guess with Flux is the same) I never encountered inf/nan in the crossentropy, unless initializing with weights too big or setting a learning rate too high. The binary crossentropy instead is more fragile, and yes adding an epsilon fixes the problem.

SimonEnsemble:
Thanks @CarloLucibello. So for two class discrimination, I should use the one-hot encoding such as 
```
y = [0 0 1; 1 1 0]
```
(for 3 data points) if I want to use the `crossentropy` loss function.

Otherwise, if I encode the target values as `y = [1 1 0]`, I will just write my own binary cross entropy loss function as above.

MikeInnes:
Correct. Flux could house a binary crossentropy function as well if you want to contribute it.

MikeInnes:
Closing this in favour of #143

PR #145, opened 2018-01-08
Improve binary cross entropy

### LINKED ISSUE #143: Binary cross entropy
See #139 and #96. Could use a better name than `binary_cross_entropy` as well.

americast:
@MikeInnes Adds binary cross entropy. Fixes #143.


MikeInnes:
Looks good, though I would make it a scalar function (since that's effectively what you're doing with the size restriction anyway) and people can broadcast it like `binarycrossentropy.(y, yhat)` when they have arrays.

Indentation looks off as well (the standard in Flux is 2-space indent).

src/layers/stateless.jl
```diff
@@ -16,6 +16,15 @@ function logitcrossentropy(logŷ::AbstractVecOrMat, y::AbstractVecOrMat)
   -sum(y .* ypred) / size(y, 2)
 end
 
+function binarycrossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat)
+               if (size(ŷ )==(1,) && size(y)==(1,))
```
CarloLucibello: why this? Here you are essentially allowing as inputs vectors of length 1

CarloLucibello:
I think the method definition here is far too strict. In tipical applications you may want to average the BCE across many pixels and across many samples, so something like

```julia
function binarycrossentropy(ŷ, y, average=true)
     bce  = -sum(y .* log_fast.(ŷ) + (1 .- y) .* log_fast.(1 - ŷ))
     average ? bce / length(y) : bce
end
```
would be more useful.

Also this needs a docstring, and maybe a little test. Also a little additive factor may be needed inside the logarithms for numerical stability, as discussed in #139

MikeInnes:
@CarloLucibello I'm thinking that we should just export the scalar kernel here and let people sum/average it themselves, any thoughts?

CarloLucibello:
>@CarloLucibello I'm thinking that we should just export the scalar kernel here and let people sum/average it themselves, any thoughts?

mhmh, doesn't look very practical to me. We can have both behaviours though, since the method above applies to scalars as well (fortunately length(scalar)=1), so `binarycrossentropy(0.1, 0.1)` and broadcasting  `binarycrossentropy.(yhat, y)` are valid operations

MikeInnes:
I mean, it would be a case of replacing

```julia
binarycrossentropy(ŷ, y, average=true)
```

with

```
mean(binarycrossentropy.(ŷ, y))
```

which doesn't seem particularly worse. My guess is that in most cases you'll be working with batches anyway and want to do something like `mean(binarycrossentropy.(...), 2)`, so this seems more flexible.

americast:
Thanks a lot for your suggestions. I have tried to take care of them.
@MikeInnes I guess the current implementation enables calling it in as a scalar function as well.
>My guess is that in most cases you'll be working with batches anyway and want to do something like `mean(binarycrossentropy.(...), 2)`

Agreed. Since there's only binary classification, I believe it would be safe to assume that `y` has a size `(1,<batch size>)`. In the current implementation, both `mean(binarycrossentropy.(...),2)` and `binarycrossentropy(...)` return the same result. We can leave it up to the user to determine which method to use. What do you think?

@CarloLucibello I've added `average` and additive factor as optional arguments. Docstring and tests have also been incorporated.

@MikeInnes I'll make the necessary changes to the docs.

MikeInnes:
Normal code shouldn't throw warnings; in that case we should just change this to take scalars only (and the `average` is then redundant).

The `add` argument would be clearer as `eps` as well. Otherwise, I think this looks good to go.

americast:
@MikeInnes Changed `add` to `eps`. Also, removed warnings for all scalar operations, even when `average` is set to `false`. The warning should appear when one is calculating loss for multi-class data in a non-scalar manner and has set `average` to `false`.

src/layers/stateless.jl
```diff
@@ -1,5 +1,6 @@
 using NNlib: log_fast
 
+const EPS = 1e-7
```
CarloLucibello: better to set this to 1f-7 so the Float32 arrays don't get promoted to Float64

americast:
@CarloLucibello Thanks for the suggestion. Updated it!

src/layers/stateless.jl
```diff
@@ -1,5 +1,6 @@
-using NNlib: logsoftmax, logσ
+using NNlib: logsoftmax, logσ, log_fast
 
+const EPS = 1f-7
```
staticfloat: @americast instead of defining `EPS` like this, you should probably use `eps(eltype(x))`; it's a builtin within Julia.  It will return an appropriate value depending on the datatype of the input:

```
julia> eps(Float32(1))
1.1920929f-7

julia> eps(Float64(1))
2.220446049250313e-16
```

americast:
Updated it! Thanks @staticfloat.

maetshju:
Is there any outstanding issue keeping this from being merged? I just spent a few hours tonight running into issues with binary cross entropy that it looks like the improved function here would have avoided.

maetshju:
I totally agree with your viewpoint in that comment. I should have been more clear though; my concern here was numerical stability, which the EPS value helped with.

MikeInnes:
Ah ok, that's easy enough to fix. Mind sending the patch?

CarloLucibello:
this has been staling for a while, I'm closing this but feel free to reopen

PR #147, opened 2018-01-10
optimize dropout

baggepinnen:
Avoids one pass through array `y`


MikeInnes:
Nice idea, the issue is that this needs to work with TrackedArrays and be differentiable in `x`, which is why we're using the vector operation `y .* x` for this. You could probably make this faster but would have to define both the forward and backward passes for it.

baggepinnen:
Ah too bad, never mind then ;)

Issue #22, opened 2017-04-27
Feature request: Smoothing splines with GCV


ghost:
It would be great if we could have an implementation of smoothing splines together with GCV in this framework, using the MXNet or TensorFlow backends for performance

MikeInnes:
RE this and #23: if this is easy to in either TensorFlow or MXNet then it should be pretty easy to add to Flux. Could you perhaps give an example of how that looks? If it's as simple as a function call it's very easy to add and I can help you with that.

alha02:
Well basically what I am looking for are these two packages
https://github.com/simonster/Lasso.jl
https://github.com/nignatiadis/SmoothingSplines.jl
but accelerated with something like TF or MXN running on GPUs.

As far as I know there are no calculations going on in there other than basic linear algebra, so as long as that is possible in the libs then I guess it should be possible to do here. Don't know how much work it would be to implement though..

But I did find this reference:
https://github.com/nfmcclure/tensorflow_cookbook,
where sections 6-8 in chapter 3 are the gist of the first package

alha02:
By the way, I'd love to help out in doing this, but I'm not quite sure how to get started. Could you perhaps give me some pointers as to how this could get integrated into Flux?

MikeInnes:
Sure, I should really write some docs on this stuff but here's a basics of it. So at the core of it you just want to write a function like

```julia
@net f(x, y) = x .* y
fm = mxnet(f)
fm(rand(5), rand(5))
```

With the inputs replaced with real inputs and the `.*` replaced by whatever linalg implements the algorithm you want.

All Flux needs to know is how to turn the linalg into a TensorFlow/MXNet graph. This is listed [here](https://github.com/MikeInnes/Flux.jl/blob/f50688e75cb9c1d1ae1074a6cf7bc54b134ef004/src/backend/mxnet/graph.jl#L22) for MXNet; you can see that we're just saying, given the `.*` function and two existing graphs (`mx.Symbol`s or `TensorFlow.Tensor`s), how do we use MXNet's API to add that op to the graph. So it's pretty simple.

So all you need to do is (1) write the algorithm you want in pure Julia (but using linear algebra that TensorFlow supports; it doesn't matter if it's slow), and (2) add `@net`, see where the conversion to TensorFlow fails, and add support for operations as required.

Let me know if you get stuck anywhere and I can try to get you going with this.

MikeInnes:
[...]

PR #153, opened 2018-01-21
Registering backward function for logsoftmax

boathit:
Registering backward function for `logsoftmax` for PR [#22](https://github.com/FluxML/NNlib.jl/pull/22). Test cases have passed but requires `logsoftmax` being added into NNlib.jl.


MikeInnes:
Thanks again @boathit!

PR #154, opened 2018-01-21
Dirty version of threadding

pevnak:
This a model-parallel version of threading I have implemented and I use for Flux. I have created because I desperately wanted to move out of TensorFlow and wanted to be faster. I would be very if someone can give me feedback, how to implement this more cleanly.

Thanks a lot.
Tomas


jekbradbury:
Just noting that this approach is usually termed "data parallelism," as distinct from "model parallelism" which shards the model over several [threads/processes/GPUs/nodes].

ViralBShah:
This looks like a pretty reasonable first start, given the current status of multi-threading.

pevnak:
I am glad you like it. If there is something how I can improve it, apart from changing terms (in which I am particularly bad), I would be happy to do so.

-------- Original Message --------

ViralBShah:
Waiting for @MikeInnes to chime in

MikeInnes:
This is cool! Do you have some idea of the speedups that you get with it?

Although it's rough, as you say, it's really useful to get an idea of what we need in Flux to support this better. One thing we'll need to think about is how to deal with `deepcopy(model)`; most Flux models, being Julia programs, are not easy to just make a copy of. Avoiding global state will help with this, and code-tracing approaches (which we might use for other optimisations) will solve it completely.

pevnak:
Hi Mike,

I will try to measure the speedup today.

I was thinking about Mike's argument regarding the deep copy and i have to say I do not understand it 100%. If I understand correctly, Flux creates the graph, but it always made from scratch and it should be unique to the combination of input and parameters of the network. From the source code I got the intuition that all other structures are allocated dynamically. This means that the dirty solution fails if deepcopy fails to copy the argument. What are the models (structures), where this can happen?

Best wishes,
Tomas

-------- Original Message --------

MikeInnes:
It's actually the exact same problem that you've already run into; technically the whole model is defined by `loss(x, y) => l`, but `deepcopy(loss)` won't work with global variables, so you need to explicitly copy the model state. This works in that case, but it's harder to deal with [this model](https://github.com/FluxML/model-zoo/blob/master/phonemes/1-model.jl) for example, where the state is more scattered.

There are a couple of potential solutions to this. One is to explicitly pass in model state (`loss(x, y, m) = mse(m(x), y)`). Another is to explicitly make layers for everything, like PyTorch. Yet another is to have a compilation step that traces through the code.

For now, I'd be happy enough to say that if you want threading, you have to make a custom layer, in which this API would work as is.

ToucheSir:
https://github.com/FluxML/DaggerFlux.jl/ now exists as a modern incarnation of model parallel training.

PR #156, opened 2018-01-24
Add support for `cat`

### LINKED ISSUE #142: Derivative of `cat`
We currently have `vcat` but could use the more general version.

americast:
@MikeInnes Adds generic `cat` to Flux. Fixes #142.
The tests have been provided separately for `cat` as it needs a parameter which is not tracked.


MikeInnes:
Looks good!

You don't need a new function here, however. You can just do `gradtest((a, b) -> cat(2, a, b), ...)` etc, as other tests for similar functions do.

americast:
Thanks, @MikeInnes! Fixed it.

MikeInnes:
Ok so two comments on this:

(1) You can avoid array allocations (and type instability) by working with tuples for `till_xs` and `xs_in_Δ`. e.g. `ntuple(i -> _, Val{ndims(Δ)})`.
(2) Can you extend this to `n` arguments? Doesn't have to be done as part of this PR, unless it's easy to do.

MikeInnes:
Oh, and here's a test case that fails: `@test gradtest((a, b) -> cat(2, a, b), rand(3), rand(3))`.

MikeInnes:
Merged with #243 -- thanks :+1:

PR #165, opened 2018-02-06
Register back! for logsigmoid and implement (logit)binarycrossentropy

boathit:
* Registering the `back!` function for  the `logsigmoid` in NNlib (requiring new tag)
* Implementing the `binarycrossentropy` and `logitbinarycrossentropy` using the `logsigmoid`
* `logitcrossentropy` is changed to use `logsoftmax`
* Unifying the code style in `test/layers/stateless.jl` and adding more test cases.

The `binarycrossentropy` and `logitbinarycrossentropy` are implemented as scalar functions and the users can broadcast to compute the batch loss.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,8 +3,13 @@ DataFlow 0.2.1
 Juno
 MacroTools 0.3.3
 NNlib
-ForwardDiff 0.5.0
 Requires
 Adapt
 GZip
 Colors
+
+# AD
+ForwardDiff 0.5.0
+DiffRules
+SpecialFunctions
+NaNMath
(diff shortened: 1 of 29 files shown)


MikeInnes:
Great stuff. I just opened https://github.com/JuliaLang/METADATA.jl/pull/13341 so we can merge once that's done.

MikeInnes:
By the way, I removed the `logsigmoid` integration as it wasn't necessary / actually doing anything. Since it's a scalar kernel ForwardDiff just handles it for free :)

boathit:
I see. ForwardDiff can offer derivative for any scalar function without bringing any overhead, that's great. But the `logsigmoid` itself is still necessary, right? My point is that the separated treatment of `sigmoid` and `log` operations still has numerical stability issue.

MikeInnes:
Yeah, sure. I haven't actually removed `logsigmoid`, you'll see it's still exported from Flux now, we just don't need any special handling for it.

Issue #137, opened 2018-01-02
Feature request: Conv3D


SimonEnsemble:
We need a `Conv3D` layer, where the filters are 3D arrays (e.g. 10 by 10 by 10) and these are convolved over another 3D array (e.g 100 by 100 by 100) to yield a e.g. (90, 90, 90) map (assuming no padding). Would this be possible in Flux.jl?

MikeInnes:
It would certainly be possible, but we don't currently have any support for it. We'd need either a pure-Julia convolution operator for CPU, or a CUDNN wrapper for GPU. If someone wants to put this together it'd be a welcome addition.

tejank10:
Hi! I would like to take a crack at it. I am working on adding a pure-Julia convolution operator for CPU.

MikeInnes:
Great! Let me know how I can help. Might be useful to have [ImageFiltering.jl](https://github.com/JuliaImages/ImageFiltering.jl) as a reference.

tejank10:
I was wondering, wouldn't it be a good idea of adding 3D versions of `im2col` and other functions from `conv.cpp`?

MikeInnes:
That would certainly work and would be an acceptable approach. I think it'd be much easier to write something in Julia though, and it'd be easy to make that handle any number of dimensions.

tejank10:
Hey, @MikeInnes! I have opened PR [#27](https://github.com/FluxML/NNlib.jl/pull/27) in NNlib. Please review it.

tejank10:
Is there anything wrong with the PR?

MikeInnes:
Thanks to @tejank10 this will be in the next release!

PR #166, opened 2018-02-06
Conv3D layer

tejank10:
Added Conv3D layer (#137 )


ViralBShah:
Bump.

MikeInnes:
Looking great! x-ref https://github.com/FluxML/NNlib.jl/pull/27, where we can discuss this.

MikeInnes:
Thanks for the effort; I actually added this in https://github.com/FluxML/Flux.jl/commit/15d1d3256b87c1e5caadda5415d99d3d824a8cb7, with the idea being to have a dimension-agnostic API so we don't end up with `Conv4d`, `Conv5D` etc.

PR #168, opened 2018-02-07
Separate AD infrastructure from types

MikeInnes:
This PR basically separates the core AD functionality (provided by `Tracked`) from the frontend types (`TrackedArray`, `TrackedScalar`) etc., which provide specific operations. This allows us to add tracking to any kind of object – I've already separated scalar tracking from 0-D array tracking, but we can also do things like tracking tuples, which is useful for primitives that return multiple values.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,8 +3,13 @@ DataFlow 0.2.1
 Juno
 MacroTools 0.3.3
 NNlib
-ForwardDiff 0.5.0
 Requires
 Adapt
 GZip
 Colors
+
+# AD
+ForwardDiff 0.5.0
+DiffRules
+SpecialFunctions
+NaNMath
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -19,7 +19,7 @@ export σ, sigmoid, relu, leakyrelu, elu, swish, softmax, logsoftmax,
 include("tracker/Tracker.jl")
 using .Tracker
 export Tracker
-import .Tracker: data, value
+import .Tracker: data
 
 include("optimise/Optimise.jl")
 using .Optimise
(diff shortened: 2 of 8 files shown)


iblis17:
>This allows us to add tracking to any kind of object

Could you provide an example usage?

MikeInnes:
In terms of adding tracking to types, all of the `Tracked*` types (`Array`, `Tuple`, `Number`) are now examples of that.

In terms of tracking something like a tuple-valued function:

```julia
import Flux.Tracker: TrackedArray, track, back!, back

twice(x) = (x, x)
twice(x::TrackedArray) = track(twice, x)
back(::typeof(twice), Δ, x) = back(x, Δ[1] + Δ[2])

x = param([1,2,3])
xs = twice(x)
l = sum(xs[1] + xs[2])
back!(l)
x.grad
```

The most significant thing here is that `Δ` is accumulated as a tuple and `back(twice, ...)` is only called once. If `twice` is a black box function (like CUDNN's RNNs) this makes the back pass more efficient.

iblis17:
Seems `params` does not work for new AD infra
```julia
julia> params(xs)
0-element Array{Any,1}
```

patch:
```diff
diff --git a/src/treelike.jl b/src/treelike.jl
index b4c9a77..5139125 100644
--- a/src/treelike.jl
+++ b/src/treelike.jl
@@ -33,9 +33,11 @@ function prefor(f, x; seen = OSet())
   return
 end

+using Flux.Tracker: istracked
+
 function params(m)
   ps = []
-  prefor(p -> p isa TrackedArray && push!(ps, p), m)
+  prefor(p -> istracked(p) && push!(ps, p), m)
   return ps
 end
```

MikeInnes:
Whoops, better add a test for that. Fixed in https://github.com/FluxML/Flux.jl/commit/f7fdfbe3a98bab8b59208b84a8a22d7583b77320.

PR #169, opened 2018-02-08
Reshape with Tuple Dimensions and Kronecker Product 

### LINKED ISSUE #149: Kronecker Product on Tracked Arrays No Method Matching *
Trying to do a Kronecker Product of two tracked arrays and have a no method matching for `*`:
```
kron(param(eye(2)),param(eye(2)))
#=> MethodError: no method matching *(::TrackedArray{…,Array{Float64,0}}, ::TrackedArray{…,Array{Float64,0}})
```

jessebett:
Fixes #149 
(though not in the more clever `∂(X⊗Y) = (∂X)⊗Y + X⊗(∂Y)` we also discussed.

### DIFF
--- a/src/tracker/array.jl
+++ b/src/tracker/array.jl
@@ -106,8 +106,22 @@ end
 Base.reshape(xs::TrackedArray, dims::Union{Colon,Int64}...) =
   track(reshape, xs, dims...)
 
+Base.reshape(xs::TrackedArray, dims::Tuple{Vararg{Int64,N}} where N) =
+  track(reshape, xs, dims)
+
 back(::typeof(reshape), Δ, xs::TrackedArray, _...) =
   back(xs, reshape(Δ, size(xs)))
+
+
+function Base.kron(mat1::AbstractMatrix,mat2::AbstractMatrix)
+    m1, n1 = size(mat1)
+    mat1_rsh = reshape(mat1,(1,m1,1,n1))
+
+    m2, n2 = size(mat2)
+    mat2_rsh = reshape(mat2,(m2,1,n2,1))
+
+    return reshape(mat1_rsh.*mat2_rsh, (m1*m2,n1*n2))
+end
 
 # Reductions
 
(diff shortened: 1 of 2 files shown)


MikeInnes:
Awesome, thanks! Can you add a quick gradient test to this? Should be easy following [these](https://github.com/FluxML/Flux.jl/blob/master/test/tracker.jl).

jessebett:
So I added some tests which made me aware of some confusing things (which I guess is the point of writing tests...)

First, I realized that my implementation of `kron` doesn't take Vector arguments, where `Base.kron` does, because it tries to access the second dimension. I guess this is what you meant by the `∂(X⊗Y) = (∂X)⊗Y + X⊗(∂Y)` implementation being more clever by not caring about dimensions.

However, after I realized this I was confused because `@test gradtest(kron,rand(5), rand(3))` passed? This can't be my `kron` method though, so how is this passing?

Further, this only passes sometimes, every tenth or so time I run the test it doesn't pass. 
What's concerning is this is true for all the tests. I was running them repeatedly and most of the time they were passing, but randomly they'd fail. 

Obviously this is now a different issue I'm describing. I think my kron implementation passes, it fails as infrequently as the others. Though I don't know why the `kron` on `Vector` arguments is working. Is mine now unnecessary?

MikeInnes:
Julia defines [these methods for vectors], so we get that for free :)

I didn't realise that Julia only defines this for matrices anyway, in which case it's not a big deal if we only support matrices.

MikeInnes:
I've updated this slightly; we shouldn't overwrite the base method, but only use our version when TrackedArrays are involved.

Otherwise, this is perfect, thanks a lot @jessebett!

jessebett:
Okay! I now see how to avoid overwriting base, cool!

At some point later I might try to implement `∂(X⊗Y) = (∂X)⊗Y + X⊗(∂Y)` but I think it'll require a bit more understanding of how Flux writes the `back` rules.

Thanks @MikeInnes!
(this was my first ever PR)

MikeInnes:
Awesome, great first PR!

BTW, looking at it again I'm not sure if that rule will help us here; you'd have to effectively solve for `dX` given `dY` and `d(XxY)` which requires a definition of the inverse of Kronecker product. That said, I'm not sure it even matters as the clever broadcast version looks pretty efficient to me. (and also looks like it would generalise to N-dimensions if anyone's up for a challenge).

I'm planning to write up some docs on the AD stuff though.

PR #170, opened 2018-02-11
treelike: fix params for TrackedTuple

iblis17:
See also: https://github.com/FluxML/Flux.jl/commit/0f7a1ec0229b80687f7548e91e88534630661025#commitcomment-27481941

### DIFF
--- a/src/tracker/scalar.jl
+++ b/src/tracker/scalar.jl
@@ -66,6 +66,7 @@ struct TrackedTuple{T<:Tuple}
 end
 
 tracker(xs::TrackedTuple) = xs.tracker
+isleaf(xs::TrackedTuple) = true
 
 accum!(x::Tuple, Δ::Tuple) = accum!.(x, Δ)
 init_grad(x::Tuple) = init_grad.(x)
--- a/test/tracker.jl
+++ b/test/tracker.jl
@@ -1,5 +1,6 @@
 using Flux.Tracker, Base.Test, NNlib
-using Flux.Tracker: TrackedReal, gradcheck
+using Flux.Tracker: TrackedArray, TrackedReal, TrackedTuple,
+                    @back, back!, track, gradcheck
 using NNlib
 
 gradtest(f, xs::AbstractArray...) = gradcheck((xs...) -> sum(sin.(f(xs...))), xs...)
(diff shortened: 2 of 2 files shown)


MikeInnes:
What's your use case for this? This seems like a very strange semantics, given that we don't look at the history of `Tracked*` objects for this in other cases.

Tracked tuples really only make as intermediates rather than parameters, but if someone does put a `TrackedTuple` directly in a layer for whatever reason, it seems most sensible just to treat that object as a parameter in itself.

iblis17:
> Tracked tuples really only make as intermediates rather than parameters, but if someone does put a `TrackedTuple` directly in a layer for whatever reason, it seems most sensible just to treat that object as a parameter in itself.

I will treat TrackedTuple as a compact version of `struct` 
(also, we will have NamedTuple in 0.7+, I think something like `(W = ..., b = ...)` is useful, for instance).
So, that's why I make `params` iter the tuple and extract `TrackedArray` (in case of this PR).

MikeInnes:
Why not do a (named) tuple of tracked arrays in that case? What does it buy you?

iblis17:
> Tracked tuples really only make as intermediates rather than parameters, but if someone does put a TrackedTuple directly in a layer for whatever reason, it seems most sensible just to treat that object as a parameter in itself.

> Why not do a (named) tuple of tracked arrays in that case? What does it buy you?

Oh, I read my code again 🤦‍. Current implementation is problematic, as you said, `params` should add TrackedTuple itself -- otherwise it will lose the history of `twice` call.

I will update my PR.

iblis17:
updated.

MikeInnes:
Ok, so why do you need to add `isleaf` here?

iblis17:
`params` will check it.
https://github.com/FluxML/Flux.jl/blob/5ea0ef6764c191a400ca27276e6435547ceca0b0/src/treelike.jl#L36-L43

MikeInnes:
Ok, but again, it seems very strange to say this here; if a TrackedTuple isn't a leaf node I'm not sure why we'd pretend that it is.

It's still very unclear to me why you want this. Why aren't you using a tuple of tracked arrays?

iblis17:
> It's still very unclear to me why you want this. Why aren't you using a tuple of tracked arrays?

Well, sorry, I do not have a practical use case.
I just like the idea of black box function and user-defined backward phase (I like the `twice` exmaple).
And I want it works with `params` and optimisers.

Maybe like this?
```julia
struct Foo
  pair
  
  Foo(x) = new(twice(param(x)))
end

struct Bar
  a
  b
  pair

  Bar = new(..., ..., twice(param(x)))
end

back(typeof::twice, ...) = #= custom backward phase =#
```
Thus, we have a `twice` function sharing the logic and abstraction.

iblis17:
> if a TrackedTuple isn't a leaf node I'm not sure why we'd pretend that it is.

Or any better `params` checking logic?

iblis17:
> Ok, but again, it seems very strange to say this here

I'm also ok with just closing this PR, since I do not have any urgent need on it.

MikeInnes:
In that case I think you'd be better off doing
```julia
struct Foo
  x
  
  Foo(x) = new(param(x))
end

function (f::Foo)
  twice_x = twice(f.x)
  ...
end
```

If you wanted to support this kind of thing consistently you'd have to make it work for arrays as well, e.g.

```julia
struct Foo
  x
  Foo(x) = new(param(x).^2)
end
```

And do similar things with looking inside the history of `x`. But I'm not sure we want to go down that route.

iblis17:
That's better way 👍 .
(and I still cannot found any counter-example)

MikeInnes:
Thanks for the discussion anyway!

PR #174, opened 2018-02-16
added simple support for regularization of output of individual layers

pevnak:
This is a simple modification of Chain, which can be used for regularization of output of layers, as is proposed for example in
Generalization in Deep Learning, Kenji Kawaguchi, Leslie Pack Kaelbling, Yoshua Bengio, 2017
https://arxiv.org/abs/1710.05468

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -28,6 +28,7 @@ include("treelike.jl")
 
 include("layers/stateless.jl")
 include("layers/basic.jl")
+include("layers/regularisedchain.jl")
 include("layers/conv.jl")
 include("layers/recurrent.jl")
 include("layers/normalisation.jl")
--- a/src/layers/regularisedchain.jl
+++ b/src/layers/regularisedchain.jl
@@ -1,0 +1,79 @@
+"""
+    RegularisedChain(layers,regularizations)
+    RegularisedChain(regularization,layers...)
+
+Chain multiple layers / functions together, so that they are called in sequence
+on a given input.
+
+If the chain is called with additional scalar parameter, appropriate regulariser is 
+called on output of each layer. Regularised chain then returns Tuple, where the first 
+item is the output of the Chain and the second is the value of regularization.
+
+An example where this construct might be useful is 
+Generalization in Deep Learning, Kenji Kawaguchi, Leslie Pack Kaelbling, Yoshua Bengio, 2017
+https://arxiv.org/abs/1710.05468
+
+```julia
+import Flux: RegularisedChain, nullregularizer, l1
+m = RegularisedChain([Dense(10, 5), Dense(5, 2)],[nullregularizer,l1])
+m = Flux.RegularisedChain(Flux.l1,Dense(10, 5), Dense(5, 2))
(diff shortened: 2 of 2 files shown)


GenaBitu:
The use of "regularise" and "regularize" is inconsistent.
Here is patch for z -> s, since that's prevailing:
[0001-z-s-for-consistency.txt](https://github.com/FluxML/Flux.jl/files/1751752/0001-z-s-for-consistency.txt)

MikeInnes:
Here's an idea for doing this that doesn't need a new layer type. We add an `activations` function (name open to suggestions) like so:

```julia
# Not sure why we need this
Base.rcum_promote_type(op, ::Type, ::Type{Any}) = Any

m = Chain(Dense(10, 5, relu), Dense(5, 2), softmax)

activations(c::Chain, x) = accumulate((x, m) -> m(x), x, c.layers)

activations(m, rand(10))
```

Then you can just regularise those activations as you normally would. Does that get you what you need?

pevnak:
This might work, though it might be slightly messy if we want the regularisation applied only for some layers.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

MikeInnes:
I think that can be a fairly clean broadcast, e.g. `call.(regularisers, activations)` (tell me if I'm missing something though).

MikeInnes:
I added this feature in https://github.com/FluxML/Flux.jl/commit/0a04e3ba61b0cb951b41184971f30d6c41dbf510

```
julia> c = Chain(Dense(10,5,σ),Dense(5,2),softmax)
Chain(Dense(10, 5, NNlib.σ), Dense(5, 2), NNlib.softmax)

julia> Flux.activations(c, rand(10))
3-element Array{Any,1}:
 param([0.403053, 0.468468, 0.550944, 0.319838, 0.566326])
 param([0.658583, -0.0318164])
 param([0.666056, 0.333944])
```

If this has problems for your use case, let me know and we can figure out if more utilities are needed here.

pevnak:
Hi Mike,

thank you very much. This seems to be very neat solution and it should work nicely.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

MikeInnes:
Great, in that case I'll close this for now. Happy to hear any other feedback.

johnnychen94:
In Flux v0.6.1, the last tutorial code sample in [regularisation](https://github.com/FluxML/Flux.jl/blob/d6a75e1289488945ade1cfa8867717fe1ed19557/docs/src/models/regularisation.md) fails

```julia
using Flux
c = Chain(Dense(10,5,σ),Dense(5,2),softmax)
Flux.activations(c, rand(10))
```
which raises `ERROR: MethodError: no method matching accumulate(::getfield(Flux, Symbol("##62#63")), ::Array{Float64,1}, ::Array{Any,1})`

It seems that `accumulate(op, A; dims::Integer, [init])` doesn't support multiple inputs as you wrote in `activations(c::Chain, x) = accumulate((x, m) -> m(x), x, c.layers)` anymore.

PR #181, opened 2018-02-21
introduce Reexport

### LINKED ISSUE #180: Exporting `softplus`
Is there a reason why Flux doesn't export `softplus`  like `relu` and the other standard activations?
```julia
julia> using Flux

julia> softplus
ERROR: UndefVarError: softplus not defined

julia> Flux.softplus
softplus (generic function with 1 method)

julia> relu
relu (generic function with 1 method)
```

iblis17:
- Reexporting NNlib

fix #180

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -9,6 +9,7 @@ GZip
 Colors
 ZipFile
 AbstractTrees
+Reexport
 
 # AD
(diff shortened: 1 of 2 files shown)


MikeInnes:
Yeah, that makes sense. Thanks!

PR #184, opened 2018-02-25
slight generalization of cat

pevnak:
I have generalized cat to allow multiple dimensions, since at the moment it cannot handle 4-dim arrays in convolutional functions. 

At the moment, the backward path breaks with
`cat(2,rand(4),rand(4))`

Appropriate tests are added.


MikeInnes:
Merged with #243 -- thanks :+1:

PR #187, opened 2018-02-28
Add `permutedims()` for tracked arrays

staticfloat:
@MikeInnes is this all that needs to happen here?  Will autograd correctly track the permutation?


MikeInnes:
Awesome, thanks! Can you add a quick grad check?

staticfloat:
Add the grad check, as far as I can tell.

MikeInnes:
Perfect, thanks! I'm hoping to document the AD stuff a bit before the next release, so this is all a bit more obvious.

PR #191, opened 2018-03-02
Use `cache.julialang.org` to store CMU phones

staticfloat:
It's annoying that when third party servers go down our tests break. Let's at least make sure that if our tests break due to server outages it's our fault.


staticfloat:
@MikeInnes the Flux tests always fail for me locally because the server you try to download the CMU phones from has gone down.  Can you either email me a local copy of the data so I can put it onto the `cache.julialang.org` server and then merge this PR, or switch it over to a new source that works?

MikeInnes:
:+1: Thanks!

PR #197, opened 2018-03-06
Implement `prod` for `TrackedArray`

chengchingwen:
Implementation of the reduce product `prod` function for `TrackedArray`

### DIFF
--- a/src/tracker/array.jl
+++ b/src/tracker/array.jl
@@ -139,6 +139,13 @@ Base.sum(f::Union{Function,Type},xs::TrackedArray) = sum(f.(xs))
 
 back(::typeof(sum), Δ, xs::TrackedArray, dim...) = back(xs, similar(xs.data) .= Δ)
 
+Base.prod(xs::TrackedArray, dim) = track(prod, xs, dim)
+Base.prod(xs::TrackedArray) = track(prod, xs)
+Base.prod(f::Union{Function, Type}, xs::TrackedArray) = prod(f.(xs))
+
+back(::typeof(prod), Δ, xs::TrackedArray, dim...) = back(xs, similar(xs.data) .= (prod(xs.data, dim) ./ xs.data) .* Δ)
+back(::typeof(prod), Δ, xs::TrackedArray) = back(xs, similar(xs.data) .= (prod(xs.data) ./ xs.data) .* Δ)
+
 Base.maximum(xs::TrackedArray, args...) = maximum(xs.data, args...)
 Base.findfirst(xs::TrackedArray, args...) = findfirst(xs.data, args...)
 
--- a/src/tracker/array.jl
+++ b/src/tracker/array.jl
@@ -144,7 +144,7 @@ Base.prod(xs::TrackedArray) = track(prod, xs)
 Base.prod(f::Union{Function, Type}, xs::TrackedArray) = prod(f.(xs))
 
 back(::typeof(prod), Δ, xs::TrackedArray, dim...) = back(xs, similar(xs.data) .= (prod(xs.data, dim...) ./ xs.data) .* Δ)
-back(::typeof(prod), Δ, xs::TrackedArray) = back(xs, similar(xs.data) .= (prod(xs.data) ./ xs.data) .* Δ)
+back(::typeof(prod), Δ, xs::TrackedArray) = back(xs, similar(xs.data) .= (reshape(.*(circshift.([reshape(xs.data, length(xs.data))], 1:length(xs.data)-1)...), size(xs.data))) .* Δ)
 
 Base.maximum(xs::TrackedArray, args...) = maximum(xs.data, args...)
 Base.findfirst(xs::TrackedArray, args...) = findfirst(xs.data, args...)
--- a/src/tracker/array.jl
+++ b/src/tracker/array.jl
@@ -143,7 +143,7 @@ Base.prod(xs::TrackedArray, dim) = track(prod, xs, dim)
 Base.prod(xs::TrackedArray) = track(prod, xs)
 Base.prod(f::Union{Function, Type}, xs::TrackedArray) = prod(f.(xs))
 
-back(::typeof(prod), Δ, xs::TrackedArray, dim...) = back(xs, similar(xs.data) .= (prod(xs.data, dim) ./ xs.data) .* Δ)
+back(::typeof(prod), Δ, xs::TrackedArray, dim...) = back(xs, similar(xs.data) .= (prod(xs.data, dim...) ./ xs.data) .* Δ)
 back(::typeof(prod), Δ, xs::TrackedArray) = back(xs, similar(xs.data) .= (prod(xs.data) ./ xs.data) .* Δ)
 
 Base.maximum(xs::TrackedArray, args...) = maximum(xs.data, args...)
--- a/test/tracker.jl
+++ b/test/tracker.jl
@@ -16,6 +16,8 @@ gradtest(f, dims...) = gradtest(f, rand.(dims)...)
 @test gradtest((w, x) -> w*x', randn(5,5), randn(5,5))
 
 @test gradtest(x -> sum(x, (2, 3)), (3,4,5))
+@test gradtest(x -> prod(x, (2, 3)), (3,4,5))
+@test gradtest(x -> prod(x), (3,4,5))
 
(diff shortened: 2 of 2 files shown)


MikeInnes:
Cool! Can you add a gradient check, like the ones in test/tracker.jl?

chengchingwen:
I'm not sure how does gradient check work, should it be something like `@test gradtest(x -> prod(x, (2, 3)), (3,4,5))` that mimic from `sum`

MikeInnes:
Yup, that looks right. It'd be work adding a check for the version without `dims` as well

chengchingwen:
But I got this for `@test gradtest(x -> prod(x, (2, 3)), (3,4,5))`
```
Tracker: Error During Test
  Test threw an exception of type ArgumentError
  Expression: gradtest((x->begin 
            prod(x, (2, 3))
        end), (3, 4, 5))
  ArgumentError: reduced dimension(s) must be integers
  Stacktrace:
   [1] reduced_indices(::Tuple{Base.OneTo{Int64},Base.OneTo{Int64},Base.OneTo{Int64}}, ::Tuple{Tuple{Int64,Int64}}) at ./reducedim.jl:35
   [2] reducedim_initarray(::Array{Float64,3}, ::Tuple{Tuple{Int64,Int64}}, ::Float64, ::Type{Float64}) at ./reducedim.jl:73
   [3] mapreducedim at ./reducedim.jl:241 [inlined]
   [4] prod at ./reducedim.jl:570 [inlined]
   [5] prod at ./reducedim.jl:572 [inlined]
   [6] back(::Base.#prod, ::Array{Float64,3}, ::TrackedArray{…,Array{Float64,3}}, ::Tuple{Int64,Int64}) at /home/peter/work/julia-pkg/Flux.jl/test/../src$
tracker/array.jl:146
   [7] back_(::Flux.Tracker.Call{Base.#prod,Tuple{TrackedArray{…,Array{Float64,3}},Tuple{Int64,Int64}}}, ::Array{Float64,3}, ::Array{Float64,3}) at /home$peter/work/julia-pkg/Flux.jl/test/../src/tracker/back.jl:25
   [8] back(::Flux.Tracker.Tracked{Array{Float64,3}}, ::Array{Float64,3}) at /home/peter/work/julia-pkg/Flux.jl/test/../src/tracker/back.jl:39
   [9] macro expansion at /home/peter/work/julia-pkg/Flux.jl/test/../src/tracker/back.jl:50 [inlined]
   [10] (::Flux.Tracker.##50#52)(::TrackedArray{…,Array{Float64,3}}, ::Array{Float64,3}) at /home/peter/work/julia-pkg/Flux.jl/test/../src/tracker/array.$l:306
   [11] foreach(::Function, ::Tuple{TrackedArray{…,Array{Float64,3}}}, ::Tuple{Array{Float64,3}}, ::Vararg{Tuple{Array{Float64,3}},N} where N) at ./abstr$ctarray.jl:1732
   [12] back_(::Flux.Tracker.Call{Flux.Tracker.Broadcasted{Base.#sin,Array{ForwardDiff.Dual{Void,Float64,1},3}},Tuple{TrackedArray{…,Array{Float64,3}}}}, ::Array{Float64,3}, ::Array{Float64,3}) at /home/peter/work/julia-pkg/Flux.jl/test/../src/tracker/back.jl:25
   [13] back(::Flux.Tracker.Tracked{Array{Float64,3}}, ::Array{Float64,3}) at /home/peter/work/julia-pkg/Flux.jl/test/../src/tracker/back.jl:39
   [14] back_(::Flux.Tracker.Call{Base.#sum,Tuple{TrackedArray{…,Array{Float64,3}}}}, ::Float64, ::Float64) at /home/peter/work/julia-pkg/Flux.jl/test/..$src/tracker/back.jl:25
   [15] back(::Flux.Tracker.Tracked{Float64}, ::Int64) at /home/peter/work/julia-pkg/Flux.jl/test/../src/tracker/back.jl:37
   [16] back!(::Flux.Tracker.TrackedReal{Float64}) at /home/peter/work/julia-pkg/Flux.jl/test/../src/tracker/scalar.jl:11
   [17] gradient(::Function, ::Array{Float64,3}, ::Vararg{Array{Float64,3},N} where N) at /home/peter/work/julia-pkg/Flux.jl/test/../src/tracker/numeric.$
l:3
   [18] gradcheck(::Function, ::Array{Float64,3}, ::Vararg{Array{Float64,3},N} where N) at /home/peter/work/julia-pkg/Flux.jl/test/../src/tracker/numeric$
jl:22
   [19] macro expansion at /home/peter/work/julia-pkg/Flux.jl/test/tracker.jl:20 [inlined]
   [20] macro expansion at ./test.jl:860 [inlined]
   [21] anonymous at ./<missing>:?
```
looks weird because it call `prod` with `Tuple{Tuple{Int,Int}}` dimension type.

chengchingwen:
We Just found another issue that this implement might have some trouble when `xs` is really small
@iblis17

iblis17:
Alternative of `prod(xs.data) ./ xs.data`
```julia
reshape(.*(circshift.([reshape(A, length(A))], 1:length(A)-1)...), size(A))
```

src/tracker/array.jl
```diff
@@ -139,6 +139,13 @@ Base.sum(f::Union{Function,Type},xs::TrackedArray) = sum(f.(xs))
 
 back(::typeof(sum), Δ, xs::TrackedArray, dim...) = back(xs, similar(xs.data) .= Δ)
 
+Base.prod(xs::TrackedArray, dim) = track(prod, xs, dim)
+Base.prod(xs::TrackedArray) = track(prod, xs)
+Base.prod(f::Union{Function, Type}, xs::TrackedArray) = prod(f.(xs))
+
+back(::typeof(prod), Δ, xs::TrackedArray, dim...) = back(xs, similar(xs.data) .= (prod(xs.data, dim) ./ xs.data) .* Δ)
```
CarloLucibello: should this be `back(::typeof(prod), Δ, xs::TrackedArray, dim)`?

CarloLucibello:
It would be more efficient here to make use of the output value computed in the forward pass to backpropagate, instead of computing it againg. Digging in the library it appears that this can be done through the `back_` method, so this would be
```
back_(::typeof(prod), y, Δ, xs::TrackedArray, dim) = back(xs, similar(xs.data) .= (y ./ xs.data) .* Δ)
```
@MikeInnes why we have both `back_` and `back`? wouldn't be better to just use the `back_`-like signature?

src/tracker/array.jl
chengchingwen: I just followed the implementation of `sum`. I think they should be the same.

src/tracker/array.jl
```diff
@@ -139,6 +139,13 @@ Base.sum(f::Union{Function,Type},xs::TrackedArray) = sum(f.(xs))
 
 back(::typeof(sum), Δ, xs::TrackedArray, dim...) = back(xs, similar(xs.data) .= Δ)
 
+Base.prod(xs::TrackedArray, dim) = track(prod, xs, dim)
+Base.prod(xs::TrackedArray) = track(prod, xs)
+Base.prod(f::Union{Function, Type}, xs::TrackedArray) = prod(f.(xs))
+
+back(::typeof(prod), Δ, xs::TrackedArray, dim...) = back(xs, similar(xs.data) .= (prod(xs.data, dim...) ./ xs.data) .* Δ)
```
CarloLucibello: why `dim...` instead of `dim`? `prod` takes only 2 arguments

chengchingwen: The first `dim...` in argument follows the implementation of `sum`
https://github.com/FluxML/Flux.jl/blob/9ccbac8b80cc3dc50f0fdbf02063d811add7b539/src/tracker/array.jl#L140
and the second `dim...` is to pass the gradient check `@test gradtest(x -> prod(x, (2, 3)), (3,4,5))`

CarloLucibello: ok, both `dim` and `sum` can take one or two arguments, so `dims...`  is the easiest way to go

MikeInnes:
Thanks @chengchingwen!

@CarloLucibello the main reason is that most of the functions don't need that extra information. It's not really a big deal either way, but it makes a lot of the code just a bit nicer.

I'm planning to add some AD docs that explain all this at some point.

PR #207, opened 2018-03-16
BatchNorm for convolutions

safnuk:
These changes allow batch norm to be used with convolutional layers (of any dimension), in addition to dense layers, all with the same api.


safnuk:
This also fixes issue #216.

MikeInnes:
This is great, thanks a lot!

iblis17:
Hi,
Any idea about that sample code in docstring doesn't work now?
```julia
julia> m = Chain(
           Dense(28^2, 64),
           BatchNorm(64, relu),
           Dense(64, 10),
           BatchNorm(10),
           softmax)
Chain(Dense(784, 64), BatchNorm(64, λ = NNlib.relu), Dense(64, 10), BatchNorm(10), NNlib.softmax)

julia> m(rand(784))
ERROR: arraysize: dimension out of range
Stacktrace:
 [1] (::Flux.BatchNorm{NNlib.#relu,TrackedArray{…,Array{Float64,1}},Array{Float64,1},Float64})(::TrackedArray{…,Array{Float64,1}}) at /home/iblis/.julia/v0.6/Flux/src/layers/normalise.jl:117
 [2] mapfoldl_impl(::Base.#identity, ::Flux.##81#82, ::Array{Float64,1}, ::Array{Any,1}, ::Int64) at ./reduce.jl:46
 [3] (::Flux.Chain)(::Array{Float64,1}) at /home/iblis/.julia/v0.6/Flux/src/layers/basic.jl:31
```

safnuk:
`Dense` expects a 2D array as input, not a vector. You should be able to evaluate `m(rand(784, 1))`.

MikeInnes:
In general we do actually support dropping the trailing batch dimension, but I that's a case where batchnorm doesn't make sense – we probably just need an improved error message here.

PR #208, opened 2018-03-16
Added CIFAR10 dataset.

fissoreg:
Here is a working example using CIFAR10: https://github.com/fissoreg/model-zoo/blob/gans/gan/cifar10.jl

Also, @staticfloat  may want to cache the dataset on https://cache.julialang.org/.

### DIFF
--- a/src/data/Data.jl
+++ b/src/data/Data.jl
@@ -11,6 +11,7 @@ function __init__()
 end
 
 include("mnist.jl")
+include("cifar10.jl")
 include("cmudict.jl")
 using .CMUDict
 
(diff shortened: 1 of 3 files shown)


Keno:
Didn't see this before, but I independently added ImageNet and CIFAR10 support to Metalhead: https://github.com/FluxML/Metalhead.jl/pull/4

MikeInnes:
I think it's probably better to avoid the duplication here. One nice thing about having a separate package is that we can easily start adding other useful tools there (e.g. normalisation routines) without adding things like Images.jl to Flux. For small/simple/non-image datasets, Flux itself can still make sense.

fissoreg:
> One nice thing about having a separate package is that we can easily start adding other useful tools there (e.g. normalisation routines) without adding things like Images.jl to Flux.

This makes a lot of sense. Also, [MLDatasets.jl](https://github.com/JuliaML/MLDatasets.jl) could be a better choice to import the datasets to Flux, so no need to merge this PR.

PR #209, opened 2018-03-17
WIP - Maxout for dense layer.

americast:
Implementation of `maxout` for `Dense` layer. ([Link to paper](https://arxiv.org/pdf/1302.4389.pdf)).

This implementation of maxout is specific to a particular layer. I've worked with `Dense` here. The architecture involves keeping `n` sets of weights and biases and obtaining `n` variants of the output. The maximum across the second axis is taken forward. For instance, a `Dense(5, 4, maxout = 3)` looks like:
Maxout
(biases have been ignored in the image for simplicity).

If this looks good, I can proceed to add support for batching and integrate `maxout` with other types of layers as well. This implementation might look slow. I guess it could be improved by adding support for `back!` for the `maximum` function.

An alternate implementation of `maxout` may be found at https://github.com/FluxML/NNlib.jl/pull/33.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -1,3 +1,4 @@
+using Flux:vcat, maximum
 """
     Chain(layers...)
 
@@ -43,6 +44,7 @@ end
 
 Creates a traditional `Dense` layer with parameters `W` and `b`.
 
+
     y = σ.(W * x .+ b)
 
 The input `x` must be a vector of length `in`, or a batch of vectors represented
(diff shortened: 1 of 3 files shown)


CarloLucibello:
I don't think we should add this use case specific and slow machinery to a fairly simple and common layer such as Dense. Better to keep it as a separete operator (e.g. as in the NNlib approach or as a new MaxoutDense type)

see also https://github.com/pytorch/pytorch/issues/805

MikeInnes:
Closing in favour of https://github.com/FluxML/NNlib.jl/pull/33

PR #193, opened 2018-03-03
Add Local Response Normalisation

americast:
Implements Local Response Normalisation. Fixes #192.

PR #210, opened 2018-03-18
Added Local Contrast Normalization.

sanskar107:


CarloLucibello:
maybe a more explicative name rather than LCN?

MikeInnes:
Looks good but needs some tests. Also, the current implementation does not seem very GPU-friendly; is there any way to vectorise it more?

ToucheSir:
#312 and #193 were added around the same time and have both been closed for inactivity. Shall we close this as well?

CarloLucibello:
yes, this is definitely the wrong approach in any case.

PR #220, opened 2018-04-01
add batches

CarloLucibello:
I think a basic data partitioning tool should be part of Flux, being batch partitioning such a common task.

Also here I add the `mat(x)` method, reshaping an array into a matrix, which is something you typically want to do before feeding the output of tensor input (e.g. the output of a convolutional layer) to a fully connected layer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@
 docs/build/
 docs/site/
 docs/flux.css
+.vscode/
 deps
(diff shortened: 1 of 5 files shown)


CarloLucibello:
@MikeInnes looks like RNN tests on gpu have been giving troubles in all recent PRs...

oxinabox:
I feel like when it comes to this kinda data mangling, MLDataUtil.jl does a great job,
and we could instead just be documenting how to use Flux and MLDataUtils to accomplish this.

PR #222, opened 2018-04-02
[wip] improve losses

CarloLucibello:
This PR tries to address some inconsistencies (well, in my opinion at least :) in the present state of Flux's losses:
- 3 functions (`crossentropy`, `logitcrossentropy` and `mse`) perform reduction, 2 don't (`binarycrossentropy`, `logitbinarycrossentropy`)
- some names are too long
- the interface should encourage to use `logsoftmax` instead of `softmax` as the input of cross_entropy for numerical stability
- the `logit` prefix makes sense for binary cross entropy, less so for the standard one, I think.
- `crossentropy` can be used only in conjunction with a onehot target representation `y`. In other frameworks, such as pytorch and knet, tipically `y` is a vector of integers in 1:K or 0:K-1. 

The change I propose here is to mimic closely pytorch's interface http://pytorch.org/docs/master/nn.html#id46

TODO
[ ] improve docs
[ ] should the average in mse and bce be along all dimensions or just the last one (present state of the pr) ?


bhvieira:
Is this one still relevant? Does it need any help here @CarloLucibello ?

CarloLucibello:
completely forgot about this. It is quite outdated, I'll close it. Some piece of code could be scavenged by future PRs

PR #223, opened 2018-04-02
merge onehotbatch into onehot and add doctring

CarloLucibello:
I see no point in having two separate methods


MikeInnes:
What if you are single encoding a vector? What if you are batch encoding an iterator / non-vector collect?

PR #224, opened 2018-04-02
NADAM optimizer

tejank10:
Added NADAM optimizer. [Paper ref](https://openreview.net/pdf?id=OM0jvwB8jIp57ZJjtNEZ). NADAM combines Adam and NAG.


CarloLucibello:
I think there are a couple of subtle mistakes in the update rule. It should read like this
```julia
function nadam(p::Param; η::Real = 0.001, β1::Real = 0.9, β2::Real = 0.999, ϵ::Real = 1e-8)
  mt = zeros(p.x)
  vt = zeros(p.x)
  β1p, β2p = β1, β2
  function ()
    @. mt = β1 * mt + (1 - β1) * p.Δ
    @. vt = β2 * vt + (1 - β2) * p.Δ^2
    @. p.Δ =  (β1*mt/(1-β1*β1p) + (1-β1)*p.Δ/(1-β1p))  / √(vt*β2/(1-β2p) + ϵ) * η
    β1p *= β1
    β2p *= β2
  end
end
```
Do you agree?

CarloLucibello:
Also, since you are at it, you could move adam and rmsprop's  epsilons inside the square roots, and add the missing epsilon to adagrad

tejank10:
Well, I had followed [blog post](http://ruder.io/optimizing-gradient-descent/index.html#nadam) while working on it. For simplicity, subtle variations were introduced in the rule. Going by the original paper, yes, that should be the correct update rule. I think we should stick to the rule in the paper. I'll make the necessary changes.

MikeInnes:
I'm happy with this, can you just fix the merge conflict?

MikeInnes:
Thanks a lot!

PR #241, opened 2018-04-18
layers: fix docstring of LSTM and GRU

iblis17:
[ci skip]

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -149,7 +149,7 @@ Base.show(io::IO, l::LSTMCell) =
   print(io, "LSTMCell(", size(l.Wi, 2), ", ", size(l.Wi, 1)÷4, ")")
 
 """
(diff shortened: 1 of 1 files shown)


MikeInnes:
:+1:

I would just describe the `init ` kwarg here (e.g. layers are by default initialised with glorot, this can be changed via `init`), rather than adding it to the signature. (Which is not that informative if you don't already know about it, and easily guessable if you do).

iblis17:
well, your comment reminds me this: 
  https://github.com/JuliaLang/julia/pull/25315#issuecomment-354526282
There are lots of layers/functions share these same kwargs and should share same descriptions as well.
Maybe I should find time to create some utils for composing docstring.
I have a tight schedule tmr. (Let me procrastinate this :p

ToucheSir:
This could probably be rebased and edited to add `initb` and `init_state`, but I assume that's better served by a new PR.

PR #242, opened 2018-04-19
Mention CUDAnative.jl's install instructions

KadeG:


MikeInnes:
Maybe it's better to link CuArrays.jl as CUDAnative isn't enough on its own. Or maybe just link the [GPU docs](http://fluxml.ai/Flux.jl/stable/gpu.html) and add any extra install info we need here?

KadeG:
[CuArrays'jl](https://juliagpu.github.io/CUDAnative.jl/stable/#Installation) only says 

> Note that you need to build Julia 0.6 from source and have CUDA available to use this package.

which is less detailed/helpful than [CUDAnative.jl's instructions.](https://juliagpu.github.io/CUDAnative.jl/stable/#Installation)

Moving GPU-specific stuff to the GPU docs makes sense.

MikeInnes:
I updated the CuArrays readme to be link to CUDAnative. We can add a mention to the GPU docs as well, so it's clear everywhere.

KadeG:
CUDAnative instructions are now mentioned in GPU docs and readme (same as your change to CuArrays readme).

PR #243, opened 2018-04-20
Support for hcat and cat

gustafsson:
* Support for `hcat` and `cat`.
* Support for `vcat` to accept arrays with more than 2 dimensions.


MikeInnes:
Awesome PR, thanks!

PR #245, opened 2018-04-26
Add AdaMax optimizer

freeboson:
This is the popular variant of Adam which is simpler yet still stable. Implementation is identical to paper, except for the epsilon when dividing by `ut`.


MikeInnes:
Awesome, thanks!

PR #247, opened 2018-04-27
Add stopping conditions

### LINKED ISSUE #227: Stopping criteria
It came up in #131 and #132,
I think we should have a number of predefined stopping callbacks,
So that you feed the training data in as an an infintity stream of epochs.
And you have stop based on criteria.
The ones I am thinking are:

- [ ]  `atol` stop when the loss is below some point (obv. 0 as default)
- [ ]  `rtol` stop when the decrease in loss from one epoch to the next is below some point (obv. 0 as default)
- [ ] Max time spent
- [ ] Max Epochs

 `atol` and `rtol` could be parameterised to use different loss functions,
including different datasets. Early stopping can be achieved by using a development dataset instead of the training dataset.

We could have more; e.g. incorporating some from of hysteresis to for looking at improvement over n-epochs, like some kind of `rⁿtol` but  that can be later.

Of course the user should also still be able to define their own custom methods.

oxinabox:
Fix #227 

these are all type unstable, because they either return `false` or `:stop`
I guess I could make them all return `:donotstop` or `:stop`.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -11,6 +11,7 @@ export Chain, Dense, RNN, LSTM, GRU, Conv, Conv2D,
   Dropout, LayerNorm, BatchNorm,
   SGD, ADAM, Momentum, Nesterov, AMSGrad,
   param, params, mapleaves, cpu, gpu
+  
 
 @reexport using NNlib
 using NNlib: @fix
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -45,6 +45,84 @@ function train!(loss, data, opt; cb = () -> ())
   end
 end
 
+
+############
+# Stopping criteria
+
+"""
+    stopif(cbs...)
+
+Given one callback, returns a callback that
+returns`:stop` if that callback returns `true` or `:stop`.
+i.e. converts a boolean callback, into a stopping condition.
+If multiple callbacks are passed, this does the logical union of them
+"""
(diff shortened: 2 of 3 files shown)


oxinabox:
I kinda feel like these are a common abstractions, that could e.g. be shared with something like Optim.jl 
https://github.com/JuliaNLSolvers/Optim.jl/issues/571

oxinabox:
No I haven't I think it is a good idea,
but I thought I should get the idea of having this callbacks predefined as a seperate thing to that 
to get your/others thoughts on them,
before changing the interface itself.
Since the interface changes would be naturally based on what makes for these kinds of things being easily defined/used.

oxinabox:
When I wrote this I though we only would be wanting  ors/ands and not ands/alls. 
But a it seems sensible to do early stopping only after the first say 500 epochs. Which be done with a Max_rounds and a rtol

MikeInnes:
I imagine some of this is still pretty useful, but will need a rebase for the stopping API in #361.

oxinabox:
Yes

ToucheSir:
I want to say this has been obsoleted by #1545?

darsnack:
Yeah I think that's correct. Re-open if you think something is missing @oxinabox.

oxinabox:
Yeah, I would much rather custom training loops now

PR #248, opened 2018-04-27
CIFAR-10 Dataset

avik-pal:
Added the CIFAR-10 Dataset under the Flux.Data Module. This allows users to directly use the CIFAR-10 Dataset similar to the MNIST Dataset.

### DIFF
--- a/src/data/Data.jl
+++ b/src/data/Data.jl
@@ -18,4 +18,6 @@ include("tree.jl")
 include("sentiment.jl")
 using .Sentiment
 
+include("cifar10.jl")
+
 end
--- a/src/data/cifar10.jl
+++ b/src/data/cifar10.jl
@@ -135,18 +135,10 @@ Returns a Dictionary where every label number has been mapped to the category ti
 
 function labelnames()
   labelmap = Dict{Int,String}()
-  label = 0
-  str = ""
   cd(dir_files) do
-    file = Char.(read(LABELS))
-    for i in file
-      if(i=='\n')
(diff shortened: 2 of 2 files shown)


ViralBShah:
I guess my question was if even MNIST should be sent into a separate package. We can't keep adding new data sources and models in the Flux repo - since that would limit its eventual scalability. Of course, ok to do this until we figure out the right solution.

ninjin:
Seconding what @ViralBShah said, the way forward should rather be to put the effort towards dataset packages; maybe a bit one for vision? I think there already is movement on one for NLP datasets.

Evizero:
I can't find the discussion, but as far as I recall when this last came up, the following argument was made (or something along those lines): Simpler datasets such as CIFAR and MNIST don't require much code and shipping basic implementations of them with Flux avoids unnecessary dependencies.

The argument remains as valid as when it was made. A generic Datasets package does come with a certain dependency baggage. For example, introducing the SVHN datasets to https://github.com/JuliaML/MLDatasets.jl forced me to add a dependency to https://github.com/JuliaIO/MAT.jl

MikeInnes:
Thanks @avik-pal for putting the work into this, but we actually already have a CIFAR10 loader along with other image tools in [Metalhead.jl](https://github.com/FluxML/Metalhead.jl). I think that become a standard place for those datasets and models (more datasets welcome)!

I should probably add a note on that to the readme here, to prevent someone submitting it as a PR every couple of weeks :)

PR #249, opened 2018-04-29
[RFC] Backpropagation for `maximum` and `minimum`

ninjin:
A first minimal implementation just to get the support into Flux. Comments more than welcome in case I got the API wrong.

Bonus points to anyone that reminds me how to run the tests when you have the repo checked out somewhere outside of `~/.julia` as I seem to have forgotten and reverted to having a single file where I do the hacking then then paste it all back in before a commit.


ninjin:
The JuliaGPU CI errors look unrelated, unless I have inadvertently broken the RNN implementation. '^^

MikeInnes:
LGTM. We are having some CUDNN RNN issues generally but I'm hoping the CUDNN 7 logging will help me track that down.

Why not keep the package in `~/.julia`?

ninjin:
> Why not keep the package in ~/.julia?

@MikeInnes: Because it messes with an install that has implications for all the experiments I am running? Pkg3 solves this wonderfully, I just have to be patient.

MikeInnes:
Fair enough. I haven't had that issue, but then I never use this thing :P

PR #257, opened 2018-05-07
Check for `Inf` and `NaN` within `back!(::TrackedReal)`

staticfloat:
This is often checked for within user code, whenever not using `Flux.train!()`, but there's no reason to do that, let's do it for them within `back!(::TrackedReal)`


staticfloat:
I should note that the reason this is a pain point at all is because I'm training a DAN that has multiple losses that get accumulated, and it's tiring to have multiple `if/else` cases for each independent loss being applied within the training loop.

MikeInnes:
Makes sense, thanks!

Issue #260, opened 2018-05-10
BatchNorm changes the data type.


ayush1999:
The BatchNorm operation is changing the data type. For example:
```

>>> a = convert(Array{Float32,4}, rand(5,5,1,1))    # Tensor of Float32.
>>> a = Tracker.TrackedArray(a)                     # TrackedArray of Float32
>>> b = BatchNorm(1, ϵ=1e-5, momentum = 0.9)
>>> b(a)
Tracked 5×5×1×1 Array{Float64,4}:
[:, :, 1, 1] =
 -0.353323  -1.22248     0.219534  -0.146294   -1.4266   
  1.39974    1.07692     0.925905   0.0425806   0.218777 
  0.920126  -1.55046     1.44969   -1.16887     1.06475  
 -0.143462   1.37654    -0.47538   -1.81898    -0.245258 
 -1.58401   -0.0710794   1.10551    0.388538    0.0175963

```

As we can see, the data changed from tensor of Float32 to tensor of Float64. As a result, performing further computation isn't possible (because the weights and inputs need to have same type)

ayush1999:
@MikeInnes This is interfering with onnx work, could you please have a look?

CarloLucibello:
Unfortunately, this is a case of #231:  `BN.γ` and  `BN.β` defaults initialization is to Float64 arrays. I'm not sure that #261, i.e. converting the final output to the input data,  would be a proper solution for this. One could argue that we would need such a convert for the output of any layer (Dense, Conv, ....)

A verbose workaround for this issue is to use pass the arguments `initβ=x->zeros(Float32, x), initγ=x->ones(Float32, x)` to the BatchNorm constructor. 

More conveniently and to generally fix #231, we should add a conversion method for the whole model, something like, `model = model |> float32`  and/or add to each constructor an eltype argument.

KristofferC:
You just need to initialize `ϵ` and `momentum` consistently.

```
b = BatchNorm(1, ϵ=Float32(1e-5), momentum = Float32(0.9))
```

PR #261, opened 2018-05-11
Fix for broadcast type matching

ayush1999:
Fixes : #260

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -141,7 +141,12 @@ function (BN::BatchNorm)(x)
   end
 
   let λ = BN.λ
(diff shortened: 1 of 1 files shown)


CarloLucibello:
What's the cause of #260 ? `BN.λ` being Float64?

ayush1999:
Looks to be the case. @CarloLucibello  Any idea why the build is failing?

MikeInnes:
For ONNX, it seems like it'd be better just to initialise the BatchNorm with f32 when you create it. This is going to add a fair bit of conversion overhead that I'd rather avoid.

CarloLucibello:
closing as stale

PR #266, opened 2018-05-20
Use broadcast for dropout

### LINKED ISSUE #265: Dropout for GPU
Dropout currently lacks a GPU implementation, so it falls back to looping over `getindex` and is 1000x slower than it should be.
The easiest thing to do would be to switch the manual loop for a broadcast call; we should also evaluate whether the current `rand!` (which calls a custom RNG from GPUArrays) should be swapped for a cuRAND invocation and whether a custom GPU call would be faster than the broadcasts.

CC @avik-pal

jekbradbury:
Should be fast enough on GPU now that it's not going to be an optimization target again for a while. Hopefully isn't meaningfully slower on CPU?

Fixes #265

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -31,15 +31,14 @@ function Dropout(p)
   Dropout{typeof(p)}(p, true)
 end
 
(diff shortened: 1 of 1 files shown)


MikeInnes:
Neat, yeah this'll be fine. Thanks!

PR #268, opened 2018-05-21
Add `dilation` kwarg to `Conv`

staticfloat:
Now that we have dilated convolution support in `NNlib`, this enables support in Flux's `Conv` layer.


staticfloat:
Thanks @tejank10!

MikeInnes:
:+1: Thanks!

PR #269, opened 2018-05-22
Julia 0.7

MikeInnes:
You can try this out with `add Flux#julia-0.7`.

- [x] Updates to handle 0.7 broadcast machinery
- [x] Fixes to tracked array show
- [x] ~~NNlib [branch](https://github.com/FluxML/NNlib.jl/pull/47)~~
- [x] ~~AbstractTrees#master (https://github.com/Keno/AbstractTrees.jl/issues/17)~~
- [ ] Depwarns

CuArrays also needs updates, but Flux itself will work fine without that for now.

### DIFF
--- a/src/treelike.jl
+++ b/src/treelike.jl
@@ -7,11 +7,22 @@ mapchildren(f, x) = x
 children(x::Tuple) = x
 mapchildren(f, x::Tuple) = map(f, x)
 
-function treelike(T, fs = fieldnames(T))
-  @eval current_module() begin
+function treelike(m::Module, T, fs = fieldnames(T))
+  @eval m begin
     Flux.children(x::$T) = ($([:(x.$f) for f in fs]...),)
     Flux.mapchildren(f, x::$T) = $T(f.($children(x))...)
   end
+end
+
+function treelike(T, fs = fieldnames(T))
+  Base.depwarn("`treelike(T)` is deprecated, use `@treelike T`", :treelike)
+  treelike(Base._current_module(), T, fs)
+end
+
+macro treelike(T, fs = nothing)
+  fs == nothing || isexpr(fs, :tuple) || error("@treelike T (a, b)")
+  fs = fs == nothing ? [] : [:($(map(QuoteNode, fs.args)...),)]
+  :(treelike(@__MODULE__, $(esc(T)), $(fs...)))
 end
 
 isleaf(x) = isempty(children(x))
--- a/src/treelike.jl
+++ b/src/treelike.jl
@@ -16,7 +16,7 @@ end
 
 isleaf(x) = isempty(children(x))
 
-function mapleaves(f, x; cache = ObjectIdDict())
+function mapleaves(f, x; cache = IdDict())
   haskey(cache, x) && return cache[x]
   cache[x] = isleaf(x) ? f(x) : mapchildren(x -> mapleaves(f, x, cache = cache), x)
 end
--- a/src/treelike.jl
+++ b/src/treelike.jl
@@ -53,7 +53,7 @@ cpu(m) = mapleaves(x -> adapt(Array, x), m)
 
 gpu_adaptor = identity
 
-@require CuArrays begin
+@init @require CuArrays="3a865a2d-5b23-5a0f-bc46-62713ec82fae" begin
   global gpu_adaptor = CuArrays.cu
 end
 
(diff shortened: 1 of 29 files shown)


ViralBShah:
Can start with the update for 0.7. Tagging will be any day now.

dfdx:
For broadcasting see MikeInnes/TakingBroadcastSeriously.jl#8 - in Julia 0.7 overloading it became much easier.

MikeInnes:
It's a bit more complex than that for us, because we don't actually want to un-fuse broadcasts, and in fact 0.6 was simpler in this case. We have to inject dual numbers into the leaves of the `Broadcasted` tree (what previously would just have been a tuple of arguments), collect the partials and push them back through the original TrackedArray leaves.

jekbradbury:
I think you can use `Broadcast.flatten` to get something closer to the 0.6 interface with pre-fused broadcasts?

MikeInnes:
Yeah, doing that works for broadcast as long as we're happy to force fusion; though it's hard to imagine where that would be an issue.

@gustafsson would you be able to look at how we overload `cat` on 0.7? We're not catching things right now, looks like the methods we need to disambiguate has changed somewhat.

src/treelike.jl
```diff
@@ -7,13 +7,24 @@ mapchildren(f, x) = x
...
+  Base.depwarn("`treelike(T)` is deprecated, use `@treelike T`", :treelike)
+  treelike(Base._current_module(), T, fs)
+end
+
+macro treelike(T, fs = nothing)
+  fs == nothing || isexpr(fs, :tuple) || error("@treelike T (a, b)")
+  fs = fs == nothing ? [] : [:($(map(QuoteNode, fs.args)...),)]
+  :(treelike(@__MODULE__, $(esc(T)), $(fs...)))
+end
+
```
jekbradbury: Your macro-fu is second to none, but isn't the `@__MODULE__` stuff only needed because of your original strategy of making a function that uses `@eval` to behave kind of like a macro? Could you accomplish the same thing with just a macro and leave the `treelike` function only for the deprecation?

MikeInnes: It's a good thought, and it'd work when the fields are explicitly provided, but otherwise we want to do `fieldnames(T)` where `T` is a type, rather than a symbol. So we have to go through at least one round of `eval` here.

zenna:
The MNIST example in ModelZoo is broken on .7 branch.
Flux.back! gives an out of memory error (which it didn't with 0.6), if I reduce the data (as in the following snippet) I get the error:

`ERROR: DimensionMismatch("A has dimensions (100,32) but B has dimensions (10,100)")`

Note that only occurs in the `back!`, `m(X)` works as expected

```julia
using Flux, Flux.Data.MNIST
using Flux: onehotbatch, argmax, crossentropy, throttle
using Base.Iterators: repeated
# using CuArrays

# Classify MNIST digits with a simple multi-layer-perceptron

imgs = MNIST.images()[1:100]
# Stack images into one large batch
X = hcat(float.(reshape.(imgs, :))...) |> gpu

labels = MNIST.labels()[1:100]
# One-hot-encode the labels
Y = onehotbatch(labels, 0:9) |> gpu

m = Chain(
  Dense(28^2, 32, relu),
  Dense(32, 10),
  softmax) |> gpu

loss(x, y) = crossentropy(m(x), y)
Flux.back!(loss(X, Y))
```

SimonDanisch:
Woot woot:
```Julia
Test Summary: | Pass  Total
Flux          |  313    313
   Testing Flux tests passed 
````
I run into some annoying issue with Gzip, though. 
The combination of https://github.com/JuliaLang/julia/pull/22828 And the shaky "build" in:

https://github.com/JuliaIO/GZip.jl/blob/b8b8d897bba62b3b0bcc373064b0f4f84a32bd18/src/zlib_h.jl#L5

Makes gzip (and so Flux) pretty flaky.

I'm a bit confused by all the names (gzip, zlib, libz), but could https://github.com/bicycle1885/CodecZlib.jl be a replacement for Gzip? It uses that sweet BinaryProvider sauce ;)

**Comment by MikeInnes** [comment]
I think I tried that at some point and didn't use it for some reason. Can't remember now though. If it's an issue I'm happy to switch over.

**Comment by JBlaschke** [comment]
Just a heads up, on macOS 10.13.6 and Julia v0.7 I get:

```julia
julia> import Flux
[ Info: Precompiling Flux [587475ba-b771-5e3f-ad9e-33799f191a9c]
ERROR: LoadError: LoadError: error compiling top-level scope: could not load library "libz"
dlopen(libz.dylib, 1): image not found
```

PR #270, opened 2018-05-23
Add `TrackedArray` support for `repeat(x; inner, outer)`

staticfloat:
This makes it really easy to do things like make `Upsample` layers, which are pretty useful for autoencoders.

### DIFF
--- a/src/tracker/array.jl
+++ b/src/tracker/array.jl
@@ -93,6 +93,26 @@ function back(::typeof(repmat), Δ, xs::TrackedVecOrMat, m, n=1)
     back(xs, Δ′)
 end
 
+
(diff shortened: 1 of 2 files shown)


MikeInnes:
Yeah I just meant using existing GPU-friendly primitives; you can sometimes hack broadcasting to do stuff like this (e.g. kronecker products) but it sounds like that's not really the case here. I'm sure we can figure out the kernel later anyway.

staticfloat:
I suppose you could write a massively parallel version where you construct a source index from every destination index, then run `A[dest_idxs] .= B[src_idxs]`.

MikeInnes:
@americast might be worth looking at what this needs to work on the GPU.

@staticfloat thanks for the patch!

americast:
Sure, will have a look :+1:

Issue #262, opened 2018-05-15
Error running code from the basic tutorial


kleskjr:
Hi, while going through the basic tutorial, I got hit by the following issue:

```julia> using Flux
julia> m = Chain(
         Dense(28^2, 128, relu),
         Dense(128, 32, relu),
         Dense(32, 10), softmax)

Chain(Dense(784, 128, NNlib.relu), Dense(128, 32, NNlib.relu), Dense(32, 10), NNlib.softmax)

julia> loss(x, y) = crossentropy(m(x), y) + sum(vecnorm, params(m))
loss (generic function with 1 method)

julia> loss(rand(28^2), rand(10))
ERROR: UndefVarError: crossentropy not defined
Stacktrace:
 [1] loss(::Array{Float64,1}, ::Array{Float64,1}) at ./REPL[7]:1
```

Exporting _crossentropy_ in the src/Flux.jl seems to resolve the issue. But is it an adequate solution to the problem?

kleskjr:
thanks, maetshju! I just created a pull request on the docs where _Flux._ is added in front of _crossentropy_. This should decrease the disambiguity for newbies like me.

MikeInnes:
I'll close this in favour of #276

PR #276, opened 2018-05-25
Solves Issue #262

kleskjr:
Makes  the running of the basic examples smoother (Issue #262).
crossentropy is not exported by Flus so, either explicit reference or explicit export should be add to run the examples.


MikeInnes:
Can we just add `using Flux: crossentropy` before the first example?

MikeInnes:
:+1: Thanks!

PR #279, opened 2018-05-30
Adds support for Depthwise Convolutions

avik-pal:
FluxML/NNlib.jl#42 needs to be merged before this is functional


MikeInnes:
This looks good to me. Can you add this layer to the docs as well?

Before merging, I'd like to get some API review from folks who have used depthwise convs before.

src/layers/conv.jl
```diff
@@ -45,3 +45,50 @@ function Base.show(io::IO, l::Conv)
...
+       stride = stride, pad = pad)
+
+Flux.treelike(DepthwiseConv)
+
+function (c::DepthwiseConv)(x)
+  σ, b = c.σ, reshape(c.bias, map(_->1, c.stride)..., :, 1)
+  σ.(depthwiseconv(x, c.weight, stride = c.stride, pad = c.pad) .+ b)
+end
+
+function Base.show(io::IO, l::Conv)
```
jekbradbury: Should this be `DepthwiseConv`?

avik-pal: Yes. Will fix this.

avik-pal:
How does it sound adding a constructor for `Depthwise Separable Convolutions` as Depthwise Convolutions are rarely used in itself, its mostly in combination with a pointwise convolution?

test/tracker.jl
```diff
@@ -172,6 +172,9 @@ end
 @test gradtest(conv, rand(10, 10, 3, 2), randn(2, 2, 3, 2))
 @test gradtest(conv, rand(10, 10, 10, 3, 2), randn(2, 2, 2, 3, 2))
 
+# NOTE: To pass this test rtol should be as high as 2.0 so commenting this out
+# @test gradtest(depthwiseconv, rand(10,10,3,2), randn(2, 2, 2, 3))
```
MikeInnes: This seems a bit suspect; is there a reason to expect this to be so high? How do we know the gradients are correct?

avik-pal: Since the gradients should be same as a special case to Convolution I ran them simultaneously and verified that manually (atleast while implementing in NNlib). Also the NNlib tests should verify that. Any other suggestions as to how should I check that? 1 issue might be I got something wrong here while setting up the tracker. So I will verify that once

avik-pal: It seems the depthwise conv tests fail on changing the mode parameter (which should not be the case). There might be some form of a bug in the implementation. Will have to figure that out.

MikeInnes: Ok, let me know how it goes

avik-pal:
This PR FluxML/NNlib.jl#57 should fix the gradient.

avik-pal:
The tests are passing locally, I will update the gradient hook up with the tracker in some time.

avik-pal:
I am not sure why the test fails here. The `gradtest` for depthwise conv is passing locally after the NNlib fix.
cc @MikeInnes

avik-pal:
Works on 0.7 as long as NNlib is on `master`

terasakisatoshi:
I know this PR is WIP. 
But I want to use the DepthwiseConv layer to define MobileNet in Flux.

I cloned your fork repository, checkout the branch `depthwiseconv` and move them into 
`~/.julia/packages/Flux/UHjNa/`

Also I did `pkg> add NNlib#master`.

precompiling seems ok but I encountered the following error that says UndefVarError of `N`.


```julia
julia> using Flux
julia> dc=Flux.DepthwiseConv((3,3), 3)
ERROR: UndefVarError: N not defined
Stacktrace:
 [1] #DepthwiseConv#73(::Tuple{Int64,Int64}, ::Tuple{Int64,Int64}, ::Type, ::TrackedArray{…,Array{Float64,4}}, ::TrackedArray{…,Array{Float64,1}}, ::Function) at /home/terasaki/.julia/packages/Flux/UHjNa/src/layers/conv.jl:77
 [2] (::getfield(Core, Symbol("#kw#Type")))(::NamedTuple{(:stride, :pad),Tuple{Tuple{Int64,Int64},Tuple{Int64,Int64}}}, ::Type{Flux.DepthwiseConv}, ::TrackedArray{…,Array{Float64,4}}, ::TrackedArray{…,Array{Float64,1}}, ::Function) at ./none:0
 [3] #DepthwiseConv#75(::Function, ::Tuple{Int64,Int64}, ::Tuple{Int64,Int64}, ::Type, ::Tuple{Int64,Int64}, ::Pair{Int64,Int64}, ::Function) at /home/terasaki/.julia/packages/Flux/UHjNa/src/layers/conv.jl:87
 [4] Type at /home/terasaki/.julia/packages/Flux/UHjNa/src/layers/conv.jl:87 [inlined] (repeats 2 times)
 [5] top-level scope at none:0
```

avik-pal:
@terasakisatoshi thanks for pointing it out. Will fix it right away.

terasakisatoshi:
Hum...Maybe we should modify from

```julia
# defined in src/layers/conv.jl line 75
DepthwiseConv(w::AbstractArray{T}, b::AbstractVector{T}, σ = identity; 
       stride = 1, pad = 0) where T = 
```
to
```julia
DepthwiseConv(w::AbstractArray{T,N}, b::AbstractVector{T}, σ = identity; 
       stride = 1, pad = 0) where {T,N} = 
```

The modification works the code below:

```
x = rand(28,28,3,1)
dc = Flux.DepthwiseConv((3,3),3=>2)
size(dc(x)) #size(dc(x)) = (26, 26, 6, 1) 
```

terasakisatoshi:
@avik-pal Ops you've just got fix it. Thank you very much.

avik-pal:
Yes the tests were fixed after the NNlib gradient PR was merged.

MikeInnes:
Great. Thanks a ton Avik!

PR #290, opened 2018-06-08
Default value of dilation

tejank10:
Changed `dilation` default from 0 to 1


tejank10:
Yes, `dilation` by default is supposed to be 1. It is in all other places, except this one.

Issue #282, opened 2018-06-04
Support for additional dimensions in Dense layer


harryfei:
Currently, the `Dense` layer convert a (in_features, N) matrix into a (out_features, N) matrix. Shall we extend this api to support additional dimensions? 

By supporting additional dimensions, the interface of `Dense` layer would be like this.
* Input Shape: (in_features, d1, d2, d3, ..., N)
* Ouput Shape: (out_features, d2, d2, d3,..., N)

References:
* [pytorch.Linear](https://pytorch.org/docs/stable/nn.html#linear)
* [keras.Dense](https://keras.io/layers/core/#dense)

ninjin:
Yes, I think that sounds like a very elegant API that makes expressing something very common easy to read and less verbose. It should also be fairly easy to implement, would you be happy to submit a PR?

fangyuanziti:
Of course, I will submit a PR for this. :smile:

MikeInnes:
I take it the inner dimensions are effectively treated as extra batch dimensions?

It seems to me like you could go either way here, e.g. they could equally belong to the input features. But if this the more common / standard case and people are generally on board, I'm ok with it.

fangyuanziti:
Actually, I have no idea about whether this the more common/standard case or not. Howerver, I found this inteface is very useful when I try modeling neural networks to process audio.

Besides, I don't think we can use input_features get the same thing as  additional dimensions.

If we use input_feautres only:

* Input Shape: (in_features * d1 * d2 * d3 * ..., N)
* Ouput Shape: (out_features * d1 * d2 *d3,..., N)
Finally, we get a dense layer with parameter size of (out_features * d1 * d2 * d3 * ..., in_features * d1 * d2 * d3 * ...).

If we can use additional dimensions.
* Input Shape: (in_features, d1, d2, d3, ..., N)
* Ouput Shape: (out_features, d1, d2, d3,..., N)
Finally, we get a dense layer whit parameter size of (out_features, in_features). 

The dense layers in the two situations are different. So I suggest to add addtional dimensions in `Dense` layer.

MikeInnes:
Actually, they are not necessarily exclusive, as long as one is explicit. For example you could support N-dimensional features with `Dense((5, 3), (2, 7))`. Then this accepts a `(5, 3, ...)` array and has a `15x14` weight matrix.

fangyuanziti:
[...]

PR #293, opened 2018-06-12
Add additional dimensions support in Dense layer.

harryfei:
By supporting additional dimensions, the interface of Dense layer would be like this.
 * Input Shape: (in_features, d1, d2, d3, ..., N)
 * Ouput Shape: (out_features, d2, d2, d3,..., N)

Related discuss: #282

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -75,7 +75,27 @@ treelike(Dense)
 
 function (a::Dense)(x)
   W, b, σ = a.W, a.b, a.σ
-  @fix σ.(W*x .+ b)
+
+  in_size = size(x)
+
+  if length(in_size) <= 2
+    return σ.(W*x .+ b)
+  else
+    extra_dims = in_size[2:end]
+    out_feature_size = size(b)[1]
+
+    out_size = (out_feature_size, extra_dims...)
+
+    A = zeros(out_size)
+
+    dim_range = Iterators.map(x -> 1:x, extra_dims)
+
+    for d in Iterators.product(dim_range...)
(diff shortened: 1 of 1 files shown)


MikeInnes:
Thanks! A couple comment before we can merge this –

I think it'd be cleaner to keep the current `Dense` definition as-is, perhaps with a `AbstractVecOrMat` type restriction. Then you can overload `AbstractArray` and do the needed reshaping.

You can do this in a single matmul by just reshaping the inputs/outputs to `Dense` appropriately, which would avoid extra copies and be more GPU-friendly.

fangyuanziti:
I will change this in next days. :smile:

harryfei:
I think we don't need this change anymore according to the above cleaner solution.
Should we add the reshape helper function in this repository? If it's Ok, I will create another PR. @MikeInnes

MikeInnes:
Sure, if you can update this PR with the reshape version, that works great for me.

drozzy:
Any eta on when this will be merged? I'm finding myself missing this basic functionality.
I wish all layers treated extra dimensions as batch dimensions.

drozzy (approved):
Tested and it works for me:
```
a = randn(10, 11, 12) 
b =  Dense(10, 20)
b(a) # 20x11x12
```

CarloLucibello:
this is not an efficient way to do it, we should go through a reshape

drozzy:
Yes, also I tried copying this code to my implementation and I seem to be having some issues with gpu. Don't quote me on this.

I would be happy for reshape to be merged, as long as we get something to help out with the case I provided above.
Thanks.

PR #294, opened 2018-06-12
Wrapper for CuDNN BatchNorm

### LINKED ISSUE #239: CUDNN BatchNorm
Our current BatchNorm implementation should be consistent with CUDNN's spatial batchnorm. Following the RNN implementation, we need to hook this up to use CUDNN where possible.

avik-pal:
Fixes #239 and #314


ViralBShah:
What is the perf improvement as a result?

avik-pal:
This wrapper actually turns out to be slower than the original implementation of BatchNorm

avik-pal:
Timing scripts:-
CuDNN wrapper -- https://gist.github.com/avik-pal/11a44f935bc01fed28b8b9017514668d
Julia BatchNorm -- https://gist.github.com/avik-pal/b602297a29f597dfbc77c833b2e7470e

avik-pal:
Version2 Scripts
1st script:-
604.576 μs (679 allocations: 26.17 KiB)
8.253 ms (699 allocations: 26.48 KiB)
808.001 ms (699 allocations: 26.48 KiB)
2nd script:-
484.460 μs (656 allocations: 24.05 KiB)
502.858 μs (676 allocations: 24.36 KiB)
276.378 ms (676 allocations: 24.36 KiB)

avik-pal:
Version 3 of scripts
1st script:-
```
644.334 μs (679 allocations: 26.17 KiB)
8.277 ms (699 allocations: 26.48 KiB)
807.791 ms (699 allocations: 26.48 KiB)
```
2nd script:-
```
511.727 μs (656 allocations: 24.05 KiB)
526.148 μs (676 allocations: 24.36 KiB)
272.391 ms (676 allocations: 24.36 KiB)
```

avik-pal:
The previous functions were internally calling the Flux BatchNorm code itself not the CuDNN wrapper. The function calls are now fixed however there is a `CUDNNError(code 3, CUDNN_STATUS_BAD_PARAM)` error which needs to be fixed before doing any furthur benchmarks.

avik-pal:
Updated benchmarks for the CUDNN Wrapper Forward Pass:-
```julia
345.658 μs (60 allocations: 2.44 KiB)
453.557 μs (60 allocations: 2.44 KiB)
1.008 ms (60 allocations: 2.44 KiB)
7.163 ms (60 allocations: 2.44 KiB)
```
The original BatchNorm GPU Speed:-
```julia
39.252 ms (678 allocations: 24.39 KiB)
40.223 ms (678 allocations: 24.39 KiB)
62.466 ms (678 allocations: 24.39 KiB)
100.317 ms (684 allocations: 24.48 KiB)
```
For the script used to benchmark : https://gist.github.com/avik-pal/2753678115589baa37b0f315077b7571

avik-pal:
Benchmarks for Combined forward and backward passes:
1. CuDNN Wrapper
```julia
(224, 224, 3, 1) --> 814.076 μs (324 allocations: 13.33 KiB)
(224, 224, 10, 1) --> 1.106 ms (324 allocations: 13.33 KiB)
(224, 224, 100, 1) --> 3.015 ms (324 allocations: 13.33 KiB)
(224, 224, 1000, 1) --> 24.255 ms (324 allocations: 13.33 KiB)
```
2. Original Code
```julia
(224, 224, 3, 1) --> 79.586 ms (992 allocations: 35.28 KiB)
(224, 224, 10, 1) --> 81.716 ms (992 allocations: 35.28 KiB)
(224, 224, 100, 1) --> 123.481 ms (992 allocations: 35.28 KiB)
(224, 224, 1000, 1) --> 201.506 ms (1000 allocations: 35.41 KiB)
```

avik-pal:
@MikeInnes could you review this. The test failure seems to be not related to these changes.

src/layers/normalise.jl
```diff
@@ -104,46 +104,46 @@ mutable struct BatchNorm{F,V,W,N}
...
-          initβ = zeros, initγ = ones, ϵ = 1e-8, momentum = .1) =
+          initβ = zeros, initγ = ones, ϵ = 1e-5, momentum = .1) =
   BatchNorm(λ, param(initβ(chs)), param(initγ(chs)),
-            zeros(chs), ones(chs), ϵ, momentum, true)
+            zeros(chs), ones(chs), ϵ, momentum, nothing, true)
 
-function (BN::BatchNorm)(x)
-  size(x, ndims(x)-1) == length(BN.β) ||
+
+function batchnorm(γ, β, x, μ, σ, momentum; cache = nothing, alpha = 1, beta = 0, eps = 1.0e-5, training = true)
```
MikeInnes: Is it possible to just overload `BatchNorm(x::CuArray)` here, rather than splitting this out?

We should not add `cache` to accomodate CUDNN if it's not doing anything on the CPU. If we need it, we can link it with a WeakKeyDict somewhere.

src/cuda/cudnn.jl
```diff
@@ -22,351 +23,179 @@ function DropoutDesc(ρ::Real; seed::Integer=0)
...
-  @back(c, unbroadcast(h, dc))
-  (dWi, dWh), db = backwardWeights(descs[m.rnn], data(x), h_, y, m.reserve)
-  @assert all(isleaf.((m.rnn.Wi, m.rnn.Wh, m.rnn.b)))
-  istracked(m.rnn.Wi) && accum_transpose!(m.rnn.Wi.grad, dWi)
-  istracked(m.rnn.Wh) && accum_transpose!(m.rnn.Wh.grad, dWh)
-  istracked(m.rnn.b) && accum_transpose!(m.rnn.b.grad, db)
+function back(::typeof(_batchnorm), Δ, g, b, x, running_mean, running_var, momentum, cache, alpha, beta, eps, training)
+  deriv_tup = ∇batchnorm(data(g), data(b), data(x), Δ, running_mean, running_var, momentum,
+                         cache = cache, alpha = alpha, beta = beta, eps = eps, training = training)
+  istracked(x) && @back(x, deriv_tup[1])
```
MikeInnes: `istracked &&` is redundant since that's what `@back` does as well

src/cuda/cudnn.jl
```diff
@@ -22,351 +23,179 @@ function DropoutDesc(ρ::Real; seed::Integer=0)
...
-# GRU: [weight, bias] × [input, hidden] × [reset, update, newmem]
-# LSTM: [weight, bias] × [input, hidden] × [input, forget, newmem, output]
-
-function params(w::CuVector, input, hidden, n = 1)
-  slice(offset, shape) = reshape(w[offset+(1:prod(shape))], shape)
-  wx = slice(0, (input, hidden*n))
-  wh = slice(length(wx), (hidden, hidden*n))
-  bias = w[length(wx)+length(wh) + (1:hidden*n)]
-  (wx, wh), bias
+mutable struct bncache
```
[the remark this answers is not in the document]
avik-pal: I don't get what would be the julia style? Also since I am setting the `mean` and `ivar` as `nothing` can I set their types?

src/layers/normalise.jl
avik-pal: It is possible to do that but is there a way to avoid handling the cases for TrackedArrays separately? This was the only reason I made the split

MikeInnes: Not certain if this is what you mean, but you can generally follow the implementation of RNNs here (e.g. RNNDesc and CuParam). Let me know if that doesn't translate to batchnorm for some reason.

avik-pal: I will look into the RNN implementation. What I meant was if we define a `BatchNorm(x::CuArray)` and x is something like `TrackedArray{..,CuArray{...}}` then the CPU code will be called.

MikeInnes: Right, that's what CuParam is for.

MikeInnes:
This is looking good but the tests are still failing for me, are they working on your machine?

avik-pal:
I am currently training a model using this branch on my fork. And the training is going on fine. (I know the cudnn gpu code is working because the training is definitely faster). But I think it would be still better to recheck.

MikeInnes:
The tests might just need some updates for your latest changes.

avik-pal:
Strangely one test continues to fail. The test for the `gamma grad`.

CarloLucibello:
given the discussion in #314, and since the cuda version here seems to work with the running average of the variance, could you make the cpu version consistent with that i.e. using variances instead of stddev as I proposed in #314 ?

avik-pal:
I will get it done.

avik-pal:
@MikeInnes I have pulled the updates of #322 into this. I have also rebased it. The only issue that remains is gradient for gamma is different in cpu and gpu. All tests pass except the gamma grad one (on gpu).

avik-pal:
@MikeInnes the tests are now passing. For some reason taking the `g` as an array of ones makes the test fail. Any other distribution causes the test to pass.

avik-pal:
Fixed conflicts and updated for Julia 0.7 but cant test until CuArrays work for 0.7

src/cuda/cudnn.jl
```diff
@@ -25,324 +24,189 @@ function DropoutDesc(ρ::Real; seed::Integer=0)
...
-  params::CuVector{T}
-  weights::NTuple{2,CuMatrix{T}}
-  bias::CuVector{T}
-  ptr::Ptr{Nothing}
-end
+BNCache() = BNCache(nothing, nothing)
 
-Base.unsafe_convert(::Type{Ptr{Nothing}}, d::RNNDesc) = d.ptr
+# CuDNN supports only 4D and 5D Tensors for BatchNorm Operations
+# so use the native julia code when doing batchnorm on a 2D Array
```
MikeInnes: Could we perhaps just do a reshape here instead?

src/layers/normalise.jl
```diff
@@ -70,23 +65,17 @@ end
...
               initβ = zeros, initγ = ones,
               ϵ = 1e-8, momentum = .1)
-
 Batch Normalization layer. The `channels` input should be the size of the
 channel dimension in your data (see below).
-
 Given an array with `N` dimensions, call the `N-1`th the channel dimension. (For
 a batch of feature vectors this is just the data dimension, for `WHCN` images
 it's the usual channel dimension.)
-
```
MikeInnes: Is it necessary to make this all one paragraph? I also think the code snippets are easier to read in source if they have a blank line after.

avik-pal: Oh sorry, this was not intentional.

MikeInnes:
Needs a rebase over #473

MikeInnes:
Thanks a lot @avik-pal!

MikeInnes:
One thing I just noticed is that this segfaults on 1-dimensional input. Would be good to fix that in a follow up.

avik-pal:
@MikeInnes Should I throw an error when the input is 1-D? I am not sure if I get what exactly is a 1-D input, is a `n-batch and 1 element per batch` or `n elements in 1 batch`

MikeInnes:
Throwing an error sounds fine, yeah. Or follow what the CPU version does. I don't think batchnorm can work with a batch of one so maybe the first interpretation would be reasonable.

ViralBShah:
Nice work @avik-pal !

PR #295, opened 2018-06-14
cpu() conversion before saving model/weights

tejank10:


MikeInnes:
Isn't this mentioned [here](https://github.com/FluxML/Flux.jl/pull/295/files#diff-1fcb61fd1d2b0210111d62d2a726a0f1R40)?

Technically you shouldn't *have* to do this, it's just a good practice. The fact that GPU weights wouldn't load at all for you is definitely a bug.

Issue #297, opened 2018-06-15
Implement einsum function/macro à la PyTorch and TF


outlace:
[ Opening a new issue as discussed in closed issue #292 ]

_Issue_: PyTorch and TensorFlow both implement a function to compute tensor contractions using einstein summation notation (the einsum function). It would be great if Flux would also have this feature.  [This PyTorch commit](https://github.com/wsttiger/pytorch/commit/e8b860eee157bba2b8188632b66d82bc78017224) shows how einsum was implemented there. It seems it can be done just by reshaping the tensors into matrices and permuting indices, then reshaping back into higher order tensors.

_Background_: Tensor contractions generalize the various kinds of tensor products, e.g. matrix-matrix multiplications, matrix-vector multiplications, trace,  dot products, outer products, etc.  For example, if you want to multiply two matrices A and B using Einstein summation notation, you first take their tensor product A x B, which forms a 4th-order tensor (i.e. a tensor with 4 indices). Then you label each index of this tensor with arbitrary labels such as "a,b,b,c." Notice the 2nd and 3rd indices use the same label, so this implies we contract (or sum) over this pair of indices (the indices must be the same size for this to work) resulting in a new tensor C which only has 2 indices (a matrix). See [here](https://physics.stackexchange.com/questions/252473/converting-between-matrix-multiplication-and-tensor-contraction) for more math details of the computation.

_Motivation_: Beyond the convenience of using a single function to implement these various products, you can use tensor contractions to implement tensor networks, which are like deep neural networks except they utilize higher order tensor products and do not generally involve an explicit non-linear function application, and yet they can be used for prediction, classification, and data compression of high-dimensional data. 

_Possible Syntax_: The TensorOperations.jl package implements a nice macro for einstein notation that is different than PyTorch or TF. A contraction of two matrices is represented as `@tensor C[a,c] = A[a,b]*B[b,c]`. In PyTorch (and TF and NumPy) this would be `torch.einsum('ab,bc->ac', (A,B))`. Another possible syntax could be `contract(A,B,(1,2,2,3))` where the first two arguments are the tensors to be contracted and the third argument is a tuple of indices labeled as integers, although this is a bit less expressive.

darsnack:
I think this can be closed. Anyone wanting this feature can use Tullio.jl. But if you want it as part of Flux, then open a new issue about reexporting Tullio.jl + friends.

DhairyaLGandhi:
What's wrong with this one?

[...]

PR #299, opened 2018-06-15
Einsum

MikeInnes:
`@einsum` as discussed in #297. The notation is based on the "function of indices" notation --

```julia
@einsum [i] -> a[i,j]                  # reduce dim 2
@einsum [i,k] -> a[i,j] * b[j,k]       # matmul
@einsum [i,k,N] -> a[i,j,N] * b[j,k,N] # batch matmul
@einsum [i,j] -> a[i] * b[j]           # outer product
```

You can enjoy the nice output, too:

```julia
julia> @expand @einsum [i,k] -> a[i,j] * b[j,k]
:(a * b)

julia> @expand @einsum [i,j] -> a[i] * b[j]
:(a .* (Flux.reshape)(b, (1, (Flux.size)(b, 1))))
```

There are various cases this doesn't handle properly yet, and I need to actually implement that batch-matmul primitive somewhere. Will probably throw together a naive CPU implementation, unless anyone knows a trick.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -26,6 +26,7 @@ export SGD, ADAM, AdaMax, Momentum, Nesterov,
        RMSProp, ADAGrad, ADADelta, AMSGrad, NADAM
 
 include("utils.jl")
+include("einsum.jl")
 include("onehot.jl")
 include("treelike.jl")
 
--- a/src/einsum.jl
+++ b/src/einsum.jl
@@ -1,0 +1,97 @@
+using MacroTools
+
+_permutedims(x, p) =
+  p == collect(1:length(p)) ? x : # TODO 0.7
+  p == [2, 1] ? :(transpose($x)) :
+  :(permutedims($x, ($(p...),)))
+
+_size(x, n) = Any[:(size($x, $i)) for i = 1:n]
+
+_reshape(x, n, s) = # TODO use transpose
+  s == _size(x, n) ? x :
+  :(reshape($x, ($(s...),)))
+
+function _expanddims(x, n, ds) # TODO use transpose
+  is = _size(x, n)
+  foreach(d -> insert!(is, d, 1), ds)
+  :(reshape($x, ($(is...),)))
+end
+
+function _squeezedims(x, n, ds)
+  is = [d for (i, d) in enumerate(_size(x, n)) if i ∉ ds]
+  :(reshape($x, ($(is...),)))
+end
+
+function _einsum_pair(a, b, dims)
+  (a, adims), (b, bdims) = a, b # TODO 0.7
+  preserved = setdiff(intersect(adims, bdims), dims)
+  broadcast = map(adims -> reduce(setdiff, (adims, preserved, dims)), (adims, bdims)) # TODO 0.7
+  # TODO move preserved dims last
+  aperm = sortperm(adims, by = i -> i in preserved ? -1 : i in broadcast[1] ? 0 : 1)
+  bperm = sortperm(bdims, by = i -> i in preserved ? -1 : i in dims ? 0 : 1)
+  a, b = _permutedims(a, aperm), _permutedims(b, bperm)
+  adims, bdims = adims[aperm], bdims[bperm]
+  if isempty(dims)
+    b = _expanddims(b, length(bdims), length(preserved)+(1:length(broadcast[1])))
+    :($a .* $b), vcat(adims[aperm], bdims[bperm][length(preserved)+1:end])
+  else
+    prod(xs) = isempty(xs) ? 0 : length(xs) == 1 ? xs[1] : :(prod(($(xs...),)))
+
+    ashape = _size(a, length(adims))
+    npreserve = prod(ashape[1:length(preserved)])
+    aaxes = 1+length(preserved):length(adims)-length(dims)
+    abroadcast = prod(ashape[aaxes])
+    asum = prod(ashape[end-length(dims)+1:end])
+    a = _reshape(a, length(adims), [abroadcast, asum]) # TODO preserve
+
+    bshape = _size(b, length(bdims))
+    bsum = prod(bshape[length(preserved)+1:end-length(broadcast[2])])
+    baxes = length(bdims)-length(broadcast[2])+1:length(bdims)
+    bbroadcast = prod(bshape[baxes])
+    b = _reshape(b, length(bdims), [bsum, bbroadcast]) # TODO preserve
+
+    ab = :($a*$b)
+    shape = vcat(ashape[[1:length(preserved)..., aaxes...]], bshape[baxes])
+    shape == [abroadcast, bbroadcast] || (ab = _reshape(ab, 2, shape))
+    axes = vcat(adims[[1:length(preserved)..., aaxes...]], bdims[baxes])
+
+    return ab, axes
+  end
+end
+
+# _einsum_pair([:a, [:i, :j]], [:b, [:j, :k]], [:j])
+# _einsum_pair([:a, [:i, :j, :N]], [:b, [:j, :k, :N]], [:j])
+
+macro einsum(ex)
+  @capture(ex, [out__] -> *(in__) | in_) || error("`@einsum [...] -> a[...] * b[...] * ...`")
+  in isa Vector || (in = [in])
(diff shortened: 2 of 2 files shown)


chengchingwen:
I found a [discuss](https://www.reddit.com/r/MachineLearning/comments/6b1qv1/d_numpyeinsum_on_gpu/) including a [relative paper](http://www.mcs.anl.gov/papers/P5361-0615.pdf) about generate GPU code and a [repo](https://github.com/dgasmith/opt_einsum) about optimize `einsum`. Might be useful somehow.

GiggleLiu:
Impressive!

I find
> error("Not supported: index $i appears more than twice")

in this PR, does it mean index appear more than twice is not supported? like contracting stars

In fact, @dclu and I are going to add support to general purposed `Einsum` in our gsoc project *Funny Tensor Networks*. We also have a plan to port `Flux.jl`, it is really nice to hear someone is also working on this direction. I'd like to hear some advices from you after we have finished our proposal.

MikeInnes:
Yeah, that's right. This is a fairly straightforward port from TensorFlow's version if anyone wants to pick it up. I can't remember if that's a limitation of the original or just something I didn't get round to. I do know that in some cases the original ends up with very poor space complexity (since it's just doing things pairwise).

It'd be nice to have this working just as a convenience and hopefully develop something more powerful over time.

GiggleLiu:
In numpy, we can do something like this
```
In [12]: einsum("iii->", random.randn(4,4,4))
Out[12]: 0.3968659339477017
```
It loops over all indices of input tensors, and cumulate the result to the output tensor/scalar. In numpy, `tensordot` is used to perform efficient pairwise contraction, which finally calls into `BLAS`.

datnamer:
Wouldn't it be more fruitful to make sure flux can compose well with existing tensor notation packages?

GiggleLiu:
@datnamer

> Wouldn't it be more fruitful to make sure flux can compose well with existing tensor notation packages?

I think the problem is Flux is unable to backward through inplace functions like `gemm!` - the backends of most tensor packages. This is why tensor packages like `TensorOperations.jl` and `ITensors.jl` can not utilize autodiff defined in `Flux.jl`.

MikeInnes:
Playing well with other packages is definitely the right long-term goal. Right now, `einsum` has the advantage that it's familiar to users coming from Python and it's very easy to implement (and AD, GPU support etc. just fall out).

In future it would certainly be nice to do fancier things, like differentiating tensor notation directly. That's something that'd be very easy to hook into Zygote.

mcabbott:
TensorOperations uses `contract!` which is just a `gemm!` which is less fussy about the order of indices, plus `add!`, `trace!`. I wrote gradients for these three basic functions in  https://github.com/Jutho/TensorOperations.jl/pull/59  after which it should all be differentiable. 

More general contractions are still easy enough to do on paper, so it can't be that hard to teach the computer to do them, and return another tensor contraction. Maybe that's roughly what this GSoC project is? Also cool to hear that ITensors.jl may exist.

TensorCast should be completely differentiable, as it just writes reshapes and broadcasting etc. which are themselves already handled. That's morally the same approach as this PR really (only mine is less functional in style, and ten times as verbose...) I did find some bugs in this PR, which I can probably dig up if someone is interested to work on it. 

(If you do, I would like to put in a vote against this arrow notation, which seems backwards. In `np.einsum` at least it's from input to output, but `out := in` is closer to paper. I would also vote against using the same name as in Einsum.jl.)

GiggleLiu:
> More general contractions are still easy enough to do on paper, so it can't be that hard to teach the computer to do them, and return another tensor contraction. Maybe that's roughly what this GSoC project is?

Yes, the simplicity is why einsum is suited as an IR for implementing back-propagation. Not only contraction, but also stars, trace et. al.
```
# forward     =>     backward code over the first input matrix
ik := ij,jk       =>     ij := ik, jk
ijk := il,jl,kl   =>     il := ijk, jl, kl
:= ii              =>     ii :=        # feeding shape
```
Correct me if some of them are incorrect.

This an elegant way to unify autodiff of trace, gemm!, and other kinds of `einsum`. All we need is a dispatchable `einsum` protocol with `Type of Contraction Topology` (key point) as type for dispatching to achieve best performance.

Then the second step is tensor networks contraction. Efficient contraction sequence algorithm like `treewidth` is needed (especially for quantum computing simulation). We were persuading JuliaGraphs guys to provide an implementation (based on arXiv:1704.05286). I wonder if you would be interested in implementing this algorithm @mcabbott 

Also, matrix factorization backwards is in progress... A lot of Heavy works

> Also cool to hear that ITensors.jl may exist.

Yes, it was release in a Julia meetup in NewYork. The memory independent tensor is really cool, @mfishman will make it open source in the future.

> (If you do, I would like to put in a vote against this arrow notation, which seems backwards. In np.einsum at least it's from input to output, but out := in is closer to paper. I would also vote against using the same name as in Einsum.jl.)

+1

MikeInnes:
FWIW, that syntax is motivated by the fact that `[i,j] -> a[i] * b[j]` is effectively equivalent to a lambda (e.g. `(i, j) -> a[i] * b[j]`, which would actually work already). It's not backwards if you see the indices as being the inputs. Seeing arrays as functions is admittedly unusual for numerical folks but also pretty foundational to compilers like Halide. `:=` seems like a perfectly reasonable alternative anyhow.

mcabbott:
OK, I see what you mean, this is the function you `map` over LHS indices. 

Re rules, those do look correct. Although I'd prefer to say that there's just one rule, ∂M_ab / ∂M_cd = δ_ac δ_bd for matrices, etc. If your output is some T_ab, then for backward mode you always want Δ_ab ∂T_ab / ∂M_cd. Then you can simplify δ_ac Z_xya = Z_xyc etc, but sometimes (as for your trace rule) δ survives. 

If you have many factors, then there will be a lot of overlap between the backward mode calculations for them. As you say it seems graph-like, in that each ∂T / ∂M removes just one M and sews in Δ there. I presume that exploiting this would become important. But it would take a me a long time to decode this arxiv paper! And perhaps this is getting off-topic for this PR.

bionicles:
Tullio does a decent job at this: https://github.com/mcabbott/Tullio.jl

The syntax for this is much simpler to reason about and I would love to have this in Flux

ToucheSir:
Yup, shall we close this since Tullio ticks all the boxes (AD, GPU, etc.)?

CarloLucibello:
yes, Tullio works great. I don't think  there is the need to reexport it from Flux, but if someone wants that we may consider it

DhairyaLGandhi:
I definitely don't think we would re-export tullio, but can keep this one around since it's a lightweight implementation of the same concept so in the future, if we have usecases for lighter dependencies for workloads, this is a viable route forward.

darsnack:
Why keep it open? Closed PRs don't disappear, and we don't need to delete the branch either. To some extent, the issue/PR inbox signals something to external users. Keeping PRs open if we have no current plans to merge is not great IMO.

DhairyaLGandhi:
Closing suggests it's not going to be considered, which is somewhat worse. An open pr isn't going to hurt as much.

ToucheSir:
I would differentiate between closing an issue ("we don't plan on implementing this") and closing a PR ("we don't plan on using this specific implementation"). Unless we are seriously considering reviving this specific implementation, I don't think it's worth keeping open. #297 may be as long as it's made clear that Tullio works _right now_ and something may or may not be created in Flux itself at some point.

darsnack:
Yeah I think closing this doesn't imply we will never consider something like `@einsum`. If anything, it's the opposite. We did consider it, decided that other packages provide this functionality quite well, and there is currently no need for Flux itself to provide it. Closing this makes clear what the decision was.

To be clear, I am not against this particular PR, I just think that closing stale PRs and issues is good practice for any open source project.

darsnack:
> #297 may be as long as it's made clear that Tullio works right now and something may or may not be created in Flux itself at some point.

Yeah if we want to track this issue, then keep the issue open. But again, we can just as easily close the issue, if we are seriously not considering it, and just say "please re-open this issue if you would like this feature built into Flux."

DhairyaLGandhi:
Fair, let's keep the branch around and revisit if needed

PR #304, opened 2018-06-21
Add Explicit Vector Jacobian Products to Flux

jessebett:
This is a start to vjp api in Flux. It's kinda buggy and inconsistent from an API perspective, but the math works.

Try running:
```
Flux.make_vjp(g)([6.],[6.],[6.];e=2) #returns vjp(v) and function eval
Flux.make_vjp(g)([6.],[6.],[6.];e=2)[1]([1.;1.]) #evaluate vjp at v
Flux.make_vjp(g,argnum=[1,3])([6.],[6.],[6.];e=2)[1]([1.,1.]) #only compute grads at specific args
Flux.make_vjp(g)([1.,2.],[1.,2.],[1.,2.];e=2)[1]([1.,1.]) #can handle nonscalar vjps

# what doesn't work:
# vjps of vjps cause flux can't do higher order
# argnum breaks if only integer is provided not array
# weird behaviour if args are different types depends on ordering??:
Flux.make_vjp(g)([6.],6.,[6.];e=2)[1]([1.;1.]) #works
Flux.make_vjp(g)(6.,[6.],[6.];e=2)[1]([1.;1.]) #breaks
```

### DIFF
--- a/src/tracker/Tracker.jl
+++ b/src/tracker/Tracker.jl
@@ -10,6 +10,7 @@ istracked(x) = tracker(x) ≠ nothing
 isleaf(x) = !istracked(x) || isleaf(tracker(x))
 data(x) = istracked(x) ? data(tracker(x)) : x
 grad(x) = grad(tracker(x))
+grad(::Void)=nothing
 
 struct Call{F,As<:Tuple}
   func::F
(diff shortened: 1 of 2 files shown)


jessebett:
Nope! I think the new forward / back interface does this exactly! Thanks Mike it looks great!

PR #306, opened 2018-06-26
Add epsilon term to binarycrossentropy

maetshju:
This PR adds a machine epsilon term to `binarycrossentropy`. I chose to have the epsilon term as a keyword argument so that a user could choose their own value or set it to 0 to avoid using the term.

EDIT: This is in reference to https://github.com/FluxML/Flux.jl/pull/145

### DIFF
--- a/test/layers/stateless.jl
+++ b/test/layers/stateless.jl
@@ -1,6 +1,8 @@
 using Base.Test
-using Flux: onehotbatch, mse, crossentropy, logitcrossentropy, 
+using Flux: onehotbatch, mse, crossentropy, logitcrossentropy,
             σ, binarycrossentropy, logitbinarycrossentropy
+
+const ϵ = 1e-7
 
 @testset "losses" begin
   # First, regression-style y's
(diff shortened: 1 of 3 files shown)


MikeInnes:
Can we do `eps(y)` for the epsilon? Just avoids it being f64 specific.

maetshju:
Absolutely1 I didn't know that function existed. I'll get on it right now.

test/layers/stateless.jl
```diff
@@ -1,7 +1,9 @@
 using Base.Test
-using Flux: onehotbatch, mse, crossentropy, logitcrossentropy, 
+using Flux: onehotbatch, mse, crossentropy, logitcrossentropy,
             σ, binarycrossentropy, logitbinarycrossentropy
 
+const ϵ = 1e-7
```
CarloLucibello: default to `1fe-7` instead, so that we don't obtain Float64 outputs when using Float32 parameters?

CarloLucibello: ah ok, never mind, just sow Mike's comment

CarloLucibello:
> Can we do eps(y) for the epsilon? Just avoids it being f64 specific.

Is eps(y) large enough to avoid numerical problems?

CarloLucibello:
Also, you may want `eps(ŷ)` since y could be integer

maetshju:
I can try running this with the network that was giving me numerical stability problems and see if they still crop up with eps(ŷ).

maetshju:
With the epsilon as `eps(ŷ)`, I did not encounter any errors with the network I was using and ran a full 200 epochs. With the epsilon value set to 0, I couldn't make through 5.

To use `ŷ` to determine the epsilon, I had to call `Flux.Tracker.Value()` on it, which doesn't feel that elegant a solution to me.

MikeInnes:
Overloading `eps` for `TrackedReal` should be enough to resolve that.

MikeInnes:
:+1: Thanks!

PR #307, opened 2018-07-03
Add ADAMW "Fixing Weight Decay Regularization in Adam"

jarvist:
I really don't know what I'm doing, so this code may be all wrong. 

The motivation behind this method is in this blog post:
http://www.fast.ai/2018/07/02/adam-weight-decay/
And the original paper:
https://arxiv.org/abs/1711.05101.pdf

I just tried to add the relevant bits in Algorithm 2 of the paper. It moves the application of decay to the parameter update. (i.e. it becomes a mixing parameter). 

The paper argues that the L2 regularisation (which is what I believe `invdecay` is doing) are not equivalent to proper weight decay, are not equivalent when you have adaptive gradients.

Just an hours fun hacking one morning - I tested it against a toy Char-RNN I found on my harddrive, and it seems to improve the learning rate epoch-by-epoch over ADAM with the same decay parameter.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -22,7 +22,7 @@ export Tracker, TrackedArray, TrackedVector, TrackedMatrix, param
 include("optimise/Optimise.jl")
 using .Optimise
 using .Optimise: @epochs
-export SGD, ADAM, AdaMax, Momentum, Nesterov,
+export SGD, ADAM, ADAMW, AdaMax, Momentum, Nesterov,
        RMSProp, ADAGrad, ADADelta, AMSGrad
 
 include("utils.jl")
--- a/src/optimise/Optimise.jl
+++ b/src/optimise/Optimise.jl
@@ -1,7 +1,7 @@
 module Optimise
 
 export train!,
-  SGD, ADAM, AdaMax, Momentum, Nesterov, RMSProp, ADAGrad, ADADelta, AMSGrad
+  SGD, ADAM, ADAMW, AdaMax, Momentum, Nesterov, RMSProp, ADAGrad, ADADelta, AMSGrad
 
 struct Param{T}
   x::T
--- a/src/optimise/interface.jl
+++ b/src/optimise/interface.jl
@@ -1,7 +1,7 @@
 call(f, xs...) = f(xs...)
 
 # note for optimisers: set to zero
-# p.Δ at the end of the weigths update
+# p.Δ at the end of the weights update
 function optimiser(ps, fs...)
   ps = [Param(p) for p in ps]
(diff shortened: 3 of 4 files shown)


MikeInnes:
Looks good! There are some tests for the optimisers in the tests dir – just adding ADAMW to the array there should be enough to add a test for this.

I'd appreciate a quick thumbs up from anyone else who knows the mathematics and can quickly verify it.

jarvist:
I certainly appreciate another set of eyes!
If the AdamW paper is correct, this implies that all adaptive gradient optimisers are typically incorrectly L2 regularised rather than actually weight decayed. 
As such, `invdecay` was not actually the intended mathematics of these methods.

jarvist:
Well adding ADAMW does run in the tests!
I must admit I'm slightly at a loss as to what the optimise tests do; 
https://github.com/FluxML/Flux.jl/blob/ce88273880730990ef2e236b775b2080eca12f4a/test/optimise.jl#L4-L17
Try and make a 10x10 random (normally distributed) matrix Hermitatian, by minimising the mean-squared error of the matrix (and it's transpose) acting on a random vector? 
Is this a good proxy for optimisation? Can you make any statement on learning rate from this?

As ADAMW===ADAM when the weight-decay is zero, should probably try them for some finite value of this.

CarloLucibello:
> Try and make a 10x10 random (normally distributed) matrix Hermitatian, by minimising the mean-squared error of the matrix (and it's transpose) acting on a random vector?
Is this a good proxy for optimisation? Can you make any statement on learning rate from this?

I think is the intent here is to test the optimizers in the simplest convex scenario, a quadratic function. Doing `w'w`  assures that the form is positive definite, so that you actually do have a minima. If I rember correctly, optimal learning rate in this case should be the Lipschitz coefficient of the gradient

jarvist:
Thanks for the pointers Carlo. I had a quick play - it seems that adding weight decay / regularisation seems to reduce both Adam and AdamW in the simple quadratic convex optimisation. 
Maybe a slightly more challenging toy system might be worth while to add? Or a different loop with << 10^5 steps?

MikeInnes:
I'll merge this for now. Could be good to add another test, but I'm not sure what'll be meaningful for this beyond making sure the equation is right.

jarvist:
Grand! 
It would be nice to have a slightly more challenging (even if still 'toy') test for the various optimisation. I guess there must be some simple system / potential-energy surface where weight-decay will benefit?

CarloLucibello:
the rosenbrock function would be a good candidate https://en.wikipedia.org/wiki/Rosenbrock_function

PR #308, opened 2018-07-05
Adapt to changes in NNlib to use `mode` externally

ayush1999:
This works on the changes made in https://github.com/FluxML/NNlib.jl/pull/53 . 
(To use mode directly from Flux).

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -15,7 +15,7 @@ Standard convolutional layer. `size` should be a tuple like `(2, 2)`.
 Data should be stored in WHCN order. In other words, a 100×100 RGB image would
 be a `100×100×3` array, and a batch of 50 would be a `100×100×3×50` array.
 
-Takes the keyword arguments `pad`, `stride` and `dilation`.
+Takes the keyword arguments `pad`, `stride`,`dilation` and `flipkernel`.
 """
 struct Conv{N,F,A,V}
   σ::F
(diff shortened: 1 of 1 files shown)


MikeInnes:
I think the API needs some thought here, and I'd rather not 0/1 (which are unclear to read). Maybe a kwarg like `reflect=true`?

I'm almost tempted to just have a separate `CrossCorrelation` layer for this, but it's probably not worth that.

ayush1999:
@MikeInnes I've added  `reflect` as an attribute of `Conv`. When `true`, reflects internally acts similar to `mode=0`.

philtomson:
It seems like convolution mode should be the default with cross correlation as the option.

MikeInnes:
I believe we already do; `mode = 0` corresponds to convolution in CUDNN.

ayush1999:
@philtomson The nomenclature is actually a bit confusing. `Convolution` means flipping the kernel and then adding the element wise products. `Cross Correlation` is the same as Convolution, but without flipping the kernel. Flux by default flips the kernel, and hence does Convolutions. I think it'd be better to have a kwarg with `mode="convolution"` or `mode="cross-correlation"`, so that there isn't any confusion.

philtomson:
Yeah, I don't like plain strings for these kinds of things (It's all the rage in Python-land, but I think we can do better).

It would be good if it was some kind of enum type, I think you can do this with @enum:

`@enum Choice convolution correlation`

MikeInnes:
Are we doing #423 instead?

ayush1999:
Yes, there was a plan to add an option to `Conv` initially, but that changed later. Closing this.

Issue #86, opened 2017-10-17
Functional AD


MikeInnes:
The eventual plan is to build a new compiler-level AD that better exploits Julia's compilation, provides a more function interface, and supports nested differentiation. A question here is how to support the `grad(f, x)` style interface while also still allowing abstraction and modularity in layers and their weights.

I see this looking something like:

```julia
W = randn(5,5)
b = randn(5)
loss(x, y) = mse(W*x .+ b, y)
dW, db = grad(loss, (W, b), x, y)
```

`W` and `b` are treated as implicit arguments to the function; this is nice in that it's essentially the ideal functional interface but without the mess of hundreds of explicit arguments.

Models will implement `params`, as they do now, and whatever arrays they return will be treated as trainable parameters (`dparams = grad(model, params(model), args...)`). We'll also have a `Freeze` layer to treat things as constant, e.g. `m = Freeze(Dense(10, 5)); params(m) == []`. Freezing parameters is a little more coarse-grained compared to now, but that's small loss compared to the gains.

dfdx:
> `W` and `b` are treated as implicit arguments to the function; this is nice in that it's essentially the ideal functional interface but without the mess of hundreds of explicit arguments.

Does this mean that a user will need to make `W` and `b` global and `loss` to be defined in the same context? This doesn't sound very flexible, to be honest. Recently I've been playing around (e.g. in [VariationalAE.jl](https://github.com/dfdx/VariationalAE.jl/blob/xdiff-based/src/core.jl)) with models as mutable structs. For your case it would look something like:

```
m = Linear(W, b)
loss(m::Linear, x, y) = mse(m.W * x .+ m.b, y)
dm = grad(loss, m, x, y)
```
where `dm` is another instance of `Linear` holding derivatives, i.e.:

```
dW = dm.W
db = dm.b
```

MikeInnes:
> Does this mean that a user will need to make W and b global and loss to be defined in the same context?

Er, no? For the most part I'm not expecting any thing else to look different; so the MNIST example would stay exactly the same. It's really no different to the current `TrackedArray` approach in that sense, just without the hacky overloading.

Your API is something we discussed as it's closer to what Knet currently has. At a minimum it only scales up well if you allow the structure to define the forward pass (e.g. via call overloading). Even then it imposes a bigger burden on user-defined types and small models, and it's harder to figure out how it plays when you get to really complex models (as one example, higher-order models that take another model as input).

dfdx:
[...]

PR #313, opened 2018-07-09
AD Overhaul

MikeInnes:
This is something of a PSA, as this patch is badly breaking for any custom derivative implementations. I don't think this is overly common, so I think it's best if we just help fix those up as they come along.

I've overhauled the internal gradient definition API to more cleanly separate gradient definitions from `Tracked` interception. This should make the eventual switch to a better AD non-breaking (just clean out the interception cruft and you're done).

- [x] CUDNN
- [x] Docs updates

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,4 @@
 julia 0.6.0
-DataFlow 0.2.1
 Juno
 MacroTools 0.3.3
 NNlib
--- a/docs/src/internals/tracker.md
+++ b/docs/src/internals/tracker.md
@@ -6,6 +6,52 @@ Backpropagation, or reverse-mode automatic differentiation, is handled by the `F
 julia> using Flux.Tracker
 ```
 
+Here we discuss some more advanced uses of this module, as well as covering its internals.
+
+## Taking Gradients
+
+In the [basics section](../models/basics.md) we covered basic usage of the `gradient` function.
+
+```julia
+using Flux.Tracker
+
+Tracker.gradient((a, b) -> a*b, 2, 3) # (3.0 (tracked), 2.0 (tracked))
+```
+
+`gradient` is actually just a thin wrapper around the backpropagator-based interface, `forward`.
+
+```julia
+using Flux.Tracker: forward
+
+y, back = forward((a, b) -> a*b, 2, 3) # (6.0 (tracked), Flux.Tracker.#9)
+
+back(1) # (3.0 (tracked), 2.0 (tracked))
+```
+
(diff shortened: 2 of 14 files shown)


MikeInnes:
This ended up going quite well, and I've added a few handy things:
* Checkpointing (trade compute for memory in large models)
* Gradient hooks
* Nested AD (though many primitives need updating for it)
* RNN parameters can now be derived from other model inputs
* Memory efficiency improvements for some primitives.

baggepinnen:
May I just ask, does "nested AD" mean one can differentiate expressions involving gradients (second order AD)?

MikeInnes:
Exactly – check out the [first example in the new docs](http://fluxml.ai/Flux.jl/latest/models/basics.html) :)

baggepinnen:
Cool! If I would like to add some of the missing primitives, how would I go about doing that?
I encounter the error thrown by this line
```julia
@grad nobacksies(f, x) = data(x), Δ -> error("Nested AD not defined for $f") # f = broadcast in my case
```
So I guess this is the function I need to implement for the operations I want to differentiate, e.g. `nobacksies(:broadcast, x) = something` and possibly `@grad broadcast` and `@grad nobacksies(:broadcast, ...`. Is this true or do I need to define a [custom gradient](http://fluxml.ai/Flux.jl/stable/internals/tracker.html#Custom-Gradients-1)? I can't seem to figure out what `nobacksies` is doing

MikeInnes:
`nobacksies` is just a guard that stops you from trying to backpropagate. I use it whenever calling `data(Δ)` because doing that drops the gradient of Δ. Otherwise, trying to backpropagate would silently give wrong answers.

Ideally, you don't need `nobacksies` at all – see [`Ac_mul_B`](https://github.com/FluxML/Flux.jl/pull/313/files#diff-ce4369f5fcf2de3495029b8e5a2c503bR308) for an example. This just works,  because `A_mul_Bt`, `transpose`, `*` etc are all themselves differentiable. That's all it takes to make nesting work. Doing this for broadcast might be a bit trickier given the interaction with dual numbers, but it should be doable.

Let me know if that's unclear. This is all more subtle than I'd like right now, but it's all going to go away with #86.

baggepinnen:
In fact, my particular example returns the correct result (apart from a factor of 2) if I simply remove [the line](https://github.com/FluxML/Flux.jl/blob/master/src/tracker/array.jl#L380) `nobacksies(:broadcast, dxs)`

It seems `nobacksies` is just a way of saying "I {can't/ haven't yet thought about} nested differentiation through this function?

PR #327, opened 2018-07-17
Julia 0.7 fixes

pevnak:
I have tried to fix the problem with `sum` taking optional parameter. Besides, I have fixed some deprecation warnings, one in the broadcast.

### DIFF
--- a/src/cuda/cudnn.jl
+++ b/src/cuda/cudnn.jl
@@ -83,7 +83,7 @@ function RNNDesc{T}(mode::Int, input::Int, hidden::Int; layers = 1) where T
   @check ccall((:cudnnSetRNNDescriptor_v6,libcudnn), cudnnStatus_t, (Ptr{Nothing},Ptr{Nothing},Cint,Cint,Ptr{Nothing},Cint,Cint,Cint,Cint,Cint),
     libcudnn_handle[],d[],hidden,layers,dropoutDesc,inputMode,direction,mode,algo,cudnnDataType(T))
 
-  w = cuzero(T, rnnParamSize(T, d[], input))
+  w = cuzeros(T, rnnParamSize(T, d[], input))
   # TODO: avoid reserve allocation here
   rd = RNNDesc{T}(mode, input, hidden, w, params(w, input, hidden, ngates(mode))..., d[])
   finalizer(rd, x ->
--- a/src/cuda/cudnn.jl
+++ b/src/cuda/cudnn.jl
@@ -83,7 +83,7 @@ function RNNDesc{T}(mode::Int, input::Int, hidden::Int; layers = 1) where T
   @check ccall((:cudnnSetRNNDescriptor_v6,libcudnn), cudnnStatus_t, (Ptr{Nothing},Ptr{Nothing},Cint,Cint,Ptr{Nothing},Cint,Cint,Cint,Cint,Cint),
     libcudnn_handle[],d[],hidden,layers,dropoutDesc,inputMode,direction,mode,algo,cudnnDataType(T))
 
-  w = cuzeros(T, rnnParamSize(T, d[], input))
+  w = cuzero(T, rnnParamSize(T, d[], input))
   # TODO: avoid reserve allocation here
   rd = RNNDesc{T}(mode, input, hidden, w, params(w, input, hidden, ngates(mode))..., d[])
   finalizer(rd, x ->
@@ -198,7 +198,7 @@ end
 
 function backwardData(rnn::RNNDesc{T}, y, dy_, dho, dco, h, c, reserve) where T
   # Same as above, any more efficient way?
-  dy = dy_ isa Integer ? zeros(y) : dy_
+  dy = dy_ isa Integer ? zero(y) : dy_
   yd = xDesc(y)
   dx = y isa AbstractVector ? similar(dy, rnn.input) : similar(dy, rnn.input, size(dy, 2))
   dh = similar(h)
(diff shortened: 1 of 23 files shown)


src/tracker/array.jl
```diff
@@ -26,7 +27,7 @@ TrackedArray(c::Call, x::A) where A <: AbstractArray =
 TrackedArray(c::Call, x::A, Δ::A) where A <: AbstractArray =
   TrackedArray{eltype(A),ndims(A),A}(Tracked{A}(c, Δ), x, Δ)
 
-TrackedArray(x::AbstractArray) = TrackedArray(Call(), x, zeros(x))
+TrackedArray(x::AbstractArray) = TrackedArray(Call(), x, similar(x) .= 0)
```
CarloLucibello: I think that the return value of .=  should not be used, since there was no agreement on what it should be. I cannot remember where this was discussed though. What about using `zero(x)`?

test/tracker.jl
```diff
@@ -14,10 +14,13 @@ gradtest(f, dims...) = gradtest(f, rand.(dims)...)
 @test gradtest((x, W, b) -> logσ.(W*x .+ b), 5, (2,5), 2)
 @test gradtest((x, W, b) -> logσ.(W*x .+ b), (5,3), (2,5), 2)
 
-@test gradtest((w, x) -> w'*x, randn(10, 2), randn(10))
-@test gradtest((w, x) -> w*x', randn(5,5), randn(5,5))
+@test gradtest((w, x) -> w'*x, randn(Float64,10, 2), randn(Float64,10))
+@test gradtest((w, x) -> w*x', randn(Float64,5,5), randn(Float64,5,5))
```
CarloLucibello: out of curiosity, why add Float64? randn(5,5) seems to be fine

pevnak: It was because at some moment, Julia-0.7 was warning me that randn will be deprecated in favour of randn(Float64,...). But I can revert this.

pevnak:
Hi Carlo, this is the deprecation warning I am getting at the moment when I use rand(2,3). I am on the latest beta.
```
julia> gradtest((x, W, b) -> σ.(W*x .+ b), 5, (2,5), 2)
┌ Warning: `rand(dims::Dims)` is deprecated, use `rand(Float64, dims)` instead.
│   caller = _broadcast_getindex_evalf at broadcast.jl:579 [inlined]
└ @ Core ./broadcast.jl:579
```

src/cuda/cudnn.jl
```diff
@@ -83,7 +83,7 @@ function RNNDesc{T}(mode::Int, input::Int, hidden::Int; layers = 1) where T
   @check ccall((:cudnnSetRNNDescriptor_v6,libcudnn), cudnnStatus_t, (Ptr{Nothing},Ptr{Nothing},Cint,Cint,Ptr{Nothing},Cint,Cint,Cint,Cint,Cint),
     libcudnn_handle[],d[],hidden,layers,dropoutDesc,inputMode,direction,mode,algo,cudnnDataType(T))
 
-  w = cuzeros(T, rnnParamSize(T, d[], input))
+  w = cuzero(T, rnnParamSize(T, d[], input))
```
MikeInnes: I think this one should not be changed

src/tracker/array.jl
```diff
@@ -204,27 +205,31 @@ Base.kron(a::AbstractMatrix, b::TrackedMatrix) = _kron(a, b)
 
 # Reductions
 
-Base.sum(xs::TrackedArray, dim) = track(sum, xs, dim)
+Base.sum(xs::TrackedArray; dims) = track(sum, xs, dims)
```
MikeInnes: FYI, you can also call `track(sum, xs, dims = dims)` to pass a keyword argument. This is handy as you don't have to pass a `nothing` derivative for keyword arguments.

src/tracker/array.jl
```diff
@@ -204,27 +205,31 @@ Base.kron(a::AbstractMatrix, b::TrackedMatrix) = _kron(a, b)
...
 Base.sum(xs::TrackedArray) = track(sum, xs)
 Base.sum(f::Union{Function,Type},xs::TrackedArray) = sum(f.(xs))
 
-@grad sum(xs, dim...) = sum(data(xs), dim...),
-  Δ -> (zero(xs) .+ Δ, map(_->nothing,dim)...)
+@grad sum(xs, dims::Int) = sum(data(xs), dims = dims),
+  Δ -> (zero(xs) .+ Δ, nothing)
+@grad sum(xs, dims) = sum(data(xs), dims = dims),
+  Δ -> (zero(xs) .+ Δ, map(_->nothing,dims)...)
```
MikeInnes: This is unnecessary, because there's only one `dims` positional argument (so you can just return a single `nothing` as on line 213)

pevnak: Weird that without this, it has not worked. I will take a look on that.

pevnak:
This test
```
@testset "Fallbacks" begin
  xs = param([1 2; 3 4])
  @test similar(xs) isa Matrix{Float64}
  # Remove this test if we do LowerTriangular properly
  L = LowerTriangular(xs)
  @test L*L' isa Matrix{TrackedReal{Float64}}
end
```

does not pass, becase LowerTriangular cannot be constructed from the TrackedArray. It quits with the following error
```
ERROR: TypeError: in LowerTriangular, in S, expected S<:AbstractArray{T,2}, got Type{TrackedArray{…,Array{Float64,2}}}
```
 It seems to me that it is because of
```
julia> typeof(xs) <: AbstractArray{T,2} where {T}
true
```

but not 
```
julia> typeof(A) <: AbstractArray{T,2} where {T<: TrackedReal{Float64}}
false
julia> T = eltype(xs)
TrackedReal{Float64}
```

pevnak:
We have merged latest julia-0.7 branch and all test pass now.

MikeInnes:
Thanks a lot @pevnak, as well as @CarloLucibello and @Sheemon7!

PR #339, opened 2018-07-31
Add Maxpool and Meanpool for convention.

yuehhua:


MikeInnes:
Looks good, thanks! A couple comments:

I think `MaxPool` and `MeanPool` would look better.

It'd be nice if I could provide the keyword arguments as scalars and have them expanded to tuples (as with convolutions now).

Could you also add some simple tests (doesn't have to do much more than checking it can be constructed and called).

yuehhua:
Done! Is it good?

yuehhua:
@MikeInnes ?

MikeInnes:
Thanks @yuehhua! I tweaked things a little to be closer to what I had in mind earlier (but didn't explain well).

PR #342, opened 2018-08-07
Adding CTC loss

maetshju:
This is the CTC loss function and associated tests. it should interface with the tracker just like the other loss functions, and it uses the `@require` macro to allow the GPU functionality to be optional.


src/layers/ctc.jl
```diff
@@ -0,0 +1,476 @@
...
+  for (i,elem) in enumerate(A)
+    if i > 1 && A[i] == A[i-1]
+      repeats += 1
+    end
+  end
+  return repeats
+end
+
+@require CUDAnative begin
+@require CuArrays begin
```
MikeInnes: This can just be `@require CuArrays`, since CUDAnative will always be a dependency.

These blocks will also need an update for 1.0.

src/layers/ctc.jl
```diff
@@ -0,0 +1,476 @@
...
+  return mean(ls), gs
+end
+end # end of `@require CuArrays`
+end # end of `@require CUDAnative`
+
+ctc(ŷ::TrackedArray, y::AbstractArray) = Flux.Tracker.track(ctc, ŷ, y)
+
+@grad function ctc(ŷ, y)
+  ls, gs = ctc(Flux.Tracker.data(ŷ), Flux.Tracker.data(y))
+  return ls, Δ -> (Δ .* gpu(gs), Δ)
```
MikeInnes: Why is the `gpu` call needed here? We should find a different way to do this.

test/layers/ctc.jl
```diff
@@ -0,0 +1,25 @@
...
+  gradvalues = [-0.317671 -0.427729 0.665241; 0.244728 -0.0196172 -0.829811; 0.0729422 0.447346 0.16457]
+
+  l, gs = ctc(x, y)
+  @test l ≈ lossvalue
+  @test any(x -> !x, gs ≈ gradvalues)
+  @require CUDAnative begin
+    @require CuArrays begin
+      using CuArrays
+      lossvalue = 3.6990738
+      l, gs = ctc(Flux.gpu(x), Flux.gpu(y))
```
MikeInnes: Can we test the CPU version as well, and make sure the outputs are the same?

Worth also having a numerical gradcheck? Might not be necessary if we have known-good values.

maetshju:
I should be able to get to this after a conference I'm going to this week. Doesn't look like it should be too much work; and I think it's a good idea to add it to the docs pages.

MikeInnes:
I wonder if the core kernel should be added to CuArrays so we can make sure that's well tested, and then we can add the interface parts to Flux separately. (@maleadt does that sound reasonable to you?)

maleadt:
Unless there's an NNlib interface to implement with this functionality, I'd rather not. We should think about where to put functionality like that, and I think of CuArrays as implementing existing array-like interfaces for the GPU rather than housing "arbitrary" functionality.

MikeInnes:
Seems fair.

@maetshju what's the status on this, is it ready on your end, or are you still hacking on it?

maetshju:
@MikeInnes the core functionality should be there. I am going to make a pass through all the code over the weekend to do a few more tests and make sure the documentation/comments are sufficient and correct, but I think it should be good to go after that.

maetshju:
I believe this is ready to go. It consistently passes the tests I've written, and I've checked that the methods properly cover all the 16 possible cases of tracked/untracked cu/non-cu array combinations. The docs should be up to date as well.

If it looks good on your end, would you like me to squash the commits together before merging?

dhairyagandhi96:
I'm going to do a quick run on the GPU also

bors try

maetshju:
I'm not sure I understand why that previous build failed. It doesn't look like Docker got everything installed before failing.

maetshju:
These last few commits fix the merge conflict, and I think they get everything working with Zygote. Admittedly, I'm not super familiar with Zygote yet, but the CPU tests are passing. I haven't had a chance to check on GPU yet though.

MikeInnes:
Thanks a lot. Will give the GPU tests a quick go to see where we are:

bors try

maetshju:
That most recent commit should fix the `UndefVarError` by using `CuArrays.functional()`. It's passing tests in my GPU environment, so hopefully it works with the CI.

MikeInnes:
bors try

Manifest.toml
```diff
@@ -74,6 +80,12 @@ git-tree-sha1 = "efdaf19ab11c7889334ca247ff4c9f7c322817b0"
 uuid = "bbf7d656-a473-5ed7-a52c-81e309532950"
 version = "0.2.0"
 
+[[Conda]]
```
bhvieira: Is this necessary? I haven't followed the PR, though.

maetshju: That's a good catch. The dependency was induced by FFTW's version 1.1.0, which in turn was induced by its MKL_jll and FFTW_jll dependencies requiring Julia 1.3, while I was running 1.1. I upgraded my version of Julia and then updated packages, which should remove the dependency on Conda.

maetshju: It should still install on older versions of Julia now, since it can pull in the older version of FFTW that requires Conda. Or at least, I was able to install this new version from the pull branch on Julia 1.1.

bhvieira: I think so. Couldn't imagine FFTW would induce the Conda dependency indirectly, will keep that in mind!

CarloLucibello:
@maetshju it is a pity if this gets lost. There are two changes impacting this PRs, #1258 and #1264 . If you will have the patience to rebase once #1264 gets in, we can have a quick review and merge

CarloLucibello:
starting a new PR all over may be easier than rebase

maetshju:
@CarloLucibello Thanks for letting me know about those PRs. I think opening a new PR once the others have been merged is probably easier, as you suggested. I would love to get this committed at some point; it's been lingering in my mind for a while.

maetshju:
Superseded by https://github.com/FluxML/Flux.jl/pull/1287.

Issue #323, opened 2018-07-16
define ≈ for tracked array


CarloLucibello:
right now we have equality but approx is missing: 
```julia
julia> a=param(rand(2))
Tracked 2-element Array{Float64,1}:
 0.568512
 0.873907

julia> a == a
true

julia> a ≈ a
ERROR: <= not defined for Flux.Tracker.TrackedReal{Float64}
Stacktrace:
 [1] <=(::Flux.Tracker.TrackedReal{Float64}, ::Flux.Tracker.TrackedReal{Float64}) at ./promotion.jl:351
 [2] #isapprox#14(::Float64, ::Int64, ::Bool, ::Base.LinAlg.#vecnorm, ::Function, ::TrackedArray{…,Array{Float64,1}}, ::TrackedArray{…,Array{Float64,1}}) at ./linalg/generic.jl:1299
 [3] isapprox(::TrackedArray{…,Array{Float64,1}}, ::TrackedArray{…,Array{Float64,1}}) at ./linalg/generic.jl:1297
 [4] macro expansion at ./REPL.jl:97 [inlined]
 [5] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73
```

johnnychen94:
@MikeInnes I think it's okay to close this issue now.

MikeInnes:
Seems so, thanks both.

Issue #301, opened 2018-06-17
vecnorm error in Regularisation tutorial


agentp5:
I get the following error:
```
MethodError: no method matching norm(::TrackedArray{…,Array{Float64,0}})[0m
Closest candidates are:
  norm([91m::RowVector[39m) at linalg\generic.jl:595
  norm([91m::RowVector[39m, [91m::Real[39m) at linalg\generic.jl:607
  norm([91m::Union{Base.ReshapedArray{T<:Union{Complex{Float32}, Complex{Float64}, Float32, Float64},1,A,MI} where MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N} where N} where A<:Union{DenseArray, SubArray{T,N,P,I,true} where I<:Tuple{Union{Base.Slice, UnitRange},Vararg{Any,N} where N} where P where N where T}, DenseArray{T<:Union{Complex{Float32}, Complex{Float64}, Float32, Float64},1}, SubArray{T<:Union{Complex{Float32}, Complex{Float64}, Float32, Float64},1,A,I,L} where L} where I<:Tuple{Vararg{Union{Base.AbstractCartesianIndex, Int64, Range{Int64}},N} where N} where A<:Union{Base.ReshapedArray{T,N,A,MI} where MI<:Tuple{Vararg{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64},N} where N} where A<:Union{DenseArray, SubArray{T,N,P,I,true} where I<:Tuple{Union{Base.Slice, UnitRange},Vararg{Any,N} where N} where P where N where T} where N where T, DenseArray}[39m, [91m::Union{Range{TI<:Integer}, UnitRange{TI<:Integer}}[39m) where {T<:Union{Complex{Float32}, Complex{Float64}, Float32, Float64}, TI<:Integer} at linalg\dense.jl:87
  ...
generic_vecnormInf(::TrackedArray{…,Array{Float64,2}}) at generic.jl:328
generic_vecnorm2(::TrackedArray{…,Array{Float64,2}}) at generic.jl:356
vecnorm(::TrackedArray{…,Array{Float64,2}}, ::Int64) at generic.jl:445
vecnorm(::TrackedArray{…,Array{Float64,2}}) at generic.jl:443
_mapreduce(::Base.LinAlg.#vecnorm, ::Base.#+, ::IndexLinear, ::Array{Any,1}) at reduce.jl:273
[...]

PR #358, opened 2018-08-18
Fix issues #354, #174, #301, #323, add Unittest for basic.jl, ignore .DS_Store and Project.toml

### LINKED ISSUE #354: Bug: broadcasting should be avoided in Dense
When I read the [Basics](http://fluxml.ai/Flux.jl/stable/models/basics.html) part of Flux tutorials, I found a behavior out of my expectation, that scalar value is accidentally broadcasted in the calculation, and hence output useless values without warning or error.

```julia
using Flux
# model input should be 10-element array
model = Chain(
  Dense(10, 5, σ),
  Dense(5, 2),
  softmax)

model(randn(10)) # 2-element array <----- expected
model(randn(1)) # raise DimensionMismatch <---- expected
model(1) # Tracked 2×10 Array <---- unexpected behavior, should be avoid as far as I concerned
```

I'm new to Flux, so I'm not sure if this is a feature or not. And I'm not sure if other layers has this kind of issues

johnnychen94:
## Bug fix #354
1. change the denifition of struct `Dense` to make it less general
2. add a `(a::Dense)(x::Number) = a([x])` to prevent broadcasting of scalar value

## Bug fixed: #174
Rewrite the `Flux.activations` function such that the last code sample in [Regularisation](https://github.com/FluxML/Flux.jl/blob/d6a75e1289488945ade1cfa8867717fe1ed19557/docs/src/models/regularisation.md) works as expected

## fix Issue #301
rewrite the tutorial code sample in regularisation.md where `vecnorm` raise MethodError, as is in [Issue #301](https://github.com/FluxML/Flux.jl/issues/301)

## fix Issue #323
Support ≈ for TrackedArray

## add .DS_Store, Project.toml to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ docs/site/
 docs/flux.css
 deps
 Manifest.toml
+Project.toml
+.DS_Store
--- a/docs/src/models/regularisation.md
+++ b/docs/src/models/regularisation.md
@@ -1,13 +1,15 @@
 # Regularisation
 
 Applying regularisation to model parameters is straightforward. We just need to
-apply an appropriate regulariser, such as `vecnorm`, to each model parameter and
+apply an appropriate regulariser, such as Lp-norm, to each model parameter and
 add the result to the overall loss.
 
 For example, say we have a simple regression.
 
 ```julia
+using Flux
 using Flux: crossentropy
+using LinearAlgebra: norm
 m = Dense(10, 5)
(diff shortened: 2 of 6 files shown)


MikeInnes:
Thanks a lot for the patch. It'd be easier to review if the patches for separate issues were split into separate PRs, would you mind doing that?

PR #361, opened 2018-08-20
Add stop() to train loop when callback conditions are met

dhairyagandhi96:
This is with regard to the conversation on /FluxML/Flux.jl#131 for exposing a `stop()` function to break the training loop when a callback condition is met.

@MikeInnes


src/Flux.jl
```diff
@@ -21,7 +21,7 @@ include("optimise/Optimise.jl")
 using .Optimise
 using .Optimise: @epochs
 export SGD, ADAM, ADAMW, AdaMax, Momentum, Nesterov,
-       RMSProp, ADAGrad, ADADelta, AMSGrad, NADAM
+       RMSProp, ADAGrad, ADADelta, AMSGrad, NADAM, stop, StopException
```
MikeInnes: I think these names are a bit too generic for us to export them. Let's just have the API be `Flux.stop()`.

src/optimise/train.jl
```diff
@@ -39,7 +58,16 @@ function train!(loss, data, opt; cb = () -> ())
     l = loss(d...)
     @interrupts back!(l)
     opt()
-    cb() == :stop && break
+    try
```
MikeInnes: How about putting the try/catch around the model as well? Then you could stop from inside the loss function.

src/optimise/train.jl
```diff
@@ -39,7 +58,16 @@ function train!(loss, data, opt; cb = () -> ())
...
+      cb()
```
MikeInnes: Let's check for `:stop` here so that it's backwards compatible; we can emit a warning that it's deprecated.

src/optimise/train.jl
```diff
@@ -39,7 +58,16 @@ function train!(loss, data, opt; cb = () -> ())
...
+    catch ex
+      if ex isa StopException
+        @info "Stop condition met"
```
MikeInnes: I guess this is for debugging, but I don't think we need to include it.

src/Flux.jl
```diff
@@ -23,6 +23,7 @@ using .Optimise: @epochs
 export SGD, ADAM, ADAMW, AdaMax, Momentum, Nesterov,
        RMSProp, ADAGrad, ADADelta, AMSGrad, NADAM
 
+
```
MikeInnes: Best to avoid spurious whitespace changes like this (see also the deleted endlines further down, Juno will take care of those for you).

src/optimise/train.jl
```diff
@@ -36,10 +55,18 @@ function train!(loss, data, opt; cb = () -> ())
...
   @progress for d in data
-    l = loss(d...)
-    @interrupts back!(l)
-    opt()
-    cb() == :stop && break
+    try
+      l = loss(d...)
+      @interrupts back!(l)
+      opt()
+      cb() == :stop && break
```
MikeInnes: do add the depwarn in this case; we'll want to stop support this way of stopping in future.

MikeInnes:
Perfect, thanks a lot @dhairyagandhi96!

PR #362, opened 2018-08-20
1.0 CUDA support

MikeInnes:
Mostly there, but I need to figure out how to support CuArrays' new broadcasting fix, given that we flatten the broadcast wrapper.


MikeInnes:
The CUDNN tests are still pretty flaky, sadly, but this is at least no worse than what we had before.

PR #364, opened 2018-08-21
fix argmax and add test

boathit:
* Fixing the deprecations in the definitions of `argmax`
* Removing the default parameter values to avoid being conflict with the `Base.argmax`(is this okay?) 
* Adding test cases


MikeInnes:
I think it may be confusing to have the two-argument `argmax` behave differently from Base's version. My suggestion is that we rename this back to `onecold`, as it was called originally, so that it doesn't conflict at all. If we do that we should also have `Flux.argmax` give a deprecation warning.

boathit:
Fair enough. How about this change?

MikeInnes:
Tweaked it a bit to avoid overwriting Base's `argmax`, but otherwise looks good -- thanks a lot!

PR #371, opened 2018-08-23
Fix issue #323

johnnychen94:
add support for ≈ operation for types `TrackedArray` and `TrackedReal`

### DIFF
--- a/src/tracker/array.jl
+++ b/src/tracker/array.jl
@@ -1,4 +1,4 @@
-import Base: *, ==, ≈
+import Base: *
 
 import LinearAlgebra
 using Statistics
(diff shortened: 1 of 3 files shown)


MikeInnes:
Thanks! Can you add a quick test for this?

johnnychen94:
Done as you said. Also fixed an existing bug.

```julia
param([2]).^2 == [4.0]
ERROR: MethodError: ==(::TrackedArray{…,Array{Float64,1}}, ::Array{Float64,1}) is ambiguous. Candidates:
==(x::TrackedArray, y) in Main.Flux.Tracker at /Users/jc/.julia/dev/Flux/src/tracker/array.jl:63
==(A::AbstractArray, B::AbstractArray) in Base at abstractarray.jl:1686
Possible fix, define
==(::TrackedArray, ::AbstractArray)
```

MikeInnes:
Looks good, thanks a lot @johnnychen94.

Just as a tip for future PRs – It's better if your title is a bit more descriptive, rather than just referencing another issue number (which I have not memorised :smile:). If you put issue references in the body of the PR they'll also get turned into links, which is handy. "Add support for ≈" would have been an ideal title for this.

johnnychen94:
I'm still in the progress of learning how to use Github as an expert =.= Thanks for your feedback I'll keep this in mind.

Issue #370, opened 2018-08-23
Flux.Data.MNIST.images() error in Julia v1.0


tingcao:
The stack trace as follow:

julia> Flux.Data.MNIST.images()
ERROR: MethodError: no method matching Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},N} where N(::Int64, ::Int64)
Closest candidates are:
  Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},N} where N(::UndefInitializer, ::Int64) where T at boot.jl:408
  Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},N} where N(::UndefInitializer, ::Int64, ::Int64) where T at boot.jl:409
  Array{ColorTypes.Gray{FixedPointNumbers.Normed{UInt8,8}},N} where N(::UndefInitializer, ::Int64, ::Int64, ::Int64) where T at boot.jl:410
  ...
Stacktrace:
 [1] rawimage(::Base.GenericIOBuffer{Array{UInt8,1}}) at /home/ticao/.julia/packages/Flux/npL2p/src/data/mnist.jl:52
 [2] (::getfield(Flux.Data.MNIST, Symbol("##5#6")){Base.GenericIOBuffer{Array{UInt8,1}}})(::Int64) at ./none:0
 [3] iterate at ./generator.jl:47 [inlined]
 [4] collect at ./array.jl:619 [inlined]
 [5] images(::Symbol) at /home/ticao/.julia/packages/Flux/npL2p/src/data/mnist.jl:88
 [6] images() at /home/ticao/.julia/packages/Flux/npL2p/src/data/mnist.jl:85
 [7] top-level scope at none:0

But if I run this statement alone, there is no error: 
src/data/mnist.jl:52   img = Array{Gray}(undef, NCOLS, NROWS)

johnnychen94:
Perhaps your downloaded dataset is incomplete. Just try redownload them.

tingcao:
Thank you for your help! After I redownloaded the data, the problem is solved. @johnnychen94

MikeInnes:
Should be fixed on master

PR #376, opened 2018-08-25
Fix issue #370

johnnychen94:
Fix issue #370 which happens when the downloaded dataset is incomplete.


johnnychen94:
close since this's already been fixed.

PR #379, opened 2018-08-28
New optimisers interface

dhairyagandhi96:
Rebase of #283


src/optimise/Optimise.jl
```diff
@@ -1,23 +1,9 @@
...
-include("interface.jl")
 include("train.jl")
 
-using Flux.Tracker: TrackedArray
-
-Param(x::TrackedArray) = Param(x.data, x.grad)
-# Base.convert(::Type{Param}, x::TrackedArray) = Param(x.data, x.grad)
-
-end
+end
```
MikeInnes: Watch out for spurious whitespace changes here. Might be an editor settings issue.

src/Flux.jl
```diff
@@ -19,8 +19,8 @@ export Tracker, TrackedArray, TrackedVector, TrackedMatrix, param
 include("optimise/Optimise.jl")
 using .Optimise
 using .Optimise: @epochs
-export SGD, ADAM, ADAMW, AdaMax, Momentum, Nesterov,
-       RMSProp, ADAGrad, ADADelta, AMSGrad, NADAM
+export Descent, ADAM, Momentum, Nesterov,
+       RMSProp, update!
```
MikeInnes: I'd prefer not to export this; it's much too generic a name.

src/tracker/array.jl
```diff
@@ -385,6 +385,10 @@ end
 
 using Requires
 
+Base.Broadcast.cat_nested(t::Base.Broadcast.Broadcasted, rest...) = (cat_nested(t.args...)..., cat_nested(rest...)...)
```
MikeInnes: Can you explain what this is for? It seems oddly general, rather than doing anything Flux or TrackedArray-specific, to be here.

test/optimise.jl
```diff
@@ -3,14 +3,18 @@ using Flux.Tracker
 using Test
 @testset "Optimise" begin
   w = randn(10, 10)
-  @testset for Opt in [SGD, Nesterov, Momentum, ADAM, AdaMax, RMSProp, ps -> ADAGrad(ps, 0.1), ADADelta, AMSGrad, NADAM]
```
MikeInnes: I guess I deleted some of these earlier, but we will have to bring back everything Flux currently supports before merging.

src/Flux.jl
dhairyagandhi96: Got it.

src/tracker/array.jl
dhairyagandhi96: These were to help flatten a nested broadcast, taken from broadcast API. Will remove, them as they aren't necessary.

test/optimise.jl
dhairyagandhi96: Got it

src/optimise/optimisers.jl
```diff
@@ -1,130 +1,308 @@
...
+
+mutable struct DescentWeightDecay
+  eta::Real
+  gamma::Real
+end
+
+DescentWeightDecay(η = 1, γ = 0) = DescentWeightDecay(η, γ)
+function update!(o::DescentWeightDecay, x,  Δ)
+  η, γ = o.eta, o.gamma
+  @. x = x - η * (Δ + γ * x)
```
MikeInnes: Seems like something went wrong here; we shouldn't be modifying `x` in the optimiser since the delta will get applied later on. Also, can't this be `Compose(WeightDecay(), Descent())`?

src/optimise/optimisers.jl
```diff
@@ -1,130 +1,308 @@
...
+  eta::Float64
 end
 
-# Ref: https://arxiv.org/abs/1711.05101.pdf
-function descentweightdecay(p::Param, η::Real,  γ::Real)
-  function ()
-    @. p.x = p.x - η * (p.Δ + γ * p.x) 
-    @. p.Δ = 0
-  end
+Descent(η = 0.1) = Descent(η)
```
MikeInnes: Note that this method causes stack overflow.

MikeInnes:
This really is nearly there now, thanks for the hard work @dhairyagandhi96. The one last thing we need is updates for the `train!` function so that it works with the new optimiser setup.

Also, would be great to get docstrings for the new decay optimisers, and tweak anything in the main docs that is out of date. We should document the optimiser setup in more detail, but that can wait for now.

src/optimise/deprecations.jl
```diff
@@ -0,0 +1,134 @@
+using Base: depwarn
+
+function check_decay(opt, decay)
+  if decay == 0.
+    opt = opt
+  else
+    if opt isa ADAMW
+      opt = Optimiser(opt, DescentWeightDecay(1, decay))
```
MikeInnes: Equivalent to `WeightDecay(decay)`?

src/optimise/deprecations.jl
```diff
@@ -0,0 +1,134 @@
...
+  β = (β1, β2)
+  opt = ADAMW(η, β)
+  opt = check_decay(opt, decay)
+  updaterule(opt, ps)
+end
+
+# Train function
+function train!(loss::Function, data, opt; cb = () -> ())
+  depwarn("train!(loss, data, opt; cb) is deprecated; use train!(model, data, loss, opt; cb) instead", :train)
+  train!(opt.ps, loss, data, opt.opt; cb = cb)
```
MikeInnes: Ideally this would be compatible with any `opt`, not just the `updaterule` defined above (and then the tests would not need to avoid `()->()`).

src/optimise/optimisers.jl
```diff
@@ -1,130 +1,295 @@
...
+Base.getindex(c::Optimiser, i::AbstractArray) = Optimiser(c.os[i]...)
+
+function update!(o::Optimiser, x, Δ)
+  for opt in o.os
+    Δ = update!(opt, x, Δ)
   end
+  return Δ
+end
+
+# TODO: decay
```
MikeInnes: I guess this is done now? :)

src/optimise/optimisers.jl
```diff
@@ -1,130 +1,295 @@
...
+  gamma::Float64
+  n::Int64
+end
+
+InvDecay(γ = 0.001) = InvDecay(γ, 0)
+
+function update!(o::InvDecay, x, Δ)
+  γ, n = o.gamma, o.n
+  Δ .*= 1 / (1 + γ * n)
+  o.n += 1
```
MikeInnes: Remember that the same optimiser object will be used in an `update!` with many different parameters. Right now `n` is counting the overall number of parameters in the model, not the number of times an individual parameter was seen. Any parameter-specific state should go in a dictionary.

src/optimise/optimisers.jl
```diff
@@ -1,130 +1,295 @@
...
+end
+
+mutable struct WeightDecay
+  eta::Real
+  wd::Real
+end
+
+WeightDecay(η = 1) = WeightDecay(η, 0)
+function update!(o::WeightDecay, x,  Δ)
+  η, wd = o.eta, o.wd
```
MikeInnes: eta is not used here, I assume it can be removed

src/optimise/optimisers.jl
```diff
@@ -1,130 +1,295 @@
...
+
+mutable struct ExpDecay
+  gamma::Float64
 end
+
+ExpDecay() = ExpDecay(0.001)
+
+function update!(o::ExpDecay, x, Δ)
+  γ = o.gamma
+  @. Δ += γ * x
```
MikeInnes: Right now this looks equivalent to weight decay

src/optimise/train.jl
```diff
@@ -1,7 +1,17 @@
 using Juno
-using Flux.Tracker: back!
+using Flux.Tracker: data, grad, back!
 import Base.depwarn
 
+function update!(opt, xs)
+  for x in xs
+    x, Δ = data(x), grad(x)
+    Δ = update!(opt, x, Δ)
```
MikeInnes: You should pass `x` without unwrapping it here, since that will be used as the key in any dictionaries.

src/optimise/train.jl
```diff
@@ -35,7 +45,7 @@ function stop()
 end
 
 """
-    train!(loss, data, opt)
+    train!(model, loss, data, opt)
```
MikeInnes: I think we should have this be `train!(loss, params, opt, data)` and not try to call `params` for the user.

src/optimise/train.jl
```diff
@@ -44,22 +54,22 @@ Takes a callback as keyword argument `cb`. For example, this will print "trainin
...
-function train!(loss, data, opt; cb = () -> ())
+function train!(ps::Array, loss, data, opt; cb = () -> ())
   cb = runall(cb)
   opt = runall(opt)
   @progress for d in data
     try
       l = loss(d...)
       @interrupts back!(l)
-      opt()
+      foreach(x -> x.data .-= update!(opt, x.data, x.grad), ps)
```
MikeInnes: Wouldn't this be the time to use the `update!` you defined above?

test/optimise.jl
```diff
@@ -3,14 +3,37 @@ using Flux.Tracker
...
+    if opt isa Descent || opt isa ADAGrad
+      opt = Opt(0.1)
+    end
+    if opt isa ADADelta
+      opt = Opt(0.9)
+    end
+    for t = 1: 10^5
+      l = loss(rand(10))
+      back!(l)
+      delta = Optimise.update!(opt, w′.data, w′.grad)
```
MikeInnes: also use the shorter `update!` method?

CarloLucibello:
Can we also choose names more consistently? 

Now we have,
```
export Descent, ADAM, Momentum, Nesterov, RMSProp,
ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM,
ADAMW, InvDecay, ExpDecay, WeightDecay, DescentWeightDecay
```
while I would go for: 
```
export Descent, Adam, Momentum, Nesterov, RmsProp,
AdaGrad, AdaMax, AdaDelta, AmsGrad,  Nadam,
Adamw, InvDecay, ExpDecay, WeightDecay, DescentWeightDecay
```

MikeInnes:
It's a good idea, perhaps something we should do in a separate PR to avoid making this one more complex (we could also save the `Descent` change for that PR, but it doesn't really matter).

MikeInnes:
This turned into a bit of a marathon but it's a great PR at this point. Thanks for all the great work @dhairyagandhi96!

PR #381, opened 2018-08-29
fix julia 1 changes in tutorial

piever:
- `foldl` only accepts a keyword argument to specify the initial value now.
- `vecnorm` is now `norm`


MikeInnes:
Thanks!

PR #384, opened 2018-09-01
added sigmoid function

tk3369:
I wasn't able to complete the tutorial in REPL since `σ` wasn't defined.  This PR updates the code sample with the sigmoid function.

### DIFF
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -110,6 +110,7 @@ W2 = param(rand(2, 3))
 b2 = param(rand(2))
 layer2(x) = W2 * x .+ b2
 
+σ(z) = @. 1 / (1 + exp(-z))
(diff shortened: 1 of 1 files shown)


johnnychen94:
check [NNLib](https://github.com/FluxML/NNlib.jl/blob/91fe82da545f5faa07be48f8afd9b97c15929c71/src/activation.jl#L7)

MikeInnes:
This should definitely be fixed so that it can be run without changes. Adding `using Flux` to this would solve it, since that exports `σ`.

tk3369:
Closing this PR. Just adding `using Flux` sounds like a great idea.

MikeInnes:
Someone still has to add that though :) Are you up for making another PR, or updating this one?

PR #390, opened 2018-09-07
Proposal: Initialise all weights as Float32

MikeInnes:
Any thoughts? I think this will be a more appropriate default for most users.

A fairly significant downside is that Float64 inputs, while having the right semantics, will hit very slow generic matmuls in base. I think that "convert your inputs to float32" is reasonable advice to give to the performance-conscious though, and we can catch a lot of cases via dispatch.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -114,3 +114,11 @@ end
 function Base.show(io::IO, l::Diagonal)
   print(io, "Diagonal(", length(l.α), ")")
 end
+
+# Try to avoid hitting generic matmul in some simple cases
+# Base's matmul is so slow that it's worth the extra conversion to hit BLAS
+(a::Dense{<:Any,W})(x::AbstractArray{T}) where {T <: Union{Float32,Float64}, W <: AbstractArray{T}} =
+  invoke(a, Tuple{AbstractArray}, x)
+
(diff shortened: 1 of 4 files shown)


ityonemo:
Somewhat opposed to internally converting to Float32.

Eventually GPUs will start going down the float16 route, so it might be unhelpful to obligatorially require a particular course of action (there may be cases where one would want to do a mixed-mode FP16/FP32 thing).  Maybe also someone might want to try complex-valued weights, etc.  Also, I'm using Flux to model deep learning algos on *very non-standard datatypes*...  As it stands Flux is literally the only framework where this is "basically painless".

I'm working on a PR which will give explicit model conversions:

https://github.com/FluxML/Flux.jl/issues/399

On top of this, though, one could have a flag where the user in code could create a globally enforced type, which will do promotion (or demotion?) to that type everywhere.

ityonemo:
ok just wanted to make sure that "default" doesn't mean "automatically convert when running/training the model".

rs1909:
Is there a way to go back to Float64? Or at least make this optional? I thought that the strength of Julia is that everything is generic and should work irrespective of the underlying type., i.e. Float32 vs Float64 or at least I should be able to specify the type.

I've tried: the proposal in #399 does not work.

MikeInnes:
Good timing: #572.

rs1909:
I can confirm that #572 does work (for me). I have used `(v1.1) pkg> add Flux#precision
`

PR #389, opened 2018-09-06
Fixing batchindex/batch/batchseq

Proyag:
Fixes #387.

* Changed [`indices` to `axes`](https://github.com/JuliaLang/julia/pull/25057) in `batchindex`.
* `batchseq` was broken because `batch` wasn't handling `Array{Int, 1}` input correctly.

PR #396, opened 2018-09-13
indices was renamed to axes in Julia 0.7

iht:
Some functions related to batches do not work on Julia 1.0 because the `indices` function has been renamed to `axes` (see https://github.com/JuliaLang/julia/pull/25057).

This is a minimal change so `batchindex` and `batchseq` can be used again in 1.0.


iht:
I just realized that #389 fixes this too. I will close this PR now.

PR #404, opened 2018-09-19
add inv/ldivide/rdivide + test

ornithos:
Added array functions:

* `inv`
* `\` (ldivide)
* `/` (rdivide)

I've added tests with a few extra nonlinearities thrown in as I've been caught out before calculating gradients which were correct only for *linear* input. A few things to be aware of:

* Not familiar with PRs, so apologies if I've done anything wrong.
* ldivide/rdivide are more numerically stable than their (mathematically) equivalent operations `inv(A)*B`, `A*inv(B)` (respectively), but for the time being, their *backward* pass is implemented using the inverses directly as I've found it significantly faster than multiple `lu` etc. solves using the `\`,`/` operators.
* My implementation currently requires the relevant arguments to be invertible even though `LinearAlgebra.\` and `LinearAlgebra./` do not (returning the least squares solution in this case). A work around is to require users to input the least squares projection (e.g. `inv(A'*A)*A'*B`) directly rather than using this MATLAB-style shortcut. Implementing the derivative of this is a little messy, and I think it's important for users to understand the overhead in what they're implicitly asking the AD engine to do.
* Thanks to Mike for help overloading the `/`,`\` operators.

### DIFF
--- a/src/tracker/array.jl
+++ b/src/tracker/array.jl
@@ -1,6 +1,8 @@
 import Base: *
 
 import LinearAlgebra
+import LinearAlgebra: inv, \, /
+
 using Statistics
 using LinearAlgebra: Transpose, Adjoint, diagm, diag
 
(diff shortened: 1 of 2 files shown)


MikeInnes:
Looks great to me. I think if it's useful to you even with those restrictions, then that's clearly an improvement over no gradient at all. Others can always improve it later if needed.

PR #408, opened 2018-09-21
Proposal: improve \ and / tracking

willtebbutt:
The current implementations of the ldiv and rdiv operations use an explicit inverse, which is generally understood to be sub-optimal numerically. The proposed implementations avoid the use of explicit inverses, and should produce reductions in both time- and memory-complexity.


MikeInnes:
Ref #404. Looks reasonable to me but pinging @ornithos for any comments.

ornithos:
Not sure where this all comes out. While I don't want to lose my performance gains for my own purposes, I think doing the _expected_ thing is really the right thing to do, so I think (with a bit of thought about second order) I would be happy with this.

ornithos:
One more thing, any chance you can stick to the convention of $\Delta$ being the function argument to the gradient function? Just makes reading down the list of definitions easier.

willtebbutt:
Thanks for the response @ornithos . A couple of things:

> The reason I gave in #404 for using the less numerically stable versions is that it's usually faster not to.

Could you explain to me why this should be the case in smaller models / provide a concrete example? I'm not sure exactly what experiments you're running for your timings, but I find it surprising that an `inv`-based implementation of the sensitivities of `ldiv` / `rdiv` is faster than an `ldiv` / `rdiv`-based implementation.

> My suspicion is that users will more commonly find use for ldivide/rdivide in smaller models and dynamical systems where (e.g.) inference routines and unusual parameterisations require it.

I'm personally primarily interested in high-dimensional Gaussians (I do a lot of Gaussian process - related work), so I will have to disagree on the ldivide / rdivide operations being used primarily for small models.

Regarding second-order gradients, I'm not sure what the correct course of action is here. I agree that it should be possible too support them, but I would expect the particular code that you propose to recurse infinitely, but I've not spend much time working with Flux's AD so I don't really know. @MikeInnes what's the state of things here? Do we have any testing set up for nested AD / higher-order AD? What is Flux's mechanism for dealing with perturbation confusion, and how would you expect it to play with this example?

Regarding the Delta-convention -- sure. I was trying to stick close to what I've generally see in the literature  (the reverse-mode sensitivity of a variable is usually indicated by adding a bar the variable name, whereas a forwards-mode is usually indicated by a dot) but I'm happy to go with a Delta if that's what Flux is going with.

willtebbutt:
> but I find it surprising that an inv-based implementation of the sensitivities of ldiv / rdiv is faster than an ldiv / rdiv-based implementation.

Apologies for this comment, I should have read #404 properly before commenting. The point you make in #404 is that you're better off avoiding multiple LU factorisations. I totally agree, but arbitrary dense abstract matrices aren't the only thing to consider here. I'm particularly interested in triangular matrices, for which `/` and `\` are of course very efficient, and explicitly computing the inverse would add overhead.

I think that the "right" solution here is to have better AD support for factorisations (I'm personally working on the `Cholesky` at the minute, but the other usual suspects are of course required), so that you can explicitly call the factorisation you want in your eg. Kalman filter and have sensible things happen when you backprop.

ornithos:
> I think that the "right" solution here is to have better AD support for factorisations

Really nice idea.

> I'm personally primarily interested in high-dimensional Gaussians

I hadn't thought about GPs, but yes this totally makes sense. 

> but I find it surprising that an inv-based implementation of the sensitivities of ldiv / rdiv is faster than an ldiv / rdiv-based implementation. 

I agree that this is surprising.  I was initially using TimerOutputs on a larger model, but the above table was just from running:

```
d = 10
X, Y = randn(d,d), randn(d,d)

@benchmark inv($X)*$Y
@benchmark $X\$Y
```

for various values d.

> Regarding the Delta-convention -- sure. I was trying to stick close to what I've generally see in the literature

I'm sure you know this stuff better than me. I'm happy with any convention as long as it's consistent. This is one for @MikeInnes!

> Regarding second-order gradients, I'm not sure what the correct course of action is here. I agree that it should be possible too support them, but I would expect the particular code that you propose to recurse infinitely

Yeah, I would too :). Maybe just stick a `nobacksies` wrapper around it to avoid the issue. I'm also not sure what the status is with second order derivatives.

I'm really happy with this  👍. The efficiency concerns are my own business, I'm happy that there is sufficient general purpose cause to throttle this a little.

MikeInnes:
FWIW I'd be fine with starting to move towards notation like `Ȳ`. The delta thing is basically unique to Flux right now and doesn't make for great notation in papers and such.

I would also generally lean towards doing the correct/general thing in the stdlib. Bit of a shame to make Alex use a custom implementation for this for now, but hopefully we can figure out how to express that extra structure using types.

Alex is right that this will currently break with nested AD, because you're dropping the gradient of `Y`. Two ways around this: recompute `Y` with tracked variables in the backpropagator or add `nobacksies` to disable nested AD entirely. The redundant compute is annoying but if you want nested AD to work I'll add a fix to avoid that.

To answer Will's specific questions, we currently test nested AD in a simple way but not numerically (which is probably not necessary). Flux's AD doesn't suffer perturbation confusion due to the way it's set up, and I think we have tests for this also; in either case you should never have to worry about this inside a gradient def.

willtebbutt:
> recompute Y with tracked variables in the backpropagator

What would this look like?

> Flux's AD doesn't suffer perturbation confusion due to the way it's set up

Is this documented / discussed somewhere?

MikeInnes:
I was thinking of something like
```julia
Y1 = data(A)/data(B)
Y1, function back(dy)
  Y2 = iA/B
  ...
```
That avoids the infinite recursion but it's not that nice to do the calculation twice. Instead you can probably just construct the tracked array `Y2` from `Y1` explicitly.

We haven't explicitly discussed perturbation confusion but I can answer any questions. It's not something that comes up nearly as often in reverse-mode systems; for example AutoGrad.jl does not suffer this problem either.

willtebbutt:
Ah I see. Yeah, the overhead of computing this twice would be pretty unacceptable, so the manual construction of the tracked array seems preferable. Do you have a standard way to do this, or should I just hack something together?

MikeInnes:
It should be something really simple like `track(Call(/, A, B), Y1)`, but let me know if you hit trouble.

ornithos:
By the way, a bug that appears to be introduced via the way #404 defined the overloading of '\', '/' is also fixed by this PR. MWE:

```julia
L, v = randn(3,3), param(randn(3))
z = L \ v
loss = -0.5*sum(z.^2)
Tracker.back!(loss)
# ERROR: DomainError with log...
```

The stacktrace is:
```
-0.057266858490486204 will only return a complex result if called with a complex argument. Try -0.057266858490486204(Complex(x)).
Stacktrace:
 [1] throw_complex_domainerror(::Float64, ::Symbol) at ./math.jl:31
 [2] log(::Float64) at ./special/log.jl:285
 [3] _forward at /Users/bob/.julia/packages/Flux/UHjNa/src/tracker/scalar.jl:55 [inlined]
 [4] #track#1 at /Users/bob/.julia/packages/Flux/UHjNa/src/tracker/Tracker.jl:50 [inlined]
 [5] track at /Users/bob/.julia/packages/Flux/UHjNa/src/tracker/Tracker.jl:50 [inlined]
 [6] log at /Users/bob/.julia/packages/Flux/UHjNa/src/tracker/scalar.jl:57 [inlined]
 [7] #218 at /Users/bob/.julia/packages/Flux/UHjNa/src/tracker/scalar.jl:66 [inlined]
 [8] back_(::Flux.Tracker.Call{getfield(Flux.Tracker, Symbol("##218#219")){Flux.Tracker.TrackedReal{Float64},Int64},Tuple{Flux.Tracker.Tracked{Float64},Nothing}}, ::Float64) at /Users/bob/.julia/packages/Flux/UHjNa/src/tracker/back.jl:23
 [9] back(::Flux.Tracker.Tracked{Float64}, ::Float64) at /Users/bob/.julia/packages/Flux/UHjNa/src/tracker/back.jl:43
 [10] foreach at ./abstractarray.jl:1836 [inlined]
...
```
(Worth mentioning that if one uses `z .* z` rather than `z .^ 2` there is no error.) I'm really unsure what's going on here, but was playing around to see if the error persisted with Will's code for this, and discovered that the error seems to go away by making the definitions:

```diff
+ \(A::TrackedMatrix, B::TrackedVecOrMat) = Tracker.track(\, A, B)
- A::TrackedArray     \ B::TrackedArray     = Tracker.track(\, A, B)
```

I presume I was overly generous with choosing types as Arrays rather than Matrices/Vectors. But I must confess I have no idea why this issue exists, or why this fixes it. Regardless, nice job on fixing it :).

willtebbutt:
I'm closing this as it's basically redundant at this point. Please feel free to re-open if for some reason it becomes relevant again.

PR #409, opened 2018-09-21
Correct Custom Gradients docs

harryscholes:
* Fixed a type signature that was incorrect
* Also, replaced `data(a)` with `a.data`. Don't know if the syntax has changed (recently). This may also need to be corrected in line 121.

MWE:

```julia
using Flux
using Flux.Tracker
using Flux.Tracker: forward, TrackedReal, track, @grad

minus(a, b) = a - b
minus(a::TrackedReal, b::TrackedReal) = Tracker.track(minus, a, b)
@grad function minus(a, b)
    return minus(a.data, b.data), Δ -> (Δ, -Δ)
end

a, b = param(2), param(4)
c = minus(a, b)  # -2.0 (tracked)
Tracker.back!(c)

Tracker.grad(a)  # 1.00
Tracker.grad(b)  # -1.00
```

### DIFF
--- a/docs/src/internals/tracker.md
+++ b/docs/src/internals/tracker.md
@@ -102,14 +102,14 @@ Firstly, we must tell the tracker system to stop when it sees a call to `minus`,
 ```julia
 using Flux.Tracker: TrackedReal, track, @grad
 
-minus(a::TrackedArray, b::TrackedArray) = Tracker.track(minus, a, b)
+minus(a::TrackedReal, b::TrackedReal) = Tracker.track(minus, a, b)
 ```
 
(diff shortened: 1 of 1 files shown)


harryscholes:
Ahh I realise `data(x)` isn't exported from Flux, so that part doesn't need to be changed. I think the type signature does need to be corrected though

MikeInnes:
The code works fine if you use `param([1,2,3])` or similar. Maybe it'd be good to show a quick usage example to clarify.

harryscholes:
Cool. I've rolled back my previous changes and added a wee example.

MikeInnes:
Cool, thanks!

PR #418, opened 2018-10-02
Add FashionMNIST

c-p-murphy:
Thought this might be a useful addition to the datasets that are included.

### DIFF
--- a/src/data/Data.jl
+++ b/src/data/Data.jl
@@ -13,6 +13,9 @@ end
 include("mnist.jl")
 export MNIST
 
+include("fashion-mnist.jl")
+export FashionMNIST
+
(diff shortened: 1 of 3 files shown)


MikeInnes:
Cool! It looks like most of the code is shared with MNIST itself, so perhaps it'd be good to reuse some of the utilities there rather than duplicating?

c-p-murphy:
Yeah, that definitely makes sense. I think it reuses everything that can be now, but if I missed anything just let me know.

MikeInnes:
Perfect. Thanks a lot!

c-p-murphy:
Awesome, happy to help.

PR #423, opened 2018-10-05
new cross convolution layer

ayush1999:
This works on top of changes in https://github.com/FluxML/NNlib.jl/pull/71. Implemented a `CrossConv` layer, which calls `crossconv` and `crossconv!` functions from NNlib.


ayush1999:
Will add tests for the layer upon approval.

MikeInnes:
Looks good to me. Can you add some quick tests that it runs through, and perhaps also a gradient check for `crosscor`.

src/tracker/array.jl
```diff
@@ -348,6 +348,16 @@ conv(x::TrackedArray,  w::AbstractArray; kw...) = track(conv, x, w; kw...)
...
 
+crosscor(x::TrackedArray,  w::TrackedArray;  kw...) = track(crosscor, x, w; kw...)
+crosscor(x::AbstractArray, w::TrackedArray;  kw...) = track(crosscor, x, w; kw...)
+crosscor(x::TrackedArray,  w::AbstractArray; kw...) = track(crosscor, x, w; kw...)
+
+@grad crosscor(x, w; kw...) =
+  crosscor(data(x), data(w); kw...),
+    Δ -> nobacksies(:crosscor,
+      (NNlib.∇conv_data(data.((Δ, x, w))...; kw...),
+       NNlib.∇conv_filter(data.((Δ, x, w))...; kw...)))
```
MikeInnes: Needs the gradient check. I guess this won't be right without `flipkernel`, no?

ayush1999: Ah, missed that. Will add it.

ayush1999: @MikeInnes We'd have to implement `flipkernel` argument for the `∇conv_data` family of functions, right?

ayush1999:
@MikeInnes Tests pass locally.

MikeInnes:
Sorry for dropping this for a while. Can you do a quick rebase? Will probably need the gradients part of the PR to move to Tracker.jl.

MikeInnes:
@staticfloat did `crosscor` dissappear as part of the NNlib overhaul?

staticfloat:
Ah, yes, it did.  It has been subsumed into `conv()`.  So the `conv()` call now takes a `DenseConvDims` object, which you set `flipkernel=true` within in order to do cross-correlation.

MikeInnes:
Ok cool. In that case @ayush1999 let's just define `crosscor` in Flux. We won't need the Tracker PR then. You'll need the updated branch of Tracker that's compatible with the new NNlib though.

src/layers/conv.jl
```diff
@@ -171,6 +171,59 @@ end
...
+
+Takes the keyword arguments `pad`, `stride` and `dilation`.
+"""
+struct CrossCor{N,F,A,V}
+σ::F
+weight::A
+bias::V
+stride::NTuple{N,Int}
+pad::NTuple{N,Int}
+dilation::NTuple{N,Int}
```
MikeInnes: Formatting is off here

MikeInnes:
Can we keep the `crosscor` function – it can just take a `DenseConvDims` and set `flipkernel=true`. Then the `CrossCor` layer can call that.

Also, can we get a GPU test for this layer as well.

staticfloat:
> Can we keep the `crosscor` function – it can just take a `DenseConvDims` and set `flipkernel=true`

@ayush1999 you can set `flipkernel` easily by using the [kwargs constructor of `DenseConvDims`](https://github.com/FluxML/NNlib.jl/blob/ed4fe9a01a3ebfa11bf728906e85632f29dde164/src/dim_helpers/DenseConvDims.jl#L55-L61):

```julia
new_cdims = DenseConvDims(cdims; flipkernel=true)
```

ayush1999:
If I'm not wrong, we'd need to tag a new release for NNlib so that the tests can pass.

staticfloat:
We're going to need to wait for #714 or #718 to be merged first, yes.

MikeInnes:
Those PRs are in, so we should be able to get this in now, just has one more conflict.

Note that you can also put branches of packages in the manifest to get tests to pass. Shouldn't be necessary now though.

ayush-1506:
I've added all these changes in a new PR (#762). (I lost access to the github account for this PR). Can you please close this and review the new PR?

Issue #424, opened 2018-10-07
Support for Kaiming Initialization


Aniket1998:
Flux currently does not have support for Kaiming Normal and Kaiming Uniform Initialization

Aniket1998:
Working on adding said initializers alongwith tests.

PR #425, opened 2018-10-07
Add Kaiming initializer

Aniket1998:
#424 
This PR adds support for kaiming normal and kaiming uniform initialization


MikeInnes:
This looks good to me. It'd be good to have an quick look over from someone who is more familiar with this though.

Paethon:
Checked it against the implementation in https://pytorch.org/docs/stable/_modules/torch/nn/init.html#kaiming_uniform_

Looks good to me.
One possible improvement could be to use `Uniform` and `Normal` from `Distributions.jl` instead of generating the samples from the correct distribution "by hand"?

Paethon:
Sorry, I think this was a misunderstanding. 

You are in essence creating samples form a specific uniform distribution with a lower and upper bound and samples from a normal distribution with a specific standard deviation. So in my opinion it makes more sense to explicitly calculate these bounds and standard deviation and directly generate samples using `Distributions.jl`

I was just thinking about something like this.

```julia
function kaiming_uniform(dims...; gain=sqrt(2))
    fan_in = ndims(w) <= 2 ? size(w)[end] : div(length(w), size(w)[end])
    bound = sqrt(3.0)*gain/sqrt(fan)
    return rand(Uniform(-bound, bound), dims)
end

function kaiming_normal(dims...; gain=sqrt(2))
    fan_in = ndims(w) <= 2 ? size(w)[end] : div(length(w), size(w)[end])
    std = gain/sqrt(fan)
    return rand(Normal(0.0, std), dims)
end
```

You will also have to add `Distributions` to the `REQUIRE` file

Aniket1998:
@Paethon thanks for the pointers! I'm baffled myself. Could this have something to do with Distributions.jl not being 1.0 compatible or am I making a terrible mistake somewhere?

Paethon:
You have to splat the dims. Seems rand does not accept tuples for the dimensions. e.g.

```julia
return rand(Normal(0.0, std), dims...)
```

Also corrected the examples above

Paethon:
The code above was just to show what I meant. That was not actually working code (sorry if you thought it was). This should be a correctly working implementation. I am also going to replace the upper versions to prevent confusion.

```julia
function kaiming_uniform(dims...; gain=sqrt(2))
    fan_in = length(dims) <= 2 ? dims[end] : div(*(dims...), dims[end])
    bound = sqrt(3.0)*gain/sqrt(fan_in)
    return rand(Uniform(-bound, bound), dims...)
end

function kaiming_normal(dims...; gain=sqrt(2))
    fan_in = length(dims) <= 2 ? dims[end] : div(*(dims...), dims[end])
    std = gain/sqrt(fan_in)
    return rand(Normal(0.0, std), dims...)
end
```

Drvi:
Hi! Just thinking -- wouldn't it be simpler to use ```2bound .* (rand(dims...) .- 0.5)``` instead of `Uniform` and ```std .* randn(dims...)``` instead of `Normal`? You could loose the Distributions dependency then.

Even Distributions.jl are doing it the same way:
https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/normal.jl#L84
https://github.com/JuliaStats/Distributions.jl/blob/master/src/univariate/continuous/uniform.jl#L109

Paethon:
That is what was there previously and IMHO it is clearer what is happening using the Distributions.jl package (that's what it is there for after all)

Not sure whether the dependency on Distributions.jl is problematic

Issue #416, opened 2018-10-02
copy(::TrackedArray) doesn't work


tpapp:
MWE:
```julia
using Flux.Tracker
x = ones(5)
tx = TrackedArray(x)
copy(tx)
```
gives the error
```julia
ERROR: MethodError: no method matching Float64(::Flux.Tracker.TrackedReal{Float64})
Closest candidates are:
  Float64(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:185
  Float64(::T<:Number) where T<:Number at boot.jl:722
  Float64(::Int8) at float.jl:60
  ...
Stacktrace:
 [1] convert(::Type{Float64}, ::Flux.Tracker.TrackedReal{Float64}) at ./number.jl:7
 [2] setindex!(::Array{Float64,1}, ::Flux.Tracker.TrackedReal{Float64}, ::Int64) at ./array.jl:769
 [3] setindex! at ./multidimensional.jl:414 [inlined]
 [4] copyto!(::Array{Float64,1}, ::TrackedArray{…,Array{Float64,1}}) at ./multidimensional.jl:822
 [5] copymutable(::TrackedArray{…,Array{Float64,1}}) at ./abstractarray.jl:823
 [6] copy(::TrackedArray{…,Array{Float64,1}}) at ./abstractarray.jl:773
 [7] top-level scope at none:0
```
on
```julia
(v1.1) pkg> status
  [587475ba] Flux v0.6.7+ #master (https://github.com/FluxML/Flux.jl.git)
julia> VERSION
v"1.1.0-DEV.361"
```

MikeInnes:
We just need to add a method to catch this. Since tracked arrays are immutable anyway it can just be a no-op.

tpapp:
What should `similar` do? Return a `Array{Flux.Tracker.TrackedReal{Float64}}`?

MikeInnes:
Similar should always return the equivalent un-tracked object; there's no reason to include tracking for that since there are no gradients anyway.

tpapp:
I am not sure, since the [Interfaces documentation](https://docs.julialang.org/en/stable/manual/interfaces/) says that `similar` should

> Return a mutable array with the same shape and element type

but in

```julia
x = TrackedArray(ones(5))
y = similar(x)
eltype(x) == eltype(y) # => false
```
which is causing the above code to fail.

johnnychen94:
[...]

PR #427, opened 2018-10-08
Support copy(::TrackedArray)

johnnychen94:
1. fix issue https://github.com/FluxML/Flux.jl/issues/416
2. change test code to pass the test: some broken tests are not broken now...

### DIFF
--- a/src/tracker/array.jl
+++ b/src/tracker/array.jl
@@ -43,6 +43,8 @@ end
 
 Base.print_array(io::IO, x::TrackedArray) = Base.print_array(io, data(x))
 
+Base.copy(x::TrackedArray) = copy(data(x))
+
 Base.setindex!(xs::TrackedArray, v, i...) =
   error("Can't differentiate `setindex!`")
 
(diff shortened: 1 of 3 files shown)


MikeInnes:
Thanks, but can you just add the `copy` method directly to fix this. Otherwise this will cause `copy` to silently drop gradients.

I think the `@test_throws` is right; we're testing that we get an ambiguity as opposed to a dropped gradient etc.

johnnychen94:
You mean codes like these right?
```julia
Base.copy(x::TrackedReal) = copy(data(x))
Base.copy(x::TrackedArray) = copy(data(x))
```

I use this `convert` code because the behavior of  `convert(::Type{T}, x::TrackedReal{S})` is clear -- the tracked gradients can be dropped.

I'm not clear about `copy` though, should we keep the gradients in `copy(::TrackedArray)`?

MikeInnes:
You can just write `Base.copy(x::TrackedReal) = x`, since we treat tracked objects as immutable anyway.

The problem with dropping gradients is that it's implicit, and (for example) in this case would have led `copy` to silently drop gradients as well; this kind of thing can easily lead to confusing bugs in code.

johnnychen94:
Alright, done as you said. 

Thanks for your patience to explain this

MikeInnes:
No problem, thank you for putting it together!

Issue #435, opened 2018-10-13
Error with Flux.crossentropy


xiaodaigh:
```
Flux.crossentropy([false, true, false, false], [1.0, 0.0, 0.0, 0.0] )
```
gives `NaN` or `Inf`

It should give `0`

Paethon:
No it should not.

A crossentropy-loss of 0.0 would mean you have a perfect fit between the expected output `[false, true, false, false]` and the observed output `[1.0, 0.0, 0.0, 0.0]`, which is clearly not the case. 

In this case `Inf` could be considered the correct output (`Nan` is obviously bad). Usually frameworks clamp values of 0.0 to a very small epsilon to circumvent numerical instability (which the Flux-implementation does not), so maybe this should be changed ...

boathit:
For the purpose of numerical stability, `crossentropy` should be discouraged and `logitcrossentropy` or `binarylogitcrossentropy` is strongly recommended.

Paethon:
Is there a specific reason `crossentropy` is not implemented in a numerically stable way? Because I would think if it is in the framework it should be implemented "properly", regardless whether its use is discouraged or not.

MikeInnes:
I agree and there's no specific reason I know of, so if you have any stability improvements they'd be very welcome.

Paethon:
Ahh, by the way, the first argument is the predicted and the second is the observed result. So this should probably be 

```julia
Flux.crossentropy([1.0, 0.0, 0.0, 0.0], [false, true, false, false])
```

opus111:
Hi Mike.  I'm new here, but a big fan of Julia and Flux.  Thank you for writing it :-D

I ran into the same stability issues with crossentropy() and wrote my own.  Here it is.  Use if you see fit

    const ε = eps(1.0)
    crossentropy(x) = y > 0.5 ? -(log(ŷ+ε)) : -(log((1.0-ŷ)+ε))

PR #437, opened 2018-10-15
Make crossentropy and binarycrossentropy numerically stable

Paethon:
Follows the implementation in Keras

The slightly complicated way of getting the epsilon is so that yhat does not have to contain Float

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -5,7 +5,7 @@ using NNlib: logsoftmax, logσ
 mse(ŷ, y) = sum((ŷ .- y).^2)/length(y)
 
 function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat;
-                      weight = 1, ϵ = eps(log(one(eltype(ŷ)))))
+                      weight = 1, ϵ = eps(one(log(one(eltype(ŷ))))))
     if typeof(ϵ) == eltype(ŷ)
         clamp!(ŷ, 0 + ϵ, 1 - ϵ)
     else
@@ -32,7 +32,7 @@ Return `-y*log(ŷ + ϵ) - (1-y)*log(1-ŷ + ϵ)`. The ϵ term provides numerica
     0.352317
     0.86167
 """
-function binarycrossentropy(ŷ, y; ϵ = eps(log(one(eltype(ŷ)))))
+function binarycrossentropy(ŷ, y; ϵ = eps(one(log(one(eltype(ŷ))))))
     ŷ= clamp(ŷ, 0 + ϵ, 1 - ϵ)
 
     -y*log(ŷ) - (1 - y)*log(1 - ŷ)
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -4,8 +4,15 @@ using NNlib: logsoftmax, logσ
 
 mse(ŷ, y) = sum((ŷ .- y).^2)/length(y)
 
-function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight = 1)
-  -sum(y .* log.(ŷ) .* weight) / size(y, 2)
+function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat;
+                      weight = 1, ϵ = eps(log(one(eltype(ŷ)))))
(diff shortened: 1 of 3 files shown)


Paethon:
Fixes #435

src/layers/stateless.jl
```diff
@@ -4,8 +4,15 @@ using NNlib: logsoftmax, logσ
 
 mse(ŷ, y) = sum((ŷ .- y).^2)/length(y)
 
-function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight = 1)
-  -sum(y .* log.(ŷ) .* weight) / size(y, 2)
+function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat;
+                      weight = 1, ϵ = eps(one(log(one(eltype(ŷ))))))
```
MikeInnes: Perhaps just `eps(float(eltype(y)))`?

src/layers/stateless.jl
```diff
@@ -4,8 +4,15 @@ using NNlib: logsoftmax, logσ
...
+    if typeof(ϵ) == eltype(ŷ)
+        clamp!(ŷ, 0 + ϵ, 1 - ϵ)
```
MikeInnes: Better to always to the out-of-place version here, which you can fuse with the rest of the computation anyway. Otherwise it could be surprising to find the input modified.

src/layers/stateless.jl
Paethon: OK, `log` uses `float()` anyway, so should be ok :)

Paethon:
OK, I implemented the changes (thanks for the help)

I also had to add versions of `float` and `eps` for `TrackedReal`. Hope I did not mess that up :)

boathit:
@MikeInnes I think adding such an epsilon to mitigate the numerical instability is deceptive and may lead to wrong loss. 

Considering a model whose output of the linear layer is `o = [-600, 102.5]` and the ground truth `y = [0.8, 0.2]`, if we use such an implementation to compute the loss what we will get is `28.83`; however, the true loss is `562.0`

```julia
function crossentropy_ϵ(ŷ::AbstractVecOrMat, y::AbstractVecOrMat;
                      weight = 1, ϵ = eps(float(eltype(ŷ))))
     -sum(y .* log.(clamp.(ŷ, 0 + ϵ, 1 - ϵ)) .* weight) / size(y, 2)
end

o = [-600, 102.5]
y = [0.8, 0.2]

crossentropy_ϵ(softmax(o), y) # 28.83

using Flux: logitcrossentropy
# true loss
logitcrossentropy(o, y) # 562.0
```

Even worse, for  all small `o[1]`  the  `crossentropy_ϵ()` will always return the same loss and this cannot correctly penalize the neuron whose output is incorrect 

```julia
o1 = [-700, 102.5]
o2 = [-1000, 102.5]
crossentropy_ϵ(softmax(o1), y) # 28.83
crossentropy_ϵ(softmax(o2), y) # 28.83

## true loss
logitcrossentropy(o1, y) # 642.0
logitcrossentropy(o2, y) # 882.0
```

A more reasonable way to define `crossentropy`, I think, is to change the interface of `crossentropy` to use the log-probabilities, i.e., `crossentropy(logŷ, y)` rather than `crossentropy(ŷ, y)`. In other words, we explicitly force the users to use `logsoftmax` instead of `softmax` as their output layer. Actually this is exactly what PyTorch does, see [KL Divergence Loss](https://pytorch.org/docs/stable/nn.html#kldivloss) and [NLL](https://pytorch.org/docs/stable/nn.html#nllloss).

MikeInnes:
@boathit it’s a good point to raise, but does that kind of thing actually come up? AFAIK cross entropy is defined on probablility distributions, in which case we can reasonably leave this undefined for numbers outside 0-1. 

I do think we should make it easier to be numerically stable here, and I think #85 will be a great way to do that, since we can just rewrite `crossentropy(softmax(x),y)` however we want to then.

boathit:
@MikeInnes In any case, it is desirable to compute the cross entropy through the logits. So I am also kind of into #85.  But to unify the interface, do we also need a lazy sigmoid function?

ToucheSir:
All of these loss functions now use an appropriate epsilon (mostly via `epseltype`), so this can be closed.

PR #441, opened 2018-10-17
Removes initn initialization

Paethon:
Is replaced with `glorot_uniform` for `Conv` following Keras


Paethon:
Once #425 is merged we could also switch to `kaiming_uniform` (which is what PyTorch is doing) under the assumption that most people will likely use `relu` as their non-linearity.

PR #445, opened 2018-10-20
Change check for CUDNN availability.

maleadt:
`libcudnn` is an implementation detail; check for the module.

### DIFF
--- a/src/cuda/cuda.jl
+++ b/src/cuda/cuda.jl
@@ -2,6 +2,6 @@ module CUDA
 
 using ..CuArrays
 
-CuArrays.libcudnn != nothing && include("cudnn.jl")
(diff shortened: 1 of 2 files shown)


maleadt:
Ah, I failed to notice that `cudnn.jl` contains quite some manual API calls to `libcudnn` so it did make sense to check for that. Although those API calls should probably move to CuArrays.CUDNN?

MikeInnes:
I'd be fine with moving that stuff to CuArrays in principle; should just be a copy-paste really.

MikeInnes:
I think this is outdated since #473.

MikeInnes:
I think we'll want to move more of this stuff to CuArrays  anyway (so that we can share it between Flux and Knet), so I'll make that happen in the near future.

PR #448, opened 2018-10-21
Match paper for Adam implementation and make epsilon use more consistent

JobJob:
@CarloLucibello @tejank10 according to the papers, I think the epsilons are supposed to be outside the square roots in adam ~and nadam~, was there a reason you wanted to add the epsilon inside the sqrt in https://github.com/FluxML/Flux.jl/pull/224? Maybe to avoid sqrts of subnormal floating point values?

Adam: https://arxiv.org/abs/1412.6980v9
adam_update

~Nadam~ADAMW: https://arxiv.org/abs/1711.05101
adamw_update

Not sure how much of an issue this is in practice, but it does change the values if `vt` is small. Also with the epsilon inside the sqrt it's similar to using a default of 1e-4.

I think we should probably match the paper for Adam ~/Nadam~. It seems unlikely the moving averages of the variance of the gradients are going to be subnormal, but not 0, very often, but maybe there's something else I'm missing.

As a data point, from a brief squiz at the pytorch code [here](https://pytorch.org/docs/stable/_modules/torch/optim/adam.html) they keep epsilon outside the sqrt: `denom = exp_avg_sq.sqrt().add_(group['eps'])`

If we don't want this then it still might be a good idea to change the default epsilon to 1e-16.

I'm not 100% about rmsprop, adagrad and adadelta, but making them the same seems sensible so that the default value of epsilon for each of them is consistent.

edited: was looking at ADAMW not NADAM 🥇 these [two](https://openreview.net/forum?id=OM0jvwB8jIp57ZJjtNEZ) [versions](http://cs229.stanford.edu/proj2015/054_report.pdf) of the Nadam paper have the epsilons inside and outside the sqrt respectively, so ¯\_(ツ)_/¯

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -149,7 +149,7 @@ function update!(o::ADAGrad, x, Δ)
   η = o.eta
   acc = get!(o.acc, x, fill(ϵ, size(x)))::typeof(x)
   @. acc += Δ^2
-  @. Δ *= η / √(acc + ϵ)
+  @. Δ *= η / (√acc + ϵ)
 end
 
 """
@@ -169,7 +169,7 @@ function update!(o::ADADelta, x, Δ)
   ρ = o.rho
   acc, Δacc = get!(o.state, x, (zero(x), zero(x)))
   @. acc = ρ * acc + (1 - ρ) * Δ^2
-  @. Δ *= √(Δacc + ϵ) / √(acc + ϵ)
+  @. Δ *= √Δacc/ (√acc + ϵ)
   @. Δacc = ρ * Δacc + (1 - ρ) * Δ^2
   return Δ
 end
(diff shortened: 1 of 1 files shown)


MikeInnes:
Any thoughts from others on this? It seems reasonable to me.

To minimise the chance of mistakes I won't merge this until #379 is in.

CarloLucibello:
seems fine

tejank10:
Seems fine to me

MikeInnes:
Alright, just needs a rebase now then. Thanks!

JobJob:
Just realised I was looking at the ADAMW paper not the NADAM paper which is [here](https://openreview.net/forum?id=OM0jvwB8jIp57ZJjtNEZ). Easy mistake to make, anyone could make it  🤦‍♂️ - there are way too many f#&9ing ADAM variants!

Anyway, it does indeed have the epsilon inside the sqrt, though [this version](http://cs229.stanford.edu/proj2015/054_report.pdf) doesn't.

Anyway, ADAM does have it outside the sqrt, so I'm still in favour of having NADAM match ADAM, and the other minor changes in this PR, just to keep the use of the default epsilon value consistent (which it wouldn't be if it was sometimes inside a sqrt and sometimes not).

PS the new optimisers interface that was just merged already had the (earlier) correct version of ADAM, so didn't need to change it after the rebase.

MikeInnes:
Sounds good. Thanks!

PR #450, opened 2018-10-22
Add dataloader

boathit:
Adding a PyTorch-like data loader that provides (random) iterators over the dataset.

The usage is quite simple:
```julia
X = rand(10, 1000)
Y = rand(1, 1000)
dataloader = DataLoader(X, Y, batchsize=256, shuffle=true)

for (x, y) in dataloader
    ...
end

# and work seamlessly with the train!()
Flux.train!(loss, trainloader, opt)
```

### DIFF
--- a/src/dataloader.jl
+++ b/src/dataloader.jl
@@ -1,0 +1,59 @@
+using Random:randperm!
+
+"""
+    DataLoader(dataset...; batchsize::Int=100, shuffle=true)
+
+DataLoader provides iterators over the dataset.
+
+```julia
+X = rand(10, 1000)
+Y = rand(1, 1000)
+
+m = Dense(10, 1)
+loss(x, y) = Flux.mse(m(x), y)
+opt = ADAM(params(m))
+
+trainloader = DataLoader(X, Y, batchsize=256, shuffle=true)
+
+Flux.train!(loss, trainloader, opt)
+```
+"""
+struct DataLoader
+  dataset::Tuple
+  batchsize::Int
+  shuffle::Bool
+  indices::Vector{Int}
+  n::Int
+end
+
+function DataLoader(dataset::NTuple{N,AbstractArray}; batchsize::Int=100, shuffle=false) where N
+  l = last.(size.(dataset))
+  n = first(l)
+  all(n .== l) || throw(DimensionMismatch("All data should have the same length."))
+  indices = collect(1:n)
+  shuffle && randperm!(indices)
+  DataLoader(dataset, batchsize, shuffle, indices, n)
+end
+
+DataLoader(dataset...; batchsize::Int=100, shuffle=false) =
+  DataLoader(dataset, batchsize=batchsize, shuffle=shuffle)
+
+function Base.iterate(it::DataLoader, start=1)
+  if start > it.n
+      it.shuffle && randperm!(it.indices)
+      return nothing
+  end
+  nextstart = min(start + it.batchsize, it.n + 1)
+  i = it.indices[start:nextstart-1]
+  element = Tuple(copy(selectdim(x, ndims(x), i)) for x in it.dataset)
+  return element, nextstart
+end
+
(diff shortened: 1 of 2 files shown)


datnamer:
Would it be better to interface with JuliaML?
https://github.com/JuliaML/MLDataPattern.jl

boathit:
To my understanding, `DataLoader` differs from the `RandomBatches` in two aspects (correct me if I am wrong):

* You need to specify the number of mini-batches (`count`) before creating an iterator with `RandomBatches`. But in most of the deep learning applications what we want to do is
  ```julia
  for epoch = 1:epochs
        for item in iterator
               ...
        end
  end
  ```
  This means that you need to pre-calculate `count` based on the number of epochs and the length of your data if you use `RandomBatches`. While `DataLoader` is designed particularly for the above iteration pattern.

* `RandomBatches` adopt the strategy of sampling with replacement whereas `DataLoader` use sampling without replacement. To my knowledge, almost all deep learning models adopt the later one, i.e., we only iterate each data point once in each iteration  and repeat it multiple times (epochs).

src/dataloader.jl
```diff
@@ -0,0 +1,59 @@
...
+"""
+struct DataLoader
+  dataset::Tuple
+  batchsize::Int
+  shuffle::Bool
+  indices::Vector{Int}
+  n::Int
+end
+
+function DataLoader(dataset::NTuple{N,AbstractArray}; batchsize::Int=100, shuffle=false) where N
```
johnnychen94: Perhaps it's more clear to write
`DataLoader(dataset::AbstractArray; batchsize::Int, shuffle)` as constructor
here and 
write two more functions
`DataLoader(dataset::AbstractArray...; batchsize::Int, shuffle)` as is in Line 38
`DataLoader(dataset::Tuple{AbstractArray, Vararg{AbstractArray}}; batchsize::Int, shuffle)` ( a new one for Tuple input)

BTW, `NTuple{N,AbstractArray}` accepts `()` as input; what you might want is `Tuple{AbstractArray, Vararg{AbstractArray}}`

boathit: Good catch.

boathit:
@Evizero Thanks for pointing it out. It seems `batchview` or `eachbatch` will ignore the last batch if its size is smaller than the given batch size, is that designed on purpose?

By the way, why do `JuliaML` split the data utility functions into several different packages `MLDataUtils.jl`, `MLDataPattern.jl`, `MLLabelUtils.jl`? Won't that be good to put them into one?

Evizero:
> will ignore the last batch if its size is smaller than the given batch size, is that designed on purpose?

yes to force a fixed batch size for buffer reuse. in order to always use all data one can specify `maxsize` instead of `size`

> By the way, why do JuliaML split the data utility functions into several different packages 

separation of concerns. it also keeps the code base manageable. MLDataUtils is more or less just a metapackage exporting the others

pshashk:
Training neural networks sometimes require having a buffered batch generator running in background. For example, when running various augmentation procedures on CPU and copying data to GPU without interrupting training to maximize hardware utilization. Maybe it is possible to make DataLoader support this pattern by: 
1. optionally wrapping it in Channel or equivalent if some kind of `prefetch` argument is specified
2. having user supplied function being applied to each batch (e.g. augmentation, padding, copying to GPU, etc.)

staticfloat:
I just implemented something like this on my `sf/training` branch of `Metalhead.jl`.  It achieves pretty high performance when pulling in the ImageNet images training set; once I open a PR against `Metalhead.jl`, I would be happy to discuss how to best generalize the infrastructure to support more general datasets than just ImageNet.

ByzanTine:
Any update for this?

colinxs:
I'm also interested in this and would be happy to lend a hand. While development on MLDataPattern has slowed, it still seems like there's a lot of good work there and that we could revamp it to accommodate the above features?

Thoughts @Evizero?

PR #455, opened 2018-10-23
Load Juno-specific code conditionally.

### LINKED ISSUE #454: make Juno dependency conditional
Because of a dependency via Juno, using Flux on Julia#master currently suffers from https://github.com/JunoLab/Media.jl/issues/19.

I wonder if dependency on Juno could be made conditional, via Requires. Looking at the code, the only difficulty I see with this is the `@progress` macro in training.

tpapp:
Use the new JunoProgressLogger API for the progress bar via @logmsg.

Fixes #454.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,4 @@
 julia 0.7
-Juno
 MacroTools 0.3.3
 NNlib
 Requires
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -2,7 +2,7 @@ module Flux
 
 # Zero Flux Given
 
-using MacroTools, Juno, Requires, Reexport, Statistics, Random
+using MacroTools, Requires, Reexport, Statistics, Random
 using MacroTools: @forward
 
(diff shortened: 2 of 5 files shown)


tpapp:
@pfitzseb, can you please review if I am using the progressbar API right?

pfitzseb:
You'll also need to set the `_id` kwarg to a constant string or symbol to associate the different `@logmsg`s.

Could also completely get rid of Juno by using the TreeViews.jl API for displaying results instead, but that's probably out of scope for this PR.
Looks good otherwise.

tpapp:
@MikeInnes: please let me know if there is anything else I can improve in this PR to get it merged.

tpapp:
Is there a chance of merging this? If not, should I make a PR along some alternative lines? With #454, Flux is difficult to use on `master`.

MikeInnes:
Can you try `] add Media#mji/dynamic`?

MikeInnes:
Media is fixed and tagged now, so this shouldn't be necessary.

PR #413, opened 2018-09-27
evaluate both 2-ary DiffRules only when needed

mcabbott:
This should fix #412. 

I suppose it creates a few hundred more functions, hope that doesn't create any issues.

PR #456, opened 2018-10-23
Use promotion in binary derivatives to avoid issues with irrationals

andreasnoack:
Man, autodiff code is convoluted. All the closures. Besides the issue with size `zero` of irrationals there was also another promotion issue where the irrational caused promotion to `Float64` even when the input is `Float32.
```julia
julia> Tracker.derivative(t -> π*t, Float32(1.0))
3.141592653589793
```
This should be fixed with this PR together with the `zero` issue.

### DIFF
--- a/src/tracker/array.jl
+++ b/src/tracker/array.jl
@@ -223,7 +223,7 @@ end
 A::TrackedArray     / B::TrackedArray     = Tracker.track(/, A, B)
 A::AbstractVecOrMat / B::TrackedArray     = Tracker.track(/, A, B)
 A::TrackedArray     / B::AbstractVecOrMat = Tracker.track(/, A, B)
-@grad function (A / B)
+@grad function (A::AbstractVecOrMat / B::AbstractVecOrMat)
     return Tracker.data(A) / Tracker.data(B), function (Δ)
         Binv = inv(B)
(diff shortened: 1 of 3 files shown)


MikeInnes:
Indeed it is. Some of the convolutedness will go away with Capstan and Zygote, though you have to learn to love the closures.

This looks like a good change, but I'm just wondering, would it be possible to hook into Base's normal promotion system here, rather than defining each method separately?

andreasnoack:
Possibly but currently a `(::S<:Real,::TrackedReal{S})` pair is promoted to `(::TrackedReal{TS},::TrackedReal{TS})` where `TS=promote_type(T,S)`. If it is generally okay instead to promote to `(::TS,TrackedReal{TS})` then I think we can fully exploit the base promotion system. Do you think that would be okay or would it break other use cases?

MikeInnes:
I think that would be ok, although I'd expect promotion to effectively take the former option? Might be getting confused.

andreasnoack:
Okay. I think this one is a better solution. This one properly uses the promotion mechanism in Base which allows us to delete the two special cased `@grad` definitions introduced in https://github.com/FluxML/Flux.jl/pull/413 as long as we add a grad definition for `x^2` aka `literal_pow`.

PR #465, opened 2018-10-26
Destroy AD graph when doing in-place gradients

MikeInnes:
This is technically breaking if you call `back!` more than once on the same tracked output, but it's pretty easy to fix.

Dismantling the graph as we go along allows memory to be reclaimed as we do the backwards pass, which is likely to be useful especially for larger convnets and such. @avik-pal would be great if you could check for an improvement with this branch.

### DIFF
--- a/src/tracker/back.jl
+++ b/src/tracker/back.jl
@@ -19,47 +19,50 @@ function scan(x)
   return
 end
 
-function back_(c::Call, Δ)
+function back_(c::Call, Δ, once)
   Δs = c.func(Δ)
   (Δs isa Tuple && length(Δs) >= length(c.args)) ||
     error("Gradient is not a tuple of length $(length(c.args))")
-  foreach(back, c.args, data.(Δs))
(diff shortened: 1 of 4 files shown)


ViralBShah:
cc @denizyuret

denizyuret:
Couple pointers that may be useful:
Some relevant discussion: https://github.com/JuliaGPU/CuArrays.jl/issues/152 
Large network that requires this type of memory handling: https://github.com/KnetML/MAC-Network 
Relevant implementation in Knet: https://github.com/denizyuret/Knet.jl/blob/master/src/gcnode.jl

MikeInnes:
Note that I'm not currently force-freeing anything, though I have been thinking about how to do this safely. One way is to detach the CuArray from the GC and bump its refcount when storing it on the tape, then free it as we throw out that node. If the GC runs at any point between the forwards and backwards passes, *and* it is safe to do so, then the buffer will get freed immediately during the reverse pass.

Anyway, I'll merge this for now and we'll see if anyone complains.

Issue #468, opened 2018-10-26
Fatal error on show of TrackedArray UnionAll


iamed2:
```
julia> TrackedArray{S, N, Array{S, N}} where {N, S<:Real}
Error showing value of type UnionAll:
ERROR: UndefVarError: A not defined
Stacktrace:
 [1] show(::IOContext{REPL.Terminals.TTYTerminal}, ::Type{SYSTEM: show(lasterr) caused an error
UndefVarError(:A)

Stacktrace:
 [1] show(::IOContext{REPL.Terminals.TTYTerminal}, ::Type{ERROR: UndefVarError: A not defined
Stacktrace:
 [1] show(::IOContext{REPL.Terminals.TTYTerminal}, ::Type{fatal: error thrown and no exception handler available.
UndefVarError(var=:A)
rec_backtrace at /Users/ericdavies/repos/julia1p0/src/stackwalk.c:94
record_backtrace at /Users/ericdavies/repos/julia1p0/src/task.c:246 [inlined]
jl_throw at /Users/ericdavies/repos/julia1p0/src/task.c:577
jl_undefined_var_error at /Users/ericdavies/repos/julia1p0/src/rtutils.c:134
show at /Users/ericdavies/.julia/packages/Flux/rcN9D/src/tracker/array.jl:36
jl_fptr_trampoline at /Users/ericdavies/repos/julia1p0/src/gf.c:1831
show_datatype at ./show.jl:526
show at ./show.jl:436
print at ./strings/io.jl:31
print at ./strings/io.jl:42
show_tuple_as_call at ./show.jl:1490
jl_fptr_trampoline at /Users/ericdavies/repos/julia1p0/src/gf.c:1831
show_spec_linfo at ./stacktraces.jl:261
#show#9 at ./stacktraces.jl:272
unknown function (ip: 0x121dce8f3)
jl_fptr_trampoline at /Users/ericdavies/repos/julia1p0/src/gf.c:1831
#show at ./none:0 [inlined]
#show_trace_entry#630 at ./errorshow.jl:471
unknown function (ip: 0x121dce19c)
jl_fptr_trampoline at /Users/ericdavies/repos/julia1p0/src/gf.c:1831
#show_trace_entry at ./none:0
unknown function (ip: 0x121dcded8)
jl_fptr_trampoline at /Users/ericdavies/repos/julia1p0/src/gf.c:1831
show_backtrace at ./errorshow.jl:574
#showerror#612 at ./errorshow.jl:79
[...]

PR #469, opened 2018-10-26
Stop type treason with show of the TrackedArray type

### LINKED ISSUE #434: Showing array of param's  crashes julia
```
using Flux
using Flux.Tracker

D, N = (4, 10_000)
H = 100

W1 = param(rand(H, D))
b1 = param(rand(H))
W2 = param(rand(1, H))
b2 = param(rand(1))

θ = [W1, b1, W2, b2]
# ^This errors and crashes julia

4-element Array{Error showing value of type Array{ERROR: UndefVarError: A not defined
Stacktrace:
 [1] show(::IOContext{REPL.Terminals.TTYTerminal}, ::Type{SYSTEM: show(lasterr) caused an error
UndefVarError(:A)
```

iamed2:
Fixes https://github.com/FluxML/Flux.jl/issues/468
Fixes #434

### DIFF
--- a/src/tracker/lib/array.jl
+++ b/src/tracker/lib/array.jl
@@ -33,9 +33,6 @@ TrackedArray(x::AbstractArray) = TrackedArray(Call(), x, zero(x))
 
 Base.eltype(x::Type{<:TrackedArray{T}}) where T <: Real = TrackedReal{T}
 
-Base.show(io::IO, ::Type{TrackedArray{T,N,A}}) where {T,N,A<:AbstractArray{T,N}} =
-  print(io, "TrackedArray{…,$A}")
-
 function Base.summary(io::IO, x::TrackedArray)
   print(io, "Tracked ")
   summary(io, data(x))
--- a/src/tracker/lib/array.jl
+++ b/src/tracker/lib/array.jl
@@ -33,6 +33,11 @@ TrackedArray(x::AbstractArray) = TrackedArray(Call(), x, zero(x))
 
 Base.eltype(x::Type{<:TrackedArray{T}}) where T <: Real = TrackedReal{T}
 
+Base.show(io::IO, t::Type{TrackedArray{T,N,A}}) where {T,N,A<:AbstractArray{T,N}} =
+  @isdefined(A) ?
+    print(io, "TrackedArray{…,$A}") :
+    invoke(show, Tuple{IO,DataType}, io, t)
+
 function Base.summary(io::IO, x::TrackedArray)
   print(io, "Tracked ")
   summary(io, data(x))


MikeInnes:
What's the type treason – is there actually something incorrect about this method or is it just a Base bug?

iamed2:
You can see [Jeff's comment](https://github.com/JuliaLang/julia/issues/24195#issuecomment-337744534) for my previous encounter on a different package.

Basically this function is used for some critical display features (like showing method tables) which must not error. The specific crash I encountered was showing `TrackedArray{S, N, Array{S, N}} where {N, S<:Real}`, which is demonstrated in #468. I believe it has to do with showing unbound TypeVars.

MikeInnes:
Seems fair in the case there's a bug on our side, but it also seems like this method should never be called with `A` unbound. Doesn't happen in these cases, for example:
```
julia> foo(::Array{Complex{T}}) where T = 1

julia> foo([1+2im])
1

julia> foo(Complex[1+2im])
ERROR: MethodError: no method matching foo(::Array{Complex,1})
Closest candidates are:
  foo(::Array{Complex{T},N} where N) where T at REPL[1]:1
Stacktrace:
 [1] top-level scope at none:0

julia> foo(::Type{Complex{T}}) where T = 1
foo (generic function with 2 methods)

julia> foo(Complex)
ERROR: MethodError: no method matching foo(::Type{Complex})
Closest candidates are:
  foo(::Array{Complex{T},N} where N) where T at REPL[1]:1
  foo(::Type{Complex{T}}) where T at REPL[4]:1
```

Maybe I'm misunderstanding some way in which those are different.

Perhaps we can work around this by checking `@isdefined A`?

MikeInnes:
What I mean is, AFAICT the method we define here is not actually applicable to `TrackedArray{Int}`, so it's not clear to me why it's being called at all.

iamed2:
It actually happens during display of `TrackedArray{Int}`, when it's displayng `TrackedArray{Int}.body`

ararslan:
Can this be merged? It's a fairly unobtrusive change and it fixes a rather nasty error.

AlexRobson:
I think it'll also fix https://github.com/FluxML/Flux.jl/issues/434

MikeInnes:
Ok looks like it can be reproduced with
```julia
julia> foo(::Type{Complex{T}}) where T = T

julia> Complex.body
Complex{T<:Real}

julia> foo(Complex.body)
ERROR: UndefVarError: T not defined
```
This is just a really strange case, I guess. Let's try the fixed method and we can revisit it if further problems come up. Thanks a lot for narrowing this down @iamed2!

PR #473, opened 2018-10-29
Update CUDNN function calls

avik-pal:


MikeInnes:
Can we try to make this somewhat backwards-compatible with the current CuArrays release? I'm thinking of something like

```julia
if isdefined(CuArrays, :libcudnn_handle)
  handle() = CuArrays.libcudnn_handle[]
else
  handle() = CuArrays.CUDNN.handle()
end
```

avik-pal:
Some tests are failing in GPU. I am not very familiar with what changed in CuArrays. Would be great to get some pointers here:-

 ```
Expression: haskey((Flux.CUDA).descs, curnn.cell)
UndefVarError: descs not defined
```

and for the batch_size 5 tests we are getting `scalar getindex is disallowed`

MikeInnes:
You just missed out the `include("cudnn.jl")`. Testing now.

PR #479, opened 2018-11-01
Fix deprecations of optimisers

### LINKED ISSUE #478: Improve deprecation handling of optimisers
Old optimiser API; `opt(params(model))` errors with the new API without warnings

This is due to the deprecations expecting arrays instead of `Params` objects.

dhairyagandhi96:
Fixes #478


MikeInnes:
Maybe we should use `Union{Params,AbstractArray}` for now, so things like `SGD([p])` still work.

dhairyagandhi96:
Something like so?

MikeInnes:
Yep that looks good, thanks!

Issue #476, opened 2018-10-31
Remove `TrackedArray` from `AbstractArray` type tree?


Roger-luo:
Here is the concern:

not all the operations of an array is supported for `TrackedArray`, some of them is because it is not differentiable, some is just not implemented (e.g previously #458 ), thus, some operation of a `TrackedArray` will fallback to a method implemented for `AbstractArray` without error. This can be dangerous and cause the result to be nothing or just wrong (in the `view` case).

Since Julia types are duck-typed, there is no need to implement `TrackedArray` as `AbstractArray`, to use some of the functionality (like pretty printing) we just need a trait that inserts the value to the array type tree, e.g: https://github.com/Roger-luo/YAAD.jl/blob/master/src/show.jl

In this way, we can guarantee that if an operator is not implemented, you get an error.

I guess it is similar for `TrackedReal`, `TrackedTuple`, etc. or maybe we only need a type called `TrackedValue`.

What do you think?

johnnychen94:
One related issue about this: https://github.com/FluxML/Flux.jl/issues/378

KristofferC:
Ref https://github.com/eschnett/SIMD.jl/pull/36.

MikeInnes:
The `view` case wasn't great, but that was actually due to `back!` not throwing an error when given a non-tracked input. We can fix that and then you'll get the error you want. In general, code like `sum(view(x))` would have worked fine, so it was really just that particular case.

Generally, I can see where you're coming from, and this is the choice that Knet took as well, but it causes a lot of its own issues as well; both the `Real` and `AbstractArray` restrictions are heavily used in places that are perfectly valid for us to differentiate.

I do however think that Flux should never silently give incorrect gradients, so in general strategies we can take to avoid that are a good idea.

Roger-luo:
> The view case wasn't great, but that was actually due to back! not throwing an error when given a non-tracked input. We can fix that and then you'll get the error you want.

Yeah, you can fix that by adding a check in `back!`, but if this is removed from type tree, I will just get an no method matching blablabla, you don't need that check at all.

> both the Real and AbstractArray restrictions are heavily used in places that are perfectly valid for us to differentiate.

I'm not sure about this, is there any example? I mean you can still use some existing codes by inserting them back to a type tree by a trait.

MikeInnes:
[...]

PR #481, opened 2018-11-01
WIP: remove tracked type from corresponding type tree

Roger-luo:
Ref: #476


Roger-luo:
printing won't working with this, so I'll add some traits to insert the value back to `AbstractArray`

PR #490, opened 2018-11-14
Add missing eps overload for TrackedReal

ChrisRackauckas:
`eps` can be called on the number type as well, and this is missing from the TrackedReal overloads.


MikeInnes:
:+1: Thanks!

PR #491, opened 2018-11-14
Add zero on TrackedArray

ChrisRackauckas:
`zero([1.0,2.0,3.0])` worked but `zero` on a TrackedArray was a missing method. To keep immutability, I just use the multiplication by zero.

### DIFF
--- a/src/tracker/lib/array.jl
+++ b/src/tracker/lib/array.jl
@@ -79,6 +79,8 @@ Base.similar(x::TrackedArray, dims::Union{AbstractUnitRange,Integer}...) =
 
 Base.similar(x::TrackedArray, T::Type) = similar(data(x), T)
 
+Base.zero(x::Flux.Tracker.TrackedArray) = 0*x
+
 for op in [:(==), :≈]
     @eval Base.$op(x::TrackedArray, y::AbstractArray) = Base.$op(data(x), y)
     @eval Base.$op(x::AbstractArray, y::TrackedArray) = Base.$op(x, data(y))


MikeInnes:
Seems to be working for me:

```
julia> xs = param([1, 2, 3])
Tracked 3-element Array{Float64,1}:
 1.0
 2.0
 3.0

julia> zero(xs)
3-element Array{Float64,1}:
 0.0
 0.0
 0.0
```

ChrisRackauckas:
But it's not a TrackedArray

MikeInnes:
So?

ChrisRackauckas:
It's standard in generic type-stable code to initialize a value to zero at the start of an iterative algorithm. What other construct would you expect someone to use here? As it stands now, `zero(x)` is the correct thing to use for any array that is not a `Flux.TrackedArray`

MikeInnes:
True, but do you actually have a case where this gives a speedup? In general the AD tape is inherently type-unstable anyway, so this strikes me as being a premature optimisation.

ChrisRackauckas:
Are you trying to say that every generic algorithm should not strictly type their types because their usage with AD can be type-unstable? That seems like an odd argument. It's not an optimization either. If you put this array into a strictly typed mutable struct you will get an error because it's the strong type.

MikeInnes:
Not at all. That's quite a separate issue which is easy to fix more generally: 325035cf60cd4fbffd5e8c19c8f61b1bd99fb43f

ChrisRackauckas:
If that works for putting it into a type then sure. I still don't get why this one type shouldn't act like the others here.

MikeInnes:
Well, consistency is another distinct issue. This is a bit less unprecendented than you think it is, though.

```julia
julia> SharedArray([1, 2, 3]) |> zero |> typeof
Array{Int64,1}

julia> 1:3 |> zero |> typeof
Array{Int64,1}

julia> view([1, 2, 3], 1:2) |> zero |> typeof
Array{Int64,1}

julia> reinterpret(Float64,[1,2,3]) |> zero |> typeof
Array{Float64,1}

julia> CartesianIndices([1,2,3]) |> zero |> typeof
Array{CartesianIndex{1},1}
```

ChrisRackauckas:
SharedArray not doing it is a bug that should be addressed.

ChrisRackauckas:
`false .* u` is a nice workaround.

MikeInnes:
I remain unconvinced that this is an interface violation, especially given that Base provides the current definition of `zero` rather than us; if you're right Base is doing the wrong thing here and should probably throw an error here. This issue needs at least a citation and ideally a test case for which this feature is clearly needed.

In any case, will have to move over to Tracker.jl.

Issue #475, opened 2018-10-31
New Optimiser Docs


MikeInnes:
All the optimiser docstrings could do with a runthrough as they have gotten a bit out of sync with the code over time.

The docs fixes should mostly just amount to tweaking how we present the training loop. But we should also now document the optimiser internals so that people can easily understand them and build their own.

PR #493, opened 2018-11-16
New Optimiser Docs

dhairyagandhi96:
These docs are a WIP version of some of the changes made to the optimisers. Feedback welcome.

Ref: #475


dhairyagandhi96:
I have to fix the first bit of `optimisers.md` still, that's a bit of a mess.

MikeInnes:
I think there are two separate things here:

(1) updating the deprecated syntax in the current docs and
(2) adding docs on the new internal interface and how optimisers can be built.

I think it'd be useful to split those PRs and do (1) first; that's a bit easier to review and we can just get a release out then.

dhairyagandhi96:
I believe the current state should suffice (1). 

Could remove docstrings for consistency as well

MikeInnes:
Thanks a lot @dhairyagandhi96. Now this is merged we can look at expanding the docs with details on the new functionality.

dhairyagandhi96:
Absolutely!

PR #496, opened 2018-11-18
support choose data type in dense layer

Roger-luo:
This PR makes it possible for `Dense` to choose data type in its constructor.


MikeInnes:
I think this was address by #572, in a way which will work well across custom layer types as well. Does that address the need for this?

Roger-luo:
Yes, I think so. closed.

PR #498, opened 2018-11-18
Supply useful arguments to the Flux.train! callback function `cb`.

### LINKED ISSUE #489: Supply useful arguments to the Flux.train! callback function `cb`.
`Flux.train!` takes an optional argument `cb`. `cb` is a function that gets called after every training pass. 

Currently, `Flux.train!` doesn't supply any useful arguments to `cb`. This means that it can be used for simple callbacks like `cb = throttle(() -> println("training"), 10)`. However, `cb` cannot evaluate any interesting information from the evaluation of the last training pass, like the model input and output, the loss, etc.

It'd be great if Flux.train! could (perhaps on demand?) supply `cb` with useful info about the last training pass. The question now is what can be considered "useful". 

Here are two parameters I'd find useful:

* the current iteration count
* the loss for the last training pass

roberthoenig:
First shot at this.

The first commit is not directly related, but rather does some cleanups of the Flux.train! docs to reflect the new parameters.

The second commit adds callback info. I'm not too happy with the CallbackInfo struct's fieldnames. Also, maybe we should return `data(loss)` instead of  `loss`, since directly returning loss carries the risk that someone inadvertently messes with the parameters?

I tested this with the `xor1.jl` example in the modle-zoo, where I adapted the `evalcb` function:

```julia
evalcb = (info) -> begin
    dump(info)
    @show batch_loss(val)
end
```

Fixes #489.

### DIFF
--- a/docs/src/training/training.md
+++ b/docs/src/training/training.md
@@ -1,15 +1,16 @@
 # Training
 
-To actually train a model we need three things:
+To actually train a model we need four things:
 
 * A *objective function*, that evaluates how well a model is doing given some input data.
+* A collection of parameters that define the behavior of the model.
 * A collection of data points that will be provided to the objective function.
 * An [optimiser](optimisers.md) that will update the model parameters appropriately.
 
 With these we can call `Flux.train!`:
 
 ```julia
-Flux.train!(objective, data, opt)
+Flux.train!(objective, params, data, opt)
 ```
 
 There are plenty of examples in the [model zoo](https://github.com/FluxML/model-zoo).
@@ -26,7 +27,7 @@ m = Chain(
 loss(x, y) = Flux.mse(m(x), y)
 
 # later
-Flux.train!(loss, data, opt)
+Flux.train!(loss, Flux.params(m), data, opt)
 ```
 
 The objective will almost always be defined in terms of some *cost function* that measures the distance of the prediction `m(x)` from the target `y`. Flux has several of these built in, like `mse` for mean squared error or `crossentropy` for cross entropy loss, but you can calculate it however you want.
@@ -78,7 +79,7 @@ julia> @epochs 2 Flux.train!(...)
 `train!` takes an additional argument, `cb`, that's used for callbacks so that you can observe the training process. For example:
 
 ```julia
-train!(objective, data, opt, cb = () -> println("training"))
+train!(loss, params, data, opt, cb = () -> println("training"))
 ```
 
 Callbacks are called for every batch of training data. You can slow this down using `Flux.throttle(f, timeout)` which prevents `f` from being called more than once every `timeout` seconds.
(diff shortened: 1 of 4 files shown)


roberthoenig:
For now, I included four fields in the CallbackInfo struct. Once this is agreed on, we can easiliy extend it to include things like the microbenchmarks that @dhairyagandhi96 mentioned.

MikeInnes:
Dropped this for a while, but I think we should take another look at it.

Firstly we should avoid using reflection here, except to deprecate old-style callbacks; we can just tell people to write `_ -> ...`.

Also, I think we should avoid documenting `CallbackInfo` and leave it as a black box. Sounds odd but if all you know is that the callback takes a thing that you can do `.loss` on, then we are free to change the representation (e.g. go from storing a field to calculating it) without breaking any code.

roberthoenig:
> Firstly we should avoid using reflection here, except to deprecate old-style callbacks; we can just tell people to write _ -> ....

Agreed.

> Also, I think we should avoid documenting CallbackInfo and leave it as a black box. Sounds odd but if all you know is that the callback takes a thing that you can do .loss on, then we are free to change the representation (e.g. go from storing a field to calculating it) without breaking any code.

Not sure if I can follow you here. What does `go from storing a field to calculating it` mean? Also, the fields in `CallbackInfo` are supposed to be accessed directly by Flux users, right? So if they access the fields anyway, then changing the representation would break their code with or without docs, no?

oxinabox:
Some of this feels the same as the  things we didn't want to do in https://github.com/FluxML/Flux.jl/pull/247

(iirc)

I feel like we have even less reason to do them now, since we can write custom training loops so easily.

although I wasn't convinced that I was wrong then and I am not convinced this is wrong now.

arnavs:
how do the custom loops work?

oxinabox:
https://fluxml.ai/Flux.jl/dev/training/training/#Custom-Training-loops-1

ToucheSir:
I think QoL changes over time, newer PRs like https://github.com/FluxML/Flux.jl/pull/1471 and higher-level frameworks like FluxTraining/FastAI mean we can safely close this.

DhairyaLGandhi:
Doesn't make sense to close this since the functionality hasn't been added or the concerns addressed without merging the code in

ToucheSir:
This is a PR and not an issue though? Is there anything in here that you feel is not addressed by the existing open issues/PRs we have?

oxinabox:
I believe this is well addressed by writing a custom loop which is now well documented..
and this this can be closed.

 `Flux.train` and particularly it's callbacks, should remain limitted.
It is an API that is good for simple cases, but adding more power to it requires adding disproportionate complexity.
and at that point one is better off switching to a custom loop.

DhairyaLGandhi:
#1471 exposes a pretty simple api. Especially for the easy cases. I am already keeping the function very limited, but wouldn't make sense to stop users from doing the most basic of tasks without having to rewrite more code than is necessary.

darsnack:
Without getting into the callback vs custom loop discussion, as @ToucheSir pointed out, this is not an issue, it's a PR. Even if we were to merge the kind of functionality in this PR, it would be via #1471 and not this PR. I think we can safely close this unless the author wants to revive and push this over #1471.

PR #499, opened 2018-11-18
Deal with <= for TrackedReals

willtebbutt:
Relates to [this Turing.jl issue](https://github.com/TuringLang/Turing.jl/issues/600). I can't see that this obviously causes trouble.


MikeInnes:
:+1: Thanks!

PR #504, opened 2018-11-22
Make eltype of TrackedArray a TrackedReal{T}

ChrisRackauckas:
It breaks some array interface assumptions to have indexing return a value that doesn't match the prescribed `eltype` of the array. This changes it so that way the `eltype` of a TrackedArray is the same as the type of the elements that come out.


bhvieira:
Move this to Tracker perhaps?

PR #505, opened 2018-11-22
TrackedArray interface matches standard usage with immutable arrays

ChrisRackauckas:
Changes which were made:

- `zero` keeps type like StaticArrays
- UniformScaling overloads work around the generic `setindex!`
- `eltype(x) == typeof(x[i])` for TrackedArrays
- Fixed a random error message that failed because T wasn't defined

Together, this makes TrackedArrays work in many generic codes, including the DiffEq non-stiff and stiff integrators.

### DIFF
--- a/src/tracker/lib/array.jl
+++ b/src/tracker/lib/array.jl
@@ -114,6 +114,16 @@ Base.:-(xs::TrackedArray) = track(-, xs)
 
 @grad -(xs) = -data(xs), Δ -> (-Δ,)
 
+function Base.:+(A::TrackedArray, J::LinearAlgebra.UniformScaling)
+  n = LinearAlgebra.checksquare(A)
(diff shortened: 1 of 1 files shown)


MikeInnes:
The typing of `TrackedArray` is pretty weird, I know, but it's kind of hard to do the right thing here. Without some kind of "computed supertype relationship", hard-coding `TrackedArray <: AbstractArray{TrackedReal}` prevents us from re-using TrackedArray for, say, complex numbers. Or at least we'd have to create a new `TrackedComplexArray` and so on to handle different element types.

I'm not fundamentally opposed to that, but things have to get pretty bad before we start generating tracked types. For now I'll look at writing a better `similar` and we'll see how much of DiffEq that fixes.

MikeInnes:
This will have to move over to Tracker.jl.

PR #506, opened 2018-11-23
Update regularization docs

harryscholes:
Couple of changes here:
 - `norm` requires `LinearAlgebra`
 - I believe l2 regularization is the sum of the squared two-norms of a model's parameters

### DIFF
--- a/docs/src/models/regularisation.md
+++ b/docs/src/models/regularisation.md
@@ -15,6 +15,7 @@ loss(x, y) = crossentropy(softmax(m(x)), y)
 We can regularise this by taking the (L2) norm of the parameters, `m.W` and `m.b`.
 
 ```julia
+using LinearAlgebra
 penalty() = norm(m.W) + norm(m.b)
 loss(x, y) = crossentropy(softmax(m(x)), y) + penalty()
 ```
--- a/docs/src/models/regularisation.md
+++ b/docs/src/models/regularisation.md
@@ -12,11 +12,11 @@ m = Dense(10, 5)
 loss(x, y) = crossentropy(softmax(m(x)), y)
 ```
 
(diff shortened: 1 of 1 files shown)


harryscholes:
Couple of sources:
* [Elements of Statistical Learning](https://web.stanford.edu/~hastie/ElemStatLearn//printings/ESLII_print10.pdf), Hastie, section 3.4.1 pp 80
* [Deep Learning](https://www.deeplearningbook.org/contents/regularization.html), Goodfellow, section 7.1.1 pp 227

But it looks like only the weights are regularised, not the weights _and_ biases:

> notice that the intercept β0 has been left out of the penalty term. Penalization of the intercept would make the procedure depend on the origin chosen for Y; that is, adding a constant c to each of the targets yi would not simply result in a shift of the predictions by the same amount c.

Also see Hastie section 11.4 pp 395 on regularization for neural nets.

staticfloat:
This seems right; we can probably save some work by not using `norm` to avoid the `sqrt(x).^2` that's going on here, but oh well.

docs/src/models/regularisation.md
```diff
@@ -28,7 +29,7 @@ julia> params(m)
  param([0.355408 0.533092; … 0.430459 0.171498])
  param([0.0, 0.0, 0.0, 0.0, 0.0])
 
-julia> sum(norm, params(m))
+julia> sum(x->norm(x, 2)^2, params(m))
 26.01749952921026 (tracked)
```
staticfloat: I think this value will need to be updated.

ToucheSir:
The relevant docs were updated in https://github.com/FluxML/Flux.jl/pull/1260 for more clarity (and correctness), so this can be closed.

DhairyaLGandhi:
Thanks for the contribution! Hopefully we have the correct notion of what we expect from the docs in master

PR #513, opened 2018-12-09
Configurable dtypes for all layers and initializers

eamartin:
Hi
I thought Flux could benefit from easier configuration of dtypes, so I tried implementing something myself. 
I based the `FloatX` default dtype variable off of Theano: http://deeplearning.net/software/theano/library/config.html . 

Things I'm looking for feedback on:
* Currently there's no good way to set `FloatX`. Should we make `FloatX` configurable by environment variable or from a config variable? Is there a good way the user could set `FloatX` before all of the functions with optional argument `dtype=FloatX` get interpreted?
* This will break any user's initializers that do not have a `(::Type, size...)` method. Is that OK at this point?
* This is some of the first Julia code I've written, so interested in anywhere I violated best practice / code hygiene / etc.

Changes:
* All layers and initializers now optionally take a dtype argument.
* There is a new variable Flux.FloatX that defines the default
  floating point type (currently Float32).
* defined eltype(TrackedReal{T}) = T

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -71,8 +71,8 @@ end
 Dense(W, b) = Dense(W, b, identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
-               initW = glorot_uniform, initb = zeros)
-  return Dense(param(initW(out, in)), param(initb(out)), σ)
+               initW = glorot_uniform, initb = zeros, dtype = FloatX)
+  return Dense(param(initW(dtype, out, in)), param(initb(dtype, out)), σ)
 end
 
 @treelike Dense
@@ -103,8 +103,8 @@ struct Diagonal{T}
   β::T
 end
 
-Diagonal(in::Integer; initα = ones, initβ = zeros) =
-  Diagonal(param(initα(in)), param(initβ(in)))
+Diagonal(in::Integer; initα = ones, initβ = zeros, dtype = FloatX) =
+  Diagonal(param(initα(dtype, in)), param(initβ(dtype, in)))
 
 @treelike Diagonal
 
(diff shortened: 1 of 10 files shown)


eamartin:
I realized that I forgot to modify the optimizers to take an optional dtype argument that is defaulted to `FloatX`.
After a brief look, I do not fully understand what's going on (particularly with the deprecations). If we're going to go with this system of configuring floating point precision, I think it should also be extended to the optimizers at some point.

Roger-luo:
I personally prefer just use a type parameter. It might not be type stable with dtype. And this is consistent with most other Julia objects, e.g Array

mcabbott:
The function `gpu` converts to `Float32`, and acts on arrays, layers, models. Perhaps it and `cpu` should take a type argument, like `Dense(784,128) |> cpu(Float16)`?

eamartin:
@Roger-luo Would adding a type parameter still still support a user directly passing in their own weight matrices? For instance, if a user tries to create a Dense layer with `Float32` weights and `Float64` bias then what should happen? (pardon my lack of Julia knowledge here)

@mcabbott I like this syntax. It would be nice to avoid the initial allocation in float64 (or any non-desired precision). This could be particularly painful with large embedding layers and/or running in a memory-limited environment.

Roger-luo:
Yes, it will prevent user from using different precision within a single layer, .e.g

`Dense` can only has one type, it is either `Float64` or `Float32`. I don't think using `Float32` and `Float64` will be a common case, in most cases I think within a single layer, the element type should be the same. And this should usually be error at the construction time, (this is sometimes painful in PyTorch, since it errors util runtime). 

If someone want to use `Float64` and `Float32` within a single layer, it is always easy to create a custom wrapper around that. 

And I'll vote for @mcabbott 's syntax as well.

mcabbott:
One more thought is that, if I understand right, `Dense()` etc. always creates the object on the CPU. If things are so large that creating and then converting / moving is an issue, then perhaps the first problem to solve is not creating it with `Float16`, but with `CuArray{Float16}`. Could some macro `@gpu Float16 Chain( Dense( ...` do this?

MikeInnes:
Thanks for this patch, and apologies for not getting back on it sooner. I think this is generally the right idea and very useful functionality but I decided to do it a little differently; in particular, rather than adding options to layers we can make this more similar to `gpu` and `cpu`, i.e. functions of any model that will work even with user-defined layers.

Thanks for pushing this along anyhow! Let me know if I can help out with any other contributions.

PR #517, opened 2018-12-12
Fix decay argument in ADAMW

### LINKED ISSUE #507: WeightDecay argument in ADAMW optimiser
https://github.com/FluxML/Flux.jl/blob/a32c8a2e60870d49475719a36c74afed305e370a/src/optimise/optimisers.jl#L231 is throwing `UndefVarError: wd not defined`

I think `wd` should be replaced by the value of parameter `decay`.

So it would be
 ```julia
ADAMW(η = 0.001, β = (0.9, 0.999), decay = 0) =
  Optimiser(ADAM(η, β), WeightDecay(decay))
```

dhairyagandhi96:


MikeInnes:
Why'd you close this?

PR #522, opened 2018-12-20
Tiny bugfix: `stack` was still calling julia 0.6 `cat`

kolia:
Also added tiny test for good measure.


dhairyagandhi96:
Thanks for pointing it out!
Some of these changes were merged in ec8dde79c3c9ad64cb3a7d8656e2a14de31cdcf6, but its nice to have more tests.

Let me know if I can be helpful in making more contributions

Issue #514, opened 2018-12-09
Gradient of `mse` is of type Float64 for data of type Float32


vchuravy:
```
args = Flux.param.((CuArray(rand(Float32,32, 32, 1, 10)), CuArray(rand(Float32, 32, 32, 1, 10))))
y, back = Flux.Tracker.forward(Flux.mse, args...)
julia> typeof(y)
Flux.Tracker.TrackedReal{Float32}

julia> typeof(back(1.0f0))
Tuple{TrackedArray{…,CuArray{Float64,4}},TrackedArray{…,CuArray{Float64,4}}}
```

This seems to be due to the division by an integer in `mse`, since `inv(::Int)::Float64`

This worked for me as a workaround:
```
mse(ŷ, y) = sum((ŷ .- y).^2) / convert(Float32, length(y))
```

but hopefully we can fix this generally.

dhairyagandhi96:
Not limited to just `mse`, same behaviour persists in `crossentropy` as well. 

Converting the `size(y, 2)` in crossentropy to `Float32` gave the correct gradient...

dhairyagandhi96:
```
using Flux, Flux.Tracker, Flux.Optimise # master
using CuArrays # master

X = trainimgs(CIFAR10);

getarray(X) = float.(permutedims(channelview(X), (2, 3, 1)));
imgs = [getarray(X[i].img) for i in 1:50000];

valset = 49001:50000
valX = cat(imgs[valset]..., dims = 4) |> gpu;
valY = labels[:, valset] |> gpu;

m = Chain(
  Conv((5,5), 3=>16, relu),
  x -> maxpool(x, (2,2)),
  Conv((5,5), 16=>8, relu),
  x -> maxpool(x, (2,2)),
  x -> reshape(x, :, size(x, 4)),
  Dense(200, 120),
  Dense(120, 84),
  Dense(84, 10),
  softmax) |> gpu

using Flux: crossentropy, Momentum

loss(x, y) = sum(crossentropy(m(x), y))
opt = Momentum(params(m))

l = loss(valX, valY)
back!(l)
```

works with defining `Base.promote_rule(::Type{Float64}, ::Type{Float32}) = Float32`. Would need to figure out if this too harsh a condition to put on the types.

MikeInnes:
How about `convert(eltype(y), length(y))`? I think the type piracy in promote_rule is not going to work out well.

PR #533, opened 2019-01-03
Couple label types with the data type

dhairyagandhi96:
Ref: #514


dhairyagandhi96:
I can understand it causing some regressions, and potentially losing efficiency of some data types. The alternative being to have `convert` statements in the loss functions explicitly.

KristofferC:
I think this can be closed now?

PR #538, opened 2019-01-06
fix promotion by avoiding integer division in mse and crossentropy

### LINKED ISSUE #514: Gradient of `mse` is of type Float64 for data of type Float32

KristofferC:
Fixes #514

### DIFF
--- a/test/layers/stateless.jl
+++ b/test/layers/stateless.jl
@@ -49,4 +49,16 @@ const ϵ = 1e-7
   @testset "logitbinarycrossentropy" begin
     @test logitbinarycrossentropy.(logŷ, y) ≈ binarycrossentropy.(σ.(logŷ), y; ϵ=0)
   end
+
+  @testset "no spurious promotions" begin
+    for T in (Float16, Float32, Float64)
+      y = rand(T, 2)
+      ŷ = rand(T, 2)
+      for f in (mse, crossentropy, logitcrossentropy)
+        fwd, back = Flux.Tracker.forward(mse, ŷ, y)
+        @test typeof(fwd) == Flux.Tracker.TrackedReal{T}
+        @test eltype(back(one(T))[1]) == Flux.Tracker.TrackedReal{T}
+      end
+    end
+  end
 end
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -2,16 +2,16 @@ using NNlib: logsoftmax, logσ
 
 # Cost functions
 
-mse(ŷ, y) = sum((ŷ .- y).^2)/length(y)
+mse(ŷ, y) = sum((ŷ .- y).^2) * 1 // length(y)
 
 function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight = 1)
-  -sum(y .* log.(ŷ) .* weight) / size(y, 2)
+  -sum(y .* log.(ŷ) .* weight) * 1 // size(y, 2)
 end
 
 @deprecate logloss(x, y) crossentropy(x, y)
 
 function logitcrossentropy(logŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight = 1)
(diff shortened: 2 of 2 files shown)


MikeInnes:
Nice trick. Do you know of any downsides to doing things this way?

I think testing this is particularly important given how tempting it's going to be to simplify it.

KristofferC:
> Do you know of any downsides to doing things this way?

Not really. I've run with this PR a bit both on CPU and GPU and it seems to work ok. 

> I think testing this is particularly important given how tempting it's going to be to simplify it.

Agreed, will add some.

KristofferC:
Added tests.

KristofferC:
Wait, this test doesn't test the correct thing!

KristofferC:
It seems there is still spurious promotion when using onehot vectors as input (see the `@test_broken` in last commit.

KristofferC:
That promotion seems to also exist in the forward pass:

```
julia> g(ŷ, y) = sum(ŷ - y)
g (generic function with 1 method)

julia> typeof.((ŷ, y))
(Flux.OneHotVector, Array{Float32,1})

julia> fwd, back = Flux.Tracker.forward(g, ŷ, y);

julia> typeof(fwd)
Flux.Tracker.TrackedReal{Float64}
```

KristofferC:
I added a fix to the onehot promotion issue by mapping the onehot arrays to Bools instead of Float64.

test/layers/stateless.jl
```diff
@@ -49,4 +49,17 @@ const ϵ = 1e-7
   @testset "logitbinarycrossentropy" begin
     @test logitbinarycrossentropy.(logŷ, y) ≈ binarycrossentropy.(σ.(logŷ), y; ϵ=0)
   end
+
+  @testset "no spurious promotions" begin
+    for T in (Float16, Float32, Float64)
+      y = rand(T, 2)
+      for ŷ in (rand(T, 2), Flux.onehot(2, [1, 2]))
+          for f in (mse, crossentropy, logitcrossentropy)
+              fwd, back = Flux.Tracker.forward(mse, ŷ, y)
```
KristofferC: `mse -> f`

KristofferC:
Actually, mapping to `Bool` doesn't work in cases where we do a `log` on the onehot vector (for example in `crossentropy`. The `Bool` will be converted to a `Float64`.

staticfloat:
@MikeInnes is there anything else you're interested in about this?  If not, I'd like to merge this in.

MikeInnes:
The `param` change isn't necessary to fix `crossentropy`, which works fine with a tracked array and a one hot vector as input. The fix is also not ideal since it won't work with in-place gradients. Given that one hot vectors are generally not turned into parameters anyway I suggest just removing that code and we'll merge the basic fix in.

[For some context, the issue with `TrackedArray{<:Integer}` is that gradients may not be integers, which causes problems; hence the conversion to floating point both for one hot vectors and things like `param([1, 2, 3])`.)

KristofferC:
I think I did the changes requested.


> For some context, the issue with TrackedArray{<:Integer} is that gradients may not be integers, which causes problems

We should probably add a test for this problem then since this PR passed even with the new param def.

MikeInnes:
:+1: Thanks!

PR #546, opened 2019-01-10
Support random numbers as constants

ChrisRackauckas:
This adds support for rngs, specifically `rand`, `randn`, and `randexp`, by treating the generated random numbers as constants. The resulting derivative is the strong derivative, i.e. the derivative defined w.r.t. the chosen sequence of random numbers. For example, when looking at geometric Brownian motion (https://en.wikipedia.org/wiki/Geometric_Brownian_motion) and throwing this through the DiffEq SDE solvers, the values that are outputted are the derivatives w.r.t. the parameters of the strong solution of the SDE. This has a number of useful properties for optimization, since adding noise and optimizing the strong solutions has been shown to give robustness to data noise and thus acts as a form of regularization (https://www.sciencedirect.com/science/article/pii/S0025556414000510). Practically, this would let someone inject random numbers into a neural net and the calculated gradient would be the same as though that sequence of random forcings were hard coded into the net.

### DIFF
--- a/src/tracker/lib/real.jl
+++ b/src/tracker/lib/real.jl
@@ -55,14 +55,9 @@ Base.promote_rule(::Type{TrackedReal{S}},::Type{T}) where {S,T} =
 
 using Random
 
-Random.rand(::Type{Flux.Tracker.TrackedReal{T}}) where {T} = convert(Flux.Tracker.TrackedReal{T},rand(T))
-Random.rand(rng::AbstractRNG,::Type{Flux.Tracker.TrackedReal{T}}) where {T} = convert(Flux.Tracker.TrackedReal{T},rand(rng,T))
-
-Random.randn(::Type{Flux.Tracker.TrackedReal{T}}) where {T} = convert(Flux.Tracker.TrackedReal{T},randn(T))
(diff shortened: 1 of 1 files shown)


MikeInnes:
These definitions seem strange to me, can you give an example usage? For example, `rand(::Real)` is not actually defined except for `rand(::Integer)`, which has a very different interpretation from the one you're providing here.

Converting a constant to a tracked real is also semantically redundant, since we also allow constants themselves to act as values with no derivative.

ChrisRackauckas:
I guess `rand(::Real)` isn't needed, but `rand(::Type{<:Real})` is standard.

MikeInnes:
Ok, in that case it seems reasonable to respect the type. I think we only needs the set of `randx(rng, type)` methods then, and we'll get the `randx(type)` method for free.

ChrisRackauckas:
I only added the others since ForwardDiff.jl had them. Maybe @jrevels can comment on that, since https://github.com/JuliaDiff/ForwardDiff.jl/pull/378 should get pruned as well then.

jekbradbury:
There are a number of other ways to differentiate stochastic computations (see https://arxiv.org/abs/1506.05254 for a formalism based on a different one), but I'm not sure what it would look like to let the user choose.

MikeInnes:
My guess is that if you want to do anything other than this, you need something significantly more sophisticated (e.g. a program transform). Might be interesting to look into further though.

PR #550, opened 2019-01-10
modernize documentation

KristofferC:
This modernizes the documentation a bit:

- Use latest Documenter version.
- Check in a Project and Manifest for the docs so we fix the versions there (as recommended by Documenter.jl).
- Use build stages in Travis for docs (as recommended by Documenter.jl)
- Check in the `flux.css` so that local builds also gets styled.
- Fix a vestigial `SGD` entry in the documentation (changed to `Descent`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,4 @@
 *.jl.mem
 docs/build/
 docs/site/
-docs/flux.css
 deps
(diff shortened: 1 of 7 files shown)


KristofferC:
We also need to execute this step:

> Remove any links to the release-X.Y branches and remove the directories from your gh-pages branch.

KristofferC:
Bump.

MikeInnes:
Ok, I'll go ahead and do that. Thanks @KristofferC!

PR #552, opened 2019-01-10
Conv constructor type mismatch fix

AShedko:
This fixes the type mismatch in a constructor for Conv op.

Found when trying to create a 1d convolution like so:
```
c = Conv(Tuple(7),300=>128)
```
Which results in the following error:
```
MethodError: no method matching Conv(::TrackedArray{…,Array{Float32,3}}, 
::TrackedArray{…,Array{Float64,1}}, ::typeof(identity); stride=1, pad=0, dilation=1) ...
```
The important parts are :TrackedArray{…,Array{**Float32**,3}}, and ::TrackedArray{…,Array{**Float64**,1}}

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -15,6 +15,9 @@ Standard convolutional layer. `size` should be a tuple like `(2, 2)`.
 Data should be stored in WHCN order. In other words, a 100×100 RGB image would
 be a `100×100×3` array, and a batch of 50 would be a `100×100×3×50` array.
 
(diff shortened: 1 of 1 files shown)


AShedko:
@MikeInnes, PTaL

MikeInnes:
The example you posted works OK for me.

AShedko:
Weights are initialized with Float64 on CPU, is that an error?
No GPU atm, so I cannot check.

ToucheSir:
Looks like this is from the Tracker era and can be closed.

PR #553, opened 2019-01-11
Updated with more detailed instructions for installing CuArrays

xiaodaigh:
I think the current Flux.jl GPU instructions are too barebone. It took me a while to figure out how to install CuArrays and configure it to work with Flux.jl. This PR addresses that by writing some explicit and easy to follow instructions which should reduce the friction in adopting Flux.jl for GPUs.


MikeInnes:
Sure; I'm pretty sure it should work with just the toolkit.

MikeInnes:
:+1: Thanks!

Issue #501, opened 2018-11-20
MaxPool not working on latest master branch


roberthoenig:
```julia
using Flux

model = Chain(
    Conv((2,2), 1=>16, relu),
    MaxPool((2,2)),
    x -> reshape(x, :, size(x, 4)),
    Dense(144, 10),
    softmax);

X = rand(8,8,1,4);

loss(x) = begin
    Flux.crossentropy(model(x), rand(10,4))
end
opt = Flux.RMSProp(Flux.params(model));

Flux.train!(loss, [(X,)], opt)
```

The above code works fine in `[587475ba] Flux v0.6.9`,  but it crashes on the latest master (commit 4cba46c2) with the following error message (pruned):

> MethodError: no method matching ∇maxpool(::Array{Float64,4}, ::Array{Float32,4}, ::Array{Float32,4}, ::Tuple{Int64,Int64}; pad=(0, 0), stride=(2, 2))
Closest candidates are:
  ∇maxpool(::A<:AbstractArray, !Matched::A<:AbstractArray, !Matched::A<:AbstractArray, ::Any; pad, stride) where A<:AbstractArray at /home/robert/.julia/packages/NNlib/0EAe7/src/conv.jl:142
Stacktrace:
 [1] (::getfield(Flux.Tracker, Symbol("##457#458")){Base.Iterators.Pairs{Symbol,Tuple{Int64,Int64},Tuple{Symbol,Symbol},NamedTuple{(:pad, :stride),Tuple{Tuple{Int64,Int64},Tuple{Int64,Int64}}}},TrackedArray{Float32,4,Array{Float32,4}},Tuple{Int64,Int64},Array{Float32,4}})(::Array{Float64,4}) at /home/robert/.julia/dev/Flux/src/tracker/lib/array.jl:393

Roger-luo:
I guess

```
X = rand(Float32, 8, 8, 1, 4)
```

might fix your error? This is because the function requires input with same array type:

https://github.com/FluxML/NNlib.jl/blob/20aa7005c2ad2e511b4332a8af7920a2bac7a067/src/conv.jl#L162

eamartin:
I think the surprise here comes from differing default Float types. My understanding is that things in Julia default to Float64. Something in this chain must default to Float32 and mess things up. I'll investigate.

Roger-luo:
The default type in flux is float32, this is because most GPU device support float 32 only so most framework like pytorch use it as default

[...]

PR #554, opened 2019-01-11
Preserve element type in stateless layers

staticfloat:
This should fix issues where Float32 weights and activations get suddenly switched to Float64 on the backwards pass.  MWE to show current problem:

```julia
using Flux

x = param(randn(Float32, 10, 10, 4, 4)
mp = MeanPool((10,10))
Flux.back!(Flux.mse(mp(x), mp(x)))
```

This currently errors out with:
```
ERROR: MethodError: no method matching ∇meanpool(::Array{Float64,4}, ::Array{Float32,4}, ::Array{Float32,4}, ::Tuple{Int64,Int64}; pad=(0, 0), stride=(10, 10))
```

This is because the backward pass of `mse()` is spuriously generating `Float64`'s.  I don't really understand why it is doing that, but this PR seems to fix it.

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -2,16 +2,21 @@ using NNlib: logsoftmax, logσ
 
 # Cost functions
 
-mse(ŷ, y) = sum((ŷ .- y).^2)/length(y)
+function mse(ŷ, y)
+  x = sum((ŷ .- y).^2)
+  return x/Tracker.tracked_eltype(x)(length(y))
+end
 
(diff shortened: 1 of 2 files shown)


staticfloat:
It's possible this will fix https://github.com/FluxML/Flux.jl/issues/501

dhairyagandhi96:
For further reference, #514, #538, #533 might be relevant.

KristofferC:
> This is because the backward pass of mse() is spuriously generating Float64's. I don't really understand why it is doing that, but this PR seems to fix it.

It does so because it divides by an `int` and `inv(::Integer)` gives a Float64. I think #514 is a cleaner solution.

Issue #385, opened 2018-09-05
Batchnorm fails on GPU


avik-pal:
When trying to run batchnorm on gpus the following error occurs
```julia
ERROR: LoadError: GPU compilation failed, try inspecting generated code with any of the @device_code_... macros
[...]

PR #555, opened 2019-01-11
Work around GPU launch bug for BatchNorm

staticfloat:
X-ref: https://github.com/FluxML/Flux.jl/issues/385

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -138,7 +138,10 @@ function (BN::BatchNorm)(x)
   end
 
   let λ = BN.λ
-    λ.(reshape(γ, affine_shape...) .* ((x .- μ) ./ sqrt.(σ² .+ BN.ϵ)) .+ reshape(β, affine_shape...))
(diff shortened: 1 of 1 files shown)


MikeInnes:
Julia's escape analysis can make for some fun gpu debugging.

Do we have a test that catches this?

staticfloat:
Yes; here you go:

```julia
using CuArrays, Flux
μ = gpu(randn(Float32, 1, 1, 8, 1)); σ = μ; β = param(μ); γ = param(μ);
x = gpu(randn(Float32, 10, 10, 8, 1))
γ .* ((x .- μ) ./ σ) .+ β # <--- This should error out
```

MikeInnes:
Actually, I guess this is not escape analysis but some failure of broadcast compilation when it gets this complex.

That's useful but also, does this issue get caught by the flux gpu tests now? If not we should just add something that calls batchnorm.

KristofferC:
This content of this PR is in #586, can be closed.

staticfloat:
Great, thanks @KristofferC!

PR #560, opened 2019-01-16
Remove release branches from docs

MikeInnes:
@KristofferC hope this looks right.


KristofferC:
Just a heads up, it is possible to change the target branch to merge into from an existing PR. It's in the top after clicking the Edit button.

KristofferC:
Yes, this is good (these are no longer used, see the version selector in the newly deployed docs https://fluxml.ai/Flux.jl/stable/).

MikeInnes:
Ah, that's useful to know, thanks.

PR #562, opened 2019-01-19
Fix Tracker.update! to work for TrackedReal

cshenton:
Replaced calls to properties `.data` and `.grad` with their generic function calls so that `update!(opt, xs)` does not cause an `UndefRefError: access to undefined reference` error.

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -4,9 +4,9 @@ import Base.depwarn
 
 function update!(opt, xs)
   for x in xs
-    Δ = update!(opt, x.data, x.grad)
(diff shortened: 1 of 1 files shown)


dhairyagandhi96:
Maybe replace `data(x) .-= Δ` with `x.data .-= Δ`?

MikeInnes:
This should use the `Tracker.update!` function instead of reaching into `x.data` directly.

bhvieira:
Move this to Tracker?

MikeInnes:
Closing for now since this will have to be rebased over Tracker.jl.

PR #571, opened 2019-01-25
Adding `nest = true` option in `Tracker.gradient`

arnaudmgh:
otherwise fails and returns an error. Note that option has to be added in both `df` and `d2f`.

### DIFF
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -10,12 +10,12 @@ using Flux.Tracker
 f(x) = 3x^2 + 2x + 1
 
 # df/dx = 6x + 2
-df(x) = Tracker.gradient(f, x)[1]
+df(x) = Tracker.gradient(f, x; nest = true)[1]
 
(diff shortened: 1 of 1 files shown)


MikeInnes:
Thanks! I think it only has to be added to `df`, unless you end up differentiating `df2`, but it's just as good to be consistent here.

arnaudmgh:
You're welcome, and thanks for the additional info, I did not know that!

So, just suggesting; you may want to drop it from the declaration of `d2f`, like below, to illustrate calls that don't need `nest = true`. 

```
# df/dx = 6x + 2
df(x) = Tracker.gradient(f, x)[1]
df(2) # 14.0 (tracked)

# d²f/dx² = 6
df(x) = Tracker.gradient(f, x; nest = true)[1] # use "nest = true" for d2f(x) below
d2f(x) = Tracker.gradient(df, x)[1]

d2f(2.0) # 6.0 (tracked)
```

PR #575, opened 2019-01-28
Clean up parameter update API

MikeInnes:
cc @dhairyagandhi96

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -2,9 +2,9 @@ using Juno
 using Flux.Tracker: data, grad, back!
 import Base.depwarn
 
-function update!(opt, xs)
+function _update_params!(opt, xs)
   for x in xs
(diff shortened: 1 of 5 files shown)


src/optimise/train.jl
```diff
@@ -1,10 +1,14 @@
 using Juno
-using Flux.Tracker: data, grad, back!
+import Flux.Tracker: data, grad, back!, update!
 import Base.depwarn
 
-function update!(opt, xs)
+function update!(opt, x, x̄)
```
oxinabox: What is this for?
It is not called anywhere and has no docs AFAICT

MikeInnes: You pass an optimiser, parameter and gradient and it modifies the parameter accordingly.

We do need to do an update on parameter docs.

PR #578, opened 2019-01-29
Some memory improvements to OneHotMatrix

Keno:
1. Parameterize OneHotVector on Integer type, to avoid using more memory
   than required for vectors of them.
2. Switch OneHotMatrix from storing a vector of OneHotVectors to only storing
   the data and the size of the vector (reconstructing the vector locally), thus
   saving half the memory required and eliminating a transpose operation for
   matmul with OneHotMatrix on TPU.

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -1,32 +1,44 @@
 import Base: *
 
-struct OneHotVector <: AbstractVector{Bool}
-  ix::UInt32
-  of::UInt32
+struct OneHotVector{T <: Integer} <: AbstractVector{Bool}
+  ix::T
+  of::T
 end
 
-Base.size(xs::OneHotVector) = (Int64(xs.of),)
+Base.size(xs::OneHotVector) = (Int(xs.of),)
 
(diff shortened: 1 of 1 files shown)


KristofferC:
This no longer works:

```
julia>  y = Flux.onehotbatch(1:3, 1:10)
10×3 Flux.OneHotMatrix{Array{Int64,1}}:
  true  false  false
 false   true  false
 false  false   true
 false  false  false
 false  false  false
 false  false  false
 false  false  false
 false  false  false
 false  false  false
 false  false  false

julia> gpu(y)
ERROR: MethodError: no method matching Flux.OneHotMatrix(::Int64, ::CuArray{Float32,1})
Closest candidates are:
  Flux.OneHotMatrix(::Int64, ::A<:(AbstractArray{#s21,1} where #s21<:Integer)) where A<:(AbstractArray{#s21,1} where #s21<:Integer) at C:\Users\Kristoffer\.julia\dev\Flux\src\onehot.jl:18

````

ToucheSir:
@darsnack I assume #1448 mostly obviates this?

darsnack:
Yeah I think so

PR #586, opened 2019-01-30
work around extreme slowdown in BatchNorm due to julia performance bug in broadcast fusion

KristofferC:
Code:

```jl
using Flux
bn = BatchNorm(32);
testimg = randn(Float32, 416, 416, 32, 1);
@time bn(testimg)
```` 

Before

```
#   8.712519 seconds (224.78 M allocations: 4.133 GiB, 4.82% gc time)
```

After

```
#  0.059702 seconds (177 allocations: 105.631 MiB, 23.61% gc time)
```

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -138,7 +138,9 @@ function (BN::BatchNorm)(x)
   end
 
   let λ = BN.λ
-    λ.(reshape(γ, affine_shape...) .* ((x .- μ) ./ sqrt.(σ² .+ BN.ϵ)) .+ reshape(β, affine_shape...))
+    temp = reshape(γ, affine_shape...) .* ((x .- μ) ./ sqrt.(σ² .+ BN.ϵ)) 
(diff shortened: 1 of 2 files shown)


KristofferC:
Seems similar in spirit to https://github.com/FluxML/Flux.jl/pull/555.

MikeInnes:
Seems good but as in that PR, would be great to have a test (even if it's GPU-only).

KristofferC:
Test for what? Number of allocations?

MikeInnes:
That'd be fine given how stark the difference is. [This comment](https://github.com/FluxML/Flux.jl/pull/555#issuecomment-453419961) also indicated that this fixes GPU compilation, but I don't know if you were aiming for that as well.

KristofferC:
For some reason it didn't seem like my previous workaround worked anymore so I changed to the one in #555.

MikeInnes:
:+1: Thanks!

PR #589, opened 2019-02-02
Sparse grads for getindex

### LINKED ISSUE #577: Sparse
Related to #206 and partly to #66, a feature request: allow sparse gradients for getindex, as [pytorch does](https://pytorch.org/docs/stable/nn.html?highlight=embedding#torch.nn.functional.embedding) for its embedding layer. 

I have a large embedding matrix and the gradient for getindex is creating a large array of zeroes every time it's called and this kills my GPU performance. I don't think the getindex should use sparse data structures in the general case, and I'm not sure whats the best API for that, but this is just a big road block for me.

I have troubles using views and sparse arrays with CuArrays and Flux (https://github.com/JuliaGPU/CuArrays.jl/issues/267#issue-403606632), so I couldn't really experiment with the idea.

Some API ideas
a) Define a minimal `Embedding` struct and use a special gradient definition for getindexing into this type. `E[:,i]` dispatches to the sparse definition of getindex because `E isa Embedding`.
b) Define a special indexing type, i.e. `X[:, sparsely(i)]` dispatches to the sparse definition of getindex becase of the resulting type of `sparsely(i)`
c) Define a function `sparsegetindex(x, i...)` that is just `getindex` with a sparse grad definition

As a workaround I guess I can split the big embbedding matrix into multiple small ones, but I'm really not looking forward working with this kind of setup.

Thanks a lot and please let me know if I can help (but my GPU and Tracker knowledge is limited).

Drvi:
This is a proposition that fixes #577. I basically delay the gradient computation to happen inplace in the `accum!()` call and not in the `@grad` definition, which allows me to avoid the copy. On an extreme cornercase:

```julia
using CuArrays
using Flux

Ea = gpu(param(randn(64, 1_000_000)));
Eb = gpu(param(randn(64, 65_535)));
i = UInt16.(collect(1:5_000));
loss(i,n) = sum(sum(Eb[:, i] .+ Ea[:, rand(1:size(Ea,2), 1)]) for _ in 1:n)

function g(n, t, i)
   for _ in 1:t
      print("loss ")
      CuArrays.@time l = loss(i, n)
      print("back ")
      CuArrays.@time Flux.back!(l)
   end
end

g(100, 10, i)
```

Before, I got the following timings:
```
loss   0.149092 seconds (35.33 k CPU allocations: 2.126 MiB) (600 GPU allocations: 245.150 MiB, 26.88% gc time of which 100.00% spent allocating)
back   7.784618 seconds (64.09 k CPU allocations: 3.001 MiB, 29.91% gc time) (900 GPU allocations: 25.882 GiB, 3.20% gc time of which 100.00% spent allocating)
```

And after:
```
loss   0.062988 seconds (32.28 k CPU allocations: 2.061 MiB) (600 GPU allocations: 245.150 MiB)
back   0.405314 seconds (78.89 k CPU allocations: 4.502 MiB, 24.22% gc time) (1.30 k GPU allocations: 734.404 MiB)
```

There is a downside: `getindex`ing into the sparse structure I use is very slow and I think there is some `getindex`ing happening in the jacobians (at least the jacobian tests were complaining). 

Please let me know what you think.

### DIFF
--- a/src/tracker/back.jl
+++ b/src/tracker/back.jl
@@ -31,6 +31,41 @@ back_(::Call{Missing}, Δ, once) = error("`back!` was already used")
 
 accum!(x, Δ) = x .+ Δ
 accum!(x::AbstractArray, Δ) = (x .+= Δ)
+struct SparseGrad{T,N,S,P,O} <: AbstractArray{T,N} where P<:Union{T,AbstractArray{T,N}}
+    Δ::P
+    i::S
+    size::NTuple{N,Int}
+    function SparseGrad(Δ::P, i::S, size::NTuple{N,Int}, x::AbstractArray{T,N}) where {T,N,S,P<:Union{T,AbstractArray{T}}}
+        new{T,N,S,P,typeof(x)}(Δ, i, Base.size(x))
+    end
+end
+accum!(x::AbstractArray, Δ::SparseGrad) = (@inbounds(x[Δ.i...] += Δ.Δ); return x)
+Base.size(x::SparseGrad) = x.size
+Base.similar(x::SparseGrad{T,N,S,P,O}) where {T,N,S,P,O} = similar(O, size(x))
+
+#FIXME: Very slow getindex.
+function Base.getindex(x::SparseGrad, i...)
+    Base.checkbounds_indices(Bool, map(Base.OneTo, size(x)), i) || throw(BoundsError(x, i))
+
+    out = zero(x)
+    @inbounds out[x.i...] = x.Δ
+    @inbounds out[i...]
+end
+function Base.getindex(x::SparseGrad{T,N,S,P,O}, i::Int...)::T where {T,N,S,P,O}
+    Base.checkbounds_indices(Bool, map(Base.OneTo, size(x)), i) || throw(BoundsError(x, i))
+
+    li = LinearIndices(size(x))
+    @inbounds nonempty = li[x.i...]
+    @inbounds queryindices = li[i...]
+
+    outidx = indexin(queryindices, nonempty)[1]
+    isnothing(outidx) ? zero(T) : @inbounds x.Δ[outidx]::T
+end
+function Base.getindex(x::SparseGrad{T,N,S,P,O}, i::Int...)::T where {T,N,O,S<:NTuple{N,Int},P<:T}
+    Base.checkbounds_indices(Bool, map(Base.OneTo, size(x)), i) || throw(BoundsError(x, i))
+    x.i == i ? x.Δ : zero(T)
+end
+
 
 function back(x::Tracked, Δ, once)
   x.isleaf && (x.grad = accum!(x.grad, Δ); return)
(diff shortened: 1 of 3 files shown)


Drvi:
I realized the type signatures might be a little bit too strong here, so I'm adjusting. 
Strangely, at one point I got following error:
```
MethodError: no method matching Flux.Tracker.SparseGrad(::CuArray{Float64,1}, ::Tuple{UnitRange{Int64}}, ::Tuple{Int64}, ::CuArray{Float32,1})
```
this would suggest that 
`Δ` (`::CuArray{Float64,1}` and `xs` (`::CuArray{Float32,1}`) in `@grad` definition had different eltypes -- is that expected? (my loss was Float32 and my weights as well).

src/tracker/lib/array.jl
```diff
@@ -97,9 +97,8 @@ Base.getindex(xs::TrackedArray, i...) = track(getindex, xs, i...)
 
 @grad function getindex(xs::AbstractArray, i...)
   data(xs)[i...], function (Δ)
-    Δ′ = zero(xs)
-    Δ′[i...] = data(Δ)
-    (nobacksies(:getindex, Δ′), map(_->nothing, i)...)
+    checkbounds(xs, i...)
```
MikeInnes: Is this needed given that we already did `xs[i...]`?

src/tracker/back.jl
```diff
@@ -31,14 +31,54 @@ back_(::Call{Missing}, Δ, once) = error("`back!` was already used")
...
+    function SparseGrad(Δ::P, i::S, size::NTuple{N,Int}, x::AbstractArray{T,N}) where {T,N,S,P}
+        new{T,N,S,P,typeof(x)}(Δ, i, Base.size(x))
+    end
+end
+accum!(x::AbstractArray, Δ::SparseGrad) = (@inbounds(x[Δ.i...] += Δ.Δ); return x)
+Base.size(x::SparseGrad) = x.size
+Base.similar(x::SparseGrad{T,N,S,P,O}) where {T,N,S,P,O} = similar(O, size(x))
+
+#FIXME: Very slow getindex.
+function Base.getindex(x::SparseGrad, i...)
```
MikeInnes: Is it possible to just implement the scalar version and have the rest fall back? Or is this need for the GPU?

MikeInnes:
I like the general idea here but I don't think this should touch `back.jl`. `SparseGrad` should be orthogonal to the AD other than being used by `getindex` (like `OneHot` is). But I think this is pretty close to that anyway.

src/tracker/back.jl
Drvi: If I remember correctly, the scalar case is worth it only for a very small queries (the `indexin()`s are expensive), hence I added the allocating version for the general case.

src/tracker/lib/array.jl
Drvi: Good point:)

src/tracker/back.jl
Drvi: If I remember correctly, the scalar case is worth it only for a very small queries (the `indexin()`s are expensive), hence I added the allocating version for the general case.

src/tracker/lib/array.jl
Drvi: Good catch:)

pshashk:
Gradient definition of `view` is very similar to `getindex`. Maybe we can use sparse gradients there as well? 
That way things like `selectdim(embeddind_matrix, 2, indices)` will avoid all kinds of copies.

Drvi:
Yes, touching things in `back.jl` didn't feel like the optimal thing to do...
I now think that I should focus more on how `SubArray` handles indexing rather than on `SparseArray`. I have some studying to do with regards the this and broadcasting system, so this may take me a while.

Issue #582, opened 2019-01-29
Typical accuracy function using onecold with a OneHotMatrix fails to compile on GPU


KristofferC:
The following crashes when being compiled to GPU:

```jl
julia> x = rand(Float32, 10, 3) |> gpu;

julia> y = Flux.onehotbatch(1:3, 1:10) |> gpu;

julia> accuracy(x, y) = Flux.onecold(x) .== Flux.onecold(y);

julia> accuracy(x, y)
ERROR: GPU compilation of #23(CuArrays.CuKernelState, CUDAnative.CuDeviceArray{Bool,1,CUDAnative.AS.Global}, Base.Broadcast.Broadcasted{Nothing,Tuple{Base.OneTo{Int64}},typeof(==),Tuple{Base.Broadcast.Extruded{CUDAnative.CuDeviceArray{Int64,1,CUDAnative.AS.Global},Tuple{Bool},Tuple{Int64}},Base.Broadcast.Extruded{Array{Int64,1},Tuple{Bool},Tuple{Int64}}}}) failed
KernelError: passing and using non-bitstype argument

Argument 4 to your kernel function is of type Base.Broadcast.Broadcasted{Nothing,Tuple{Base.OneTo{Int64}},typeof(==),Tuple{Base.Broadcast.Extruded{CUDAnative.CuDeviceArray{Int64,1,CUDAnative.AS.Global},Tuple{Bool},Tuple{Int64}},Base.Broadcast.Extruded{Array{Int64,1},Tuple{Bool},Tuple{Int64}}}}.
That type is not isbits, and such arguments are only allowed when they are unused by the kernel.
```

zsz00:
i meet the  same problem..

KristofferC:
The PR got reverted, so would be good to reopen.

darsnack:
This is fixed with all the latest changes to one-hot-related stuff (see #1448)

PR #591, opened 2019-02-02
Fix OneHotVector/Matrix performance on GPU

dhairyagandhi96:
This PR aims to fix performance of OneHotMatrix performance on the GPU. It additionally adds some missing functionality to their behaviour.

### DIFF
--- a/test/cuda/cuda.jl
+++ b/test/cuda/cuda.jl
@@ -38,6 +38,12 @@ Flux.back!(sum(l))
 
 end
 
+@testset "onecold gpu" begin
+  x = rand(Float32, 10, 3) |> gpu;
+  y = Flux.onehotbatch(1:3, 1:10) |> gpu;
+  @test_nowarn Flux.onecold(x) .== Flux.onecold(y)
+end
+
 if CuArrays.libcudnn != nothing
     @info "Testing Flux/CUDNN"
     include("cudnn.jl")
--- a/test/cuda/cuda.jl
+++ b/test/cuda/cuda.jl
@@ -39,9 +39,10 @@ Flux.back!(sum(l))
 end
 
 @testset "onecold gpu" begin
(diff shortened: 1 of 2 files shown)


KristofferC:
Does this help with https://github.com/FluxML/Flux.jl/issues/582?

dhairyagandhi96:
For that, we want to ensure `onecold` returns a `CuArray`. I'll add a commit to make it so.

dhairyagandhi96:
This should help with #582

MikeInnes:
LGTM, but are there any tests we can add for this? e.g. returns CuArray on the GPU.

test/cuda/cuda.jl
```diff
@@ -38,6 +38,12 @@ Flux.back!(sum(l))
 
 end
 
+@testset "onecold gpu" begin
+  x = rand(Float32, 10, 3) |> gpu;
+  y = Flux.onehotbatch(1:3, 1:10) |> gpu;
+  @test_nowarn Flux.onecold(x) .== Flux.onecold(y)
```
KristofferC: `@test_nowarn` doesn't work anymore.

```
julia> using Test

julia> @test_nowarn @warn "foo"
┌ Warning: foo
└ @ Main REPL[5]:1
```

dhairyagandhi96: Oops. Perhaps?
```julia
@testset "onecold gpu" begin
     x = zeros(Float32, 10, 3) |> gpu;
     y = Flux.onehotbatch(ones(3), 1:10) |> gpu;
     res = Flux.onecold(x) .== Flux.onecold(y)
     @test res isa CuArray
end
```

KristofferC: That seems to indeed better test what we want to check.

dhairyagandhi96: Thanks bunches!

PR #592, opened 2019-02-04
Layer normalisation for images

### LINKED ISSUE #406: Layer normalisation does not work for images
The layer uses the `normalise` (stateless) function as defined [here](https://github.com/FluxML/Flux.jl/blob/master/src/layers/stateless.jl#L42-L53). This function calculates mean and std on `dims=1` but for images we need `dims=(1,2,3)` leaving out only the batch dimension. The following function should work:
```julia
function normalise(x)
    d = ndims(x) == 4 ? [1,2,3] : [1,]
    mu = Flux.mean(x, dims=d)
    sd = Flux.std(x, dims=d, mean=mu)
    return (x.-mu)./sd
end
```
also the type of `x` has to change in the function signature to allow for images, currently `x::AbstractVecOrMat` fails for e.g. `rand(Float32, 84,84,1,1)` since it allows only 1d or 2d arrays.

MJ10:
Resolves #406 

As suggested by @MikeInnes and @skariel, added `dims` argument to function, with default value as 1 and changed type of `x` to `AbstractArray`

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -40,12 +40,12 @@ but it is more numerically stable.
 logitbinarycrossentropy(logŷ, y) = (1 - y)*logŷ - logσ(logŷ)
 
 """
-    normalise(x::AbstractVecOrMat)
+    normalise(x::AbstractArray, dims::Int=1)
 
-Normalise each column of `x` to mean 0 and standard deviation 1.
+    Normalises x to mean 0 and standard deviation 1, across the dimensions given by dims. Defaults to normalising over columns.
 """
-function normalise(x::AbstractVecOrMat)
-  μ′ = mean(x, dims = 1)
-  σ′ = std(x, dims = 1, mean = μ′, corrected=false)
+function normalise(x::AbstractArray, dims::Int=1)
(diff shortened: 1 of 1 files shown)


MikeInnes:
LGTM. I think the docstring is a bit confusing. Maybe something like, "Normalises `x` to mean 0 and standard deviation 1, across the dimensions given by `dims`. Defaults to normalising over columns."

MJ10:
That sounds better, I'll make the changes.

dhairyagandhi96:
Oops, looks like GitHub's Web UI isn't happy about the conflict resolution. Can you take a look at the spurious changes please?

Many thanks!

MJ10:
Sorry about the multiple pushes, made wrong commit by mistake.
I've resolved the conflicts now.

MJ10:
Really sorry about that, fixed it.

dhairyagandhi96:
Absolutely no worries. Thanks for the PR!

Could you add the keyword `corrected` as is there in master currently?
Looks good to me!

MJ10:
I missed that while resolving the conflicts, sorry!

PR #595, opened 2019-02-05
Add higher order autodiff for getindex

Keno:
This fixed higher order autodiff or getindex (and by extension vcat, since
it's backward pass uses getindex). This makes tracker able to handle
the third order derivatives needed to implement [1].

[1] Physics-Informed Generative Adversarial Networks for Stochastic Differential Equations. https://arxiv.org/abs/1811.02033

### DIFF
--- a/src/tracker/lib/array.jl
+++ b/src/tracker/lib/array.jl
@@ -95,12 +95,27 @@ end
 
 Base.getindex(xs::TrackedArray, i...) = track(getindex, xs, i...)
 
-@grad function getindex(xs::AbstractArray, i...)
-  data(xs)[i...], function (Δ)
-    Δ′ = zero(xs)
-    Δ′[i...] = data(Δ)
-    (nobacksies(:getindex, Δ′), map(_->nothing, i)...)
+struct ∇getindex{T,S}
+    xs::T
+    i::S
+end
+
+function (g::∇getindex)(Δ)
+  Δ′ = zero(g.xs)
+  Δ′[g.i...] = Δ
+  (Δ′, map(_->nothing, g.i)...)
+end
+(g::∇getindex)(Δ::TrackedArray) = track(g, Δ)
+
+@grad function (g::∇getindex)(Δ)
+  z, back = g(data(Δ)), function(Δ′′)
+    (Δ′′[1][g.i...],)
   end
+  z, back
+end
+
+@grad function getindex(xs::AbstractArray, i...)
+  data(xs)[i...], ∇getindex(xs, i)
 end
 
 Base.view(x::TrackedArray, inds...) = track(Base.view, x, inds...)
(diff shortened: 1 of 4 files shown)


AzamatB:
@Keno are you implementing [1]? Will you add it to the model zoo?

Keno:
We're working on an extension of that work. I can see if we can put the baseline implementation in the model zoo.

AzamatB:
@Keno thanks a lot! Would be great to have that implementation in the model zoo :)

src/tracker/lib/array.jl
```diff
@@ -95,14 +95,27 @@ end
 
 Base.getindex(xs::TrackedArray, i...) = track(getindex, xs, i...)
 
-@grad function getindex(xs::AbstractArray, i...)
-  data(xs)[i...], function (Δ)
-    Δ′ = zero(xs)
-    Δ′[i...] = data(Δ)
-    (nobacksies(:getindex, Δ′), map(_->nothing, i)...)
+struct getindex_back{T,S}
```
MikeInnes: Would prefer this to follow the usual naming conventions.

src/tracker/lib/array.jl
```diff
@@ -95,14 +95,27 @@ end
...
+    xs::T
+    i::S
+end
+
+function (g::getindex_back)(Δ)
+  Δ′ = zero(g.xs)
+  Δ′[g.i...] = data(Δ)
```
MikeInnes: Calling `data` here drops the gradient for `Δ`, which you won't want here.

test/tracker.jl
```diff
@@ -36,6 +36,15 @@ gradtest(f, dims...) = gradtest(f, rand.(Float64, dims)...)
 
 @testset "indexing & slicing" begin
   gradtest(x->view(x, 1:2, 1:2), rand(4, 4))
+
+  # Nested AD for getindex
+  @test gradient([1.0, 2.0, 3.0]; nest=true) do x
```
MikeInnes: Would be good to have a numerical check (or one against ForwardDiff, if that's too unstable).

src/tracker/lib/array.jl
```diff
@@ -95,14 +95,27 @@ end
...
+  (Δ′, map(_->nothing, g.i)...)
+end
+
+@grad function (g::getindex_back)(Δ)
```
MikeInnes: This doesn't actually get called at the moment.

src/tracker/lib/array.jl
```diff
@@ -95,14 +95,27 @@ end
...
+  g(Δ), function(Δ′′)
+    (Δ′′[g.i...], map(_->nothing, g.i)...)
```
MikeInnes: The `nothing`s aren't necessary here.

test/tracker.jl
Keno: yeah, I realized I got it wrong ;). Will push a fix shortly.

Keno:
> Thanks, this seems like generally the right idea. If your model is training with this though, it shouldn't be :)

It's not. I had also realized I got it wrong. Thanks for the review though. I'll fix it up.

Keno:
Alright, try this.

Drvi:
Since the `∇getindex` struct holds the gradient and the destination indices, I think MR is not that far from the idea in #589 (I try to avoid allocating the full array of zeros and then copying the grad into it by specializing the `accum!()` function to just add the grad inplace). It really helped performance in my case.

The logic for getindex is getting more complex here, so I'll see if I can make #589 compatible.

src/tracker/lib/array.jl
```diff
@@ -95,12 +95,27 @@ end
...
+
+@grad function (g::∇getindex)(Δ)
+  z, back = g(data(Δ)), function(Δ′′)
+    (Δ′′[1][g.i...],)
   end
+  z, back
+end
+
+@grad function getindex(xs::AbstractArray, i...)
+  data(xs)[i...], ∇getindex(xs, i)
```
MikeInnes: You could simplify this by having `∇getindex` return an array and wrapping it to return the right tuple. Then the changes to TrackedTuples and backprop wouldn't be needed.

Keno: so have a closure here that returns the tuple or sth to that extent?

MikeInnes: Yeah exactly

Issue #542, opened 2019-01-08
No method for computing determinant of TrackedMatrix


kylejbrown17:
Simple test case:
```julia
using Flux, LinearAlgebra

det(param(ones(2,2)))
```
yields the following error message:
```
ERROR: MethodError: no method matching lu!(::TrackedArray{…,Array{Float64,2}}, ::Val{true}; check=false)
Closest candidates are:
[...]

PR #596, opened 2019-02-05
Fixed issue #542.

IvanYashchuk:
Added tracking of LinearAlgebra.det and its grad method.

The gradient of the determinant is taken from
http://www.cs.cmu.edu/~zkolter/course/15-884/linalg-review.pdf (Section 4.5)

It is also mentioned that $det{A} A^{-T}$ can be used, this form is also written in 
Matrix Cookbook Sec. 2.1.2
https://www.ics.uci.edu/~welling/teaching/KernelsICS273B/MatrixCookBook.pdf

I don't know what variant is more appropriate.

### DIFF
--- a/src/tracker/lib/array.jl
+++ b/src/tracker/lib/array.jl
@@ -1,7 +1,7 @@
 import Base: *
 
 import LinearAlgebra
-import LinearAlgebra: inv, det, \, /
+import LinearAlgebra: inv, det, logdet, logabsdet, \, /
 
 using Statistics
 using LinearAlgebra: Transpose, Adjoint, diagm, diag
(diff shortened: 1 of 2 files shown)


KristofferC:
Regarding your first implementation, note that `adj` here means Adjugate matrix and not `adjoint` (and in fact `adj(A) = 1/det(A) * inv(A)`

IvanYashchuk:
Yes, when I thought about it again, I've noticed it.
I changed it already.

IvanYashchuk:
Can someone guide me, how should it be tested?

The result matches ForwardDiff.
```
julia> f(x) = det(x)
julia> df(x) = Flux.Tracker.gradient(f, x)[1]
julia> a = rand(3, 3)
3×3 Array{Float64,2}:
 0.92519   0.494903   0.380887
 0.688972  0.0461206  0.226083
 0.961183  0.358504   0.60728
julia> ForwardDiff.gradient(f, a) -  df(a)
Tracked 3×3 Array{Float64,2}:
  0.0          -5.55112e-17   8.32667e-17
 -2.77556e-17   2.77556e-17   0.0
 -4.16334e-17   2.77556e-17  -5.55112e-17
```

mcabbott:
For tests I think you just need to add this to `test/tracker.jl`:
```
@test gradtest(det, (4,4))
```
May I suggest adding also `logdet` and `logabsdet` at the same time? (Just `Δ * transpose(inv(xs)`.) 

Is it possible to cache the forward result? I was going to suggest the following, but it breaks second derivatives: 
```
@grad function det(xs) 
    d = det(data(xs))
    d, Δ -> (Δ * d * transpose(inv(xs)),)
end
```
Note that both [det](https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/generic.jl#L1293) and [inv](https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/dense.jl#L715) first work out `lu(x)` so even more re-use might be possible, but would add a lot of complexity.

MikeInnes:
This is great. Thanks a lot for the patch!

PR #599, opened 2019-02-06
Handle various cases of multiplying transpose-wrapped matrices

Keno:
See test cases. I hit these while taking third-order derivatives of
matrix multiplies (whose gradient definitions use transpose).

### DIFF
--- a/src/tracker/lib/array.jl
+++ b/src/tracker/lib/array.jl
@@ -361,6 +361,9 @@ x::TrackedVector  * y::TrackedVector  = track(*, x, y)
 @grad a::AbstractMatrix * b::AbstractVecOrMat =
   data(a)*data(b), Δ -> (Δ * transpose(b), transpose(a) * Δ)
 
+@grad a::TrackedMatrix{<:Any, <:Transpose} * b::AbstractVecOrMat =
+  data(a)*data(b), Δ -> (transpose(b * transpose(Δ)), transpose(a) * Δ)
+
 # NNlib
 
 using NNlib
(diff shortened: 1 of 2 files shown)


src/tracker/lib/array.jl
```diff
@@ -361,6 +361,9 @@ x::TrackedVector  * y::TrackedVector  = track(*, x, y)
 @grad a::AbstractMatrix * b::AbstractVecOrMat =
   data(a)*data(b), Δ -> (Δ * transpose(b), transpose(a) * Δ)
 
+@grad a::TrackedMatrix{<:Any, <:Transpose} * b::AbstractVecOrMat =
+  data(a)*data(b), Δ -> (transpose(b * transpose(Δ)), transpose(a) * Δ)
```
MikeInnes: Worth commenting that this is due to the type restriction on `grad`. I'm thinking of loosening that in future, in which case we could remove this.

ToucheSir:
Can this be closed or moved to the Tracker repo?

DhairyaLGandhi:
Things that should be moved to tracker should be moved. Instead of being closed. Tracker is still pretty usable despite not getting updates, so small bugs like these can be merged there.

CarloLucibello:
Since even you did care enough to reopen but not enough to move the pr over, let's close it. It will wait with the other 26 issues I labeled as move to tracker more than a month ago and no one ever cared about, and likely never will, no point in keeping this issue open in flux forever

DhairyaLGandhi:
So the reason for not doing it immediately was that it was 4 in the morning for me. I'll move the issues, but the prs will do them over the week

PR #600, opened 2019-02-06
Adapt to the new CUDAdrv.CuPtr pointer type.

maleadt:
FWIW, this would have been easier with some type aliases in the `ccall` signatures (from CuArrays.jl, `cudnnTensorDescriptor_t`, `Ptr{cudnnTensorFormat_t}`, etc).

There also seems to have sneaked in a `allowscalar` violation, but I don't have the time to hunt that down right now.

The dep on `CUDAdrv` would also need the Requires-style handling like CuArrays, so it might be easier to move those wrappers to CuArrays not to have to do that.

Ref https://github.com/JuliaGPU/CUDAdrv.jl/pull/125

### DIFF
--- a/src/cuda/cuda.jl
+++ b/src/cuda/cuda.jl
@@ -5,14 +5,14 @@ import ..CuArrays.CUDAdrv: CuPtr, CU_NULL
 using Pkg.TOML
 
 function version_check()
-  minor_version = 9
+  major_version = 1
   project = joinpath(dirname(pathof(CuArrays)), "../Project.toml")
   project = TOML.parse(String(read(project)))
   version = VersionNumber(get(project, "version", "0.0.0"))
-  if !(version.major == 0 && version.minor == minor_version)
+  if version.major != major_version
     @warn """
-    Flux is only supported with CuArrays v0.$minor_version.
-    Try running `] pin CuArrays@0.$minor_version`.
+    Flux is only supported with CuArrays v$major_version.x.
+    Try running `] pin CuArrays@$major_version`.
     """
   end
 end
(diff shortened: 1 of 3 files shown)


MikeInnes:
The scalar indexing issue is probably unrelated; pull 777571d4b4fdd96e4187819bcbe0e5c52a40a26e to fix that.

KristofferC:
Im getting

```
┌ Warning: Package Flux does not have CUDAdrv in its dependencies:
│ - If you have Flux checked out for development and have
│   added CUDAdrv as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Flux
└ Loading CUDAdrv into Flux from project dependency, future warnings for Flux are suppressed.
```

with this.

MikeInnes:
Yeah, we'll have to import CUDAdrv indirectly (e.g import `.CuArrays.CUDAdrv`).

maleadt:
Added import of CUDAdrv stuff through CuArrays.

dhairyagandhi96:
Lgtm

MikeInnes:
@dhairyagandhi96 can you take a look at fixing that and getting this in?

dhairyagandhi96:
Sure thing. I added a commit that takes care of the warning.

maleadt:
LGTM. I changed the CuArrays version check logic to only look at the major version, as per semver.
I'll tag CuArrays 1.0 then.

maleadt:
Passes tests locally, except for the spurious CUDNN failures.

PR #601, opened 2019-02-06
Revert "Fix OneHotVector/Matrix performance on GPU"

MikeInnes:
Reverts FluxML/Flux.jl#591

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -9,8 +9,6 @@ Base.size(xs::OneHotVector) = (Int64(xs.of),)
 
 Base.getindex(xs::OneHotVector, i::Integer) = i == xs.ix
 
-Base.getindex(xs::OneHotVector, ::Colon) = xs
-
 A::AbstractMatrix * b::OneHotVector = A[:, b.ix]
 
 struct OneHotMatrix{A<:AbstractVector{OneHotVector}} <: AbstractMatrix{Bool}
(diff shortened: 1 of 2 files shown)


MikeInnes:
@dhairyagandhi96 I'm reverting this because the test it added appears to fail. Hopefully you can figure that out.

KristofferC:
Test passes for me...

```
julia>   res = Flux.onecold(x) .== Flux.onecold(y)
3-element CuArray{Bool,1}:
 true
 true
 true

julia> @test res isa CuArray
Test Passed
```

MikeInnes:
It fails on CI due to scalar indexing.

KristofferC:
Ah, I see.

dhairyagandhi96:
After a bit of digging, the problem really seems to be in `Flux.onecold(x)` in the test, which admittedly does still need `argmax` across a dimension for `CuArray`s. The failure in the issue really was the broadcast, which would be resolved if `Flux.onecold(y)` returns a `CuArray` which it does (with the original PR) without scalar indexing. So in that case, would testing `Flux.onecold(y) isa CuArray` be equivalent to the intention of the test to begin with? 

```Julia
julia> using CuArrays: allowscalar

julia> allowscalar(false)
false

julia> y = Flux.onehotbatch(1:300, 1:10000) |> gpu;

julia> using BenchmarkTools

julia> @btime Flux.onecold(y)
  15.456 μs (57 allocations: 1.92 KiB)
300-element CuArray{Int64,1}:
   1
   2
   3
   4
...
```

MikeInnes:
Yeah, it's best to test the specific intent of the changes in the PR. It seemed like there were a couple of other changes so tests for them might be good to, unless they all directly enable that specific test to work.

PR #606, opened 2019-02-08
Add `corrected` argument to std 

pshashk:
Makes `normalise` differentiable

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -44,13 +44,13 @@ logitbinarycrossentropy(logŷ, y) = (1 - y)*logŷ - logσ(logŷ)
 
     Normalises x to mean 0 and standard deviation 1, across the dimensions given by dims. Defaults to normalising over columns.
 """
-function normalise(x::AbstractArray; dims::Int=1)
+function normalise(x::AbstractArray; dims=1)
   μ′ = mean(x, dims = dims)
   σ′ = std(x, dims = dims, mean = μ′, corrected=false)
   return (x .- μ′) ./ σ′
 end
 
-function normalise(x::AbstractArray, dims::Int=1)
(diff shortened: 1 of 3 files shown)


MikeInnes:
> Makes `normalise` differentiable

Can you add a test for that too? Otherwise looks good.

pshashk:
Maybe we should change `dims` in `normalise` to keyword argument for consistency?

MikeInnes:
I was just thinking the same. It would need a deprecation for the current signature though.

src/layers/stateless.jl
```diff
@@ -40,12 +40,17 @@ but it is more numerically stable.
 logitbinarycrossentropy(logŷ, y) = (1 - y)*logŷ - logσ(logŷ)
 
 """
-    normalise(x::AbstractArray, dims::Int=1)
+    normalise(x::AbstractArray; dims::Int=1)
 
     Normalises x to mean 0 and standard deviation 1, across the dimensions given by dims. Defaults to normalising over columns.
 """
-function normalise(x::AbstractArray, dims::Int=1)
+function normalise(x::AbstractArray; dims::Int=1)
```
MikeInnes: I think this type restriction can also be removed; it's fine for us to forward tuples of ints, etc.

src/layers/stateless.jl
```diff
@@ -40,12 +40,17 @@ but it is more numerically stable.
 logitbinarycrossentropy(logŷ, y) = (1 - y)*logŷ - logσ(logŷ)
 
 """
-    normalise(x::AbstractArray, dims::Int=1)
+    normalise(x::AbstractArray; dims::Int=1)
```
MikeInnes: Watch out for the docstring also

MikeInnes:
Thanks a lot!

PR #608, opened 2019-02-09
Batch dimensioning for single channel inputs

thebhatman:
Let's say you have a batch of thousand 32x32 single-channel images. Then x will be a three-dimensional array and weights will be a four-dimensional array. So, in this case, we need the size(x) to be 32x32x1x1000. This takes care of it.


thebhatman:
Travis CI says "Failed to connect to storage.googleapis.com port 443: Connection timed out
gzip: stdin: unexpected end of file".

dhairyagandhi96:
I've restarted the job, hopefully that was a spurious fail.

thebhatman:
@dhairyagandhi96 , I am now trying to implement Batch dimensioning using dispatch(I am new to this). My last commit was using dispatch. Can you tell me if I am on the right track?

MikeInnes:
You're being overly restrictive in the type parameters here; `Conv` needs to work on more than just 2D inputs and on number types other than Float64.

Also, this makes the channel dimension optional rather than the batch dimension.

thebhatman:
Yes. I agree. I started off by trying to solve this small subproblem. Now I am working on the generalization of the conditions required.

ToucheSir:
Given the lack of activity, need to generalize to higher dimensions (e.g. Flux3D) and existence of `Flux.unsqueeze`, I think this can be safely closed.

PR #609, opened 2019-02-09
Handle back! and isleaf sensibly

oxinabox:
calling `back!` should not change if something `isleaf` or not.
and it should not  cause checking such to give missing related errors.

This solves https://discourse.julialang.org/t/typeerror-non-boolean-missing-used-in-boolean-context-when-collecting-parameters/20593

Wether or not this is a wise thing that a user should do or not,
the behavour with this PR is more sensible.

### DIFF
--- a/src/tracker/Tracker.jl
+++ b/src/tracker/Tracker.jl
@@ -16,13 +16,17 @@ grad(x) = grad(tracker(x))
 grad(::Nothing) = nothing
 data(x) = x
 
+abstract type ComputationSentinel end
+struct NoComputation <: ComputationSentinel end
+struct FreedComputation <: ComputationSentinel end
+
 struct Call{F,As<:Tuple}
   func::F
   args::As
 end
 
 Call(f::F, args::T) where {F,T} = Call{F,T}(f, args)
-Call() = Call(nothing, ())
+Call() = Call(NoComputation(), ())
 
 # When deserialising, the object_id changes
 a::Call == b::Call = a.func == b.func && a.args == b.args
@@ -40,7 +44,7 @@ mutable struct Tracked{T}
 end
 
 istracked(x::Tracked) = true
-isleaf(x::Tracked) = x.f == Call()
+isleaf(x::Tracked) = x.f isa Call{<:ComputationSentinel}
 grad(x::Tracked) = x.grad
 
 track(f::Call, x) = Tracked{typeof(x)}(f)
--- a/src/tracker/back.jl
+++ b/src/tracker/back.jl
@@ -1,3 +1,5 @@
+
+
 init_grad(x) = zero(x)
 zero_grad!(x) = zero(x)
 zero_grad!(x::AbstractArray) = (x .= 0)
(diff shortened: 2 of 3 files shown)


src/tracker/Tracker.jl
```diff
@@ -16,13 +16,17 @@ grad(x) = grad(tracker(x))
...
+struct FreedComputation <: ComputationSentinel end
+
 struct Call{F,As<:Tuple}
   func::F
   args::As
 end
 
 Call(f::F, args::T) where {F,T} = Call{F,T}(f, args)
-Call() = Call(nothing, ())
+Call() = Call(NoComputation(), ())
```
staticfloat: I guess the only thing we need to be careful about now is that behavior of actually calling `Call(nothing, ())` has changed.  @MikeInnes that's not a concern, is it?

src/tracker/back.jl
```diff
@@ -1,3 +1,5 @@
+
+
```
staticfloat: kill the extra whitespace please

test/tracker.jl
```diff
@@ -329,6 +329,15 @@ end
   @test gs[W] == dW
 end
 
+@testset "Params after back!" begin
+  w = param(1)
+  model(z) = w*z
+  params_before = params(model)
+  Flux.back!(model, 1.0; once=true)
+  @test params(model) == params_before()
+end
```
staticfloat: Nice, but let's make sure that we're not putting in a test for a function that is about to disappear.  @MikeInnes do we need to test that `Flux.back!()` still works?

oxinabox: We can always delete the tests when we have the function.
Writing code that changes behavour, and does not have tests to prove it
is not a good  thing.

Plenty of other tests have to be deleted if `back!` goes away

src/tracker/Tracker.jl
oxinabox: Call is  not exported, is it?
And there are no instances of it called as `Call(nothing,   ())` in the  current  code  base , AFAICT. (If there were some  tests would have failed, I hope)

staticfloat: I don't think there are, but I'm just worried that if someone puts a `nothing` just sitting alone in their model that [things like this](https://github.com/FluxML/Flux.jl/blob/e438587aefb3fd71d4bedd29354b7d2487658352/src/tracker/Tracker.jl#L54-L57) won't generate a `Call(nothing, ())` in their AD graph.  Anyway, never hurts to ask the Mike, he did write this after all.

src/tracker/Tracker.jl
```diff
@@ -40,7 +44,7 @@ mutable struct Tracked{T}
 end
 
 istracked(x::Tracked) = true
-isleaf(x::Tracked) = x.f == Call()
```
MikeInnes: Wouldn't just changing this to `===` fix the bug?

oxinabox: My belief is that
It would stop the error, 
but it would give wrong results, since it would not return any of the params that had back'ed recently, since those would change type.

MikeInnes: I think it would be fine, since `isleaf` types don't have their call changed to `missing` anyway. Worth giving it a try.

ToucheSir:
Methinks this could be closed or moved to Tracker?

PR #612, opened 2019-02-09
Fixes OneHotMatrix/Vector GPU Performance

dhairyagandhi96:
Added tests in conjunction with changes made to the behaviour of OneHotVector/Matrix
cc @MikeInnes @KristofferC


test/cuda/cuda.jl
```diff
@@ -38,6 +38,13 @@ Flux.back!(sum(l))
 
 end
 
+@testset "onecold gpu" begin
+  CuArrays.allowscalar(false)
```
MikeInnes: This shouldn't be necessary as part of the test (we call it at the beginning of this file I think).

src/onehot.jl
```diff
@@ -22,6 +24,22 @@ Base.getindex(xs::OneHotMatrix, i::Integer, j::Integer) = xs.data[j][i]
...
+Base.getindex(xs::Flux.OneHotMatrix, j::Base.UnitRange, i::Int) = xs.data[i][j]
+
+Base.getindex(xs::OneHotMatrix, ::Colon, ::Colon) = xs
+Base.getindex(xs::OneHotMatrix, i::Integer, ::Colon) = map(x -> x[i], xs.data)
+
+# handle special case when we want the whole column
+function Base.getindex(xs::Flux.OneHotMatrix{T}, ot::Union{Base.Slice, Base.OneTo}, i::Int) where {T<:AbstractArray}
+  res = similar(xs, size(xs, 1), 1)
+  if length(ot) == size(xs, 1)
+    res = xs[:,i]
```
MikeInnes: Why do we need this branch? Are there any cases where they aren't equivalent?

dhairyagandhi96: W/O this branch - `136.165 ms (50001 allocations: 1.99 MiB)`
```Julia
julia> A = Flux.onehotbatch(1:300, 1:10000) |> gpu;

julia> d = 1
1

julia> a = Base.Slice(axes(A, d))
Base.Slice(Base.OneTo(10000))

julia> A[a, 5]
10000-element Array{Bool,1}:
 false
 false
 false
...
```

With - `15.930 μs (7 allocations: 10.16 KiB)`
```Julia
julia> A = Flux.onehotbatch(1:300, 1:10000) |> gpu;

julia> a = Base.Slice(axes(A, d))
Base.Slice(Base.OneTo(10000))

julia> A[a, 5]
10000-element Flux.OneHotVector:
 false
 false
 false
...
```

Performance and avoiding the allocation of the vector, basically.

test/cuda/cuda.jl
dhairyagandhi96: Right, I'll remove it

src/onehot.jl
MikeInnes: Is that also true on CPU? Is the slowdown due to scalar indexing? It seems like this might need to be something that's fixed at the CuArrays level rather than being special cased here.

dhairyagandhi96: I can remove it, if that behaviour is expected and should be maintained.

dhairyagandhi96: Scalar indexing happens when we try to get a column out of the `.data` field from `OneHotMatrix` currently, which does affect performance.

MikeInnes: It's not so much about whether the behaviour is expected as where the bug should be filed. If it can be fixed in CuArrays instead then it should be. It's still not clear to me whether or not that's the case, but I'll take a closer look at the code.

src/onehot.jl
```diff
@@ -22,6 +24,22 @@ Base.getindex(xs::OneHotMatrix, i::Integer, j::Integer) = xs.data[j][i]
 Base.getindex(xs::OneHotMatrix, ::Colon, i::Integer) = xs.data[i]
 Base.getindex(xs::OneHotMatrix, ::Colon, i::AbstractArray) = OneHotMatrix(xs.height, xs.data[i])
 
+Base.getindex(xs::Flux.OneHotMatrix, j::Base.UnitRange, i::Int) = xs.data[i][j]
```
MikeInnes: Perhaps just change the above definition to
```julia
Base.getindex(xs::OneHotMatrix, i::Union{Integer,AbstractVector}, j::Integer)
```

src/onehot.jl
```diff
@@ -22,6 +24,22 @@ Base.getindex(xs::OneHotMatrix, i::Integer, j::Integer) = xs.data[j][i]
...
+
+Base.getindex(xs::OneHotMatrix, ::Colon, ::Colon) = xs
```
MikeInnes: I think this definition is already handled above. Probably best to leave it; even though it's faster to avoid the allocation, it's more correct to return an independent array.

test/onehot.jl
```diff
@@ -15,5 +15,4 @@ end
 @testset "onehotbatch indexing" begin
   y = Flux.onehotbatch(ones(3), 1:10)
   @test y[:,1] isa Flux.OneHotVector
-  @test y[:,:] isa Flux.OneHotMatrix
```
MikeInnes: These tests should both still pass, no?

dhairyagandhi96: This goes into scalar indexing and allocates the whole array. I had that function to avoid it..

MikeInnes:
My old comment got deleted from the diff, so just to reiterate; there's a line
```julia
Base.getindex(xs::OneHotMatrix, i::Integer, j::Integer) = xs.data[j][i]
```

I think if `i::Integer` is changed to `i::Union{AbstractArray,Integer}` then we won't need the extra `getindex` special cases here.

dhairyagandhi96:
I am trying to figure out a nice way to do the reduction in the case of an `AbstractArray`, such that I can avoid allocating intermediaries.`mapreducedim!` currently returns an LLVM IR error for me, complaining about a call to `jl_box_float32`.

test/onehot.jl
dhairyagandhi96: Worth it to have something like `getindex(xs::OneHotMatrix, ::Colon, ::Colon) = deepcopy(xs)`

MikeInnes: Best to avoid `deepcopy` -- perhaps `copy(xs.data)` and reconstruct.

src/onehot.jl
```diff
@@ -18,9 +20,12 @@ end
 
 Base.size(xs::OneHotMatrix) = (Int64(xs.height),length(xs.data))
 
-Base.getindex(xs::OneHotMatrix, i::Integer, j::Integer) = xs.data[j][i]
+Base.getindex(xs::OneHotMatrix, i::Union{Integer, AbstractVector}, j::Integer) = xs.data[j][i]
 Base.getindex(xs::OneHotMatrix, ::Colon, i::Integer) = xs.data[i]
 Base.getindex(xs::OneHotMatrix, ::Colon, i::AbstractArray) = OneHotMatrix(xs.height, xs.data[i])
+# Base.getindex(xs::OneHotMatrix, ::Colon, ::Colon) = OneHotMatrix(xs.height, copy(xs.data))
```
MikeInnes: Is this meant to be commented out?

dhairyagandhi96: Ah, I found that the second I pushed, fixed with 2952bcd

dhairyagandhi96: Thanks, Mike!

dhairyagandhi96:
bors try

dhairyagandhi96:
Hmm.. bors seems to have given up, but the tests have finished https://gitlab.com/JuliaGPU/Flux.jl/pipelines/58592993

MikeInnes:
I guess the internal error is just the usual spurious one? If so feel free to merge.

dhairyagandhi96:
Those messages indeed are the ones CuArrays has been showing. Will fix the merge conflicts and merge.

Issue #613, opened 2019-02-10
generic_matmul! hit in `back!` because type-promotion in activation function


oxinabox:
Sometimes `generic_matmul!` is hit in `back!`
For examopole adding a `leak` too unit can be done 
by writing an activation function like

```
    leaky_relu6(x) = 0.01x + clamp(x, 0, 6)
```

And this is well and good, of `x` is a `Float64`.
But if `x` is a `Float32` this will trigger a type-promotion.
Which is bad, because the user almost certainly did not intend the type promotion.
But worse,
it means rather than hitting fast BLAS, we fall back to slow `generic_matmul!`.

Here is a MWE:

```
function flux_model()
    return Chain(
#        Dense(1280, 64, x->0.1f0x),    # Uncomment one of these lines
#        Dense(1280, 64, x->0.1e0x),   # Uncomment one these lines
        Dense(64, 1),
    )
end

function demo_flux()
    mdl = flux_model()
    features = rand(Float32, (1280, 1000))
    
    Flux.train!(
        params(mdl),
        [(features,)],
        Flux.ADAM()
    ) do xs
        sum(mdl(xs))
    end
end
```

### Time if it has to promote: `@time demo_flux()`
0.143774 seconds (607 allocations: 19.635 MiB)

### Time normally: `@time demo_flux()`
0.016475 seconds (568 allocations: 13.218 MiB, 47.67% gc time)

That is a 10x time diifference, and it scales up as your matrix sizes scale up.

KristofferC:
Isn't this expected though? Why aren't you using the eltype of the input to determine the type of the float constants?

oxinabox:
Depends what you mean by "expected".
It follows the rules of the language, yes.
But the fact that a mistake like this can trash your performance is not great.

As a rule when looking for code that might be causing a slow down one doesn't immediately go looking for constants not type matching.
This is fully type stable, it doesn't allocate etc etc.
Even looking at the profiler output it took @staticfloat and I a while to work this out.

It is of course obvious in retrospect,
but not in from inspection of the code base that this is causing a order of magnitude slow down.

We could certainly consider giving a surpressable warning if the return type of the activation does not match it's inputs.
(Or even just if it switched between floating point types).

Or we could do other things.

oxinabox:
If nothing else we can have a "performance tips" page, saying to be careful of the types of literals in your activation functions.

[...]

PR #615, opened 2019-02-11
Create performance tips docs section

oxinabox:
In part in response to #613,
but more generally I think good to have.


docs/make.jl
```diff
@@ -19,6 +19,7 @@ makedocs(modules=[Flux, NNlib],
                   "One-Hot Encoding" => "data/onehot.md",
                   "GPU Support" => "gpu.md",
                   "Saving & Loading" => "saving.md",
+                  "Performance Tips" => "performance_tips.md",
```
MikeInnes: Nit, I admit, but maybe just `performance.md`?

docs/src/performance_tips.md
```diff
@@ -0,0 +1,78 @@
+# Performance Tips
+
+All the usual [Julia performance tips apply](https://docs.julialang.org/en/v1/manual/performance-tips/).
+As always [profiling your code](https://docs.julialang.org/en/v1/manual/profile/#Profiling-1) is generally a useful way of finding bottlenecks.
+Below follow some Flux specific tips/reminders.
+
+## Don't use more precision than you need.
+
+Flux works great with all kinds of number types.
+But often you do not need to be working with say `Float64` (let along `BigFloat`).
```
MikeInnes:
```suggestion
But often you do not need to be working with say `Float64` (let alone `BigFloat`).
```
Though `BigFloat` references might be confusing for those not steeped in Julia / numerical stuff.

docs/src/performance_tips.md
```diff
@@ -0,0 +1,78 @@
...
+Switching to `Float32` can give you a significant speed up,
+not because the operations are faster, but because the memory usage is halved.
+Which means allocations occur much faster.
+And you use less memory.
+
+
+
+## Make sure your custom activation functions preserve the type of their inputs
+Not only should your activation functions be [type-stable](https://docs.julialang.org/en/v1/manual/performance-tips/#Write-%22type-stable%22-functions-1),
+they should also preserve the type of there inputs.
```
MikeInnes:
```suggestion
they should also preserve the type of their inputs.
```

docs/src/performance_tips.md
```diff
@@ -0,0 +1,78 @@
...
+While one could change your activation function (e.g. to use `0.01f0*x`) to avoid this when ever your inputs change,
+the idiomatic (and safe way) is to use `oftype`.
+
+```
+    leaky_tanh(x) = oftype(x/1, 0.01) + tanh(x)
+```
+
+
+
+## Evaluate batches as Matrix's of features, rather than sequences of Vector features
```
MikeInnes: matrices or matrixes

oxinabox:
CI failure is an unrelated network issue with downloading some test data.
Lets merge this?

docs/src/performance.md
```diff
@@ -0,0 +1,75 @@
...
+All the usual [Julia performance tips apply](https://docs.julialang.org/en/v1/manual/performance-tips/).
+As always [profiling your code](https://docs.julialang.org/en/v1/manual/profile/#Profiling-1) is generally a useful way of finding bottlenecks.
+Below follow some Flux specific tips/reminders.
+
+## Don't use more precision than you need.
+
+Flux works great with all kinds of number types.
+But often you do not need to be working with say `Float64` (let alone `BigFloat`).
+Switching to `Float32` can give you a significant speed up,
+not because the operations are faster, but because the memory usage is halved.
```
KristofferC: Seems relevant to mention `Float32` on GPU here. Also, operations do tend to be faster since you can fit more numbers in a SIMD lane at a given size.

docs/src/performance.md
```diff
@@ -0,0 +1,75 @@
...
+And you use less memory.
+
+
+## Make sure your custom activation functions preserve the type of their inputs
+Not only should your activation functions be [type-stable](https://docs.julialang.org/en/v1/manual/performance-tips/#Write-%22type-stable%22-functions-1),
+they should also preserve the type of their inputs.
+For example:
+
+```
+    my_tanh(x) = Float64(tanh(x))
```
KristofferC: This seems a bit artificial, perhaps something like `tanh(x) + 2.0` or `5.0 * tanh(x)` etc.

docs/src/performance.md
```diff
@@ -0,0 +1,75 @@
...
+function loss_total(xs::AbstractVector{<:Vector}, ys::AbstractVector{<:Vector})
+    sum(zip(xs, ys)) do (x,y_target)
+        y_pred = model(x) #  evaluate the model
+        return loss(y_pred, y_target)
+    end
+end
+```
+
+It is much faster to concatenate them into a matrix,
+as this will hit BLAS matrix-matrix multiplication, which is much faster than the equivelent sequence of matrix-vector multiplications.
```
KristofferC:
```suggestion
as this will hit BLAS matrix-matrix multiplication, which is much faster than the equivalent sequence of matrix-vector multiplications.
```

docs/src/performance.md
```diff
@@ -0,0 +1,75 @@
...
+Even though this means allocating new memory to store them continously.
```
KristofferC:
```suggestion
Even though this means allocating new memory to store them continuously.
```

docs/src/performance.md
oxinabox: I wanted something very obvious, for the first example, the one below is less obvious.

docs/src/performance.md
oxinabox: Feel free to use the github suggestion feature, or to PR after it is merged.
I know little of GPU so someone else is better to write it.

ViralBShah:
Merge?

MikeInnes:
Great stuff and sorely needed, thanks a lot @oxinabox.

docs/src/performance.md
```diff
@@ -0,0 +1,76 @@
...
+not because the operations are faster, but because the memory usage is halved.
+Which means allocations occur much faster.
+And you use less memory.
+
+
+## Make sure your custom activation functions preserve the type of their inputs
+Not only should your activation functions be [type-stable](https://docs.julialang.org/en/v1/manual/performance-tips/#Write-%22type-stable%22-functions-1),
+they should also preserve the type of their inputs.
+
+A very artificial example using an activatioon function like
```
KristofferC: "activatioon"

docs/src/performance.md
```diff
@@ -0,0 +1,76 @@
...
+
+Flux works great with all kinds of number types.
+But often you do not need to be working with say `Float64` (let alone `BigFloat`).
+Switching to `Float32` can give you a significant speed up,
+not because the operations are faster, but because the memory usage is halved.
+Which means allocations occur much faster.
+And you use less memory.
+
+
+## Make sure your custom activation functions preserve the type of their inputs
```
KristofferC: Add empty line

docs/src/performance.md
KristofferC: You could just remove the "not because the operations are faster, but because the memory usage is halved." part?

docs/src/performance.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+will result in performance on `Float32` input orders of magnitude slower than the normal `tanh` would,
+because it results in having to use slow mixed type multiplication in the dense layers.
+
+Which means if you change your data say from `Float64` to `Float32` (which should give a speedup: see above),
+you will see a large slow-down
+
+This can occur sneakily, because you can cause type-promotion by interacting with a numeric literals.
+E.g. the following will have run into the same problem as above:
```
KristofferC:
```suggestion
E.g. the following will run into the same problem as above:
```

docs/src/performance.md
```diff
@@ -0,0 +1,76 @@
...
+
+```
+    leaky_tanh(x) = 0.01x + tanh(x)
+```
+
+While one could change your activation function (e.g. to use `0.01f0x`) to avoid this when ever your inputs change,
```
KristofferC:
```suggestion
While one could change the activation function (e.g. to use `0.01f0x`) to avoid this when ever your inputs change,
```

oxinabox:
@KristofferC  ah noo.
Seconds too late.
Ah well, an easy PR to fix it?

Issue #618, opened 2019-02-11
Method overwrite warning when precompiling


KristofferC:
https://travis-ci.org/FluxML/Flux.jl/jobs/491652217#L476

These two methods:

https://github.com/FluxML/Flux.jl/blob/f17a5acd2bdf1c41c7574d1319522ac4a1ce6218/src/layers/stateless.jl#L47-L56

indeed look like they both match for `normalise(::AbstractArray)`

MikeInnes:
Yeah, we should just remove the `=1` so that `dims` isn't optional.

PR #619, opened 2019-02-11
Remove dims=1 from normalise

pshashk:
Fixes https://github.com/FluxML/Flux.jl/issues/618


MikeInnes:
We don't actually need to remove it everywhere -- just from the deprecated version of the function, so that it doesn't overlap with the non-deprecated version.

MikeInnes:
Thanks a lot!

PR #622, opened 2019-02-14
Implement missing conversions

MikeInnes:
That I was too lazy to fix at the time. Need to add some tests though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ os:
   # - osx
 
 julia:
-  - 1.0
+  - 1.1
   - nightly
 
 matrix:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -6,7 +6,7 @@ os:
   # - osx
 
 julia:
-  - 1.0
```
KristofferC: Flux claims to support `1.0` so should probably still keep testing on `1.0`.

MikeInnes: Any thoughts on dropping 1.0 support?

The issue is only a small annoying method ambiguity so it's easy enough to fix, anyway.

nickrobinson251: Please keep supporting (and testing on) the long-term support release of Julia -- it'd be hard to argue for using Flux in Production if that's not the case. (And thanks for your work on Flux!)

bhvieira:
Move to Tracker?

Issue #624, opened 2019-02-15
Error while running BatchNorm on OSX CPU


Sklan:
The last test case fails on OSX CPU
```julia 
BatchNorm: Test Failed at /Users/sklan/JuliaProjects/Flux.jl/test/layers/normalisation.jl:108
  Expression: #= /Users/sklan/JuliaProjects/Flux.jl/test/layers/normalisation.jl:108 =# @allocated(m(x)) < 100000000
   Evaluated: 3588496016 < 100000000
Stacktrace:
 [1] top-level scope at /Users/sklan/JuliaProjects/Flux.jl/test/layers/normalisation.jl:108
 [2] top-level scope at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.1/Test/src/Test.jl:1083
 [3] top-level scope at /Users/sklan/JuliaProjects/Flux.jl/test/layers/normalisation.jl:33
Test Summary: | Pass  Fail  Total
BatchNorm     |   12     1     13
ERROR: LoadError: Some tests did not pass: 12 passed, 1 failed, 0 errored, 0 broken.
```

MikeInnes:
cc @KristofferC, ref #586. Looks like that performance fix isn't working there but I can't think why that would change.

KristofferC:
Pff... I have no idea either. @Sklan could you report on what you get running

```
using Flux
bn = BatchNorm(32);
testimg = randn(Float32, 416, 416, 32, 1);
@time bn(testimg)
@time bn(testimg)
```

Sklan:
This is the output.
```julia
Last login: Fri Feb 15 21:26:25 on ttys003
MBP:~ sklan$ exec '/Applications/Julia-1.1.app/Contents/Resources/julia/bin/julia'
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.1.0 (2019-01-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> using Flux

julia> bn = BatchNorm(32);

julia> testimg = randn(Float32, 416, 416, 32, 1);

julia> @time bn(testimg)
 14.939357 seconds (225.12 M allocations: 3.799 GiB, 2.11% gc time)
[...]

PR #633, opened 2019-02-20
Improving BatchNorm

Sklan:
Fixes 2 issues
1. ϵ is declared but never used.
2. Makes BatchNorm notation similar to that of the original paper with similar performance.

### Ubuntu 18.04 GPU K-80

Test
```julia
bn = gpu(BatchNorm(32))
testimg = gpu(randn(Float32, 416, 416, 32, 1))
@time gpu(bn(testimg))
@time gpu(bn(testimg))
```

Old Code
```julia
3.327136 seconds (3.25 M allocations: 167.726 MiB, 1.83% gc time)
0.001245 seconds (155 allocations: 5.391 KiB)
```

New Code
```julia
3.542283 seconds (3.46 M allocations: 178.192 MiB, 2.74% gc time)
0.001185 seconds (149 allocations: 5.281 KiB)
```

### OSX CPU

Test
```julia
using Flux
bn = BatchNorm(32)
testimg = randn(Float32, 416, 416, 32, 1)
@time bn(testimg)
@time bn(testimg)
```

Old Code
```julia
14.939357 seconds (225.12 M allocations: 3.799 GiB, 2.11% gc time)
12.857950 seconds (215.97 M allocations: 3.342 GiB, 1.32% gc time)
```
New Code
```julia
2.701686 seconds (8.37 M allocations: 487.830 MiB, 9.61% gc time)
0.077704 seconds (135 allocations: 63.381 MiB, 24.40% gc time)
```

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -133,7 +133,7 @@ function (BN::BatchNorm)(x)
     # update moving mean/std
     mtm = data(convert(T, BN.momentum))
     BN.μ = (1 - mtm) .* BN.μ .+ mtm .* reshape(data(μ), :)
-    BN.σ² = (1 - mtm) .* BN.σ² .+ mtm .* reshape(data(σ²), :) .* m ./ (m - 1)
+    BN.σ² = (1 - mtm) .* BN.σ² .+ (mtm * m / (m - 1)) .* reshape(data(σ²), :)
   end
 
   let λ = BN.λ
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -113,34 +113,32 @@ BatchNorm(chs::Integer, λ = identity;
 function (BN::BatchNorm)(x)
   size(x, ndims(x)-1) == length(BN.β) ||
     error("BatchNorm expected $(length(BN.β)) channels, got $(size(x, ndims(x)-1))")
-  γ, β = BN.γ, BN.β
   dims = length(size(x))
   channels = size(x, dims-1)
   affine_shape = ones(Int, dims)
   affine_shape[end-1] = channels
   m = prod(size(x)[1:end-2]) * size(x)[end]
-
+  γ = reshape(BN.γ, affine_shape...)
+  β = reshape(BN.β, affine_shape...)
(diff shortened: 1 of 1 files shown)


dhairyagandhi96:
Ref: #624 
Thanks for the PR :)

KristofferC:
Locally I get

```
julia> @time bn(testimg);
  0.032135 seconds (170 allocations: 63.381 MiB, 15.82% gc time)
```

for old code on cpu.

src/layers/normalise.jl
```diff
@@ -113,34 +113,32 @@ BatchNorm(chs::Integer, λ = identity;
...
     axes = [1:dims-2; dims] # axes to reduce along (all but channels axis)
     μ = mean(x, dims = axes)
     σ² = sum((x .- μ) .^ 2, dims = axes) ./ m
-
+    ϵ = data(convert(T, BN.ϵ))
     # update moving mean/std
     mtm = data(convert(T, BN.momentum))
     BN.μ = (1 - mtm) .* BN.μ .+ mtm .* reshape(data(μ), :)
-    BN.σ² = ((1 - mtm) .* BN.σ² .+ mtm .* reshape(data(σ²), :) .* m ./ (m - 1))
+    BN.σ² = (1 - mtm) .* BN.σ² .+ mtm .* reshape(data(σ²), :) .* m ./ (m - 1)
```
dhpollack: You can change `.* m ./ (m - 1)` to `.* (m / (m - 1))`, since m is a scalar.  This saves some allocations on the initial run and 1 allocation thereafter.  You could also move mtm into the parathesis since it's also a scalar, but it makes the code less readable.

staticfloat:
This seems fine to me, but I'm not sure how it would have such a performance difference.  @Sklan what version are you testing when you say "old"?

Sklan:
@staticfloat I am using Pkg.dev that clones the master branch.
The performance difference is only on the mac cpu and I can't figure out why this works either.

MikeInnes:
GPU tests pass locally, so I think we can take this. Thanks a lot @Sklan!

PR #634, opened 2019-02-20
instance normalization

dhpollack:
While playing with Flux, I implemented [Instance Normalization](https://arxiv.org/abs/1607.08022), which is popular in style transfer.   I've copied the functionality of batch norm.  Thought I'd submit a pull request in case you were interested in adding it to the library.

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -203,9 +203,9 @@ InstanceNorm(chs::Integer, λ = identity;
   InstanceNorm(λ, param(initβ(chs)), param(initγ(chs)),
             zeros(chs), ones(chs), ϵ, momentum, true)
 
-function (IN::InstanceNorm)(x)
-  size(x, ndims(x)-1) == length(IN.β) ||
-    error("InstanceNorm expected $(length(IN.β)) channels, got $(size(x, ndims(x)-1))")
+function (in::InstanceNorm)(x)
+  size(x, ndims(x)-1) == length(in.β) ||
+    error("InstanceNorm expected $(length(in.β)) channels, got $(size(x, ndims(x)-1))")
   ndims(x) > 2 ||
     error("InstanceNorm requires at least 3 dimensions. With 2 dimensions an array of zeros would be returned")
   # these are repeated later on depending on the batch size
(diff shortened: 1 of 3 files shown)


test/layers/normalisation.jl
```diff
@@ -104,3 +104,84 @@ end
...
+  end
+
+  let m = InstanceNorm(2), sizes = (2, 4, 1, 2, 3),
+      x = param(reshape(collect(1:prod(sizes)), sizes))
+    y = reshape(permutedims(x, [3, 1, 2, 4, 5]), :, 2, 3)
+    y = reshape(m(y), sizes...)
+    @test m(x) == y
+  end
+
+  let m = BatchNorm(32), x = randn(Float32, 416, 416, 32, 1);
```
staticfloat: Is this supposed to be an `InstanceNorm`?

src/layers/normalise.jl
```diff
@@ -157,3 +157,102 @@ function Base.show(io::IO, l::BatchNorm)
...
+
+    ϵ = data(convert(T, IN.ϵ))
+    axes = 1:dims-2 # axes to reduce along (all but channels and batch size axes)
+    μ = mean(x, dims = axes)
+    σ² = mean((x .- μ) .^ 2, dims = axes)
+
+    # update moving mean/std
+    mtm = data(convert(T, IN.momentum))
+    IN.μ = reshape(mean(repeat((1 - mtm) .* IN.μ, outer=[1, bs]) .+ mtm .* reshape(data(μ), (c, bs)), dims = 2), :)
+    IN.σ² = reshape(mean((repeat((1 - mtm) .* IN.σ², outer=[1, bs]) .+ reshape(data(σ²), (c, bs)) .* (mtm * m / (m - 1))), dims = 2), :)
```
staticfloat: If this is being applied to a 1-spatial-dimension input (e.g. a `W x C x N` image) or a 3-spatial dimension input (e.g. a `W x H x D x C x N` image) will the `dims=2` still hold?  I think it will because `μ` will always be a `C x N` vector, but I just want to make sure.  It might be worthwhile to add some tests to ensure that the shapes of running 3 and 5-rank tensors through have the right shape.  :)

dhpollack: Yes, `μ` will be `C x N`, but you only want to store `C`, because you don't want the parameter to be dependent on the batch size, so we get the mean over the second dimension which is `N`.  But I'll create a test for this with a rank-N tensor.

test/layers/normalisation.jl
dhpollack: you are correct.  I'll make this change.

src/layers/normalise.jl
dhpollack: @staticfloat just to be a bit more clear, the reshape of `μ` is from `1 x ... x C x N` -> `C x N` and earlier I always expand `IN.μ` to `1 x ... x C x N` with the `expand_inst` function.

src/layers/normalise.jl
```diff
@@ -157,3 +157,103 @@ function Base.show(io::IO, l::BatchNorm)
...
+  momentum::N
+  active::Bool
+end
+
+InstanceNorm(chs::Integer, λ = identity;
+          initβ = (i) -> zeros(Float32, i), initγ = (i) -> ones(Float32, i), ϵ = 1f-5, momentum = 0.1f0) =
+  InstanceNorm(λ, param(initβ(chs)), param(initγ(chs)),
+            zeros(chs), ones(chs), ϵ, momentum, true)
+
+function (IN::InstanceNorm)(x)
```
MikeInnes: I would prefer lowercase here for a value.

dhpollack: Lowercase IN? I copied the form that was used in batchnorm. I can change it here but then it'll be non-standard. I feel like this and the batchnorm pull should be approved then another opened to change variable names what such.

MikeInnes:
@staticfloat are you happy with the changes? Needs a quick update for recent changes but otherwise GPU tests pass and I think we're good to go.

src/layers/normalise.jl
MikeInnes: Consistency doesn't really matter here since it's not exposed in any way. The `BN` for batchnorm must've slipped through at some point.

dhpollack: Ok. I'm away from the computer but I can make this change tomorrow. But I do think you should consider keeping the uppercase here and then using pull #658 to make all the variable name changes at once to maintain consistency.

staticfloat: Let's use lowercase `in` here, then I can fix up `BN` -> `bn` later.  #658 is unrelated.

Once that is done, and it is rebased on top of the latest `master`, I think we're good to go.

dhpollack: alright, changed  :+1:

staticfloat:
LGTM!  Thanks @dhpollack!

MikeInnes:
Thanks @dhpollack and @staticfloat!

@dhpollack I just added a new NEWS.md, new policy is that you have to show off your work there :) Can you add an entry to it?

PR #647, opened 2019-02-27
Add MaxOut layer

oxinabox:


docs/src/models/layers.md
```diff
@@ -28,6 +30,13 @@ GRU
 Flux.Recur
 ```
 
+## Hipster Layers
+These are marginally more obscure layers that you probably haven't heard of.
```
aviks: I know you mean it as a joke, but this can come out as condescending.

oxinabox: I put it in knowing that in code reveiw someone would complain and hopefully come up with something better.

KristofferC: Just "Other"?

src/layers/basic.jl
```diff
@@ -117,10 +125,47 @@ function Base.show(io::IO, l::Diagonal)
...
-# Base's matmul is so slow that it's worth the extra conversion to hit BLAS
-(a::Dense{<:Any,W})(x::AbstractArray{T}) where {T <: Union{Float32,Float64}, W <: AbstractArray{T}} =
-  invoke(a, Tuple{AbstractArray}, x)
 
-(a::Dense{<:Any,W})(x::AbstractArray{<:Real}) where {T <: Union{Float32,Float64}, W <: AbstractArray{T}} =
-  a(T.(x))
+"""
+    MaxOut(over)
+
+MaxOut is a neural network layer, which has a number of internal layers,
```
KristofferC:
```suggestion
`MaxOut` is a neural network layer, which has a number of internal layers,
```

src/layers/basic.jl
```diff
@@ -117,10 +125,47 @@ function Base.show(io::IO, l::Diagonal)
...
+    over::FS
+end
+
+"""
+    MaxOut(f, n_alts, args...; kwargs...)
+
+Constructs a MaxOut layer over `n_alts` instances of  the layer given  by `f`.
+All other arguements (`args` & `kwargs`) are passed to the constructor `f`.
+
+For example the followeExample usage
```
KristofferC: Something seems off here

docs/src/models/layers.md
oxinabox: Jules Merckx   came up with "esoteric layers",
which I like better than Other, it is like Misc
Anything could be in Other where as Esoteric says it is not the run of the mill layer.

src/layers/basic.jl
```diff
@@ -117,10 +125,47 @@ function Base.show(io::IO, l::Diagonal)
...
+In Proceedings of the 30th International Conference on International Conference on Machine Learning - Volume 28 (ICML'13),
+Sanjoy Dasgupta and David McAllester (Eds.), Vol. 28. JMLR.org III-1319-III-1327.
+https://arxiv.org/pdf/1302.4389.pdf
+"""
+struct MaxOut{FS<:Tuple}
+    over::FS
+end
+
+"""
+    MaxOut(f, n_alts, args...; kwargs...)
```
MikeInnes: Why not just pass in those layers, like `MaxOut(a, b, c)`?

src/layers/basic.jl
```diff
@@ -117,10 +125,47 @@ function Base.show(io::IO, l::Diagonal)
...
+    MaxOut(Dense, 4, insize, outsize)
+```
+"""
+function MaxOut(f, n_alts, args...; kwargs...)
+  over = Tuple(f(args...; kwargs...) for _ in 1:n_alts)
+  return MaxOut(over)
+end
+
+function (mo::MaxOut)(input::AbstractArray)
+    mapreduce(f -> f(input), (acc, out) -> max.(acc, out), mo.over)
```
MikeInnes: I'd find `max.(map(f -> f(input), mo.over)...)` clearer.

oxinabox: hmmm...
Will the compiler unroll the `map` and the splat?

It should be possible to do so since everything is tuples.

docs/src/models/layers.md
KristofferC: Do you have a list of those layers that will classify as "esoteric"? Seems like such a notation will depend on the person and the application. This seciton will likely effectively just work as an "Other" section so might as well call it that.

src/layers/basic.jl
oxinabox: Mapreduce is noticably faster.
over 2x faster.

```
julia> foo(fs,x) = max.(map(f->f(x), fs)...)

julia> bar(fs,x) = mapreduce(f->f(x), (a,b)->max.(a,b), fs)
bar (generic function with 1 method)

julia> @btime foo((x->2x, x->0.5x, x->3x, x->4x, x->6x), $[1:1000.0;]);
  9.469 μs (9 allocations: 47.72 KiB)

julia> @btime bar((x->2x, x->0.5x, x->3x, x->4x, x->6x), $[1:1000.0;]);
  4.989 μs (11 allocations: 71.47 KiB)
```

src/layers/basic.jl
oxinabox: In general, using different layers is not a thing.
Also using layers that are not dense/linear is not a thing.
The paper is just about identically structured dense linear layers.

Allowing them to be different, and allowing them to be any kind of
felt like a natural generalisation, and it made testing easier.


Here the line in my code using this
```
 MaxOut(Dense, 4, insize, hlsize)
```

Here is the Keras code this is replacing
```
MaxoutDense(hlsize, input_dim=input_units, nb_feature=4, bias=True
```

What are your proposals?

docs/src/models/layers.md
oxinabox: I want a way to be clear that these are not domain specific ones.
In contrast to RNN or CNN.

Perhaps "Other General Purpose Layers" ?

src/layers/basic.jl
MikeInnes: Makes sense to optimise for the common case of course. But in that case I think `
MaxOut(_ -> Dense(insize, outsize), 4)` gives a clearer sense that this is doing something with 4 dense layers (it looks like `ntuple`, which is basically what it is).

src/layers/basic.jl
MikeInnes: Just to record what I mentioned on slack, tracked arrays might have slightly different perf here (broadcast is much heavier and SIMD probably won't kick in either way), so it'd be worth giving that case a quick test.

oxinabox: Even for TrackedArrays

```
julia> data = Flux.TrackedArray(collect(1:1000));

julia> using BenchmarkTools

julia> foo(fs,x) = max.(map(f->f(x), fs)...)

julia> bar(fs,x) = mapreduce(f->f(x), (a,b)->max.(a,b), fs)
bar (generic function with 1 method)

julia> @btime bar((x->2x, x->0.5x, x->3x, x->4x, x->6x), $data);
  7.891 μs (65 allocations: 73.08 KiB)

julia> @btime foo((x->2x, x->0.5x, x->3x, x->4x, x->6x), $data);
  12.359 μs (56 allocations: 49.13 KiB)
```

We hafve resolved in slack that we will use mapreduce
and if anyone complains it is too slow rewrite it using a generated function.

src/layers/basic.jl
oxinabox: This is growing on me.

Another option is:
`MaxOut(Dense(insize, outsize), 4)`,
and then we call `deepcopy` on the first argument to get 4 versions of it.
Not sure I like that.

oxinabox: Ok, I've sold myself on your idea.
It seperates which thigns are arguments of the inner layer, and which things are arguments of the outer layer.
Going forward we sould make this a convention if it comes up again.

nickrobinson251:
Very minor: the name `Maxout` would be consistent with the paper (and with `Dropout`), as opposed to `MaxOut` :)

oxinabox:
Done

docs/src/models/layers.md
```diff
@@ -28,6 +32,25 @@ GRU
 Flux.Recur
 ```
 
+## Other General Purpose Layers
+These are marginally more obscure than the Basic Layers.
+But incontrast to the layers described in the other sections are not readily grouped around a paparticular purpose (e.g. CNNs or RNNs).
```
MikeInnes:
```suggestion
But in contrast to the layers described in the other sections are not readily grouped around a paparticular purpose (e.g. CNNs or RNNs).
```

docs/src/models/layers.md
```diff
@@ -28,6 +32,25 @@ GRU
...
+
+```@docs
+MaxOut
```
nickrobinson251:
```suggestion
Maxout
```

src/Flux.jl
```diff
@@ -6,8 +6,10 @@ using Base: tail
 using MacroTools, Juno, Requires, Reexport, Statistics, Random
 using MacroTools: @forward
 
-export Chain, Dense, RNN, LSTM, GRU, Conv, ConvTranspose, MaxPool, MeanPool,
-       DepthwiseConv, Dropout, LayerNorm, BatchNorm,
+export Chain, Dense, MaxOut,
```
nickrobinson251:
```suggestion
export Chain, Dense, Maxout,
```

docs/src/models/layers.md
nickrobinson251:
```suggestion
But in contrast to the layers described in the other sections are not readily grouped around a particular purpose (e.g. CNNs or RNNs).
```

src/layers/basic.jl
```diff
@@ -117,10 +125,47 @@ function Base.show(io::IO, l::Diagonal)
...
+will construct a `Maxout` layer over 4 dense linear layers,
+each identical in structure (784 inputs, 128 outputs).
+```julia
+    insize = 784
+    outsie = 128
+    Maxout(Dense, 4, insize, outsize)
+```
+"""
+function Maxout(f, n_alts, args...; kwargs...)
+  over = Tuple(f(args...; kwargs...) for _ in 1:n_alts)
```
MikeInnes: The signature needs an update now that we're supporting the closure-ised version instead of passing args through.

Also I think it'd make sense to use `ntuple(f, n_alts)` to construct the tuple. It's clean and it means you can write `_ -> Dense(...)`, which saves a character :) but the `n` might be useful in some cases also.

oxinabox: It _could_ be useful.
I worry that we are overengineering generality in to this that is not present in the paper.

If we wanted to make it easier/less complex to just do the thing in the paper (and the thing Karas lets you do)
could add.

`MaxoutDense(insize, outsize, n_alts)  = Maxout(_->Dense(insize, outsize),  n_alts)`

oxinabox: I've thought about it more.
I can't imagine passing in the number would every be useful.
Maxout is defined as the max over structurally identical internal layers.

Most things you might thing about doing with it are

 - going to break it `Maxout(n->Dense(128, 128n), 4)`
 - pointless `Maxout(n->n*Dense(128, 128), 4)`
 - Weird `Maxout(n->Dense(128,128; σ=(relu, identity, tanh, sigmoid)[n])`
 - Complex enough that they should be written seperately: 
```
Maxout(n->Chain(Dense(128,128n, σ=sigmoid), Dense(128n, 128)), 4)
```
vs
``
inner_layers = ntuple(n->Chain(Dense(128,128n, σ=sigmoid), Dense(128n, 128)),  4)
Maxout(inner_layers)
```

I think it is better to be able to say if you want to construct differing internal layers,
then you should do so explictly and pass them in as a tuple

This extraflexibility  I think makes the code harder to read (even if it can be 1 character shorter).
it is readily apparent what a `()->f(z)` does, and how I can change/tweak it.
where as the purpose of the argument in `_->f(z)` is not clear, and it isn't obvious what the `n` is doing in  `n->g(z,n)`

**Comment by oxinabox** [comment]
Any further thoughts?

**Comment by MikeInnes** [comment]
I think we are in general agreement on how to tweak the API; I think it'll be good to go once that's done.

**Comment by oxinabox** [comment]
API tweaks are done.
It matches as discussed in https://github.com/FluxML/Flux.jl/pull/647#discussion_r264437446

File: src/layers/basic.jl [hunk 10]
```diff
@@ -117,10 +125,48 @@ function Base.show(io::IO, l::Diagonal)
...
+In Proceedings of the 30th International Conference on International Conference on Machine Learning - Volume 28 (ICML'13),
+Sanjoy Dasgupta and David McAllester (Eds.), Vol. 28. JMLR.org III-1319-III-1327.
+https://arxiv.org/pdf/1302.4389.pdf
+"""
+struct Maxout{FS<:Tuple}
+    over::FS
+end
+
+"""
+    Maxout(f, n_alts, args...; kwargs...)
```
MikeInnes: If we're supporting the closure-based API we can remove the splatting from the signature.

I think we may as well document the signature as `Maxout(() -> ...)` if that's always the intended use.

oxinabox: Ah yes, mmy bad i forget to remove the arguments.

oxinabox: it is `Maxout(f, n_alts)`
I feel like `Maxout(()->..., n_alts)` is a bit less clear than having `f` then being able to reference `f` in the docstring below.

The docstring does incluide an example
```
Maxout(()->Dense(insize, outsize), 4)
```

MikeInnes:
Alright, I think this is pretty perfect. Thanks a lot @oxinabox!

Issue #630, opened 2019-02-15
train! doesn't work with Trackedreal


merckxiaan:
A model with a Trackedreal parameter crashes during the backward pass with the `train!` function
`type TrackedReal has no field grad`

alhirzel:
I believe I am seeing this too, here is one example:

```julia
using Flux: mse, param, params
using Flux.Tracker: update!, TrackedReal
using Flux.Optimise
using Random: seed!
using Plots

struct LinRegParams
	slope     :: TrackedReal{Float64}
	intercept :: TrackedReal{Float64}
end

LinReg(m, b) = LinRegParams(param(m), param(b)) # instantiator
(m::LinRegParams)(x) = m.slope*x .+ m.intercept # evaulator

m = LinReg(1, 0) # m is callable and has attached data

loss(x, y) = mse(m(x), y)
ps = params(m.slope, m.intercept)
data = [(0,3), (1,5)]  # should get us to m=2, b=3

train!(loss, ps, data, Descent(0.1))
@show m(0)
```

alhirzel:
I think this will be addressed when #562 is merged.

kota7:
This is also related, perhaps. https://github.com/FluxML/Flux.jl/issues/214

PR #649, opened 2019-02-27
Solved the problem of train! with TrackedReal

thebhatman:
train! was giving an error with TrackedReal saying grad member not found. This PR fixes the issue.

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -8,8 +8,16 @@ end
 
 function _update_params!(opt, xs)
   for x in xs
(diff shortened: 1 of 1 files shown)


thebhatman:
#630

MikeInnes:
This PR needs updating. It would be best to make sure that `update!` for parameters is generic.

bhvieira:
Flux doesn't depend on Tracker anymore, so I guess unless someone fork it to keep Tracker this PR is not compatible anymore with the library

PR #650, opened 2019-02-28
Add Parallel layer

swiesend:
I've improved the implementation of my Parallel layer with a MapReduce approach. I hope this is interesting for stacked models. I started this work to enable bidirectional LSTMs, but the current implementation is not only for recurrent models.

The approach goes like this: The user adds some layers and defines custom mappings and inverses for the input. After the mapping-phase comes the reduce-phase, which defaults to concatenation of the parallel layers:

```julia
mutable struct Parallel
    layers::Vector
    map::Vector{Function}
    inv::Vector{Function}
    reduce::Function
end

function (p::Parallel)(xs))
    # y = map(x) |> f |> map^-1
    apply(l) = p.inv[l](p.layers[l](p.map[l](xs)))
    
    Z = map(l -> apply(l), eachindex(p.layers))

    p.reduce(Z)
end
```

For the full implementation see:

https://github.com/swiesend/Flux.jl/blob/parallel/src/layers/parallel.jl

Here I've added run some example models to showcase the usage of the `Parallel` layer and some other convinience layers, like `Bi` or `BiLSTM`:

https://github.com/swiesend/Flux.jl/blob/parallel/test/layers/parallel.jl

For an example of a bidirectional LSTM see:

https://github.com/swiesend/Flux.jl/blob/ce33f606fded250ec2524bd13c47d75245e2de52/test/layers/parallel.jl#L29

I've also added a new peephole LSTM layer (`PLSTM`) in `recurrent.jl`:
https://github.com/swiesend/Flux.jl/blob/ce33f606fded250ec2524bd13c47d75245e2de52/src/layers/recurrent.jl#L202

How to run the examples:
```sh
git clone git@github.com:swiesend/Flux.jl.git .
cd Flux.jl
git checkout parallel
julia
```

```julia-repl
julia> using Pkg; Pkg.activate("."); Pkg.instantiate(); include("test/layers/parallel.jl")
```

## Issues

* The introduced `Base.reverse()` function on Flux data-structures lacks unit tests

* There are some issues, like GPU support, which I could not solve. I ran into a `ReadOnlyMemoryError()`.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -6,7 +6,7 @@ using Base: tail
 using MacroTools, Juno, Requires, Reexport, Statistics, Random
 using MacroTools: @forward
 
-export Chain, Dense, RNN, LSTM, GRU, Conv, ConvTranspose, MaxPool, MeanPool,
+export Chain, Dense, RNN, LSTM, GRU, Parallel, Bi, BiLSTM, PLSTM, BiPLSTM, Conv, ConvTranspose, MaxPool, MeanPool,
        DepthwiseConv, Dropout, LayerNorm, BatchNorm,
        params, mapleaves, cpu, gpu, f32, f64
 
@@ -32,6 +32,7 @@ include("layers/stateless.jl")
 include("layers/basic.jl")
 include("layers/conv.jl")
 include("layers/recurrent.jl")
+include("layers/parallel.jl")
 include("layers/normalise.jl")
 
 include("data/Data.jl")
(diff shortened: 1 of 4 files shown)


MikeInnes:
This looks generally good, but would be easier to review if it were a bit split up. For example we could probably take the peephole LSTM without many changes, so that may as well be its own PR.

Depending on your familiarity with git, it's probably easiest to do this by splitting a clean branch off `master` and making the changes in a new PR.

PR #651, opened 2019-02-28
Refactor training loop

MikeInnes:
It turns out a lot of people look at this, show it needs to do a better job of showing the actual Flux APIs. Addresses parts of #607.


oxinabox:
Yesss this is much better

PR #655, opened 2019-03-01
Added support for Float64 for DepthwiseConv

thebhatman:
DepthwiseConv was giving errors for Float64. This fixes the issue.

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -165,6 +165,12 @@ function Base.show(io::IO, l::DepthwiseConv)
   print(io, ")")
 end
 
+(a::DepthwiseConv{<:Any,<:Any,W})(x::AbstractArray{T}) where {T <: Union{Float32,Float64}, W <: AbstractArray{T}} =
+  invoke(a, Tuple{AbstractArray}, x)
+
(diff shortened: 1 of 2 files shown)


Tokazama:
Just out of curiosity, does Flux usually limit to Float32? I know it's best for GPUs but I thought any floats we're fine.

thebhatman:
I don't think Flux limits to Float32. Float32 is the default for convolution layers because it can store a decimal number with good enough precision and computation with Float32 is much faster than Float64. I guess Float64 is used only when you want to store decimals with very high precision, which is generally less important when compared to computation time in a neural network.

avik-pal:
@Tokazama Flux supports both Float64 and Float32. Only recently we changed the CPU default to Float32. Apparently, DepthwiseConv was left out while transitioning.

dhairyagandhi96:
Can we add a couple tests just to catch these cases?

thebhatman:
@dhairyagandhi96 I have added tests for Float64 in DepthwiseConv.

dhairyagandhi96:
Bors ping

thebhatman:
I am not sure why the tests failed. The details say it failed to clone from github.

dhairyagandhi96:
Bors r+

PR #656, opened 2019-03-03
Added AlphaDropout which is used in SNNs.

thebhatman:


MikeInnes:
Seems generally reasonable, but the implementation is currently dropping gradients. You'll need to avoid in-place broadcast to get this working without `param` and `collect`.

thebhatman:
Thank You @MikeInnes. I have made the necessary changes so that gradients are not lost. I am no longer using collect and param.

src/layers/normalise.jl
```diff
@@ -43,6 +43,35 @@ end
...
 
+"""
+    AlphaDropout(p)
+A dropout layer. It is used in Self-Normalizing Neural Networks. 
+(https://papers.nips.cc/paper/6698-self-normalizing-neural-networks.pdf)
+The AlphaDropout layer ensures that mean and variance of activations remains the same as before.
+"""
+mutable struct AlphaDropout{F}
+	p::F
+  active::Bool
```
CarloLucibello: indentation

src/layers/normalise.jl
```diff
@@ -43,6 +43,35 @@ end
...
+function AlphaDropout(p)
+  @assert 0 ≤ p ≤ 1
+  AlphaDropout{typeof(p)}(p,true)
+end
+
+function (a::AlphaDropout)(x)
+  a.active || return x
+  α = -1.75813631
+  noise = randn(Float64, size(x.data))
+  x.data .= x.data .* (noise .> (1 - a.p)) + α .* (noise .<= (1 - a.p))
```
CarloLucibello: Do not assign to x.data, otherwise you  . Just do
```julia
 x = x* (noise .> (1 - a.p)) + α .* (noise .<= (1 - a.p))
```
or even better
```julia
d = rand(size(x)) .> a.p
 x = @. x * d + α * !d
```

src/layers/normalise.jl
```diff
@@ -43,6 +43,35 @@ end
...
+
+function AlphaDropout(p)
+  @assert 0 ≤ p ≤ 1
+  AlphaDropout{typeof(p)}(p,true)
+end
+
+function (a::AlphaDropout)(x)
+  a.active || return x
+  α = -1.75813631
+  noise = randn(Float64, size(x.data))
```
CarloLucibello: `randn` gives normally distributed numbers, maybe you want `rand` here

src/layers/normalise.jl
```diff
@@ -43,6 +43,35 @@ end
...
+end
+
+function AlphaDropout(p)
+  @assert 0 ≤ p ≤ 1
+  AlphaDropout{typeof(p)}(p,true)
+end
+
+function (a::AlphaDropout)(x)
+  a.active || return x
+  α = -1.75813631
```
CarloLucibello: this is going to change the return value to Float64 irrespective of `x`. α should be initialized to the same element type of `x`

src/layers/normalise.jl
thebhatman: I think randn is more suitable here, as we are required to have a uniform normally distributed output.

thebhatman:
Thank you @CarloLucibello. I have made the changes suggested by you. Just curious, I wanted to know what would have been the problem if we assigned to x.data.

MikeInnes:
This looks generally reasonable to me. Would like a quick look over from @staticfloat that it makes sense.

> Just curious, I wanted to know what would have been the problem if we assigned to x.data.

Any time you work with `.data` you drop gradients; the reason we wrap the data in a `TrackedArray` is exactly so that we can record what happens to it and get gradients.

MikeInnes:
Also this needs some numerical tests, e.g. comparing hand-coded output from another implementation, or similar.

src/layers/normalise.jl
```diff
@@ -43,6 +43,35 @@ end
...
+function AlphaDropout(p)
+  @assert 0 ≤ p ≤ 1
+  AlphaDropout{typeof(p)}(p,true)
+end
+
+function (a::AlphaDropout)(x)
+  a.active || return x
+  α = eltype(x)(-1.75813631)
+  noise = randn(eltype(x), size(x))
+  x = @. x*(noise .> (1 - a.p)) + α .* (noise .<= (1 - a.p))
```
staticfloat: I think if you use `@.` you don't need the `.>` and `.*` and `.<=` operators; either use `@.` to get them all, or manually write them all as dotted operators.

src/layers/normalise.jl
```diff
@@ -43,6 +43,35 @@ end
...
+end
+
+function AlphaDropout(p)
+  @assert 0 ≤ p ≤ 1
+  AlphaDropout{typeof(p)}(p,true)
+end
+
+function (a::AlphaDropout)(x)
+  a.active || return x
+  α = eltype(x)(-1.75813631)
```
staticfloat: You should use the same `α` parameter as given by the `selu()` function in NNlib, see the definition here: https://github.com/FluxML/NNlib.jl/blob/d07ac0bfd3c71c3a29bc9c22becbba19227bbeb5/src/activation.jl#L100-L104

src/layers/normalise.jl
```diff
@@ -43,6 +43,35 @@ end
...
+The AlphaDropout layer ensures that mean and variance of activations remains the same as before.
+"""
+mutable struct AlphaDropout{F}
+  p::F
+  active::Bool
+end
+
+function AlphaDropout(p)
+  @assert 0 ≤ p ≤ 1
+  AlphaDropout{typeof(p)}(p,true)
```
staticfloat: You shouldn't need `{typeof(p)}` here; it should be able to determine this from the `p` parameter:

```
julia> struct Foo{F}
       p::F
       active::Bool
       end

julia> Foo(1.0, true)
Foo{Float64}(1.0, true)
```

src/layers/normalise.jl
```diff
@@ -43,6 +43,35 @@ end
...
+end
+
+function (a::AlphaDropout)(x)
+  a.active || return x
+  α = eltype(x)(-1.75813631)
+  noise = randn(eltype(x), size(x))
+  x = @. x*(noise .> (1 - a.p)) + α .* (noise .<= (1 - a.p))
+  A = (a.p + a.p * (1 - a.p) * α ^ 2)^0.5
+  B = -A * α * (1 - a.p)
+  x = @. A .* x .+ B
```
staticfloat: Same thing here; this `@.` is not doing anything, because you have already dotted your operators.

thebhatman:
I realized randn(), which I am using to initialize noise, is not giving a standard normal distribution. Due to which, the mean and variance is slightly different from the mean and variance of input. This can be solved by making use of Normal() which is available in Distributions package. But Flux doesn't have a dependency on Distributions. What would be the best way to get a standard normal distribution? Shall I add Distributions among Flux Dependencies?

thebhatman:
Thank you @staticfloat. I have used the value of lambda and alpha as defined in NNlib.jl. I have removed the redundant dot operations. I will be adding test cases. But before that, I wanted to solve the problem of randn() not giving a standard normal distribution. Can I use Distributions package and use Normal() so that I get a distribution with mean = 0 and variance = 1?

staticfloat:
> I realized randn(), which I am using to initialize noise, is not giving a standard normal distribution. 

What do you mean by this?  Why do you think that `randn()` is not a standard normal distribution?  Note that the `Normal` distribution from `Distributions.jl` [uses `randn()` to generate its random numbers internally](https://github.com/JuliaStats/Distributions.jl/blob/817bd83326f9d562ec88c0e782e60d25d64862ad/src/univariate/continuous/normal.jl#L90).

thebhatman:
`julia> x = randn(2,2)
2×2 Array{Float64,2}:
 0.332405  -0.813627
 0.446137  -0.644151

julia> using Statistics

julia> mean(x)
-0.16980890659280456

julia> std(x)
0.650925294719868

`

thebhatman:
I was just testing randn() and it doesn't seem to be giving a standard normal distribution.

MikeInnes:
That's just sampling error, the mean is only in the limit of a large number of samples.

staticfloat:
This is completely expected; the mean and standard deviation you compute on normally-distributed random numbers will not always be exactly zero and one.  That's not the way randomness works.  When you take `mean(x)`, you are calculating an _estimate_ of the true mean of the population that your four numbers were taken from.

Let's imagine we are measuring the height of people in a city.  Let's further assume that across the entire city, the heights form a normal distribution with a certain mean and standard deviation.  If I just take 4 random people from the city, the mean of their heights will probably not be the "true" mean; it will just be an estimate of it.  I may be unlucky and grab four people that are all on the very tall side of things.  The "error" of my mean estimate may be very high.

You can read [here](https://en.wikipedia.org/wiki/Standard_error) for more information on calculating the error bounds of your statistical estimates, but to convince yourself, try making a matrix that is much larger than `2x2`, say, `10000x10000`, and see what the statistics of that look like.  I guarantee they will be much closer to a mean of `0` and a variance of `1`, but they will also not be _exactly_ `0` and `1`.

thebhatman:
Oh! That was a nice example. Thank you for clearing my doubt.

src/layers/normalise.jl
```diff
@@ -43,6 +43,37 @@ end
...
+
+function AlphaDropout(p)
+  @assert 0 ≤ p ≤ 1
+  AlphaDropout(p,true)
+end
+
+function (a::AlphaDropout)(x)
+  a.active || return x
+  λ = 1.0507009873554804934193349852946
+  α = 1.6732632423543772848170429916717
```
staticfloat: I think it would be better to have `λ` and `α` be `oftype(eltype(x), ....)` so that we don't have to convert lower on down; that might help us generate code to run on platforms where, for instance, Float64 literals are not supported (that's not the case on CPUs, but for things like TPUs it's helpful to stay in the proper datatype as much as possible)

thebhatman: Shouldn't the arguments of oftype() be two values?

thebhatman:
The Travis CI build failed after I resolved the merge conflict, which was raised due to InstanceNorm.

src/layers/normalise.jl
thebhatman: Done.

staticfloat:
Thanks @thebhatman this looks pretty good!

MikeInnes:
Awesome, thanks a lot @thebhatman and @staticfloat for the review!

@thebhatman can you also add an entry to NEWS.md about this?

Issue #503, opened 2018-11-21
Provide aliases for unicode fields and keywords


donm:
I'm thinking of adding aliases for keyword arguments and structure fields that currently require using unicode characters to access.  Examples include the `σ` in the `Conv` field names, and almost all of the fields in `BatchNorm`:

```
julia> fieldnames(BatchNorm)
(:λ, :β, :γ, :μ, :σ, :ϵ, :momentum, :active)
```

I only have three data points so far, but when other people start to use Flux this has been something that bothers them.   

Questions:
- Is this something we should do?
- Would you prefer having ascii names added as aliases, or replacing the current instances of unicode with ascii names and then adding the current unicode names back as aliases?
- Should the ascii name simply spell out the unicode symbol, or describe meaning of the field/variable?  For example, should the alias for `σ` in `Conv` be `activation` or `sigma`?  (My preference would be `activation`.)

MikeInnes:
I generally agree that anything user-facing should not have unicode in it (and there are a few places we need to fix this). At the same time fields like this are generally considered private, and it's better form to provide accessor functions (e.g. `Flux.activation(layer)`).

On the flip side of *that*, it's probably not that reasonable to have an accessor function for every possible component of a layer, especially for things like batchnorm. If people want to access those fields (is this common?) then it'd make sense to rename them clearly as you suggest.

Open to discussion on this.

donm:
When you say user-facing interfaces should not have unicode, do you mean they shouldn't force the use of unicode, or shouldn't use it at all?

I'm not sure if accessing those types of fields is common, but it's definitely common for me and others I work with.  Being able to reach in and quickly manipulate weights and other state in the middle of training is a huge benefit of having the whole library written in Julia.  

I agree that the number of accessor functions could be unwieldy, especially because many of the fields would need both getters and setters.

MikeInnes:
Right, no one should be forced to use unicode to build a model with Flux -- which sounds like it's the case for you.

We can change the fieldnames but should also add a deprecation for the current names. I'm open to PRs on this.

namanbiyani:
Hi I would like to work on this . please tell which field names should be changed in which files are they present?

MikeInnes:
If you want to start with batchnorm, that's in `src/layers/normalise.jl`.

CarloLucibello:
[...]

PR #658, opened 2019-03-04
Update normalise.jl

namanbiyani:
#503 Provided aliases for unicode fields in batchnorm and dropout

### DIFF
--- a/src/cuda/cudnn.jl
+++ b/src/cuda/cudnn.jl
@@ -194,7 +194,7 @@ end
 # Flux Interface
 
 (BN::Flux.BatchNorm)(x::Union{CuParam{T,2},CuParam{T,4},CuParam{T,5}}, cache = nothing) where T<:Union{Float32, Float64} =
-  batchnorm(BN.γ, BN.β, x, BN.μ, BN.σ², BN.momentum; cache = cache, alpha = 1, beta = 0, eps = BN.ϵ, training = BN.active)
+  batchnorm(BN.scale, BN.bias, x, BN.moving_mean, BN.moving_std, BN.momentum; cache = cache, alpha = 1, beta = 0, eps = BN.ϵ, training = BN.active)
 
 batchnorm(g::TrackedArray, b::TrackedArray, x::TrackedArray, running_mean::CuArray{T},
           running_var::CuArray{T}, momentum; kw...) where T<:Union{Float32, Float64} =
--- a/src/cuda/cudnn.jl
+++ b/src/cuda/cudnn.jl
@@ -194,7 +194,7 @@ end
 # Flux Interface
 
 (BN::Flux.BatchNorm)(x::Union{CuParam{T,2},CuParam{T,4},CuParam{T,5}}, cache = nothing) where T<:Union{Float32, Float64} =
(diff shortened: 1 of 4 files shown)


src/layers/normalise.jl
```diff
@@ -95,25 +95,25 @@ m = Chain(
...
+  BatchNorm(activation_function, param(initβ(chs)), param(initγ(chs)),
+            zeros(chs), ones(chs), epsilon, momentum, true)
 
 function (BN::BatchNorm)(x)
-  size(x, ndims(x)-1) == length(BN.β) ||
-    error("BatchNorm expected $(length(BN.β)) channels, got $(size(x, ndims(x)-1))")
-  γ, β = BN.γ, BN.β
+  size(x, ndims(x)-1) == length(BN.bias) ||
+    error("BatchNorm expected $(length(BN.bias)) channels, got $(size(x, ndims(x)-1))")
+  scale, bias = BN.scale, BN.bias
```
avik-pal: IMO it's cleaner to use unicode characters internally.

src/cuda/cudnn.jl
```diff
@@ -194,7 +194,7 @@ end
 # Flux Interface
 
 (BN::Flux.BatchNorm)(x::Union{CuParam{T,2},CuParam{T,4},CuParam{T,5}}, cache = nothing) where T<:Union{Float32, Float64} =
-  batchnorm(BN.γ, BN.β, x, BN.μ, BN.σ², BN.momentum; cache = cache, alpha = 1, beta = 0, eps = BN.ϵ, training = BN.active)
+  batchnorm(BN.scale, BN.bias, x, BN.moving_mean, BN.moving_std, BN.momentum; cache = cache, alpha = 1, beta = 0, eps = BN.ϵ, training = BN.active)
```
CarloLucibello: `moving_std` should really be `moving_var`

namanbiyani:
Please tell why the checks are getting failed . I am unable to understand this

test/layers/normalisation.jl
```diff
@@ -78,7 +78,7 @@ end
     @test !m.active
 
     y = m(x).data
-    @test isapprox(y, data(sigmoid.((x .- m.μ) ./ sqrt.(m.σ² .+ m.ϵ))), atol = 1.0e-7)
+    @test isapprox(y, data(sigmoid.((x .- m.running_mean) ./ sqrt.(m.moving_var .+ m.eps))), atol = 1.0e-7)
```
avik-pal: Should be 
```suggestion
    @test isapprox(y, data(sigmoid.((x .- m.running_mean) ./ sqrt.(m.running_var .+ m.eps))), atol = 1.0e-7)
```

namanbiyani:
Please review @MikeInnes

dhpollack:
@namanbiyani  Won't this cause problems with both #633 and #634 which both seem ready to go?  I actually agree the names should be unicode, but approving this and before those will be a bit of a headache.

MikeInnes:
Note that as discussed in #503, we'll need to provide deprecation warnings for anything we change here.

namanbiyani:
What should I do then @MikeInnes ?

MikeInnes:
You can provide a deprecation by overloading `getproperty`, so that if someone tries to use the old names it'll work but with a warning to use the new name.

Issue #657, opened 2019-03-04
`collect` drops gradients


MikeInnes:
```julia
julia> xs = param([1, 2, 3])
Tracked 3-element Array{Float64,1}:
 1.0
 2.0
 3.0

julia> collect(xs)
3-element Array{Float64,1}:
 1.0
 2.0
 3.0
```

Sklan:
So, I looked into this. I figured this is a method call ambiguity between ```Base.collect``` and ```Tracker.collect```.  
```julia
julia> xs = param([1, 2, 3])
Tracked 3-element Array{Float64,1}:
 1.0
 2.0
 3.0

julia> collect(xs)
3-element Array{Float64,1}:
 1.0
 2.0
 3.0

julia> Tracker.collect(xs)
Tracked 3-element Array{Float64,1}:
 1.0
 2.0
 3.0
```
I tried to fix it by explicitly exporting collect and creating a type annotation
from
```julia 
collect(xs)
```
to
```julia
collect(xs::TrackedArray)
```
But turns out its a bit messy. Since there exists a function in array.jl in Base,
```julia 
collect(A::AbstractArray)
```
and given that 
```julia
julia> xs isa TrackedArray
true

julia> xs isa AbstractArray
true
```
So, before opening a pull request I wanted some input on how to correctly handle this.

MikeInnes:
`Tracker.collect` is actually a separate function which behaves slightly differently. `Base.collect` should just be the identity function on TrackedArrays. You can fix that by remove `collect` from [this line](https://github.com/FluxML/Flux.jl/blob/d6cf116a740c2879ff04de95174d1528b7e43cd6/src/tracker/lib/array.jl#L76) and redefining it.

gxyd:
Since Flux now uses Zygote, does this issue still makes sense?

PR #661, opened 2019-03-05
Remove collect

sklan:
Solves #657
```julia 
julia> xs = param([1, 2, 3])
Tracked 3-element Array{Float64,1}:
 1.0
 2.0
 3.0

julia> collect(xs)
3-element Array{Flux.Tracker.TrackedReal{Float64},1}:
 1.0
 2.0
 3.0
```

### DIFF
--- a/src/tracker/lib/array.jl
+++ b/src/tracker/lib/array.jl
@@ -73,7 +73,7 @@ end
 
 # Fallthrough methods
 
-for f in :[Base.size, Base.ndims, Base.collect].args
+for f in :[Base.size, Base.ndims].args
   @eval @inline $f(x::TrackedArray, a...) = $f(data(x), a...)
 end
 
--- a/src/tracker/lib/array.jl
+++ b/src/tracker/lib/array.jl
@@ -71,9 +71,9 @@ function update!(x::TrackedArray, Δ)
   return x
 end
 
-collect(x::TrackedArray) = x
-collect(x::TrackedVector) = x
-collect(x::TrackedMatrix) = x
+Base.collect(x::TrackedArray) = x
+Base.collect(x::TrackedVector) = x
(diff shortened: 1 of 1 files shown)


MikeInnes:
Can you add a definition for `collect` that just returns `x`? Keeping this as a TrackedArray is more efficient.

Sklan:
Yeah, sure. I'll do that.

Sklan:
@MikeInnes It still keeps calling the ```Base.collect```
```julia
julia> xs = param([1, 2, 3])
Tracked 3-element Array{Float64,1}:
 1.0
 2.0
 3.0

julia> collect(xs)
3-element Array{Flux.Tracker.TrackedReal{Float64},1}:
 1.0
 2.0
 3.0

julia> @which collect(xs)
collect(A::AbstractArray) in Base at array.jl:546
```
I am confused. Am I missing something?

MikeInnes:
You need to define `Base.collect` (otherwise it creates a new function, `Flux.collect`).

src/tracker/lib/array.jl
```diff
@@ -71,9 +71,9 @@ function update!(x::TrackedArray, Δ)
   return x
 end
 
-collect(x::TrackedArray) = x
-collect(x::TrackedVector) = x
-collect(x::TrackedMatrix) = x
+Base.collect(x::TrackedArray) = x
+Base.collect(x::TrackedVector) = x
```
MikeInnes: The `Vector` and `Matrix` definitions are redundant here, since they are covered by `TrackedArray`.

MikeInnes:
Unfortunately this got caught when we split out Tracker.jl. You should be able to easily PR the same branch against that repo, though.

sklan:
Okay, I'll add a PR in Tracker.jl.

PR #663, opened 2019-03-06
Add PLSTM and FCLSTM layer

swiesend:
a peephole LSTM (`PLSTM`) and a fully connected LSTM (`FCLSTM`).

Both implementations are straight forward, as much of the original LSTM implementation was reused. see: `src/layers/recurrent.jl`

There are some very basic unit test under: `test/layers/recurrent.jl`

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -198,3 +198,114 @@ See [this article](http://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
 """
 GRU(a...; ka...) = Recur(GRUCell(a...; ka...))
+
+# peephole LSTM
+
+mutable struct PeepholeLSTMCell{A,V}
+  Wx::A
+  Wc::A
+  b::V
+  h::V
+  c::V
+end
+
+function PeepholeLSTMCell(in::Integer, out::Integer;
+                          init = glorot_uniform)
+  cell = PeepholeLSTMCell(
+      param(init(out*4, in)),  # Wx
+      param(init(out*4, out)), # Wc
+      param(zeros(out*4)),     # b
+      param(init(out)),        # h
+      param(init(out)))        # c
+  cell.b.data[gate(out, 2)] .= 1
+  return cell
+end
+
+function (m::PeepholeLSTMCell)((h, c), x)
+  o = size(h, 1)
+  g = m.Wx*x .+ m.Wc*c .+ m.b
+  input = σ.(gate(g, o, 1))
+  forget = σ.(gate(g, o, 2))
+  cell = tanh.(gate(g, o, 3))
+  output = σ.(gate(g, o, 4))
+  c = forget .* c .+ input .* cell
+  h = output .* tanh.(c)
+  return (h, c), h
+end
+
+hidden(m::PeepholeLSTMCell) = (m.h, m.c)
+
+@treelike PeepholeLSTMCell
+
+Base.show(io::IO, l::PeepholeLSTMCell) =
+  print(io, "PeepholeLSTMCell(", size(l.Wx, 2), ", ", size(l.Wx, 1) ÷ 4, ")")
+
+"""
+    PLSTM(in::Integer, out::Integer)
+
+Peephole LSTM layer. Behaves like an LSTM, but instead of using the hidden state
+`h`, the LSTM cell gets direct access to the hidden cell state `c`.
+
+See the following articles for internals:
+
+* [Gers, F. A.; Schmidhuber, J. (2001). "LSTM Recurrent Networks Learn Simple Context Free and Context Sensitive Languages". IEEE Transactions on Neural Networks. 12 (6): 1333–1340. doi:10.1109/72.963769.](ftp://ftp.idsia.ch/pub/juergen/L-IEEE.pdf)
+* [Gers, Felix & Schraudolph, Nicol & Schmidhuber, Jürgen. (2002). "Learning Precise Timing with LSTM Recurrent Networks". Journal of Machine Learning Research. 3. 115-143.](http://www.jmlr.org/papers/volume3/gers02a/gers02a.pdf)
+"""
+PLSTM(a...; ka...) = Recur(PeepholeLSTMCell(a...; ka...))
+
+
+# fully connected LSTM
+
+mutable struct FullyConnectedLSTMCell{A,V}
+  Wx::A
+  Wh::A
+  Wc::A
+  b::V
+  h::V
+  c::V
+end
+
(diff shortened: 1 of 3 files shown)


swiesend:
Hey @avik-pal until know I totally ignored, that a LSTM needs special GPU code work with the CuArrays.

I think the proposed `PLSTM` is quite the same as the LSTM, but the `FCLSTM` would need more adoption for its additional weight matrix.

What can be reused and what would need to change to also work on the GPU?

avik-pal:
Hi @swiesend I haven't worked on the RNN Cuda/cudnn support for Flux (I simply dumped the contents of the cudnn.jl file to curnn.jl). I am afraid I can't help you in how to add the cuda support for these layers.

Maybe you can look up the contributors in the `cudnn.jl` file to find out who added the code there.

src/layers/recurrent.jl
```diff
@@ -198,3 +198,114 @@ See [this article](http://colah.github.io/posts/2015-08-Understanding-LSTMs/)
...
+
+Peephole LSTM layer. Behaves like an LSTM, but instead of using the hidden state
+`h`, the LSTM cell gets direct access to the hidden cell state `c`.
+
+See the following articles for internals:
+
+* [Gers, F. A.; Schmidhuber, J. (2001). "LSTM Recurrent Networks Learn Simple Context Free and Context Sensitive Languages". IEEE Transactions on Neural Networks. 12 (6): 1333–1340. doi:10.1109/72.963769.](ftp://ftp.idsia.ch/pub/juergen/L-IEEE.pdf)
+* [Gers, Felix & Schraudolph, Nicol & Schmidhuber, Jürgen. (2002). "Learning Precise Timing with LSTM Recurrent Networks". Journal of Machine Learning Research. 3. 115-143.](http://www.jmlr.org/papers/volume3/gers02a/gers02a.pdf)
+"""
+PLSTM(a...; ka...) = Recur(PeepholeLSTMCell(a...; ka...))
```
MikeInnes: I would prefer PeepholeLSTM here

MikeInnes:
The easiest way to get CUDA support, by far, is to simply re-use the existing LSTM layer inside the new ones. Is that possible? If not, we have to see if CUDA has an option to support these.

Will also need to be added to the docs, and a news item.

swiesend:
Reusing the `LSTM` layer could work for the `PeepholeLSTM`, but definitively not for the `FullyConnectedLSTM` as it carries one more weight matrix.

I general I was not very pleased when I saw the plumbing of the `LSTM` and CUDA, as it did not seem very maintainable or adaptable.

MikeInnes:
That's CUDNN for you. By design it's a special case and we have to treat it that way.

I should clarify that the CUDNN stuff is only a performance optimisation, rather than an absolute necessity for CUDA support. The normal CuArrays fallback will work, perhaps with a bit of a perf hit compared to CUDNN. So if you can reuse LSTM it's a good idea but not essential.

dbadrian:
Has there been any progress on these recently (wtr to the cuda side of things)? I am interested to see these merged, especially since I am intending to implement a ConvLSTM cell. As far as I can tell, I won't be able to make a pure mashup of an LSTM and conv operations, but also write a custom cell requiring custom cuda code - in this sense, I was wondering if there is something to built upon here, or gotta venture out on my own and try.

src/layers/recurrent.jl
```diff
@@ -198,3 +198,114 @@ See [this article](http://colah.github.io/posts/2015-08-Understanding-LSTMs/)
...
+  cell.b.data[gate(out, 2)] .= 1
+  return cell
+end
+
+function (m::FullyConnectedLSTMCell)((h, c), x)
+  o = size(h, 1)
+  g = m.Wx*x .+ m.Wh*h .+ m.Wc*c .+ m.b
+  input = σ.(gate(g, o, 1))
+  forget = σ.(gate(g, o, 2))
+  cell = tanh.(gate(g, o, 3))
```
dbadrian: This update of the cell state is not complete equivalent to the paper is it? In the paper (eq.9, inside the tanh), only the respective parts of Wx*x and Wh*h (=bias) are considered, here you additionally have Wc*c added?

Prolly wont stop it from learning, but what are the side effects?

src/layers/recurrent.jl
```diff
@@ -198,3 +198,114 @@ See [this article](http://colah.github.io/posts/2015-08-Understanding-LSTMs/)
...
+  output = σ.(gate(g, o, 4))
```
dbadrian: also the output o_t (eq.10) is using the new cell state c_t not c_{t-1}, isnt it?

swiesend:
Hey @dbadrian, yeah there seem to be some odd things here:

1. Yeah, `cell` (9) should probably be without access to `Wc`.

2. According to (1) and (2) we apply `Wc` conditionally on `o` (N) on the other three gates, right?!
So: `σ.(gate(o > 1 ? gc : g, o, $))` for `$ in [1,2,4]`. 

3. and according to (11), `cell` (9) should be used, but I kept it like in the `LSTMCell` implementation with `c`:
https://github.com/FluxML/Flux.jl/blob/8c467a71c0c1e6b2c4d903406b0e8c600c2103be/src/layers/recurrent.jl#L131-L141

So I suggest something like this
```suggestion
function (m::FullyConnectedLSTMCell)((h, c), x)
  o = size(h, 1)
  g = m.Wx*x .+ m.Wh*h .+ m.b
  gc = g .+ m.Wc*c
  input = σ.(gate(o > 1 ? gc : g, o, 1))
  forget = σ.(gate(o > 1 ? gc : g, o, 2))
  cell = tanh.(gate(g, o, 3))
  output = σ.(gate(o > 1 ? gc : g, o, 4))
  c = forget .* c .+ input .* cell
  h = output .* tanh.(c)
  return (h, c), h
end
```

Hey @MikeInnes:
* What is actually the source for the `LSTMCell`?
* And what is the best way to avoid conditionals here?

swiesend:
@dbadrian, sorry for the many updates on the [comment](https://github.com/FluxML/Flux.jl/pull/663#issuecomment-524048306). But I think its fine now.

src/layers/recurrent.jl
swiesend: > Prolly wont stop it from learning, but what are the side effects?

I think it depends on how to interpret the job of the `tanh` of the `cell`. But giving it even more access shouldn't hurt much, I guess.

dbadrian: Probably not, but then we should document it. It's not too uncommon for an implementation to differ from the paper, but it should be made clear.

And possibly, beforehand, make a comparison on at least some tests how the performance is? Maybe you already used it for some projects successfully?

CarloLucibello:
any interest in rebasing this? otherwise we should close

Issue #643, opened 2019-02-26
Train/test mode


MikeInnes:
Right now layers like `BatchNorm` and `Dropout` have a flag to put them in `train` or `test` mode. However, once Zygote lands (#628) we can do something much clevererer: we enable the regularisation only in a gradient context. Then it will automatically be on during the training loop and off at test time.

We can of course still have a manual override here (just set the `enabled` flag to `:auto` by default), but it's interesting to consider whether we even need this; I suspect we will but don't know of any explicit use cases for it.

Currently I think this aligns well with how I've seen people use these layers in practice, avoids some predictable mode-boilerplate, and gets rid of one more usage of `mapleaves`. However, it does make a fairly strong assumption about how these layers get used, so I'm on the lookout for cases where this might lead to counter-intuitive or unexpected behaviour, compared to the explicit approach.

jekbradbury:
I think I agree with this! The only case I've ever encountered where the layer mode diverges from the autodiff context is that batch norm can sometimes operate in a third mode where the running mean and variance are used for normalization and _also_ updated based on changing statistics of the data; this is sometimes used for online inference. I think that's sufficiently niche to ignore 🙂

CarloLucibello:
this has already landed

wsshin:
After reading this thread, it is not clear to me whether or not I need to use `trainmode!()` and `testmode!()`.  Can someone clarify?

Issue #637, opened 2019-02-22
New New Optimisers


MikeInnes:
If we go full steam ahead with more functional-style AD (#628), then we'll need to rework optimisers a little. I think at the core the update step will look very similar, if a bit more functional, something like:

```julia
apply(::Optimiser, x, dx, state = ...) -> (dx', state')
```

Then we have an `update` function which actually applies the gradient. In general, the `x`, `dx` and `state` can be structs and we'll rebuild `x` by recursing over it (somewhat like mapleaves, though we won't need to depend on `treelike` since we can just use reflection here).

I think this is all reasonably straightforward; the main wrinkle in my mind is how we enable in-place updates as an optimisation (since it's not ideal to have two copies of ResNet at once). I'm not aware of any great solution to this right now, so we might need to define an `ismutable` trait and declare it for types we care about.

staticfloat:
Can you spell out the difficulties regarding in-place updates?  I don't see it as obvious.

MikeInnes:
One difficulty is that mutability is not part of the array API; there isn't an automatic way to discover whether an array is mutable without trying it and checking for an error. So that's where we'll need an `ismutable` trait that effectively records a database of things we are allowed to mutate, and users with new array types will have to overload `Flux.ismutable` (which isn't the end of the world, just kind of ugly).

The second difficulty is working out the semantic issues around mutability. If you have a ref around an immutable model, presumably "do this update in place" actually means update the ref. What if the model actually has a mix of immutable and mutable arrays?

Then we need to figure out how to expose this choice as an API, while also sharing mutating and non-mutating code as much as possible so that plugging in your own types is easy (not having to do everything twice). And that both for "leaf types" that get updated themselves and containers that only update their contents.

It's all doable, but quite a lot fiddlier than it initially looks, and will probably take some time to work out well.

Roger-luo:
So come from slack, I think it would be quite useful if we could move the optimizers to a single package (even share code with other things like Optim, but move to a package first maybe). We are currently using the Flux optimizer in Yao, but Flux itself is a quite heavy dependency for just Optimizers.

ToucheSir:
Now that https://github.com/SciML/ArrayInterface.jl#ismutablex exists, at least part of this is in place. WRT avoiding the "two copies of resnet in memory" problem, however, is there a more convenient API than the explicit param-passing that jax, haiku and co. use?

Roger-luo:
[...]

Issue #628, opened 2019-02-15
Gradient Interface Design


MikeInnes:
We expose a pullback/vjp-based API for gradients (`y, back = forward(f, x); x̄ = back(ȳ)`), with `x̄ = gradient(f, x)` as simple syntax sugar of top of this. This interface is pretty awesome – `gradient` aligns nicely with the mathematical and intuitive notions of a derivative operator, it naturally expresses nested derivatives, and you can build pretty much any other AD-related functionality (checkpointing, forward mode, gradient hooks, etc) on top of pullbacks, without having to go into AD internals. So far I haven't come across anything that pullbacks *can't* do straightforwardly; in [one case](https://github.com/FluxML/Flux.jl/pull/607#issuecomment-461889413) the PyTorch-style `back!` may be slightly more convenient, but it's overall more cumbersome and requires more knowledge of internals.

However, a challenge of the "mathematical" gradient operator is that it's cumbersome to pass in all our parameter arrays explicitly (`gradient(resnet, W1, b1, W2, b2, ...)`). So we need to be able to handle taking gradients of large models without it being cumbersome. There are currently two ideas about how to do this: the structural approach and the implicit approach.

## Structural Gradients

The first approach (which Zygote will support whatever happens, and could be added to Flux) is to take the gradients w.r.t. some structure containing all of the parameters. The structure could be a dictionary or list, but it's usually convenient to combine the weight structure with the definition of the forward pass. This is effectively a closure, which we refer to as a "layer". Layers can contain other layers (we often call a compound layer a "model", but there's no fundamental difference). Taking a gradient looks like this:

```julia
m = Chain(Dense(10, 5, relu), Dense(5, 2), softmax)
x, y = ...
m̄ = gradient(m -> loss(m(x), y), m)
```

This looks pretty weird at first but makes a lot of sense once it clicks. One then carries out the update step `m .+= m̄`.

## Implicit Gradients

The implicit approach is what Flux supports natively, though it works in Zygote as well. In this case we ask for gradients of a shapeless set of parameters, which are implicitly used at some point during the forward pass. In this case we have something more like:

```julia
m = Chain(Dense(10, 5, relu), Dense(5, 2), softmax)
x, y = ...
θ = params(m)
θ̄ = gradient(() -> loss(m(x), y), θ)
```

`θ̄` is a dictionary from param to gradient. One then loops over the parameters, doing `p .+= θ̄[p]`.

## Do we need implicit parameters?

[...]

Issue #666, opened 2019-03-06
add `trainstep!`


oxinabox:
Following up  from https://github.com/FluxML/Flux.jl/pull/607

We should expose functionality,  that lets the user write a training loop,
while thinking only about  loss.
Rather than thinking about `gradient` and `update!`.
`loss` is a higher level concept than gradients.

Custom  traing loops are  important since many things do not coomfortably fit into the abstraction of  
`train!(args->loss(args...), data,  params, opt,callbacks)`.
The  `train!` functioin is good for things that comfortably fit supervised training,
and while it can do anything it  becomes increasing akward the further you are from  that.
At the other  end is writing a custom training  loop,  invoking `gradients` and `update`.
This is fully general and you can do  all kinds of things l like messing with  the gradients during the training loop.
But there is a middle ground,
where you can define the loss, but you have nothing to say about the gradients.

For  this I think we should have 

`train_step!(getloss, ps, opt)`, where `getloss` is a 0 arg closure returning the loss (and using the model).
This  wouldwould have pleasing symetry in name and arguments,
to `train!(loss, ps, data, opt)` where `loss` is a closure taking args as provided by iterating `data`.

This  would  be useful because you are not being required to use the abstraction of having `data`, but you have the  rest.

The implementation would be very simple, but I feel the abstraction away from `gradients` is worth it.

```
function train_step!(getloss, ps, opt)
    gs = gradient(getloss, ps)
    update!(opt, ps, gs)
end
```

This would go into the core of `train!`.
Replacing
https://github.com/FluxML/Flux.jl/blob/3a4c6274fadf1b94468da455a06f070b2cda6a64/src/optimise/train.jl#L71-L74
with 
```
trainstep!(x->loss(d...), ps, opt)
```

MikeInnes:
I'm on board. It will make sense to have this as the default interface for gradients if we go full force on #628, since the gradient object will be a little more complex than usual.

I think this should be written `step!(opt) do ...` (it's kind of method of the optimiser). It may as well also return the loss. 

Initially it will have to be `step!(opt, ps) do ...`, but we can remove the `ps` with #628 (and also reorder `train!` arguments to be more consistent with this).

@oxinabox up for sketching it out?

oxinabox:
Do you really think  the `train!` needs to have it's argument order changed too?
I like the consistency,
But it #628 will remove `ps`, 
then reorder args will mean just an extra round of depreciations for little gain.

[...]

PR #669, opened 2019-03-08
using Zygote

MikeInnes:
Otherwise known as "break all the things". This will be a huge change so I'm beginning to prepare now, even though Zygote is still a couple of months off from being really ready. **Do not try this at home** (yet) – this branch is eventually aimed at beta testers, but isn't even ready for that yet.

The idea is to break as little code as possible, which means supporting the current `Params` API; but I also want to start prototyping the nicer things discussed in #628 and other issues.

Blocking issues:

* [x] Get the tests passing.
* [x] Check tests on GPU.
* [x] Rewrite all the docs.
* [x] Cache invalidation (https://github.com/jrevels/Cassette.jl/issues/6).
* [x] Moving over adjoints (https://github.com/FluxML/Zygote.jl/issues/81).
* [x] General Zygote robustness.

Nice to have:

* [ ] Robust nested AD (may not be a blocker if one can still use Tracker with Flux).
* [x] Zygote support for modules / globals as discussed in #628, along with #637.
* [x] Better train/test mode as in #643.

If you're the kind of person who ignores triangular road signs, you can try this with

```julia
]add Flux#zygote Zygote#master
```


MikeInnes:
I initially had started doing #666 and #637 on this branch, but that's turning into a big project, so I've stripped the commits for now (preserved on [mji/step](https://github.com/FluxML/Flux.jl/tree/mji/step)). As excited as I am to get rid of `Params` I think the right move is to shelve that for now and focus on the core AD issues.

staticfloat:
In my own experiments trying to use `Zygote` with `Flux` (I just do `model = mapleaves(Flux.data, model)` first, then define my own Zygote-based update step:

```julia
zyg_update!(opt, model, updates::Nothing) = nothing
function zyg_update!(opt, model::AbstractArray, updates::AbstractArray)
    # Sub off to Flux's ADAM optimizer
    Δ = Flux.Optimise.update!(opt, model, updates)
    return model .-= Δ
end

function zyg_update!(opt, model, updates)
    if nfields(model) == 0
        return model
    end

    for field_idx in 1:nfields(model)
        zyg_update!(opt, getfield(model, field_idx), getfield(updates, field_idx))
    end
end
```

Things actually work fairly well, except BatchNorm freaks out, complaining about mutating arrays.  To work around this, I am using [my own `BatchNorm` implementation](https://gist.github.com/staticfloat/a509b1e1cb1fb556028779722c2531e6), re-architected to work with Zygote.  Not sure if that is the direction you want to go with this Mike, but it worked well for us on TPU.  I will note, anecdotally, that my convolution and batchnom-heavy workload (a convolutional autoencoder for large images) uses ~20% less memory with Zygote than with Tracker.

staticfloat:
I wanted to use this with the new NNlib overhaul, so I rebased this branch on top of `master`; I'm not certain I did everything right, but `sf/zygote_updated` contains my rebased version.  Mike, if you like it, you can just force-push it to `#zygote` and keep on working, or you can just do the rebase yourself.

test/cuda/cuda.jl
```diff
@@ -6,8 +6,6 @@ using Zygote
 
 @testset "CuArrays" begin
 
-CuArrays.allowscalar(false)
```
MikeInnes: This is there for a reason. Why not fix the failing test?

test/cuda/cudnn.jl
```diff
@@ -1,48 +1,47 @@
...
 
         @test cpu(data(cy)) ≈ data(y)
 
         g = rand(size(y)...)
-        Flux.back!(y, g)
-        Flux.back!(cy, gpu(g))
+        # Flux.back!(y, g)
+        # Flux.back!(cy, gpu(g))
 
-        @test m.γ.grad ≈ cpu(cm.γ.grad)
```
MikeInnes: @thebhatman you may already be aware, but any tests that use `back!` and `grad` like this will have to be changed to get a gradient using Zygote and test that.

test/layers/normalisation.jl
```diff
@@ -1,29 +1,28 @@
...
+
+trainmode(f, x...) = forward(f, x...)[1]
 
 @testset "Dropout" begin
   x = [1.,2.,3.]
-  @test x == testmode!(Dropout(0.1))(x)
-  @test x == Dropout(0)(x)
-  @test zero(x) == Dropout(1)(x)
+  @test x == Dropout(0.1)(x)
+  @test x == trainmode(Dropout(0), (x))
```
MikeInnes: Some redundant brackets have slipped in here

test/cuda/cuda.jl
```diff
@@ -14,7 +16,7 @@ cx = gpu(x)
 
 x = Flux.onehotbatch([1, 2, 3], 1:3)
 cx = gpu(x)
-@test cx isa Flux.OneHotMatrix && cx isa CuArray
+@test cx isa Flux.OneHotMatrix && cx.data isa CuArray
```
thebhatman: Can we do `cx.data` here? I think this would be an error.

dhairyagandhi96: Why would it be an error?

MikeInnes: `cx` is a OneHotMatrix, which is a wrapper around another array (called `data`). So yeah, it's fine, it's not the same as the `data` for a tracked array.

thebhatman: Ooh okay! My bad, I thought `.data` was referring to the Tracker field data.

dhairyagandhi96: Cheers!

test/cuda/cudnn.jl
```diff
@@ -14,13 +14,11 @@ trainmode(f, x...) = forward(f, x...)[1]
...
-        g = rand(size(y)...)
-        # Flux.back!(y, g)
-        # Flux.back!(cy, gpu(g))
+        g = gradient(()->sum(m(x)), params(m))
+        cg = gradient(()->sum(cm(cx), params(cm))
 
-        @test m.γ ≈ cpu(cm.γ)
-        @test m.β ≈ cpu(cm.β)
-        @test x ≈ cpu(x)
+        @test g.grads[m.γ] ≈ cpu(cg.grads[cm.γ])
```
MikeInnes: You should index `g` directly here, rather than accessing `.grads`

src/layers/normalise.jl
```diff
@@ -136,7 +136,7 @@ function (BN::BatchNorm)(x)
   dims = length(size(x))
   channels = size(x, dims-1)
   affine_shape = ntuple(i->i == ndims(x) - 1 ? size(x, i) : 1, ndims(x))
-  m = prod(size(x)[1:end-2]) * size(x)[end]
+  m = trunc(Int, prod(size(x))/channels)
```
MikeInnes: Use `÷` here.

thebhatman: That gives the same problem as `div`... I will just add `@nograd div` in zygote as you suggested..

MikeInnes: Yeah, that's fine. You can send a PR to Zygote and then update this code + the manifest.

thebhatman: So I change the manifest to depend on Zygote master, right? Do I just remove this line https://github.com/FluxML/Flux.jl/blob/b779d43aca84de06e0e9ff8618904d130eec2cbd/Manifest.toml#L344

MikeInnes: you shouldn't edit the manifest by hand. you can use the package manager to check the status (it's already on master, but and older commit), and just use `up` to get the latest master.

src/layers/normalise.jl
```diff
@@ -136,7 +136,7 @@ function (BN::BatchNorm)(x)
   dims = length(size(x))
   channels = size(x, dims-1)
   affine_shape = ntuple(i->i == ndims(x) - 1 ? size(x, i) : 1, ndims(x))
-  m = trunc(Int, prod(size(x))/channels)
+  m = div(prod(size(x)), channels)
```
MikeInnes: Did you update the manifest, too?

thebhatman: Yes I did pkg up , but it didn't seem to have changed the Manifest. Just to verify, I did a fresh add of this branch and Zygote master, and it is working...

MikeInnes: You need to `activate` the Zygote project in order to use the package versions in the manifest, and also to affect it with `update`.

thebhatman: Thanks! I have updated the manifest now.

thebhatman:
So once the adjoint for dropdims is merged https://github.com/FluxML/Zygote.jl/pull/264, backprop on InstanceNorm will work.. Only backprop on GroupNorm is failing due to this issue https://github.com/FluxML/Zygote.jl/issues/263

dhairyagandhi96:
Very cool!

test/cuda/curnn.jl
```diff
@@ -8,27 +8,27 @@ using Flux, CuArrays, Test
...
 
-      Δ = randn(size(y))
+      #Δ = randn(size(y))
 
-      Flux.back!(y, Δ)
-      Flux.back!(cuy, gpu(Δ))
+      #Flux.back!(y, Δ)
+      #Flux.back!(cuy, gpu(Δ))
 
-      @test x.grad ≈ collect(cux.grad)
```
MikeInnes: The point of this code was to check gradients. Now it just checks that the cuda version of the model has the same parameters as the CPU version (and gradients don't work currently).

docs/src/training/optimisers.md
```diff
@@ -3,25 +3,25 @@
 Consider a [simple linear regression](../models/basics.md). We create some dummy data, calculate a loss, and backpropagate to calculate gradients for the parameters `W` and `b`.
 
 ```julia
-using Flux, Flux.Tracker
+using Flux, Flux.Zygote
```
MikeInnes: @dhairyagandhi96 Flux already exports `gradient`, so this may not be necessary

docs/src/training/optimisers.md
```diff
@@ -58,3 +58,78 @@ AMSGrad
 NADAM
 ADAMW
 ```
+
+## Optimiser Interface
```
MikeInnes: These new docs are great, but could we have them as a separate PR? Since they seem broadly independent of the Zygote changes.

docs/src/training/optimisers.md
```diff
@@ -3,25 +3,25 @@
...
-predict(x) = W*x .+ b
+predict(x) = (W * x) .+ b
 loss(x, y) = sum((predict(x) .- y).^2)
 
 x, y = rand(5), rand(2) # Dummy data
 l = loss(x, y) # ~ 3
 
 θ = Params([W, b])
-grads = Tracker.gradient(() -> loss(x, y), θ)
+grads = Zygote.gradient(() -> loss(x, y), θ)
```
MikeInnes:
```suggestion
grads = gradient(() -> loss(x, y), θ)
```
@dhairyagandhi96

docs/src/training/optimisers.md
dhairyagandhi96: Fixed

MikeInnes:
Seem to be some issues with our GPU CI, so just merging.

src/layers/recurrent.jl
```diff
@@ -42,21 +42,6 @@ end
...
-    truncate!(rnn)
-
-Truncates the gradient of the hidden state in recurrent layers. The value of the
-state is preserved. See also `reset!`.
-
-Assuming you have a `Recur` layer `rnn`, this is roughly equivalent to
-
-    rnn.state = Tracker.data(rnn.state)
-"""
-truncate!(m) = prefor(x -> x isa Recur && (x.state = _truncate(x.state)), m)
```
iblis17: well, is there an alternative?

Issue #317, opened 2018-07-12
Iterators for batches and epochs


MikeInnes:
These are simple to do but it's just convenient to have explicit versions of them. We should also document patterns such as using `partition` for batches; and really any other patterns that come up in models a lot.

shreyas-kowshik:
I've implemented a basic function with the following code for generating batches. This could be useful for generating batches.

```
using Base.Iterators: partition

function batch_iter(x::AbstractArray{T,N},batch_size::Int64) where {T,N}
    """
    Given an array of N-dimensions, this function will return an iterator over batches on this array
    """
    num_examples = size(x)[end]
    remaining_size = size(x)[1:end-1]
    BATCH_ALL = prod(remaining_size)
    
    return [reshape(float(hcat(data)),remaining_size) for data in partition(x,BATCH_ALL)]
end
```

Was this what was required? @MikeInnes

Tokazama:
Would it be warranted to have some sort of batch type that annotates width, height, channels, minibatch size, etc.? As I've been doing digging I see these values extracted frequently in the code.

MikeInnes:
That sounds more like #326. I do think something AxisArrays-like would be great for that kind of use case though.

Issue #150, opened 2018-01-18
Train iterator helper function


ZacCranko:
I couldn't see one of these, but it seemed to me that something like

```julia
function train_iterator(inputs::AbstractArray, labels::AbstractArray; epochs = 1, batch_size = last(size(inputs)), shuffle = false)
    @assert ndims(inputs) >= 1  "incorrectly shaped inputs with size $(size(inputs))"
    @assert ndims(labels) >= 1  "incorrectly shaped labels with size $(size(labels))"
    @assert (n = last(size(inputs))) == last(size(labels)) "inputs don't pair with labels: got $(last(size(inputs))) inputs and $(last(size(labels))) labels"

    data_inds, label_inds = ntuple((_)->:, ndims(inputs) - 1), ntuple((_)->:, ndims(labels) - 1)
    
    indices = shuffle ? Base.OneTo(n) : randperm(n)
    iter = ((view(inputs,  data_inds..., batch), view(labels, label_inds..., batch)) 
                    for _     in Base.OneTo(epochs) 
                    for batch in Iterators.partition(indices, batch_size))
    return iter
end
```

would be helpful for calling `train!`. Does anyone have any thoughts or improvements?

MikeInnes:
Seems like a good idea, although it'd be good to split this into composable pieces; e.g. we could write it as `epochs(batches(shuffle(xs), 100), 10)` or something (where shuffle actually already exists).

MikeInnes:
Closing in favour of #317.

PR #675, opened 2019-03-09
Added Iterator For Batches

shreyas-kowshik:
Addresses #317 
I am unable to understand exactly what needs to be done in case of epochs. Can someone please clarify?

### DIFF
--- a/src/data_iter.jl
+++ b/src/data_iter.jl
@@ -24,19 +24,27 @@ mutable struct BatchIter{T,N}
     ind::Any # indices
 end
 
-function _batch_iter(c::AbstractArray,n::Int,ind::Any)
+function _batch_iter(c::AbstractArray,n::Int,ind::Any,to_shuffle::Bool)
     b = BatchIter(c,n,ind)
-    indices = collect(1:size(c)[end])[shuffle(1:end)]
+    indices = collect(1:size(c)[end])
+    if to_shuffle == true
+        indices = indices[shuffle(1:end)]
+    end
     ind = collect(partition(indices,n))
     b.ind = ind
     return b
 end
 
-batch_iter(c::AbstractArray,n::Int) = _batch_iter(c,n,nothing)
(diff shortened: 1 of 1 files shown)


dhairyagandhi96:
We need to consider what all cases need to be handled when we create iterators.
- One major one being that of having labelled data (esp. if shuffling is involved).
- This function doesn't seem to create iterators, rather creates copies of the data, which is best avoided.
- We also need to take care of classic cases, with images etc.
- We currently don't return batches, rather just singular datapoints

I would expect something like this to work as
```julia
julia> x = rand(28,28,3, 100);

julia> it = batch_iter(x, 5);

# `it` is an iterator that returns a batch of 5 random images, picked from `x`

julia> first(it) # should be ((28,28,3,5),)
(28, 28, 3)
```

I also feel we should document examples of using `partition` clearly. Note that the way you've used it makes it such that we create chunks of `BATCH_ALL` size, instead of the batches we need.

Also might be worth checking out https://github.com/FluxML/Flux.jl/issues/150 if you haven't already. We need to make it such that we can ask for N epochs from calling the batches over as many times.

Thanks for the patch! :)

shreyas-kowshik:
@dhairyagandhi96 Sorry for the late reply. I have changed the basic structure to address generic Arrays of images and feature data. An example of using the `batch_iter` is also added.
This approach now does not require copies of the original data. Rather it just needs to store a partition over the batch indices.

dhairyagandhi96:
We should also have a nice way of demanding these batches for an entire epochs worth of training

shreyas-kowshik:
@dhairyagandhi96 Added the options for training over multiple epochs. The feature to shuffle the batches is made optional.

johnnychen94:
Please follow the [Iterator Interface](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration-1), also check https://github.com/JuliaLang/julia/blob/master/base/iterators.jl#L75-L103 for inspiration.

datnamer:
There are some pretty cool data iterators here: https://mldatautilsjl.readthedocs.io/en/latest/data/pattern.html

PR #678, opened 2019-03-10
Added Adaptive Pooling Layers.

thebhatman:
These Adaptive Pooling layers will be useful in the middle of a long convolutional chain when the user doesn't know the size of input but wants a required target output size.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -6,8 +6,8 @@ using Base: tail
 using MacroTools, Juno, Requires, Reexport, Statistics, Random
 using MacroTools: @forward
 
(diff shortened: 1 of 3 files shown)


MikeInnes:
Do these have precedent in other ML libraries?

It'd also be good to expose these as function (e.g. `adaptivemaxpool(x, (w, h))`) and then build the layers on top of that.

thebhatman:
Pytorch has adaptive pooling layers. These layers are used occasionally though. Yes it would be better to expose them as functions.

ToucheSir:
Should be closed by https://github.com/FluxML/Flux.jl/pull/1239 already.

PR #679, opened 2019-03-11
Added Group Normalization Layer

shreyas-kowshik:
Added the Group Normalization layer that is known to improve the overall accuracy on several segmentation and calssification tasks.


johnnychen94:
Hi Sereyas,

I just had a glance into your PR, according to my experience, it's better to break your current PR into several PRs so that others can review it more easily. For example:

* add iris dataloader
* add maxout layer
* add groupnorm layer

and you need more test cases to persuade others that your PRs are robust and amazing. 

Just try to make your PR diff small as possible as you can

PR #680, opened 2019-03-11
Added new loss functions.

thebhatman:
I have added the KL Divergence Loss function, Poisson loss function, Logcosh loss, and Hinge loss function.


src/layers/stateless.jl
```diff
@@ -54,3 +54,31 @@ function normalise(x::AbstractArray, dims)
...
+    Poisson Loss function
+Poisson loss function is a measure of how the predicted distribution diverges from the expected distribution.
+
+"""
+Poisson(ŷ, y) = sum(ŷ .- y .* log.(ŷ)) *1 // size(y,2)
+
+"""
+    Logcosh Loss function
+"""
+
```
MikeInnes: There are some unnecessary newlines going around here (in the doc strings as well)

MikeInnes:
Seems good; as a point of style it'd be better if the function names were lowercase (we use uppercase for layers / types).

Can you also put a reference for these in the docs?

thebhatman:
Thank you @MikeInnes. I have made the necessary changes.

src/layers/stateless.jl
```diff
@@ -54,3 +54,30 @@ function normalise(x::AbstractArray, dims)
...
+https://isaacchanghau.github.io/post/loss_functions/
+"""
+poisson(ŷ, y) = sum(ŷ .- y .* log.(ŷ)) *1 // size(y,2)
+
+"""
+    Logcosh Loss function
+"""
+logcosh(ŷ, y) = sum(log.(cosh.(ŷ .- y)))
+
+hinge(ŷ, y) = sum(max.(0.0, 1 .-  ŷ .* y)) *1 // size(y,2)
```
CarloLucibello: 0.0 -> 0 to avoid promotion to Float64

src/layers/stateless.jl
```diff
@@ -54,3 +54,30 @@ function normalise(x::AbstractArray, dims)
...
+    Poisson Loss function
+Poisson loss function is a measure of how the predicted distribution diverges from the expected distribution.
+https://isaacchanghau.github.io/post/loss_functions/
+"""
+poisson(ŷ, y) = sum(ŷ .- y .* log.(ŷ)) *1 // size(y,2)
+
+"""
+    Logcosh Loss function
+"""
+logcosh(ŷ, y) = sum(log.(cosh.(ŷ .- y)))
```
CarloLucibello: this at the same time not very common and easy for users to define, are we sure we want it in Flux?

thebhatman: Yes, I agree it is easy to define. But most ML libraries have it predefined and users might expect it to be predefined.

MikeInnes: Usually the reason for having it predefined is numerical stability; should we be doing something for that here?

Also, is there a reason for this to not be doing the averaging that other functions do?

thebhatman: Thank you @MikeInnes for pointing it out. It was a mistake from my side. logcosh loss has to be averaged. I have corrected it.

pshashk: Maybe we could move logcosh activation to NNLib as something like
```
logcosh(x) = x + softplus(-2 * x) - log(2)
∇logcosh(x) = tanh(x)
```
and the loss function will trivially become `logcosh.(ŷ - y)`

thebhatman: @MikeInnes would it be better if I remove logcosh function from here if it isn't actually required?

MikeInnes: Yeah, it might be better to remove it, as the name doesn't make it obvious that it's meant to be used as a loss.

thebhatman: I have removed the logcosh function.

thebhatman:
Where can I add reference to the new loss functions in the docs?

MikeInnes:
Create a section for docstrings [here](https://fluxml.ai/Flux.jl/stable/training/training/#Loss-Functions-1). We should add an API reference at some point, but that'll do for now.

pshashk:
Submitted `logcosh` to NNlib https://github.com/FluxML/NNlib.jl/pull/108/
Should I add `∇logcosh(x) = tanh(x)` gradient to Tracker?

thebhatman:
@MikeInnes, I have added the loss functions to the docs.

staticfloat:
> Should I add ∇logcosh(x) = tanh(x) gradient to Tracker?

For now, yes.  Eventually we'll need to add it to Zygote as well.

thebhatman:
Is there anything else to be done here?

docs/src/training/training.md
```diff
@@ -32,6 +32,18 @@ Flux.train!(loss, ps, data, opt)
 
 The objective will almost always be defined in terms of some *cost function* that measures the distance of the prediction `m(x)` from the target `y`. Flux has several of these built in, like `mse` for mean squared error or `crossentropy` for cross entropy loss, but you can calculate it however you want.
 
+In-built loss functions:
```
MikeInnes: This should be in the API reference, not tutorial-style docs. Maybe in the layers section, if there's nowhere better. Good to have though.

dhairyagandhi96:
Bump

src/layers/stateless.jl
```diff
@@ -49,3 +49,29 @@ function normalise(x::AbstractArray; dims=1)
   σ′ = std(x, dims = dims, mean = μ′, corrected=false)
   return (x .- μ′) ./ σ′
 end
+
+function normalise(x::AbstractArray, dims)
```
MikeInnes: Is this intended to be part of this patch?

src/layers/stateless.jl
```diff
@@ -49,3 +49,29 @@ function normalise(x::AbstractArray; dims=1)
...
+end
+
+"""
+    Poisson Loss function
+Poisson loss function is a measure of how the predicted distribution diverges from the expected distribution.
+https://isaacchanghau.github.io/post/loss_functions/
+"""
+poisson(ŷ, y) = sum(ŷ .- y .* log.(ŷ)) *1 // size(y,2)
+
+hinge(ŷ, y) = sum(max.(0, 1 .-  ŷ .* y)) *1 // size(y,2)
```
MikeInnes: `hinge` needs a doc string.

src/layers/stateless.jl
thebhatman: Oops. Yeah I have removed it now.

docs/src/models/layers.md
```diff
@@ -65,3 +65,15 @@ AlphaDropout
 LayerNorm
 GroupNorm
 ```
+
+## In-built loss functions:
```
MikeInnes:
```suggestion
## Loss Functions

```

src/layers/stateless.jl
```diff
@@ -49,3 +49,28 @@ function normalise(x::AbstractArray; dims=1)
   σ′ = std(x, dims = dims, mean = μ′, corrected=false)
   return (x .- μ′) ./ σ′
 end
+
+"""
+    Kullback Leibler Divergence(KL Divergence)
```
MikeInnes: Could use a signature. Right now this title will show up as code.

src/layers/stateless.jl
```diff
@@ -49,3 +49,28 @@ function normalise(x::AbstractArray; dims=1)
...
+"""
+    Poisson Loss function
+Poisson loss function is a measure of how the predicted distribution diverges from the expected distribution.
+https://isaacchanghau.github.io/post/loss_functions/
+"""
+poisson(ŷ, y) = sum(ŷ .- y .* log.(ŷ)) *1 // size(y,2)
+
+"""
+    Hinge Loss function
+Measures the loss given the prediction ŷ and true labels y(containing 1 or -1). This is usually used for measuring whether two inputs are similar or dissimilar
```
MikeInnes: Full stop at the end. The last sentence also doesn't sound that helpful right now (isn't that just the definition of a loss?); maybe it just needs rewording or a reference.

docs/src/models/layers.md
```diff
@@ -66,7 +66,7 @@ LayerNorm
 GroupNorm
 ```
 
-## In-built loss functions:
+## Loss functions:
```
MikeInnes: The other headings in this section are capitalised, and don't end with a colon.

thebhatman:
I have made the suggested changes in documentation, are there any more changes to be done?

thebhatman:
Bump. Is there anything left to do here?

docs/src/models/layers.md
```diff
@@ -65,3 +65,15 @@ AlphaDropout
 LayerNorm
 GroupNorm
 ```
+
+## Loss Functions
```
MikeInnes: I think we should perhaps label these "cost functions", since we sometimes refer to the whole objective function (including the model) as the "loss".

MikeInnes:
```suggestion
## Cost Functions
```

MikeInnes:
Once you've changed the line I mentioned, feel free to merge.

bors delegate

MikeInnes:
bors d+

MikeInnes:
Is bors borked again? @dhairyagandhi96 do we have any idea how to fix it?

dhairyagandhi96:
Seems like it, the gitlab yaml was changed recently, but I've noticed this with bors only recently. I'll take a look at that.

dhairyagandhi96:
You might want to try again, updated some keys in gitlab.

dhairyagandhi96:
Also, can you add a test with `Float32` and make sure no spurious promotions to `Float64` occur

thebhatman:
Done. I have added the tests to check for spurious promotions.

PR #681, opened 2019-03-12
add Flux.stop to training docs

dellison:
In a recent [post](https://discourse.julialang.org/t/rmsprop-and-early-stopping-in-flux-jl/21065) on the Julia discourse, someone new to Flux didn't know about using `Flux.stop` to exit the training loop. This PR adds a note about this feature in the training section of the docs to (hopefully!) make this feature a little more discoverable.


MikeInnes:
Great idea, thanks!

Issue #640, opened 2019-02-24
Error: "back! was already used"


jpsamaroo:
Opening this issue as recommended by @staticfloat 

I'm pretty new to Flux and was just trying out a simple LSTM model on a sequence learning task, so this error took me completely by surprise. Here's the MWE:

```
using Flux
using Statistics

Ni = 8
Nn = 8
Nt = 100

# Custom onehot
seq = Vector{Float32}[]
for i in 1:Nt
    idx = rand(1:Ni)
    vec = zeros(Float32, Ni)
    vec[idx] = 1.0
    push!(seq, vec)
end
X = seq[1:end-1]
Y = seq[2:end]
dataset = collect(Iterators.zip(X,Y))

model = Chain(
    LSTM(Ni, Nn),
    softmax
) |> f32

loss(x,y) = Flux.mse(model(x), y)
accuracy() = Statistics.mean([Flux.Tracker.data(loss(x,y)) for (x,y) in dataset])

opt = ADAM()

accs = Float32[]
function cb()
    # Uncommenting this line deepens the stacktrace considerably but does not change the error
    #push!(accs, accuracy())
    println("cb")
end

Flux.train!(loss, Flux.params(model), dataset, opt; cb=cb)
```

And here's the error:
```
ERROR: LoadError: `back!` was already used
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] back_(::Flux.Tracker.Call{Missing,Tuple{}}, ::Array{Float32,1}, ::Bool) at /home/jpsamaroo/.julia/packages/Flux/8XpDt/src/tracker/back.jl:30
 [3] back(::Flux.Tracker.Tracked{Array{Float32,1}}, ::Array{Float32,1}, ::Bool) at /home/jpsamaroo/.julia/packages/Flux/8XpDt/src/tracker/back.jl:46
 [4] #3 at /home/jpsamaroo/.julia/packages/Flux/8XpDt/src/tracker/back.jl:26 [inlined]
 [5] foreach at ./abstractarray.jl:1869 [inlined]
[...]

PR #685, opened 2019-03-15
Add note on reset! usage in recurrence docs

jpsamaroo:
Closes https://github.com/FluxML/Flux.jl/issues/640

### DIFF
--- a/docs/src/models/recurrence.md
+++ b/docs/src/models/recurrence.md
@@ -114,3 +114,13 @@ truncate!(m)
 Calling `truncate!` wipes the slate clean, so we can call the model with more inputs without building up an expensive gradient computation.
 
 `truncate!` makes sense when you are working with multiple chunks of a large sequence, but we may also want to work with a set of independent sequences. In this case the hidden state should be completely reset to its original value, throwing away any accumulated information. `reset!` does this for you.
+
+In general, when training with recurrent layers in your model, you'll want to call `reset!` or `truncate!` for each loss calculation:
+
+```julia
+function loss(x,y)
+  l = Flux.mse(m(x), y)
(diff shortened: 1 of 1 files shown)


docs/src/models/recurrence.md
```diff
@@ -114,3 +114,13 @@ truncate!(m)
 Calling `truncate!` wipes the slate clean, so we can call the model with more inputs without building up an expensive gradient computation.
 
 `truncate!` makes sense when you are working with multiple chunks of a large sequence, but we may also want to work with a set of independent sequences. In this case the hidden state should be completely reset to its original value, throwing away any accumulated information. `reset!` does this for you.
+
+Note that when training with recurrent layers in your model, `reset!` needs to be called for each loss calculation:
```
MikeInnes: I think something like `In general, you'll want to...` since it's not strictly required depending on what behaviour you want.

Also, you can use either `truncate!` or `reset!`.

jpsamaroo: In what cases can `reset!`/`truncate!` be avoided in the loss calculation? I wasn't able to figure out an alternative with my LSTM example.

MikeInnes: You wouldn't usually want to for something like a simple LSTM. For something like char-rnn you could in theory backprop over all input the model has seen; things are also a little more complex with things like moving window BPTT.

Of course ideally we'd document all these patterns, but with all the APIs about to shift I think it's ok to wait until Zygote is in to be really thorough here.

jpsamaroo: PR updated

MikeInnes:
:+1: Thanks!

Issue #671, opened 2019-03-08
Deprecate initialiser keyword arguments


MikeInnes:
These don't really work, they aren't flexible enough and are no more convenient that just supplying your own weights e.g. 

```julia
Dense(randn(2, 5), rand(2))
```

This is partly a docs issue as people don't realise that this is a supported option. Although like most things, it will of course get more convenient with #669.

shreyas-kowshik:
I am willing to work on this. 
But I need to clarify a few things beforehand : 
```
using Flux
using Distributions:Normal

dist = Normal(0.0,1.0)
function initW(out,in)
    W = rand(dist,out,in)
end

Dense(2,3,sigmoid,initW=initW,initb=zeros)
```

This piece of code does work right?...So why is there a need to depreciate it?...
I apologize if I'm being silly here. In case if this is not what is relevant to this issue, then can you please explain what needs to be done exactly?

MikeInnes:
The original idea was that you could choose initilialisers for a large model with the `init` kwarg (e.g. `ResNet50(init = xorot_uniform)`. But we ended up using different initialisers on a per-parameters basis (`initW` and `initb`) which defeats the purpose, since now you have to deal with every weight separately anyway. On top of that, this extra API must be explicitly added to each new layer, which makes it harder to achieve consistency (most newer layers don't have it at all).

So given those issues and the fact that you can just as easily write this without `initW`, we can simplify things a fair bit by just removing it.

shreyas-kowshik:
So all that needs to be done is to add deprecation warnings to ```Dense(...)```...?
Also for the warning, is this message fine : 
```
Dense(in,out,act,initW,initb) is deprecated; use Dense(in,out) instead
```
OR is this required : 
```
Dense(in,out,act,initW,initb) is deprecated; use Dense(in,out,init) instead
```

Accordingly I shall make changes to the API then.

Can you please clarify it a bit here @MikeInnes. 
Thank You.

johnnychen94:
https://github.com/FluxML/Flux.jl/blob/7418a2d7d721451e4af5e5835abc424acfd2a80c/src/layers/basic.jl#L65-L71

`Dense(in,out,act,initW,initb) is deprecated; use Dense(in,out,init) or Dense(W, b, σ) instead`

shreyas-kowshik:
@MikeInnes added the deprecation warning in #722 .

PR #696, opened 2019-03-25
Added GroupNorm Layer

shreyas-kowshik:
The previous PR was closed due to some silly git errors from my side. Here is the original code added.


src/layers/normalise.jl
```diff
@@ -286,3 +286,90 @@ function Base.show(io::IO, l::InstanceNorm)
   (l.λ == identity) || print(io, ", λ = $(l.λ)")
   print(io, ")")
 end
+
+"""
+Group Normalization. Known to improve the overall accuracy in case of classification and segmentation tasks.
+Link : https://arxiv.org/pdf/1803.08494.pdf
```
staticfloat: Can you add onto this description a little bit more?  Contrast it to Instance Normalization and Batch Normalization so that the user might understand what this is without needing to read the paper?  Look at the docstrings for `InstanceNorm` or `BatchNorm` for inspiration.

src/layers/normalise.jl
```diff
@@ -286,3 +286,90 @@ function Base.show(io::IO, l::InstanceNorm)
...
+end
+
+GroupNorm(chs::Integer, G::Integer,N::Integer, λ = identity;
+          initβ = (i) -> zeros(Float32, i), initγ = (i) -> ones(Float32, i), ϵ = 1f-5, momentum = 0.1f0) =
+  GroupNorm(G,N,λ, param(initβ(chs)), param(initγ(chs)),
+            zeros(G,N), ones(G,N), ϵ, momentum, true)
+
+function(gn::GroupNorm)(x)
+  size(x,ndims(x)-1) == length(gn.β) || error("Group Norm expected $(length(gn.β)) channels, but got $(size(x,ndims(x)-1)) channels")
+  ndims(x) > 2 || error("Need to pass atleast 3 channels for Group Norm to work")
```
staticfloat: `atleast` -> `at least`

src/layers/normalise.jl
```diff
@@ -286,3 +286,90 @@ function Base.show(io::IO, l::InstanceNorm)
...
 end
+
+"""
+Group Normalization. Known to improve the overall accuracy in case of classification and segmentation tasks.
+Link : https://arxiv.org/pdf/1803.08494.pdf
+"""
+
+mutable struct GroupNorm{F,V,W,N,T}
+  G::T # number of groups
+  N::T # Batch Size
```
staticfloat: I don't think we need to bake `N` into the `GroupNorm`; we should be able to allow variable batch sizes with this operation.

src/layers/normalise.jl
```diff
@@ -286,3 +286,90 @@ function Base.show(io::IO, l::InstanceNorm)
...
+GroupNorm(chs::Integer, G::Integer,N::Integer, λ = identity;
+          initβ = (i) -> zeros(Float32, i), initγ = (i) -> ones(Float32, i), ϵ = 1f-5, momentum = 0.1f0) =
+  GroupNorm(G,N,λ, param(initβ(chs)), param(initγ(chs)),
+            zeros(G,N), ones(G,N), ϵ, momentum, true)
+
+function(gn::GroupNorm)(x)
+  size(x,ndims(x)-1) == length(gn.β) || error("Group Norm expected $(length(gn.β)) channels, but got $(size(x,ndims(x)-1)) channels")
+  ndims(x) > 2 || error("Need to pass atleast 3 channels for Group Norm to work")
+  (size(x,ndims(x) -1))%gn.G == 0 || error("The number of groups ($(gn.G)) must divide the number of channels ($(size(x,ndims(x) -1)))")
+  (size(x,ndims(x)) == gn.N) || error("Number of samples in batch not equal to that passed")
```
staticfloat: I don't think we nee this check; we should be able to deal with variable batch sizes.

src/layers/normalise.jl
```diff
@@ -286,3 +286,90 @@ function Base.show(io::IO, l::InstanceNorm)
...
+  # γ : (1,1...,C,1)
+  # β : (1,1...,C,1)
```
staticfloat: I'm not sure what these comments mean.  Can you expand them or remove them?

src/layers/normalise.jl
```diff
@@ -286,3 +286,90 @@ function Base.show(io::IO, l::InstanceNorm)
...
+
+  dims = length(size(x))
+  groups = gn.G
+  channels = size(x, dims-1)
+  batches = size(x,dims)
+  channels_per_group = convert(Int32,div(channels,groups))
```
staticfloat: Is there a reason you explicitly want `channels_per_group` to be an `Int32`?  `div()` should already give you an integral type.

src/layers/normalise.jl
```diff
@@ -286,3 +286,90 @@ function Base.show(io::IO, l::InstanceNorm)
...
+  γ = reshape(gn.γ, affine_shape...)
+  β = reshape(gn.β, affine_shape...)
+  if !gn.active
+    μ = reshape(gn.μ, affine_shape...)
+    σ² = reshape(gn.σ², affine_shape...)
+    ϵ = gn.ϵ
+  else
+    T = eltype(x)
+    og_shape = size(x)
+    x = reshape(x,((size(x))[1:end-2]...,channels_per_group,groups,batches))
```
staticfloat: Clever use of `reshape()`, I suggest that you name this something other than `x` though, as it makes looking at things like `ndims(x)` below needlessly confusing.

src/layers/normalise.jl
shreyas-kowshik: @staticfloat The size of the mean and variance matrices depends on N and is (Channels/Groups,Batch SIze). So should'nt that need the implementation of N?
I apologize if it's a silly question but this is what I understand so far.

staticfloat: You're right that the size of `μ` and `σ²` will change depending on `N`; but we don't need to know it beforehand.  We can just broadcast `μ` and `σ²` up to the proper size when we get a new batch.  (You are correct that we cannot have fully variable batch sizes; e.g. running with `N=8` and then `N=16` won't work; we will need to reset the `μ` and `σ²` before doing that).

You can initialize `μ` and `σ²` as size `(G, 1)` where `G` is the number of batches, then when you do this line:

```
gn.μ = (1 - mtm) .* gn.μ .+ mtm .* reshape(data(μ), (groups,batches))
```

`gn.u` will be automatically broadcast up to the proper size.

shreyas-kowshik:
@staticfloat Thank you for your feedback. I have made the requested changes.

staticfloat:
Add some tests as well; I think there might be some problems with variable names and multiple code branches, so it will be good to do tests (both with an `active` layer and a `!active` layer).  Take a look at the instance normalization tests for inspiration.

shreyas-kowshik:
@staticfloat I have added the tests for Group Normalization. They are passing on my machine. Can you please review it once?

src/layers/normalise.jl
```diff
@@ -286,3 +286,110 @@ function Base.show(io::IO, l::InstanceNorm)
...
+
+"""
+Group Normalization. 
+Known to improve the overall accuracy in case of classification and segmentation tasks.
+
+GroupNorm(chs::Integer, G::Integer, λ = identity;
+          initβ = (i) -> zeros(Float32, i), initγ = (i) -> ones(Float32, i), 
+          ϵ = 1f-5, momentum = 0.1f0)
+
+chs is the numebr of channels, the channeld dimension of your input.
```
staticfloat: `numebr`, `channeld` both look to be typos.

src/layers/normalise.jl
```diff
@@ -286,3 +286,110 @@ function Base.show(io::IO, l::InstanceNorm)
   (l.λ == identity) || print(io, ", λ = $(l.λ)")
   print(io, ")")
 end
+
+"""
+Group Normalization. 
+Known to improve the overall accuracy in case of classification and segmentation tasks.
```
staticfloat: Let's make the second line mention that this is a normalization layer that can perform better than Batch or Instance normalization.

src/layers/normalise.jl
```diff
@@ -286,3 +286,110 @@ function Base.show(io::IO, l::InstanceNorm)
...
+
+GroupNorm(chs::Integer, G::Integer, λ = identity;
+          initβ = (i) -> zeros(Float32, i), initγ = (i) -> ones(Float32, i), 
+          ϵ = 1f-5, momentum = 0.1f0)
+
+chs is the numebr of channels, the channeld dimension of your input.
+For an array of N dimensions, the (N-1)th index is the channel dimension.
+
+G is the number of groups along which the statistics would be computed.
+The number of groups must divide the number of channels for this to work.
```
staticfloat: I think a better way of saying this is that `the number of channels must be an integer multiple of the number of groups`.

src/layers/normalise.jl
```diff
@@ -286,3 +286,110 @@ function Base.show(io::IO, l::InstanceNorm)
...
+  if !gn.active
+    og_shape = size(x)
+    μ = reshape(gn.μ, μ_affine_shape...) # Shape : (1,1,...C/G,G,1)
+    σ² = reshape(gn.σ², μ_affine_shape...) # Shape : (1,1,...C/G,G,1)
+    ϵ = gn.ϵ
+    y = reshape(x,((size(x))[1:end-2]...,channels_per_group,groups,batches))
+  else
+    T = eltype(x)
+    og_shape = size(x)
+    y = reshape(x,((size(x))[1:end-2]...,channels_per_group,groups,batches))
```
staticfloat: Since you're calculating the same thing for `y` in each branch, you can just pull that out of the `if` statement and do it once before.

test/layers/normalisation.jl
```diff
@@ -1,5 +1,5 @@
 using Flux: testmode!
-using Flux.Tracker: data
+using Flux.Tracker: data
```
staticfloat: extra whitespace

test/layers/normalisation.jl
```diff
@@ -200,3 +200,108 @@ end
...
+    @test size(m.μ) == (m.G,1)
+    @test size(m.σ²) == (m.G,1)
+    @test size(y) == sizes
+  end
+
+  # show that group norm is the same as instance norm when the group size is the same as the number of channels
+  let IN = InstanceNorm(4), GN = GroupNorm(4,4), sizes = (2,2,3,4,5),
+      x = param(reshape(collect(1:prod(sizes)), sizes))
+    @test IN(x) ≈ GN(x)
+  end
```
staticfloat: I really like this test.  Can we get one for BatchNorm too, when the group size is `1`?

staticfloat:
It's coming together!  I have more minor comments this time around, once these are addressed I think we'll be ready to merge!

shreyas-kowshik:
@staticfloat Thank you for the feedback. I have made the requested changes.

test/layers/normalisation.jl
```diff
@@ -200,3 +200,114 @@ end
...
+
+  # show that group norm is the same as batch norm for a group of size 1 and batch of size 1
+  let BN = BatchNorm(4), GN = GroupNorm(4,4), sizes = (2,2,3,4,1),
+      x = param(reshape(collect(1:prod(sizes)), sizes))
+    @test BN(x) ≈ GN(x)
+  end
```
staticfloat: This is wrong, it should be `GroupNorm(4,1)`, right?

shreyas-kowshik: A group size of ```1``` would imply 4 groups...right? GroupNorm takes in the number of channels and the number of groups. I referred to this figure for understanding : 
https://shaoanlu.files.wordpress.com/2018/03/gn_fig2.jpg?w=788

Am I wrong somewhere...If so can you please tell me where I am wrong...

shreyas-kowshik: The test is passing with this piece of code. I think it makes sense as the second parameter as ```4``` , we have the same case as instance norm which is the same as batch norm for a batch size of 1.
Do you think something is wrong here? @staticfloat

staticfloat: I'm sorry, you're correct.  I switched `LN` with `BN` in my head.  This is good.  Thanks!

MikeInnes:
Looks like the tests are failing; if it's unrelated it might just need a merge of master.

src/Flux.jl
```diff
@@ -6,10 +6,8 @@ using Base: tail
 using MacroTools, Juno, Requires, Reexport, Statistics, Random
 using MacroTools: @forward
 
-export Chain, Dense, Maxout,
```
johnnychen94: `Maxout` is accidentally deleted here.

src/layers/normalise.jl
```diff
@@ -286,3 +286,109 @@ function Base.show(io::IO, l::InstanceNorm)
...
+
+"""
+Group Normalization. 
+This layer can outperform Batch-Normalization and Instance-Normalization.
+
+GroupNorm(chs::Integer, G::Integer, λ = identity;
+          initβ = (i) -> zeros(Float32, i), initγ = (i) -> ones(Float32, i), 
+          ϵ = 1f-5, momentum = 0.1f0)
+
+chs is the number of channels, the channel dimension of your input.
```
johnnychen94: best practice of writting variable name is to add `` around it -- REPL can recognize it.
* chs --> `chs`
* G --> `G`

src/layers/normalise.jl
```diff
@@ -286,3 +286,109 @@ function Base.show(io::IO, l::InstanceNorm)
   (l.λ == identity) || print(io, ", λ = $(l.λ)")
   print(io, ")")
 end
+
+"""
+Group Normalization. 
+This layer can outperform Batch-Normalization and Instance-Normalization.
+
+GroupNorm(chs::Integer, G::Integer, λ = identity;
```
johnnychen94: According to https://docs.julialang.org/en/v1/manual/documentation/
> Always show the signature of a function at the top of the documentation, with a four-space indent so that it is printed as Julia code.

src/layers/normalise.jl
```diff
@@ -286,3 +286,109 @@ function Base.show(io::IO, l::InstanceNorm)
...
+G is the number of groups along which the statistics would be computed.
+The number of channels must be an integer multiple of the number of groups.
+
+Example:
+```
+m = Chain(Conv((3,3), 1=>32, leakyrelu;pad = 1),
+          GroupNorm(32,16)) # 32 channels, 16 groups (G = 16), thus 2 channels per group used          
+```
+
+Link : https://arxiv.org/pdf/1803.08494.pdf
```
johnnychen94: Personally, I prefer to add a title to this link to info the users what it might be
"""
References: 
[1] Wu, Y., & He, K. (2018). Group normalization. In Proceedings of the European Conference on Computer Vision (ECCV) (pp. 3-19). https://arxiv.org/abs/1803.08494
"""

src/Flux.jl
shreyas-kowshik: @johnnychen94 Thanks!

src/layers/normalise.jl
```diff
@@ -286,3 +286,109 @@ function Base.show(io::IO, l::InstanceNorm)
...
+mapchildren(f, gn::GroupNorm) =  # e.g. mapchildren(cu, BN)
+  GroupNorm(gn,G,gn.λ, f(gn.β), f(gn.γ), f(gn.μ), f(gn.σ²), gn.ϵ, gn.momentum, gn.active)
+
+_testmode!(gn::GroupNorm, test) = (gn.active = !test)
+
+function Base.show(io::IO, l::GroupNorm)
+  print(io, "GroupNorm($(join(size(l.β), ", "))")
+  (l.λ == identity) || print(io, ", λ = $(l.λ)")
+  print(io, ")")
+end
```
johnnychen94: usually, it's best practice to add one newline at EOF

src/layers/normalise.jl
```diff
@@ -286,3 +286,109 @@ function Base.show(io::IO, l::InstanceNorm)
...
+  γ::V  # scale
+  μ::W  # moving mean
+  σ²::W  # moving std
+  ϵ::N
+  momentum::N
+  active::Bool
+end
+
+GroupNorm(chs::Integer, G::Integer, λ = identity;
+          initβ = (i) -> zeros(Float32, i), initγ = (i) -> ones(Float32, i), ϵ = 1f-5, momentum = 0.1f0) =
```
johnnychen94: What you need is not add `Float32` here. Instead, you need to do type conversion in the implementation details `function(gn::GroupNorm)(x)`

src/layers/normalise.jl
```diff
@@ -286,3 +286,109 @@ function Base.show(io::IO, l::InstanceNorm)
...
+Example:
+```
+m = Chain(Conv((3,3), 1=>32, leakyrelu;pad = 1),
+          GroupNorm(32,16)) # 32 channels, 16 groups (G = 16), thus 2 channels per group used          
+```
+
+Link : https://arxiv.org/pdf/1803.08494.pdf
+"""
+
+mutable struct GroupNorm{F,V,W,N,T}
```
johnnychen94: I prefer to add constraints `GroupNorm{F<:Function, V<:Number, W<:Number, T<:Integer}` and do type conversions with new constructors.
And, do we really need `N` here, is it possible to be absorbed by `V` or `W`?

shreyas-kowshik:
@MikeInnes The tests related to GroupNorm have passed. The issue was due to the missing export of Maxout as pointed out by @johnnychen94 .

src/layers/normalise.jl
shreyas-kowshik: Thanks! I'll have a look at these and incorporate them in my commit.

staticfloat:
Thanks @shreyas-kowshik!

MikeInnes:
Can you add this layer to the docs, and also an entry to NEWS.md?

johnnychen94:
I'm still suspect about the default initializer, we can't just let some type (in this case `Float32`) be the default type.

```julia
GroupNorm(chs::Integer, G::Integer, λ = identity;
          initβ = (i) -> zeros(Float32, i), initγ = (i) -> ones(Float32, i), ϵ = 1f-5, momentum = 0.1f0)
```

But since there'll be deprecation on these `init` keywords, it doesn't matter much. https://github.com/FluxML/Flux.jl/issues/671

shreyas-kowshik:
@MikeInnes #728 adds GroupNorm to docs and NEWS.md

PR #698, opened 2019-03-25
make Maxout trainable

oxinabox:
Fix what I failed to do in #647


MikeInnes:
We never added a news item for this either, can you add a quick line there?

Issue #700, opened 2019-03-26
Conv with non quadratic size broken


jw3126:
The following used to work in older versions of Flux, but errors now:
```julia
julia> using Flux

julia> randn(Float32, 10,10,1,1) |> Conv((6,6), 1=>1, Flux.σ)
Tracked 5×5×1×1 Array{Float32,4}:
[:, :, 1, 1] =
 0.295454   0.541948  0.0864369  0.959286   0.891815 
 0.41697    0.218508  0.4821     0.746911   0.499474 
 0.500112   0.898844  0.592479   0.637593   0.874677 
 0.0124033  0.168184  0.0669264  0.024803   0.109923 
 0.286872   0.219004  0.0657297  0.0456955  0.0598319

julia> randn(Float32, 10,10,1,1) |> Conv((6,1), 1=>1, Flux.σ)
ERROR: DimensionMismatch("new dimensions (5, 5, 6, 1, 1) must be consistent with array size 300")
```

MikeInnes:
@tejank10 can you take a look at this?

thebhatman:
After quite a bit of debugging, I found out the error lies in https://github.com/FluxML/NNlib.jl/blob/d07ac0bfd3c71c3a29bc9c22becbba19227bbeb5/src/impl/conv.jl#L21. Here, as you can see, ouput height dimension is using K[1], instead of K[2], S[1] instead of S[2]. Which is why, it was working only for kernel sizes where width and height were equal.

thebhatman:
Changing the output size calculation to this, fixes the issue.
```
function output_size(c::ConvDims{I,K,C,S,P,D,F}) where {I, K, C, S, P, D, F}
    O_w = div(I[1] + P[1] + P[2] - (K[1] - 1) * D[1] - 1, S[1]) + 1
    O_h = div(I[2] + P[3] + P[4] - (K[2] - 1) * D[2] - 1, S[2]) + 1
    return (O_w, O_h)
end
```

tejank10:
Thanks, @jw3126 for pointing this out and adding tests, and @thebhatman for a quick fix!

MikeInnes:
This is fixed on newer NNlibs. We'll tag a release with that soon.

nmoran:
Also possible to override library version in local file until fix makes it to stable version.

For example:
```
using Flux: NNlib.ConvDims

[...]

PR #99, opened 2017-10-30
TrackedArray: implement `mean`

iblis17:
```julia
julia> p
Tracked 2×3 Array{Float64,2}:
 1.0  3.0  5.0
 2.0  4.0  6.0
```

Before
```julia
julia> @benchmark Flux.Tracker.back!(sum($p, 2) ./ size($p, 2), ones(2, 1))
BenchmarkTools.Trial:
  memory estimate:  3.44 KiB
  allocs estimate:  75
  --------------
  minimum time:     20.438 μs (0.00% GC)
  median time:      21.239 μs (0.00% GC)
  mean time:        22.354 μs (1.68% GC)
  maximum time:     3.811 ms (98.51% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

After
```julia
julia> @benchmark Flux.Tracker.back!(mean($p, 2), ones(2, 1))
BenchmarkTools.Trial:
  memory estimate:  1008 bytes
  allocs estimate:  21
  --------------
  minimum time:     5.973 μs (0.00% GC)
  median time:      6.310 μs (0.00% GC)
  mean time:        6.630 μs (1.96% GC)
  maximum time:     680.709 μs (97.28% GC)
  --------------
  samples:          10000
  evals/sample:     6
```

PR #701, opened 2019-03-26
add broken test for #700

jw3126:
See #700. This is not a fix, just test cases.

### DIFF
--- a/test/layers/conv.jl
+++ b/test/layers/conv.jl
@@ -61,3 +61,29 @@ end
   x_hat = ConvTranspose((3, 3), 1 => 1)(y)
   @test size(x_hat) == size(x)
 end
+
+@testset "Conv with non quadratic window #700" begin
+  data = fill(0f0, 7,7,1,1)
+  data[4,4,1,1] = 1
+
+  l = Conv((3,3), 1=>1)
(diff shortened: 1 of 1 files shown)


test/layers/conv.jl
```diff
@@ -33,3 +33,29 @@ end
...
+
+  l = Conv((3,3), 1=>1)
+  expected = fill(zero(eltype(l.weight)),5,5,1,1)
+  expected[2:end-1,2:end-1,1,1] = l.weight
+  @test expected == l(data)
+
+  l = Conv((3,1), 1=>1)
+  expected = fill(zero(eltype(l.weight)),5,7,1,1)
+  expected[2:end-1,4,1,1] = l.weight
+  @test_broken expected == l(data)
```
tejank10: Using `@test_broken` is fine for now. Once [#99](https://github.com/FluxML/NNlib.jl/pull/99) gets merged, we would have to change it to `@test`.

jw3126: So should I change it or not? Is the plan is to merge this PR first or not?

dhairyagandhi96: Could change it to honor non square sizes now

jw3126: Done!

tejank10:
Thanks @jw3126 , good work!

PR #709, opened 2019-03-27
Small fix for recurrence documentation

DoktorMike:
The code

```julia
julia> RNN(10, 5)
Recur(RNNCell(Dense(15, 5)))
```
is changed to 

```julia
julia> RNN(10, 5)
Recur(RNNCell(Dense(10, 5)))
```

Which is the correct output.


MikeInnes:
Looks like it should be

```julia
julia> RNN(10, 5)
Recur(RNNCell(10, 5, tanh))
```
on recent versions.

DoktorMike:
Right you are. I've updated the PR.

MikeInnes:
:+1: Thanks!

Issue #702, opened 2019-03-26
`activations` is broken


MikeInnes:
In Flux v0.6.1, the last tutorial code sample in [regularisation](https://github.com/FluxML/Flux.jl/blob/d6a75e1289488945ade1cfa8867717fe1ed19557/docs/src/models/regularisation.md) fails

```julia
using Flux
c = Chain(Dense(10,5,σ),Dense(5,2),softmax)
Flux.activations(c, rand(10))
```
which raises `ERROR: MethodError: no method matching accumulate(::getfield(Flux, Symbol("##62#63")), ::Array{Float64,1}, ::Array{Any,1})`

It seems that `accumulate(op, A; dims::Integer, [init])` doesn't support multiple inputs as you wrote in `activations(c::Chain, x) = accumulate((x, m) -> m(x), x, c.layers)` anymore.

_Originally posted by @johnnychen94 in https://github.com/FluxML/Flux.jl/pull/174#issuecomment-414040344_

johnnychen94:
Here's my original patch, I closed the PR https://github.com/FluxML/Flux.jl/pull/355 because @MikeInnes at that time you fixed it by yourself.

```julia
activations(m::Any, x, rst=[]) = push!(rst,m(x))
activations(c::Chain, x, rst=[]) = begin
    rst = activations(c[1], x, rst)
    if length(c) >= 2
        rst = activations(c[2:end], rst[end], rst)
    end
    return rst
end
```

MikeInnes:
`reduce` is surprisingly broken in Julia. It'd be nice just to have a simple reduce-over-tuples (should be a couple lines) and use that. Or figure out why else this went wrong.

johnnychen94:
A temporary patch is implemented in https://github.com/FluxML/Flux.jl/pull/710
However, this issue should be kept open unless fix on upstream is impossible or there's a real reason to have a Flux-version `reduce`

MikeInnes:
Meh. That would be nice but the important thing is that this works, and your implementation is totally reasonable. So I think we can close this for now and change it if `accumulate` every improves.

PR #710, opened 2019-03-28
naive implementation of activations

johnnychen94:
As a temporary patch to https://github.com/FluxML/Flux.jl/issues/702
It's highly possible that there's a better implementation. But this can be a start-point for a stable `activations` function.

According to https://github.com/FluxML/Flux.jl/commit/9d1d5187f349252365e73adb7a2da66caf29bfcf#commitcomment-30267188, we don't export this function.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -47,6 +47,9 @@ Calculate the forward results of each layers in Chain `c`
 """
 activations(m, x) = activations(Chain(m), x)
 function activations(c::Chain, x)
+    if length(c) == 0
+        return []
+    end
     rst = [c[1](x), ]
     if length(c) == 1
         return rst
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -46,14 +46,14 @@ end
 Calculate the forward results of each layers in Chain `c`
 """
 function activations(c::Chain, x)
-    if isempty(c)
-        return [x, ]
-    end
-    rst = [x, c[1](x)]
-    for l in c[2:end]
-        push!(rst, l(rst[end]))
-    end
-    return rst
+  if isempty(c)
+    return [x, ]
+  end
+  rst = [x, c[1](x)]
+  for l in c[2:end]
+    push!(rst, l(rst[end]))
+  end
+  return rst
 end
 
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -42,19 +42,18 @@ end
 
 
 # This is a temporary and naive inplementation
+# it might be replaced in the future to get better performance
 # see issue https://github.com/FluxML/Flux.jl/issues/702
+# Johnny Chen -- @johnnychen94
 """
-    activations(c::Chain, x)
-Calculate the forward results of each layers in Chain `c`
+    activations(c::Chain, input)
+Calculate the forward results of each layers in Chain `c` with `input` as model input.
 """
-function activations(c::Chain, x)
-  rst = Array{Any,1}()
-  if isempty(c)
-    return rst
-  end
-  push!(rst, c[1](x))
-  for l in c[2:end]
-    push!(rst, l(rst[end]))
+function activations(c::Chain, input)
+  rst = []
+  for l in c
+    x = get(rst, length(rst), input)
+    push!(rst, l(x))
   end
   return rst
 end
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -41,8 +41,8 @@ function Base.show(io::IO, c::Chain)
 end
 
 
-# This is a temporary and naive inplementation
-# it might be replaced in the future to get better performance
+# This is a temporary and naive implementation
+# it might be replaced in the future for better performance
 # see issue https://github.com/FluxML/Flux.jl/issues/702
 # Johnny Chen -- @johnnychen94
 """
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -42,18 +42,14 @@ end
 
 
 """
-    activations(c::Union{Chain, Any}, x)
+    activations(c::Chain, x)
 Calculate the forward results of each layers in Chain `c`
 """
-activations(m, x) = activations(Chain(m), x)
 function activations(c::Chain, x)
-    if length(c) == 0
-        return []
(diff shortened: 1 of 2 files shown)


src/layers/basic.jl
```diff
@@ -40,7 +40,26 @@ function Base.show(io::IO, c::Chain)
...
+"""
+    activations(c::Union{Chain, Any}, x)
+Calculate the forward results of each layers in Chain `c`
+"""
+activations(m, x) = activations(Chain(m), x)
+function activations(c::Chain, x)
+    if length(c) == 0
+        return []
+    end
+    rst = [c[1](x), ]
```
MikeInnes: I think it'd be better to just have `rst = []` followed by the loop. There's no guarantee that all activations will have the same type.

test/layers/basic.jl
```diff
@@ -1,6 +1,21 @@
 using Test, Random
+import Flux: activations
 
 @testset "basic" begin
+    @testset "helpers" begin
+        @testset "activations" begin
+            dummy_model = Chain(Dense(10,5,σ),Dense(5,2),softmax)
+            x = rand(10)
+            @test_nowarn activations(dummy_model[1], x)
```
MikeInnes: Why is `@test_nowarn` particularly necessary?

johnnychen94: I use `test_nowarn` for interface test, and the following `@test` is used for numerical test

src/layers/basic.jl
johnnychen94: this's is necessary since in the loop we need an initialization: `rst[end]`

johnnychen94: Oh I think I made a mistake here
We should add `x` into the result list of `activations` just like `accumulate` does.

test/layers/basic.jl
MikeInnes: Seems easiest to just keep the functional tests. These aren't adding much over them.

src/layers/basic.jl
```diff
@@ -40,7 +40,26 @@ function Base.show(io::IO, c::Chain)
   print(io, ")")
 end
 
-activations(c::Chain, x) = accumulate((x, m) -> m(x), c.layers, init = x)
+
+"""
+    activations(c::Union{Chain, Any}, x)
+Calculate the forward results of each layers in Chain `c`
+"""
+activations(m, x) = activations(Chain(m), x)
```
MikeInnes: I'd prefer to just keep the `Chain` method alone for now.

src/layers/basic.jl
```diff
@@ -40,7 +40,22 @@ function Base.show(io::IO, c::Chain)
...
+
+"""
+    activations(c::Chain, x)
+Calculate the forward results of each layers in Chain `c`
+"""
+function activations(c::Chain, x)
+    if isempty(c)
+        return [x, ]
+    end
+    rst = [x, c[1](x)]
```
johnnychen94:
```julia
rst = [x, ]
for l in c
    push!(rst, l(rst[end]))
end
```
triggers a conversion fail.

johnnychen94:
I changed the tab from 4-spaces to 2-spaces to match Flux.jl style
In case you don't know the tip: add `?w=1` to the end of url to review the changes with whitespaces ignored.

src/layers/basic.jl
```diff
@@ -40,7 +40,22 @@ function Base.show(io::IO, c::Chain)
...
+
+"""
+    activations(c::Chain, x)
+Calculate the forward results of each layers in Chain `c`
+"""
+function activations(c::Chain, x)
+  if isempty(c)
+    return [x, ]
+  end
+  rst = [x, c[1](x)]
```
MikeInnes: You don't need to initialise this with `x` (`accumulate` doesn't behave this way). This will also have issues if anything after `c[1]` yields a different type.

So just do `rst = []` and then have the loop.

johnnychen94:
Could you provide a test case that returns different types?

Johnny Chen

陈久宁

School of Mathematical Sciences, East China Normal University, Shanghai

华东师范大学数学科学学院

2019年4月5日 +0800 00:15 Mike J Innes <notifications@github.com>，写道：

@MikeInnes commented on this pull request.

________________________________

In src/layers/basic.jl<https://apc01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FFluxML%2FFlux.jl%2Fpull%2F710%23discussion_r272258011&data=02%7C01%7C%7Ccf2e3db4fca342a77a2708d6b918c25b%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C636899913315946713&sdata=4xiue6746emcnwRDN24IX9uwG8wYyyGvxhW%2FdlLWQW0%3D&reserved=0>:

> @@ -40,7 +40,22 @@ function Base.show(io::IO, c::Chain)
   print(io, ")")
 end

-activations(c::Chain, x) = accumulate((x, m) -> m(x), c.layers, init = x)
+
+"""
+    activations(c::Chain, x)
+Calculate the forward results of each layers in Chain `c`
+"""
+function activations(c::Chain, x)
+  if isempty(c)
+    return [x, ]
+  end
+  rst = [x, c[1](x)]


You don't need to initialise this with x (accumulate doesn't behave this way). This will also have issues if anything after c[1] yields a different type.

So just do rst = [] and then have the loop.

MikeInnes:
`Chain(identity, identity, x -> :foo)`.

test/layers/basic.jl
```diff
@@ -1,63 +1,76 @@
...
-        # numeric test should be put into testset of corresponding layer
+  @testset "helpers" begin
+    @testset "activations" begin
+      dummy_model = Chain(Dense(10,5,σ),Dense(5,2),softmax)
+      x = rand(10)
+      @test activations(Chain(), x) == []
+      @test activations(dummy_model, x)[1] == dummy_model[1](x)
+      @test activations(dummy_model, x)[2] == x |> dummy_model[1] |> dummy_model[2]
+
+      @test_nowarn activations(Chain(identity, x->:foo), 1) # different types
```
MikeInnes: `@test ... == :foo` would be better here.

MikeInnes:
Thanks @johnnychen94!

PR #716, opened 2019-03-29
Added Local Contrast Normalization

shreyas-kowshik:
Added code and tests for the same

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -1,3 +1,5 @@
+using NNlib
+
 """
     testmode!(m)
     testmode!(m, false)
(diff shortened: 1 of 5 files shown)


MikeInnes:
Looks like this has pulled in some unrelated commits.

Not a fan of the name `LCN` for this. Will also need docs changes and NEWS.md

shreyas-kowshik:
@MikeInnes Sorry for the delayed response. I have changed the name of the layer to 
```
LocalContrastNorm
```

Added the layer to docs and NEWS.md.

staticfloat:
This is going to need a rebase on top of latest `master`.

shreyas-kowshik:
@staticfloat Made the changes.

staticfloat:
You've got a type error; mixing `Float64` and `Float32`.

src/layers/normalise.jl
```diff
@@ -392,3 +394,65 @@ function Base.show(io::IO, l::GroupNorm)
...
+
+struct LocalContrastNorm{N}
+    r::N # Radius of kernel
+end
+
+function(lcn::LocalContrastNorm)(x)
+    radius = lcn.r
+    
+    filter_shape = (radius,radius,size(x)[3],1)
+    filter = gaussian_filter(filter_shape)
```
staticfloat: We don't want to be calculating the filter again and again every time we invoke this function; we should create the filter, lazily, within this function, but keep it around for next time and only recreate if we need to.  I'm thinking something like this:

```julia
if size(lcn.filter,3) != size(x,3)
    lcn.filter = gaussian_filter((lcn.r, lcn.r, size(x,3), 1))
end
```

And we can initalize `lcn.filter` to `nothing` or a zero-dimensional array or something like that to make sure it is created the first time it's called.

src/layers/normalise.jl
```diff
@@ -392,3 +394,65 @@ function Base.show(io::IO, l::GroupNorm)
...
+    mid_w = div(filter_shape[1],2) + 1
+    mid_h = div(filter_shape[2],2) + 1
+    
+    for i in 1:filter_shape[1]
+        for j in 1:filter_shape[2]
+            out[i,j,:,:] .= gaussian(i - mid_w,j - mid_h)
+        end
+    end
+    
+    return Float64.(out/sum(out))
```
staticfloat: We don't want to force it to `Float64`; we actually want it to be the same type as `x`.

shreyas-kowshik:
@staticfloat Made the requested changes.

src/layers/normalise.jl
```diff
@@ -392,3 +394,72 @@ function Base.show(io::IO, l::GroupNorm)
   (l.λ == identity) || print(io, ", λ = $(l.λ)")
   print(io, ")")
 end
+
+"""
+  LocalContrastNorm(3)(param(rand(10,10,3,5)))
+
+Local Contrast Normalization Layer can serve an importanat role in achieving better accuracies in object recognition datasets.
```
staticfloat: `importanat`

test/layers/normalisation.jl
```diff
@@ -311,3 +311,8 @@ end
   end
 
 end
+
+@testset "LocalContrastNorm" begin
+    x = rand(10,10,3,5)
+    @test size(x) == size(LocalContrastNorm(3)(x))
```
staticfloat: Please add some tests that compare against known-good values, e.g. run a test in PyTorch or TensorFlow with a LCN, save the values, then compare here so we can make sure we don't break this in the future. :)

shreyas-kowshik: The issue here is that both tensorflow and pytorch don't have a LCN implementation.
So would hand written tests for cases checked mathematically do?

ToucheSir:
Given this uses custom kernels with loops, it ought to be in NNlib(CUDA) instead with a Flux layer added later. Feel free to re-open this PR there if you're still interested @shreyas-kowshik, but otherwise I think we should close this.

PR #717, opened 2019-03-29
Fix uses of norm in regularisation docs

dlozeve:
`norm` is part of the `LinearAlgebra` package, and is not available by default.

The output of `params` has also changed. Finally, the `activations` function has to be imported, but is broken: https://github.com/FluxML/Flux.jl/issues/702 Until this is fixed, maybe the example should be removed from the docs?


dlozeve:
Closed as near duplicate of https://github.com/FluxML/Flux.jl/pull/506
(sorry, I searched for "regularisation" and not "regularization"...)

PR #718, opened 2019-03-29
Add asymmetric padding

staticfloat:
This goes on top of https://github.com/FluxML/Flux.jl/pull/714, since NNlib now has support for asymmetrical padding this provides the Flux-level API for it.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -27,6 +27,12 @@ git-tree-sha1 = "055eb2690182ebc31087859c3dd8598371d3ef9e"
 uuid = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
 version = "0.5.3"
 
+[[CSTParser]]
+deps = ["LibGit2", "Test", "Tokenize"]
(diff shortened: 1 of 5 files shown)


MikeInnes:
May as well fold into #714?

staticfloat:
Sure, if you like, just close #714 and merge this one when https://github.com/FluxML/Tracker.jl/pull/11 is merged.

staticfloat:
@MikeInnes looks like we're good to go, merge at your leisure.

MikeInnes:
Thanks a lot @staticfloat.

I suspect it'd be cleaner to store the `Dims` object in the layer, rather than storing the same settings and reconstructing it on the fly. But that's not important for now.

staticfloat:
> I suspect it'd be cleaner to store the Dims object in the layer, 

I thought of that, but considering that we need to re-create it for every differently-sized `x`, it wasn't high on my priority list.  It may save us a little bit of time, but not that much.  :P

MikeInnes:
Ah yes, I didn't realise you're actually including the size of the input. This is probably the right way to do it then.

PR #720, opened 2019-03-30
Implemented vectorized LRNorm

thebhatman:
Added function for Local Response Normalization. Reference: https://papers.nips.cc/paper/4824-imagenet-classification-with-deep-convolutional-neural-networks.pdf. The previous PRs on adding LRNorm used for loops and indexing, due to which automatic differentiation was not supported. I have implemented a vectorized version of LRNorm.

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -407,27 +407,31 @@ m(x) #x is a 4-D input
 ```
 """
 mutable struct LRNorm{F, V, W, N}
-  α :: F
-  β :: V
-  n :: W
-  k :: N
-  active :: Bool
+  α::F
+  β::V
+  n::W
+  k::N
+  active::Bool
 end
 
 LRNorm(α = 10.0 ^ (-4), β = 0.75, n = 5, k = 2.0) = LRNorm(α, β, n, k, true)
 
 function (lrn::LRNorm)(x)
   num_channels = size(x, 3)
-  y = zeros(eltype(x), size(x))
-  for i in size(y, 3)
-    lower_lim = max(1, trunc(Int, i + 1 - lrn.n/2))
-    upper_lim = min(num_channels, trunc(Int, i + 1 + lrn.n/2))
-    y[:, :, lower_lim:upper_lim, :] .= 1
+  if typeof(x) <: TrackedArray
+    y = zeros(eltype(x.data), 1, 1, num_channels, num_channels)
+  else
+    y = zeros(eltype(x), 1, 1, num_channels, num_channels)
   end
-  y = lrn.α .* sum(y .* (x .^ 2), dims = 3)
-  y = y .+ lrn.k
-  y = y .^ lrn.β
-  return x ./ y 
+  for i in num_channels
+    lower_lim = max(1, trunc(Int, i - lrn.n/2))
+    upper_lim = min(num_channels, trunc(Int, i + lrn.n/2))
+    y[1, 1, lower_lim:upper_lim, i] .= 1
+  end
+  norm = lrn.α .* NNlib.conv(x .^ 2, y)
+  norm = norm .+ lrn.k
+  norm = norm .^ lrn.β
+  return x ./ norm 
 end 
 
 children(lrn::LRNorm) =
@@ -437,3 +441,8 @@ mapchildren(f, lrn::LRNorm) =  # e.g. mapchildren(cu, BN)
   LRNorm(lrn.α, f(lrn.β), f(lrn.n), f(lrn.k), lrn.active)
 
 _testmode!(lrn::LRNorm, test) = (lrn.active = !test)
+
+function Base.show(io::IO, l::LRNorm)
+  print(io, "LRNorm(α = $(join(l.α))")
+  print(io, ",β = $(l.β) ,n = $(l.n),k = $(l.k))")
+end
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -420,8 +420,8 @@ function (lrn::LRNorm)(x)
   num_channels = size(x, 3)
   y = zeros(eltype(data(x)), 1, 1, num_channels, num_channels)
   for i in 1:num_channels
-    lower_lim = max(1, trunc(Int, i - lrn.n/2))
-    upper_lim = min(num_channels, trunc(Int, i + lrn.n/2))
+    lower_lim = max(1, trunc(Int, i - div(lrn.n,2)))
+    upper_lim = min(num_channels, trunc(Int, i + div(lrn.n,2)))
     y[1, 1, lower_lim:upper_lim, i] .= 1
   end
   norm = lrn.α .* NNlib.conv(x .^ 2, y)
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -392,3 +392,48 @@ function Base.show(io::IO, l::GroupNorm)
   (l.λ == identity) || print(io, ", λ = $(l.λ)")
   print(io, ")")
 end
+
+"""
+    LRNorm(α, β, n, k)
+Local Response Normalization layer. The `n` input should be the number of 
+adjacent kernel maps to sum over.
+See [ImageNet Classification with Deep Convolutional
+Neural Networks](https://papers.nips.cc/paper/4824-imagenet-classification-with-deep-convolutional-neural-networks.pdf).
+
+Example : 
+```julia
+m = LRNorm(0.0001, 0.75, 5, 2.0)
+m(x) #x is a 4-D input
+```
+"""
+mutable struct LRNorm{F, V, W, N}
+  α :: F
+  β :: V
+  n :: W
+  k :: N
+  active :: Bool
+end
+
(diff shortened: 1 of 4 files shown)


thebhatman:
I have added tests for Local Response Normalisation.

thebhatman:
Travis CI details say there were 4 errored tests. But the tests I added for LRNorm passed. I am not sure where the errored tests are coming from...

src/layers/normalise.jl
```diff
@@ -392,3 +392,48 @@ function Base.show(io::IO, l::GroupNorm)
...
+m(x) #x is a 4-D input
+```
+"""
+mutable struct LRNorm{F, V, W, N}
+  α :: F
+  β :: V
+  n :: W
+  k :: N
+  active :: Bool
+end
```
staticfloat: Please follow the style of the rest of the file, no spaces between `n :: W`, for instance.

src/layers/normalise.jl
```diff
@@ -392,3 +392,48 @@ function Base.show(io::IO, l::GroupNorm)
...
+  for i in size(y, 3)
+    lower_lim = max(1, trunc(Int, i + 1 - lrn.n/2))
+    upper_lim = min(num_channels, trunc(Int, i + 1 + lrn.n/2))
+    y[:, :, lower_lim:upper_lim, :] .= 1
+  end
+  y = lrn.α .* sum(y .* (x .^ 2), dims = 3)
+  y = y .+ lrn.k
+  y = y .^ lrn.β
+  return x ./ y 
+end
```
staticfloat: I.... don't think this is calculating the output properly.  The output value for each output channel is a sum across the input channels; it's more like a convolution, whereas this output sums across the input channels only once.

staticfloat (changes requested):
I don't think this is calculating the output properly; I think you need to read the paper again, particularly the equation in section `3.3`.  I wrote [a suggestion on a previous attempt to write LRN for Flux](https://github.com/FluxML/Flux.jl/pull/312#issuecomment-467816832) that may be helpful.

src/layers/normalise.jl
thebhatman: Oh! I realized the error. When I take sum along the dimension of channels, the output dimensions will no longer be the same as the input dimension. So when I compared the output of my LRNorm function with your suggestion of 1x1 convolution, the output matched exactly only in one channel. I will correct this. Thank you!

thebhatman:
@staticfloat I have made the requested changes. What kind of tests would be appropriate?

test/layers/normalisation.jl
```diff
@@ -311,3 +311,7 @@ end
   end
 
 end
+@testset "LRNorm" begin
+  x = param(reshape(Float32[1:96;], 4, 4, 3, 2))
+  @test size(LRNorm()(x)) == size(x) 
+end
```
staticfloat: Please add some tests that compare against known-good values, e.g. run a test in PyTorch or TensorFlow with a LRN, save the values, then compare here so we can make sure we don't break this in the future.  :)

src/layers/normalise.jl
```diff
@@ -392,3 +392,57 @@ function Base.show(io::IO, l::GroupNorm)
...
+
+LRNorm(α = 10.0 ^ (-4), β = 0.75, n = 5, k = 2.0) = LRNorm(α, β, n, k, true)
+
+function (lrn::LRNorm)(x)
+  num_channels = size(x, 3)
+  if typeof(x) <: TrackedArray
+    y = zeros(eltype(x.data), 1, 1, num_channels, num_channels)
+  else
+    y = zeros(eltype(x), 1, 1, num_channels, num_channels)
+  end
```
staticfloat: This can be made more succint as just `y = zeros(eltype(data(x)), 1, 1, num_channels, num_channels)`

staticfloat (changes requested):

thebhatman:
I have added tests by comparing with tensorflow's LRN. The output values are approximately equal and comparable. So I have rounded off both outputs to two digits and then compared them.

staticfloat:
While I do expect some difference between the two implementations, we should be able to get a lot more than just 2 decimal places of accuracy.  If we have to round with `digits=2`, that makes me think there might be something wrong with our algorithm.  Can you show the actual TF output and the comparison with our algorithm?

thebhatman:
Here's the computed output by our function:
Input is ```x = param(reshape(Float32[1:20;], 1, 1, 5, 4))```
```
julia> m = LRNorm()
LRNorm(α = 0.0001,β = 0.75 ,n = 5,k = 2.0)
julia> m(x)
Tracked 1×1×5×4 Array{Float64,4}:
[:, :, 1, 1] =
 0.5942915817132711

[:, :, 2, 1] =
 1.1878710105259485

[:, :, 3, 1] =
 1.7801403935901126

[:, :, 4, 1] =
 2.3736092915813667

[:, :, 5, 1] =
 2.9674555452942113

[:, :, 1, 2] =
 3.547816324791048

[:, :, 2, 2] =
 4.126683180908961

[:, :, 3, 2] =
 4.698799923539177

[:, :, 4, 2] =
 5.29318127413301

[:, :, 5, 2] =
 5.891985429255813

[:, :, 1, 3] =
 6.436172025553092

[:, :, 2, 3] =
 6.971188482472584

[:, :, 3, 3] =
 7.49092944638354

[:, :, 4, 3] =
 8.102437972776501

[:, :, 5, 3] =
 8.72667895604141

[:, :, 1, 4] =
 9.214966343739228

[:, :, 2, 4] =
 9.665769310057835

[:, :, 3, 4] =
 10.092066777866355

[:, :, 4, 4] =
 10.748286820459093

[:, :, 5, 4] =
 11.430093727457718
```

And here is tensorflow LRN output for same input :
```
[[[[ 0.5945813]]

  [[ 1.1890287]]

  [[ 1.7832088]]

  [[ 2.3769882]]

  [[ 2.9702334]]]


 [[[ 3.5628128]]

  [[ 4.154593 ]]

  [[ 4.745444 ]]

  [[ 5.3352346]]

  [[ 5.923835 ]]]


 [[[ 6.5111175]]

  [[ 7.0969534]]

  [[ 7.681217 ]]

  [[ 8.263785 ]]

  [[ 8.844531 ]]]


 [[[ 9.423337 ]]

  [[10.000079 ]]

  [[10.574641 ]]

  [[11.146905 ]]

  [[11.716756 ]]]]
```

thebhatman:
There is actually a considerable difference in the output.

staticfloat:
Agreed; I suggest tracking down the tensorflow computation and seeing if you can figure out where the difference lies; I think this is more than just floating-point noise; I think there is an actual difference in algorithm.

thebhatman:
I realized the difference between our algorithm and the actual algorithm. The actual algorithm sums upto n adjacent channels. Now as long ```i - n/2``` is less than 1, it keeps adding the square of the neuron in channel 1. Likewise, as long as ```i + n/2``` is greater than number of channels, it keeps adding the square of the neuron in the last channel. But our algorithm, adds the square of these neurons(which are on the first and last kernels) only once. The actual algorithm requires to sum over n adjacent channels for every neuron in each channel. But the way we have implemented it, it sums over less than n adjacent channels at the channels which are on the edge.

thebhatman:
I was trying tensorflow's local_response_normalization for an input of all ones and track the way it is computing the normalization term. The parameters were : depth_radius n = (I varied it from 2 to 4. It is supposed to be the number of adjacent channels to sum over.), bias k = 0, alpha = 1, beta = 1. This ensures that the normalization term is sum of squares of activations in the local region of interest. 
The input was a batch of ones with width = height = 2, channels = 4, batch_size = 1.
Now the output of tensorflow's lrn layer is the same for n = 2, 3, 4. The output was :
```
[[[[0.5 0.5]
   [0.5 0.5]]

  [[0.5 0.5]
   [0.5 0.5]]

  [[0.5 0.5]
   [0.5 0.5]]

  [[0.5 0.5]
   [0.5 0.5]]]]
```
Also when the input was a batch with width = height = 4, channels = 1, batch_size = 1.
Now since num_channels is 1, LRNorm across channels is supposed to sum over just that channel itself for any n > 1. But tensorflow's lrn layer gives this output for n = 2:
```
[[[[0.33333334 0.25       0.25       0.33333334]
   [0.33333334 0.25       0.25       0.33333334]
   [0.33333334 0.25       0.25       0.33333334]
   [0.33333334 0.25       0.25       0.33333334]]]]
```
and for n = 3 and 4:
```
[[[[0.25 0.25 0.25 0.25]
   [0.25 0.25 0.25 0.25]
   [0.25 0.25 0.25 0.25]
   [0.25 0.25 0.25 0.25]]]]
```
I think this proves that tensorflow's lrn layer is normalizing within a channel by summing over the square of n adjacent activations(spatially) in the same channel. But the paper deals with lrn which sums over n adjacent activations across different channels at same spatial location.

thebhatman:
The two types of local response normalization : Across channels and within channels has been discussed here https://stats.stackexchange.com/questions/145768/importance-of-local-response-normalization-in-cnn

thebhatman:
I have kinda hit a wall here. Any suggestions would be highly appreciated.

CarloLucibello:
closing as stale

PR #722, opened 2019-03-31
Added deprecation warning to Dense(in,out,act,initW,initb)

shreyas-kowshik:
Addresses issue #671 . 
@MikeInnes is this what was needed?

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -73,8 +73,18 @@ end
 Dense(W, b) = Dense(W, b, identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
-               initW = glorot_uniform, initb = zeros)
-  depwarn("Dense(in,out,σ,initW,initb) is deprecated; use Dense(W,b) instead",:Dense)
+               initW = nothing, initb = nothing)
+  if initW != nothing || initb != nothing
+    depwarn("Dense(in,out,σ,initW,initb) is deprecated; use Dense(W,b) instead",:Dense)
+  end 
+
+  if initW == nothing
+    initW = glorot_uniform
+  end 
+  if initb == nothing
+    initb = zeros
+  end 
+
   return Dense(param(initW(out, in)), param(initb(out)), σ)
 end
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -1,3 +1,5 @@
+using Base: depwarn
+
 """
     Chain(layers...)
 
@@ -72,6 +74,7 @@ Dense(W, b) = Dense(W, b, identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
                initW = glorot_uniform, initb = zeros)
+  depwarn("Dense(in,out,σ,initW,initb) is deprecated; use Dense(W,b) instead")
   return Dense(param(initW(out, in)), param(initb(out)), σ)
 end
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -74,7 +74,7 @@ Dense(W, b) = Dense(W, b, identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
                initW = glorot_uniform, initb = zeros)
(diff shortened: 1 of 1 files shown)


src/layers/basic.jl
```diff
@@ -72,6 +74,7 @@ Dense(W, b) = Dense(W, b, identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
                initW = glorot_uniform, initb = zeros)
+  depwarn("Dense(in,out,σ,initW,initb) is deprecated; use Dense(W,b) instead",:Dense)
```
johnnychen94: we can't just deprecate the whole method here. The tricky part is that keyword arguments are not counted in multiple dispatch.
 
My solution (not sure if it works)
1. change `initW = glorot_uniform` to `initW = nothing`
2. manual check before initialization
```julia
if initW is nothing
  initW = glorot_uniform;
else
  depwarn
end
```

shreyas-kowshik: I see...My bad...I'll fix this accordingly...

johnnychen94:
Looks like this strategy works.

Now you need to have a look at the codebase and deprecate other functions such as `Diagonal`, `ConvTranspose`. And don't forget to modify the testcases.

BTW, it might be useful to provide a link to https://github.com/FluxML/Flux.jl/issues/671 as a comment alongside your `depwarn`.

shreyas-kowshik:
I suppose that the other layers do not have this issue and this mostly pertained to the Dense layer. 
Conv and ConvTranspose only have a single init in them. So I don't think that they should need deprecation.
At least this was what I could infer from #671 .
Is this all or am I getting it wrong? @MikeInnes

johnnychen94:
It's not about single init keyword, it's about consistency. If you deprecate `init` as a keyword argument, deprecate them all.

MikeInnes:
Let's get the `Dense` deprecation done and then do a review of other layers. I think we probably will want to get rid of all of them though.

I think the `kwargs` method could be defined separately. You should be able to do something like

```julia
@deprecate Dense(in::Integer, out::Integer; initW = ...) = Dense(initW(out, in), ...)
```

which separates the deprecated code from the new code nicely, and gives a clearer instruction to the user.

johnnychen94:
@MikeInnes `@deprecate` might not work well for keyword arguments. They don't participate in dispatch.

```julia
julia> foo(a,b;k=1) = println()

julia> @deprecate foo(a,b;k=1) foo(a,b)

julia> foo(1,2)
┌ Warning: `foo(a, b; k=1)` is deprecated, use `foo(a, b)` instead.
│   caller = top-level scope at none:0
└ @ Core none:0
```

https://github.com/JuliaDocs/Documenter.jl/blob/c59fe92f880e7acfbf2610fed0c70dc364e4c3a0/src/Documenter.jl#L473-L513 is an example of such deprecation work.

MikeInnes:
It actually works if you do it the other way around:
```julia
julia> @deprecate foo(a,b;k=1) foo(a,b)

julia> foo(a,b;k=1) = println()

julia> foo(1, 2)
```

johnnychen94:
If it's all about deprecating `Dense`, then I think this PR is ready to merge. @MikeInnes

src/layers/basic.jl
```diff
@@ -71,7 +73,18 @@ end
 Dense(W, b) = Dense(W, b, identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
-               initW = glorot_uniform, initb = zeros)
+               initW = nothing, initb = nothing)
+  if initW != nothing || initb != nothing
+    depwarn("Dense(in,out,σ,initW,initb) is deprecated; use Dense(W,b) instead",:Dense)
```
johnnychen94: I think "keyword argument `initW` is deprecated; use `Dense(W,b)` to initialize" is more descriptive to the current one.

CarloLucibello:
closing as we are going in a different direction with #1440

PR #727, opened 2019-04-04
add other optimizers to documentation

hossein-pourbozorg:


dhairyagandhi96:
Thank you for the patch!

Issue #684, opened 2019-03-14
Add tests of ExpDecay


gustafsson:
https://github.com/FluxML/Flux.jl/blob/22cb732657f127416822b86b752fa06bb5b2c283/src/optimise/optimisers.jl#L310

From what I can see the value of `decay` is never changed after the constructor, did I miss something?

I think we need some tests to check the documented behavior:
> Schedule the learning rate `eta` by `decay` every `decay_step` till a minimum of `clip`.

thebhatman:
The way I understand it, the learning rate decays to {learning_rate * decay_rate ^ (step / current_step)}. Here decay is actually the decay_rate, which is a constant. This decay happens only at discrete steps, that is when current_step % steps == 0. That is what the decay_step means. This is how I understand exponential decay of learning rate. So the decay is initialized in the beginning and kept constant.

gustafsson:
@thebhatman Thanks for looking into this.

When `when current_step % steps == 0` then `eta` is changed but `decay` doesn't depend on `eta`. So effectively we could just have 
```
function apply!(o::ExpDecay, x, Δ)
  @. Δ *= o.decay
end
```
But that doesn't make much sense. I think that `ExpDecay` is broken.

MikeInnes:
I guess that's meant to be eta. cc @dhairyagandhi96

dhairyagandhi96:
Yeah, the decay needs to be swapped for eta there. We scale the eta by decay every decay_step steps till a minimum eta of clip.

dhairyagandhi96:
Good to be closed now?

ToucheSir:
Bump for close :)

darsnack:
Yes, this is fixed now.

PR #733, opened 2019-04-11
Fixed ExpDecay

thebhatman:
This patch is regarding issue https://github.com/FluxML/Flux.jl/issues/684. The credit goes to @gustafsson for pointing this out.

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -307,7 +307,7 @@ function apply!(o::ExpDecay, x, Δ)
     η = max(η * decay^(s / n), o.clip)
     o.eta = η
   end
-  @. Δ *= decay
+  @. Δ *= η
 end
 
 """
--- a/test/optimise.jl
+++ b/test/optimise.jl
@@ -56,7 +56,7 @@ end
 
 @testset "ExpDecay" begin
     w = randn(10, 10)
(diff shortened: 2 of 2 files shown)


MikeInnes:
Is there a test we can add here?

dhairyagandhi96:
We should add a test in here, but otherwise lgtm

thebhatman:
I have added tests for Exponential Decay.

dhairyagandhi96:
How about logging when the decay happened and comparing it with a known ground truth. Also probably a good idea to check that the clip value is honoured.

thebhatman:
Yes. I shall do that. But I did not understand the part about the known ground truth. What will be the ground truth?

dhairyagandhi96:
The ground truth for our purposes can be
A. An array of steps of decay (eg. decay should happen every 1000 steps, so an array along the lines of 1000:1000:10000)
B. We have to honor the clip, so assuming we do enough iterations (10^5 is plenty) we would hit the clip and go beyond. We have to ensure that once the clip is hit, we don't make any further decay, so the ground truth will be the decay parameter we provided.

dhairyagandhi96:
Thanks @thebhatman, this looks great

thebhatman:
Thank you for your help!!!

PR #735, opened 2019-04-14
Added PixelShuffle Operation

shreyas-kowshik:
Added Pixel Shuffle Operation used frequently in super-resolution networks for up-sampling.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -6,7 +6,7 @@ using Base: tail
 using MacroTools, Juno, Requires, Reexport, Statistics, Random
 using MacroTools: @forward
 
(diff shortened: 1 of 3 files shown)


dhairyagandhi96:
I think this makes more sense as a function. We aren't exactly storing any parameters, so as long as the Ops are differentiable, we should be good.

shreyas-kowshik:
@dhairyagandhi96 So where do you suggest moving this file to? I had initially modified `conv.jl` thinking it would make more sense to put it in there as all convolution related functions would be together.

dhairyagandhi96:
I would put this in perhaps in NNlib.jl.

shreyas-kowshik:
@dhairyagandhi96 Added it here : https://github.com/FluxML/NNlib.jl/pull/112

ToucheSir:
We can close this now that #1468 is merged :)

PR #738, opened 2019-04-15
Add `leaves()` to make it simple to collect all parameters in a model

staticfloat:
Useful for implementing things like L2 regularization, typically
requires a filtering afterwards such as collecting only
`AbstractArrays`, etc...

### DIFF
--- a/src/treelike.jl
+++ b/src/treelike.jl
@@ -9,10 +9,26 @@ children(x::NamedTuple) = x
 mapchildren(f, x::Tuple) = map(f, x)
 mapchildren(f, x::NamedTuple) = map(f, x)
 
+function leaves(model, params = Any[])
+  # Get all children of the current `model` object
+  children = Flux.children(model)
+  # If there are no more children, we know we are at a leaf
+  if isempty(children)
+    push!(params, model)
+  else
+    # If there are more children, recurse
+    for c in children
+      leaves(c, params)
+    end
+  end
+  # Return the `params` collection in the end
+  return params
+end
+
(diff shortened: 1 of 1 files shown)


src/treelike.jl
```diff
@@ -9,10 +9,26 @@ children(x::NamedTuple) = x
...
+  end
+  # Return the `params` collection in the end
+  return params
+end
+
 function treelike(m::Module, T, fs = fieldnames(T))
   @eval m begin
     Flux.children(x::$T) = ($([:(x.$f) for f in fs]...),)
-    Flux.mapchildren(f, x::$T) = $T(f.($children(x))...)
+    Flux.mapchildren(f, x::$T) = $T(map(f, children(x))...)
```
staticfloat: I changed this to `map` just to be consistent with the `mapchildren()` definition above on line 10

MikeInnes:
Can you talk a bit about how this compares with `params`?

staticfloat:
Oh, I didn't really know about `Flux.params()`, but it doesn't seem to work the same way:

```julia
using Flux, Metalhead
m = Metalhead.VGG19()

@show length(params(m))
@show length(leaves(m))
```

```
length(params(m)) = 0
length(leaves(m)) = 162
```

MikeInnes:
Right now `params` filters out `TrackedArray`, but we'll need to remove that restriction anyway so that things work with Zygote. Removing that should get you the behaviour you want from `leaves` as well, I think.

staticfloat:
Cool, then I think there's nothing to see here.

PR #745, opened 2019-04-19
Fix typo in Maxout

Tokazama:


dhairyagandhi96:
Thanks for catching this!

PR #747, opened 2019-04-22
Swap comma for full stop

jctops:
"ERROR: LoadError: UndefVarError: G not defined" caused by "gn,G" rather than "gn.G" in line 386. Swapping for full stop should fix this


MikeInnes:
:+1: Thanks!

PR #755, opened 2019-04-26
add some docs for onehot & onecold

Roger-luo:
I'm preparing some tutorial based on Flux... but these function used in MNIST example in model zoo are not documented (only documented on website)...

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -39,6 +39,29 @@ adapt_structure(T, xs::OneHotMatrix) = OneHotMatrix(xs.height, adapt(T, xs.data)
   cudaconvert(x::OneHotMatrix{<:CuArray}) = OneHotMatrix(x.height, cudaconvert(x.data))
 end
 
+"""
+    onehot(l, labels[, unk])
+
+Create an [`OneHotVector`](@ref) wtih `l`-th element be `true` based on possible `labels` set.
+If `unk` is given, it retruns `onehot(unk, labels)` if the input label `l` is not find in `labels` or
+it will error.
+
+## Examples
+
+```jldoctest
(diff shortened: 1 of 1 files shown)


src/onehot.jl
```diff
@@ -39,6 +39,29 @@ adapt_structure(T, xs::OneHotMatrix) = OneHotMatrix(xs.height, adapt(T, xs.data)
   cudaconvert(x::OneHotMatrix{<:CuArray}) = OneHotMatrix(x.height, cudaconvert(x.data))
 end
 
+"""
+    onehot(l, labels[, unk])
+
+Create an [`OneHotVector`](@ref) wtih `l`-th element be `true` based on possible `labels` set.
+If `unk` is given, it retruns `onehot(unk, labels)` if the input label `l` is not find in `labels` or
```
MikeInnes:
```suggestion
If `unk` is given, it retruns `onehot(unk, labels)` if the input label `l` is not find in `labels`; otherwise
```

src/onehot.jl
```diff
@@ -51,9 +74,49 @@ function onehot(l, labels, unk)
...
 
+"""
+    onecold(y[, labels = 1:length(y)])
+
+Inverse operations of [`onehot`](@ref).
+
+## Examples
+
+```jldoctest
+julia> onecold(ans, [:a, :b, :c])
```
MikeInnes: `ans` doesn't work here out of context, maybe you need to copy over a `onehot` example to here.

Roger-luo: Oh, yeah. forgot that, thanks

MikeInnes:
:+1: Thanks!

Issue #703, opened 2019-03-26
Shape-propagating Chain


MikeInnes:
It'd be nice to be able to write something like

```julia
model = @Chain(
  Input(28^2),
  Dense(32, relu),
  Dense(10),
  softmax)
```

It's a relatively minor convenience but it does avoid some redundancy when specifying chains, which is tedious to correct and easy to get wrong when trying different layer sizes.

Here's roughly how I imagine this working. The `@Chain` would expand to something like

```julia
shape = nothing
layer1, shape = fromshape(Input, shape, 10)
layer2, shape = fromshape(Dense, shape, 32, relu)
...
Chain(layer1, layer2, ...)
```

`fromshape` can then forward to an appropriate constructor or error for non-supported layers. Hopefully this strikes the right balance of simplicity/generality and we don't end up having to turn it into a full shape inference system.

chengchingwen:
Looks like it would need to define a inference system for every possible function of Chain? For the example above, it might be easy to handle the layer type. But for functions(like `softmax` above) or even anonymous function, it would really be a problem.

MikeInnes:
Yes, this is not in any way a general system. We'd just define it for `Dense`, `softmax`, `Conv` etc., things that are common with `Chain`, and people could of course hook in custom layers as well. But anonymous functions would never work with this.

It is a downside, but as long as it's clear in the docs that this is only a very light sugar that shouldn't be used for anything complex -- we should still for example show people how to use normal `Chain` -- I think that can be OK.

darsnack:
This could be quite a bit simpler now that we have `outputsize`. @DhairyaLGandhi and @mcabbott what do you think? If it's desired, I can put together a PR.

ToucheSir:
Bikeshedding syntax, would it be clearer to have something like:
```julia
buildmodel = @[macro name] Chain(
  Dense(_, 32, relu),
  Dense(_, 10),
  softmax))
...
model = buildmodel(28^2)
```
Does outputsize allow us to support arbitrary layers?

darsnack:
Yeah it should support pretty much anything. Missing primitives if there are any (e.g. `sin`) can easily be added as they come up.

mcabbott:
Neat idea to build this using `outputsize`. I like the underscore idea, as that avoids needing another syntax for every layer. I guess the macro will still have to understand most of them, but could fall back to `_ = size(x,1)`? Maybe something like
```
[...]

PR #756, opened 2019-04-26
Change `DepthwiseConv()` to use `in=>out` instead of `in=>mult`.

staticfloat:
This is an API change, but I think it makes more sense, and is more consistent with our `Conv()` API.  This also dumps the `DepthwiseConv((3,3), C_in)` API, as I'm not sure why you would want to specify only the input channel count and default the output to a channel multiplier of 1; if anything I would think you'd want to specify the channel output and leave the input to be default.  In any case, I think consistency with `Conv()` is the best thing to chase after here.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,6 @@
+# v0.9.0
+* [Depthwise comvolutional layer API changes](https://github.com/FluxML/Flux.jl/pull/756) from `in => mult` channel specification to `in => out` channel specification, and deprecates implicit `out` constructor.
+
 # v0.8.0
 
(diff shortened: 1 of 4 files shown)


MikeInnes:
Agreed on all counts. Especially with leaving out the input channel; we should prepare our APIs for this so we can eventually do #703.

It's just a shame we can't deprecate this. But I guess for the most part people will get a fairly obvious shape error, if not an assertion error.

MikeInnes:
Oh, worth adding a news item for this though!

staticfloat:
Anything else to do here?

dhairyagandhi96:
Lgtm

Issue #757, opened 2019-04-27
BatchNorm gradients return Float64 instead of Float32


aterenin:
MWE
```
x = randn(Float32, 8,8,3,16)
v = randn(Float32, 8,8,3,16)
bn = BatchNorm(3)
(y, back) = Flux.Tracker.forward(bn,x) # (Array{Float32,4}, λ)
back(v)[1] # Array{Float64,4)
```
Whatever the issue is, it's a bit subtle, because it occurs in `back` - the type of `y` in the forward pass is correct. I'll make a PR if I find out.

bhvieira:
Tracker was deprecated, the issue is not reproducible anymore. BatchNorm might have changed as well, I can't even get the forward pass.

aterenin:
Last I looked at Flux's source, the underlying issue is still present in spite of Tracker deprecation, and I suspect this will reproduce on Zygote. I am happy to finish my PR given appropriate input.

bhvieira:
For sure! I've been running down the issues tracker searching for deprecated stuff, but on this issue of yours I actually thought it could be still active due to BN (it probably is), but I couldn't run your MWE, nor even the forward pass.

```
julia> bn(x)
ERROR: UndefVarError: isnothing not defined
Stacktrace:
 [1] _isactive(::BatchNorm{typeof(identity),Array{Float32,1},Array{Float32,1},Float32}) at C:\Users\bhebl\.julia\packages\Flux\NpkMm\src\layers\normalise.jl:5
 [2] (::BatchNorm{typeof(identity),Array{Float32,1},Array{Float32,1},Float32})(::Array{Float32,4}) at C:\Users\bhebl\.julia\packages\Flux\NpkMm\src\layers\normalise.jl:182
 [3] top-level scope at none:0
```

CarloLucibello:
I don't see any conversion problem on master:
```julia
julia> x = rand(Float32, 2,3)
2×3 Array{Float32,2}:
 0.0336944  0.897612   0.230159
 0.503094   0.0609328  0.815861

julia> x = rand(Float32, 2,3);

julia> bn = BatchNorm(2);

julia> gradient(()->sum(bn(x)), params(bn)).grads
IdDict{Any,Any} with 5 entries:
  :(Main.x)         => Float32[4.76837f-7 -4.76837f-7 0.0; 2.38419f-7 0.0 0.0]
[...]

PR #758, opened 2019-04-27
Workaround for BatchNorm Float64 grad with Float32 input

aterenin:
This works around #757.

The cause is as follows: BatchNorm gradients get promoted to Float64 is due to element-wise integer division. I don't know why Tracker is doing that - I could look into it but don't understand Tracker internals particularly well. Simply performing floating point division instead of integer division eliminates the issue.


KristofferC:
See https://github.com/FluxML/Flux.jl/pull/538 for another way of working around the same issue.

aterenin:
Is there any reason to prefer one fix to the other? I can modify my PR if your fix is preferable.

MikeInnes:
Either seems fine to me. It would be nice to have a test though.

aterenin:
It seems this issue got lost in my email inbox half a year ago. Sorry about that.

I'm not sure what kind of test to write for it. Any suggestions?

aterenin:
Oops, accidentally pressed the wrong button and closed this. My apologies.

PR #762, opened 2019-05-01
CrossCor layer

ayush-1506:
Same as #423 (which could be edited since I lost access to that github account).

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -198,6 +198,76 @@ end
 
 (a::DepthwiseConv{<:Any,<:Any,W})(x::AbstractArray{<:Real}) where {T <: Union{Float32,Float64}, W <: AbstractArray{T}} =
   a(T.(x))
+"""
+    CrossCor(size, in=>out)
+    CrossCor(size, in=>out, relu)
+  
+Standard cross convolutional layer. `size` should be a tuple like `(2, 2)`.
+`in` and `out` specify the number of input and output channels respectively.
+    
+Example: Applying CrossCor layer to a 1-channel input using a 2x2 window size,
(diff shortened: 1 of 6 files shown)


src/layers/conv.jl
```diff
@@ -199,6 +199,66 @@ end
 (a::DepthwiseConv{<:Any,<:Any,W})(x::AbstractArray{<:Real}) where {T <: Union{Float32,Float64}, W <: AbstractArray{T}} =
   a(T.(x))
 
+  """
```
MikeInnes: Formatting is off here

test/cuda/cuda.jl
```diff
@@ -36,6 +36,10 @@ c = gpu(Conv((2,2),3=>4))
 l = c(gpu(rand(10,10,3,2)))
 Flux.back!(sum(l))
 
+c = gpu(CrossCor((2,2),3=>4))
+l = c(gpu(rand(10,10,3,2)))
+Flux.back!(sum(l))
```
MikeInnes: Would be nice to have some numerical tests here, e.g. comparing to a regular convolution. Also, let's have a news item for this.

dhairyagandhi96:
Not to get ahead of myself, we'll need a News item.

dhairyagandhi96:
bors try

dhairyagandhi96:
Could you do a quick conflict resolve?

ayush-1506:
Done!

dhairyagandhi96:
bors try

MikeInnes:
Can we also link `CrossCor` in the layer reference?

@avik-pal once the docs are updated, if you're happy with the tests here go ahead and merge with bors.

MikeInnes:
Thanks @ayush-1506!

PR #764, opened 2019-05-04
Fix for onecold broadcast bug

DhairyaLGandhi:
Using `mapreduce` was returning an `Array` when used with `OneHotMatrix{CuArray}`. This causes issues down the chain. `map` avoids that, and also avoids the reduction giving slightly better speed (~40x).

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -121,8 +121,7 @@ onecold(y::AbstractVector, labels = 1:length(y)) = labels[Base.argmax(y)]
 onecold(y::AbstractMatrix, labels...) =
   dropdims(mapslices(y -> onecold(y, labels...), y, dims=1), dims=1)
 
-onecold(y::OneHotMatrix, labels...) =
-  mapreduce(x -> Flux.onecold(x, labels...), |, y.data, dims = 2, init = 0)
+onecold(y::OneHotMatrix, labels...) = map(x -> Flux.onecold(x, labels...), y.data)
 
 function argmax(xs...)
   Base.depwarn("`argmax(...)` is deprecated, use `onecold(...)` instead.", :argmax)
--- a/test/onehot.jl
+++ b/test/onehot.jl
@@ -10,6 +10,13 @@ using Test
   @test onecold(A) == [3, 1, 4]
   @test onecold(a, labels) == 'C'
   @test onecold(A, labels) == ['C', 'A', 'D']
+
+  data = [:b, :a, :c]
+  labels = [:a, :b, :c]
+  hot = onehotbatch(data, labels)
+  cold = onecold(hot, labels)
+
+  @test cold == data
 end
 
 @testset "onehotbatch indexing" begin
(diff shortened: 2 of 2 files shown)


dhairyagandhi96:
bors try

KristofferC:
What was the bug? Could the existing test be tweaked or added to, to catch it?

dhairyagandhi96:
`mapreduce` is returning an `Array` with a `OneHotMatrix{CuArray}`.

I think we already have the test that is supposed to check for this https://github.com/FluxML/Flux.jl/blob/e9912280472569006acf9ae126e3f3f67e88fa47/test/cuda/cuda.jl#L43

Maybe we should add an extra one that checks for the broadcasting just like in the `accuracy` function?

MikeInnes:
The diff looks odd to me -- if this can be done with a simple `map`, why was it written with `mapreduce` before? Is there a behaviour change or was the original code just redundant?

If there's no behaviour change then this is certainly good to go.

dhairyagandhi96:
We needed to hit an optimised kernel with the GPU at the time (iirc, because it's been a while since this has been up), but things have moved on massively since then and the regular way works fine now.

No change in behaviour

darsnack:
I think this will address #864 issues as well. Any other delays on this other than possibly wanting to be consistent with #1006?

dhairyagandhi96:
Bump

MikeInnes:
I'm not totally clear on the testing situation; if there's a test for this already, why isn't it failing currently? Can we add the test that would fail without this patch?

lassepe:
A test that would fail right now is the following identity:

```julia
using Flux: onehotbatch, onecold
using Test

data = [:b, :a, :c]
labels = [:a, :b, :c]
hot = onehotbatch(data, labels)
cold = onecold(hot, labels)

@test cold == data
```

This is something that works with the new `map` version of `onecold`.

However, I don't know why the the `reduce` with `|` was necessary (I guess there is/was a good reason to do it like this). From https://github.com/FluxML/Flux.jl/issues/1214 one can at least infer that at some point `onehotbatch` used to produce matrices with eltype `Bool`.

lassepe:
Perhaps another relevant data point:

When the `mapreduce` version of `onecold` was first introduced, it seems like there was no test for this. The only thing tested is the  [1-argument-version](https://github.com/FluxML/Flux.jl/pull/612/files#diff-32a5c7576af715722d5fb1ce0db1822eR43). In that case the reduce is simply a `0 | x::Int` which should just give `x`. The 2-argument version where (potentially non-Integer) labels are passed is only tested with `Base.Matrix` not with `Flux.OneHotMatrix`.

darsnack:
I think the original reason was for some CUDA optimization. But this PR really should get some attention.

CarloLucibello:
@DhairyaLGandhi could you add the test from https://github.com/FluxML/Flux.jl/pull/764#issuecomment-646118484 and get this merged?

PR #766, opened 2019-05-07
store apply function in Chain

hossein-pourbozorg:
I replace recursive function `applychain` with [a function composition](https://docs.julialang.org/en/v1.3-dev/manual/functions/#Function-composition-and-piping-1) and store result in Chain type.
It can speed up execution a little.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -18,7 +18,8 @@ m(x) == m[2](m[1](x))
 """
 struct Chain{T<:Tuple}
   layers::T
-  Chain(xs...) = new{typeof(xs)}(xs)
+  apply::Any
+  Chain(xs...) = new{typeof(xs)}(xs, isempty(xs) ? identity : reduce(∘, reverse(xs)))
 end
 
 @forward Chain.layers Base.getindex, Base.length, Base.first, Base.last,
@@ -27,10 +28,7 @@ end
 children(c::Chain) = c.layers
 mapchildren(f, c::Chain) = Chain(f.(c.layers)...)
 
-applychain(::Tuple{}, x) = x
-applychain(fs::Tuple, x) = applychain(tail(fs), first(fs)(x))
-
-(c::Chain)(x) = applychain(c.layers, x)
+(c::Chain)(x) = c.apply(x)
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i]...)
 


MikeInnes:
Does this actually improve runtimes, or just compile time? We can likely just put an `@nospecialize` somewhere to get the same effect.

This current patch would break compatibility with stricter backends, like XLA.

hossein-pourbozorg:
I extract relevant parts to a file alongside my suggest and here are the results:
```julia
julia> versioninfo()
Julia Version 1.1.0
Commit 80516ca202 (2019-01-21 21:24 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-4702MQ CPU @ 2.20GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, haswell)
```
```julia
module MyChains

    using Base: tail

    export my_main

    struct Chain{T<:Tuple}
        layers::T
        Chain(xs...) = new{typeof(xs)}(xs)
    end

    applychain(::Tuple{}, x) = x
    applychain(fs::Tuple, x) = applychain(tail(fs), first(fs)(x))

    struct MyNewChain{T<:Tuple}
        layers::T
        apply::Any
        MyNewChain(xs...) = new{typeof(xs)}(xs, isempty(xs) ? identity : reduce(∘, reverse(xs)))
    end

    (c::MyNewChain)(x) = c.apply(x)

    function my_main()
        n::BigInt = 10_000
        seq = (
            sin,
            cos,
            exp,
        )

        for s in (Chain, MyNewChain)
            @show s
            f = s(seq...)
            @show f(n)
            sleep(1)
            @timev begin
                for i in 1:1_000_000
                    f(n)
                end
            end
        end
    end
end

using .MyChains
my_main()
```
```
s = Main.MyChains.Chain
f(n) = 2.595196602429303356279772286389813670990364682927454466749796957506537907032024
 11.750475 seconds (26.00 M allocations: 808.716 MiB, 0.22% gc time)
elapsed time (ns): 11750475102
gc time (ns):      25954007
bytes allocated:   848000024
pool allocs:       8000000
malloc() calls:    5000000
realloc() calls:   13000000
free() calls:      5000000
GC pauses:         37
s = Main.MyChains.MyNewChain
f(n) = 2.595196602429303356279772286389813670990364682927454466749796957506537907032024
 11.731370 seconds (26.00 M allocations: 808.716 MiB, 0.24% gc time)
elapsed time (ns): 11731370324
gc time (ns):      27569308
bytes allocated:   848000000
pool allocs:       8000000
malloc() calls:    5000000
realloc() calls:   13000000
free() calls:      5000000
GC pauses:         37
```

jpsamaroo:
Maybe it's just me, but the supposed improvement in performance is not made apparent to me in your benchmark. Maybe you can separate out compilation time from pre-compiled runtime by placing a single `@timev f(n)` before the longer benchmark?

MikeInnes:
Yeah, it seems like the conclusion is that the timings are identical up to noise.

Worth using BenchmarkTools for this kind of timing FWIW.

hossein-pourbozorg:
I used BenchmarkTools, and obtained this results: (output of @ benchmark)

For old Chain:
```julia
BenchmarkTools.Trial: 
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     61.923 ns (0.00% GC)
  median time:      65.420 ns (0.00% GC)
  mean time:        67.171 ns (0.00% GC)
  maximum time:     151.527 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     979
```
And for new Chain:
```julia
BenchmarkTools.Trial: 
  memory estimate:  16 bytes
  allocs estimate:  1
  --------------
  minimum time:     71.996 ns (0.00% GC)
  median time:      77.529 ns (0.00% GC)
  mean time:        88.165 ns (7.75% GC)
  maximum time:     50.552 μs (99.82% GC)
  --------------
  samples:          10000
  evals/sample:     972
```

If it's true, I offered a bad suggest :disappointed:
Sorry for my impractical pull request.

MikeInnes:
No worries. I'll close this for now but we welcome any other ideas!

PR #767, opened 2019-05-07
Some cleanup on performance tips docs

oxinabox:


docs/src/performance.md
```diff
@@ -60,7 +61,7 @@ end
 
 It is much faster to concatenate them into a matrix,
 as this will hit BLAS matrix-matrix multiplication, which is much faster than the equivalent sequence of matrix-vector multiplications.
-Even though this means allocating new memory to store them contiguously.
+The improvement is enough that it is worthwild allocating new memory to store them contiguously.
```
MikeInnes:
```suggestion
The improvement is enough that it is worthwile allocating new memory to store them contiguously.
```

docs/src/performance.md
```diff
@@ -14,18 +14,19 @@ Which means allocations occur much faster.
...
-A very artificial example using an activatioon function like
+A very artificial example using an activation function like
 
 ```
     my_tanh(x) = Float64(tanh(x))
 ```
 
 will result in performance on `Float32` input orders of magnitude slower than the normal `tanh` would,
 because it results in having to use slow mixed type multiplication in the dense layers.
+Similar can occur in the loss function during backpropagation.
```
KristofferC:
```suggestion
Similar situations can occur in the loss function during backpropagation.
```

MikeInnes:
:+1: Thanks!

PR #772, opened 2019-05-11
delete redundant section

johnnychen94:
This section is redundant to section `## Normalisation & Regularisation` in the same file


MikeInnes:
Odd that this got duplicated. Thanks for fixing it.

PR #662, opened 2019-03-06
Add Parallel and Bi layer

swiesend:
@MikeInnes I've cleaned the parallel branch and separated the peephole LSTM into another branch for another [pull request](https://github.com/FluxML/Flux.jl/pull/663).

To enable a bidirectional model `Base.reverse()` is [defined](https://github.com/swiesend/Flux.jl/blob/258140f4ab1e9839015a7326cf89db8c67d3a9c4/src/layers/parallel.jl#L101-L186) for common Flux data structures.

PR #773, opened 2019-05-12
parallel layer

adam-r-kowalski:
This is aimed to simply the case when you wish to call multiple layers in parallel rather than in sequence. For example in the resnet style approach you apply a layer and the identity function then sum them. This can be written like so

```
model = Chain(Parallel(Conv((3, 3), 3=>3, pad=1),
                       x -> x),
              sum)

inputs = randn(28, 28, 3, 1)
model(inputs)
```

If you wish to have a set of convolutional layers that can extract different sized features and then concat them together of the feature dimension.

```
model = Chain(Parallel(Conv((3, 3), 3=>3),
                       Conv((5, 5), 3=>3, pad=1),
                       Conv((7, 7), 3=>3, pad=2)),
              x -> cat(x..., dims=3))


inputs = randn(28, 28, 3, 1)
model(inputs)
```

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -59,6 +59,46 @@ function activations(c::Chain, input)
 end
 
 
+
+"""
+    Parallel(layers...)
+
+Call multiple layers / functions in parallel on a given input.
+
+```julia
+m = Parallel(x -> x^2, x -> x+1)
+m(5) == [25, 6]
+
+m = Chain(Dense(10, 5), Dense(5, 2))
+x = rand(10)
+m(x) == [m[1](x), m[2](x)]
+```
+
+`Parallel` also supports indexing and slicing, e.g. `m[2]` or `m[1:end-1]`.
+`m[1:3](x)` will calculate the output of the first three layers.
+"""
+struct Parallel{T<:Tuple}
+  layers::T
+  Parallel(xs...) = new{typeof(xs)}(xs)
+end
+
+@forward Parallel.layers Base.getindex, Base.length, Base.first, Base.last,
+  Base.iterate, Base.lastindex
+
+children(p::Parallel) = p.layers
+mapchildren(f, p::Parallel) = Parallel(f.(c.layers)...)
+
+(p::Parallel)(x) = [layer(x) for layer in p.layers]
+
+Base.getindex(p::Parallel, i::AbstractArray) = Parallel(p.layers[i]...)
+
+function Base.show(io::IO, p::Parallel)
+  print(io, "Parallel(")
+  join(io, p.layers, ", ")
+  print(io, ")")
+end
+
+
 """
     Dense(in::Integer, out::Integer, σ = identity)
 
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -6,8 +6,8 @@ using Base: tail
 using MacroTools, Juno, Requires, Reexport, Statistics, Random
 using MacroTools: @forward
 
-export Chain, Dense, Maxout, RNN, LSTM, GRU, Conv, ConvTranspose, MaxPool, MeanPool,
-       DepthwiseConv, Dropout, AlphaDropout, LayerNorm, BatchNorm, InstanceNorm, GroupNorm, 
+export Chain, Parallel, Dense, Maxout, RNN, LSTM, GRU, Conv, ConvTranspose, MaxPool, MeanPool,
+       DepthwiseConv, Dropout, AlphaDropout, LayerNorm, BatchNorm, InstanceNorm, GroupNorm,
        params, mapleaves, cpu, gpu, f32, f64
 
 @reexport using NNlib
--- a/test/layers/basic.jl
+++ b/test/layers/basic.jl
@@ -19,6 +19,14 @@ import Flux: activations
     # numeric test should be put into testset of corresponding layer
   end
 
+  @testset "Parallel" begin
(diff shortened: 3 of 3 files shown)


jpsamaroo:
I'm guessing this layer is similar to the one implemented in https://github.com/FluxML/Flux.jl/pull/662? If so, maybe @swiesend can provide a review.

adam-r-kowalski:
I think this is more similar to what Jax does. [here is there resnet implementation](https://github.com/google/jax/blob/master/examples/resnet50.py#L38).

Currently my layer doesn't actually do anything in parallel, it's more about the semantics of saying I would like to apply all these layers to my input and then do something with the result such as `sum` `reduce` `cat` etc rather then doing them in a sequential fashion.

jpsamaroo:
> Currently my layer doesn't actually do anything in parallel

AFAICT neither does #662, which I definitely think is a good thing, since adding multiple threads/processes/etc. into the mix can make things a lot more complicated. If we want actual parallelism, I think that should be done as a separate PR so as to not block this very useful functionality being added to Flux.

jpsamaroo:
Also, I think we'll need a few tests for this layer. Maybe you can draw inspiration from #662, if things are similar enough.

swiesend:
I like the `Tuple` approach from this #773.

The thing with #662 is, that it should maybe be better called `MapReduce` instead of `Parallel`, as both phases are fused into one layer. It is possible to do nothing in the reduce phase, but only by applying the `identity` function, which is quite silly.

swiesend:
The only concern I have is, that a thing like this should also be able to work smoothly for sequential models, to overcome the border between the convolutional and sequential models... or is there none?

adam-r-kowalski:
@jpsamaroo I have added some tests based on tests for the `Chain` layer. It is honestly almost identical to that layer. The only difference is that instead of calling in sequence, I do a list comprehension and call each layer on the input.

HenriDeh:
This does not seem to be compatible with `Flux.gpu`:

```
julia> m = Chain(Dense(10,5), Parallel(Dense(5,5), Dense(5,1))) |> gpu
julia> m(cu(rand(10)))
ERROR: ArgumentError: cannot take the CPU address of a CuArray{Float32,1,Nothing}
Stacktrace:
 [1] unsafe_convert(::Type{Ptr{Float32}}, ::CuArray{Float32,1,Nothing}) at /home/dehaybe/.julia/packages/CuArrays/ZYCpV/src/array.jl:212
 [2] gemv!(::Char, ::Bool, ::Array{Float32,2}, ::CuArray{Float32,1,Nothing}, ::Bool, ::CuArray{Float32,1,Nothing}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/LinearAlgebra/src/blas.jl:587
 [3] gemv!(::CuArray{Float32,1,Nothing}, ::Char, ::Array{Float32,2}, ::CuArray{Float32,1,Nothing}, ::Bool, ::Bool) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/LinearAlgebra/src/matmul.jl:463
 [4] * at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/LinearAlgebra/src/matmul.jl:66 [inlined]
 [5] (::Dense{typeof(σ),Array{Float32,2},Array{Float32,1}})(::CuArray{Float32,1,Nothing}) at /home/dehaybe/.julia/packages/Flux/oX9Pi/src/layers/basic.jl:102
 [6] Dense at /home/dehaybe/.julia/packages/Flux/oX9Pi/src/layers/basic.jl:113 [inlined]
 [7] #50 at ./none:0 [inlined]
 [8] iterate at ./generator.jl:47 [inlined]
 [9] collect(::Base.Generator{Tuple{Dense{typeof(σ),Array{Float32,2},Array{Float32,1}},Dense{typeof(elu),Array{Float32,2},Array{Float32,1}}},var"#50#51"{CuArray{Float32,1,Nothing}}}) at ./array.jl:622
 [10] (::Chain{Tuple{Dense{typeof(elu),CuArray{Float32,2,Nothing},CuArray{Float32,1,Nothing}},Parallel{Tuple{Dense{typeof(σ),Array{Float32,2},Array{Float32,1}},Dense{typeof(elu),Array{Float32,2},Array{Float32,1}}}}}})(::CuArray{Float32,1,Nothing}) at /home/dehaybe/Documents/Doctorat/Code/Julia/Stochastic Inventory Problem/src/custom_layers.jl:27
 [11] top-level scope at none:0
```

The faulty line appears to be `(p::Parallel)(x) = [layer(x) for layer in p.layers]` which creates an array on the cpu.

src/layers/basic.jl
```diff
@@ -59,6 +59,46 @@ function activations(c::Chain, input)
...
+struct Parallel{T<:Tuple}
+  layers::T
+  Parallel(xs...) = new{typeof(xs)}(xs)
+end
+
+@forward Parallel.layers Base.getindex, Base.length, Base.first, Base.last,
+  Base.iterate, Base.lastindex
+
+children(p::Parallel) = p.layers
+mapchildren(f, p::Parallel) = Parallel(f.(c.layers)...)
```
HenriDeh: c.layers should be p.layers

ToucheSir:
We now have https://github.com/FluxML/Flux.jl/pull/1462, so this can be closed.

PR #778, opened 2019-05-16
modifications to support proximal optimization

salbert83:
In some of my tests, using proximal optimization for regularization functions (like L1 and L2 norms) has been quite efficient,


ToucheSir:
L1 and L2 norm clipping landed in https://github.com/FluxML/Flux.jl/pull/1133, so this can be closed.

PR #791, opened 2019-06-10
[WIP] Integrating Zygote with Flux

thebhatman:
I have added no-op definitions for param and data. Tests for optimisers have been modified to support zygote and move away from Tracker. Temporarily I have commented out the tests for cudnn.jl and curnn.jl. I will be modifying those tests as well so that they support zygote.


MikeInnes:
Best to make this PR directly against master, so we can see the full set of changes.

thebhatman:
So this is the current status of this PR :  All tests are passing except two errorred tests. One test that is erring is the training loop test in optimise.jl and another test that is erring is the test for "no spurious promotions". The latter is because of this issue. https://github.com/FluxML/Zygote.jl/issues/228. I am trying to debug what exactly is causing the error in both the test cases.

thebhatman:
The tests are passing. But there is still one failing test in "no spurious promotions", when back is called on Float16. I have temporarily removed the test for Float16....

dhairyagandhi96:
So this seems close to done, save for the issues from https://github.com/FluxML/Zygote.jl/issues/247 now, @thebhatman?

thebhatman:
Yes! This PR is almost done. I think I review over the entire PR especially the cuda part and the tests for cudnn and curnn would be very helpful.

MikeInnes:
Great! @dhairyagandhi96 can you give it a first pass, and then I'll take a look over it?

dhairyagandhi96:
Yup!

dhairyagandhi96:
```julia
julia> m = Chain(
         LSTM(N, 128),
         LSTM(128, 128),
         Dense(128, N),
         softmax)
ERROR: type Array has no field data
Stacktrace:
 [1] getproperty(::Any, ::Symbol) at ./Base.jl:20
 [2] #LSTMCell#105(::typeof(Flux.glorot_uniform), ::Type{Flux.LSTMCell}, ::Int64, ::Int64) at /home/dhairyagandhi96/zygote_thebhatman/Flux.jl/src/layers/recurrent.jl:112
 [3] Flux.LSTMCell(::Int64, ::Int64) at /home/dhairyagandhi96/zygote_thebhatman/Flux.jl/src/layers/recurrent.jl:110
 [4] #LSTM#106(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(LSTM), ::Int64, ::Vararg{Int64,N}
where N) at /home/dhairyagandhi96/zygote_thebhatman/Flux.jl/src/layers/recurrent.jl:144
 [5] LSTM(::Int64, ::Vararg{Int64,N} where N) at /home/dhairyagandhi96/zygote_thebhatman/Flux.jl/src/layers/recurrent.jl:144
 [6] top-level scope at REPL[18]:1
```

thebhatman:
This is due to the `.data` that stayed back in LSTMCell here https://github.com/thebhatman/Flux.jl/blob/517219ba23a7f6cd448a55424a52eeb4749eb457/src/layers/recurrent.jl#L112. My bad. Will correct it.

dhairyagandhi96:
While training the char-rnn from model-zoo, we run into Zygote producing `nothing`s for the gradients for the biases

```julia
julia> gs.grads
IdDict{Any,Any} with 16 entries:
  LSTMCell(68, 128)                          => RefValue{Any}((Wi = Float32[-1.06695e-5 -0.000227912 … 0.0 0.0; -7.4389e-8 -1.857…
  LSTMCell(128, 128)                         => RefValue{Any}((Wi = Float32[0.024442 -0.0290818 … -0.0176835 -0.00408542; -0.0377…
  Float32[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0… => nothing
  Recur(LSTMCell(68, 128))                   => RefValue{Any}((cell = RefValue{Any}((Wi = Float32[-1.06695e-5 -0.000227912 … 0.0 …
  Float32[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0… => nothing
  Float32[0.0594126 0.0281752 … 0.01424 -0.… => Float32[-0.00999993 0.00999994 … 0.0099999 0.00999959; 0.00999583 -0.00999654 … -…
  Float32[-0.0441741 0.0700965 … 0.0171228 … => Float32[-0.00999064 -0.00999956 … 0.0 0.0; -0.00881501 -0.00994644 … 0.0 0.0; … ;…
  Recur(LSTMCell(128, 128))                  => RefValue{Any}((cell = RefValue{Any}((Wi = Float32[0.024442 -0.0290818 … -0.017683…
  Float32[-0.0905313 0.0952912 … -0.0840143… => Float32[0.024442 -0.0290818 … -0.0176835 -0.00408542; -0.0377915 0.0451328 … 0.02…
  Float32[-0.0140954 0.0965541 … 0.0922624 … => Float32[0.00302187 0.0114205 … 0.0175195 0.00522623; -0.00473368 -0.0178943 … -0.…
  Float32[0.0413398, 0.0967397, -0.0907992,… => Float32[0.0381598, -0.0595871, -0.00486297, -0.00818275, 0.0699488, -0.222861, -0…
  Float32[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0… => nothing
  Float32[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0… => nothing
  Float32[0.0762027 -0.171703 … 0.060722 -0… => Float32[0.0418269 0.161186 … 0.248163 0.0737012; -0.116159 -0.447983 … -0.691102 …
  Float32[-0.0484179, 0.101334, 0.0993247, … => Float32[-0.00999996, 0.00999736, -0.00999798, 0.00999998, -0.01, -0.00999999, -0.…
  Float32[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0… => Float32[0.539081, -1.49885, -1.83964, -1.45842, -1.70783, -5.60054, 0.358645, 0.4…
```

thebhatman:
So I was testing backprop and gradient computation of Zygote on all Flux layers one by one. Zygote failed for the following layers : 
```
     Layer                                                        Error Thrown

1.) Maxout                                                     AssertionError: Blockas are not in domtree order.

2.) Dropout                                                   Can't differentiate gc_preserve_end expression

3.) AlphaDropout                                           Non-differentiable function le_float

4.) BatchNorm                                               Mutating Arrays is not supported

5.) InstanceNorm                                           Mutating Arrays is not supported

6.) GroupNorm                                              Mutating Arrays is not supported

7.) ConvTranspose                                         Mutating Arrays is not supported

8.) RNN                                                         Gives nothing as gradient for a few members in params.     
                                                                     Hence update! fails.

9.) LSTM                                                        Gives nothing as gradient for a few members in params.
                                                                     Hence update! fails.

10.) GRU                                                       Gives nothing as gradient for a few members in params.
                                                                     Hence update! fails.
```

dhairyagandhi96:
For normalizing layers, maybe allocation of the output matrix would be necessary?

Can you merge Zygote#mutate to your branch and try once?

MikeInnes:
We definitely don't want to rely on the mutate branch for this. Might be useful information, but things like dropout might just need custom adjoints (which should be easy).

dhairyagandhi96:
Yeah, it's just for debugging to see what's holding us back. As it turns out `nograd` to zeros causes nothing's in the grads for the recurrent layers. How should we best handle this?

MikeInnes:
`nothing` seems like the right gradient if you call `zero`. Does it cause a problem, and if so where? We might have to e.g. update the optimiser loop to handle `nothing` somewhere.

dhairyagandhi96:
`nothing` is indeed the correct gradient for `zero` but while doing a backwards pass through an LSTM, the `nothing`s that show up in the fields need to be handled

The gradients from Zygote and Tracker are different however:

For Tracker:
```julia
julia> gs[m[1].cell.h]
Tracked 128-element Array{Float32,1}:
 -0.004275216f0
 -0.012256544f0
  0.018217055f0
...
```

With this PR
```julia

julia> gs[m[1].cell].x.h

julia>
```

MikeInnes:
Ok, yes, it makes sense that for the hidden state you'd get `nothing` after the first data is seen, so we just need to handle that in the optimiser.

Obviously numerically different gradients are a whole other issue, we should definitely narrow that down.

MikeInnes:
Closing this in favour of #669 (I've moved all commits over there).

Issue #797, opened 2019-06-24
Hypernetwork API


jessebett:
A [HyperNetwork](https://arxiv.org/pdf/1609.09106.pdf) is a neural network which outputs the parameters for another neural network. This issue is to discuss the implementation of a convenient API for expressing these in Flux.

The immediate difficulty working with hypernets is that Flux model creation prefers implicit dependence on parameters. This came up with neural ODEs and DiffEqFlux has a solution to make an implicitly parameterized function into an explicit one by `destructure`ing and `restructructure`ing. Issue #742 discusses bringing this, or some other solution to implicit->explicit, into Flux. 

Using the `DiffEqFlux: destructure, restructure` solution allows me to define a hypernet by first defining a model in the usual way, `destructure`ing to a vector of model parameters just to get the length. This gives the output dimension for a hypernet which is defined in the usual way as well.

```
hyper_dim = 20
nn_dim = 20

m_nn = Chain(Dense(1,nn_dim,tanh),Dense(nn_dim,1))
num_p = length(DiffEqFlux.destructure(m_nn))
hyper = Chain(Dense(1,hyper_dim,tanh),Dense(hyper_dim,hyper_dim,tanh),Dense(hyper_dim,num_p))

function model(x,t)
  p = hyper([t])
  m = restructure(m_nn,p)
  return m(x)
end
```
In this example the hypernet takes a single scalar hyperparameter, `t`. The `model` takes `x,t` , evaluates the hypernet on `t` to produce the parameters for the model neural net `m_nn` which is restructured and then evaluated on `x`.

@MikeInnes suggests we can have a much more elegant API for this, something like `m = Hypernet(100, Dense(10,5))` would take a 100-dim hyperparameter input and return a new `Dense(10,5)`.
The goal of this sort of API would be to have `Hypernet` be a layer in the chain which outputs a parameterized layer, e.g. `Dense`, or a chain of layers, e.g. the parameters to `m_nn` above. 
Then the model would look like `model(x,t) = hyper([t])(x)`

@MikeInnes in your proposed API you're having `Hypernet` return a layer: `Hypernet(100, Dense(10,5))` Do you anticipate difficulties having the hypernet return anything treelike? `Hypernet(100, Chain(Dense(10,5),Dense(5,10)))`?

MikeInnes:
So `Hypernet` would just be:

```julia
Hypernet(n, m) =
  Chain(Dense(n, length(destructure(m))), w -> restructure(m, w))
```

[...]

Issue #742, opened 2019-04-18
Implicit to Explicit Parameterization of Flux Models


jessebett:
I think DiffEqFlux's `restructure` and `destructure` should be brought to Flux and documented as a way of defining Flux models as implicit, but then converting them to explicitly depend on their parameters. 

```
nn = Dense(1,1,tanh)
p = destructure(nn)
params = Params([p])
function f(z, p)
    m = restructure(nn,p)
    return m(z)
end
```

This can be useful for instance when defining hypernets -- networks whose parameters are the output of another network.

yuehhua:
This should be solved in #986

PR #799, opened 2019-06-28
Towards Hypernet Layer

jessebett:
Implementation towards the discussion in #797 including the explicit/implicit parameterization tools discussed in #742. 

@MikeInnes Can you take a look at the `HyperNet` struct. In particular, do you think it should have type annotation like `HyperNet{H,M}` I don't know how to get such a thing working.

Also, I ran into a problem where `restructure` expects the parameters to be tracked going in. This is annoying as it fails silently, and the resulting model is just reparamterized with nothing as you can see in the broken test. If you supply `restructure` with an untracked array it will create the model with an empty `Any[]` array and fail silently. The check for `TrackedArray` was included in the DiffEqFlux implementation. @ChrisRackauckas do you remember why? If this is required it should produce an error possibly.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -98,6 +98,35 @@ function batchseq(xs, pad = nothing, n = maximum(length(x) for x in xs))
   [batch([xs_[j][i] for j = 1:length(xs_)]) for i = 1:n]
 end
 
+# Parameterization
+"""
+Takes a model and returns a vector of its parameters and a function which
+reparameterizes a vector of the same size producing a new model with these
+parameters.
+"""
+function destructure(m)
+  ps = []
+  mapleaves(m) do p
+    p isa TrackedArray && push!(ps, p)
+    return p
+  end
+  return vcat(vec.(ps)...), ps->restructure(m,ps)
+end
+
+"""
+Takes a model `m` and a vector of parameters `ps` and returns a new model like
+`m` parameterized by `ps`.
+"""
+function restructure(m, ps)
+  i = 0
+  mapleaves(m) do p
+    p isa TrackedArray || return p #TODO: should this be robust to restructuring with untracked?
+    p = reshape(ps[i.+(1:length(p))], size(p))
+    i += length(p)
+    return p
+  end
+end
+
 # Other
 
 """
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -9,6 +9,7 @@ using MacroTools: @forward
 export Chain, Dense, Maxout, RNN, LSTM, GRU, Conv, CrossCor, ConvTranspose, MaxPool, MeanPool,
        DepthwiseConv, Dropout, AlphaDropout, LayerNorm, BatchNorm, InstanceNorm, GroupNorm,
        SkipConnection,
+       HyperNet, HyperDense,
        params, mapleaves, cpu, gpu, f32, f64
 
 @reexport using NNlib
@@ -33,6 +34,7 @@ include("layers/basic.jl")
 include("layers/conv.jl")
 include("layers/recurrent.jl")
 include("layers/normalise.jl")
+include("layers/hypernet.jl")
 
 include("data/Data.jl")
 
--- a/src/layers/hypernet.jl
+++ b/src/layers/hypernet.jl
@@ -1,0 +1,22 @@
+using Flux: @treelike
+
+# should this be HyperNet{H,M} ?
+struct HyperNet
+    h
+    m
+end
+
+@treelike HyperNet
+
+function (H::HyperNet)(t)
+    _, re = destructure(H.m)
+    ps = H.h(t)
+    return re(ps)
+end
+
+function HyperDense(in_dim,m,σ)
+    p,_ = destructure(m)
+    return Dense(in_dim,length(p),σ)
+end
+
+HyperDense(in_dim,m) = HyperDense(in_dim,m,identity)
--- a/test/layers/hypernet.jl
+++ b/test/layers/hypernet.jl
@@ -1,0 +1,21 @@
+using Flux
+using Flux: destructure,restructure
+using Test
+
+
+@testset "Correct Dimensions for HyperDense" begin
+    m = Chain(Dense(4,5),Dense(5,2))
+    ps, re = destructure(m)
+    hyper = Chain(Dense(1,10,tanh),HyperDense(10,m))
+    @test length(hyper([1])) == length(ps)
+    hyper = Chain(Dense(1,10,tanh),HyperDense(10,m,tanh))
+    @test length(hyper([1])) == length(ps)
+end
+
+@testset "Define HyperNet for Chain" begin
+    m = Chain(Dense(5,10,σ),Dense(10,2))
+    h = Chain(Dense(1,10,tanh),HyperDense(10,m))
+    hyper = HyperNet(h,m)
+    x = rand(5)
+    @test restructure(m,h([1]))(x) == hyper([1])(x)
+end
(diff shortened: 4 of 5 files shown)


ChrisRackauckas:
I do not remember why.

MikeInnes:
The `TrackedArray` stuff is following the implementation of `params`. That does need to be changed in future but for now, being consistent is fine.

With your definition of `Hypernet`, I'm not sure there's much purpose in passing `h`, since `Hypernet(h, m)` could equally be written `Chain(h, Hypernet(m))`. Even with that change I suspect this definition is less useful than the one proposed in #797 (what you've now called `Hyperdense`), because it doesn't avoid you having to call `destructure` yourself in order to get the number of parameters.

e.g. now:

```julia
ps, _ = destructure(m)
h = Chain(..., Dense(10, length(ps))
m = Hypernet(h, m)
```

vs.
```julia
ps, re = destructure(m)
m = Chain(..., Dense(10, length(ps)), re)
```

Is there a particular use case where you see `Hypernet` (without the dense layer) adding convenience?

jessebett:
@MikeInnes it's possible I am misunderstanding your question, but it seems like you're suggesting that Hypernets should always look like a `Chain(...,Dense(_,length(ps))`. However, I wanted to leave room for hypernets with different structure, not sure whether my proposed implementation accomplishes this. So here I am passing `h` as the full parameter network, with `HyperDense` as a convenience layer. 

To answer your last question, how would you use the parameter network defined in your third option `Chain(..., HyperDense(10, m))` ? I saw `Hypernet` being used as a wrapper for the restructuring of the output of `Chain(..., HyperDense(10, m))` into the parameters of `m`. Are you proposing all that restructuring gets rolled into `HyperDense` and have that output be the new parameterized `m`?

MikeInnes:
To clarify, I'm definitely on board with hypernet models that don't use `Dense` layers. But what I'm exploring above is whether the `Hypernet(h, m)` interface (example 1) is actually more concise and clear than just using `restructure` directly (example 2). Unless I'm missing something, it seems not, in which case we can just add `restructure` alone.

I had misread your definition of `Hyperdense`, but yes, I'm imagining that layer returning a model. Although it doesn't cover all hypernets it seems like a useful special case that does save some code when it's applicable (example 3).

jessebett:
Ah I see. So currently `HyperDense` is a convenience function which given `m` returns a Dense layer with `length(destructure(m)[1])` output dimension. What you're proposing is to also have that `HyperDense` layer handle restructuring those parameters back into `m` and returning that new model? 

I agree, that would remove the need for `HyperNet`. I can try that.

jessebett:
@MikeInnes  As written the `HyperDense` gives the `Dense` layer, and the restructuring is handled by `HyperNet`. This separation of concerns allows the hypernet parameters (including those of the HyperDense layer) to be easily accessed with `params(Chain(...,HyperDense(10,m)))`. 

I'm having trouble approaching the API you're proposing (not that I disagree with it). How do I implement this in such a way that the parameters for `HyperDense` will be those of that final dense layer, and not the parameterized model produced? How do I expose those to `params`, I thought `treelike` needed a struct.

src/utils.jl
```diff
@@ -98,6 +98,35 @@ function batchseq(xs, pad = nothing, n = maximum(length(x) for x in xs))
...
+end
+
+"""
+Takes a model `m` and a vector of parameters `ps` and returns a new model like
+`m` parameterized by `ps`.
+"""
+function restructure(m, ps)
+  i = 0
+  mapleaves(m) do p
+    p isa TrackedArray || return p #TODO: should this be robust to restructuring with untracked?
```
jessebett: @MikeInnes @ChrisRackauckas Actually this is incredibly annoying. This resulted in me spending quite a long time chasing down an effectively invisible bug. Since `==` does not care about tracked vs untracked I was running into the case where `restructure(p_tracked) != restructure(p_untracked)` but `p_tracked==p_untracked`

MikeInnes: That seems like it comes down reference and value equality being used for models and arrays, rather than being due to this function; unless I've misunderstood the issue.

MikeInnes:
I'm not sure I understand the issue; you already have the right definition in `HyperNet`, you more-or-less just need to tweak the constructor so that `h` is a dense layer. Then `params` will pull out the parameters of that layer.

src/utils.jl
jessebett: Not exactly sure what you mean. This problem showed up when I tried to use a hypernet in DiffEqFlux dynamics. In the case where I was defining `dudt_` for tracked arrays it was fine, because the parameters being restructured are tracked. However the dispatch for `dudt_` on untracked arrays silently behaved completely different. The resulting hypernet was not the same, and also very hard to debug. 
```
using Flux
using Flux: destructure

m = Chain(Dense(1,10),Dense(10,1))
h = Chain(Dense(1,5),HyperDense(5,m))

H = HyperNet(h,m)

p, re = destructure(H)

re(p)([1.])([1.]) #0.08 tracked
re(Flux.data(p))([1.])([1.]) #-0.97
```

I would expect these to be equal up to tracking, but they are not.

MikeInnes: I just mean that I can reproduce the "parameters equal but models not equal" behaviour without `restructure`; in that sense it's expected, though I now see that you're not talking about literal equality though, but equality in terms of the output of the model (and the bug you describe isn't caused by equality semantics).

<details>

```julia
julia> W = randn(5, 10); b = rand(10);

julia> Dense(W, b) == Dense(W, b)
true

julia> Dense(W, b) == Dense(copy(W), copy(b))
false
```
</details>

It would be really useful for `re` (as in the returned version of `restructure`) to have a length check on its input. If you do that then this test case will produce an error rather than returning different results. Secondly, `HyperNet` should close over `re`, not the target model itself; if you do that then this test case will behave correctly.

jessebett: I'm working on addressing the other comments now, but let me just point out that I do not believe this is a situation of the paramters being equal but the models being unequal in the way you've described it. I'm not testing the equality between the two models, I'm testing the equality between the outputs of the models with equal (up to tracking) parameters and inputs.

A fair comparison should have looked like:
```
julia> using Flux: destructure
julia> using Flux: data
julia> W = randn(10,5); b = rand(10);
julia> x = randn(5);

julia> Dense(W,b)(x) == Dense(copy(W),copy(b))(x)
true

julia> Dense(W,b)(x) == Dense(data(W),data(b))(x)
true
```

I was going to write another test to show that this isn't true for de/restructured, but since `W` and `b` are untracked this shows the behaviour which I did not expect more clearly:
```
julia> p, re = destructure(Dense(W,b));

julia> p
0-element Array{Any,1}
```

I know you are saying that I need to do a length check for `p` when I restructure it, which is fine and I will do this. However, is the intended behaviour of `destructure` to only get the Tracked parameters?

jessebett: I've resolved the request to assert restructured `p` length. @MikeInnes can you check that this is a reasonable approach to that?

The above is still the behaviour of `destructure` on untracked `Dense`

jessebett:
I have made the changes you suggested that `HyperDense` returns the `Hypernet`. Also now `HyperNet` takes in the `restructure` function, not the target model itself, which is much nicer.

src/utils.jl
MikeInnes: Yeah, as mentioned in another comment, I do think the behaviour of `destructure` needs to change (it will have to when we switch to Zygote and get rid of tracking). But I think this is the right behaviour in the short term given that its consistent with what `params` and training consider to be parameters.

Silent numerical errors are of course of immediate concern, but having the shape check should mean that those errors are at least no longer silent.

XVilka:
There are conflicts now that should be resolved.

Issue #803, opened 2019-07-06
prefor does not update seen


DrChainsaw:
I'm guessing that IdSet is there to avoid infinite recursion, but nothing is ever put into it.

For example:
```julia
using Flux
mutable struct A a end
Flux.@treelike A

trouble = A(nothing)
trouble.a = trouble

params(trouble)
ERROR: StackOverflowError:
```

Happens on 0.8.3, but source on master looks the same to me.

MikeInnes:
Yes, I think also to avoid collecting parameters twice from layers that are used multiple times, which would be another interesting test case to come up with.

Would be happy to give pointers if you're interested in submitting a PR for this.

DrChainsaw:
Thanks,

Sorry for not sending a PR right away. I'm new to the language so I wanted to make sure there was no other intention.

PR sent: #805

MikeInnes:
Closed by #805.

PR #805, opened 2019-07-08
Fix for #803

DrChainsaw:
The case with same layer used multiple times was already handled in the p collector function in params. 

Added a test for it anyways.


MikeInnes:
Perfect, thankyou!

Issue #807, opened 2019-07-09
Initial state in RNNs should not be learnable by default


tanhevg:
Currently, the RNN cells are initialised as `param`'s. (e.g. [here](https://github.com/FluxML/Flux.jl/blob/3d5b76c0dfe9d9e885aa5ea62b312833a8d0759a/src/layers/recurrent.jl#L86) and [here](https://github.com/FluxML/Flux.jl/blob/3d5b76c0dfe9d9e885aa5ea62b312833a8d0759a/src/layers/recurrent.jl#L125-L126)). This causes the initial state to be modified during the backprop,  which can in turn affect the model when `reset!` is called. 

The default behaviour of the initial cell state should be for it to stay constant and not to be affected by the backprop. Having it learned, as per now, is still useful in some contexts, so this should stay as an option

mkschleg:
Has there been any updates on this? When I'm using RNNs with `Flux.reset!`, there are gradients like mentioned at the end of #808. My workaround is to just keep a copy of the initial hidden state around outside the recur cell and use that to reset the hidden state.

CarloLucibello:
@jeremiedb has this been fixed?

jeremiedb:
Treating the initial state as learnable parameters is still the default behavior for RNN, nothing was changed in the latest PR.  

My position on the subject is however that the initial state should continue to be treated as learnable parameters. It's debatable whether one case is more prevalent to the other, on my end, for NLP or time-series, learnable has been the desired case. 
The more objective argument I would have, is that the CUDNN RNN handles initial states as learnable parameters, and as such I think it makes it an expected default behavior. Adding that ignoring the initial state as a learnable parameter is fairly trivial, I have some difficulty seeing how changing the current behavior brings improvement. 

I could perhaps add a quick section in the docs about that initial state handling, given that I skipped discussing explicitly that question.

jeremiedb:
Actually, an option to make learnable parameters a more first class citizen could be to use to same approach taken with the `bias` option for `Dense` layer. So we could add `learnable_init=true` option in the RNN cells for that purpose. It would result in the initial state to be set to Zeros, that is non-learnable zeros if `learnable_init=false`.

CarloLucibello:
Unless the are some cases in which one needs a non-learnable non-zero initial state, it seems a good solution. 
That and showing in the docs some examples along the lines of 
```julia
# Non-trainable state0 for all RNN cell
[...]

Issue #648, opened 2019-02-27
Backprop through time


MikeInnes:
Continuing our series "cool things we can't have yet", and inspired by [this comment](https://github.com/FluxML/Flux.jl/pull/607#issuecomment-461889413) I was thinking about how we'll expose BPTT. Currently, given a forward pass like this:

```julia
for word in seq
  loss += model(word)
end
loss
```

If we don't want to backprop over the whole sequence at once (`gradient` outside the loop) or over only a single step at a time (`gradient` inside the loop) then we need to split the loop as follows:

```julia
for chunks in seq
  θ̄ = gradient() do
    for word in chunk
      loss += model(word)
    end
    loss
  end
end
```

An alternative to this is to just expose primitives that let us fiddle with time steps directly. Consider:

```julia
record = Capacitor(5)
for x in seq
  θ̄ = gradient() do
    record() do
      model(x)
    end
  end
end
```

Alright, bear with me here. This is written as if we were backprop-ing only across a single time step at a time, but with model evaluation wrapped in `record`. The idea is that `record` will log 5 previous backpropagators for the closure it is passed, and then chain these together for the backwards pass, which means we can actually backpropagate through `n` *previous* iterations of the loop -- i.e. backpropagation through time.

What's cool about this is that it makes BPTT completely orthogonal to the structure of the forward pass. The recorder can equally well be set up to backprop the last `n` steps each iteration (sliding window BTTF) or only every `n`th iteration (normal BTTF), or anything in between, and this can be set up differently for different parts of the model. It also isn't specific to any particular RNN implementation, e.g. this will work even though we have to backprop through `h` over loop iterations:

```julia
record = Capacitor(5)
h = ...
for word in seq
  θ̄ = gradient() do
    record() do
      y, h = model(word)
      loss(word, y)
    end
  end
end
```

The main question is whether this is actually going to be intuitive for people (who aren't travelling at 88mph). If it looks weird right now I think that's partly because we're not used to using `gradient` this way, so getting used to that will make the extra feature easier to reason about. At least for sliding windows, I think it's strictly better than flow-based alternatives.

philtomson:
What about the _train_step!_ proposed at: https://github.com/FluxML/Flux.jl/pull/607#issuecomment-462070498 ?

I'm trying to translate some PyTorch that does BPTT and in the code I'm translating (ENAS-Pytorch) they seem to have put in an explicit loop inside their _forward_ function to do the time steps (35 time steps in this case).

[...]

PR #808, opened 2019-07-09
Do not learn RNN initial state by default

tanhevg:
Do not learn RNN initial state during backprop by default; leave this as an option
#807

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 docs/build/
 docs/site/
 deps
+.tags*
--- a/docs/src/models/recurrence.md
+++ b/docs/src/models/recurrence.md
@@ -124,3 +124,11 @@ function loss(x,y)
   return l
 end
 ```
+
+## Learning the Initial State
+
+By default, the hidden state of RNN cells is initialised to zero, and is reset to zero every time `Flux.reset!` is called. Some models treat this initial state as a learnable parameter, which must be optimised during training. For such models, *learn_initial_state!* must be invoked:
+
+```julia
+Flux.learn_initial_state!(m)
+```
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -69,22 +69,34 @@ Assuming you have a `Recur` layer `rnn`, this is roughly equivalent to
 """
 reset!(m) = prefor(x -> x isa Recur && (x.state = x.init), m)
 
+"""
+    learn_initial_state!(rnn)
+
+Make the initial state of the RNN a learnable parameter, such that its gradients
+are taken and its value is updated during training
+"""
+learn_initial_state!(rnn) = prefor(x -> x isa Recur && learn_initial_state_!(x), rnn)
+
+function learn_initial_state_!(rnn)
+  rnn.state = rnn.init = param(rnn.init)
+end
+
 flip(f, xs) = reverse(f.(reverse(xs)))
 
 # Vanilla RNN
 
-mutable struct RNNCell{F,A,V}
+mutable struct RNNCell{F,A,V,S}
   σ::F
   Wi::A
   Wh::A
   b::V
-  h::V
+  h::S
 end
 
 RNNCell(in::Integer, out::Integer, σ = tanh;
         init = glorot_uniform) =
   RNNCell(σ, param(init(out, in)), param(init(out, out)),
-          param(init(out)), param(zeros(out)))
+          param(init(out)), zeros(out))
 
 function (m::RNNCell)(h, x)
   σ, Wi, Wh, b = m.σ, m.Wi, m.Wh, m.b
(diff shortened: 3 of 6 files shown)


MikeInnes:
This is a fairly big change, given that it could break people's models, and it isn't obvious to me that not learning by default is better than learning by default. It'd be good to have some discussion of the pros and cons here.

The other important thing is that any new APIs we expose for this kind of thing need to be reasonably compatible with the new Zygote-y way of doing RNNs, which will land soon (#648). Otherwise we're adding this just to soon remove it again.

tanhevg:
This PR came up after a Slack thread with @oxinabox . My main argument "for" would be that  I have never come across trainable initial state in literature, including articles that are referenced from [Flux documentation](https://fluxml.ai/Flux.jl/stable/models/recurrence/) (e.g. [this one](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)). So I thought that keeping the initial state constant at zero is the default convention in the DL community. Happy to change it the other way around, if persuaded. 

As for "Zygote-y way" - this change does not go inside `Tracker`, so from (briefly) reading the [intro to Zygote](https://github.com/FluxML/Zygote.jl) I thought that the API should still be compatible. Please let me know if that is not the case. Is there a place where I could have a look how Flux with Zygote is shaping up?

MikeInnes:
Zygote completely removes any tracking or notion of parameters, _only_ about what happens inside the call to `gradient`, so that's likely to turn the intuition about this a bit on its head.

The good news is that the Zygote version of this will actually do what you want by default.

```julia
m = LSTM(...)
reset!(m) # redundant here but usually you'd want this per sample
gradient(m) do m
  for x in xs
    m(x) ...
```

Inside `gradient`, the cell's hidden state parameter is never actually touched (since we copied it to `Recur.state`), so the gradient is zero. Things are actually different if you do:

```julia
m = LSTM(...)
gradient(m) do m
  reset!(m)
  for x in xs
    m(x) ...
```

Since now there's an explicit link between the state of the RNN and the parameter of the cell, so this trains your hidden state.

Unfortunately there's no way to simulate this distinction with Tracker (part of the reason that Zygote exists). So we'll have to wait for that; in the mean time, I think it'd be OK to add a (non-breaking) utility around this if it's needed. But I think you can already get that, effectively, with `truncate!`, so that might be enough in the short term.

tanhevg:
> But I think you can already get that, effectively, with truncate!, so that might be enough in the short term.

That would indeed be the case had `truncate!` erased the gradients for `init`. But it does not; it only clears them for `state`.  Should both of them be erased at the same time?

As a side note, I got a completely wrong idea of what `truncate!` does from the documentation. The doc does not mention that it affects only the hidden state; I thought that model weights would also be truncated. Also, I thought that the gradient will resume accumulating after its history is wiped by `truncate!`, whereas in reality there will be no gradient at all.

>  I think it'd be OK to add a (non-breaking) utility around this if it's needed. 

If Zygote will break things anyway, I think there is even less value in maintaining the current behaviour, as opposed to maintaining the mathematical conventions.

MikeInnes:
Your understanding of what `truncate!` is supposed to do from the docs seems right to me. It should drop gradients for *anything that was used to calculate* `state`, which includes model weights from past time steps as well as `init`.

What could be happening here is that `state === init` is a bit of an edge case that we're potentially not handling correctly. If that's the case then it should be easy to fix.

tfburns:
Initial states should be trainable. These are used in some RNN literature to provide initial "input" or "bias" for a sequence.

oxinabox:
@tfburns the question is not: should it be trainable, but rather should it be trainable by default.
(To which I say it should not.)

oxinabox:
why is this closed now?

tanhevg:
accidental force push to my fork. reopening

dhairyagandhi96:
Would be good to update it as per master with Zygote

MikeInnes:
This is no longer be necessary for the reasons mentioned in https://github.com/FluxML/Flux.jl/pull/808#issuecomment-510864610. We definitely do need more docs and tooling around RNNs with Zygote to show how to do these things, though.

AlexLewandowski:
> Zygote completely removes any tracking or notion of parameters, _only_ about what happens inside the call to `gradient`, so that's likely to turn the intuition about this a bit on its head.
> 
> The good news is that the Zygote version of this will actually do what you want by default.
> 
> ```julia
> m = LSTM(...)
> reset!(m) # redundant here but usually you'd want this per sample
> gradient(m) do m
>   for x in xs
>     m(x) ...
> ```
> 
> Inside `gradient`, the cell's hidden state parameter is never actually touched (since we copied it to `Recur.state`), so the gradient is zero. Things are actually different if you do:
> 
> ```julia
> m = LSTM(...)
> gradient(m) do m
>   reset!(m)
>   for x in xs
>     m(x) ...
> ```
> 
> Since now there's an explicit link between the state of the RNN and the parameter of the cell, so this trains your hidden state.
> 
> Unfortunately there's no way to simulate this distinction with Tracker (part of the reason that Zygote exists). So we'll have to wait for that; in the mean time, I think it'd be OK to add a (non-breaking) utility around this if it's needed. But I think you can already get that, effectively, with `truncate!`, so that might be enough in the short term.

@MikeInnes  Any idea why the following code snippet computes a gradient for the initial state? Of course, I can just change my code to use `gradient(m) do m` but I'm curious why this is different.

```julia
rnn = Flux.RNN(2, 3)
seq = [rand(2) for i = 1:3]

grads = gradient(Flux.params(rnn)) do
    loss = 0f0
    for s in seq
        loss += sum(rnn(s))
    end
    loss
end

grads[rnn.init]
```

oxinabox:
@AlexLewandowski i susggest discussion this on #807

PR #814, opened 2019-07-24
Rewrite `binarycrossentropy`

Sleort:
The current `binarycrossentropy` introduces an epsilon factor for numerical stabilization. That shouldn't be necessary in this rewrite. As a bonus, now only one call to `log` is needed...

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -13,17 +13,16 @@ function logitcrossentropy(logŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight
 end
 
 """
-    binarycrossentropy(ŷ, y; ϵ=eps(ŷ))
-
-Return `-y*log(ŷ + ϵ) - (1-y)*log(1-ŷ + ϵ)`. The ϵ term provides numerical stability.
-
+    binarycrossentropy(ŷ, y)
+Return `-log(y*ŷ + (1-y)*(1-ŷ))`, which is the same as `-y*log(ŷ) - (1-y)*log(1-ŷ)`
+    for `y ∈ (0,1)` but numerically more stable.
     julia> binarycrossentropy.(σ.([-1.1491, 0.8619, 0.3127]), [1, 1, 0.])
     3-element Array{Float64,1}:
     1.4244
     0.352317
     0.86167
 """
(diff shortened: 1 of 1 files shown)


MikeInnes:
The tests need to be changed so that they pass.

Would appreciate a review from someone who's worked on this code previously, maybe @boathit?

Sleort:
I just checked the tests, which reads
```
  logŷ, y = randn(3), rand(3)
  @testset "binarycrossentropy" begin
    @test binarycrossentropy.(σ.(logŷ), y; ϵ=0) ≈ -y.*log.(σ.(logŷ)) - (1 .- y).*log.(1 .- σ.(logŷ))
    @test binarycrossentropy.(σ.(logŷ), y) ≈ -y.*log.(σ.(logŷ) .+ eps.(σ.(logŷ))) - (1 .- y).*log.(1 .- σ.(logŷ) .+ eps.(σ.(logŷ)))
  end
```
In other words, they assume that `y` can be any number on the unit interval, not just 0 or 1 as I assumed. I'm confused. Is that really what we want from a *binary* cross entropy?

pshashk:
Label smoothing can improve generalization.
https://arxiv.org/abs/1906.02629
Also used in GANs for discriminator.

Sleort:
Hmm... Fair enough. One could argue that it's not really a *binary* cross entropy in that case, but nitpicking aside, maybe the restriction to only `y=0` and `y=1` is too rigid then? (Despite the few advantages of my version of `binarycrossentropy`...)

pshashk:
If your version is faster it can be dispatched when `y` is integer or boolean. But that needs to be benchmarked (both forward and backward pass).

MikeInnes:
I think the "binary" comes from it being a bernoulli distribution (binary outcome) rather than because the parameter is binary. Perhaps it should be `bernoullicrossentropy` (or even `crossentropy(Bernoulli(y), Bernoulli(yhat))`?), but the term is standard.

An alternative to dispatch would be to just check `yhat in (1, 0)` and branch.

Sleort:
I think you misunderstood. The parameter `yhat` should (of course) be allowed to take any value on the unit interval. I was considering `y`...

However, checking with [authoritative sources](https://en.wikipedia.org/wiki/Cross_entropy) I realize that my understanding of cross entropy may not have been complete [see Eq. (1)] so I'll concede. (Returning to the nitpicking, it would maybe be better if the inputs where denoted by something like `p` and `q` instead of `y` and `yhat`, then..?)

PR #818, opened 2019-07-26
Added Huber loss

dhairyagandhi96:


src/layers/stateless.jl
```diff
@@ -12,6 +12,14 @@ function logitcrossentropy(logŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight
   return -sum(y .* logsoftmax(logŷ) .* weight) * 1 // size(y, 2)
 end
 
+function smooth_l1_loss(y, fx; δ = 1)
+  α = abs(y - fx)
+  abs(α) <= δ && return 0.5f0 * α ^ 2
+  δ * α - (0.5f0 * δ ^ 2)
+end
+
+huber_loss = smooth_l1_loss
```
hossein-pourbozorg: You should use [const](https://docs.julialang.org/en/latest/base/base/#const) for alias functions, And write:
```julia
const huber_loss = smooth_l1_loss
```
It can improve performance!

hossein-pourbozorg: Example:
```julia
julia> using BenchmarkTools
julia> fact = factorial;
julia> const fact2 = factorial;
julia> @btime factorial(5);
  3.159 ns (0 allocations: 0 bytes)
julia> @btime fact(5);
  20.700 ns (0 allocations: 0 bytes)
julia> @btime fact2(5);
  3.782 ns (0 allocations: 0 bytes)
```

dhairyagandhi96: Thanks!

src/layers/stateless.jl
```diff
@@ -12,6 +12,14 @@ function logitcrossentropy(logŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight
   return -sum(y .* logsoftmax(logŷ) .* weight) * 1 // size(y, 2)
 end
 
+function smooth_l1_loss(y, fx; δ = 1)
```
MikeInnes: We shouldn't require unicode to use the API here; best to pick an ascii name.

Also best not to have multiple names. Is there precedent in other libraries for this? Could we just call it `huber` unambiguously?

PR #827, opened 2019-07-31
Momentum doesn't need params

ChrisRackauckas:


dhairyagandhi96:
Thanks for the patch!

PR #832, opened 2019-08-05
Fix #816: unnecessary typeasserts in implementations of apply!

### LINKED ISSUE #816: Unnecessary typeasserts in Flux.Optimise.apply! cause training to fail
Hi, all --

Here's an MWE where I try to learn a linear transform using a custom `AbstractMatrix` subtype:

```julia
using Flux
import Zygote: @adjoint, forward, gradient

struct MyMatrix{T, A <: AbstractMatrix{T}} <: AbstractMatrix{T}
    W :: A
    
    MyMatrix(W::AbstractMatrix) = begin
        # This example is a little contrived -- MyMatrix just wraps around copy(W')'.
        # If we skip this line then the rest of this code works fine.
        W = copy(W')'
        new{eltype(W),typeof(W)}(W)
    end
end

Flux.params(A::MyMatrix) = Flux.Params([A])
Tracker.istracked(::MyMatrix) = true   # Make Tracker think MyMatrix is a TrackedArray

Base.size(A::MyMatrix, args...) = size(A.W, args...)
Base.getindex(A::MyMatrix, args...) = getindex(A.W, args...)
Base.setindex!(A::MyMatrix, args...) = setindex!(A.W, args...)
Base.:(*)(A::MyMatrix, x::AbstractVector) = A.W * x
Base.:(*)(A::MyMatrix, x::AbstractMatrix) = A.W * x
@adjoint Base.:(*)(A::MyMatrix, x::AbstractVector) = forward(Base.:(*), A.W, x)
@adjoint Base.:(*)(A::MyMatrix, x::AbstractMatrix) = forward(Base.:(*), A.W, x)

model = Dense(randn(8,8) |> MyMatrix, zeros(8) |> param, identity)

loss(x,y) = Flux.mse(model(x),y)
∇ = gradient(() -> loss(randn(8), randn(8)), params(model))

### THIS WORKS:
Flux.Optimise.update!(ADAM(), params(model), ∇)

### THIS FAILS:
Flux.Optimise.update!(RMSProp(), params(model), ∇)
```
The error caused by the last line is due to a typeassert in [`apply!(o::RMSProp, x, Δ)`](https://github.com/FluxML/Flux.jl/blob/master/src/optimise/optimisers.jl#L83). [`apply!(o::ADAM, x, Δ)`](https://github.com/FluxML/Flux.jl/blob/master/src/optimise/optimisers.jl#L101) doesn't contain any typeasserts and thus works fine.

There are similar typeasserts in the definitions of `apply!` for `Momentum`, `Nesterov`, and `ADAGrad`, but not in any of the other optimizers. Would it be possible to get rid of these? Or are they actually necessary?

Thanks!

kernelmethod:
As I mention in #816, typeasserts of the form `get!(...)::typeof(data(x))` cause training to fail in some edge cases involving custom layers. This change would just remove the typeasserts from the definitions of `apply!` for the optimisers `Momentum`, `Nesterov`, `RMSProp`, and `ADAGrad` (the only optimisers where #816 is relevant).

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -37,7 +37,7 @@ Momentum(η = 0.01, ρ = 0.9) = Momentum(η, ρ, IdDict())
 
 function apply!(o::Momentum, x, Δ)
   η, ρ = o.eta, o.rho
-  v = get!(o.velocity, x, zero(x))::typeof(data(x))
+  v = get!(o.velocity, x, zero(x))
   @. v = ρ * v - η * Δ
   @. Δ = -v
 end
@@ -57,7 +57,7 @@ Nesterov(η = 0.001, ρ = 0.9) = Nesterov(η, ρ, IdDict())
 
 function apply!(o::Nesterov, x, Δ)
   η, ρ = o.eta, o.rho
-  v = get!(o.velocity, x, zero(x))::typeof(data(x))
+  v = get!(o.velocity, x, zero(x))
   d = @. ρ^2 * v - (1+ρ) * η * Δ
   @. v = ρ*v - η*Δ
   @. Δ = -d
@@ -80,7 +80,7 @@ RMSProp(η = 0.001, ρ = 0.9) = RMSProp(η, ρ, IdDict())
 
 function apply!(o::RMSProp, x, Δ)
   η, ρ = o.eta, o.rho
-  acc = get!(o.acc, x, zero(x))::typeof(data(x))
+  acc = get!(o.acc, x, zero(x))
   @. acc = ρ * acc + (1 - ρ) * Δ^2
   @. Δ *= η / (√acc + ϵ)
 end
@@ -147,7 +147,7 @@ ADAGrad(η = 0.1) = ADAGrad(η, IdDict())
 
 function apply!(o::ADAGrad, x, Δ)
   η = o.eta
-  acc = get!(o.acc, x, fill(ϵ, size(x)))::typeof(data(x))
+  acc = get!(o.acc, x, fill(ϵ, size(x)))
   @. acc += Δ^2
   @. Δ *= η / (√acc + ϵ)
 end


MikeInnes:
It seems like a good idea to support custom matrix types, but generally they'll have `zero(x)::typeof(x)`. But I guess it can't hurt to have this (except perhaps for performance / inferability, but we're reworking this stuff anyway).

wshand:
Well, the custom matrix type I'd like to use comes from https://arxiv.org/abs/1903.03129, where I'm using locality-sensitive hashing to drop some of the outputs of the matrix multiply. Setting up my matrix type requires sampling some functions from an LSH family, generating hash tables, etc. which is fairly expensive in comparison to just defining `zero(::MyMatrixType)` to return `Matrix{dtype}`.

Re: performance, I'll write a benchmarking script to test speed with the current and revised versions of `apply!` and try them out later when I can access our research computing.

wshand:
Had a lot of trouble getting this to run in parallel on a CPU (never mind a GPU), so I don't know how robust/useful this is. I adapted the following code from https://github.com/FluxML/model-zoo/blob/master/vision/mnist/conv.jl (basic CNN for MNIST) and tested it on an Intel Xeon E5-2680 v3 with `JULIA_NUM_THREADS=12`:

```julia
using Flux, Flux.Data.MNIST, Random
using Flux: onehotbatch, crossentropy
using BenchmarkTools: @benchmark
using Base.Iterators: partition
using Zygote: gradient

Random.seed!(0)

# Load labels and images from Flux.Data.MNIST
train_labels, train_imgs = MNIST.labels(), MNIST.images()

# Bundle images together with labels and group into minibatches
function make_minibatch(X, Y, idxs)
    X_batch = Array{Float32}(undef, size(X[1])..., 1, length(idxs))
    for i in 1:length(idxs)
        X_batch[:, :, :, i] = Float32.(X[idxs[i]])
    end
    Y_batch = onehotbatch(Y[idxs], 0:9)
    return (X_batch, Y_batch)
end

mb_idxs = partition(1:length(train_imgs), 512)  # batch size of 512
train_set = [make_minibatch(train_imgs, train_labels, i) for i in mb_idxs]

model = Chain(
    Conv((3, 3), 1=>16, pad=(1,1), relu),
    MaxPool((2,2)),
    Conv((3, 3), 16=>32, pad=(1,1), relu),
    MaxPool((2,2)),
    Conv((3, 3), 32=>32, pad=(1,1), relu),
    MaxPool((2,2)),
    x -> reshape(x, :, size(x, 4)),
    Dense(288, 10),
    softmax,
)

# Momentum omitted because issue #816 is already resolved for the Momentum
# optimiser on the zygote branch
for opt in (Nesterov(1e-3), RMSProp(), ADAGrad())
	ps = params(model)

	println("Speed of Flux.Optimise.update! for $(typeof(opt))")
	@benchmark(
		Flux.Optimise.update!($opt, $ps, ∇),
		setup = begin 
			x, y = rand(train_set)
			∇ = gradient(() -> loss(x,y), $ps)
		end
	) |> display
	print("\n\n")
end
```

(This code actually triggers the same issue in #816 for `ADAGrad` despite not using any specialized types, since the global constant `ϵ` is `Float64` in https://github.com/FluxML/Flux.jl/blob/zygote/src/optimise/optimisers.jl#L5.)

There's definitely a stark difference in performance pre- and post-commit (though it's still a fraction of the time to compute `∇`):

```julia
#=
Pre-commit (note: most of the results from @benchmark have been cut out to
make this snippet shorter)
=#
Speed of Flux.Optimise.update! for Nesterov
BenchmarkTools.Trial: 
  memory estimate:  267.61 KiB
  median time:      350.237 μs (0.00% GC)

Speed of Flux.Optimise.update! for RMSProp
BenchmarkTools.Trial: 
  memory estimate:  134.00 KiB
  median time:      290.110 μs (0.00% GC)

Speed of Flux.Optimise.update! for ADAGrad
BenchmarkTools.Trial: 
  memory estimate:  134.00 KiB
  median time:      242.163 μs (0.00% GC)

#=
Post-commit
=#
Speed of Flux.Optimise.update! for Nesterov
BenchmarkTools.Trial: 
  memory estimate:  270.86 KiB
  median time:      407.918 μs (0.00% GC)

Speed of Flux.Optimise.update! for RMSProp
BenchmarkTools.Trial: 
  memory estimate:  137.25 KiB
  median time:      365.980 μs (0.00% GC)

Speed of Flux.Optimise.update! for ADAGrad
BenchmarkTools.Trial: 
  memory estimate:  136.38 KiB
  median time:      318.854 μs (0.00% GC)
```

I'd be interested to see how that performance penalty is changed by the sections of code you mentioned that are being reworked.

PR #835, opened 2019-08-06
Fix  cuzeros deprecation

Moelf:


dhairyagandhi96:
Thanks for the patch, @Moelf!

bors r+

PR #840, opened 2019-08-15
Fix backprop through BatchNorm in testmode

baggepinnen:
This bug only appeared while taking gradients of networks is in testmode (`training=false`)


baggepinnen:
Any reason this is held up?

MikeInnes:
What's the reason backprop currently fails? Should we make sure `squeeze` is differentiable rather than just replacing it here? Would be great to have a test too.

baggepinnen:
Well
- the `dims`-keyword is not supplied to `sum`
- `squeeze` is deprecated in favor of `dropdims`

MikeInnes:
Ah, I thought Flux had defined its own version of `squeeze` at some point, but I must be thinking of something else. Just needs the conflict fixed and the test then, thanks!

CarloLucibello:
closing as outdated

PR #844, opened 2019-08-22
better printing for optimizers

Roger-luo:
some optimizer will print a super long `IdDict` by default, removed printing that in this PR.

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -5,7 +5,7 @@ using MacroTools: @forward
 const ϵ = 1e-8
 
 # TODO: should use weak refs
-print_opt(io, name::String, args...) = print(io, name, "(", join(args, ", "), ")")
+print_opt(io, name::Union{String, Symbol}, args...) = print(io, name, "(", join(args, ", "), ")")
 print_opt(io, x::T, args...) where T = print_opt(io, nameof(T), args...)
 
 """
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -5,6 +5,8 @@ using MacroTools: @forward
 const ϵ = 1e-8
 
 # TODO: should use weak refs
+print_opt(io, name::String, args...) = print(io, name, "(", join(args, ", "), ")")
+print_opt(io, x::T, args...) where T = print_opt(io, nameof(T), args...)
 
 """
     Descent(η)
@@ -42,6 +44,8 @@ function apply!(o::Momentum, x, Δ)
   @. Δ = -v
 end
 
+Base.show(io::IO, x::Momentum) = print_opt(io, x, x.eta, x.rho)
+
 """
     Nesterov(eta, ρ = 0.9)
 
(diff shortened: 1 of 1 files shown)


dhairyagandhi96:
Bump!

src/optimise/optimisers.jl
```diff
@@ -5,6 +5,8 @@ using MacroTools: @forward
 const ϵ = 1e-8
 
 # TODO: should use weak refs
+print_opt(io, name::Union{String, Symbol}, args...) = print(io, name, "(", join(args, ", "), ")")
```
MikeInnes: I think this should be something like `ADAM(a, b, ...)`. The `...` just to indicate that there's hidden state beyond what's printed.

Roger-luo: I print in this way to make it parsable. I find it is pretty useful when someone want to serialize it.

So I basically just remove all `IdDict` and let the human input-able parameters printed.

src/optimise/optimisers.jl
```diff
@@ -174,6 +189,9 @@ function apply!(o::ADADelta, x, Δ)
   return Δ
 end
 
+Base.show(io::IO, x::ADADelta) = print_opt(io, x, x.rho)
+
+
```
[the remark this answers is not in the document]
Roger-luo: no problem!

src/optimise/optimisers.jl
Roger-luo: It is not my favorite solution tho... I have been wondering if there is way to separate the optimizer definition with the intermediate states (the `IdDict`s).

Roger-luo: Is there any use cases that one need to input the state to their optimizer?

dhairyagandhi96: The idea is to actually get rid of the state from the optimiser.

The use case could be if someone wants to resume training from exactly some point.

Roger-luo: > The use case could be if someone wants to resume training from exactly some point.

Yeah, however, IIUC, this happens when someone trained part of the model already. but currently serialization of `IdDict` is not supported in a bunch of serialization format (e.g JLD, since it requires the id of variables), one cannot recover the state at all. That's only of the reason I want the printing to be parsable: so instead of doing serialization directed on the object, I can just serialize the string...

MikeInnes: Serialising and storing via the `show` method seems like an odd use case to me. I agree that optimisers should be serialisable, but I'm not sure that's the way to do it.

If nothing else, the concern here is that you're not actually serialising the optimiser, but only a description of it (since it also contains a bunch of other hidden information). You know that already, and I know that already, but I'm concerned that it might lead to confusion for people who don't know it, which is where a visual cue would help. I'd like to support your use case better, but we should do that via the redesign, and the current show serialisation doesn't strictly need library support.

Roger-luo: Is there any proposal for the re-design of optimizers recently?

MikeInnes: There's some thinking in #637. Right now we're obviously pretty focused on the Zygote upgrade, but @dhairyagandhi96 has some more plans to tackle this after that.

ToucheSir:
I _think_ this should be unnecessary now that we're moving to `IdDict`-less optimizers from Optimisers.jl.

Roger-luo:
yeah I'll close this.

PR #847, opened 2019-08-24
Fix CuArrays.libcudnn imports

### LINKED ISSUE #846: CuArrays.CUDNN.libcudnn not defined
Hey, my Flux suddenly started emitting an error when `using CuArrays` (using Julia 1.2.0, Flux 0.8.3, CuArrays 1.2.1).

The error occurs due to an import of `CuArrays.CUDNN.libcudnn` in [curnn.jl](https://github.com/FluxML/Flux.jl/blob/aab3c4e05214b81a9cbbfa33d8da2a8d4cdba2f5/src/cuda/curnn.jl#L1) (and [cudnn.jl](https://github.com/FluxML/Flux.jl/blob/aab3c4e05214b81a9cbbfa33d8da2a8d4cdba2f5/src/cuda/cudnn.jl#L1)).
CuArrays seems to have removed `libcudnn` from the `CUDNN` module.
Instead, it is in `CuArrays.libcudnn`, so the corresponding line should be added instead.

janEbert:
Closes #846.

### DIFF
--- a/src/cuda/cudnn.jl
+++ b/src/cuda/cudnn.jl
@@ -1,5 +1,6 @@
-using .CuArrays.CUDNN: @check, libcudnn, cudnnStatus_t, cudnnTensorDescriptor_t,
+using .CuArrays.CUDNN: @check, cudnnStatus_t, cudnnTensorDescriptor_t,
   cudnnBatchNormMode_t, cudnnHandle_t, cudnnDataType, TensorDesc, FilterDesc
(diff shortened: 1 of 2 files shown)


dhairyagandhi96:
bors try

dhairyagandhi96:
bors r+

Thanks for the patch @janEbert

janEbert:
Thank you as well!

Issue #850, opened 2019-08-26
Problems using binarycrossentropy()


wisehe:
Hello, I'm working on a binary classification problem and I'm having trouble with using binarycrossentropy as my loss function. Here is an example of what I am trying to do. Forgive me if I made a mistake early on and that is why it's not working.

```
using Flux

nb_in_nodes = 20
nb_hidden_nodes = 8
nb_out_nodes = 1

feature = rand(nb_in_nodes, 100)
response = rand([0,1], 100)
dataset = [(feature, response)]

model = Chain(Dense(nb_in_nodes, nb_hidden_nodes, relu),
              Dense(nb_hidden_nodes, nb_out_nodes),
              softmax)

loss(x, y) = Flux.binarycrossentropy(model(x), y)
opt = ADAM()

Flux.train!(loss, params(model), dataset, opt)
```

I get the following error:
```
MethodError: no method matching eps(::TrackedArray{…,Array{Float32,2}})
...
Stacktrace:
 [1] binarycrossentropy(::TrackedArray{…,Array{Float32,2}}, ::Array{Int64,1}) at /home/henry/.julia/packages/Flux/qXNjB/src/layers/stateless.jl:26
```

So it looks like julia's eps function doesn't like working with the TrackedArray's that Flux.Train! eventually sends along. So I then tried the same thing but replaced the definition of the loss function on line 15 with the following, just defining epsilon directly so eps() was never called.

`loss(x, y) = Flux.binarycrossentropy(model(x), y; ϵ=1e-15)`

Then if run the above, I get the following:

> MethodError: no method matching +(::TrackedArray{…,Array{Float32,2}}, ::Float64)
> ...
> Stacktrace:
>  [1] binarycrossentropy(::TrackedArray{…,Array{Float32,2}}, ::Array{Int64,1}) at /home/henry/.julia/packages/Flux/qXNjB/src/layers/stateless.jl:26

So it looks like julia doesn't want to add an Array to a float (epsilon), like so.

> julia> [1, 2, 3] + 1.0
> ERROR: MethodError: no method matching +(::Array{Int64,1}, ::Float64)

Those are the problems I'm having.

Is binarycrossentropy() supposed to have broadcasting in it for the arithmetic and log taking, or perhaps is it supposed to only be getting a single value for ŷ and y at a time? Here is the code for binarycrossentropy as found here on github and on my own machine
[...]

PR #851, opened 2019-08-27
Add binary crossentropy fallback functions

janEbert:
These fallbacks assume that if the first argument is a 1-element
`Vector` (as given by a model's output) and the second argument is a
scalar, the first argument should also be interpreted as a scalar.
Ref #850.

The use case is putting a model's scalar output right into the loss function instead of having to manually index into the 1-element `Vector` which is easily forgotten.

Not sure if this is desired, otherwise just close this. It does have the benefit of preventing not so obvious errors that – in my opinion – should not happen.

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -25,6 +25,11 @@ Return `-y*log(ŷ + ϵ) - (1-y)*log(1-ŷ + ϵ)`. The ϵ term provides numerica
 """
 binarycrossentropy(ŷ, y; ϵ=eps(ŷ)) = -y*log(ŷ + ϵ) - (1 - y)*log(1 - ŷ + ϵ)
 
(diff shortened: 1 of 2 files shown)


MikeInnes:
Yeah, I'm not sure about this as is. Helpful coercions like this are tempting in each case, but adding them makes it gradually harder to tell what a function is actually meant to do. I think it might be preferable to document clearly that `binarycrossentropy` is meant to be used in a broadcasted way.

One other option is to make `binarycrossentropy` automatically broadcast if not given two scalars. This would actually address the same problem, but with a slightly more standard motivation (this lets us work with batches of outputs, which things like `crossentropy` already support via broadcasting.

janEbert:
Yeah, too much magic is always bad. I can change it to your suggestion if you want that instead.

CarloLucibello:
Not relevant anymore

Issue #848, opened 2019-08-24
MethodError when taking gradient of Flux.mse on GPU


janEbert:
Hey, I am getting the following errors using Julia 1.2.0, Flux [master](https://github.com/FluxML/Flux.jl/tree/6494f73c7878fffac6ac9dab667f94a0fda3c99a), CuArrays 1.2.1, Zygote [master](https://github.com/FluxML/Zygote.jl/tree/d74f3cf5ed3c185969ff7787ee36d15c105e6653). Everything works just fine if I am not `using CuArrays`.
Setup:
```julia
julia> using Flux, CuArrays, Zygote

julia> m = LSTM(10, 10) |> gpu;

julia> x = rand(10) |> gpu;

julia> loss(y_hat, y) = Flux.mse(y_hat, y);

julia> ps = params(m);

julia> y_hat = m(x);
```
Using the Tracker backend:
```julia
julia> gs = Flux.Tracker.gradient(() -> loss(y_hat, x), ps)
┌ Warning: Performing scalar operations on GPU arrays: This is very slow, consider disallowing these operations with `allowscalar(false)`
└ @ GPUArrays ~/.julia/packages/GPUArrays/J4c3Q/src/indexing.jl:16
ERROR: MethodError: no method matching culiteral_pow(::typeof(^), ::ForwardDiff.Dual{Nothing,Float32,1}, ::Val{2})
Closest candidates are:
  culiteral_pow(::typeof(^), ::Union{Float32, Float64}, ::Val{2}) at /home/jan/.julia/packages/CuArrays/wXQp8/src/broadcast.jl:46
  culiteral_pow(::typeof(^), ::Union{Float32, Float64}, ::Val{p}) where p at /home/jan/.julia/packages/CuArrays/wXQp8/src/broadcast.jl:48
  culiteral_pow(::typeof(^), ::Union{Float32, Float64}, ::Val{0}) at /home/jan/.julia/packages/CuArrays/wXQp8/src/broadcast.jl:44
  ...
Stacktrace:
 [1] #2 at ./broadcast.jl:299 [inlined]
[...]

PR #852, opened 2019-08-27
RFC: Replace Requires with direct CuArrays dependency.

maleadt:
Instead of depending indirectly on CuArrays through Requires, this PR adds a proper dependency on the package, and uses CUDAapi to figure out if we can actually load CuArrays. Advantages: actual version compatibility, which should improve the usability of the package. Disadvantages: some additional packages are installed on every system, even without a GPU. Proper conditional dependencies would be a better solution, but for now this seems like an improvement.

Requires a tag of CUDAnative: https://github.com/JuliaRegistries/General/pull/2978
Ref https://discourse.julialang.org/t/cuarrays-cudanative-psa-simplified-package-loading/27897


MikeInnes:
Awesome, thanks!

janEbert:
Hey, would it be possible to also add a function which can disable the loading of CuArrays?
With #848 and other possible errors, it would be nice to selectively turn off GPU usage which would otherwise require modifying `JULIA_LOAD_PATH` or writing a wrapper around `Flux.gpu`.

MikeInnes:
Couldn't you just not call `Flux.gpu`?

janEbert:
But then if I wanted to use the GPU I'd have to change every usage.

maleadt:
I'd even argue that calling `gpu` when the GPU stack is not functional should error, or at least warn.

janEbert:
I was gonna say that that is what I like about the function – that you can just call it always because it does the right thing but I guess then Flux could just always call it internally. I see your points. :)

MikeInnes:
You can just define something like
```julia
enable_gpu = true
foo(x) = enable_gpu ? gpu(x) : x

m = Chain(...) |> foo
```

xukai92:
When is the check of CuArrys done? I have been having issue on `Flux.gpu` since this PR is merged. While my machine does have CUDA and CuArrays, the `Flux.gpu` sometimes just doesn't work, usually after I do `] update` and everything got compiled or built again.

My `Flux.gpu` is not working ATM. Any information I can provide to help diagnosis this?

maleadt:
There is one corner case I have not found a workaround, and you might be running into this: if at some point your GPU environment was broken, and CuArrays failed to import (or CUDAapi decided you didn't have a GPU/CUDA), Flux will be precompiled "without" GPU support, since those checks happen in global scope when the package is precompiled.

To fix this, we'd need to consult CUDAapi from init, but also be able to somehow declare a precompile dependency on CuArrays even if it failed to import (in order to trigger recompilation of Flux when CuArrays gets updated). I haven't found a way to do that.

Workaround; delete `~/.julia/compiled`.

xukai92:
Can I ask Julia to re-precompile Flux somehow to achive the same thing?

xukai92:
I mean by a command or function.

maleadt:
```
julia> Base.compilecache(Base.PkgId("Flux"))
```

xukai92:
Cool. Thanks!

xukai92:
> There is one corner case I have not found a workaround, and you might be running into this: if at some point your GPU environment was broken, and CuArrays failed to import (or CUDAapi decided you didn't have a GPU/CUDA), Flux will be precompiled "without" GPU support, since those checks happen in global scope when the package is precompiled.

I think my case is exactly this. In my uni, all the server uses the same afs to store files. I have used Flux on a sever with only CPUs and then started to use one with GPUs. And it doesn't seem to be re-precompiled in this case as the filesytem is shared.

maleadt:
I guess we could do the CUDAapi check during `__init__`, and that would cover most (including your) cases. Let me have a look.

xukai92:
Also how should it work with environment? I think logically Flux should not load GPU if there is no CuArrays in the activated environment, even the machine has GPUs, right?

maleadt:
Flux now depends on CuArrays, so it's always in the environment.

xukai92:
Oh yes you are right.

PR #617, opened 2019-02-11
Support for additional dimensions in Dense layer via reshape

pshashk:
Another attempt at https://github.com/FluxML/Flux.jl/issues/282 (https://github.com/FluxML/Flux.jl/pull/293)

PR #856, opened 2019-09-02
Dense with optional reshape

### LINKED ISSUE #282: Support for additional dimensions in Dense layer

mcabbott:
Fixes #282, by allowing `Dense((5, 3), (2, 7))` as suggested. Also closes #293, #617.

For `d = Dense(15, 14)` I wanted to leave the original behaviour alone, but attempting to do this by dispatch landed me in ambiguity hell, so it's an `if` statement (which I think ought to get compiled out). Possibly this should have `&& ndims(x)<=2` so that such `d` does allow higher-dimensional data. 

I also added a line to allow `Dense(15 => 14)` and `Dense((5, 3) => (2, 7))`, more like `Conv` and less confusing I think. But will remove if anyone objects / thinks this deserves its own issue. 

Note by the way the following `::Any` problem. Since this persists after `] free Flux`, I don't think it's my fault. (This is on Julia 1.1)
```
julia> d = Dense(5,2)
Dense(5 => 2)

julia> @code_warntype d(rand(Float32, 5))
Body::Any
```

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -60,7 +60,7 @@ end
 
 
 """
-    Dense(in::Integer, out::Integer, σ = identity)
+    Dense(in => out, σ = identity)
 
 Creates a traditional `Dense` layer with parameters `W` and `b`.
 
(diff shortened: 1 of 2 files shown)


dhairyagandhi96:
This is interesting! We should also add a depwarn to the changed API.

mcabbott:
Glad someone likes `=>`! But what I propose just to leave `Dense(5,2)` working, so that nothing breaks. If that's not too confusing?

I can add a news line, but should it go under v0.10.0?

src/layers/basic.jl
```diff
@@ -57,9 +57,10 @@ function activations(c::Chain, input)
   return rst
 end
 
+using Base: depwarn
 
 """
-    Dense(in::Integer, out::Integer, σ = identity)
+    Dense(in => out, σ = identity)
```
CarloLucibello: Shouldn't we have in this docstring, or as separete docstrings, also the other constructors?
```julia
Dense(in, out)
Dense(W, b)
```
```

mcabbott: See if you like this now. I’ve also tried to make clearer that `σ` is not a keyword.

mancellin:
What is the status of this?
I've been drafting a small PR with some improvements of Dense docstring, but most of it would actually be fixed (or outdated) by this PR from @mcabbott.

mcabbott:
Would still be nice, IMO, but didn't gain traction? I guess an alternative would be to make a nice `Reshape()` layer.

Instead of `bias=false`, there's also a proposal in #873 to write `bias=Zeros()`.

CarloLucibello:
+1 for the Reshape layer, to me it seems more manageable and of broader utility

bhvieira:
+1 for the Reshape layer as well

drozzy:
Please merge this or the other solution (https://github.com/FluxML/Flux.jl/pull/293)  :-) It seems having this features is super useful - and I would take any solution over no solution. Additionally the existing functionality will not suffer since it is behind an if statement.
If there is an efficiency issue, I'm sure a patch could be applied in the future. (Just my 2c)
I spent about a day trying to see if Flux supports this and I'm sure many more people will (esp. pytorch/keras folks).

Issue #201, opened 2018-03-08
error for crossentropy on GPU: ERROR: Broadcast output type Any is not concrete


osofr:
Running the LSTM example on GPUs with `crossentropy`  loss produces the following error. Different losses (e.g., `mse`) work fine. Snippet below, for full replication run example in #189.

```julia
julia> function losscu(xs, ys)
         l = sum(crossentropy.(predict.(xs), gpu.(ys)))
         Flux.truncate!(mcu)
         return l
       end
losscu (generic function with 1 method)

julia> predict(x) = collect(x) |> gpu |> mcu
predict (generic function with 1 method)

julia> @time Flux.train!(losscu, zip(Xs[1:1], Ys[1:1]), opt, cb = throttle(evalcb, 30))
ERROR: Broadcast output type Any is not concrete
Stacktrace:
 [1] broadcast_t at /home/ubuntu/.julia/v0.6/CuArrays/src/broadcast.jl:34 [inlined]
 [2] broadcast_c at /home/ubuntu/.julia/v0.6/CuArrays/src/broadcast.jl:63 [inlined]
 [3] broadcast at ./broadcast.jl:455 [inlined]
 [4] tracked_broadcast(::Function, ::Flux.OneHotMatrix{CuArray{Flux.OneHotVector,1}}, ::TrackedArray{…,CuArray{Float32,2}}, ::Int64) at /home/ubuntu/.julia/v0.6/Flux/src/tracker/array.jl:278
 [5] #crossentropy#71(::Int64, ::Function, ::TrackedArray{…,CuArray{Float32,2}}, ::Flux.OneHotMatrix{CuArray{Flux.OneHotVector,1}}) at /home/ubuntu/.julia/v0.6/Flux/src/layers/stateless.jl:8
 [6] crossentropy(::TrackedArray{…,CuArray{Float32,2}}, ::Flux.OneHotMatrix{CuArray{Flux.OneHotVector,1}}) at /home/ubuntu/.julia/v0.6/Flux/src/layers/stateless.jl:8
 [7] (::##61#62)(::Flux.OneHotMatrix{Array{Flux.OneHotVector,1}}, ::Flux.OneHotMatrix{Array{Flux.OneHotVector,1}}) at ./<missing>:0
[...]

PR #861, opened 2019-09-11
GPU CI maintainance 

dhairyagandhi96:


dhairyagandhi96:
CI failure seems unrelated

maleadt:
I'd drop the branch restrictions, that way you get automatic testing for people who push to branches on Flux.jl (and only need Bors for external contributors).

dhairyagandhi96:
Thanks @maleadt!

Does this look better?

maleadt:
Yes. Any particular reason you define your own `flux` target and don't use the templates (e.g. https://raw.githubusercontent.com/JuliaGPU/gitlab-ci/master/templates/v4/test_v1.2.yml)?

dhairyagandhi96:
We used to need it since we had to add CuArrays as a dependency for testing on the GPUs. But since we don't need to do that anymore, we can safely get rid of it.

MikeInnes:
Is this ready to go?

dhairyagandhi96:
I think the code is there, but I am seeing `access to undefined reference` on Julia v1.1 and v1.2

maleadt:
@MikeInnes reported something similar yesterday, I'll have a closer look.

dhairyagandhi96:
Am I doing something incorrect here with regard to the error from the bors job above?

maleadt:
No, bors just reported that the build failed.

maleadt:
Interesting, this is a legitimate issue now that we depend on CuArrays from Flux and as a result the CuArrays `__init__` functions might not have run before Flux' global code (which performs an allocation in global scope). I'll push a quick fix on a stable branch of CuArrays, but I've fixed the issue in the PR that moves over Flux DNN functionality (including that global allocation) to CuArrays. https://github.com/JuliaGPU/CuArrays.jl/pull/404#issuecomment-532732574 was related after all :smile: 

@dhairyagandhi96 could you try the CuArrays `stable` branch? https://github.com/JuliaGPU/CuArrays.jl/tree/stable

dhairyagandhi96:
bors try

Thanks!

dhairyagandhi96:
`ERROR: LoadError: LoadError: LoadError: CUDA error: invalid device context (code #201, ERROR_INVALID_CONTEXT)`

Seems related to the initialisation

maleadt:
Maybe it's easier to include the fix here instead: https://github.com/JuliaGPU/CuArrays.jl/pull/404/commits/5048e892b8556b4ed79b697c3f592e6767d3981f

dhairyagandhi96:
bors try

dhairyagandhi96:
Apologies for the delay, I saw the message on here just then. I think I have to fix the toml files here.

dhairyagandhi96:
Thanks for the help @maleadt!!

bors try

dhairyagandhi96:
bors try

dhairyagandhi96:
bors try

dhairyagandhi96:
bors try

dhairyagandhi96:
bors try

maleadt:
Why do you need to extend manually? The failures before didn't seem CI infrastructure-related.

dhairyagandhi96:
In the default the template does a `dev` of the project per 
https://github.com/JuliaGPU/gitlab-ci/blob/52d59dadbc16b7b5cbb87f6a276cb50f68ecf554/templates/v4/common.yml#L25

With having dependencies on the `master` in a couple places, it is possible that the exact commit is not picked up (or rather the tag is), hence returning the `pullback not defined`

Trying this out to see if this if this way allows for better reproducibility.

maleadt:
> With having dependencies on the `master` in a couple places, it is possible that the exact commit is not picked up (or rather the tag is), hence returning the `pullback not defined`

There shouldn't be a difference between `develop $PATH` and `--project $PATH instantiate`, right? Ah wait, I guess the former ignores any Manifest. But isn't that risky for CI, where regular package installation ignores the Manifest anyhow?

MikeInnes:
The benefits are that (a) the CI builds are completely reproducible and (b) it makes it substantially easier to do WIP on multiple packages at once (e.g. we can update Flux for Zygote changes and make patches as needed, rather than needing to tag a Zygote release, then doing several patch releases for issues Flux finds).

Of course at release time we need to make sure we're using release versions; the manifest then serves as a clear log of what needs tags and/or compat updates, without us having to keep track of that throughout the release cycle.

dhairyagandhi96:
Right. It does mean that any breaking changes in the dependencies can break CI. This comes against the reproducibility of the environment (since the Manifest is ignored)

maleadt:
That sounds like a good fit for CUDAnative/CuArrays/GPUArrays too. @dhairyagandhi96 the `v3` templates used `instantiate` instead of `develop`, I can update those so that we can continue using them.

dhairyagandhi96:
Okay, sounds good! We can then just use the corresponding targets without rewriting them here.

PR #870, opened 2019-09-25
Fix printing of SkipConnection

mcabbott:
Before:
```
julia> SkipConnection(Dense(2,2),+)
SkipConnection(Error showing value of type SkipConnection:
ERROR: MethodError: no method matching iterate(::Dense{typeof(identity),TrackedArray{…,Array{Float32,2}},TrackedArray{…,Array{Float32,1}}})

julia> SkipConnection(Chain(Dense(2,3), Dense(3,2), LayerNorm(2)),+)
SkipConnection(Dense(2, 3), Dense(3, 2), LayerNorm(2))

julia> SkipConnection(Dense(2, 3), Dense(3, 2), LayerNorm(2))
ERROR: MethodError: no method matching SkipConnection(::Dense{typeof(identity),TrackedArray{…,Array{Float32,2}},TrackedArray{…,Array{Float32,1}}}, ::Dense{typeof(identity),TrackedArray{…,Array{Float32,2}},TrackedArray{…,Array{Float32,1}}}, ::LayerNorm{TrackedArray{…,Array{Float32,1}}})
```
After:
```
julia> SkipConnection(Dense(2,2),+)
SkipConnection(Dense(2, 2), +)

julia> SkipConnection(Chain(Dense(2,3), Dense(3,2), LayerNorm(2)),+)
SkipConnection(Chain(Dense(2, 3), Dense(3, 2), LayerNorm(2)), +)

julia> SkipConnection(Dense(2,2), (a,b) -> a .+ b./2)
SkipConnection(Dense(2, 2), #9)
```

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -208,7 +208,6 @@ sm = SkipConnection(m, (mx, x) -> cat(mx, x, dims=3))
 size(sm(x)) == (5, 5, 11, 10)
 ```
 """
-function SkipConnection end
 struct SkipConnection
   layers
   connection  #user can pass arbitrary connections here, such as (a,b) -> a + b
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -217,7 +217,7 @@ end
 
 function Base.show(io::IO, b::SkipConnection)
   print(io, "SkipConnection(")
-  b.layers isa Chain ? join(io, b.layers, ", ") : print(io, b.layers)
-  print(io, ",", b.connection)
-  print(io, ")")
+  b.layers isa Chain ? print(io, "Chain(", join(b.layers, ", "), "), ") :
+    print(io, b.layers, ", ")
+  print(io, b.connection, ")")
 end
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -192,17 +192,23 @@ end
 """
     SkipConnection(layers, connection)
 
-Creates a Skip Connection, which constitutes of a layer or `Chain` of consecutive layers
-and a shortcut connection linking the input to the block to the
-output through a user-supplied callable.
+Creates a Skip Connection, of a layer or `Chain` of consecutive layers
+plus a shortcut connection. The connection function will combine the result of the layers
+with the original input, to give the final output.
 
-`SkipConnection` requires the output dimension to be the same as the input.
+The simplest 'ResNet'-type connection is just `SkipConnection(layer, +)`,
+and requires the output of the layers to be the same shape as the input.
+Here is a more complicated example:
+```
+m = Conv((3,3), 4=>7, pad=(1,1))
+x = ones(5,5,4,10);
+size(m(x)) == (5, 5, 7, 10)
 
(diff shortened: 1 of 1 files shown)


dhairyagandhi96:
Lgtm!

bhvieira:
Whoops, I knew the print method was poor, but didn't notice it failed!

src/layers/basic.jl
```diff
@@ -190,17 +190,17 @@ function (mo::Maxout)(input::AbstractArray)
...
+Creates a Skip Connection, which constitutes of a layer or `Chain` of consecutive layers
 and a shortcut connection linking the input to the block to the
 output through a user-supplied callable.
 
 `SkipConnection` requires the output dimension to be the same as the input.
 
 A 'ResNet'-type skip-connection with identity shortcut would simply be
 ```julia
-    SkipConnection(layer, (a,b) -> a + b)
+    SkipConnection(layer, +)
```
bhvieira: I think I had this example spelling out the `(a,b)` tuple because I wanted to make obvious the usage on non-"standard" callables. Perhaps it's more obvious to other users than I thought though.

mcabbott: Is there a simple example of a function which isn’t `+` which people use? I was a bit mystified as to why it was written out like this.

bhvieira: `cat` is big (think DenseNet). Also, it could involve another layer (1x1 convolutions are common for this).

bhvieira: The thing with `cat` is that you have to spell out `dims`, so the plain call wouldn't work.

mcabbott: See if you like my `cat` example. Seems a little crazy how long this is now for a one-line function!

MikeInnes:
Looks good, thanks a lot -- @mcabbott is this ready to go?

src/layers/basic.jl
bhvieira: I think it's good (it was good already before the last commit as well).

mcabbott:
Yes I’m done!

src/layers/basic.jl
```diff
@@ -192,17 +192,23 @@ end
...
+size(m(x)) == (5, 5, 7, 10)
 
-A 'ResNet'-type skip-connection with identity shortcut would simply be
-```julia
-    SkipConnection(layer, +)
+sm = SkipConnection(m, (mx, x) -> cat(mx, x, dims=3))
+size(sm(x)) == (5, 5, 11, 10)
 ```
 """
+function SkipConnection end
```
MikeInnes: This seems like an odd addition, what's it for? Doesn't this make redefining SkipConnection as a struct break?

MikeInnes: I got confused by the diff -- I see you've deleted this.

mcabbott: Should be gone, I somehow committed the version before deleting this by mistake.

mcabbott: Right, sorry about five thumbs. That was cause of test failure. No code that actually does anything should now be changed at all.

Issue #868, opened 2019-09-24
Convolution without bias


pshashk:
Many [vision networks ](https://github.com/search?utf8=%E2%9C%93&q=%22bias%3DFalse%22+repo%3Apytorch%2Fvision+extension%3Apy&type=Code&ref=advsearch&l=&l=) doesn't add bias after convolution. Can we add a flag to disable it in layer constructor?

MikeInnes:
About to be addressed by #873.

CarloLucibello:
reopening until #873 gets merged

dhairyagandhi96:
Closing through #873

PR #873, opened 2019-09-27
Make bias optional

dhairyagandhi96:
Addresses #868


pshashk:
Thank you for your prompt PR. For completeness sake this option could also be added to dense, recurrent and normalization layers (to disable affine transformation after normalization).

src/layers/conv.jl
```diff
@@ -30,29 +30,34 @@ struct Conv{N,M,F,A,V}
   stride::NTuple{N,Int}
   pad::NTuple{M,Int}
   dilation::NTuple{N,Int}
+  use_bias::Bool
 end
 
 function Conv(w::AbstractArray{T,N}, b::AbstractVector{T}, σ = identity;
-              stride = 1, pad = 0, dilation = 1) where {T,N}
+              stride = 1, pad = 0, dilation = 1, use_bias = true) where {T,N}
```
CarloLucibello: It is weird though that when calling this constructor with `use_bias=false` one has to pass a vector `b` as well.
I would suggest the following non-breaking change instead of the `use_bias` flag:
  - relax the signature to 
```julia
Conv(w::AbstractArray{T,N}, b::Union{Nothing,AbstractVector{T}}, σ = identity; ...)
```
and have a call to 
```julia
Conv(w, nothing)
```
construct a Conv layer with no bias

MikeInnes:
How about if we did this by setting the bias to `nothing`?

We could even use e.g. some kind of `Zero` type that does the right thing when you broadcast over it; then we can automatically use that in a bunch of layers, rather than modifying each one.

pshashk:
Passing `nothing` as weight initializer (stuff like initW, initb, initβ, initγ) seems like the most flexible solution to me.

dhairyagandhi96:
Yeah, holding the bias felt awkward to say the least. I had wanted to keep the signature more or less the same to avoid breaking code that depended on that behavior. I had also previously tried just setting the bias to a `zero(eltype(T))`, but the extra compute I felt could be avoided.

I'll modify this to have a nothing instead. That looks clean

mcabbott:
I was going to suggest setting the bias to `false`, as this ought to drop out of `σ.(conv(x, c.weight, cdims) .+ b)` when applying the layer, without needing further checking. And that constructing this `Conv(…, bias=false)` would then be both nice and accurate.

dhairyagandhi96:
We could introduce something like
```
struct ZeroType <: Number end
+(a::Number, ::ZeroType) = a
reshape(::ZeroType, args...) = ZeroType()
```

And use this whenever we need it for similar use cases. The constructor can just accept this type or a `Nothing` and we can get the bias (or any switch) that way

dhairyagandhi96:
With the `false` approach, how do we handle the call to `reshape` in the forward pass?

mcabbott:
Good point about reshape, and `[false]`is ugly, so I change my vote to branching on `b===nothing`.

dhairyagandhi96:
Now we should be able to construct `Conv(w, nothing)`, `Conv(w, ZeroType(...))` and `Conv((2,2), 1=>3, use_bias = false)`

MikeInnes:
The `Conv(weight, bias)` should be fine as is; how about the following setup for the convenience constructor:

```julia
Conv((k, k), in=>out; weight = convweight((k, k), in=>out), bias = convbias(out)) = ...
```

Then we can do `Conv((2, 2), 3=>6, bias = zero)`. And as a bonus it's easy to generate the weight matrix outside of the conv layer, which is a nice utility to have.

dhairyagandhi96:
For people who want to provide their own weight matrices, `Conv(weight, bias)` is the more intuitive constructor, and I like having `convweight` to help that constructor more than the convenience one, as it makes the API a bit symmetric. `Conv(convweight((k,k), in=>out), convbias(out))`

A `weight` kwarg might make it harder for folks to find `Conv(weight, bias)`, hence will document it explicitly.

I also worry that if provided, it effectively makes the first few args to the convenience constructor redundant.

janEbert:
This ought to be documented in the "convenience" constructor as well! Great change, too; I always wrote silly extra constructors until now.

src/layers/conv.jl
```diff
@@ -32,18 +32,35 @@ struct Conv{N,M,F,A,V}
   dilation::NTuple{N,Int}
 end
 
-function Conv(w::AbstractArray{T,N}, b::AbstractVector{T}, σ = identity;
+"""
+    Conv(weight::AbstractArray, bias::AbstractArray)
+    Conv(weight::AbstractArray, bias::AbstractArray, relu)
```
MikeInnes: Seems like it should be something like `activation` rather than `relu`. Otherwise it looks like you have to pass that specific function in.

src/layers/conv.jl
```diff
@@ -32,18 +32,35 @@ struct Conv{N,M,F,A,V}
...
   dilation = expand(Val(N-2), dilation)
   return Conv(σ, w, b, stride, pad, dilation)
 end
 
-Conv(k::NTuple{N,Integer}, ch::Pair{<:Integer,<:Integer}, σ = identity;
-     init = glorot_uniform,  stride = 1, pad = 0, dilation = 1) where N =
-  Conv(init(k..., ch...), zeros(ch[2]), σ,
+convweight(k::NTuple{N,Integer}, ch::Pair{<:Integer,<:Integer};
+  init = glorot_uniform) where N = init(k..., ch...)
+const convbias = zeros
```
MikeInnes: This doesn't seem necessary.

src/layers/conv.jl
```diff
@@ -32,18 +32,35 @@ struct Conv{N,M,F,A,V}
...
   stride = expand(Val(N-2), stride)
   pad = expand(Val(2*(N-2)), pad)
   dilation = expand(Val(N-2), dilation)
   return Conv(σ, w, b, stride, pad, dilation)
 end
 
-Conv(k::NTuple{N,Integer}, ch::Pair{<:Integer,<:Integer}, σ = identity;
-     init = glorot_uniform,  stride = 1, pad = 0, dilation = 1) where N =
-  Conv(init(k..., ch...), zeros(ch[2]), σ,
+convweight(k::NTuple{N,Integer}, ch::Pair{<:Integer,<:Integer};
```
MikeInnes: While we're naming things, maybe this should be called `filter`; that'd probably be more familiar to many people, and `weight` is a bit meaningless.

test/layers/conv.jl
```diff
@@ -20,6 +20,17 @@ end
...
   @test size(m(r)) == (10, 5)
+
+  # Test bias switch
+  bias = Conv(ones(Float32, 2, 2, 1, 3), ones(Float32, 3))
+  ip = zeros(Float32, 28,28,1,1)
+
+  op = bias(ip)
+  @test sum(op) == prod(size(op))
+
+  bias = Conv((2,2), 1=>3, bias = zero(3))
```
MikeInnes: Why use the `zero` function here? Also, are we still interested in a `zero` type?

src/layers/conv.jl
dhairyagandhi96: Right I was following the structure from earlier, but we of it's alright in this PR, I'd like to change docs wherever applicable

test/layers/conv.jl
dhairyagandhi96: Ah, did passing a `bias = zero` (from https://github.com/FluxML/Flux.jl/pull/873#issuecomment-537980961) point to the zero type?

src/layers/conv.jl
dhairyagandhi96: Gotcha

test/layers/conv.jl
MikeInnes: yeah, sorry, I was imagining a definition like `zero = Zero()`. That might not work given that `zero` exists as a function, but I think we'll need a new type to guarantee that it won't be updated by optimisers.

dhairyagandhi96: Right, in that case I'll just get the definition of `ZeroType` back and we can have the guarantee baked into that

MikeInnes:
For the zero type, how's this for an interface. `Flux.Zeros(sz::Int...)` behaves as an array of zeros of shape `sz` (the `eltype` should probably be `Bool`). `Zeros()` creates a 0D array, which should, I think, automatically do the right thing when broadcasting. Then this is as convenient as `Conv((k, k), in=>out, bias = Zeros())`, without anything being too much of a special case.

src/layers/conv.jl
dhairyagandhi96: Should be able to support `Conv(w, nothing)` and make it a bit more extensible now, I think

dhairyagandhi96:
Yeah, I was thinking of adding a convenience around that as I added the docstrings. The interface looks good also.

src/layers/conv.jl
dhairyagandhi96: Ideas about replacing `weight` with something more specific?

MikeInnes: Just find/replace to `filter` everywhere?

janEbert:
I think `Zeros` isn't the best name for a non-learnable array. A new user seeing code with `Conv(..., bias=Zeros())` would most likely assume that it's a learnable version of `zeros` instead of a non-learnable array (that is ignored during computation as well).
`ZeroType` also seems a bit awkward to me. Maybe something along the lines of `Ignored`, `Neutral`, `PassOver`, `Disregarded` or `NotLearned`? (I personally like `PassOver` best.)

dhairyagandhi96:
Thoughts on `ZeroLike`?

MikeInnes:
If we were to use a different type like that, we could just use `nothing`. But the idea is to make it a "normal" array type as far as possible, so you don't have to modify layers to handle nothing (and `PassOver` is a weird name for an array).

I see that `Zeros()` could be seen as a constructor for a normal zero array, but once you've seen that it returns a `Zero` type, I think it's intuitive that the value can't be changed to contain other values, by definition.

janEbert:
> Thoughts on `ZeroLike`?

Love it!

> If we were to use a different type like that, we could just use nothing. But the idea is to make it a "normal" array type as far as possible, so you don't have to modify layers to handle nothing (and PassOver is a weird name for an array).

I get that, I just think `Zeros` would be a very unfortunate name. (And you're right, `PassOver` _is_ weird.)

MikeInnes:
When I read `ZeroLike` I think: zero like what? It sounds closer to what Julia's current `zero(x)` function does.

The issue is, from the perspective of the forward pass, it's either zero or it's not. We don't have any kind of naming convention for things that only affect backwards pass / training semantics.

janEbert:
I see your issues with the name but the difference from `ZeroLike` to `zero` is much clearer than the difference from `Zeros` to `zeros`. What about `ZerosLike`?
In terms of naming conventions, I think it's important for users to have an easily understandable interface to the library, although I cannot come up with a good name at the moment either.

MikeInnes:
Yes, it's definitely important that it's intuitive. I think we should go with `Zero()` for now, and we can always change if we have a suggestion that better gets across "array of zeros that doesn't get updated by optimisers" in a concise way.

janEbert:
`StaticZeros`?

MikeInnes:
Unfortunately the word "static" (with "dynamic") is already way too heavily overloaded in the Julia world, and elsewhere. Using it to mean trainable/constant would definitely cause other confusion.

dhairyagandhi96:
Looking for some feedback on the current way its handled (as of https://github.com/FluxML/Flux.jl/pull/873/commits/4a183aeaf02a9de9a98f21ee5eddfd0e7f8219f4). Could also alter the forward pass to not reshape in the presence of a `Zeros` object.

MikeInnes:
So me and @dhairyagandhi96 had some discussion of whether the zero type should be a scalar or array. While a scalar is obviously simpler, an array has the benefit of not needing to be a special case in things like the `Conv` code (which needs to do a reshape of the bias), and it won't cause problems with `<:AbstractArray` type hints.

@dhairyagandhi96 has had the idea of just implementing `reshape` for the scalar zero as a no-op; I feel somewhat uneasy about that as it slightly abuses the meaning of `reshape`, but it may not cause any issues in practice. If it does, I suppose we could always have both `Zero` and `Zeros`.

dhairyagandhi96:
Note that it is possible to have it be an array type as well. The conversion should hold better in this case. I would like to hold the size information in that case too so we respect the pedagogical standards of how we interact with arrays.

I think I'll make that change here.

MikeInnes:
Ok, so you're going with the arrays approach now -- I think it's worth some notes here on the reasoning for that if so, just in case we end up circling back to this decision.

dhairyagandhi96:
Right, so the reasoning is simply that we can probably use this to express more clearly what our intentions with this are, and extend it to be used in cases that come hence. It is also easier to reason about in corner cases, I feel, since the precedent is simply what would an array with zeros do here, does it meaningfully change the underlying logic if replaced with something similar to `Zeros`.

It also naturally extends a bunch of the Array logic that we can take lean on as well.

dhairyagandhi96:
The disadvantage is probably in how it interacts with the GPU. Without external hints, it may be that we need to specialise some functions to return `CuArrays` as necessary.

src/layers/conv.jl
```diff
@@ -32,18 +36,49 @@ struct Conv{N,M,F,A,V}
...
+layer.
+
+Takes the keyword arguments `pad`, `stride` and `dilation`.
+"""
+function Conv(w::AbstractArray{T,N}, b::Union{Nothing, Zeros, AbstractVector{T}}, σ = identity;
               stride = 1, pad = 0, dilation = 1) where {T,N}
   stride = expand(Val(N-2), stride)
   pad = expand(Val(2*(N-2)), pad)
   dilation = expand(Val(N-2), dilation)
+  b = b isa Nothing ? Zeros((size(w, ndims(w)), )) : b
```
MikeInnes: This seems to defeat the idea of `Zeros` somewhat; if we make API exceptions like this, we'll end up having to change all layers to behave consistently.

src/layers/conv.jl
```diff
@@ -32,18 +36,49 @@ struct Conv{N,M,F,A,V}
...
+Constructs a standard convolutional weight matrix with given `filter` and
+channels from `in` to `out`.
+
+Accepts the keyword `init` (default: `glorot_uniform`) to control the sampling
+distribution.
+
+See also: [`depthwiseconvfilter`](@ref)
+"""
+convfilter(filter::NTuple{N,Integer}, ch::Pair{<:Integer,<:Integer};
+  init = glorot_uniform) where N = init(filter..., ch...)
```
MikeInnes: Syntax nit: `init = ` should start aligned with its scope (the opening `(`), with the function body on a new line. (This is true in a couple of places).

src/utils.jl
```diff
@@ -139,6 +139,106 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+    Zeros(Type, size...)
+
+Acts as a stand-in for an array of zeros that can be
+used during training which is ignored by the optimisers.
+
+Useful to turn bias off for a forward pass of a layer.
+
+!!! warning
+    Zeros acts a scalar while broadcasting, so does not
+    expand dims. Checks for shape compatibility by default.
```
MikeInnes: Seems like it'd be best to have it behave consistently here.

src/utils.jl
```diff
@@ -139,6 +139,106 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+3×3 Array{Float64,2}:
+ 0.198739  0.490459  0.785386
+ 0.779074  0.39986   0.66383
+ 0.854981  0.447292  0.314497
+
+julia> bias_less_conv = Conv((2,2), 1=>3, bias = Flux.Zeros())
+Conv((2, 2), 1=>3)
+```
+"""
+struct Zeros{T,N} <: AbstractArray{T,N}
```
MikeInnes: Do we need the flexibility to customise the eltype? It's not hurting, obviously, but I'd imagined that `Bool` would be sufficient. Perhaps it's useful for layer type restrictions?

src/utils.jl
```diff
@@ -139,6 +139,106 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+Base.copy(xs::Zeros{T,N}) where {T,N} = Zeros(T, size(xs)...)
+
+isZeros(x::Zeros) = true
+isZeros(x) = false
+
+function Base.copyto!(dest::AbstractArray, bc::Broadcasted{Broadcast.ArrayStyle{Flux.Zeros}})
+  bc = Broadcast.flatten(bc)
+
+  i = isZeros(first(bc.args)) ? 2 : 1 # findfirst(!isZeros, bc.args)
+  dest .= bc.args[i]
```
MikeInnes: Is this valid in general? It seems like this is applying to _all_ broadcasted operations, but in general e.g. `x .* Zeros()` or more general `f.(x, Zeros())`, we can't know that it behaves as the identity.

I think it might be simpler to only overload `broadcasted` for some specific types to make it a no-op, allowing the rest of the broadcast machinery to do its thing normally.

src/layers/conv.jl
dhairyagandhi96: Gotcha

src/utils.jl
dhairyagandhi96: Yeah, I'd added it prior to hooking it in to broadcasting, it's not valid anymore. We have it behave consistently like an array now.

src/utils.jl
dhairyagandhi96: I was imaging just for dispatch and `collect` kind of operations, so if we do need to allocate it somewhere, we are consistent with the type that the user needs in that case. The default case is `Bool`, of course.

src/utils.jl
dhairyagandhi96: Ah, right, so broadcasting over `Zeros()` would (for the overloaded cases) be a no-op, and for others would follow the usual broadcasting rules. How would we handle cases where scalar indexing could creep in?

Also, would thinking of `Zeros()` as a scalar vs. `Zeros(1,2,3..)` as an `Array` be useful in this case? Or would it lend itself more to larger complications?

src/layers/conv.jl
dhairyagandhi96: This was only if we want to use `nothing` to set the bias, if we are okay without it, removing it is fine.

src/utils.jl
MikeInnes: I think they'd probably end up with an identical set of `broadcasted` rules, so it scalar vs array makes no difference here.

If you do something like `CuArray .* Zeros`, I think it'll just work with the current broadcasting rules (might be worth checking explicitly). This should work because `Zeros` is a bits type, so can just live on the GPU and be indexed there.

dhairyagandhi96:
CI failure is curious, but seems to come from somewhere else

src/utils.jl
MikeInnes: So can this warning be removed in that case?

src/utils.jl
```diff
@@ -139,6 +139,130 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+Base.axes(xs::Zeros) = Base.OneTo.(size(xs))
+
+Base.IndexStyle(::Type{<:Zeros}) = IndexLinear()
+
+Base.getindex(xs::Zeros{T,N}, I::Int) where {T,N} = zero(T)
+Base.getindex(xs::Zeros{T,N}, inds::Union{Base.OneTo, Base.UnitRange}) where {T,N} =
+              Zeros(T, inds.stop)
+
+Base.setindex(xs::Zeros, args...) =
+              error("setindex disallowed on Zeros Array")
```
MikeInnes: `setindex` could work just by promoting to a regular array. So might be best to just leave this unimplemented, rather than indicating that it can't be supported (someone might implement it for us :)

src/utils.jl
```diff
@@ -139,6 +139,130 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+  res .= zero(S)
+end
+
+Base.copy(xs::Zeros{T,N}) where {T,N} = Zeros(T, size(xs)...)
+
+# Define broadcasting behaviour
+for op in (:+, :-)
+  @eval function broadcasted(::typeof($op), a::AbstractArray, b::Zeros)
+    sz = similar(a, Broadcast.broadcast_shape(size(a), size(b)))
+    sz .= a
```
MikeInnes: Can't we just return `a`?

MikeInnes:
Looks good. The broadcasting rules look like they're a bit more complex than they need to be though; seems like we should only need a few rules and just need to return `x` or `-x` without doing anything tricky.

src/utils.jl
dhairyagandhi96: This is to capture any cases of broadacsting the values over multiple channels etc

src/utils.jl
dhairyagandhi96: I thought that that should be an opt-in, and can be done with `collect`?

dhairyagandhi96: Happy to remove it, though, of course

src/utils.jl
dhairyagandhi96: Yes, will make that change

src/utils.jl
MikeInnes: We of course want to avoid materialising where possible, but if it's the only way to implement an operation, then I think it's fine to implement it that way in principle.

src/utils.jl
MikeInnes: Ah I see, the zeros might actually change the shape of `a`. We might want to check explicitly for that case somehow, since otherwise it loses the performance advantage of `Zeros`; but maybe we can kick that down the road for now.

src/utils.jl
dhairyagandhi96: Okay, will do so

src/utils.jl
```diff
@@ -139,6 +139,126 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+
+broadcasted(::typeof(+), a::Zeros, b::AbstractArray) = broadcasted(typeof(+), b, a)
+broadcasted(::typeof(-), a::Zeros, b::AbstractArray) = broadcasted(typeof(+), -b, a)
+
+function broadcasted(::typeof(*), a::AbstractArray, b::Zeros)
+  sz = similar(a, Broadcast.broadcast_shape(size(a), size(b)))
+  sz .= zero(a)
+end
+
+broadcasted(::typeof(*), a::Zeros, b::AbstractArray) = broadcasted(typeof(*), b, a)
```
MikeInnes: This looks wrong; `*` should be the argument to broadcasted, not `typeof(*)`. There are a few cases like this, do they have any tests?

src/utils.jl
MikeInnes: Could we remove these methods entirely? It seems like as-is, it's just reimplementing what broadcasting would do by default anyway (albeit without an extra addition, but that seems unlikely to make a noticeable performance difference). Correct me if I'm wrong ofc.

dhairyagandhi96: We hit a case of scalar indexing in that case:

with these methods
```julia
julia> @btime $(rand(3000)) .+ $(Flux.Zeros(3000,3000,2));
  61.056 ms (6 allocations: 137.33 MiB)

julia> @btime $(rand(3000)) .+ $(zeros(3000,3000,2));
  69.290 ms (2 allocations: 137.33 MiB)
```

w/o these methods

```julia
julia> @btime $(rand(3000)) .+ $(Flux.Zeros(3000,3000,2));
  43.785 s (359603510 allocations: 8.98 GiB)
```

I am not sure why it did that, perhaps something in the array definition?

dhairyagandhi96: Another option is to do a check for the sizes and return, else do the usual thing.

src/utils.jl
```diff
@@ -139,6 +139,121 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+  res .= zero(S)
+end
+
+function *(a::Zeros{T,2}, b::AbstractArray{S,2}) where {T,S}
+  @assert size(a,2) == size(b,1) throw(DimensionMismatch("A has dimensions $(size(a)) but B has dimensions $(size(b))"))
+  res = similar(b, size(a,1), size(b,2))
+  res .= zero(S)
+end
+
+Base.copy(xs::Zeros{T,N}) where {T,N} = Zeros(T, size(xs)...)
```
MikeInnes: As `Zeros` is immutable, can't this be a no-op?

src/utils.jl
```diff
@@ -139,6 +139,121 @@ function throttle(f, timeout; leading=true, trailing=false)
...
++(a::Zeros, b::AbstractArray) = b + a
+-(a::Zeros, b::AbstractArray) = -b + a
+
+function *(a::AbstractArray{S,2}, b::Zeros{T,2}) where {T,S}
+  @assert size(a,2) == size(b,1) throw(DimensionMismatch("A has dimensions $(size(a)) but B has dimensions $(size(b))"))
+  res = similar(a, size(a,1), size(b,2))
+  res .= zero(S)
+end
+
+function *(a::Zeros{T,2}, b::AbstractArray{S,2}) where {T,S}
```
MikeInnes: What's the motivation for this method?

src/utils.jl
dhairyagandhi96: Yes, it can, will make it so.

src/utils.jl
dhairyagandhi96: Largely to define matmul here to avoid scalar operations, iirc

dhairyagandhi96:
```Julia
Flux.Zeros(3000,3000)) * rand(3000,3000);
```
would hit `generic_matmul!` here, which is slow too

```Julia
julia> @btime $(zeros(3000,3000)) * $(rand(3000,3000));
  564.237 ms (2 allocations: 68.66 MiB)

julia> @btime $(Flux.Zeros(3000,3000)) * $(rand(3000,3000));
  7.953 ms (6 allocations: 68.66 MiB)
```

src/utils.jl
dhairyagandhi96: Fixed

dhairyagandhi96:
@MikeInnes I added the kwarg only constructors, and made some docs corrections too. Perhaps good to get this in for now?

dhairyagandhi96:
The error happens with or without the bias machinery

src/utils.jl
MikeInnes: That makes sense, I was just surprised that this method came up at all since `Zeros()` usually gets used as a bias. Did this come up from trying to support `Dense` layers with `Zeros()` for the weight?

If so it'd be good to discuss whether that's something we want to support anyway; it seems like an odd thing to want to do.

src/utils.jl
MikeInnes: That's a surprising slowdown; it seems worth digging into that, otherwise we're going to have to implement a lot of methods for things which should just be fast to begin with.

Of course, we could always merge this PR without the performance fixes, and follow up with an investigation on what's needed to make these things as fast as possible.

src/utils.jl
```diff
@@ -186,6 +186,121 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+    sz .= a
+  end
+end
+
+broadcasted(::typeof(+), a::Zeros, b::AbstractArray) = broadcasted(+, b, a)
+broadcasted(::typeof(-), a::Zeros, b::AbstractArray) = broadcasted(+, -b, a)
+
+function broadcasted(::typeof(*), a::AbstractArray, b::Zeros)
+  sz = similar(a, Broadcast.broadcast_shape(size(a), size(b)))
+  sz .= zero(a)
```
MikeInnes: I have the same concern with this as with the methods above, i.e. if we're going to allocate anyway we should just be able to use the built-in fallback.

dhairyagandhi96: We could also potentially just have it return a correctly shaped `Zeros` object to avoid allocating.

dhairyagandhi96: But I understand the concern, and I agree that that shouldn't be necessary

src/utils.jl
dhairyagandhi96: Partly to cover our ground in case it ends up being used outside of the context of a bias, and to cover the basic ops so it doesn't end up being accidentally slow

src/utils.jl
MikeInnes: Yeah, that seems like the best route for this particular method.

src/utils.jl
MikeInnes: There's an unlimited number of methods we could special case like this though, right? Given that we can't add everything I think we need to decide what we're supporting and why. For broadcasting `+` and `-` that's obvious enough (it covers the bias case) but I don't know that we want to implement every matmul method without a good use case for it.

dhairyagandhi96: Fair point

src/utils.jl
dhairyagandhi96: Wondering if this works as a workaround for now

dhairyagandhi96:
The run passes here https://travis-ci.org/FluxML/Flux.jl/jobs/658191484?utm_medium=notification&utm_source=github_status

CarloLucibello:
why we need to define our own types when we can use FillArrays.Zeros? FullArrays is already used by zygote.

It's very hard to review this PR, since it comes with a lot of unrelated doc additions

MikeInnes:
> FullArrays is already used by zygote.

I think the reasoning is that we want to represent not just an array of zeros, but an array which is held fixed with respect to gradient descent. There are a couple of ways to do this; either Zygote or FillArrays could decide that the gradient of a `Fill` is zero, but it's a bit hard to justify that outside of Flux (and probably breaks other uses of `Fill`). Alternatively Flux/Optimisers.jl could decide that `Fill` is a special case and doesn't get updated regardless of its gradient. This could again be a bit surprising and it's not clear how we'd document it (the type itself is the obvious choice, but we don't own that).

`import Flux: Zeros` sends a clear signal that this is a zero type that Flux knows about, which hopefully makes its behaviour feel more intuitively obvious, if not quite self-documenting.

Issue #871, opened 2019-09-26
ADAM does not accept keyword arguments


jessebett:
`?Flux.ADAM` gives the help string `ADAM(η = 0.001, β = (0.9, 0.999))`

But if you try to create an optimizer with those (unicode) keyword arguments then you get a method error for unexpected arguments.

MikeInnes:
Unfortunately this is a notation clash between [named] default arguments and keyword arguments. In signatures it's unambiguous because you have to use `;` for kwargs, but this doesn't work when you actually call a function.

How about writing the signature as

```julia
ADAM(η, β)
ADAM() => ADAM(0.001, (0.9, 0.999))
```

jessebett:
Yes I think the help string mislead me to think those were keyword arguments, but you're right that I should have looked for the `;`.

ToucheSir:
Can we call this closed by https://github.com/FluxML/Flux.jl/pull/904 and subsequent PRs?

darsnack:
The initial issue is the same, but that's because it is the standard way to document this situation in a docstring. I agree that having to look for `;` isn't helpful. I am pushing for only LR as a positional argument in Optimisers.jl (which would make this a non-issue).

Issue #836, opened 2019-08-10
Optimiser interface is not documented


purzelrakete:
The optimiser api is not documented anywhere. Here are some questions that came up for me:

- the relationship between `apply!` and `update!` is not clear just based on the names of these methods. It appears that `apply!` updates gradients, while `update!` updates parameters.
- it is not immediately obvious what `apply!` should mutate. Maybe just the optimiser? This is exacerbated by the fact that it returns `Δ`.  Is there a reason to return `Δ` when it is already being mutated? Some light documentation would quickly dispel any confusion.
- There is no documentation. I looked at the pull request for the optimiser changes as well as the commit message and the code itself. 
- Parameters are named `x`. This is not what I would expect, since x usually refers to the features. This could quickly be cleared up with some docstrings. 
- It is not immediately clear why parameters are being passed into `apply!`. It seems that we are not allowed to mutate `x`, and it also isn't used by any of the optimisers to compute gradient updates. It would help to document the the lifecycle of the `apply!` method to understand that we need to attach optimiser state to each individual parameter. The reason is that `apply!` can get called many times in the same update step, once for each parameter. Some documentation would be helpful. 
- The docstrings are not in sync with the coder, or not appropriate for the methods being described. Example: "Classic gradient descent optimiser with learning rate `η`. For each parameter `p` and its gradient `δp`, this runs `p -= η*δp`.". Names of the parameters don't match the code here, which uses `x` and `Δ` instead of `p` and `δp`. 

I think it might be helpful to explicity write out an interface along with some documentation. Then each Optimiser will need less documentation. Ignoring the fact that `Optimiser` already exists, something like this maybe:

```julia:
"""
Apply changes to gradients. 

Optimisers are called once for each parameter in each update step. Because of this, the parameter `x` is also passed to each `apply!` call. Optimisers can then do bookkeeping by namespacing changes to that parameter only, e.g. by keeping an internal dictionary of parameters to internal counters.
"""
abstract type Optimiser; end

"""
Applies changes to the gradient.

# Arguments
- `o`: the optimiser. Keep optimiser state in here and update on apply! as appropriate.
- `x`: the parameter being optimised. It may not be changed by apply.
- `Δ`: the gradients w.r.t the parameter. Gradients should be mutated by apply using elementwise operations.

# Returns
- `Δ`: the updated gradients w.r.t the parameter.
"""
apply!(o::Optimiser, x, Δ) = error("not implemented")

"""
    Descent(η)

[...]

PR #875, opened 2019-09-27
Documenting Optimiser Interface

dhairyagandhi96:
This PR introduces some docs to the optimiser interface, and applies basic fixes to the docstrings for the various optimisers
Ref #871 and #836

### DIFF
--- a/docs/src/training/optimisers.md
+++ b/docs/src/training/optimisers.md
@@ -66,14 +66,16 @@ Flux's optimsers are built around a `struct` that holds all the optimiser parame
 In this manner Flux also allows one to create custom optimisers to be used seamlessly. Let's work this with a simple example.
 
 ```julia
-mutable struct Momentum{T,S,D}
-  eta::T
-  rho::S
-  velocity::D
+mutable struct Momentum
+  eta
+  rho
+  velocity
 end
+
+Momentum(eta::Real, rho::Real) = Momentum(eta, rho, IdDict())
 ```
 
-The `Momentum` type will act as our optimiser in this case. Notice that we have added all the parameters as fields, along with the velocity which we will use as our state. **Note that this behaviour is set to change in consequent versions of Flux**. We can now define the rule applied when this optimiser is invoked.
+The `Momentum` type will act as our optimiser in this case. Notice that we have added all the parameters as fields, along with the velocity which we will use as our state dictionary. Each parameter in our models will get an entry in there. We can now define the rule applied when this optimiser is invoked.
 
 ```julia
 function apply!(o::Momentum, x, Δ)
@@ -85,19 +87,22 @@ end
 ```
 
 This is the basic definition of a Momentum update rule given by:
-$v = ρ * v - η * Δ$
-$w = w - v$
+
+```math
+v = ρ * v - η * Δ
+w = w - v
+```
 
 The `apply!` defines the update rules for an optimiser `opt`, given the parameters and gradients. It returns the updated gradients usually. Here, every parameter `x` is retrieved from the running state `v` and subsequently updates the state of the optimiser.
 
-Flux internally calls on this function via the `update!` function. It shares the API with `apply!` but ensures that multiple parameters are handled gracefully. In the future, it will also be delegating immutable update operations.
+Flux internally calls on this function via the `update!` function. It shares the API with `apply!` but ensures that multiple parameters are handled gracefully.
 
 ## Composing Optimisers
 
 Flux defines a special kind of optimiser called simply as `Optimiser` which takes in a arbitrary optimisers as input. Its behaviour is similar to the usual optimisers, but differs in that it acts by calling the optimisers listed in it sequentially. Each optimiser produces a modified gradient
 that will be fed into the next, and the resultant update will be applied to the parameter as usual. A classic use case is where adding decays is desirable. Flux defines some basic decays including `ExpDecay`, `InvDecay` etc.
 
-``julia
+```julia
 opt = Optimiser(ExpDecay(0.001, 0.1, 1000, 1e-4), Descent())
 ```
 
--- a/docs/src/training/optimisers.md
+++ b/docs/src/training/optimisers.md
@@ -94,7 +94,7 @@ Flux internally calls on this function via the `update!` function. It shares the
 
 ## Composing Optimisers
 
-Flux defines a special kind of optimiser called simply as `Optimiser` which takes in a arbitrary optimisers as input. Its behaviour is similar to the usual optimisers, but differs in that it acts by calling the optimsers listed in it sequentially. Each optimiser produces a modified gradient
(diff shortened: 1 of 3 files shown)


src/optimise/optimisers.jl
```diff
@@ -23,7 +24,8 @@ function apply!(o::Descent, x, Δ)
 end
 
 """
-    Momentum(η = 0.01; ρ = 0.9)
+    Momentum() => Momentum(η, ρ)
```
mcabbott: Making clear that the parameters aren’t keywords is great, but these `=>` things seem very odd to me. Isn’t it enought to write the defaults next to or below the definition? 
```suggestion
    Momentum(η, ρ) # Defaults: η = 0.01, ρ = 0.9
```

dhairyagandhi96: I think we also need to make clear that users don't need to specify the parameters if they want to use the defaults. That was kind of the hope with the previous docstring since Julia doesn't do named positional arguments, and defining defaults values like that is pretty common practice.

I'd be happy to make this change if there is consensus.

mcabbott: Agree with the goals, I just think `=>` is confusing. Maybe another way is to write this: 
```
    Momentum(η, ρ)
    Momentum() # Defaults: η = 0.01, ρ = 0.9
```

dhairyagandhi96: Right, I think the last commit is close to this, save the `$opt()`

Maybe we add that in

MikeInnes: Maybe we can just use `=`. That's exactly how you'd define it in Julia anyway, so it should be unambiguous.

MikeInnes: If we use something like `Defaults`, it should definitely be in the text rather than the signature; perhaps with a description of the parameters. That might actually be better since we can expand eta to "learning rate" etc. Probably it's a good idea to do both things.

src/optimise/optimisers.jl
```diff
@@ -25,7 +27,10 @@ end
 
 """
     Momentum(η, ρ)
-    Defaults: η = 0.01, ρ = 0.9
+
+    Calls to `Momentum()` default to:
```
MikeInnes: It's still strange to have text / a list inside the code block like this, when it could be normal markdown.

I'm really imagining something like a "parameters" section which outlines the meanings of parameters, and their defaults.

src/optimise/optimisers.jl
```diff
@@ -7,10 +7,29 @@ const ϵ = 1e-8
 # TODO: should use weak refs
 
 """
-    Descent(η)
+  Descent(η)
 
+## Description
```
MikeInnes: Looks like you have a stray `description` header here.

docs/src/training/optimisers.md
```diff
@@ -58,3 +58,78 @@ AMSGrad
...
+
+```julia
+mutable struct Momentum{T,S,D}
+  eta::T
+  rho::S
+  velocity::D
+end
+```
+
+The `Momentum` type will act as our optimiser in this case. Notice that we have added all the parameters as fields, along with the velocity which we will use as our state. **Note that this behaviour is set to change in consequent versions of Flux**. We can now define the rule applied when this optimiser is invoked.
```
MikeInnes: Reading this, it's not clear what the struct fields actually are. I think it'd be better, for pedagogical purposes, to leave the type parameters off entirely and have an example constructor like `Momentum(eta::Real, rho::Real) = ...`, to show the IdDict being constructed. We can then explain that the IdDict stores a state for each model parameter.

I think the warning is also unnecessary; many things could change, but we'll deprecate those things when we need to.

docs/src/training/optimisers.md
```diff
@@ -58,3 +58,78 @@ AMSGrad
...
+end
+```
+
+This is the basic definition of a Momentum update rule given by:
+$v = ρ * v - η * Δ$
+$w = w - v$
+
+The `apply!` defines the update rules for an optimsier `opt`, given the parameters and gradients. It returns the updated gradients usually. Here, every parameter `x` is retrieved from the running state `v` and subsequently updates the state of the optimiser.
+
+Flux internally calls on this function via the `update!` function. It shares the API with `apply!` but ensures that multiple parameters are handled gracefully. In the future, it will also be delegating immutable update operations.
```
MikeInnes: I would leave the "in future" comment off here too; users here want to understand the current API and I think this might distract from that.

docs/src/training/optimisers.md
```diff
@@ -58,3 +58,78 @@ AMSGrad
...
+
+## Composing Optimisers
+
+Flux defines a special kind of optimiser called simply as `Optimiser` which takes in a arbitrary optimisers as input. Its behaviour is similar to the usual optimisers, but differs in that it acts by calling the optimsers listed in it sequentially. Each optimiser produces a modified gradient
```
MikeInnes:
```suggestion
Flux defines a special kind of optimiser called simply as `Optimiser` which takes in a arbitrary optimisers as input. Its behaviour is similar to the usual optimisers, but differs in that it acts by calling the optimisers listed in it sequentially. Each optimiser produces a modified gradient
```

docs/src/training/optimisers.md
```diff
@@ -58,3 +58,78 @@ AMSGrad
...
+  @. v = ρ * v - η * Δ
+  @. Δ = -v
+end
+```
+
+This is the basic definition of a Momentum update rule given by:
+$v = ρ * v - η * Δ$
+$w = w - v$
+
+The `apply!` defines the update rules for an optimsier `opt`, given the parameters and gradients. It returns the updated gradients usually. Here, every parameter `x` is retrieved from the running state `v` and subsequently updates the state of the optimiser.
```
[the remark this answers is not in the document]
dhairyagandhi96: Gotcha

docs/src/training/optimisers.md
```diff
@@ -58,3 +58,78 @@ AMSGrad
...
+  η, ρ = o.eta, o.rho
+  v = get!(o.velocity, x, zero(x))::typeof(x)
+  @. v = ρ * v - η * Δ
+  @. Δ = -v
+end
+```
+
+This is the basic definition of a Momentum update rule given by:
+$v = ρ * v - η * Δ$
+$w = w - v$
```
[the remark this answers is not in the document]
MikeInnes: We probably need to upgrade Documenter to use Katex instead of Mathjax. Can do that as a follow up though

dhairyagandhi96:
bump

src/optimise/optimisers.jl
```diff
@@ -7,10 +7,28 @@ const ϵ = 1e-8
...
+
+opt = Descent(0.3) # use provided η
+
+ps = params(model)
+
+gs = gradient(ps) do
+  loss(x, y)
+end
+
+Flux.Optimise.update(opt, ps, gs)
```
MikeInnes:
```suggestion
Flux.Optimise.update!(opt, ps, gs)
```

dhairyagandhi96:
Closed via #904

PR #877, opened 2019-09-29
Fix functor's `params!` to work with complex numbers

PhilipVinc:


MikeInnes:
Good catch, thanks – the definition should probably just be `<:Number` to get everything..

PhilipVinc:
I thought there was a reason to use `<:Real`.
Like, to avoid irrationals or other things.

But in hindsight, maybe not.
I'll fix it

PhilipVinc:
Give me a couple days and I'll find the time to do it

PR #882, opened 2019-10-03
Check if CUDA availability changed during init.

maleadt:
With this PR, Flux checks using CUDAapi if CUDA is available during initialization, and forces recompilation if that does not agree with what was decided during precompilation. This avoids the scenario where Flux was precompiled without GPU support, consequently not allowing use of the GPU even if the user fixed his CUDA/GPU set-up because that does not force recompilation (and we can't add precompilation dependencies on stuff that doesn't exist).

However, we can't do the same for the case where we have a GPU/CUDA but CuArrays fails to import (checking if it imports during `__init__` would be much too expensive, if even possible), so this PR removes support for having CUDA/a GPU but CuArrays being broken. That's a little risky now that Flux depends on CuArrays, but the package is pretty mature and I haven't seen many bug reports failing to load it recently.

Fixes https://github.com/FluxML/Flux.jl/pull/852#issuecomment-538028314

cc @MikeInnes @xukai92


maleadt:
One case where the latter situation would crop up is if the user is using a terribly old GPU that CUDAnative does not support, which would make CuArrays and thus Flux fail to load. Maybe we should allow overriding using an environment variable (`FLUX_USE_CUDA`)?

MikeInnes:
This is a good idea; I worry a bit about the text being confusing (maybe "please restart Julia" would be harder to misunderstand?); in my experience new users tend not to think of the "just try again" route unless it's really explicit. But it hopefully won't come up very often, so we can see how it goes.

Issue #878, opened 2019-10-01
Functor differentiability


MikeInnes:
`fmap` currently errors when differentiating it, which is an issue mainly for RNNs (because of `reset!`). I think it's mostly just a case of fixing up adjoints for things like the set operations used internally by `fmap`; everything else should be pretty simple.

One test case:

```julia
struct Foo
  x
  y
end

@functor Foo

gradient(Foo(1, 2)) do f
  f = fmap(x -> 2x, f)
  f.y
end
```

This is an immediate issue due to the fact that we want to differentiate `reset!` in RNNs.

CarloLucibello:
it works now 
```julia
julia> using Flux, Functors

julia> struct Foo
         x
         y
       end

julia> @functor Foo

julia> gradient(Foo(1, 2)) do f
         f = fmap(x -> 2x, f)
         f.y
       end

((x = nothing, y = 2),)

```

DhairyaLGandhi:
Yeah this was fixed some time back

PR #885, opened 2019-10-04
WIP: `fmap` Adjoint

dhairyagandhi96:
Addresses #878 

I am a little unsure about whether the correct result to return here would be a `Foo(2,2)` object or a simple `2`.

cc @MikeInnes

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -39,10 +39,10 @@ end
 
 @adjoint function Flux.fmap(f, x)
   op = Flux.fmap(f, x)
-  back = Flux.fmap(x) do x_
+  back(del) = Flux.fmap(del) do x_
     f'(x_)
   end
-  op, Δ -> (nothing, back)
+  op, Δ -> (nothing, back(Δ))
 end
 
 trainable(m) = functor(m)[1]
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -32,13 +32,23 @@ function fmap1(f, x)
   re(map(f, func))
 end
 
-@adjoint function fmap1(f, x)
-  back(del) = fmap1(del) do x_
-    x_ isa Nothing && return
-    f'(x_)
+@adjoint function Base.map(f, x::NamedTuple)
+  function back(del)
+    del = map(del) do x_
+      x_ isa Nothing && return
+      f'(x_)
+    end
   end
-  fmap1(f, x), Δ -> (nothing, back(Δ))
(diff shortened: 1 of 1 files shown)


MikeInnes:
I was actually imagining that we'd fix anything in Zygote that prevented `fmap` from running (e.g. when I tried the RNN example it was an issue with set operations), and let `fmap` itself get derived.

An explicit adjoint may eventually be useful for performance, but we should avoid that unless we have to.

dhairyagandhi96:
Ah alright, that would be adjoints for `isleaf`, `haskey` and `setindex!`

I'll make the changes

dhairyagandhi96:
The char-rnn does work on current Flux#master + Zygote#master, but we really do need to have these adjoints regardless, just an interesting observation

MikeInnes:
I was wondering about that, given that the other test case worked. I wonder why that changed.

If that's the case, then for the purposes of this issue we should just add a test representing that char-rnn example, so it doesn't break again.

dhairyagandhi96:
I believe these would be useful beyond this case as well, so good to have. Also, the mwe would need to be addressed.

I'll add some tests

MikeInnes:
@kshyatt found a case that triggers the original error: `cpu(x)`. Would be great to turn that into a minimal example.

dhairyagandhi96:
Right, I think the issue is triggered most easily when `gpu` and `cpu` are involved. Will use the adjoint in this PR to verify the adjoints I'm writing. I think `functor` tests should have the MWE, gpu and cpu and perhaps a run of the char-rnn model (on GitLab)

src/functor.jl
```diff
@@ -32,19 +32,21 @@ function fmap1(f, x)
...
 
-@adjoint function Flux.fmap(f, x)
-  op = Flux.fmap(f, x)
-  back(del) = Flux.fmap(del) do x_
-    x_ isa Nothing && return
-    f'(x_)
-  end
-  op, Δ -> (nothing, back(Δ))
-end
+@adjoint haskey(cache, x) = false, _ -> nothing
```
MikeInnes: This seems like it should be in Zygote. Can't the adjoints for fmap1/isleaf be derived?

dhairyagandhi96:
Yeah, I really should've opened it against Zygote

dhairyagandhi96:
Moved to FluxML/Zygote.jl#372

PR #893, opened 2019-10-14
Update Dependencies

MikeInnes:


maleadt:
I did an update in https://github.com/FluxML/Flux.jl/pull/916 that passed CI, so I guess this can be closed.

PR #899, opened 2019-10-19
Update travis config and readme file for coverage

### LINKED ISSUE #89: Coverage
It would be great to

1) [ ] Turn on coverage in the Travis script and put the banner in the README.
2) [ ] Improve the coverage. Right now it is around 38% according to `Coverage.jl`

hossein-pourbozorg:
Fixes #89

I updated travis config file and added coverage tools to it. Also, I inserted their badges to readme file too. We should use travis-ci.com instead of travis-ci.org because of travis-ci announced that users and projects (both private and public) should only use travis-ci.com. ([ref1](https://blog.travis-ci.com/2018-05-02-open-source-projects-on-travis-ci-com-with-github-apps) [ref2](https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/))

You can see the results for my fork:
https://codecov.io/gh/hossein-pourbozorg/Flux.jl/branch/issue-89
https://coveralls.io/github/hossein-pourbozorg/Flux.jl

https://travis-ci.com/hossein-pourbozorg/Flux.jl/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,27 +3,28 @@ language: julia
 
 os:
   - linux
-  # - osx
 
 julia:
   - 1.1
+  - 1.2
   - nightly
 
 matrix:
   allow_failures:
     - julia: nightly
+  fast_finish: true
+
+codecov: true
+
+coveralls: true
 
 jobs:
   include:
-    - stage: "Documentation"
+    - stage: Documentation
       julia: 1.0
-      os: linux
-      script:
-        - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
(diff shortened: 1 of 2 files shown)


MikeInnes:
Thanks for the effort, but I don't think we have a great need for coverage (and adding it is not without downsides, e.g. noise on issues). I do think we should keep improving test coverage as features come in, but the most urgent thing is to do more combined CPU/GPU tests, which is not well-measured by coverage stats.

Looking at the travis CI refs, it also looks as though that advice only applies to new projects, and existing ones need to be migrated; the [Flux page on travis-ci.com](https://travis-ci.com/FluxML/Flux.jl) is currently a 404.

PR #906, opened 2019-10-23
Add CUDA tests for stateless layers

kshyatt:
A lot of `@test_broken` here because of CUDAnative `log` fighting and scalar `getindex` being disallowed.

### DIFF
--- a/test/cuda/cuda.jl
+++ b/test/cuda/cuda.jl
@@ -25,11 +25,27 @@ cm = gpu(m)
 @test all(p isa CuArray for p in params(cm))
 @test cm(gpu(rand(10, 10))) isa CuArray{Float32,2}
 
+x = rand(3) 
+y = rand(3) 
+cx = gpu(x)
+cy = gpu(y)
+@test Flux.mse(cx, cy) ≈ Flux.mse(x, y)
 x = [1,2,3]
 cx = gpu(x)
 @test Flux.crossentropy(x,x) ≈ Flux.crossentropy(cx,cx)
 @test Flux.crossentropy(x,x, weight=1.0) ≈ Flux.crossentropy(cx,cx, weight=1.0)
 @test Flux.crossentropy(x,x, weight=[1.0;2.0;3.0]) ≈ Flux.crossentropy(cx,cx, weight=cu([1.0;2.0;3.0]))
+x = [-1.1491, 0.8619, 0.3127]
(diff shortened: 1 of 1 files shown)


kshyatt:
bors try

test/cuda/cuda.jl
```diff
@@ -25,11 +25,27 @@ cm = gpu(m)
...
 cx = gpu(x)
 @test Flux.crossentropy(x,x) ≈ Flux.crossentropy(cx,cx)
 @test Flux.crossentropy(x,x, weight=1.0) ≈ Flux.crossentropy(cx,cx, weight=1.0)
 @test Flux.crossentropy(x,x, weight=[1.0;2.0;3.0]) ≈ Flux.crossentropy(cx,cx, weight=cu([1.0;2.0;3.0]))
+x = [-1.1491, 0.8619, 0.3127]
+y = [1, 1, 0]
+cx = gpu(x)
+scx = σ.(cx)
+cy = gpu(y)
+@test_broken Flux.binarycrossentropy.(σ.(x), y) ≈ Flux.binarycrossentropy.(scx, cy)
```
MikeInnes: I'm curious why you split out `scx` here, is it just a style thing or does it change the bug/behaviour somehow?

MikeInnes:
Hoping this will let you `try`:

bors d+

test/cuda/cuda.jl
kshyatt: Was trying to see if the sigma was causing the compilation error, or the log.

kshyatt:
bors try

dhairyagandhi96:
Failure on 1.0 seems network related

dhairyagandhi96:
bors try

dhairyagandhi96:
bors try

Since #940 was merged, I think the normalise is what we miss currently

dhairyagandhi96:
Yup, seems like it

dhairyagandhi96:
@maleadt can say more, but I think now we don't want to compare GPU Arrays with regular ones.

CarloLucibello:
@kshyatt if you don't mind I'll add these tests to #1264

ToucheSir:
#1264 was merged with non-broken gpu losses, so can this be closed?

DhairyaLGandhi:
Yup, thanks for nudging in the right direction @kshyatt !

PR #907, opened 2019-10-24
Change `gate` function to `view` instead of copy

janEbert:
This speeds up code with large inputs by quite a lot. I only added it to the function accepting an `AbstractVector` as input as copying matrices may be faster than viewing them due to caching (they are sliced per row so will the data will not necessarily have a low stride).


MikeInnes:
Seems like a good idea, thanks!

bors r+

PR #913, opened 2019-10-31
Update has_cudnn

xukai92:
The current master of CuArrays doesn't export `libcudnn` string any more. It also doesn't export `has_cudnn()` during precompilation time for other packages. So I implement a `has_cudnn()` inside Flux using CUDAapi.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -26,6 +26,7 @@ const consider_cuda = allow_cuda()
 
 using CUDAapi
 const use_cuda = consider_cuda && has_cuda()
+has_cudnn() = find_cuda_library("cudnn", find_toolkit()) !== nothing
 if use_cuda
   try
     using CuArrays
--- a/src/cuda/cuda.jl
+++ b/src/cuda/cuda.jl
@@ -2,7 +2,7 @@ module CUDA
 
 using ..CuArrays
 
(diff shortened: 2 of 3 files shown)


maleadt:
Weird, I didn't get a mail for this review request. Anyhow, `has_cudnn` is still there, so there's no need to reimplement it here. The current master of CuArrays is incompatible with Flux too, as there's some package loading changes, and I'll create a PR to fix all that next week.

xukai92:
Is it available at precompilation period though? I can access to has_cudnn when using CuArrays directly but not when Flux is precompiling. I look at the init function of CuArrays and believe it is exactly what coded to behave.

maleadt:
> Is it available at precompilation period though? I can access to has_cudnn when using CuArrays directly but not when Flux is precompiling.

No, and that's one of the changes incoming in https://github.com/FluxML/Flux.jl/pull/916. The CuArrays.CUDNN module and its functionality is always going to be available, but the use thereof should be conditioned on `CuArrays.has_cuda()`. Otherwise the state of the precompilation image depends on the availability of CUDNN, and that causes several issues as documented in that PR.

xukai92:
I see. Thanks. I'm closing this PR now.

PR #919, opened 2019-11-05
update Zygote version in compat

findmyway:


MikeInnes:
We should be able to `free Zygote` and `rm IRTools ZygoteRules` now, too, so that fewer things are on master branches.

MikeInnes:
Not at all, I think this change was right to do (but I've gone ahead and done it in https://github.com/FluxML/Flux.jl/pull/920).

PR #924, opened 2019-11-07
CUDA package initialization improvements

maleadt:
Addresses https://github.com/FluxML/Flux.jl/pull/916#issuecomment-550758133 (cc @findmyway).

@MikeInnes Does Flux need to warn if CUDA isn't available?


MikeInnes:
No, I don't think so, lots of people are happy with just CPU.

MikeInnes:
FWIW, this is fine to merge whenever you're happy with it.

Issue #921, opened 2019-11-06
`train!` method for `Chain`


heliosdrm:
The second argument of `train!` is meant to be a set of tracked parameters of the model, typicall obtained from `params(model)`. While there is no theoretical restriction on how the user makes that `model` -- except that it should contain tracked parameters --, it is typically made as a  `Chain` of multiple layers, i.e. `typeof(model) <: Flux.Chain`.

Is there any reason (e.g. performance, etc.) against defining a method for `train!(loss, model::Chain, data, opt; kwargs...) = train!(loss, params(model), data, opt; kwargs...)`? I think that this interface would be more intelligible, and straightforward in many use cases.

If this is a reasonable idea, I can make a PR.

MikeInnes:
I think this would have the potential to cause confusion, without adding all that much. We're talking about the difference between `params(model)` and `model`, which isn't a big saving. We generally expect that people will learn how to write their own training loops quite quickly, so hiding `params` could make that process slower.

heliosdrm:
I agree that this suggestion does not save much typing. The motivation was to provide an interface that is one step closer to the interface of other ML libraries, where the functions for training take the model as argument. Honestly, the current interface of `train!` did make me feel confused the first time I saw it. ("hm... where is the model?")

Perhaps this interface has the purpose of helping users understand that it is the parameters of the model what they actually "train" -- or modify in the training, not the model itself. If it is like that, I'm ok with leaving as it is. Maybe it's the documentation and not the function what might be modified to make this clearer?

dhairyagandhi96:
Would really appreciate a note in the docs!

PR #927, opened 2019-11-08
Extend docs about `train!`

heliosdrm:
Related to #921: explain why it is not needed to pass the model as argument.

### DIFF
--- a/docs/src/training/training.md
+++ b/docs/src/training/training.md
@@ -1,6 +1,6 @@
 # Training
 
-To actually train a model we need three things:
+To actually train a model we need three things, in addition to the tracked parameters that will be fitted:
 
 * A *objective function*, that evaluates how well a model is doing given some input data.
 * A collection of data points that will be provided to the objective function.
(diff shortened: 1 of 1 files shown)


docs/src/training/training.md
```diff
@@ -11,6 +11,7 @@ With these we can call `Flux.train!`:
 ```julia
 Flux.train!(objective, params, data, opt)
 ```
+At first glance it may seem strange that the model that we want to train is not part of the input arguments of `Flux.train!`. However the target of the optimizer is not the model itself, but the objective function that represents the departure between modelled and observed data. In other words, the model is implicitly defined in the objective function, and there is no need to give it explicitly. Passing the objective function instead of the model and a cost function separately (see below) provides more flexibility, and the possibility of optimizing the calculations.
```
MikeInnes: I think this should be the first paragraph of the "loss functions" section just below, but otherwise LGTM.

heliosdrm: Really the first paragraph, or maybe rather the last one?

MikeInnes: Sure, that works well too.

heliosdrm: Moved to the suggested place.

docs/src/training/training.md
```diff
@@ -1,6 +1,6 @@
 # Training
 
-To actually train a model we need three things:
+To actually train a model we need three things, in addition to the tracked parameters that will be fitted:
```
MikeInnes: This should probably just be changed to "four things" and list parameters in the right place.

heliosdrm: Made as suggested, and also included a sub-section about the params argument. There is not a sub-section about the optimiser, but I guess this is already covered by the link to the page dedicated to it.

docs/src/training/training.md
```diff
@@ -1,8 +1,9 @@
 # Training
 
-To actually train a model we need three things:
+To actually train a model we need four things:
 
 * A *objective function*, that evaluates how well a model is doing given some input data.
+* The parameters of the model.
```
MikeInnes:
```suggestion
* The trainable parameters of the model.
```

MikeInnes:
Thanks @heliosdrm!

PR #932, opened 2019-11-15
Travis: test on 1.0

MikeInnes:


MikeInnes:
I had to avoid running a `GroupNorm` gradient test on 1.0 to get this running. Not ideal but it's going to take a bunch of time to properly debug why Zygote compiler things are going wrong on 1.0 here.

MikeInnes:
@dhairyagandhi96 pointed out that this bors has been testing 1.0 for a while, so the issue can't be due to bitrot on 1.0. In fact, it looks like an issue with one of the backports to 1.0.5. We'll narrow this down further if anyone complains, but more likely it's easier to just use 1.0.4.

PR #934, opened 2019-11-19
Fix Glorot initialization, add He initialization

### LINKED ISSUE #442: Glorot initialization not correctly implemented
I think the Glorot initializations are not correctly implemented.

https://github.com/FluxML/Flux.jl/blob/cb773e54c0dbb5337bbd3318acb3ae79c1979b01/src/utils.jl#L4-L5

e.g. `glorot_uniform` should sample from a uniform distribution [-limit, limit] where limit = sqrt(6 / (fan_in + fan_out))

For a dense layer the `sum(dims)` will give the correct result for (fan_in + fan_out), but I think this does not hold in the case of convolutions anymore. I think here it should be `*(dims[1:end-1]...) + dims[end]`

Sleort:
Should fix #442 .
Adds He weight initialization as a bonus :-)

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,6 +1,12 @@
 # Arrays
-glorot_uniform(dims...) = (rand(Float32, dims...) .- 0.5f0) .* sqrt(24.0f0/sum(dims))
-glorot_normal(dims...) = randn(Float32, dims...) .* sqrt(2.0f0/sum(dims))
+nfan(n_in, n_out) = n_in, n_out #fan-in, fan-out
+nfan(dims...) = prod(dims[1:end-2]) .* (dims[end-1], dims[end]) #In case of convolution kernels
+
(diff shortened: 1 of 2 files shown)


MikeInnes:
Why the close, is this superceded by something?

At any rate, it'd be best to split this into two PRs.

MikeInnes:
Just saw #937.

Sleort:
> Just saw #937.

Sorry about that. I was too trigger-happy with the first commit, and discovered I needed to fix a couple of things before resubmitting...

(I'm kind in a learning phase regarding git / Github, so please let me know if this wasn't the proper way of doing it... )

MikeInnes:
Yes, you can change the PR so there's generally no need to open a new one for the same thing. Adding new commits and pushing them is the simplest way; if you want to go a bit more advanced and clean up history the most useful option to learn about is `git rebase -i`.

Sleort:
Thanks for the tip!

Issue #923, opened 2019-11-06
Spurious RNN failure with CUDNN


maleadt:
On 8a0745f, which uses CuArrays 1.4.2:

```
R = Flux.LSTM: Error During Test at /builds/JuliaGPU/Flux.jl/test/cuda/curnn.jl:4
  Got exception outside of a @test
  CUDNNError(code CUDNN_STATUS_BAD_PARAM, CUDNN_STATUS_BAD_PARAM)
  Stacktrace:
   [1] cudnnRNNForwardTraining(::Ptr{Nothing}, ::CuArrays.CUDNN.RNNDesc{Float32}, ::Int64, ::Array{CuArrays.CUDNN.TensorDesc,1}, ::CuArray{Float32,1,Nothing}, ::CuArrays.CUDNN.TensorDesc, ::CuArray{Float32,1,Nothing}, ::CuArrays.CUDNN.TensorDesc, ::CuArray{Float32,1,Nothing}, ::CuArrays.CUDNN.FilterDesc, ::CuArray{Float32,1,Nothing}, ::Array{CuArrays.CUDNN.TensorDesc,1}, ::CuArray{Float32,1,Nothing}, ::CuArrays.CUDNN.TensorDesc, ::CuArray{Float32,1,Nothing}, ::CuArrays.CUDNN.TensorDesc, ::CuArray{Float32,1,Nothing}, ::CuArray{UInt8,1,Nothing}, ::Int64, ::CuArray{UInt8,1,Nothing}, ::Int64) at /root/.julia/packages/CuArrays/YbsYr/src/dnn/error.jl:19
[...]

PR #935, opened 2019-11-19
Fix AMSGrad on GPU

baggepinnen:
The previous initialization created a CPU array. Now, the same type of array as `x` is created.


MikeInnes:
CI failure is due to #923

PR #937, opened 2019-11-19
Fix Glorot initialization, add He initialization

### LINKED ISSUE #442: Glorot initialization not correctly implemented

Sleort:
Should fix #442 .
Adds He weight initialization as a bonus :-)

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,12 +1,14 @@
 # Arrays
-nfan(n_in, n_out) = n_in, n_out #fan-in, fan-out
+nfan() = 1, 1 #fan_in, fan_out
+nfan(n) = 1, n #A vector is treated as a n×1 matrix
+nfan(n_out, n_in) = n_in, n_out #In case of Dense kernels: arranged as matrices
 nfan(dims...) = prod(dims[1:end-2]) .* (dims[end-1], dims[end]) #In case of convolution kernels
 
 glorot_uniform(dims...) = (rand(Float32, dims...) .- 0.5f0) .* sqrt(24.0f0 / sum(nfan(dims...)))
 glorot_normal(dims...) = randn(Float32, dims...) .* sqrt(2.0f0 / sum(nfan(dims...)))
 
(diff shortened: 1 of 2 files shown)


MikeInnes:
Best to split this into two PRs to make it easier to review and merge.

Sleort:
I couldn't figure out how to split the PR into two, so I deleted the He initialization stuff (for now). 

(And sorry for the mess. I'm not yet completely familiar/comfortable with how Git(hub) works...)

MikeInnes:
No worries. Splitting the commit is possible but it's usually easier just to delete code and/or copy-paste it somewhere else for the `he` PR, so that's the right thing to do.

Sleort:
Anything else I should do to get this PR reviewed / merged?

MikeInnes:
I'm a bit uncomfortable with the heuristic this uses, but at least it has good precedent in other libraries. We can always redesign in future if needed.

Thanks @Sleort!

bors r+

Sleort:
> I'm a bit uncomfortable with the heuristic this uses, but at least it has good precedent in other libraries. We can always redesign in future if needed.

I agree (see the discussion at #442). A more "proper" way of treating this would, however, require breaking changes, since we would have to dispatch on layer type in addition to dimensionality. Actually, ideally we should maybe make the initialization depend on the activation function as well, such that the "right" initialization could be automatically selected (Glorot for `tanh`, He for `relu`, etc.)...?

MikeInnes:
We don't generally consider changing a layer's default initialisation to be breaking, so we could make the defaults smarter at any time, even if it isn't exposed as an API. A more general user-accessible initialisation API would be nice to have though.

PR #946, opened 2019-11-28
compat, pkg up

MikeInnes:
Needed to finalise the release process.

We still have a dependency on CuArrays master here. Usually we'd try to get on release branches, but we know CuArrays is solid apart from an annoying CUDNN bug that is a problem for our CI. So I think we can keep it here and release a patch for CuArrays later.

cc @dhairyagandhi96

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.9.0"
+version = "0.10.0"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -24,8 +24,17 @@ ZipFile = "a5390f91-8eb1-5f08-bee0-b1d1ffed6cea"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
+AbstractTrees = "0.2"
+Adapt = "1"
+CodecZlib = "0.5, 0.6"
+Colors = "0.8, 0.9"
 CuArrays = "1.4.3"
```
dhairyagandhi96: Shouldn't CuArrays be bumped too?

MikeInnes: Given how rare the RNN issue is, I think we should just let the resolver install 1.4.3 if it really needs to, rather than erroring out. Of course in most cases it should just install the most recent version anyway.

MikeInnes:
CI is flaky again, but tested locally.

PR #954, opened 2019-12-03
[WIP] Decaydocs

baggepinnen:
I found the docstring to leave a lot for the imagination regarding the workings of the decays optimizers. It's still unclear to me why we need to provide a step size to `ExpDecay` when there is a step size in the wrapped optimizer.


dhairyagandhi96:
LGTM

dhairyagandhi96:
Thanks for the fixes!

PR #957, opened 2019-12-05
Add some gradient checking tests on GPUs

dhairyagandhi96:
Good to add generic tests for tracking gradients through the various layers on the GPU.

### DIFF
--- a/test/cuda/layers.jl
+++ b/test/cuda/layers.jl
@@ -1,0 +1,86 @@
+# Test layers and data/model movements on and off the GPU
+# Add tests for layers and their gradients on the GPU
+# Most of the forward passes should be fine being applied
+# to bitstype objects, but this gives higher coverage for our use-cases
+# Check that getting the gradients does not throw
+
+# generic movement tests
+@test_broken gradient(x -> sum(gpu(x)), rand(3,3)) isa Tuple
+@test gradient(x -> sum(cpu(x)), gpu(rand(3,3))) isa Tuple
+
+function gradtest(layers, args...; name = "Conv", xs = rand(Float32, 28, 28, 1, 1))
+	@testset "$name GPU grad tests" begin
+		for layer in layers
+			@testset "$layer GPU grad test" begin
+				l = gpu(layer(args...))
+				xs = gpu(xs)
+				if l isa DepthwiseConv
+					@test_broken gradient(Flux.params(l)) do
+						sum(l(xs))
+					end isa Flux.Zygote.Grads
+				else
+					@test gradient(Flux.params(l)) do
+						sum(l(xs))
+					end isa Flux.Zygote.Grads
+				end
+			end
+		end
+	end
+end
+
+# Repeats from Conv, CrossCor
+
+conv_layers = [Conv, ConvTranspose, CrossCor, DepthwiseConv]
+gradtest(conv_layers, (2,2), 1=>3, name = "Conv")
+
+pooling_layers = [MaxPool, MeanPool]
+gradtest(pooling_layers, (2,2), name = "Pooling")
+
+dropout_layers = [Dropout, AlphaDropout]
+gradtest(dropout_layers, 0.5f0, name = "Dropout")
+
+norm_layers = [LayerNorm, BatchNorm]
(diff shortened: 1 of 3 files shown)


MikeInnes:
Seems like a good idea, but CI is very unhappy for some reason.

dhairyagandhi96:
From the CI, it looks like something is missing a `git-tree-sha` in the Manifest.

dhairyagandhi96:
Looks like its CuArrays.

dhairyagandhi96:
bors try

dhairyagandhi96:
CI Failure: https://gitlab.com/JuliaGPU/Flux.jl/-/jobs/378522748

`cpu` seems to dabble in scalar indexing. Or rather, hashing `CuArray`s seems to trigger it.

dhairyagandhi96:
bors try

dhairyagandhi96:
A segfault shows up in the test for Dropouts. Unless there's anything incorrect in the tests, might be worth investigating

test/cuda/layers.jl
```diff
@@ -0,0 +1,86 @@
...
+@test_broken gradient(x -> sum(gpu(x)), rand(3,3)) isa Tuple
+@test_throws ErrorException gradient(x -> sum(cpu(x)), gpu(rand(3,3))) isa Tuple
+
+function gradtest(layers, args...; name = "Conv", xs = rand(Float32, 28, 28, 1, 1))
+  @testset "$name GPU grad tests" begin
+    for layer in layers
+      @testset "$layer GPU grad test" begin
+        l = gpu(layer(args...))
+        xs = gpu(xs)
+        if l isa DepthwiseConv
```
MikeInnes: It seems a little weird to have this information here. I guess it's not too bad if it's only one case though.

test/cuda/layers.jl
```diff
@@ -0,0 +1,86 @@
...
+          @test_broken gradient(Flux.params(l)) do
+            sum(l(xs))
+          end isa Flux.Zygote.Grads
+        else
+          @test gradient(Flux.params(l)) do
+            sum(l(xs))
+          end isa Flux.Zygote.Grads
```
MikeInnes: Would this perhaps look clearer with a short-form lambda?

test/cuda/layers.jl
```diff
@@ -0,0 +1,86 @@
...
+# Add tests for layers and their gradients on the GPU
+# Most of the forward passes should be fine being applied
+# to bitstype objects, but this gives higher coverage for our use-cases
+# Check that getting the gradients does not throw
+
+# generic movement tests
+@test_broken gradient(x -> sum(gpu(x)), rand(3,3)) isa Tuple
+@test_throws ErrorException gradient(x -> sum(cpu(x)), gpu(rand(3,3))) isa Tuple
+
+function gradtest(layers, args...; name = "Conv", xs = rand(Float32, 28, 28, 1, 1))
```
MikeInnes: Probably best to avoid these defaults, otherwise it might be confusing if someone forgets to supply these arguments.

Something obviously-wrong or just an `error("...")` might be better.

dhairyagandhi96: Done

test/cuda/layers.jl
dhairyagandhi96: Could you take a look at this now?

test/cuda/layers.jl
dhairyagandhi96: Yeah, only one of them outright borks, so just thought adding it here is easier, as long as the case can be caught

dhairyagandhi96:
bors try

dhairyagandhi96:
So the refresh seems to have revealed 4 layers that need attention.

### Scalar Indexing
- `DepthwiseConv`
- `InstanceNorm`
- `GroupNorm`

### Compilation Error
- `AlphaDropout`

cc @MikeInnes

dhairyagandhi96:
Interesting. The same commit works on the local setup, being able to differentiate the `gpu` and `cpu` calls. Also, it segfaults for some reason.

dhairyagandhi96:
bors try

PR #960, opened 2019-12-06
Added utility function outdims to compute output dimensions of a layer

darsnack:
Based on Slack chatter, I added a utility function, `outdims`, that computes the output dimensions for given input dimensions.

Example
```julia
layer = Conv((3, 3), 3 => 16)
outdims(layer, (10, 10)) # returns (8, 8)
```


test/layers/conv.jl
```diff
@@ -107,3 +107,23 @@ end
     true
   end
 end
+
+@testset "conv output dimensions" begin
+  m = Conv((3, 3), 3 => 16)
```
baggepinnen: Would be good to have some tests for different padding, stride and dilation

MikeInnes:
I can see this being useful. Note that not every method needs a docstring; you can just give one docstring with a couple of examples, otherwise the docs for `outdims` will be pretty noisy.

You can probably reuse the tools in NNlib to calculate this for convolutions, which will make it much less likely to miss anything.

darsnack:
Okay switched to NNlib.jl, added some more tests w.r.t. pad, stride, etc., and also added a small section to the docs.

darsnack:
Any ideas on the build errors? I don't really follow the stack trace, and I am not able to reproduce it on my local system.

MikeInnes:
Quite unlikely to be your fault, but we should get it passing. If you push a new commit the error might just go away.

darsnack:
> Quite unlikely to be your fault, but we should get it passing. If you push a new commit the error might just go away.

So I guess this didn't work. Have you seen this issue pop up with any other builds since I last pushed?

MikeInnes:
We have had a bit of flakiness lately. 

bors try

ianshmean:
It would be great to revive this idea. Perhaps worth rebasing and re-running tests?

darsnack:
@MikeInnes looks like the build is passing now? Though Travis is only testing for Julia 1.3?

dhairyagandhi96:
bors try

Let's try this once to see if things are settled down. Post which thoughts on documenting it's scope, since it might be odd to get it running with arbitrary functions?

darsnack:
Sure I can push a commit updating the doc section to specify which layers it works on?

ianshmean:
Perhaps the docs could state which layers it works for out of the box, and give the outdims conv method code to give a handy starting point for people to generate an outdims method for custom layers?

ianshmean:
This is looking good to me and I'm keen to start using these functions. 
Is it ready to be merged?

darsnack:
> Is it ready to be merged?

I am satisfied with merging it in the current form.

dhairyagandhi96:
Happy to merge this now
bors r+

PR #964, opened 2019-12-09
Allowing for function replacement in LSTM and GRU layers

NTimmons:
This makes the behaviour of the LSTM and GRU recurrent layers consistent with other layers.


MikeInnes:
This seems like a good thing to have, although perhaps with more informative names than `sigmRepl` etc; is there a better name for these gates?

Also, the CUDA bindings will have to be updated, since only layers with the default setup will be able to use CUDNN.

NTimmons:
Ah, I hadnt seen the CUDA bindings. I'll get on that

I'll take a look to see if the papers call the gates something particular. I was going to go with σ and tanh for the names to stay in sync with the other layers which just use σ but thought that would be extra confusing hah

NTimmons:
For the CUDA bindings, without going into CUDNN, I don't think there is much I can do except throw a message saying that the selected functions aren't GPU compatible? I looked at the CUDNN docs (https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html) and it looks like it has a hardcoded implementation? I am not sure...

Probably would restrict submission of a custom function LSTM/GRU to the GPU here though? 
```
curnn.jl:10
function CUDNN.RNNDesc(m::CuRNNs{T}) where T
  h, i = length(m.h), size(m.Wi, 2)
  mode = m isa CuRNN ?
    (m.σ == tanh ? CUDNN.CUDNN_RNN_TANH : CUDNN.CUDNN_RNN_RELU) :
    m isa CuGRU ? CUDNN.CUDNN_GRU : CUDNN.CUDNN_LSTM
  r = CUDNN.RNNDesc{T}(mode, i, h)
  return r
end
```

For the replacement names - we would need to call them the input, output and forget gates.
But the input and output gate functions would be a pair (sigm, tanh). As seen here...

forgetgate

I would rather keep the current setup with just the two functions because I want to use it for replacing sigmoid with better sigmoids and tanh with better tanh. 

But, I know that others are interested in controlling the actual model where they may want to vary the input and output gates independently. 

I am happy either way though, so let me know what you think would be best.

MikeInnes:
RE cuda, you can look [here](https://github.com/FluxML/Flux.jl/blob/ac4c49b63e546a6830f669920c65114b6b25dfda/src/cuda/curnn.jl#L5). The situation is the same for vanilla RNNs, we allow the activation function to be changed but only two are supported by CUDNN. Changing the `CuRNN` signature prevents incompatible RNNs from ever dispatching to CUDNN, they just fall through to the Julia implementation instead; the same should be possible for LSTMs.

NTimmons:
Hi,
I have done the changes required for CUDA.
Is there any performance benchmarks/tests I could run to ensure that my changes haven' t caused any performance regression?

NTimmons:
Updated the repo with the CUDA support. Added tests to the Cuda test files so that we can check that explicitly stating the sigmoid and tanh functions works as well as a test with a random function.

NTimmons:
Is anything else needed for this to be merged?

src/cuda/curnn.jl
```diff
@@ -1,17 +1,17 @@
...
+CuGRU{T}  = Flux.GRUCell{<:CuArray{T,2},<:CuArray{T,1}}#Flux.GRUCell{<:CuArray{T,2},<:CuArray{T,1},<:typeof(sigmoid),<:typeof(tanh)}
+CuLSTM{T} = Flux.LSTMCell{<:CuArray{T,2},<:CuArray{T,1},<:typeof(sigmoid),<:typeof(tanh)}
 CuRNNs{T} = Union{CuRNN{T},CuGRU{T},CuLSTM{T}}
 
 function CUDNN.RNNDesc(m::CuRNNs{T}) where T
   h, i = length(m.h), size(m.Wi, 2)
   mode = m isa CuRNN ?
     (m.σ == tanh ? CUDNN.CUDNN_RNN_TANH : CUDNN.CUDNN_RNN_RELU) :
-    m isa CuGRU ? CUDNN.CUDNN_GRU : CUDNN.CUDNN_LSTM
+    (m isa CuGRU ? CUDNN.CUDNN_GRU : (m isa CuLSTM ? CUDNN.CUDNN_LSTM : println("No CUDNN RNN available.")))
```
MikeInnes: This check was left out because e.g. users can't hit this path by accident. Don't object to adding it but shouldn't it be an error?

src/layers/recurrent.jl
```diff
@@ -98,31 +98,39 @@ RNN(a...; ka...) = Recur(RNNCell(a...; ka...))
...
 
-mutable struct LSTMCell{A,V}
+mutable struct LSTMCell{A,V,F,G}
   Wi::A
   Wh::A
   b::V
   h::V
   c::V
-end
+  sigmRepl::F
```
MikeInnes: Not wholly convinced by these names. "forget activation" or similar perhaps?

src/layers/recurrent.jl
```diff
@@ -98,31 +98,39 @@ RNN(a...; ka...) = Recur(RNNCell(a...; ka...))
...
-  cell = tanh.(gate(g, o, 3))
-  output = σ.(gate(g, o, 4))
-  c = forget .* c .+ input .* cell
-  h′ = output .* tanh.(c)
+  b, o   = m.b, size(h, 1)
+  g      = m.Wi*x .+ m.Wh*h .+ b
+  input  = m.sigmRepl.(gate(g, o, 1))
+  forget = m.sigmRepl.(gate(g, o, 2))
+  cell   = m.tanhRepl.(gate(g, o, 3))
+  output = m.sigmRepl.(gate(g, o, 4))
```
MikeInnes: It would be best to unpack these before the broadcast. Otherwise we try to send `m` to the GPU and I'd expect that to break (though it seems you have this tested?)

src/cuda/curnn.jl
```diff
@@ -1,17 +1,17 @@
-import ..Flux: Flux, relu
+import ..Flux: Flux, relu, sigmoid
 using CuArrays.CUDAnative
 using CuArrays: @cuindex, cudims
 
-CuRNN{T} = Flux.RNNCell{<:Union{typeof(tanh),typeof(relu)},<:CuArray{T,2},<:CuArray{T,1}}
-CuGRU{T} = Flux.GRUCell{<:CuArray{T,2},<:CuArray{T,1}}
-CuLSTM{T} = Flux.LSTMCell{<:CuArray{T,2},<:CuArray{T,1}}
+CuRNN{T}  = Flux.RNNCell{<:Union{typeof(tanh),typeof(relu)},<:CuArray{T,2},<:CuArray{T,1}}
+CuGRU{T}  = Flux.GRUCell{<:CuArray{T,2},<:CuArray{T,1}}#Flux.GRUCell{<:CuArray{T,2},<:CuArray{T,1},<:typeof(sigmoid),<:typeof(tanh)}
```
MikeInnes: Why is this commented out?

MikeInnes (commented):
Looking good, thanks! I think we need to bikeshed names a bit but otherwise this is good to go.

MikeInnes:
bors try

src/layers/recurrent.jl
MikeInnes: Perhaps another option is to have an "activations" tuple without trying to name them individually

src/cuda/curnn.jl
NTimmons: Whoops. Little mistake when merging after some tests. Thanks for catching that, this version also had some debug prints in it still. All corrected now.
The right side is correct, only select CuGRU if the correct functions are set.

src/layers/recurrent.jl
NTimmons: I don't think they map directly like that as we would need 3(GRU)/5(LSTM) function replacement choices.
'Activations' tuple might be better.

I guess it depends on if you want to replace each stage or just replace the concept of each function.

src/cuda/curnn.jl
NTimmons: Changed to error.

src/layers/recurrent.jl
NTimmons: Seems to pass the current GPU tests I have written? (Tested on NV 1080 desktop, NV 2060 laptop)

I am not 100% on how it is submitted for GPU though, is there any additional tests/tools I can use to check it is working as expected?

src/layers/recurrent.jl
MikeInnes: I think we can just do the simple replacement with a tuple for now. In future we could dispatch on tuple size to decide whether we're replacing both functions or each gate.

MikeInnes: I think we can just do the simple replacement with a tuple for now. In future we could dispatch on tuple size to decide whether we're replacing both functions or each gate.

src/cuda/curnn.jl
MikeInnes: It seems like you tested this case, so it's a little strange that the test passed despite this bug. Any idea why?

NTimmons: Checking now

NTimmons: Annoyingly, both seem to work. It shouldn't as the signature doesn't match. So I am digging to see what it is actually creating... then can add a test for that too.

MikeInnes: Great. Best to figure it out as otherwise it might mean the GPU side isn't being tested properly.

NTimmons: So,
If we have: `CuGRU{T}  = Flux.GRUCell{<:CuArray{T,2},<:CuArray{T,1}}`
This function is able to be called with custom activations, which should not be:
```
function (m::CuGRU{T})(h::CuArray{T}, x::CuArray{T}) where T <: Union{Float32,Float64}
  println("This is m::CuGRU")
  y, h′ = CUDNN.forward(desc(m), x, h)
  return h′, y
end
```
With: `Flux.GRUCell{<:CuArray{T,2},<:CuArray{T,1}},<:typeof(sigmoid),<:typeof(tanh)}`
That function is not called, instead as expected, it is using the definition in recurrent.jl:177 as it cannot create a cuGRU. Which I believe to be the correct behaviour?

However, I am not sure why it does not throw an error with the shortened definition when it is attempted to be instantiated. I did a little check to make sure I wasn't going mad:
'''
mutable struct Foo{A,B,C}
    a::A
    b::B
    c::C
end

inst = Foo(1,2.0,3.0f0)
'''
-> Foo{Int64,Float64,Float32}(1, 2.0, 3.0f0)
"""
Bar{T}  = Foo{Int64}
"""
-> Foo{Int64,B,C} where C where B
"""
Bar(1, 2.0, 3.0f0)
"""
MethodError: no method matching Foo{Int64,B,C} where C where B(::Int64, ::Float64, ::Float32)

And I would expect a similar error with the instantiation of a cuGRU with only two types defined as we have extended its signature to have four - unless it is being created in a way I am not familiar with?

src/layers/recurrent.jl
NTimmons: Done!

NTimmons:
bors try

NTimmons:
Damn
I think the normal checking is broken. Seems to be failing on downloading data...
Local testing all passes.

NTimmons:
This looks like a network error.
Can we rerun the test and merge this?

CarloLucibello:
@NTimmons . Then we can merge if tests pass

tbenst:
The docs currently say: [`LSTM(in::Integer, out::Integer, σ = tanh)`](https://fluxml.ai/Flux.jl/v0.5/models/layers.html#Flux.LSTM). But this fails:
```
julia> LSTM(1,1,σ=tanh)
ERROR: MethodError: no method matching Flux.LSTMCell(::Int64, ::Int64; σ=tanh)
```

I presume docs were updated prematurely before this was merged..?

NTimmons:
I don't believe this has been merged. I need to check it and retest it. Been busy on other projects and missed the update.  I will get on this ASAP

ToucheSir:
Those docs are also from v0.5. The [latest reference](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.LSTM) looks correct.

CarloLucibello:
closing as stale

PR #969, opened 2019-12-16
add lookahead optimiser

chengchingwen:
This add the implementation of Lookahead optimiser from [this paper](https://arxiv.org/abs/1907.08610) and based on the [pytorch impl](https://github.com/lonePatient/lookahead_pytorch/blob/master/optimizer.py)

### DIFF
--- a/src/optimise/Optimise.jl
+++ b/src/optimise/Optimise.jl
@@ -4,12 +4,13 @@ using LinearAlgebra
 import ArrayInterface
 
 export train!, update!,
-	Descent, ADAM, Momentum, Nesterov, RMSProp,
+	Descent, ADAM, Momentum, Nesterov, RMSProp, Lookahead,
 	ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW,RADAM, OADAM, AdaBelief,
 	InvDecay, ExpDecay, WeightDecay, stop, skip, Optimiser,
 	ClipValue, ClipNorm
 
 include("optimisers.jl")
+include("lookahead.jl")
 include("train.jl")
 
 end
--- a/src/optimise/lookahead.jl
+++ b/src/optimise/lookahead.jl
@@ -1,0 +1,117 @@
+"""
+    Lookahead(opt, α = 0.5, k = 6, momentum_handler = default_momentum_handler!)
+
+Implements the Lookahead optimiser.
+
+## Parameters
+  - Inner Optimiser
+  - alpha (`\alpha::Float64`): The slow weights (outer loop) step size. Defaults to `0.5`.
+  - k : The inner-loop step. The outer loop will update per k inner-loop step.
+  - momentum_handler: function for handling the inner optimiser momentums. Default do nothing.
+
+## Examples
+```julia
+inner_opt = ADAM() # create the inner optimiser
+
+opt = Lookahead(inner_opt, α = 0.7, k = 10)
+```
+
+For using with custom Optimiser, overload the corresponding methods. see also [`has_momentum`](@ref), [`reset_momentum_handler!`], [`pullback_momentum_handler!`](@ref).
+
+## References
+[Lookahead](https://arxiv.org/pdf/1907.08610.pdf) optimiser.
+"""
+mutable struct Lookahead{O, MH <: Function}
+  opt::O
+  momentum_handler::MH
+  alpha::Float64
+  k::Int
+  state::IdDict
+end
+
+const MomentumOptim = Union{Momentum, RMSProp, Nesterov, ADAM, RADAM, AdaMax, ADAGrad, ADADelta, AMSGrad, NADAM}
+
+"""
+  has_momentum(o)
+
+return `true` if the optimiser has momentum.
+"""
+has_momentum(o) = false
+has_momentum(o::MomentumOptim) = true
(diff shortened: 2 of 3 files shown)


src/optimise/Optimise.jl
```diff
@@ -2,10 +2,12 @@ module Optimise
 
 export train!,
 	SGD, Descent, ADAM, Momentum, Nesterov, RMSProp,
-	ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW,RADAM, 
-	InvDecay, ExpDecay, WeightDecay, stop, Optimiser
+	ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW,
+  RADAM, Lookahead, InvDecay, ExpDecay, WeightDecay,
```
bhvieira: The indentation here (and the next line) moved. Just letting you know.

chengchingwen: There miss a space between `ADAMW` and `RADAM` but with the space that line looks too long, so I moved it to the next line.

bhvieira: Sure, but the indentation is different now. You could make every line use a double-space as well, or move back to tab everywhere, I'm not sure Flux follow any single style.

chengchingwen: what do you mean? all lines are two spaces indent.

bhvieira: Weird, my browser renders it as if the line starting with RADAM was a double-space while the other one above was a tab.

chengchingwen: Ah, I see. I set all tab as double-space width.

src/optimise/lookahead.jl
```diff
@@ -0,0 +1,117 @@
...
+"""
+mutable struct Lookahead{O, MH <: Function}
+  opt::O
+  momentum_handler::MH
+  alpha::Float64
+  k::Int
+  state::IdDict
+end
+
+const MomentumOptim = Union{Momentum, RMSProp, Nesterov, ADAM, RADAM, AdaMax, ADAGrad, ADADelta, AMSGrad, NADAM}
```
dhairyagandhi96: Perhaps an `AbstractOptimiser` type would help with some of the repitition.

chengchingwen: But some optimiser don't have momentum term. Maybe we also need a `AbstractMomentumOptimiser`?

dhairyagandhi96: Thinking if we could use traits here?

chengchingwen: what's that?

chengchingwen:
```julia
abstract type OptimiserStyle end
struct Momentumlike <: OptimiserStyle end
struct NonMomentumlike <: OptimiserStyle end

has_momentum(o::Type{Descent}) = NonMomentumlike()
has_momentum(o::Type{Momentum}) = Momentumlike()
has_momentum(o::Type{Adam}) = Momentumlike()
```
like this?

dhairyagandhi96: Something like that, yeah.

dhairyagandhi96: Also, ref https://github.com/oxinabox/oxinabox.github.io/blob/master/notebook_posts/Dispatch%2C%20Traits%20and%20Metaprogramming%20Over%20Reflection.ipynb

chengchingwen:
@dhairyagandhi96  I define a `AbstractOptimiser`and the `GradientStyle`traits`.

CarloLucibello:
I would step back to the original problem and just add the Lookahead optimizer

chengchingwen:
@CarloLucibello  Sure! should I revert the commit to 510cfaa or 14956c2 ?

CarloLucibello:
yes please. If we want to introduce a type hierarchy for the optimizers let's first discuss it in a separate issue/PR

chengchingwen:
@CarloLucibello Reverted.

CarloLucibello:
In the paper, I read 
"""
We have the choice of maintaining, interpolating, or resetting the internal state (e.g. momentum) of
the inner optimizer. We evaluate this tradeoff on the CIFAR dataset (where every choice improves
convergence) in Appendix D.1 and maintain internal state for the other experiments.
....
D.1 Inner Optimizer State
Throughout our paper, we maintain the state of our inner optimizer for simplicity. For SGD with
heavy-ball momentum, this corresponds to preserving the momentum. Here, we present a sensitivity
study by comparing the convergence of Lookahead when maintaining the momentum, interpolating
the momentum, and resetting the momentum. All three improve convergence versus SGD.
"""
So let's keep this simple and maintain the state.

This also needs a rebase

CarloLucibello:
> So let's keep this simple and just maintain the state.

I'm saying this because the specific needs of this PR, we probably need a state setter/getter API for the optimizers, so we should think of it more carefully in a separate issue/PR, then we can revisit LookAhead when we have that in place.
So I think it's better to just merge the minimal LookAhead implementation here, with no state reset

chengchingwen:
@DhairyaLGandhi rebased and updated

CarloLucibello:
sorry we droppebd the ball here. Closing since optimisers are now in Optimisers.jl

PR #973, opened 2019-12-23
Give `NNPACK` a bit of numerical leeway

staticfloat:
We were very strict on NNlib's ability to fudge numbers a bit; this loosens some of the requirements, since `NNPACK` is definitely running in the DNN-world equivalent of `@fastmath`.


dhairyagandhi96:
Thanks! @staticfloat

Issue #887, opened 2019-10-08
normalise is not GPU compatible


MikeInnes:
```julia
julia> using Flux, CuArrays

julia> CuArrays.allowscalar(false)

julia> Flux.normalise(gpu(rand(10)))
ERROR: scalar getindex is disallowed
```

This is actually due to `std` with a dims argument:

```julia
julia> std(gpu(rand(10)), dims = 1)
ERROR: scalar getindex is disallowed
```

an-awesome-guy:
`julia> using Flux,CuArrays

julia> CuArrays.allowscalar(false)

julia> Flux.normalise(gpu(rand(10)))
10-element CuArray{Float32,1,Nothing}:
 -1.806264   
  0.78976214 
  1.5305954  
  0.86513764 
 -0.45081395 
  1.1337769  
 -1.0279455  
 -0.47165877 
 -0.5008726  
 -0.061718248

julia> Flux.normalise(gpu(reshape(collect(1:9), 3, 3)),dims=2)
3×3 CuArray{Float32,2,Nothing}:
 -1.22474  0.0  1.22474
 -1.22474  0.0  1.22474
 -1.22474  0.0  1.22474`
@MikeInnes I wrote a seperate normalise function for CuArray. I think this might solve this bug. May I send a pull request?

an-awesome-guy:
```
julia> using Flux,CuArrays

julia> CuArrays.allowscalar(false)

julia> Flux.normalise(gpu(rand(10)))
10-element CuArray{Float32,1,Nothing}:
 -0.4204     
 -1.0460519  
  1.3730303  
  0.9785272  
 -1.2841846  
  1.0471098  
  1.0527792  
 -1.3440493  
 -0.46324056 
  0.106479086

julia> test = gpu(reshape(collect(1:9), 3, 3))
3×3 CuArray{Float32,2,Nothing}:
 1.0  4.0  7.0
 2.0  5.0  8.0
 3.0  6.0  9.0

julia> Flux.normalise(test,dims=1)
3×3 CuArray{Float32,2,Nothing}:
 -1.22474  -1.22474  -1.22474
  0.0       0.0       0.0    
  1.22474   1.22474   1.22474

julia> Flux.normalise(test,dims=2)
3×3 CuArray{Float32,2,Nothing}:
 -1.22474  0.0  1.22474
 -1.22474  0.0  1.22474
 -1.22474  0.0  1.22474
```
[...]

PR #981, opened 2020-01-01
add normalise function for CuArrays

an-awesome-guy:


an-awesome-guy:
This solves #887

CarloLucibello:
Hi, 
#887  is better fixed by adding `std` and `var` specializations to CuArrays.jl, see https://github.com/JuliaGPU/CuArrays.jl/issues/411.
Also, it needs to be extended to support multi-dimensional reductions (e.g. `dims=(1,2)`) and you should add some tests.

MikeInnes:
Yes, this should definitely live in CuArrays rather than here. Best to open a patch on CuArrays and get feedback over there. Thanks for the effort though!

PR #983, opened 2020-01-03
Activation functions for GRU layer

sdobber:
Some papers (for example Lai et al. "Modeling Long- and Short-Term Temporal Patterns with Deep Neural Networks", arXiv:1703.07015v3) use activation functions different than `sigmoid` and `tanh` for gated recurrent units. Such activation functions can now be included in the call to `GRU(...)`. The default behaviour reverts to the current implementation.

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -146,23 +146,26 @@ LSTM(a...; ka...) = Recur(LSTMCell(a...; ka...))
 
 # GRU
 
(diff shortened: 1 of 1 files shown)


sdobber:
I just saw that this is more or less the same as [pull request 964](https://github.com/FluxML/Flux.jl/pull/964) - sorry for the unnecessary PR.

MikeInnes:
Thanks for the effort!

PR #984, opened 2020-01-05
Adding CompatHelper

aminya:


ViralBShah:
@dhairyagandhi96 Can you still review this?

ChrisRackauckas:
The Flux progress bar should probably update to the platform-independent setup instead of using `Julia.@progress` (which is where the compat comes from).

PR #985, opened 2020-01-06
Minor doc improve

aminya:
**Ready**
Doesn't overlap with https://github.com/FluxML/Flux.jl/pull/853

### DIFF
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -2,7 +2,7 @@
 
 ## Taking Gradients
 
-Flux's core feature is taking gradients of Julia code. The `gradient` function takes another Julia function `f` and a set of arguments, and returns the gradient with respect to each argument. (It's a good idea to try pasting these examples in the Julia terminal.)
+Flux's core feature is taking gradients of Julia code. The [`gradient`](@ref) function takes another Julia function `f` and a set of arguments, and returns the gradient with respect to each argument. (It's a good idea to try pasting these examples in the Julia terminal.)
 
 ```jldoctest basics
 julia> using Flux
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -41,6 +41,8 @@ julia> x = [2, 1];
 
 julia> y = [2, 0];
 
+julia> f(x, y) = sum((x .- y).^2);
+
 julia> gs = gradient(params(x, y)) do
          f(x, y)
        end
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -25,12 +25,12 @@ When a function has many parameters, we can get gradients of each one at the sam
 ```jldoctest basics
 julia> f(x, y) = sum((x .- y).^2);
 ```
-
-Now we calculate gradient of `f` for `x=2, y=2` and `x=1, y=0`:
+For this `f`, we can get the gradient with respect to both `x` and `y` in a single call:
 ```jldoctest basics
 julia> gradient(f, [2, 1], [2, 0])  # same as: df(x,y) = gradient(f,x,y); df([2,1], [2,0])
 ([0, 2], [0, -2])
 ```
+where `x = [2, 1]` and `y = [2, 0]`
 
 But machine learning models can have *hundreds* of parameters! To handle this, Flux lets you work with collections of parameters, via `params`. You can get the gradient of all parameters used in a program without explicitly passing them in.
 
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -35,13 +35,9 @@ where `x = [2, 1]` and `y = [2, 0]`
 But machine learning models can have *hundreds* of parameters! To handle this, Flux lets you work with collections of parameters, via `params`. You can get the gradient of all parameters used in a program without explicitly passing them in.
 
 ```jldoctest basics
-julia> using Flux
-
 julia> x = [2, 1];
 
 julia> y = [2, 0];
-
-julia> f(x, y) = sum((x .- y).^2);
 
 julia> gs = gradient(params(x, y)) do
          f(x, y)
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -20,7 +20,7 @@ julia> d2f(2)
 6
 ```
 
-When a function has many parameters, we can get gradients of each one at the same time by passing vectors of corresponding inputs for each argument to `gradient`:
+When a function has many parameters, we can get gradients of each one at the same time by passing each one as an argument:
 
 ```jldoctest basics
 julia> f(x, y) = sum((x .- y).^2);
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -27,10 +27,10 @@ julia> f(x, y) = sum((x .- y).^2);
 ```
 For this `f`, we can get the gradient with respect to both `x` and `y` in a single call:
 ```jldoctest basics
-julia> gradient(f, [2, 1], [2, 0])  # same as: df(x,y) = gradient(f,x,y); df([2,1], [2,0])
+julia> gradient(f, [2, 1], [2, 0])
 ([0, 2], [0, -2])
 ```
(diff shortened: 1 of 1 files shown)


docs/src/models/basics.md
```diff
@@ -2,42 +2,49 @@
...
+julia> import Flux: gradient
+
 julia> f(x) = 3x^2 + 2x + 1;
 
-julia> df(x) = gradient(f, x)[1]; # df/dx = 6x + 2
+julia> df(x) = gradient(f, x; nest = true)[1]; # df/dx = 6x + 2
 
 julia> df(2)
-14
+14.0 (tracked)
```
CarloLucibello: variables are not tracked since the move from Tracker to Zygote for AD

aminya: I'm on Flux#master. Running the code gives this. Apparently, Zygote isn't used by default.

aminya: I added `using Zygote` and the issue is gone now.

aminya: @CarloLucibello my Flux had been downgraded automatically because old compat bounds. Adding #master again fixes the problem

docs/src/models/basics.md
```diff
@@ -20,12 +20,15 @@ julia> d2f(2)
 6
 ```
 
-When a function has many parameters, we can get gradients of each one at the same time:
+When a function has many parameters, we can get gradients of each one at the same time by passing vectors of corresponding inputs for each argument to `gradient`:
```
MikeInnes:
```suggestion
When a function has many parameters, we can get gradients of each one at the same time by passing each one as an argument:
```

docs/src/models/basics.md
```diff
@@ -38,6 +41,8 @@ julia> x = [2, 1];
 
 julia> y = [2, 0];
 
+julia> f(x, y) = sum((x .- y).^2);
```
MikeInnes: Is it necessary to repeat this definition here, after it's defined right above?

docs/src/models/basics.md
```diff
@@ -20,12 +20,15 @@ julia> d2f(2)
...
 
-When a function has many parameters, we can get gradients of each one at the same time:
+When a function has many parameters, we can get gradients of each one at the same time by passing vectors of corresponding inputs for each argument to `gradient`:
 
 ```jldoctest basics
 julia> f(x, y) = sum((x .- y).^2);
+```
 
-julia> gradient(f, [2, 1], [2, 0])
+Now we calculate gradient of `f` for `x=2, y=2` and `x=1, y=0`:
```
MikeInnes: This is not what the example is intended to show; the point is that you can get `x` and `y` at the same time, not that you can get gradients at multiple points at the same time (which is true only in very specific cases).

docs/src/models/basics.md
aminya: No, because the example starts with `using Flux` again, it implies that it is a new example

docs/src/models/basics.md
aminya: Could you edit my sentence to show what you are trying to say, I still don't understand this part of doc. It is not clear.

MikeInnes: Maybe it would be best to skip these parts then and just do the definite improvements; we can look at this later.

scimas: The `f` in that example is simply the square of the [Euclidean distance](https://en.wikipedia.org/wiki/Euclidean_distance) between two vectors `x ≡ (x1, x2, ..., xN)` and `y ≡ (y1, y2, ..., yN)`. So the arguments being passed are the `x ≡ (2, 1)` and `y ≡ (2, 0)` points.

So I would word it this way:
```
For this `f`, we can get the gradient with respect to both `x` and `y` in a single call:
<The gradient example here>
where `x = [2, 1]` and `y = [2, 0]`
```
Sorry about butting into the conversation. I'm trying out Julia / Flux and found a bunch of minor mistakes / inconsistencies in the documentation that I would like to see fixed too.

MikeInnes: A wording like this sounds good to me. Perhaps if @aminya wants to revert the current version you could put up a separate PR. It's also possible to PR against this PR but that's a little complicated.

MikeInnes: A wording like this sounds good to me. Perhaps if @aminya wants to revert the current version you could put up a separate PR. It's also possible to PR against this PR but that's a little complicated.

aminya: I can commit it to this PR. no need for a separate PR.

aminya:
Could you merge this? Doc improvement PRs should be in bit size, so they get merged fast.
Notice how https://github.com/FluxML/Flux.jl/pull/853 is waiting to get merged for several months now!

I will create more PRs, as I read more

docs/src/models/basics.md
MikeInnes: Let's remove the `using Flux` instead, in that case.

janEbert:
> Could you merge this? Doc improvement PRs should be in bit size, so they get merged fast.
> Notice how #853 is waiting to get merged for several months now!
> 
> I will create more PRs, as I read more

You are right. It's still possible to split that PR up, though!
I just thought it would be nice to have a global overhaul.

aminya:
What is holding this back from getting merged? 😞

docs/src/models/basics.md
```diff
@@ -20,20 +20,21 @@ julia> d2f(2)
 6
 ```
 
-When a function has many parameters, we can get gradients of each one at the same time:
+When a function has many parameters, we can get gradients of each one at the same time by passing each one as an argument:
```
CarloLucibello, line 23: I would keep only this change and leave the rest as it is since it seems redundant

aminya: What changes are redundant? Do you mean the examples!?

CarloLucibello: all changes except the line above, in my opinion. I prefer the previous version

aminya: I am describing the syntax. I don't dee this as redundant.

ToucheSir, line 23:
```suggestion
When a function has many parameters, we can get the gradient with respect to each in a single call:
```

docs/src/models/basics.md
```diff
@@ -20,20 +20,21 @@ julia> d2f(2)
...
 
 ```jldoctest basics
 julia> f(x, y) = sum((x .- y).^2);
-
+```
+For this `f`, we can get the gradient with respect to both `x` and `y` in a single call:
```
ToucheSir, line 28:
```suggestion
```
If the "with respect to" is moved up top, this line is no longer necessary.

docs/src/models/basics.md
```diff
@@ -20,20 +20,21 @@ julia> d2f(2)
...
+```jldoctest basics
 julia> gradient(f, [2, 1], [2, 0])
 ([0, 2], [0, -2])
 ```
+where `x = [2, 1]` and `y = [2, 0]`. The above syntax is the same as: `df(x,y) = gradient(f,x,y); df([2,1], [2,0])`
```
ToucheSir, line 33:
```suggestion
The above syntax is the same as: `df(x,y) = gradient(f,x,y); df([2,1], [2,0])`
```

ToucheSir (changes requested):
Apologies for letting this fall through the cracks. I've added some suggestions to deduplicate the wording and slim down the copy a bit. @CarloLucibello do you mind taking another look?

docs/src/models/basics.md
ToucheSir: Having a statement like "where ..." is a little cumbersome. Since `df(x, y)` already names the parameters, there's no need to define them in prose. Alternatively, you could actually define them in the doctest with `x, y = ...`, but that might be confusing given `f(x, y)` has already been defined earlier (users who aren't comfortable with scoping may conflate the two).

aminya:
Feel free to update the branch.

CarloLucibello:
closing as stale

PR #986, opened 2020-01-06
Destructure/restructure for models

MikeInnes:
Provides the ability to "flatten" a model into a single weight vector and back again, as in #747. Useful for hypernetworks as in #797 and #799. This is essentially the DiffEqFlux code, but tweaked to the interface described in #797 and made differentiable with Zygote.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,8 +1,10 @@
+# This file is machine-generated - editing it directly is not advised
+
 [[AbstractFFTs]]
 deps = ["LinearAlgebra"]
-git-tree-sha1 = "380e36c66edfa099cd90116b24c1ce8cafccac40"
+git-tree-sha1 = "051c95d6836228d120f5f4b984dd5aba1624f716"
 uuid = "621f4979-c628-5d54-868e-fcf4e3e8185c"
-version = "0.4.1"
+version = "0.5.0"
 
 [[AbstractTrees]]
 deps = ["Markdown", "Test"]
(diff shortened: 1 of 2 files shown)


MikeInnes:
@dhairyagandhi96 do we know what's happening with bors?

dhairyagandhi96:
Likely some of the keys are need updating. Will re-add it to the GitLab CI

MikeInnes:
Great. Would be great if you could merge this once that's in place.

dhairyagandhi96:
bors ping

dhairyagandhi96:
bors ping

dhairyagandhi96:
bors try

maleadt:
The problem is that GitLab for some reason doesn't manage to push the build status to GitHub, and as a result Bors "fails". You can see the build status by inspecting the branch on the GitLab repo page, e.g., https://gitlab.com/JuliaGPU/Flux.jl/pipelines/108796083

dhairyagandhi96:
Exactly right, I figured that must have been the case when I saw no GitLab status reporting on the `trying` branch which bors monitors.

dhairyagandhi96:
Also, will merge this so it can be used by folks easier, but the build failures on 1.3- would need to be addressed

jessebett:
@ChrisRackauckas has said that their `Fast___` layers (e.g. `FastDense`) which are written for explicit parameterization are considerably faster than using `destructure restructure` on an implicit layer. Do we have an idea for why that is?

`restructure` on a `Dense` layer should perform identically to a `FastDense` that was written to be explicit.

ChrisRackauckas:
For reference: https://github.com/JuliaDiffEq/DiffEqFlux.jl/pull/126 . I couldn't figure out how to make restructure/destructure faster, and this is easier to use anyways, so 🤷‍♂

MikeInnes:
Yes, it would be much easier to (a) make `restructure` faster and (b) have better support for structs in more places so that `restructure` is unnecessary in the first place.

Note that those DiffEqFlux layers are not "more explicit" than the Flux ones. The only difference is whether you store weights in a single array or allow data structures. In general, I think we'd like to generalise more of the ecosystem (e.g. optimisers) to support structures, which would avoid this kind of issue.

Issue #965, opened 2019-12-13
Unable to get gradients of "Dense" models when sparse arrays are involved


SimonMandlik:
```julia
using Flux, SparseArrays

md = Dense(2,2)
ms = Dense(sparse(randn(2,2)), sparse(randn(2)))
xd = randn(2, 2)
xs = sparse(randn(2, 2))

gradient(() -> sum(md(xd)), Flux.params(md))
gradient(() -> sum(ms(xs)), Flux.params(ms))
gradient(() -> sum(ms(xd)), Flux.params(ms))
gradient(() -> sum(md(xs)), Flux.params(md))
```

The last call fails (on 0.10.0):

```julia
ERROR: MethodError: no method matching zero(::Type{Any})
Closest candidates are:
  zero(::Type{Union{Missing, T}}) where T at missing.jl:105
  zero(::Type{Missing}) at missing.jl:103
  zero(::Type{LibGit2.GitHash}) at /Users/sabae/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/LibGit2/src/oid.jl:220
  ...
Stacktrace:
 [1] zero(::Type{Any}) at ./missing.jl:105
 [2] _zeros_eltypes at /Users/sabae/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/SparseArrays/src/higherorderfns.jl:203 [inlined]
 [3] _noshapecheck_map(::Zygote.var"#1457#1464", ::SparseMatrixCSC{Any,Int64}) at /Users/sabae/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/SparseArrays/src/highero
rderfns.jl:159
 [4] map(::Zygote.var"#1457#1464", ::SparseMatrixCSC{Any,Int64}) at /Users/sabae/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/SparseArrays/src/higherorderfns.jl:143
 [5] adjoint at /Users/simon.mandlik/.julia/packages/Zygote/8dVxG/src/lib/broadcast.jl:126 [inlined]
 [6] _pullback at /Users/simon.mandlik/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:47 [inlined]
 [7] adjoint at /Users/simon.mandlik/.julia/packages/Zygote/8dVxG/src/lib/lib.jl:139 [inlined]
[...]

PR #987, opened 2020-01-06
added support for SparseArray input in Dense layer

mungu42:
Fixes Issue #965 
Dense Layer with is able to take SparseArray as input


MikeInnes:
Can you describe why this is needed a bit more?

It seems like it's the same definition that's already covered by `(a::Dense)(x::AbstractArray)`, since sparse arrays are abstract arrays (and the method body hasn't changed).

mungu42:
For some reason gradients of Dense layer with an Arrays its weight and a SparseMatrix as input fails

`using Flux, SparseArrays

md = Dense(2,2)
ms = Dense(sparse(randn(2,2)), sparse(randn(2)))
xd = randn(2, 2)
xs = sparse(randn(2, 2))

gradient(() -> sum(md(xd)), Flux.params(md))
gradient(() -> sum(ms(xs)), Flux.params(ms))
gradient(() -> sum(ms(xd)), Flux.params(ms))
gradient(() -> sum(md(xs)), Flux.params(md))`

mungu42:
For some reason gradients of Dense layer with an Arrays as it's weight and a SparseMatrix as input fails.
```

using Flux, SparseArrays

md = Dense(2,2)
ms = Dense(sparse(randn(2,2)), sparse(randn(2)))
xd = randn(2, 2)
xs = sparse(randn(2, 2))

gradient(() -> sum(md(xd)), Flux.params(md)) #Works
gradient(() -> sum(ms(xs)), Flux.params(ms)) #Works
gradient(() -> sum(ms(xd)), Flux.params(ms)) #Works
gradient(() -> sum(md(xs)), Flux.params(md)) #Fails
```

I tried using  `sum(md.σ.(md.W * xs .+ md.b)) `in its place which works. Maybe something to do with BLAS calls and SparseArrays in the backward pass?

MikeInnes:
Right you are, but this error actually comes from trying to convert the sparse array to Float32 to match the layer:

```julia
julia> gradient(x -> sum(Float32.(x)), sparse(randn(2)))
ERROR: MethodError: no method matching zero(::Type{Any})
```

Your patch avoids the conversion, but really we should just fix this error in Zygote.

mungu42:
So what can be defined for `zero(::Type{Any})` As I understand it the zero function is missing for the Type Any?

MikeInnes:
Well, at some point we're calling `zero(Any)`. That can't be right, so it needs digging into to find out why that's happening.

racinmat:
The main problem is caused by mixing Float32 and Float64, which results in `zero(Any)`. 
When we have consistent types, it's working.
```julia
W = randn(Float32, 2,2)
b = randn(Float32, 2)
md = Dense(W, b)
xs = sparse(randn(Float32, 2, 2))
gradient(() -> sum(md(xs)), Flux.params(md))

W = randn(Float64, 2,2)
b = randn(Float64, 2)
md = Dense(W, b)
xs = sparse(randn(Float64, 2, 2))
gradient(() -> sum(md(xs)), Flux.params(md))
```
are working.

But
```
W = randn(Float32, 2,2)
b = randn(Float32, 2)
md = Dense(W, b)
xs = sparse(randn(Float64, 2, 2))
gradient(() -> sum(md(xs)), Flux.params(md))

W = randn(Float64, 2,2)
b = randn(Float64, 2)
md = Dense(W, b)
xs = sparse(randn(Float32, 2, 2))
gradient(() -> sum(md(xs)), Flux.params(md))
```
throw the above-mentioned error.
The main question is: how should we approach it?
Since Flux uses `Float32` by default, but Julia itself is using usually `Float64` by default, it's very easy to run into this inconsistency.

racinmat:
For Zygote itself, it's working, see:
```
W = randn(Float32, 2,2)
b = randn(Float32, 2)
xs = sparse(randn(Float32, 2, 2))
gradient(() -> sum(W*xs .+ b), Params([W,b]))

W = randn(Float32, 2,2)
b = randn(Float32, 2)
md = Dense(W, b)
xs = sparse(randn(Float64, 2, 2))
gradient(() -> sum(W*xs .+ b), Params([W,b]))

W = randn(Float64, 2,2)
b = randn(Float64, 2)
md = Dense(W, b)
xs = sparse(randn(Float32, 2, 2))
gradient(() -> sum(W*xs .+ b), Params([W,b]))

W = randn(Float64, 2,2)
b = randn(Float64, 2)
md = Dense(W, b)
xs = sparse(randn(Float64, 2, 2))
gradient(() -> sum(W*xs .+ b), Params([W,b]))
```

all of them are passing, so I don't think this is problem with Zygote.

racinmat:
Hm, it seems the conversion is problematic. Dense layer explicitly casts input data to type of parameter in https://github.com/FluxML/Flux.jl/blob/master/src/layers/basic.jl#L138.

And Zygote can't pullback through it, following Zygote code
```

W = randn(Float32, 2,2)
b = randn(Float32, 2)
md = Dense(W, b)
xs = sparse(randn(Float64, 2, 2))
gradient(() -> sum(W*Float32.(xs) .+ b), Params([W,b]))

W = randn(Float64, 2,2)
b = randn(Float64, 2)
md = Dense(W, b)
xs = sparse(randn(Float32, 2, 2))
gradient(() -> sum(W*Float64.(xs) .+ b), Params([W,b]))
```
crashes. The question is: should this be fixed in Zygote or in Flux?

mungu42:
I guess if mixed-precision compute is supported this shouldn't be an issue.

racinmat:
I made issue on Zygote with MWE, we'll see https://github.com/FluxML/Zygote.jl/issues/810

ToucheSir:
Since this turned out to be about eltype mismatches instead of sparse array support, I think it can be safely closed.

Issue #942, opened 2019-11-22
Feature request: Weight normalization


ararslan:
[Weight normalization](https://arxiv.org/abs/1602.07868) is a reparameterization of the weights in terms of two quantities which are optimized instead of the weights directly. The authors provide [a reference implementation](https://github.com/TimSalimans/weight_norm) in Python using Theano, and the functionality is available in PyTorch as a hook.

I've toyed just a tiny bit with implementing this, but I think my knowledge of Flux's internals is insufficient. I was looking at implementing it as a layer that wraps another layer, for which the `params` are the *g* and **v** parameterization of the weights of the wrapped layer. Perhaps there's a better way.

CarloLucibello:
another attempt here https://github.com/FluxML/Flux.jl/pull/1005

PR #993, opened 2020-01-16
Added Weight Normalization

an-awesome-guy:


an-awesome-guy:
Added feature for #942 @MikeInnes @dhairyagandhi96

CarloLucibello:
It would nice to have a utility function to apply this to any module as pytorch does:
https://github.com/pytorch/pytorch/blob/master/torch/nn/utils/weight_norm.py
https://pytorch.org/docs/stable/nn.html#weight-norm

bhvieira:
> It would nice to have a utility function to apply this to any module as pytorch does:
> https://github.com/pytorch/pytorch/blob/master/torch/nn/utils/weight_norm.py
> https://pytorch.org/docs/stable/nn.html#weight-norm

I've been thinking of ways of doing that.
I think we could define an enclosure (like Chain) whose parameters are defined in the Weight Normalization fashion in regards to the parameters of its constituents.

Or we could do something with custom adjoints too, that's done explicitly in the paper.

bhvieira:
I thought of a better way that mimics PyTorch, will write a PR soon.

bhvieira:
BTW in the paper the norm should be taken by output dimension, so g is a vector and v is a matrix.
Your current implementation equals to PyTorch with dim=None.

CarloLucibello:
we need something more general, along the lines of #1005. Closing this now, feel free to open a new PR in case you want to tackle the problem again

PR #994, opened 2020-01-16
Add custom training loops to docs

oxinabox:


bhvieira:
Oh god yes, we needed this. Especially for people trying wacky stuff with recurrence, `train!` becomes cumbersome fast.

oxinabox:
Bump

dhairyagandhi96:
Should we add in a mention of it conceptually being a eval, backprop, update loop? 
Otherwise lgtm! Thanks so much!

oxinabox:
> Should we add in a mention of it conceptually being a eval, backprop, update loop?

I'm not sure what that would add.
It is what it is.

At some point we will do #666. 
and then there will be two examples in this section.
The current one, which will be marked as an advanced training loop " for if you need access to the gradient, e.g for gradient clipping or logging purposes.

And a new one that never explictly calculates the gradient, and just does
```
function my_custom_train!(loss, ps, data, opt)
  ps = Params(ps)
  for d in data
    step!(opt, ps) do
      training_loss = loss(d...)
      # Insert what ever code you want here that needs Training loss, e.g. logging
      return training_loss
    end

    # Here you might like to check validation set accuracy, and break out to do early stopping
  end
end
```

oxinabox:
Merging now, additional enhancements can come in n follow up PRs

dhairyagandhi96:
Thanks Lyndon!

PR #996, opened 2020-01-18
Added the DenseNet-type connectivity pattern

bhvieira:
With the DenselyConnected constructor now it is possible to define the DenseNet-type connectivity pattern.

Comments welcome!


bhvieira:
We could do away with the recursion (both here and in Chain as well), but since these recursions aren't likely to be that deep and I'm not sure how much the compiler would benefit from a loop implementation, for now it's a recursion, which I also think is easier to read anyways.

bhvieira:
Can this be merged? I think this is pretty non-controversial and nice to have for CNNs.
ping @MikeInnes  @dhairyagandhi96

src/layers/basic.jl
```diff
@@ -226,3 +226,48 @@ end
...
+```
+layers = (Dense(1,1), Dense(2,1), Dense(3,1))
+connection = (x,y) -> vcat(x, y)
+db = DenselyConnected(layers, connection)
+x = zeros(Float32,1,2)
+size(db(x)) == (4,2)
+```
+"""
+
+
```
CarloLucibello: remove whitelines so that the docstring is correctly captured

src/layers/basic.jl
```diff
@@ -226,3 +226,48 @@ end
...
+It is up to the user to select the appropriate sizes for the layers.
+
+A simple example:
+```
+layers = (Dense(1,1), Dense(2,1), Dense(3,1))
+connection = (x,y) -> vcat(x, y)
+db = DenselyConnected(layers, connection)
+x = zeros(Float32,1,2)
+size(db(x)) == (4,2)
+```
```
CarloLucibello: I think it should be indent and no thicks for julia code

src/layers/basic.jl
```diff
@@ -226,3 +226,48 @@ end
 function Base.show(io::IO, b::SkipConnection)
   print(io, "SkipConnection(", b.layers, ", ", b.connection, ")")
 end
+
+"""
+    DenselyConnected(layers, connection)
+
+Creates a dense-connection of a Tuple of consecutive layers plus a
+connectivity pattern. The connection is an arbitrary callable, often a function.
+This means that the input of each layer is given by the output of all previous layers.
```
CarloLucibello: It is not clear from this definition how `connection` would be used to produce the input to each layer

bhvieira: I agree, I now improved the description, is it better now?

src/layers/basic.jl
bhvieira: I think I fixed this one, please check it later :)

src/layers/basic.jl
bhvieira: Done, thanks for pointing out!

src/layers/basic.jl
```diff
@@ -253,3 +253,51 @@ end
...
+
+"""
+    DenselyConnected(layers, connection)
+
+Creates a dense-connection of a Tuple of consecutive layers plus a
+connectivity pattern. In this structure layers are applied sequentially
+to the combined output of all previous layers in the structure.
+The connection is an arbitrary callable, often a function,
+and is used to combine all outputs from previous layers before applying the next
+layer.
```
CarloLucibello: It should be made explicit that  `connection` has to be callable with 2 arguments, and how a two-argument function is supposed to create the input to an intermediate layer by combining all previous inputs

src/layers/basic.jl
CarloLucibello: mhmhm, still seeing backticks

bhvieira: Where? In the example? All other docstrings have it

MikeInnes: Either backticks or indent is ok, but don't use both since that'll give the rendered code a big margin.

MikeInnes:
I haven't looked at this in great detail, but as a heads up, the bar for adding new layers is going to be pretty high. It would help a lot if there's any precedent we can point to in other frameworks. Right now it's not totally clear to me what this does from its name or description; that should be really obvious. It might be helpful to have some design discussion showing how code looks with and without the layer (e.g. the simplest possible unrolled version, something that uses `reduce` or a loop or whatever is appropriate, etc.) to show what the code saving looks like.

bhvieira:
Thinking about it better now I think the functionality is easily provided by the existing `SkipConnection` in the case of concatenating the features between layers, they are identical if you call `SkipConnection` on all the layers of your Chain, and I predict this would be the biggest use of this layer (in the case of taking averages it would be just a little more involved, but doable), so there's no need for this PR.

Issue #930, opened 2019-11-11
Error with regularization using norm() and Zygote


appleparan:
When trying to use L2 regularization with Zygote + Flux#master, Flux shows `mutating arrays is not supported` error. Here is a code snippet to reproduce and Stacktrace. It works if I just use `Flux.mse` only to loss function.

Moreover, even if I redefine loss function (`loss`) , It seems Zygote doesn't understand that loss function has changed (showing same error). I should restart Julia to work. Is it intended result?
```
using Flux, LinearAlgebra

m = Dense(2, 5)
data = [(rand(2, 3), rand(5, 3)) for _ in 1:10]
loss(x, y) = Flux.mse(m(x), y) + sum(LinearAlgebra.norm, Flux.params(m))
opt = Flux.ADAM()
Flux.train!(loss, Flux.params(m), data, opt)
```
```
julia> Flux.train!(loss, Flux.params(m), data, opt)
ERROR: Mutating arrays is not supported
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] (::getfield(Zygote, Symbol("##986#987")))(::Nothing) at /home/appleparan/.julia/packages/Zygote/8dVxG/src/lib/array.jl:42
 [3] (::getfield(Zygote, Symbol("##2607#back#988")){getfield(Zygote, Symbol("##986#987"))})(::Nothing) at /home/appleparan/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:49
 [4] push! at /home/appleparan/.julia/packages/Zygote/8dVxG/src/compiler/interface.jl:65 [inlined]
 [5] (::typeof(∂(push!)))(::Nothing) at /home/appleparan/.julia/packages/Zygote/8dVxG/src/compiler/interface2.jl:0
 [6] params! at /home/appleparan/.julia/packages/Flux/bqklb/src/functor.jl:42 [inlined]
 [7] (::typeof(∂(params!)))(::Nothing) at /home/appleparan/.julia/packages/Zygote/8dVxG/src/compiler/interface2.jl:0
 [8] params! at /home/appleparan/.julia/packages/Flux/bqklb/src/functor.jl:48 [inlined]
 [9] (::typeof(∂(params!)))(::Nothing) at /home/appleparan/.julia/packages/Zygote/8dVxG/src/compiler/interface2.jl:0
 [10] params! at /home/appleparan/.julia/packages/Flux/bqklb/src/functor.jl:48 [inlined]
[...]

PR #997, opened 2020-01-19
use Buffer when computing params

piever:
This is an attempt to address the discussion in https://github.com/FluxML/Flux.jl/issues/930#issuecomment-574202189. One would still need to figure out how to create a `Params` object from a vector in a differentiable way (the current approach is `push!` based), and I'm not 100% sure how to do that.

It may be better to figure that out first before merging here, maybe there is a smart trick that makes `push!(ps::Zygote.Params, x)` differentiable and makes this change unnecessary. UPDATE: a differentiable constructor for `Zygote.Params` is now implemented in https://github.com/FluxML/Zygote.jl/pull/471.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -39,9 +39,9 @@ end
 
 trainable(m) = functor(m)[1]
 
-params!(p::Params, x::AbstractArray{<:Number}, seen = IdSet()) = push!(p, x)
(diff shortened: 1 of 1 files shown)


MikeInnes:
Looks good, but it also seems like we might be able to keep `push!` for `Params`, and then this might be unnecessary.

piever:
Yes, this was made unnecessary by https://github.com/FluxML/Zygote.jl/pull/476

PR #1009, opened 2020-01-29
Added Bilinear layer

bhvieira:
A basic implementation inspired on https://pytorch.org/docs/stable/nn.html#bilinear

I haven't exported it, because I think this layer is a bit more esoteric compared with others.

It basically computes interactions between two sets of inputs.

I thought about augmenting it to also include the non-interaction terms (this can easily be done, eg. augmenting the data with a row of ones) but for now it simply mirrors PyTorch's one.

I had to use splatting `vcat(x...)` and `hcat(x...)` in the forward pass. I wanted to avoid it, but with `reduce` I couldn't get gradients. But I think this can be improved.


mcabbott:
Seems like a reason to want https://github.com/FluxML/NNlib.jl/pull/100

But even without that, I think it's not hard to avoid the splats, and go almost 100x faster. Some scribbles here: https://gist.github.com/mcabbott/29cc74f287a95724d6f561f4ed285624

bhvieira:
> But even without that, I think it's not hard to avoid the splats, and go almost 100x faster. Some scribbles here: https://gist.github.com/mcabbott/29cc74f287a95724d6f561f4ed285624

Cool stuff, didn't know about @einsum (as a physicist, I'm specially pleased with it!), really concise. Feel free to push your code to my branch!

mcabbott:
I updated the gist, now OMEinsum's `@ein` is the fastest way. And I believe it will work on GPUs too, which the others may not? Haven't tried as mine is broken.

bhvieira:
If we can't add that dependency we can just fallback to your previous implementation, thanks for the commit @mcabbott!

src/layers/basic.jl
```diff
@@ -226,3 +226,85 @@ end
...
+
+function Bilinear(in1::Integer, in2::Integer, out::Integer, σ = identity;
+  initW = glorot_uniform, initb = zeros)
+  return Bilinear(initW(out, in1, in2), initb(out), σ)
+end
+
+function (a::Bilinear)(x::AbstractMatrix, y::AbstractMatrix)
+  if size(x, 2) ≠ size(y, 2)
+    error("Bilinear expected equal number of samples in both streams. Got $(size(x,2)) and $(size(y,2))")
+  end
```
mcabbott:
```suggestion
  end
  W, b, σ = a.W, a.b, a.σ
```
This got lost in copy-paste I think.

bhvieira:
@mcabbott where does `eachcol` come from?

bhvieira:
Oh, is it `DataFrames.jl`? I think we could define our own here, `DataFrames.jl` would be a big dependency to include.

mcabbott:
No, it's in Base, but perhaps only since 1.1? It is one line though.

src/layers/basic.jl
```diff
@@ -226,3 +226,86 @@ end
...
+(a::Bilinear)(x::AbstractArray) = a(x,x)
+(a::Bilinear)(x::AbstractVector, y::AbstractVector) = vec(a(reshape(x, :,1), reshape(y, :,1)))
+
+function Base.show(io::IO, l::Bilinear)
+  print(io, "Bilinear(", size(l.W, 2), ", ", size(l.W, 3), ", ", size(l.W, 1))
+  l.σ == identity || print(io, ", ", l.σ)
+  print(io, ")")
+end
+
+collecteachcol(x::AbstractMatrix) = collect(eachcol(x))
```
mcabbott:
```suggestion
collecteachcol(x::AbstractMatrix) = [view(x, :,c) for c in axes(x,2)]
```
May actually be quicker without views, `x[:, c]`, but haven't checked.

bhvieira:
> No, it's in Base, but perhaps only since 1.1? It is one line though.

Yeah, I tend to stick to the LTS versions, I'll check it

src/layers/basic.jl
bhvieira: In the source code it's a Generator though , `eachcol(A::AbstractVecOrMat) = (view(A, :, i) for i in axes(A, 2))`

mcabbott: Right, but I added `collect`. You could try without, I'm not sure whether the gradients still work correctly.

bhvieira: Forgot to click the commit suggestion button, sorry for that, was testing locally

bhvieira:
The only error remaining is about `reducehcat`, probably about it's adjoint as it appears in the gradient call in the last test.

dhairyagandhi96:
eachcol is think is Julia v1.1+ only, so will fail on earlier versions

bhvieira:
I got a new implementation working now though, using `Zygote.Buffer`. It's faster than the previous one, and now the code uses it:

```
#current
@btime b($x...); #  48.201 μs (609 allocations: 103.31 KiB)
@btime gradient(() -> sum(abs2.(b($x...))), params(b)); #  11.179 ms (62819 allocations: 8.26 MiB)

#previous
@btime b($x...); #  53.300 μs (1022 allocations: 69.66 KiB)
@btime gradient(() -> sum(abs2.(b($x...))), params(b)); #  11.262 ms (62819 allocations: 8.26 MiB)
```

arnavs:
@dhairyagandhi96 Do you think this code is ready to include in projects? i.e., if I copy your PR into a fork of Flux.

(Also, nice to see that `eachcol` is getting some use... it was a pain to merge in, but I'm glad it happened.)

arnavs:
Actually, I"m having trouble `Chain`ing this to other layers. I see 

```
MethodError: no method matching (::Chain{Tuple{Bilinear{Array{Float32,3},Array{Float64,1},typeof(tanh)},Dense{typeof(tanh),Array{Float32,2},Array{Float32,1}},Dense{typeof(identity),Array{Float32,2},Array{Float32,1}}}})(::Array{Float64,1}, ::Array{Float64,1})
Closest candidates are:
  Any(::Any) at /Users/arnavsood/.julia/packages/Flux/2i5P1/src/layers/basic.jl:32

Stacktrace:
 [1] top-level scope at In[99]:1
```

bhvieira:
@dhairyagandhi96 can this one be merged?

arnavs:
@bhvieira Any reason to not add the `chain` methods? They were what I needed to get this to work.

bhvieira:
@arnavs I think you deleted a comment or something? Didn't see your suggestion for some reason, I might look into it, but could do it in another PR as well

arnavs:
Yeah, I'd made an earlier comment just asking if this was ready to merge. And then a follow up with the bug report. 

Thanks for looking into it. Basically we just need chains to act on two arguments, otherwise you can't use a bilinear layer as the first in a chain. So those two lines work for me, but perhaps there are better ways.

bhvieira:
I fixed that issue @arnavs without touching `Chain`, see if it works for you now 🙂

mcabbott:
Note that `batched_mul` is now merged, FluxML/NNlib.jl#100, and has a gradient https://github.com/FluxML/Zygote.jl/pull/531. Not yet hooked up for cuarrays, but will surely be. I think this PR ought to use that, instead of hacking its own version.

bhvieira:
@mcabbott Gosh this never stops haha. It's cool that we can rely on `batched_mul`, but I wouldn't call using `Buffer` 'hacking' by any means. Anyways, can you open a PR against my branch again? Checks are failing because I probably did something wrong, so I'll look into that when I have the time later today.

mcabbott:
Sorry, no insult intended, if that came off wrong, I'm guilty of earlier hacks. But this is a common operation which, like `*`, should ideally be outsourced to the professionals. And now at last we can easily do so. 

```julia
function (a::Bilinear)(x::AbstractMatrix, y::AbstractMatrix)
    W, b, σ = a.W, a.b, a.σ

    d_z, d_x, d_y = size(W)
    d_x == size(x,1) && d_y == size(y,1) || throw(DimensionMismatch("number of rows in data must match W"))
    size(x,2) == size(y,2) || throw(DimensionMismatch("data inputs must agree on number of columns"))

    # @einsum Wy[o,i,s] := W[o,i,j] * y[j,s]
    Wy = reshape(reshape(W, (:, d_y)) * y, (d_z, d_x, :))

    # @einsum Z[o,s] := Wy[o,i,s] * x[i,s]
    Wyx = batched_mul(Wy, reshape(x, (d_x, 1, :)))
    Z = reshape(Wyx, (d_z, :))

    # @einsum out[o,s] := σ(Z[o,i] + b[o])
    σ.(Z .+ b)
end
```

src/layers/basic.jl
```diff
@@ -253,3 +253,78 @@ end
...
+      for o in axes(W, 1)
+          Z[o, s] = x[:,s]' * W[o,:,:] * y[:,s]
+      end
+  end
+  σ.(copy(Z) .+ b)
+end
+
+(a::Bilinear)(x::AbstractArray) = a(x,x)
+(a::Bilinear)(x::AbstractVector, y::AbstractVector) = vec(a(reshape(x, :,1), reshape(y, :,1)))
+(a::Bilinear)(x::AbstractVector) = a(x...)
```
mcabbott: This seems like a bad idea, if `B(x)` means `y=x` for matrix `x`, and vector `x` is otherwise like a one-column matrix. Insisting on a tuple seems better.

bhvieira: Agreed, I think you are seeing an outdated version of the code though, it's limited to Tuples only

mcabbott: Oh right, sorry!

bhvieira:
With the timely PRs by @mcabbott, I think we are set here and the functionality is better than ever. Is there anything else you think we should do here @dhairyagandhi96?

bhvieira:
Btw, should it be exported? Similarly "uncommon" functionalities aren't exported, so I did not include it, but I can add it you deem it useful.

src/layers/basic.jl
```diff
@@ -253,3 +253,78 @@ end
...
+
+`z[i] = σ.(x' * W[i,:,:] * y .+ b[i])`
+
+If `x` and `y` are matrices, then each column of the output `z = B(x, y)` is of this form.
+
+If `y` is not given, it is taken to be equal to `x`, i.e. `B(x) == B(x, x)`
+The two inputs may also be provided as a tuple, `B((x, y)) == B(x, y)`,
+which is accepted as the input to a `Chain`.
+
+```
```
CarloLucibello:
```suggestion
```julia
```

src/layers/basic.jl
```diff
@@ -253,3 +253,78 @@ end
 function Base.show(io::IO, b::SkipConnection)
   print(io, "SkipConnection(", b.layers, ", ", b.connection, ")")
 end
+
+"""
+    B = Bilinear(in1, in2, out)
```
CarloLucibello:
```suggestion
    Bilinear(in1, in2, out)
```
Also, please reflect truthfully the constructor's signatures here

CarloLucibello:
looks good! I would leave it unexported

src/layers/basic.jl
mcabbott: I'm to blame here, the reason to write `B = ...` was to refer to this later, `z = B(x,y)` etc. It doesn't break the signature though?

bhvieira: Is there a consensus here? @mcabbott ?

CarloLucibello: We should put  the correct signatures here:
```
Bilinear(W, b) 
Bilinear(in1, in2, out, σ=identity;  initW=glorot_uniform, initb=zeros)
```
And B should be changed to Bilinear in the text below

mcabbott: OK, now it's clear what you meant by signatures (although perhaps it should have `W::Array, b::Vector`). But in things below, I'm not sure if you are misreading this or suggesting something else, but replacing `B` in `z = B(x, y)` would be incorrect.

bhvieira:
I changed the text slightly to accommodate the definition of `B`, do you agree with the changes @mcabbott, @CarloLucibello ?

CarloLucibello:
this only needs a mention in the docs

bhvieira:
Will do asap @CarloLucibello

bhvieira:
@CarloLucibello The failed test is GroupNorm Julia nightly, but Bilinear tests passed everywhere. Can this be merged?

CarloLucibello:
glad this finally made it

bors r+

CarloLucibello (approved):

src/layers/basic.jl
```diff
@@ -276,3 +276,79 @@ end
...
+  Wyx = batched_mul(Wy, reshape(x, (d_x, 1, :)))
+  Z = reshape(Wyx, (d_z, :))
+
+  # @einsum out[o,s] := σ(Z[o,i] + b[o])
+  σ.(Z .+ b)
+end
+
+(a::Bilinear)(x::AbstractVecOrMat) = a(x, x)
+(a::Bilinear)(x::AbstractVector, y::AbstractVector) = vec(a(reshape(x, :,1), reshape(y, :,1)))
+(a::Bilinear)(x::NTuple{2, AbstractArray}) = a(x...)
```
DhairyaLGandhi, line 348:
```suggestion
(a::Bilinear)(x::NTuple{2, AbstractArray}) = a(x[1], x[2])
```

To help with inference

src/layers/basic.jl
```diff
@@ -276,3 +276,79 @@ end
...
+  # @einsum Z[o,s] := Wy[o,i,s] * x[i,s]
+  Wyx = batched_mul(Wy, reshape(x, (d_x, 1, :)))
+  Z = reshape(Wyx, (d_z, :))
+
+  # @einsum out[o,s] := σ(Z[o,i] + b[o])
+  σ.(Z .+ b)
+end
+
+(a::Bilinear)(x::AbstractVecOrMat) = a(x, x)
+(a::Bilinear)(x::AbstractVector, y::AbstractVector) = vec(a(reshape(x, :,1), reshape(y, :,1)))
```
DhairyaLGandhi, line 347: Why do we need this separately? Is there any issue in us not restricting the forward pass to AbstractMatrix

src/layers/basic.jl
```diff
@@ -276,3 +276,79 @@ end
...
+  initW = glorot_uniform, initb = zeros)
+  return Bilinear(initW(out, in1, in2), initb(out), σ)
+end
+
+function (a::Bilinear)(x::AbstractMatrix, y::AbstractMatrix)
+  W, b, σ = a.W, a.b, a.σ
+
+  d_z, d_x, d_y = size(W)
+  d_x == size(x,1) && d_y == size(y,1) || throw(DimensionMismatch("number of rows in data must match W"))
+  size(x,2) == size(y,2) || throw(DimensionMismatch("data inputs must agree on number of columns, got $(size(x,2)) and $(size(y,2))"))
```
DhairyaLGandhi, line 333:
```suggestion
  size(x,2) == size(y,2) || throw(DimensionMismatch("Data inputs must agree on number of columns, got $(size(x,2)) and $(size(y,2))"))
```

DhairyaLGandhi (commented):
Generally looks good. One question, could we also test it on the GPU. I would also want to see the performance of the backwards pass on it.

src/layers/basic.jl
```diff
@@ -276,3 +276,79 @@ end
...
   print(io, "SkipConnection(", b.layers, ", ", b.connection, ")")
 end
+
+"""
+    Bilinear(in1, in2, out)
+
+Creates a Bilinear layer, which operates on two inputs at the same time.
+It has parameters `W` and `b`, and its output given vectors `x`, `y` is of the form 
+
+`z[i] = σ.(x' * W[i,:,:] * y .+ b[i])`
```
CarloLucibello, line 286:
```suggestion
    z[i] = σ.(x' * W[i,:,:] * y .+ b[i])
```

CarloLucibello:
needs a rebase

bhvieira:
The testset `"output dimensions"` is failing

mcabbott:
That testset doesn't look like it tests anything related to this layer? Outdims changed & got renamed, but perhaps tests for it should go in https://github.com/FluxML/Flux.jl/blob/master/test/outputsize.jl.

bhvieira:
Weird, I might've added it by mistake when rebasing @mcabbott . Can't see what I did wrong

CarloLucibello (dismissed):
@DhairyaLGandhi merge?

bhvieira:
@DhairyaLGandhi mentioned GPU tests, I could provide such in a few days, it would take me a while to spin Julia in a cluster to test it locally

src/layers/basic.jl
DhairyaLGandhi: In that case maybe we also want a forward pass that accepts a tuple and forwards it here, so that modelling is easier. The previous layer if any can also return a tuple and that is handled gracefully.

Also, wouldn't inputting vectors return vectors by default anyway? Maybe there's some place that assumes matrices which can be relaxed?

src/layers/basic.jl
```diff
@@ -253,3 +253,79 @@ end
...
   print(io, "SkipConnection(", b.layers, ", ", b.connection, ")")
 end
+
+"""
+    Bilinear(in1, in2, out)
+
+Creates a Bilinear layer, which operates on two inputs at the same time.
+It has parameters `W` and `b`, and its output given vectors `x`, `y` is of the form 
+
+    z[i] = σ.(x' * W[i,:,:] * y .+ b[i])
```
mcabbott, line 263: If I'm reading this correctly, the dots here don't do anything:
```suggestion
It has parameters `W` and `b`, and its output given vectors `x`, `y` is another vector `z` with 

    z[i] = σ(x' * W[i,:,:] * y + b[i])
```
It could also be written `dot(x, W[i,:,:], y) + b[i]` but I don't know if that's clearer.

src/layers/basic.jl
mcabbott: > In that case maybe we also want a forward pass that accepts a tuple and forwards it here

See the next line.

mcabbott (dismissed):
This looks fine to me, will rebase #1440 on this to remove `initW` once it's in.

mcabbott:
Bump? This needs a rebase, sadly, as it went cold. 

`reshape` and `batched_mul` have their own GPU tests, so (IMO) that's shouldn't block this.

src/layers/basic.jl
```diff
@@ -253,3 +253,79 @@ end
...
+  # @einsum Z[o,s] := Wy[o,i,s] * x[i,s]
+  Wyx = batched_mul(Wy, reshape(x, (d_x, 1, :)))
+  Z = reshape(Wyx, (d_z, :))
+
+  # @einsum out[o,s] := σ(Z[o,i] + b[o])
+  σ.(Z .+ b)
+end
+
+(a::Bilinear)(x::AbstractVecOrMat) = a(x, x)
+(a::Bilinear)(x::AbstractVector, y::AbstractVector) = vec(a(reshape(x, :,1), reshape(y, :,1)))
```
DhairyaLGandhi, line 324:
```suggestion
function (a::Bilinear)(x, y)
  W, b, σ = a.W, a.b, a.σ

  d_z, d_x, d_y = size(W)
  d_x == size(x,1) && d_y == size(y,1) || throw(DimensionMismatch("number of rows in data must match W"))
  size(x,2) == size(y,2) || throw(DimensionMismatch("Data inputs must agree on number of columns, got $(size(x,2)) and $(size(y,2))"))

  # @einsum Wy[o,i,s] := W[o,i,j] * y[j,s]
  Wy = reshape(reshape(W, (:, d_y)) * y, (d_z, d_x, :))

  # @einsum Z[o,s] := Wy[o,i,s] * x[i,s]
  Wyx = batched_mul(Wy, reshape(x, (d_x, 1, :)))
  Z = reshape(Wyx, (d_z, :))

  # @einsum out[o,s] := σ(Z[o,i] + b[o])
  σ.(Z .+ b)
end

(a::Bilinear)(x::AbstractVecOrMat) = a(x, x)
(a::Bilinear)(x::AbstractVector, y::AbstractVector) = vec(a(x, y))
```

src/layers/basic.jl
DhairyaLGandhi: I'm still unsure about why we need separate methods here. See the suggestion, which should generalise the function somewhat.

DhairyaLGandhi (commented):
Looks like we still need to add GPU tests, and export the layer

src/layers/basic.jl
mcabbott: This has the disadvantage of producing harder to understand error messages on e.g. `x, y = rand(4,2,1), rand(3,2,1)`. And doesn't reduce the number of method definitions. Seems also to make the main method harder to read, IMO, as you aren't told the dimensions. What's gained?

DhairyaLGandhi: Harder how, it's largely unchanged? I would be happy to remove the vector based method entirely but that would return a column vector.

mcabbott: Harder to read because if, at the top of the function, you are told that x,y are matrices, then that is the only possibility you need to imagine when reading the body. And that is the case for which the comments make sense.

DhairyaLGandhi: Well, the error messages would be caught in the messages, and we can improve upon them, we also might want to generalise this to higher dimensions which I'm picking up from your previous comments. In that case, only methods that need the types should have them. If there really is a big difference in implementation for different arrays (which there doesn't seem to be in this case), then sure, else it's unnecessary

mcabbott: I'm not sure what "would be caught in the messages" is meant to mean. The actual error messages are
```
ERROR: MethodError: no method matching (::Bilinear{Array{Float64, 3}, Vector{Float64}, typeof(identity)})(::Array{Float64, 3}, ::Array{Float64, 3})
```
vs:
```
ERROR: MethodError: no method matching *(::Matrix{Float64}, ::Array{Float64, 3})
Closest candidates are:
  *(::Any, ::Any, ::Any, ::Any...) at operators.jl:597
  *(::StridedMatrix{T}, ::StridedVector{S}) where {T<:Union{Float32, Float64, ComplexF32, ComplexF64}, S<:Real} at /Users/me/.julia/dev/julia/usr/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:44
  *(::StridedMatrix{var"#s832"} where var"#s832"<:Union{Float32, Float64}, ::StridedMatrix{var"#s831"} where var"#s831"<:Union{Float32, Float64, ComplexF32, ComplexF64}) at /Users/me/.julia/dev/julia/usr/share/julia/stdlib/v1.7/LinearAlgebra/src/matmul.jl:158
  ...
Stacktrace:
```
It's pretty hard to argue that the second is an improvement. 

The way to generalise to higher dims is to make another method `(a::Bilinear)(x::AbstractArray, y::AbstractArray)` which reshapes all further dims as batch dim. But please make a new PR after this to discuss, it's a bit more complicated than the Dense case, since there can be multiple inputs. Importantly, this catch-all-higher-dims method needs to be *less* specific than the one which implements things for AbstractMatrix. Which is another reason that shouldn't be `(::Any, ::Any)`.

bhvieira:
Thanks for updating the branch @mcabbott . I'll look into the GPU tests later today if I get some time.

mcabbott (approved):
By me it's OK that we test the constituents, we could leave adding integration tests for (say) #1472.

CarloLucibello:
weird. Want to try to approve again? maybe your rights have changed since last approval

mcabbott (dismissed):
bors r+

DhairyaLGandhi:
I think @bhvieira mentioned that they would be able to look at tests soon. Let's wait on that perhaps

mcabbott:
18 days ago, and it's been open a year? Would be good to keep things moving. The constituents are pretty well tested.

DhairyaLGandhi:
Ah I was referring to https://github.com/FluxML/Flux.jl/pull/1009#issuecomment-769225927 which was two days ago.

bors r-

DhairyaLGandhi:
Good to have it, and also with the appropriate testing.

CarloLucibello:
this layer is just doing batched_mul and a bunch of reshape, since this PR has been open for so long let's not waste further time, we can leave without gpu tests. I can add them later

bors r+

bhvieira:
Yay @mcabbott, I'm pretty preocuppied right now (my PhD scholarship requires yearly reports on the grant and also scientific merit, leading to a renewal request, so I'm finding spots between other obligations to work on this PR, which I'm not even using right now). Setting the gpu test is actually pretty simple, I just couldn't find the time to work on it.

test/cuda/layers.jl
```diff
@@ -122,4 +122,15 @@ end
...
+end
+
+@testset "Two-streams Bilinear" begin
+  x = zeros(Float32,10,9) |> gpu
+  y = zeros(Float32,2,9) |> gpu
+  b = Flux.Bilinear(10, 2, 3) |> gpu
+  @test size(b(x,y)) == (3,9)
+  @test sum(abs2, b(x,y)) ≈ 0f0
+  @test_nowarn gs_gpu = gradient(() -> sum(abs2.(b(x,y))), params(b))
+  gs_cpu = gradient(() -> sum(abs2.(b(x |> cpu,y |> cpu) |> cpu)), params(b))
```
CarloLucibello, line 134:
```suggestion
   b_cpu, x_cpu, y_cpu = b |> cpu, x |> cpu, y |> cpu
  gs_cpu = gradient(() -> sum(abs2.(b_cpu(x_cpu, y_cpu)), params(b_cpu))
```

test/cuda/layers.jl
```diff
@@ -122,4 +122,15 @@ end
...
+  @test gs_cpu ≈ gs_gpu
```
CarloLucibello, line 135:
```suggestion
  @test Array(gs_cpu) ≈ gs_gpu
```

test/cuda/layers.jl
```diff
@@ -122,4 +122,15 @@ end
...
+
+@testset "Two-streams Bilinear" begin
+  x = zeros(Float32,10,9) |> gpu
+  y = zeros(Float32,2,9) |> gpu
+  b = Flux.Bilinear(10, 2, 3) |> gpu
+  @test size(b(x,y)) == (3,9)
+  @test sum(abs2, b(x,y)) ≈ 0f0
+  @test_nowarn gs_gpu = gradient(() -> sum(abs2.(b(x,y))), params(b))
+  gs_cpu = gradient(() -> sum(abs2.(b(x |> cpu,y |> cpu) |> cpu)), params(b))
+  @test Array(gs_cpu) ≈ gs_gpu
```
CarloLucibello, line 136:
```suggestion
  @test gs_cpu ≈ Array(gs_gpu)
```

bhvieira:
Would the `@test_nowarn` not return `gs_gpu` to the local scope  @CarloLucibello?

CarloLucibello:
could be. I didn't even know it existed though. I'll just remove the test

test/cuda/layers.jl
```diff
@@ -122,4 +122,16 @@ end
...
\ No newline at end of file
+end
+
+@testset "Two-streams Bilinear" begin
+  x = zeros(Float32,10,9) |> gpu
+  y = zeros(Float32,2,9) |> gpu
+  b = Flux.Bilinear(10, 2, 3) |> gpu
+  @test size(b(x,y)) == (3,9)
+  @test sum(abs2, b(x,y)) ≈ 0f0
+  @test_nowarn gs_gpu = gradient(() -> sum(abs2.(b(x,y))), params(b))
```
CarloLucibello, line 133:
```suggestion
  gs_gpu = gradient(() -> sum(abs2.(b(x, y))), params(b))
```

test/cuda/layers.jl
```diff
@@ -122,4 +122,16 @@ end
...
+@testset "Two-streams Bilinear" begin
+  x = zeros(Float32,10,9) |> gpu
+  y = zeros(Float32,2,9) |> gpu
+  b = Flux.Bilinear(10, 2, 3) |> gpu
+  @test size(b(x,y)) == (3,9)
+  @test sum(abs2, b(x,y)) ≈ 0f0
+  gs_gpu = gradient(() -> sum(abs2.(b(x, y))), params(b))
+  b_cpu, x_cpu, y_cpu = b |> cpu, x |> cpu, y |> cpu
+  gs_cpu = gradient(() -> sum(abs2.(b_cpu(x_cpu, y_cpu))), params(b_cpu))
+  @test gs_cpu ≈ Array(gs_gpu)
```
CarloLucibello, line 136:
```suggestion
  for (pgpu, pcpu) in zip(params(b), params(b_cpu))
    @test gs_cpu[pcpu] ≈ Array(gs_gpu[pgpu])
  end
```

CarloLucibello:
I really hope this goes green, this commit suggestion thing is becoming painful :sweat_smile:

CarloLucibello (dismissed):

CarloLucibello:
victory!

bors r+

bhvieira:
@CarloLucibello thanks for the efforts haha. I had no idea a simple equality test between gpu and cpu would take so much. Are gpu gradients stored as gpu arrays? Perhaps if we moved it back to the cpu it would've worked.

CarloLucibello:
@DhairyaLGandhi maybe you should just merge manually here

CarloLucibello (approved):

DhairyaLGandhi (approved):

PR #1010, opened 2020-01-30
document correction

mrchaos:
using Flux: update! ==> using Flux.Optimise: update!


docs/src/training/optimisers.md
```diff
@@ -14,14 +14,14 @@ loss(x, y) = sum((predict(x) .- y).^2)
 x, y = rand(5), rand(2) # Dummy data
 l = loss(x, y) # ~ 3
 
-θ = Params([W, b])
+θ = params([W, b])
```
bhvieira: This modification is not necessary, though `Params` isn't exported. Otherwise, you could just call `params(W, b)` as well.

CarloLucibello:
Part of this has been addressed in #1052 . We could use the `params` correction though

PR #1013, opened 2020-01-31
Adapt to GPUArrays/CuArrays changes

maleadt:
Changes in response to https://github.com/JuliaGPU/GPUArrays.jl/commit/a29df67184a0fe2afda71331bb9833ef00ad4be3 and https://github.com/JuliaGPU/CuArrays.jl/pull/576. I suppose the next CuArrays release will need to be breaking because of this.

Maybe the `crossentropy` signature needs to be adjusted to support integer vectors, but I'll leave that decision up to Flux developers. This at least is the quick fix to get the tests passing again.


maleadt:
Bump; this breaks Flux reverse testing on CuArrays.jl.

dhairyagandhi96:
Sgtm

bors r+

PR #1017, opened 2020-02-03
Refactor train!

dhairyagandhi96:
This is in response to #666 where we can visualise `train!` as a wrapper around the `step!` function, while maintaining the same api.

To actually accumulate the loss as in https://github.com/FluxML/Flux.jl/issues/666#issuecomment-471307139, we could of course go with the loss function being a simple closure here, getting rid of the need to send around the mini batch.

With the changes to Zygote, it might be nice to actually have this in our interface.

cc @MikeInnes @oxinabox

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -42,6 +42,13 @@ function stop()
   throw(StopException())
 end
 
+function step!(loss, ps, minibatch, opt)
+  gs = gradient(ps) do
+    loss(minibatch...)
+  end
+  update!(opt, ps, gs)
+end
+
 """
     train!(loss, params, data, opt; cb)
 
@@ -65,10 +72,11 @@ function train!(loss, ps, data, opt; cb = () -> ())
   cb = runall(cb)
   @progress for d in data
     try
-      gs = gradient(ps) do
-        loss(d...)
-      end
-      update!(opt, ps, gs)
+      # gs = gradient(ps) do
+      #   loss(d...)
+      # end
+      # update!(opt, ps, gs)
+      step!(loss, ps, d, opt)
       cb()
     catch ex
       if ex isa StopException


src/optimise/train.jl
```diff
@@ -42,6 +42,13 @@ function stop()
   throw(StopException())
 end
 
+function step!(loss, ps, minibatch, opt)
```
oxinabox: Can this have a docstring, and be exported?

dhairyagandhi96: Yeah, it would be good to converge on its implementation, with that I'll add the doctrings and a basic test too.

src/optimise/train.jl
```diff
@@ -42,6 +42,13 @@ function stop()
...
+  gs = gradient(ps) do
+    loss(minibatch...)
+  end
+  update!(opt, ps, gs)
```
CarloLucibello: Isn't this return incompatible with the interface proposed in https://github.com/FluxML/Flux.jl/issues/666#issuecomment-516240099?

dhairyagandhi96: It returns `nothing`, which should be fine for most cases. For that comment specifically, yes, but it really is just a matter of agreeing on the semantics desired out of it. returning the loss here should be fine, but seeing as it is stepping through a batch, and doing a training step, I feel the correct return should be a nothing, with a slightly more trained model as it were

dhairyagandhi96:
Pinging @MikeInnes for his thoughts on the semantics that would make most sense here.

MikeInnes:
I agree that closing over the batch would make the most sense. I also agree it's best to avoid returning things like losses, since as discussed earlier we can just close over those things, and we will eventually want an out-of-place version of this that returns a model.

`train!` will likely need a fair bit of refactoring once we figure out new optimisers + accelerator support, but we can figure that out later.

PR #1019, opened 2020-02-05
Added some loss functions

AdarshKumar712:
I have added following loss functions:
1. Mean absolute error
2. Mean squared Logarithmic error
3. Huber loss
4. Squared Hinge loss

I have also done some function documentation improvements. Kindly review and suggest further changes(if required).


src/layers/stateless.jl
```diff
@@ -2,9 +2,93 @@ using CuArrays
 using NNlib: logsoftmax, logσ
 
 # Cost functions
+"""
+    mae(ŷ, y)
+L1 loss function. Computes the mean of absolute error between prediction and true values
+"""
+mae(ŷ, y) = sum(abs.(ŷ .- y)) * 1 // length(y)
```
oxinabox:
```suggestion
mae(ŷ, y) = sum(abs.(ŷ .- y)) / length(y)
```

Issue #1016, opened 2020-02-03
crossentropy should have label smoothing 


sambitdash:
`crossentropy` method should support label smoothing. The equivalent API can be seen in TensorFlow at:

https://www.tensorflow.org/api_docs/python/tf/keras/losses/BinaryCrossentropy

CarloLucibello:
Do we need this? It is not very common AFAIK, pytorch doesn't implement it, and users can implement it very easily by passing to the loss function the shifted labels

sambitdash:
> Do we need this? It is not very common AFAIK, pytorch doesn't implement it, and users can implement it very easily by passing to the loss function the shifted labels

Enhancement request is getting implemented in pytorch. 

https://github.com/pytorch/pytorch/issues/7455

PR #1025, opened 2020-02-07
Implementation of label smoothing with crossentropy 

sambitdash:
crossentropy and binarycrossentropy can have label smoothing. 

Addressing as per issue #1016.


test/layers/stateless.jl
```diff
@@ -22,31 +22,42 @@ const ϵ = 1e-7
...
     @test logitcrossentropy(logŷ, y, weight = [.5, .5]) ≈ lossvalue/2
+    @test logitcrossentropy(logŷ, y, weight = ones(2), label_smoothing=0.2) ≈ lossvalue
+    @test logitcrossentropy(logŷ, y, weight = [.5, .5],label_smoothing=0.2) ≈ lossvalue/2
     @test logitcrossentropy(logŷ, y, weight = [2, .5]) ≈ 1.5049660054074199
   end
 
   logŷ, y = randn(3), rand(3)
+  ydb = [0.1, 0.9, 0.9, 0.1]
+  yb  = [1., 1., 0., 0.]
+  yr  = [2.0828626352604234, 0.3250829733914482, 2.0828626352604234, 0.3250829733914482]
```
oxinabox: Can you indicate the source of this number?

sambitdash: Hard coded numbers have been removed.

sambitdash:
> Can we have some tests that:
> 
> 1. show label smoothing clearly differing from not using it
> 2. show properties of label soothing, e.g test that the loss with increased smoothing is larger/smaller the decreased etc

The label smoothing is a regularizer and the most significant is when the distribution of the data is skewed to one value. label smoothing ensures all zeros for weights do not end up getting to a statistically maximal category as a local minima. One a symmetric distribution over the categories the label smoothing has no effect.

oxinabox:
> (2) is kind of very subjective of the label smoothing will reduce losses for dissimilar distributions and increases losses for similar distributions.

That sounds testable.

test/layers/stateless.jl
```diff
@@ -25,10 +25,21 @@ const ϵ = 1e-7
   ylp = [0.9 0.1]'
   logylp = [0.0 v]'
 
+
```
oxinabox:
```suggestion
  # Construct `sim`ilar and `dis`imilar versions of the dataset so we can test effect of smoothing
  # smoothing should decrease loss on disimilar and increase the loss on similar, compared to 
  # the loss without smoothing
```

src/layers/stateless.jl
```diff
@@ -5,56 +5,81 @@ using NNlib: logsoftmax, logσ
...
-  return -sum(y .* log.(ŷ)) .* weight * 1 // size(y, 2)
-end
+_crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat, weight::Union{Number, AbstractVector}) =
+  _crossentropy(ŷ, y.*weight, nothing)
 
-function _crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat, weight::AbstractVector)
-  return -sum(y .* log.(ŷ) .* weight) * 1 // size(y, 2)
+function crossentropy(ŷ::AbstractVecOrMat, y::AbstractVecOrMat; weight=nothing, label_smoothing=zero(eltype(y)))
+  if !iszero(label_smoothing)
+    y = y .* (1 - label_smoothing) .+ label_smoothing* 1 // size(y, 1)
```
oxinabox: Is it worth being explict:
```suggestion
    y = y .* (1 - label_smoothing) .+ (label_smoothing // size(y, 1))
```

sambitdash: Yes. label_smoothing can be non-integral type that the operator `//` may not support.

oxinabox: Ah..
```suggestion
    y = y .* (1 - label_smoothing) .+ (label_smoothing * 1 // size(y, 1))
```

sambitdash:
@MikeInnes can you take a look at this change and suggest if it's good to merge.

sambitdash:
Any reasons this is not being merged?

CarloLucibello:
I'm not convinced that this feature is popular enough to be worth adding. Of course, I may be wrong

oxinabox:
> I'm not convinced that this feature is popular enough to be worth adding. Of course, I may be wrong

We have a dozen less popular activation functions.
TensorFlow has this build in we should thus also.
Plus it is tricky enough to implement in a way that plays nicely with all types, that i would like a library for it.
And what other librarry but Flux?

gxyd:
Hi, any plan to get to this in? Do we need checks to make sure that `label_smoothing` is between `0` and `1`?

I'll be interested in seeing this in.

gxyd:
Hi, I can send in a new PR with my commits on top of this PR's commits, would that be ok?

sambitdash:
> Hi, I can send in a new PR with my commits on top of this PR's commits, would that be ok?

@gxyd, I am no longer tracking this PR nor have the relevant set up to track this PR. I am presuming when the reviewers find the change acceptable they shall be able to merge the suitable changes. If you want to take over this PR and merge to your code, feel free to do so.

sambitdash:
closing was completely unintentional. Reopening.

oxinabox:
@DhairyaLGandhi this is still waiting for your review.
It's an API change

DhairyaLGandhi:
Needs a rebase, imo better to hide the control flow in dispatch since we should check for runtimes for the backwards pass and correctness

PR #1028, opened 2020-02-08
Common questions answered in docs

dhairyagandhi96:
cc @MikeInnes


docs/src/models/basics.md
```diff
@@ -219,3 +219,21 @@ Flux.@functor Affine
...
+
+By default all the fields in the `Affine` type are collected as its parameters, however, in some cases it may be desired to hold other metadata in our "layers" that may not be needed for training, and are hence supposed to be ignored while the parameters are collected. With Flux, it is possible to mark the fields of our layers that are trainable in two ways.
+
+The first way of achieving this is through overloading the `trainable` function.
+
+```julia
+Flux.trainable(a::Affine) = (a.W, a.b,)
+```
+
+To add other fields is simply to add them to the tuple.
```
CarloLucibello:
```suggestion
Only the fields returned by `trainable` will be collected as trainable parameters of the layer when calling `Flux.params`. 
```

docs/src/models/basics.md
```diff
@@ -219,3 +219,21 @@ Flux.@functor Affine
...
+
+Another way of achieving this is through the `@functor` macro. Here, wee can mark the fields we are interested in like so:
+
```
CarloLucibello:
```suggestion
Another way of achieving this is through the `@functor` macro. Here, we can mark the fields we are interested in by grouping them in the second argument:
```

dhairyagandhi96:
Only question to answer is that using the `functor` macro can end up with making it incompatible with other constructs, since it needs some proper constructors defined for `fmap` to pick up, which might be confusing for newcomers. In that case, `trainable` is a safer bet to override. Should we even mention `functor` in that case?

dhairyagandhi96:
Bump

MikeInnes:
This is good content, thanks @dhairyagandhi96. My only concern is about interrupting the flow of these docs; it's good to have detail but we also want the `basics` and `training` sections to be a bit more tutorial like, and these additions could be overwhelming to someone who just wants to get started. So it might be good to split these out, e.g. link to advanced notes on training somewhere or the `@functor` docstring which can detail how to freeze parameters.

Mainly just a thought; if it's hard to do that in a nice way, this would be fine too.

dhairyagandhi96:
Happy to move this in an advanced usage type section and see flow like that.

johnnychen94:
May I add one more question here? :P

I don't quite see how normalizing layers switch their behaviors wrt train/test mode. It says it's in an :auto mode, but the only place I find is https://github.com/FluxML/Flux.jl/blob/35f6998be7572bb557948d3cee65797be22c9019/src/layers/normalise.jl#L1

and this "constant" function looks really weird at first glance... It would be great if it can be explained.

dhairyagandhi96:
That's a neat trick. Essentially, through the `@adjoint` macro, Zygote "registers" how to handle a particular method definition. It expects to return the forwards pass value, and a pullback function. This method definition is only exposed when in the "context" of a `gradient` call. So by having the actual definition of what the forwards pass is doing inside that context, rather than when using it as a normal function, we can separate its behaviour normally vs during training, at compile time.

dhairyagandhi96:
Could we merge this too?

docs/src/models/advanced.md
```diff
@@ -0,0 +1,61 @@
...
+
+`ps` now holds a reference to only the parameters of the layers passed to it.
+
+During training, now the gradients would only be applied to the last `Dense` layer (and the `softmax` layer, but that is stateless so doesn't have any parameters), so only that would have its parameters changed.
+
+`Flux.params` also takes multiple inputs to make it easy to collect parameters from heterogenous models with a single call. A simple demonstration would be if we wanted to omit optimising the second `Dense` layer in the previous example. It would look something like this:
+
+```julia
+Flux.params(m[1], m[3:end])
+```
```
CarloLucibello: Could you add something like this, until we have a better solution for freezing  a single parameter:

In order to freeze a specific parameter of a specific layer which already entered the `Params` object `ps`, we can
simply delete it from `ps`:
```julia
ps = params(m)
delete!(ps, m[2].bias) 
```

Issue #1026, opened 2020-02-07
Type Promotion often Unwieldy and day Ruining


caseykneale:
So I've blown 2 days tracking pervasive errors in making basic toy models for a project at work... The errors are Undef Ref errors... Somewhere in Zygote on an @adjoint function...
I believe the issue is reshape in a chain or a model or something explodes the world... Below is a minimum not working example...

```Julia
using Flux
function permute_data( batch::Array )
    s = size(batch)
    reshape( batch, (s[2],1,1,s[1]) )
end

Flux.@nograd permute_data
squish(z) = reshape(z, :, size(z, 4))

Flux.@nograd squish
squishgrad(z) = reshape(z, ( prod( size(z)[1:3]),  size(z, 4) ) )

stop_gradient(x) = x
Flux.@nograd stop_gradient
#Let's make a model....
obs, vars       = ( 10, 200 )
proxydata       = randn(obs,vars)
proxydata       = permute_data( proxydata )
ToEmbeddingA    = Flux.Conv( ( 5, 1 ),  1 => 16, relu)
ToEmbeddingB    = Flux.Conv( ( 5, 1 ),  16 => 24, relu)

freeze_forelayer(x) = ToEmbeddingA(x) |> ToEmbeddingB
Flux.@nograd freeze_forelayer
newproxy        = freeze_forelayer(proxydata)
newproxysize    = size(reshape(newproxy, :, size(newproxy, 4)))

EmbeddingA      = Flux.Dense( newproxysize[1], 32,  relu )
EmbeddingB      = Flux.Dense( 32, 32,  relu )

regression_head = Flux.Dense( 32, 1, relu )
siamese_head    = Flux.Dense( 32, 1, relu )

regr_model = Chain(  freeze_forelayer, squishgrad,
                    EmbeddingA, EmbeddingB,
                    regression_head )


siam_model2 = Chain(  ToEmbeddingA, ToEmbeddingB, squishgrad,
                    EmbeddingA, EmbeddingB,
                    regression_head )

SGD = Descent( 10.0 )
mseloss( x1, y1 ) = sum( regr_model( x1 ) .- y1 )
smseloss( x1, y1 ) = sum( siam_model( x1 ) .- y1 )

ex = proxydata

yeye =  Matrix(transpose(randn(10)))

println(siam_model.layers[1].weight[1,1,1,1])
Flux.train!( mseloss, Flux.params( regr_model ), [ (ex, yeye) ], SGD )
[...]

Issue #815, opened 2019-07-25
Flux#zygote slower than Tracker


crstnbr:
MWE:

```julia
using Flux, BenchmarkTools
using Flux: crossentropy, onecold, onehotbatch, throttle, @epochs
using Printf, Statistics, Random
using Base.Iterators: repeated

confs_left = rand(64,4000)
confs_right = rand(64,4000)

# set up as training data
neach = size(confs_left, 2)
X = hcat(confs_left, confs_right)
labels = vcat(fill(1, neach), fill(0, neach))
Y = onehotbatch(labels, 0:1)
dataset = repeated((X, Y), 10)

# create neural network with 10 hidden units and 2 output neurons
Random.seed!(123)
m = Chain(
  Dense(64, 10, relu),
  Dense(10, 2),
  softmax)

# define cost-function
loss(x, y) = crossentropy(m(x), y)
accuracy(x, y) = mean(onecold(m(x)) .== onecold(y))

opt = ADAM()

println("-------- Training")
@btime Flux.train!($loss, params($m), $dataset, $opt)
```

I get `955.648 ms (28965221 allocations: 587.11 MiB)` with Flux + Zygote and `47.298 ms (7471 allocations: 63.52 MiB)` with Flux + Tracker.

Is this expected because I'm running Julia 1.1?

merckxiaan:
I can reproduce this issue.
To make sure the performance regression happens in the backwards pass, I ran gradient on both versions. (Tracker: left; Zygote: right)
![image](https://user-images.githubusercontent.com/31353884/61908911-11e52900-af31-11e9-8c8e-734164eb93f2.png)
I also tested on Julia 1.3 and the issue remains.

mkschleg:
So I've been playing around with Zygote a bit, and something I've noticed is it suffers terrible performance when you are doing type conversions in the loss you are optimizing over, and Tracker will also have worse performance but isn't effected as much as zygote. If you run the above code, except w/ 

```julia
confs_left = rand(Float32, 64,4000)
confs_right = rand(Float32, 64,4000)
```

you will get much better performance. Basically, you want to make sure your inputs match the type of your model (at least I think). For me the speed up is significant:

W/ Float64s
```julia
julia> @btime Flux.train!($loss, Flux.params($m), $dataset, $opt)
  1.028 s (26084765 allocations: 522.98 MiB)
```

W/ Float32s
```julia
[...]

PR #1031, opened 2020-02-08
Debugging messages about type mismatches

mcabbott:
This adds a few warning messages about conversions to Float64 or worse, which seem to be a common problem, today #1026, last week #815.

These are behind `@debug`, which I see is also used by NNlib for some such warnings. Another option would be to make these errors, controlled by some `Flux.allowconvert(false)` along the lines of what CuArrays does for scalar access, see earlier commit. 

I'm not sure they will catch all the activation-function bugs mentioned in `performance.md`. Perhaps it would be better to attach the warning to every layer, not to `Chain`.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -41,28 +41,20 @@ function Base.show(io::IO, c::Chain)
   print(io, ")")
 end
 
-"""
-    allowconvert(flag=true)
-Setting this to `false` will cause an error on some type mismatches (instead of silently converting).
-"""
-allowconvert(flag::Bool=true) = convert_allowed[] = flag
-
-const convert_allowed = Ref(true)
-
-convert_check(arg...) = convert_allowed[] || error(string(arg..., ".\n",
-  "Flux.allowconvert(true) will disable this error."))
-
-Zygote.@nograd convert_check
-
 @adjoint function (c::Chain)(x)
   y, back = Zygote.pullback(applychain, c.layers, x)
   T = typeof(y)
   y, dy -> begin
-    eltype(y) == eltype(dy) ||
-      convert_check("Chain(...) has output ", eltype(y), " but receives gradient ", eltype(dy))
+    eltype(y) == eltype(dy) || debug_string(
+      "Chain(...) has output of eltype ", eltype(y),
+      " but receives gradient of eltype ", eltype(dy))
     back(dy)
   end
 end
+
+Zygote.@nograd debug_string
+
+debug_string(arg...) = @debug string(arg...)
 
 """
     outdims(c::Chain, isize)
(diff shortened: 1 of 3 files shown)


src/layers/basic.jl
```diff
@@ -39,6 +39,20 @@ function Base.show(io::IO, c::Chain)
   print(io, ")")
 end
 
+@adjoint function (c::Chain)(x)
+  y, back = Zygote.pullback(applychain, c.layers, x)
```
dhairyagandhi96: Not all models will exhibit changes like in `Chain`. Many of the culprits identified earlier were actually coming from the loss functions, so might not get caught here.

mcabbott: I thought this would catch loss function problems, but not activation functions within layers?

dhairyagandhi96: Interesting, will this need to be implemented for all cases of non homogenous eltypes for every layer separately?

dhairyagandhi96:
I'm a big fan of providing users with the option to get a better hold of performance and debugging that comes through so easily, thanks @mcabbott !!

Could you please do a quick rebase?

src/layers/basic.jl
mcabbott: Right, this doesn't catch the following:
```
julia> m = Chain(Dense(2,2, x->x + 0.1)); x= rand(Float32, 2); y=rand(Float32, 2);

julia> gradient(() -> sum(m(x) .- y), params(m))
Grads(...)

julia> m(x) |> typeof
Array{Float64,1}
```
I think I had a go at hooking onto `applychain` but it got messy.

mcabbott:
Sure. I guess a remaining question is whether these want to be a bigger warning, say `@warn`. And if so, whether there needs to be a mechanism for turning them off. An earlier commit had one.

dhairyagandhi96:
From your comment on the other issue, you'd shown a significant speedup that could have been avoided if we could be type stable, which makes it critical.

Having tooling around being able to find these gotchas I think is a more stable and longer term useful situation, so as long as people can access it easily, without performance penalties to the actual runtime, I think is a fair deal, so a warn, though very tempting, might be unnecessary.

src/layers/basic.jl
dhairyagandhi96: Comparing input eltype to output eltype would catch it, no?

mcabbott: Of the whole chain? Yes I guess that would work, done!

dhairyagandhi96:
Could do with tests that demonstrate the cases this catches and ones it misses

mcabbott:
Perhaps this is ready?

What it doesn't do is tell you which layer of a Chain messes up the forward pass. Nor will it catch at all if a particular layer messes up the backward pass, although that doesn't seem to be common.

dhairyagandhi96:
Pinging @MikeInnes for his thoughts on this

racinmat:
@MikeInnes  pinging again as this would really help, e.g. with hunting down obscure bugs like https://github.com/FluxML/Flux.jl/issues/965

mcabbott:
It'll have to be a fresh PR as apparently rebasing has disabled the "reopen" button. IMO this would still be worth doing, and should be a warning not an `@debug` message (i.e. printed by default).

racinmat:
@CarloLucibello  seemed active last days.
I would love to see it merged, it would have saved us so much time we spent hunting bugs 😅

mcabbott:
https://discourse.julialang.org/t/flux-con-warning/49456 is today’s example of a nice friendly this-will-be-slow warning.

PR #1033, opened 2020-02-15
added GroupedConvolutions and ChannelShuffle

gartangh:


Manifest.toml
```diff
@@ -8,15 +8,15 @@ version = "0.5.0"
 
 [[AbstractTrees]]
 deps = ["Markdown"]
-git-tree-sha1 = "8201f932428d25a2e2903300764515754847d87d"
```
maleadt: Are the Manifest changes required? If not, leave them out.

src/layers/basic.jl
```diff
@@ -214,7 +214,7 @@ size(sm(x)) == (5, 5, 11, 10)
 """
 struct SkipConnection
   layers
-  connection  #user can pass arbitrary connections here, such as (a,b) -> a + b
+  connection  # user can pass arbitrary connections here, such as (a,b) -> a + b
```
maleadt: needless whitespace change

test/layers/basic.jl
```diff
@@ -28,7 +28,7 @@ import Flux: activations
   end
 
   @testset "Dense" begin
-    @test  length(Dense(10, 5)(randn(10))) == 5
```
maleadt: same

maleadt:
I guess this will need to use cuDNN's support for grouped convolutions to run efficiently on the GPU? Did you write this code with that in mind? Also, @contributors, would that need an API in NNlib? We've kind-of been moving away from Flux using cuDNN internals.

Manifest.toml
gartangh: I undid the manifest changes [in this commit](https://github.com/FluxML/Flux.jl/pull/1033/commits/0feabec499fa6cc10cab452240e7b29ac4b24671).

src/layers/basic.jl
gartangh: I undid the whitespace change [in this commit](https://github.com/FluxML/Flux.jl/pull/1033/commits/0feabec499fa6cc10cab452240e7b29ac4b24671).

test/layers/basic.jl
gartangh: I undid the whitespace change [in this commit](https://github.com/FluxML/Flux.jl/pull/1033/commits/0feabec499fa6cc10cab452240e7b29ac4b24671).

gartangh:
@maleadt , I believe the [cuDNN API for grouped convolutions](https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#grouped-convolutions), can only run more efficiently when all `paths` consist of the same convolution and `split=true`. This is the case for Xception and ResNeXt.
This implementation also provides support for each path acting on the complete input (with `split=false`) and paths being different among each other (so different operations on different parts of the input). This is used in Inception variants.

maleadt:
> the [cuDNN API for grouped convolutions](https://docs.nvidia.com/deeplearning/sdk/cudnn-developer-guide/index.html#grouped-convolutions), can only run more efficiently when all `paths` consist of the same convolution and `split=true`

Right, so your current design does not allow for use of cuDNN. If you, e.g. move the split field into the typevar you could dispatch on that (and T being a homogeneous NTuple, I presume).

dhairyagandhi96:
Could you add some precedent here for these layers and their general popularity? PyTorch does this via a `groups` kwarg, which seems cleaner since NNlib can already support that

PR #1037, opened 2020-02-16
update compat to Juno 0.8

heliosdrm:


heliosdrm:
No longer needed if #1039 is merged

oxinabox:
This is better than #1039 as it also bumps the version so a release can be tagged.

dhairyagandhi96:
Isn't a compat entry in general needed for everyone who would want to use older versions of Flux with newer versions of Juno?

PR #1041, opened 2020-02-20
add NNlib docs + misc docs improvements

CarloLucibello:
Partially addressing https://github.com/FluxML/NNlib.jl/issues/137.

Also, I'm leaving out the `σ`  activation and using its alias `sigmoid`, since `σ` conveys little information and it is also used to denote a generic activation in the Dense layer. I think we should deprecate `σ` in NNlib, has this been discussed already?

In an ideal world, before merging this, we should get NNlib to either unexport or add docs to its undocumented exports


CarloLucibello:
I chose to expose in the documentation only a subset, which should be the most relevant to users, of the functions exported by Flux and re-exported by NNlib (most of which are undocumented in any case). 

We should for https://github.com/FluxML/NNlib.jl/pull/172 to be merged  before merging this PR

CarloLucibello:
This is ready for review. We should wait for the new release of NNlib before merge

dhairyagandhi96:
Can you rebase on master please?

dhairyagandhi96:
Ah, the log and exp bugs showing up in GitLab.

CarloLucibello:
ah, is current master broken? How do we fix it?

dhairyagandhi96:
I would do a run with Zygote master to see if we caught the codepath with the recent changes there

CarloLucibello:
> I would do a run with Zygote master to see if we caught the codepath with the recent changes there

how do we do that?

CarloLucibello:
how do we get rid of these cuda exp errors?

CarloLucibello:
@maleadt any hint on this?

CarloLucibello:
it should work now

bors r+

PR #1046, opened 2020-02-23
Bump Colors compat to include 0.10, 0.11

ianshmean:
I think 0.10 was erroneously skipped in the auto-generated #1038


dhairyagandhi96:
Yeah, this seems better here, thanks @ianshmean!

PR #1047, opened 2020-02-24
Edit description of convolutional layer

MotJuMi:
"# batches" is understood as number of batches, but it N in WHCN means number of data points in a single batch.

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -17,7 +17,7 @@ Example: Applying Conv layer to a 1-channel input using a 2x2 window size,
     out = 16
     Conv((2, 2), 1=>16, relu)
 
-Data should be stored in WHCN order (width, height, # channels, # batches).
+Data should be stored in WHCN order (width, height, # channels, batch size).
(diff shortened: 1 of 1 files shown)


src/layers/conv.jl
```diff
@@ -17,7 +17,7 @@ Example: Applying Conv layer to a 1-channel input using a 2x2 window size,
     out = 16
     Conv((2, 2), 1=>16, relu)
 
-Data should be stored in WHCN order (width, height, # channels, # batches).
+Data should be stored in WHCN order (width, height, # channels, batch size).
```
dhairyagandhi96: Maybe Data should be read as batch/ minibatch and data can then be a collection of batches. However in the case of just one large dataset, the `step!` function can be used instead?

MotJuMi: It's not about design and implementation, it's just about description. "# batches" is interpreted as "number of minibatches" but the real meaning is "length of each batch".

dhairyagandhi96: Fair enough, we should probably be a bit more clear about what we want to say here anyway

Issue #1056, opened 2020-02-28
@functor needs better documentation


freemin7:
I was studying the Flux source code how to create me own layer types.
I came across the macro `@functor` whose purpose remained unclear to me. After asking in Slack it was signaled to me that those question comes up a lot and that it should be answered here.

I could imagine the documentation taking these forms:

- Make really good clear examples and explanation of `@functor`  in docs and explain when you need it and what for
- an expansion of the affine tutorial which includes @functor

dhairyagandhi96:
Relevant #1028

HamletWantToCode:
One thing about `@functor` still confuse me. Inside `functor`'s definition, e.g.
```julia
functor(x::AbstractArray) = x, y -> y
```

it returns a `Tuple`. The 2nd element of the tuple is a function, however, it seems that this function has never been used ( I searched through the repository, but just find `functor(m)[1]`, which uses the 1st element of the returned tuple ).

So:
1. Why `functor`'s return value include e.g. `y->y` ?

2. (continue the 1st) If it's necessary, where & when should we use it ?

ToucheSir:
For anyone happening upon this issue, the [Functors README](https://github.com/FluxML/Functors.jl) has a good explanation of what's returned and what it's all used for. In this case, `y -> y` is the `re(construct)` function that allows you to "rehydrate" a structure from its constituent parts. Since arrays are considered un-destructureable base values, no logic is required to restructure them and `re` is just the identity function.

PR #1059, opened 2020-03-01
Make really good clear examples and explination of @functor  in docs

findmyway:


johnnychen94:
Thanks for the explanation! Should be enough to close #1056

It would be fantastic if there's some brief docstring to `@functor` 😄

CarloLucibello:
@findmyway I started to add docstrings to the `functor.jl` file, then I recalled about this PR. Please fill free to take/adapt anything you want from #1068 to add here, even all of it.

findmyway:
I just switched the base branch from master to your PR directly, in case you need to update the docs further 😃

ToucheSir:
@CarloLucibello seems like this fell off the map and never made it into master? Would it be possible to cherry-pick this commit? The [current docs](https://fluxml.ai/Flux.jl/stable/gpu/) don't mention functors at all!

(happy to put in a PR for this)

DhairyaLGandhi:
Yeah, i think putting up the PR a clear example of functor would be good. The motivation may not need to define the internals of the `functor` function etc. Just a thought

ToucheSir:
I stand corrected, there was already a nice explanation on https://fluxml.ai/Flux.jl/stable/models/advanced/ and I just wasn't finding it through the docs search. Mea culpa :)

PR #1062, opened 2020-03-01
docstring ensure signature code formatting

visr:
by using a four space indent instead of two

Fixes issues seen here:

image

Where the type signature has no code formatting, and a code block is introduced that throws off the rest of the formatting.


CarloLucibello:
thanks!

bors r+

CarloLucibello:
don't know what went wrong with bors, trying again

bors r+

PR #1063, opened 2020-03-02
Feature : Some optimizers added

sriyash421:
Added  the following function optimizers :
- Rprop
- SparseAdam
- HeavyBall


dhairyagandhi96:
Thank you for the contribution! Could you speak about precedent for having these optimisers available in other libraries?

sriyash421:
> Thank you for the contribution! Could you speak about precedent for having these optimisers available in other libraries?

Rprop: It takes care of vanishing and exploding gradients. And also an advantage of RProp is having a different step size for each weight. If one weight is already very close to its optimal value while a second weight still needs to be changed a lot, this is not a problem for RProp. Other variants can have many more problems with such a situation, especially because the gradient magnitudes can be misleading here.

SparseAdam : Effective for sparse gradients

HeavyBall: Reduces variation in gradients i.e the zigzagging trajectories during optimization.

bhvieira:
I think @dhairyagandhi96 meant if other frameworks/libraries have these too, like MXNet, PyTorch or TF/TF2, for example.

sriyash421:
> I think @dhairyagandhi96 meant if other frameworks/libraries have these too, like MXNet, PyTorch or TF/TF2, for example.

All of the frameworks have the first two but not the last one.

src/optimise/optimisers.jl
```diff
@@ -70,6 +70,49 @@ function apply!(o::Momentum, x, Δ)
...
+Rprop(α = 0.01, η = (0.5, 1.2), step_limit = (1e-6, 50.0)) = Rprop(α, η, step_limit, IdDict(), IdDict()) 
+
+function apply!(o::Rprop, x, Δ)
+  α, η = o.alpha, o.eta
+  step_limit = o.step_limit
+  state= get!(o.state, x, zero(Δ))::typeof(Δ) 
+  step = get!(o.step, x, one(x)*α)::typeof(x)
+  temp = Δ.*state
+  sign = oftype(temp,temp.>0)*η[2]+oftype(temp,temp.<0)*η[1]+oftype(temp,temp.==0)
+  signΔ= zero(Δ)+oftype(Δ,Δ.>0)-oftype(Δ,Δ.<0)
```
dhairyagandhi96, line 110: I think you can get rid of the oftype calls by using the appropriate eltype from the params or gradients to initialise scalar zeros

dhairyagandhi96:
Needs a rebase

src/optimise/optimisers.jl
```diff
@@ -70,6 +70,49 @@ function apply!(o::Momentum, x, Δ)
   @. Δ = -v
 end
 
+"""
+    Rprop(α, η, step_limit)
+
+Implements the resilient backpropagation algorithm.
+
+## Parameters
+    - Learning Rate (`α`): Amount by which gradients are discounted before updating the weights. Defaults to `0.01`.
```
CarloLucibello, line 79: everywhere else the learning rate is called `η`, can we change this?

CarloLucibello:
I think SparseAdams is meant to be work in conjunction with sparse arrays. As implemented here doesn't have much utility, please remove it from this PR

CarloLucibello:
Is HeavyBall implemented correctly? It doesn't seem to correspond to eq. 3 in
https://arxiv.org/pdf/1412.7457.pdf

src/optimise/optimisers.jl
sriyash421: Will do that.

sriyash421:
I have implemented the expanded equation. If we substitute p in the second equation it's same as eq3. of the paper.

sriyash421:
Extremely sorry but the above PR had some commits made directly into the master which messes things up a lot. So I had to hard reset the master and open a new PR from another branch. #1139

PR #1066, opened 2020-03-02
fix travis for documentation build

johnnychen94:
The previous build doesn't trigger the documentation stage because the matrix doesn't get expanded for the sole job.

Not very clear how Travis reads the config but this change fixes the issue.

😕 weird that it doesn't allow failures on nightly here... The one in my fork works as expected. https://github.com/johnnychen94/Flux.jl/runs/479502998

cc: @CarloLucibello

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,8 @@ julia:
   - 1.3
   - nightly
 
-matrix:
-  allow_failures:
-    - julia: nightly
+notifications:
+  email: false
 
 jobs:
   include:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,6 +23,5 @@ jobs:
         - julia --project=docs/ docs/make.jl
       after_success: skip
 
-## uncomment the following lines to override the default test script
-script:
- - julia --color=yes -e 'using Pkg; Pkg.activate(); Pkg.instantiate(); Pkg.test()'
```
CarloLucibello: this means we won't be using the current Manifest for testing though

CarloLucibello:
nice, thanks!

bors r+

johnnychen94:
* de40476beb059e35e9d99723e956d534b3ddc8f9 is the last change on traivs but seems that's unrelated @dhairyagandhi96 
* 6499344af397db698706b3325d2ba6831178ac65 is the first commit that doesn't get documentation triggered @CarloLucibello

This might be an upstream Travis issue. But I'll just stop here not investigate anymore.

PR #1068, opened 2020-03-03
[WIP] docs for functor.jl

CarloLucibello:


docs/src/gpu.md
```diff
@@ -38,6 +38,40 @@ m = fmap(cu, m)
...
+
+Now you can see that the inner fields of `actor` and `critic` are transformed into `CuArray`. So what does the `@functor` macro do here? Basically, it will create a function like this:
+
+```julia
+Flux.functor(m::ActorCritic) = (actor = m.actor, critic=m.critic), fields -> ActorCritic(fields...)
+```
+
+And the `functor` will be called recursively in `fmap`. As you can see, the result of `functor` contains two parts, a *destructure* part and a *reconstrucutre* part. The first part is to make the customized model structure into `trainable` data structure known to `Flux` (here is a `NamedTuple`). The goal is to turn `m` into `(actor=cu(ones(2,2)), critic=cu(ones(2)))`. The second part is to turn the result back into a `ActorCritic`, so that we can get `ActorCritic(cu(ones(2,2)),cu(ones(2)))`.
+
+By default, the `@functor` macro will transform all the fields in your customized structure. In some cases, you may only want to transform several fields. Then you just specify those fields manually like `Flux.@functor ActorCritic (actor,)` (note that the fields part must be a tuple). And make sure the `ActorCritic(actor)` constructor is also implemented.
```
johnnychen94:
```suggestion
By default, the `@functor` macro will transform all the fields in your customized structure. To transform only a subset of fields, you could just specify those fields manually. For example, `FLux.@functor ActorCritic (actor, )` (note that the fields part must be a tuple). In this case, the constructor `ActorCritic(actor)` should be defined, otherwise, a `MethodError` would be thrown later when you call `fmap`.

**Comment by dhairyagandhi96** [comment]
The docs still give me pause about whether I understand what these functions are doing. Maybe linking to what a functor is would be helpful. Also linking to the advanced section would be best, if we move it away from the GPU docs.

File: docs/src/gpu.md [hunk 2]
```diff
@@ -38,6 +38,40 @@ m = fmap(cu, m)
...
+ActorCritic{CuArray{Float32,2,Nothing},CuArray{Float32,1,Nothing}}(Float32[1.0 1.0; 1.0 1.0], Float32[1.0, 1.0])
+```
+
+Now you can see that the inner fields of `actor` and `critic` are transformed into `CuArray`. So what does the `@functor` macro do here? Basically, it will create a function like this:
+
+```julia
+Flux.functor(m::ActorCritic) = (actor = m.actor, critic=m.critic), fields -> ActorCritic(fields...)
+```
+
+And the `functor` will be called recursively in `fmap`. As you can see, the result of `functor` contains two parts, a *destructure* part and a *reconstrucutre* part. The first part is to make the customized model structure into `trainable` data structure known to `Flux` (here is a `NamedTuple`). The goal is to turn `m` into `(actor=cu(ones(2,2)), critic=cu(ones(2)))`. The second part is to turn the result back into a `ActorCritic`, so that we can get `ActorCritic(cu(ones(2,2)),cu(ones(2)))`.
```
cossio: `reconstrucutre` -> `reconstructure` ?

ToucheSir:
Now that functors have been moved out of Flux, there are a couple ways we could handle this:

1. Write the docs in Flux and link to them from Functors.jl
2. Write the docs in Functors and link to them from Flux (with a "learn more" link of sorts)
3. Write the core docs around generic structure traversal in Functors, and consolidate documentation (or link to it) on all related functionality (trainable, de/restructure, params) into a central location in Flux.

DhairyaLGandhi:
A mix of 2/3 is the way to go imo.

CarloLucibello:
Yeah, better have docs where they belong and have some leaner references here. 
Maybe I can add some part of this to Functors.jl's README

cossio:
Bump

docs/src/gpu.md
darsnack: Maybe `restructure`?

ToucheSir:
I think the docs and docstrings for everything that's still defined in Flux are useful here. `params`, `cpu` and `gpu` in particular. Everything else probably needs to go.

CarloLucibello:
closing as stale

PR #1070, opened 2020-03-03
Prevent breakage due to new `active` field in normalise layers

ianshmean:
Prevents breakage where the normalise structs, such as `BatchNorm`, have been directly defined but missing the new `active` field

cc. @darsnack

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.10.2"
+version = "0.10.3"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -80,9 +80,6 @@ mutable struct AlphaDropout{F}
   end
 end
 
-# TODO: deprecate in v0.11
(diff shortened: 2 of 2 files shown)


dhairyagandhi96:
Could you link the PR that added the new field?

darsnack:
Here: https://github.com/FluxML/Flux.jl/pull/1044#issuecomment-594049857

CarloLucibello:
thank you for this, that was an oversight. Could you bump the minor version so that we tag immediately? Could you also add a comment `# TODO: remove in v0.11` on top of those constructors?

CarloLucibello:
> thank you for this, that was an oversight. Could you bump the minor version so that we tag immediately? Could you also add a comment `# TODO: remove in v0.11` on top of those constructors?

actually `# TODO: deprecate in v0.11`  is better

ianshmean:
I'm seeing
```
WARNING: Method definition (::Type{Flux.AlphaDropout{F} where F})(Any) in module Flux at /home/ian/.julia/dev/Flux/src/layers/normalise.jl:78 overwritten at /home/ian/.julia/dev/Flux/src/layers/normalise.jl:84.
  ** incremental compilation may be fatally broken for this module **
```
Looking into it

ianshmean:
@CarloLucibello Ready for review

CarloLucibello:
looks good, thanks!

bors r+

PR #1073, opened 2020-03-06
WIP: Make optimize work on structs

Roger-luo:
So I tried to implement what was proposed in #637 
I guess this is what @MikeInnes wanted? But I didn't see why need `isimmutable` trait tho. With this PR, the following would work at least and a bit faster since it's type stable now.

```julia
using Flux
using Flux.Optimise: update!, ADAM

m = Chain(Dense(100, 100, tanh), Dense(100, 100, tanh))
opt = ADAM()
x = rand(Float32, 100)
loss(m, x) = sum(m(x))
∇m, _ = gradient(loss, m, x)
update!(opt, m, ∇m)
```

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -26,7 +26,7 @@ update!(opt, x::AbstractArray, x̄::Nothing) = nothing
 update!(opt, m::M, ∇m::Nothing) where M = nothing
 
 function update!(opt, x::AbstractArray, x̄)
-  x .-= apply!(opt, x::AbstractArray, x̄)
+  x .-= apply!(opt, x, x̄)
   return
 end
 
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -29,12 +29,6 @@ end
 
 update!(opt, m::M, ∇m::Nothing) where M = nothing
 
-function update!(opt, m::Tuple, ∇m::Tuple)
-  for (x, ∇x) in zip(m, ∇m)
-    update!(opt, x, ∇x)
-  end
-end
-
 # NOTE: since there won't be real loop in a struct
 #       we could always flatten it, which is a bit
 #       faster.
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -29,11 +29,22 @@ end
 
 update!(opt, m::M, ∇m::Nothing) where M = nothing
 
-function update!(opt, m::M, ∇m) where M
-  for each in fieldnames(M)
-    update!(opt, getfield(m, each), getfield(∇m, each))
+function update!(opt, m::Tuple, ∇m::Tuple)
+  for (x, ∇x) in zip(m, ∇m)
+    update!(opt, x, ∇x)
   end
-  return
+end
+
+# NOTE: since there won't be real loop in a struct
+#       we could always flatten it, which is a bit
+#       faster.
+@generated function update!(opt, m::M, ∇m) where M
+  body = Expr(:block)
+  for each in fieldnames(M)
+    each = QuoteNode(each)
+    push!(body.args, :(update!(opt, getfield(m, $each), getfield(∇m, $each))))
+  end
+  return body
 end
 
 function update!(opt, xs::Params, gs)
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -20,15 +20,15 @@ function update!(x::AbstractArray, x̄)
   return
 end
 
+# skip, if gradient is nothing
 update!(x::AbstractArray, x̄::Nothing) = nothing
 update!(opt, x::AbstractArray, x̄::Nothing) = nothing
+update!(opt, m::M, ∇m::Nothing) where M = nothing
 
 function update!(opt, x::AbstractArray, x̄)
   x .-= apply!(opt, x::AbstractArray, x̄)
   return
 end
-
-update!(opt, m::M, ∇m::Nothing) where M = nothing
 
 # NOTE: since there won't be real loop in a struct
 #       we could always flatten it, which is a bit
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -44,7 +44,6 @@ end
 
 function update!(opt, xs::Params, gs)
   for x in xs
-    gs[x] == nothing && continue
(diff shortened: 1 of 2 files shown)


dhairyagandhi96:
I think this is pretty close to what we were thinking of, just getting update to work with regular structs

Roger-luo:
So I guess should work on regular structs use `AbstractArray` as its parameters, but there could be things like `prelu` that contains one parameter, not sure if Flux has it somewhere. I think the way torch deal with training is just letting users define it (at least at its early stages). might also related to #666

src/optimise/train.jl
```diff
@@ -17,17 +17,42 @@ Update the array `x` according to `x .-= x̄`.
...
+update!(x::AbstractArray, x̄::Nothing) = nothing
+
+function update!(opt, x::AbstractArray, x̄)
+  x .-= apply!(opt, x::AbstractArray, x̄)
+  return
+end
+
+update!(opt, m::M, ∇m::Nothing) where M = nothing
+
+function update!(opt, m::Tuple, ∇m::Tuple)
```
CarloLucibello: Is this needed? Won't `@generated function update!(opt, m::M, ∇m) where M` cover also this case?

src/optimise/train.jl
```diff
@@ -17,17 +17,42 @@ Update the array `x` according to `x .-= x̄`.
...
   x .-= x̄
+  return
 end
 
-function update!(opt, x, x̄)
-  x .-= apply!(opt, x, x̄)
+update!(x::AbstractArray, x̄::Nothing) = nothing
+
+function update!(opt, x::AbstractArray, x̄)
+  x .-= apply!(opt, x::AbstractArray, x̄)
```
CarloLucibello: why the `::AbstractArray` annotation?

Roger-luo: or this will conflict with `update!(opt, m::M, ∇m) where M`

src/optimise/train.jl
Roger-luo: oh didn't notice that. yeah you are right.

src/optimise/train.jl
CarloLucibello: I meant here `apply!(opt, x::AbstractArray, x̄)`

MikeInnes:
The goal of #637 is not to support updating structs, but to support functional-style training loops; that's why the API proposed in that issue takes and returns a model and optimiser state rather than updating in-place. (That's also why we need something like an `isimmutable` trait: we want to do things in place as an optimisation, but need an automatic way to tell when that's possible, which it isn't for some array types.)

To meet that goal it's not enough to build on top of the current optimisers, since they use very-not-functional `IdDict`s internally which loses all of the properties we want.

Also, as a style note, we definitely shouldn't be writing generated functions here. `fmap` designed for exactly this kind of thing.

I think it'd be great to use this branch to hack out a design for this stuff, but just to point out that this is a fairly comprehensive rework with a few tricky design challenges involved.

Roger-luo:
Ah, @MikeInnes  I see what you mean, so actually this PR is not very related to #637 actually since it just changed the `update!` function, but #637 will need a rework on the entire optimizers structure with this new API IIUC.

src/optimise/train.jl
Roger-luo: oops, yeah I see.

Roger-luo:
`fmap` will make `update!` type unstable IIUC, since the cache is a `IdDict{Any, Any}`. If it's type unstable at least this will make AutoPreallocation hard to sort things out.

CarloLucibello:
How is this interacting with parameters sharing? e.g.
```julia
l = Dense(100, 100, tanh)
m = Chain(l, l)
....
```
We would have two ADAM steps, one for each partial derivative, instead of a single ADAM step with the sum of the partial derivates, right? This is not what we want

Roger-luo:
I think this simply just feed the optimizer the reference of the Array, (unless you are using some immutable structure to keep your parameters). 

Zygote will create a corresponding NamedTuple with new allocated memory, thus you can just feed them to different optimizers as before. I don't see why this won't work.

CarloLucibello:
Let's make a concrete example
```julia
julia> l = Dense(2,2,tanh)
Dense(2, 2, tanh)

julia> m = Chain(l,l)
Chain(Dense(2, 2, tanh), Dense(2, 2, tanh))

julia> loss(m, x) = sum(m(x))
loss (generic function with 1 method)

julia> x = rand(Float32, 2)
2-element Array{Float32,1}:
 0.6158823
 0.3728192

julia> ∇m, _ = gradient(loss, m, x)
((layers = ((W = Float32[0.94506115 0.57208484; -0.76053166 -0.46038148], b = Float32[1.5344834, -1.2348653], σ = nothing), (W = Float32[0.13190015 0.296789; 0.1380123 0.310542], b = Float32[0.9556952, 0.99998146], σ = nothing)),), Float32[0.4527306, -1.2399194])

julia> ∇m.layers[1].W
2×2 Array{Float32,2}:
  0.945061   0.572085
 -0.760532  -0.460381

julia> ∇m.layers[2].W
2×2 Array{Float32,2}:
 0.1319    0.296789
 0.138012  0.310542
```
Let's call `W` the shared weight matrix. According to this PR, a call to `update!(opt, m, ∇m)` will result into two calls to apply:
```julia
W .-= apply!(opt, W, ∇m.layers[1].W)
W .-= apply!(opt, W, ∇m.layers[2].W)
```
This is wrong for optimizers with internal state. The correct behavior is 
```julia
W .-= apply!(opt, W, ∇m.layers[1].W + ∇m.layers[2].W)
```

Roger-luo:
True. I see your point, it seems we need to store the mapping between parameter reference and its corresponding gradient somehow. Only struct gradient seems not sufficient. I don't have a good idea yet, let me think about it too.

MikeInnes:
This isn't really any more type-stable, it's just that the IdDict is hidden inside the current optimiser objects (which results in the weird interactions Carlo points out). Part of the "why not?" is that `fmap` handles subtleties like DAG-structures in a correct way, and we can pretty easily fix the cache type stability for immutable objects.

ToucheSir:
@Roger-luo can we close this in favour of Optimisers.jl?

PR #1079, opened 2020-03-13
fixed docstrings in the module

Dsantra92:
the docstrings in the module for functions onhot and onecold  proposed a function with some examples that returns 0 or 1 instead of boolean .

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -56,15 +56,15 @@ julia> using Flux: onehot
 
 julia> onehot(:b, [:a, :b, :c])
 3-element Flux.OneHotVector:
- 0
- 1
- 0
+ false
+ true
+ false
 
 julia> onehot(:c, [:a, :b, :c])
 3-element Flux.OneHotVector:
- 0
- 0
- 1
+ false
+ false
+ true
 ```
 """
 function onehot(l, labels)
@@ -92,9 +92,9 @@ julia> using Flux: onehotbatch
 
 julia> onehotbatch([:b, :a, :b], [:a, :b, :c])
 3×3 Flux.OneHotMatrix{Array{Flux.OneHotVector,1}}:
- 0  1  0
- 1  0  1
- 0  0  0
+ false   true  false
+  true  false   true
+ false  false  false
 ```
 """
 onehotbatch(ls, labels, unk...) =


CarloLucibello:
docstring seems correct though
```julia
julia> onehot(:b, [:a, :b, :c])
3-element Flux.OneHotVector:
 0
 1
 0
```

Dsantra92:
I thought it should return true and false instead of 0 and 1s.

bhvieira:
Docs are conflicting on this right now. Going by `src/data/onehot.md`, you are right. Ping @janEbert

@CarloLucibello the code for `OneHotVector` implies it returns booleans, and it is true on my Flux (v0.10.2).

janEbert:
Hey, thanks for the ping! You are absolutely right, the docstrings are wrong, but in the inverted way, as @bhvieira said.
They mention the function would return a vector with `true` at the `l`-th position which is incorrect. I actually find I rewrote that docstring very badly; due to calling it the `l`-th position and the first argument also being named `l`, it is very confusing. It should rather be written with the `k`-th position, for example. The docstring in general does not really explain what the function does to a newbie so it could use a better, newer one.

Finally, @bhvieira mentioned [docs/src/data/onehot.md](docs/src/data/onehot.md) which also needs to be updated to use `0` and `1` instead of `false` and `true`. I'd recommend making those docstrings doctests as well by placing `jldoctest` after the initial backticks instead of `julia` or nothing, like this:

    ```jldoctest
    julia> onehot(:a, [:a, :b, :c])
    [...]
    ```

That way, these kinds of mistakes should not happen. Sorry I made them.

janEbert:
Hey @Dsantra92, friendly ping in case you forgot about this.

Dsantra92:
Sorry I really got busy with college work and thank you for the reminder @janEbert .I will be working on it from today.

janEbert:
No need to apologize at all. :)
I just don't want to confuse more people with my mistake. So, thanks that you're going to fix it!

Dsantra92:
@janEbert I have commited the changes to one hot function that you requested.

janEbert:
Thank you so much!

Dsantra92:
Already merged in #1256

PR #1082, opened 2020-03-14
There were some problem docstrings in onehot.jl

Dsantra92:
In the file onehot.jl the functions onehot ,onecold etc. had docstrings which asserts that the functions will return 0's and 1's instead of true and false,I have fixed this docstring.


CarloLucibello:
why are you opening the same PR as #1079 ? please don't make unnecessary noise in the repo

PR #1085, opened 2020-03-14
Tell people to stop using global models

oxinabox:
People keep doing this and it is bad.
Having it in the performance tips should at at least help.

But throughout the docs we violate this rule nearly constantly.
My bad example was basically copied straight from https://fluxml.ai/Flux.jl/stable/training/training/#Loss-Functions-1

But this PR is just going to add it to the performance tips.
Its beyond scope of this PR to rewrite every example that violates this

### DIFF
--- a/docs/src/performance.md
+++ b/docs/src/performance.md
@@ -4,6 +4,59 @@ All the usual [Julia performance tips apply](https://docs.julialang.org/en/v1/ma
 As always [profiling your code](https://docs.julialang.org/en/v1/manual/profile/#Profiling-1) is generally a useful way of finding bottlenecks.
 Below follow some Flux specific tips/reminders.
 
+## Don't write loss functions that use a non-constant globally declared model.
+This is a special case of one of the most important [Julia Performance Tips](https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-global-variables-1).
+Non-constant global variables are slow.
+We repeat it here as it is a common mistake.
+
+This advice is appliable also to writing callbacks, and more generally to all Julia code.
+
+### Don't write:
+```julia
+data = ...
+m = Chain(Dense(784, 32, σ), Dense(32, 10), softmax)
+loss(x, y) = Flux.mse(m(x), y)
+
+Flux.train!(loss, Flux.params(m), data, Descent(0.1))
+```
+In this bad code, the model `m` is a non-constant global.
+It is being used inside the function `loss`, which is one of the most performance critical parts of this code.
+It will be slow, as the compiler can't rely on `m` always being the same type -- it is a mutable global, it could change at any time.
+
+### Correct alternatives:
+#### Mark the model `const`
+```julia
+data = ...
+const m = Chain(Dense(784, 32, σ), Dense(32, 10), softmax)
+loss(x, y) = Flux.mse(m(x), y)
+
+Flux.train!(loss, Flux.params(m), data, Descent(0.1))
+```
+Similarly anything else that is a non-constant global that is used in functions should also be made constant
+
+#### Put everything in a main function:
+For more flexibility, you could even make this take `m` as a argument -- it doesn't matter of `m` was originally declared as a non-const global once it has been passed in as a argument because it then becomes a local variable.
+```julia
+function main(data)
+    m = Chain(Dense(784, 32, σ), Dense(32, 10), softmax)
+    loss(x, y) = Flux.mse(m(x), y)
+
+    Flux.train!(loss, Flux.params(m), data, Descent(0.1))
+end
+```
+
+#### Make the loss function actually close over `m`.
+Closures can be very useful.
+
+```julia
+data = ...
+m = Chain(Dense(784, 32, σ), Dense(32, 10), softmax)
+get_loss_function(mdl) = (x, y) -> Flux.mse(mdl(x), y)
+
+Flux.train!(get_loss_function(m), Flux.params(m), data, Descent(0.1))
+```
+This example is particularly applicable to callbacks.
+
 ## Don't use more precision than you need
 
 Flux works great with all kinds of number types.


docs/src/performance.md
```diff
@@ -4,6 +4,59 @@ All the usual [Julia performance tips apply](https://docs.julialang.org/en/v1/ma
...
+function main(data)
+    m = Chain(Dense(784, 32, σ), Dense(32, 10), softmax)
+    loss(x, y) = Flux.mse(m(x), y)
+
+    Flux.train!(loss, Flux.params(m), data, Descent(0.1))
+end
+```
+
+#### Make the loss function actually close over `m`.
+Closures can be very useful.
```
johnnychen94:
```suggestion
[Closures](https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-captured-1) can be very useful.
```

docs/src/performance.md
```diff
@@ -4,6 +4,59 @@ All the usual [Julia performance tips apply](https://docs.julialang.org/en/v1/ma
...
+data = ...
+const m = Chain(Dense(784, 32, σ), Dense(32, 10), softmax)
+loss(x, y) = Flux.mse(m(x), y)
+
+Flux.train!(loss, Flux.params(m), data, Descent(0.1))
+```
+Similarly anything else that is a non-constant global that is used in functions should also be made constant
+
+#### Put everything in a main function:
+For more flexibility, you could even make this take `m` as a argument -- it doesn't matter of `m` was originally declared as a non-const global once it has been passed in as a argument because it then becomes a local variable.
```
CarloLucibello:
```suggestion
For more flexibility, you could even make this take `m` as a argument -- it doesn't matter if `m` was originally declared as a non-const global once it has been passed in as a argument because it then becomes a local variable.
```

dhairyagandhi96:
Do we link to https://docs.julialang.org/en/v1/manual/performance-tips/ appropriately too?

docs/src/performance.md
```diff
@@ -4,6 +4,59 @@ All the usual [Julia performance tips apply](https://docs.julialang.org/en/v1/ma
...
+### Correct alternatives:
+#### Mark the model `const`
+```julia
+data = ...
+const m = Chain(Dense(784, 32, σ), Dense(32, 10), softmax)
+loss(x, y) = Flux.mse(m(x), y)
+
+Flux.train!(loss, Flux.params(m), data, Descent(0.1))
+```
+Similarly anything else that is a non-constant global that is used in functions should also be made constant
```
bhvieira: Perhaps explicitly call it data? Or inputs and outputs. That way most bases are covered, I can't imagine other things being used in forward passes other than models and data.

bhvieira:
I'll be frank here, I didn't know declaring the models constant helped, because the parameters are type-stable. I declare models inside functions, so I'm not hit by this when training, but I've been running some post-hoc analysis on the trained models and this should benefit me when loading trained models into memory, right? I'll try to benchmark this later.

bhvieira:
```
using Flux
using BenchmarkTools

m = Chain(Dense(784, 32, σ), Dense(32, 10), softmax);
const const_m = Chain(Dense(784, 32, σ), Dense(32, 10), softmax);
loss(x, y) = Flux.mse(m(x), y);
const_loss(x, y) = Flux.mse(const_m(x), y);
x, y = (randn(Float32, 784, 100), softmax(randn(Float32, 10, 100)));
const const_x, const_y = (randn(Float32, 784, 100), softmax(randn(Float32, 10, 100)));

#tests without `$` interpolation, I guess it makes sense since we want to measure the effect of globals
@benchmark loss(x, y)
```

>BenchmarkTools.Trial:
>memory estimate:  46.88 KiB
>allocs estimate:  20
>
>minimum time:     228.200 μs (0.00% GC)
>median time:      249.301 μs (0.00% GC)
>mean time:        279.954 μs (1.77% GC)
>maximum time:     5.604 ms (94.28% GC)
>
>samples:          10000
>evals/sample:     1 

```
@benchmark loss(const_x, const_y)
```

>BenchmarkTools.Trial:
>memory estimate:  46.88 KiB
>allocs estimate:  20
>
>minimum time:     227.500 μs (0.00% GC)
>median time:      242.500 μs (0.00% GC)
>mean time:        266.806 μs (1.45% GC)
>maximum time:     4.452 ms (94.59% GC)
>
>samples:          10000
>evals/sample:     1

```
@benchmark const_loss(x, y)
```

>BenchmarkTools.Trial:
>memory estimate:  46.88 KiB
>allocs estimate:  20
>
>minimum time:     228.600 μs (0.00% GC)
>median time:      241.200 μs (0.00% GC)
>mean time:        272.915 μs (1.30% GC)
>maximum time:     3.870 ms (92.49% GC)
>
>samples:          10000
>evals/sample:     1

```
@benchmark const_loss(const_x, const_y)
```

>BenchmarkTools.Trial:
>memory estimate:  46.86 KiB
>allocs estimate:  19
>
>minimum time:     230.500 μs (0.00% GC)
>median time:      242.700 μs (0.00% GC)
>mean time:        280.045 μs (0.89% GC)
>maximum time:     3.766 ms (0.00% GC)
>
>samples:          10000
>evals/sample:     1

jondeuce:
Echoing @bhvieira's thoughts, I'm not seeing much of a performance hit in practice, with the exception of very small models. Consider the following examples:

```julia
m_large = Flux.Chain(Flux.Dense(768,1024), Flux.Dense(1024,10), Flux.softmax)
m_small = Flux.Chain(Flux.Dense(1,5), Flux.Dense(5,2), Flux.softmax)
const const_m_large = deepcopy(m_large)
const const_m_small = deepcopy(m_small)

loss_large(x,y) = Flux.mse(m_large(x), y)
loss_small(x,y) = Flux.mse(m_small(x), y)
grad_large(x,y) = Flux.gradient(() -> loss_large(x,y), Flux.params(m_large))
grad_small(x,y) = Flux.gradient(() -> loss_small(x,y), Flux.params(m_small))
loss_const_large(x,y) = Flux.mse(const_m_large(x), y)
loss_const_small(x,y) = Flux.mse(const_m_small(x), y)
grad_const_large(x,y) = Flux.gradient(() -> loss_const_large(x,y), Flux.params(const_m_large))
grad_const_small(x,y) = Flux.gradient(() -> loss_const_small(x,y), Flux.params(const_m_small))

makexy(in, out, batch) = rand(Float32, in, batch), rand(Float32, out, batch)

@btime loss_large(x,y)       setup = ((x,y) = makexy(768, 10, 1024)); # 6.491 ms (19 allocations: 8.20 MiB)
@btime loss_const_large(x,y) setup = ((x,y) = makexy(768, 10, 1024)); # 6.515 ms (18 allocations: 8.20 MiB)

@btime grad_large(x,y)       setup = ((x,y) = makexy(768, 10, 1024)); # 18.681 ms (161 allocations: 18.54 MiB)
@btime grad_const_large(x,y) setup = ((x,y) = makexy(768, 10, 1024)); # 18.459 ms (157 allocations: 18.54 MiB)

@btime loss_small(x,y)       setup = ((x,y) = makexy(1, 1, 1)); # 1.059 μs (12 allocations: 944 bytes)
@btime loss_const_small(x,y) setup = ((x,y) = makexy(1, 1, 1)); # 1.021 μs (11 allocations: 928 bytes)

@btime grad_small(x,y)       setup = ((x,y) = makexy(1, 1, 1)); # 15.880 μs (133 allocations: 5.81 KiB)
@btime grad_const_small(x,y) setup = ((x,y) = makexy(1, 1, 1)); # 12.370 μs (129 allocations: 5.72 KiB) <--- Performance hit
```

Interestingly, the ~3 μs performance hit in the gradient of the small model is fairly robust to changes in model size.

To be honest, I'm not exactly sure of the point I'm trying to make; of course, declaring models as `const` is strictly better than not, and reinforcing this in the documentation (especially considering it is good general Julia practice) is a good idea. Perhaps, however, it can be noted that for large computations it is not likely to seriously denigrate performance, as the internal model parameters themselves are type stable etc.

bhvieira:
@jondeuce Yeah, I think that's the takehome message here. The microsecond difference might be due to that few allocations we get without `const`

CarloLucibello:
given the above benchmarks, the tone here is too alarmistic and should be moderated

PR #1091, opened 2020-03-19
Docstrings for pad, stride, dilation

saswatpp:
Added Docstrings for expected argument types for pad, stride and dilation in conv layers


saswatpp:
Sorry there was a git issue from my part

Issue #1077, opened 2020-03-10
Documentation of `pad` keyword


wesselb:
If specified as a tuple, the `pad` keyword seems to require two integers per dimension, one specifying the padding on either side. It would be good to document this behaviour.

The following behaviour is confusing:

```julia
julia> size(ConvTranspose(ones(3,1,1), zeros(1); pad=2)(ones(5,1,1)))
(3, 1, 1)

julia> size(ConvTranspose(ones(3,1,1), zeros(1); pad=(2,2))(ones(5,1,1)))
(3, 1, 1)

julia> size(ConvTranspose(ones(3,1,1), zeros(1); pad=(2,))(ones(5,1,1)))
ERROR: MethodError: no method matching DenseConvDims(::Tuple{Int64,Int64}, ::Tuple{Int64,Int64,Int64}; stride=(1,), padding=(2,), dilation=(1,))
Closest candidates are:
  DenseConvDims(::Tuple{Vararg{T,M}} where T, ::Tuple{Vararg{T,M}} where T; stride, padding, dilation, flipkernel) where M at /Users/wessel/.julia/packages/NNlib/FAI3o/src/dim_helpers/DenseConvDims.jl:22
Stacktrace:
 [1] conv_transpose_dims(::ConvTranspose{1,1,typeof(identity),Array{Float64,3},Array{Float64,1}}, ::Array{Float64,3}) at /Users/wessel/.julia/packages/Flux/dkJUV/src/layers/conv.jl:112
 [2] (::ConvTranspose{1,1,typeof(identity),Array{Float64,3},Array{Float64,1}})(::Array{Float64,3}) at /Users/wessel/.julia/packages/Flux/dkJUV/src/layers/conv.jl:122
 [3] top-level scope at REPL[58]:1
```

PR #1093, opened 2020-03-20
docstring for pad, stride, dilation

saswatpp:
Added Docstrings for expected argument types for pad, stride and dilation in conv layers


src/layers/conv.jl
```diff
@@ -100,6 +104,10 @@ In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 
 Takes the keyword arguments `pad`, `stride` and `dilation`.
+For Input dimension N, 
+`pad` should be a single Integer or a tuple of with 2*(N-2) parameters indicating
```
CarloLucibello: of with

src/layers/conv.jl
```diff
@@ -27,6 +27,10 @@ In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 
 Takes the keyword arguments `pad`, `stride` and `dilation`.
+For Input dimension N,
```
CarloLucibello: input

src/layers/conv.jl
saswatpp: fixed

src/layers/conv.jl
saswatpp: fixed

src/layers/conv.jl
```diff
@@ -27,6 +27,10 @@ In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 
 Takes the keyword arguments `pad`, `stride` and `dilation`.
+For input dimension N, 
+`pad` should be a single Integer or a tuple with 2*(N-2) parameters indicating
```
CarloLucibello:
```suggestion
`pad` should be a single Integer or a tuple of length 2*(N-2) indicating 
```

src/layers/conv.jl
```diff
@@ -27,6 +27,10 @@ In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 
 Takes the keyword arguments `pad`, `stride` and `dilation`.
```
CarloLucibello: can you add these keywords to the function signature on the top of this docstring?

src/layers/conv.jl
CarloLucibello: Also, maybe better use something like "for D spatial dimensions, `pad` should be a single Integer or a tuple of length 2*D"

src/layers/conv.jl
bhvieira: I think they were there already? in `src/layers/conv.jl` the signature already reads (correctly):
```
    Conv(size, in => out, σ = identity; init = glorot_uniform,
         stride = 1, pad = 0, dilation = 1)
```

src/layers/conv.jl
```diff
@@ -27,6 +27,10 @@ In other words, a 100×100 RGB image would be a `100×100×3×1` array,
...
+For input dimension N, 
+`pad` should be a single Integer or a tuple of length 2*(N-2) indicating
```
bhvieira, line 31: Not entirely true, pad can be a `N-2`-D `Tuple` as well, that would apply symmetric padding.

bhvieira:
This PR should close #1077

dhairyagandhi96:
Bump. What's the status on this?

src/layers/conv.jl
CarloLucibello: they are not in the docstring's signature:
```
"""
    Conv(size, in=>out)
    Conv(size, in=>out, relu)

Standard convolutional layer. `size` should be a tuple like `(2, 2)`.
```

src/layers/conv.jl
CarloLucibello: good catch

src/layers/conv.jl
bhvieira: They are though; actually it's the sole signature presented.

bhvieira: I think you are not in `master`, prior to #853 being merged

src/layers/conv.jl
saswatpp: made changes in new commit

Issue #1094, opened 2020-03-21
Cosine_embedding_loss could be added to Flux.jl


ven-k:
Adding **Cos_embedding_loss** to Flux.jl could be useful.

This can be used for,
- Semi-supervised learning
- Learning nonlinear embeddings
-  For finding similarity or dissimilarity of 2 inputs.

PR #1095, opened 2020-03-21
Added Cosine embedding loss

ven-k:
Cosine embedding loss could be useful in Flux.jl for
- Semi-supervised learning
- Learning nonlinear embeddings
- For finding similarity or dissimilarity of 2 inputs.

**cos-embedding_loss( x1, x2, y, margin, pad = false)** 
It takes, 
- x1 and x2 as inputs
- y = 1 or -1 as mode
- margin = 0 which can range from -1 to 1 and has 0 as its default value
- pad has default value 'false'. If pad = true is passed zeros are padded and the cos_embedding_loss() is calculated. It requires PaddedViews.jl to be added. 

This PR resolves issue #1094

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -210,3 +210,51 @@ by linearizing all values for each element in the batch.
 function flatten(x::AbstractArray)
   return reshape(x, :, size(x)[end])
 end
+"""
+cos_embedding_loss takes
+ x1 and x2 as inputs
+ y = 1 or -1 as mode
+ margin = 0 which can range from -1 to 1 and has 0 as its default value
+ pad has default value 'false'. If pad = true is passed zeros are padded and the cos_embedding_loss() is calculated. It requires PaddedViews.jl to be added.
+"""
+function cos_theta(x1::Array{Float32,N}, x2::Array{Float32,N}) where N
+    return  dot(x1,x2) / (norm(x1) * norm(x2))
+end
+
+function cos_embedding_loss(x1::AbstractMatrix, x2::AbstractMatrix, y::Any; margin=0, pad = false)
+    @assert (margin <= 1 && margin >= -1)
+    @assert y ∈ (1,-1)
+    if (pad == false && (size(x1) != size(x2)))
+        throw(DimensionMismatch("If you wish to calculate the loss by padding zeros, pass 'pad = true' as an arguement and add PaddedView.jl to the environment"))
+    elseif (pad == true && (size(x1) != size(x2)))
+        size(x1) > size(x2) ? x2 = PaddedView(0.0, x2, size(x1)) : x1 = PaddedView(0.0, x1, (size(x2)))
+    end
+
+    x1 = Float32.(x1); x2 = Float32.(x2); y = Int64(y); margin = Float32(margin)
+
+    cos_embedding_loss(x1,x2,y,margin)
+end
+
+function cos_embedding_loss(x1::AbstractArray, x2::AbstractArray, y::Any; margin=0, pad = false)
+    @assert (margin <= 1 && margin >= -1)
+    @assert y ∈ (1,-1)
+    if (pad == false && (size(x1) != size(x2)))
+        throw(DimensionMismatch("If you wish to calculate the loss by padding zeros, pass 'pad = true' as an arguement and add PaddedView.jl to the environment"))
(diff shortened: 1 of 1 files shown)


src/layers/stateless.jl
```diff
@@ -210,3 +210,53 @@ by linearizing all values for each element in the batch.
...
   return reshape(x, :, size(x)[end])
 end
+
+"""
+cos_embedding_loss takes
+ x1 and x2 as inputs
+ y = 1 or -1 as mode
+ margin = 0 which can range from -1 to 1 and has 0 as its default value
+ pad has default value 'false'. If pad = true is passed zeros are padded and the cos_embedding_loss() is calculated. It requires PaddedViews.jl to be added.
+"""
```
CarloLucibello: docstring should be placed just above the corresponding function and formatted as other losses' docstrings

src/layers/stateless.jl
```diff
@@ -210,3 +210,53 @@ by linearizing all values for each element in the batch.
...
+function cos_embedding_loss(x1::AbstractArray, x2::AbstractArray, y::Any; margin=0, pad = false)
+    @assert (margin <= 1 && margin >= -1)
+    @assert y ∈ (1,-1)
+    if (pad == false && (size(x1) != size(x2)))
+        throw(DimensionMismatch("If you wish to calculate the loss by padding zeros, pass 'pad = true' as an arguement and add PaddedView.jl to the environment"))
+    elseif (pad == true && (size(x1) != size(x2)))
+        size(x1) > size(x2) ? x2 = PaddedView(0.0, x2, size(x1)) : x1 = PaddedView(0.0, x1, (size(x2)))
+    end
+
+    x1 = Float32.(x1); x2 = Float32.(x2); y = Int64(y); margin = Float32(margin)
```
CarloLucibello: you should not cast types

ven-k: Instead, should I print that inputs mismatch hence the cosine distance is maximum or can't be computed? Or is there anything else?

src/layers/stateless.jl
ven-k: `"""
 cos_embedding_loss(x1, x2, y; margin=0, pad=false)

 Finds the cosine distance between `x1` and `x2` matrices for `y` = 1 or -1.
 `margin` can take values from [-1,1].
"""
function cos_embedding_loss(x1::Array{Float32,N}, x2::Array{Float32,N}, y::Int64, margin) where N
    @assert (margin <= 1 && margin >= -1)
    @assert y ∈ (1,-1)
`...

Does this docstring follow the format?

Also, now I feel having AbstractArray is sufficient. (i.e I should remove the block for Abstract matrix) Or maybe, as mentioned below I think I'll remove even AbstractArray block and check for type similarity and compute everything in the main block itself.

ven-k:
I've asked users to add PaddedViews only if they want padded results necessary. The reason is: Padding doesn't happen if one input is a vector and the other is a matrix. How should I proceed?

src/layers/stateless.jl
bhvieira: It should be up to the user to certify that his inputs are the correct types, imo. That's the stance elsewhere in Flux.

ven-k: Yes. I have done the changes, and just like other functions, I'll test the format in ```test``` files.

```x1 = [1.0 2.0 3.0; 4.0 5.0 6.0]
  x2 = [0.1 0.2 0.3; 0.4 0.5 0.6]
  y = [1,-1]
  @testset "cos_embedding_loss" begin
    @test cos_embedding_loss(x1, x2, y[1]) ≈ 0.0
    @test cos_embedding_loss(x1, x2, y[1], margin=0.5) ≈ 0.0
    @test cos_embedding_loss(x1, x2, y[2], margin=-1) ≈ 2.0
  end


  @testset "no spurious promotion for cos_embedding_loss" begin
    for T in (Float32, Float64)
      x1 = rand(T,4)
      x2 = rand(T,4)
      for y in (-1,1)
        fwd, back = Flux.pullback(cos_embedding_loss, x1, x2, y)
        @test typeof(fwd) <: AbstractFloat
        @test eltype(back(one(T))[1]) <: AbstractFloat
      end
    end
  end
```
There is an issue: The Project.toml doesn't get updated with "LinearAlgebra". I've raised the issue in Slack. 
Once resolved, can I commit this?

PR #1098, opened 2020-03-25
Allow CuArrays v2.x

ararslan:


dhairyagandhi96:
bors try

dhairyagandhi96:
Could you also bump the version in the Manifest please? Thanks @ararslan!

ararslan:
Done!

ararslan:
Hm, I'm not sure what's going on with the Travis failure but it does look related. I guess it's expecting a CUDA installation, which is why it worked on the GPU-enabled Gitlab CI.

ararslan:
The amazing @maleadt found that the source of the issue is with CUDAdrv. He'll be addressing that (with tests to catch it) upstream soon.

ararslan:
Ah, looks like this will need https://github.com/JuliaGPU/CuArrays.jl/pull/650 too.

maleadt:
https://github.com/JuliaRegistries/General/pull/11588. I think this will also need https://github.com/FluxML/Flux.jl/pull/1050 though.

CarloLucibello:
@maleadt we should drop cuarrays 1.6 support if #1050 gets in, right?

dhairyagandhi96:
Merged #1050 here, could use a manifest bump

maleadt:
> @maleadt we should drop cuarrays 1.6 support if #1050 gets in, right?

Correct.

dhairyagandhi96:
Is it limited to 1.6 or would we also drop 1.7 support here?

maleadt:
> Is it limited to 1.6 or would we also drop 1.7 support here?

Sorry, you also need to drop 1.7. The change is only in 2.0.

maleadt:
bors try

dhairyagandhi96:
Thanks guys!

jgreener64:
Patch release would be really useful.

PR #1101, opened 2020-03-26
Implement APIs of freeze parameters and freeze layers

### LINKED ISSUE #1022: freeze parameters
I guess this came out already, but I couldn't find the corresponding issue. 
The question is: how do I freeze (i.e. no gradient computations and updates) some specific parameter in a Flux model?
The best solution that comes to my mind is to implement a dictionary-like `delete!` method for the `Zygote.Params` type, so that freezing the bias term in the first layer of my network would look like this:
```julia
opt = ADAM(η)
ps = params(model)
delete!(ps, model.layers[1].bias)  # <-- Proposed Solution: implement delete! for Zygote.Params
gs = Flux.gradient(ps) do
	ŷ = model(x)
        loss(ŷ, y)
end
```
Should I make a PR to Zygote? Or is there something better we can do?

foldfelis:
This fix #1022


dhairyagandhi96:
Also check out the docs here https://fluxml.ai/Flux.jl/stable/models/advanced/#Freezing-Layer-Parameters-1

foldfelis:
Oh, I didn't notice that the `Zygote.delete!` function has been import to Flux already

foldfelis:
Due to the `delete!` function works like charm, I removed the duplicate `freezeparam!` function

foldfelis:
Sorry I pressed the close button by mistake.

CarloLucibello:
we already this kind of interface, since for chains we already have `params(m[[1,3]])`

PR #1103, opened 2020-03-28
Bump version to 0.10.4

ararslan:
Since 0.10.3, the only changes have been dependency compatibility bumps, new features, and documentation improvements. See https://github.com/FluxML/Flux.jl/compare/v0.10.3...master.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.10.3"
+version = "0.10.4"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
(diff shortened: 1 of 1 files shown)


dhairyagandhi96:
Actually, since cuarrays 2.x is julia 1.3 and up, we'll need to drop 1.0 support to update this one...

ararslan:
Seems Flux isn't being tested on 1.0 anyway: https://github.com/FluxML/Flux.jl/blob/master/.travis.yml#L9. And I believe there was a change to package registration that allows bumping the patch when dropping a Julia version. I'd personally vote for dropping 1.0, but of course this is a decision for the Flux maintainers. 🙂

ararslan:
Actually, Flux as-is won't work on 1.0 anyway; it seems to be using some features introduced in later minor releases, e.g. `isnothing` [here](https://github.com/FluxML/Flux.jl/blob/master/src/layers/normalise.jl#L57).

dhairyagandhi96:
Fair cop, we should at least set the Julia version correctly, could you bump the Julia version required?

ararslan:
Done. Set to 1.3 to reflect the updated requirement due to the CuArrays dependency.

ararslan:
Also have a companion PR to Zygote to update it for CuArrays 2: https://github.com/FluxML/Zygote.jl/pull/565

PR #1106, opened 2020-03-30
Added margin_ranking_loss

ven-k:
I would like to **add margin_ranking_loss(x1, x2, y; margin=0, mode=identity, pad=false)** to Flux. (counterpart exists in Pytorch)

loss = max( 0, -y * (x1 - x2) + margin)

Measures the margin ranking loss between x1, x2 for y = 1 or -1. mode can be additionally specified as mean, sum.

This can be used for ranking problems and GANs.


src/layers/stateless.jl
```diff
@@ -210,3 +211,20 @@ by linearizing all values for each element in the batch.
...
+
+Measures the margin ranking loss between `x1`, `x2` for `y` = 1 or -1 and `margin`. `mode` can be additionally specified as mean, sum.
+
+```math
+\text{loss}(x, y) = \max(0, -y * (x1 - x2) + \text{margin})
+```
+"""
+
+function margin_ranking_loss(x1, x2, y; margin=zero(eltype(x1)), mode=identity, pad=false)
+  if (y ∉ (1,-1)) error("Mode can be either 1 or -1, got $y") end
```
dhairyagandhi96: Could use something like `y... || error...`

src/layers/stateless.jl
```diff
@@ -1,4 +1,5 @@
 # Cost functions
+using Statistics, PaddedViews
```
dhairyagandhi96: Is paddedviews a dependency already?

dhairyagandhi96:
Will need correctness tests

src/layers/stateless.jl
ven-k: No. It is not. But I can't add any file to the dependencies of Flux.jl. But I've tested this by adding it in the global scope.

src/layers/stateless.jl
ven-k: Ok. I will change it.

ven-k:
I've written this test.
```
x1 = [1.2 2.3 3.4]
x2 = [9.8 8.7 7.6]
y = 1
@testset "margin_ranking_loss" begin
    @test margin_ranking_loss(x1, x2, y) ≈ [8.6 6.4 4.2]
    @test margin_ranking_loss(x1, x2, y, margin=1.0) ≈ [9.6 7.4 5.2]
    @test margin_ranking_loss(x1, x2, y, margin=1.0, mode=mean) ≈ 7.4
    @test margin_ranking_loss(x1, x2, y, margin=1.0, mode=sum) ≈ 22.2
end
```

The test for "no spurious promotion" gives "Dimension mismatch" for
 ```
@test eltype(back(one(T))[1]) == T
 ```
I'll debug it and commit the test file.

dhairyagandhi96:
I don't think we should add PaddedViews as a dependency, considering it's not come up often and we have a pretty sizable dependency tree already.

ven-k:
I will remove the padding option then.

src/layers/stateless.jl
```diff
@@ -210,3 +211,13 @@ by linearizing all values for each element in the batch.
...
   return reshape(x, :, size(x)[end])
 end
+
+"""
+  margin_ranking_loss(x1, x2, y; margin=0, mode=identity)
+
+Measures the margin ranking loss between `x1`, `x2` for `y` = 1 or -1 and `margin`. `mode` can be additionally specified as mean, sum.
+"""
+function margin_ranking_loss(x1, x2, y; margin=zero(eltype(x1)), mode=identity)
+  y ∈ (1, -1) || error("Mode can be either 1 or -1, got $y")
```
ToucheSir, line 221: Should this error mention y instead of Mode?

ToucheSir:
@ven-k looks like all the outstanding review feedback was addressed, would you still like to push this through?

ven-k:
API has changed since the previous commit. I will modify and push a commit

ToucheSir (commented):
Do you mind adding a test in https://github.com/FluxML/Flux.jl/blob/master/test/cuda/losses.jl as well? Otherwise I think this is good to go, barring any objections on the calculation itself.

test/losses.jl
```diff
@@ -206,3 +206,24 @@ end
...
+
+@testset "no spurious promotions for margin_ranking_loss" begin
+  for T in (Float32, Float64)
+    x1 = rand(T, 2)
+    x2 = rand(T, 2)
+    for y in (1,-1)
+      fwd, back = Flux.pullback(Flux.margin_ranking_loss, x1, x2, y)
+      @test eltype(fwd) == T
+    end
+  end
```
ToucheSir, line 228: Seems this is not necessary if you add `margin_ranking_loss` to `ALL_LOSSES` in https://github.com/FluxML/Flux.jl/blob/v0.12.6/test/losses.jl#L8. Then it'll be tested as part of https://github.com/FluxML/Flux.jl/blob/v0.12.6/test/losses.jl#L168.

test/cuda/losses.jl
```diff
@@ -35,4 +36,16 @@ y = [1  0  0  0  1
...
+@testset "margin_ranking_loss" begin
+     x1 = [1.2 2.3 3.4]
+     x2 = [9.8 8.7 7.6]
+     y = 1
+   
+     @test Flux.margin_ranking_loss(x1, x2, y) ≈ Flux.margin_ranking_loss(gpu(x1), gpu(x2), gpu(y)) |> cpu
+     @test Flux.margin_ranking_loss(x1, x2, y, margin=1.0) ≈ 
+          Flux.margin_ranking_loss(gpu(x1), gpu(x2), gpu(y), margin=gpu(1.0)) |> cpu atol=1e-6
+     @test Flux.margin_ranking_loss(x1, x2, y, margin=1.0, mode=sum) ≈
+          Flux.margin_ranking_loss(gpu(x1), gpu(x2), gpu(y), margin=gpu(1.0), mode=sum)
```
ToucheSir, line 48: Why not put this in the testset above (right under `focal_loss`)? I'd even be fine with having it as a nested testset.

ToucheSir (changes requested):
Thanks again! Just a couple final nits.

src/losses/functions.jl
```diff
@@ -503,3 +503,32 @@ end
...
+1
+
+julia> Flux.margin_ranking_loss(x1, x2, y, margin=1.0, mode=sum) ≈ 22.2
+true
+```
+"""
+function margin_ranking_loss(x1, x2, y; margin=zero(eltype(x1)), mode=identity)
+  y ∈ (1, -1) || error("Rank can be either 1 or -1, got $y")
+  return (max.(zero(eltype(x1)), mode(((-y) .* (x1 .- x2)) .+ margin)))
+end
```
ToucheSir, line 531:
```suggestion
function margin_ranking_loss(x1, x2, y; margin=zero(eltype(x1)), agg=mean)
  any(∈((1, -1)), y) || error("Rank can be either 1 or -1, got $y")
  agg(max.(zero(eltype(x1)), -y .* (x1 .- x2) .+ margin))
end
```
Coming back to this after reading https://pytorch.org/docs/stable/generated/torch.nn.MarginRankingLoss.html, it seems that y can be a vector as well and `mode` is really a reduction. This suggestion should reflect that, and I think the tests may need updating as well.

test/losses.jl
ven-k: It is necessary because, unlike the ones in `ALL_LOSSES`, there are 3 inputs instead of just 2.

test/cuda/losses.jl
ven-k: That'll lower the coverage of test; also, the testset above takes in 2 inputs unlike this function

src/losses/functions.jl
```diff
@@ -503,3 +503,32 @@ end
 ```@meta
 DocTestFilters = nothing
 ```
+
+"""
+  margin_ranking_loss(x1, x2, y; margin=0, mode=identity)
```
CarloLucibello, line 508:
```suggestion
    margin_ranking_loss(x1, x2, y; margin=0, mode=identity)

```
Also please add the definition to the docstring as other losses do

CarloLucibello: Btw it should be updated with agg instead of mode

test/losses.jl
ToucheSir: Got it, I assumed `y` was optional since it was only ever provided as a scalar. For this PR, `y` should also be tested as a vector of `{1, -1}` as well. Longer term, we'll want to look at extending `ALL_LOSSES` to work with losses with arity > 2.

test/cuda/losses.jl
ToucheSir: Copying my reply from the comment thread above:
> Got it, I assumed `y` was optional since it was only ever provided as a scalar. For this PR, `y` should also be tested as a vector of `{1, -1}` as well. Longer term, we'll want to look at extending `ALL_LOSSES` to work with losses with arity > 2.

Also RE test coverage, moving tests and test suites around does not reduce coverage because it is calculated from how many lines of source code are covered and not how many tests/test suites exist. At least that's my understanding of the concern, feel free to clarify if you had something else in mind.

CarloLucibello:
closing as stale

PR #1127, opened 2020-04-16
Removed deprecated SGD exports

### LINKED ISSUE #1121: SGD exported but not defined
`Flux.Optimise` exports `SGD`, but SGD is not defined anywhere, likely renamed `Descent` at some point. When did that happen? Do we want to remove the SGD export or use it as an alias for Descent? I'm a bit torn here: I like SGD being an easily discoverable common name, but I don't like having two names for the same thing

bhvieira:
Closes #1121


dhairyagandhi96:
Thanks!

Issue #1128, opened 2020-04-17
CuArrays + huber_loss iterate(::nothing) error


HenriDeh:
Working on the gpu with huber_loss seems to break Zygote. Similar problem was discussed in #900. It seems to come from CuArrays since it does not happen otherwise. Maybe this can help understand the origin of the problem. 
 
MWE:
```
using Flux
nn = Chain(Dense(10,1))
l(x,y) = Flux.huber_loss(nn(x),y)
b = (rand(10,10), rand(1,10))
gradient(Flux.params(nn)) do
    l(b...)
end # works ok

nn = gpu(nn)
b = gpu(b)
gradient(Flux.params(nn)) do
    l(b...)
end 
ERROR: MethodError: no method matching iterate(::Nothing)
Closest candidates are:
  iterate(::Core.SimpleVector) at essentials.jl:600
  iterate(::Core.SimpleVector, ::Any) at essentials.jl:600
  iterate(::ExponentialBackOff) at error.jl:218
  ...
Stacktrace:
 [1] (::Zygote.var"#1906#1911"{Zygote.var"#1763#1767"})(::CuArray{Float32,2,Nothing}) at /home/dehaybe/.julia/packages/Zygote/4tJp5/src/lib/broadcast.jl:193
 [2] (::Zygote.var"#4#back#1912"{Zygote.var"#1906#1911"{Zygote.var"#1763#1767"}})(::CuArray{Float32,2,Nothing}) at /home/dehaybe/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:49
 [3] (::Zygote.var"#173#174"{Zygote.var"#4#back#1912"{Zygote.var"#1906#1911"{Zygote.var"#1763#1767"}},Tuple{NTuple{4,Nothing},Tuple{}}})(::CuArray{Float32,2,Nothing}) at /home/dehaybe/.julia/packages/Zygote/4tJp5/src/lib/lib.jl:159
 [4] (::Zygote.var"#334#back#175"{Zygote.var"#173#174"{Zygote.var"#4#back#1912"{Zygote.var"#1906#1911"{Zygote.var"#1763#1767"}},Tuple{NTuple{4,Nothing},Tuple{}}}})(::CuArray{Float32,2,Nothing}) at /home/dehaybe/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:49
 [5] broadcasted at ./broadcast.jl:1237 [inlined]
[...]

PR #1129, opened 2020-04-17
Added dropgrad in huber_loss

HenriDeh:
Workaround to prevent `iterate(::nothing)` when working with CuArrays. See issue #1128

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -48,7 +48,7 @@ given the prediction `ŷ` and true values `y`.
 """
 function huber_loss(ŷ, y;  δ=eltype(ŷ)(1))
    abs_error = abs.(ŷ .- y)
-   temp = abs_error .<  δ
+   temp = Zygote.dropgrad(abs_error .<  δ)
    x = eltype(ŷ)(0.5)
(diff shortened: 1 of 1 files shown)


bhvieira:
I think the change makes sense (not sure if it's in style with the rest of the library though). Without that Zygote would look into gradients through `temp`

HenriDeh:
Yes, I think it should only be a work around and removed when Zygote can handle it directly.
________________________________

dhairyagandhi96:
Is the issue a missing `.<` kernel for CuArrays?

HenriDeh:
I don't think so. 
```
julia> using CuArrays

julia> cu(rand(10)) .< cu(rand(10))
10-element CuArray{Bool,1,Nothing}:
 0
 1
 0
 1
 1
 0
 0
 0
 0
 1
```
works fine.

dhairyagandhi96:
Right, is that with scalar indexing? Just confirming

HenriDeh:
If you are asking if it works also when `CuArrays.allowscalar(false)` then yes. Otherwise I don’t understand your question.

CarloLucibello:
Filed an issue for this https://github.com/FluxML/Zygote.jl/issues/678

In the meantime let's merge this workaround 
bors r+

dhairyagandhi96:
We shouldn't have to merge this workaround, the issue should be fixed in Zygote.

PR #1130, opened 2020-04-17
Function "params" is lowercase

GoranDubajic:


bhvieira:
`Params` and `params` are different functions. `Params` is not exported, and comes through Zygote. The usage in the example is right, though.
Perhaps you could make it (they are all equivalent):

```
θ = params(W, b)
```

dhairyagandhi96:
If something needs fixing here, it would be `params(W, b)` but beyond that, I think its fine?

dhairyagandhi96:
Closing since it seems like things are fine, we could potentially add a note on the difference between the two.

PR #1133, opened 2020-04-20
add ClipValue and ClipNorm

AStupidBear:


bhvieira:
Keras has these, good to have it on Flux as well

src/optimise/optimisers.jl
```diff
@@ -533,3 +533,37 @@ function apply!(o::WeightDecay, x, Δ)
...
   @. Δ += wd * x
 end
+
+"""
+    ClipValue(thresh)
+
+Clip gradients when their absolute value exceeds `thresh`.
+
+# Parameters
+- Clipping threshold (`thresh`)
```
bhvieira, line 543: Nitpicking a bit too much on my part perhaps, but I think when you have few parameters the argument list is not necessary. See `LayerNorm` and `Dropout` docstrings for examples.

AStupidBear:
@bhvieira The docstrings are simpified now.

dhairyagandhi96:
Please add them to the tests as well

mcabbott:
Should both of these be called clipping, given that one clamps and the other scales?

HenriDeh:
norm uses scalar getindex. This throws an error when working with CuArrays. eg:
```
norm(cu(rand(10,10)), 2)
```

AStupidBear:
> Should both of these be called clipping, given that one clamps (only components outside range) and the other scales (all components)? Also they act on the gradient not the value. Maybe `ClampGrad` and, I don't know, `LimitGradNorm`, `BoundGrad`?
> 
> Also: is the norm of the whole array always what you want? If this is some batch of observations, then it may make more sense to scale each of them by its own norm. In which case this would need to take `dims` somehow.



> Should both of these be called clipping, given that one clamps (only components outside range) and the other scales (all components)? Also they act on the gradient not the value. Maybe `ClampGrad` and, I don't know, `LimitGradNorm`, `BoundGrad`?
> 
> Also: is the norm of the whole array always what you want? If this is some batch of observations, then it may make more sense to scale each of them by its own norm. In which case this would need to take `dims` somehow.

I think it's better to just follow Keras's convention. Gradients are not batched, so...

AStupidBear:
> norm uses scalar getindex. This throws an error when working with CuArrays. eg:
> 
> ```
> norm(cu(rand(10,10)), 2)
> ```
> 
> The issue comes from CuArrays. However, `norm(cu(rand(10,10)))` is implemented.

@HenriDeh Thanks for pointing that out.

mcabbott:
> Gradients are not batched, so...
> Gradients are not batched, so...

Oh sorry, this only applies to parameter updates, not to gradients in general. Which is obvious in the code,  `apply!`,  but not in the present docstrings.

AStupidBear:
Since this PR is just a simple feature add-on, how about merging this?

dhairyagandhi96:
bors try

CarloLucibello:
We need usage examples in the docstrings and we should add the two to the `docs/`

Project.toml
```diff
@@ -10,6 +10,7 @@ Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
 CuArrays = "3a865a2d-5b23-5a0f-bc46-62713ec82fae"
 DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
 Juno = "e5e0dc1b-0480-54bc-9374-aad01c23163d"
+LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
```
bhvieira, line 13: Is there any particular reason for this change in uuid?

src/optimise/optimisers.jl
```diff
@@ -533,3 +533,31 @@ function apply!(o::WeightDecay, x, Δ)
...
+    ClipNorm(thresh)
+
+Clip gradients when their L2 norm exceeds `thresh`.
+"""
+mutable struct ClipNorm{T}
+    thresh::T
+end
+
+function apply!(o::ClipNorm, x, Δ)
+    Δnrm = norm(Δ, 2)
```
HenriDeh, line 558: don't forget to change this to `norm(Δ)`. It is equivalent but works with CuArrays.

Project.toml
AStupidBear: LinearAlgebra's norm is used in ClipNorm

src/optimise/optimisers.jl
AStupidBear: change now

CarloLucibello:
this seems good to go, thanks

bors r+

PR #1136, opened 2020-04-21
add bilinear upsample layer

kczimm:
Implement a 2D bilinear upsample layer like the one found in [tensorflow](https://www.tensorflow.org/api_docs/python/tf/keras/layers/UpSampling2D).

### DIFF
--- a/src/layers/upsample.jl
+++ b/src/layers/upsample.jl
@@ -1,0 +1,41 @@
+struct BilinearUpsample{T<:Integer}
+    factor::Tuple{T,T}
+end
+
+function (b::BilinearUpsample)(x::AbstractArray)
+    W, H, C, N = size(x)
+
+    newW = W * b.factor[1]
+    newH = H * b.factor[2]
+
+    out = similar(x, (newW, newH, C, N))
+
+    for n = 1:N, c = 1:C, w = 1:newW, h = 1:newH
+        w₀ = (w - 0.5) / b.factor[1] + 0.5
+        h₀ = (h - 0.5) / b.factor[2] + 0.5
+
+        w1 = floor(Int, w₀)
+        w2 = w1 + 1
+        h1 = floor(Int, h₀)
+        h2 = h1 + 1
+
+        i1 = clamp(w1, 1, W)
+        i2 = clamp(w2, 1, W)
+        j1 = clamp(h1, 1, H)
+        j2 = clamp(h2, 1, H)
+
+        out[w, h, c, n] =
+            (
+                x[i1, j1, c, n] * (w2 - w₀) * (h2 - h₀) +
+                x[i1, j2, c, n] * (w2 - w₀) * (h₀ - h1) +
+                x[i2, j1, c, n] * (w₀ - w1) * (h2 - h₀) +
+                x[i2, j2, c, n] * (w₀ - w1) * (h₀ - h1)
+            ) / (w2 - w1 * h2 - h1)
+    end
+
+    out
+end
+
+function Base.show(io::IO, b::BilinearUpsample)
+    print(io, "BilinearUpsample(", b.factor[1], ", ", b.factor[2], ")")
+end
--- a/src/layers/upsample.jl
+++ b/src/layers/upsample.jl
@@ -30,7 +30,7 @@ function (b::BilinearUpsample)(x::AbstractArray)
 
     out = similar(x, (newW, newH, C, N))
 
-    for n = 1:N, c = 1:C, w = 1:newW, h = 1:newH
+    @inbounds for w = 1:newW, h = 1:newH
         w₀ = (w - 0.5) / b.factor[1] + 0.5
         h₀ = (h - 0.5) / b.factor[2] + 0.5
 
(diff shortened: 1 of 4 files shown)


src/layers/upsample.jl
```diff
@@ -0,0 +1,41 @@
...
+                x[i1, j1, c, n] * (w2 - w₀) * (h2 - h₀) +
+                x[i1, j2, c, n] * (w2 - w₀) * (h₀ - h1) +
+                x[i2, j1, c, n] * (w₀ - w1) * (h2 - h₀) +
+                x[i2, j2, c, n] * (w₀ - w1) * (h₀ - h1)
+            ) / (w2 - w1 * h2 - h1)
+    end
+
+    out
+end
+
```
johnnychen94, line 38: `@functor BilinearUpsample`

Ref: https://github.com/FluxML/Flux.jl/pull/1068

src/layers/upsample.jl
```diff
@@ -0,0 +1,41 @@
...
+function (b::BilinearUpsample)(x::AbstractArray)
+    W, H, C, N = size(x)
+
+    newW = W * b.factor[1]
+    newH = H * b.factor[2]
+
+    out = similar(x, (newW, newH, C, N))
+
+    for n = 1:N, c = 1:C, w = 1:newW, h = 1:newH
```
johnnychen94, line 13: This scalar loop would be extremely slow when `x` is `CuArray`

src/layers/upsample.jl
kczimm: Do I need this even though the structure parameters are not meant to be learned during training?

src/layers/upsample.jl
kczimm: Agreed, but I must admit I don't have experience optimizing Julia for CUDA usage. Do you have suggestions in this case?

AriMKatz: Broadcast would be fast on GPU

src/layers/upsample.jl
johnnychen94: Ah, I see. In that case we don't need `@functor`

kczimm:
build failed perhaps due to the Github outage I was experiencing? HTTP 500 errors on clones. I don't know how to trigger a rebuild without pushing another commit.

src/layers/upsample.jl
```diff
@@ -0,0 +1,61 @@
...
+ 0.845325  0.675933  0.337149  0.167757
+ 0.888679  0.710044  0.352773  0.174138
+ 0.910357  0.7271    0.360586  0.177329```
+"""
+struct BilinearUpsample{T<:Integer}
+    factor::Tuple{T,T}
+end
+
+function (b::BilinearUpsample)(x::AbstractArray)
+    W, H, C, N = size(x)
```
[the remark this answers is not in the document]
kczimm: The data in Flux is stored in WHCN order isn’t it? https://github.com/FluxML/Flux.jl/blob/7a32a703f0f2842dda73d4454aff5990ade365d5/src/layers/conv.jl#L17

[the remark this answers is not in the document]
johnnychen94: It is, but I prefer to read it as a misunderstanding.

kczimm:
I put this in a new file which I was hesitant to do. Perhaps it belongs with the conv code and docs because this is really used in convolutional networks?

johnnychen94:
> I put this in a new file which I was hesitant to do. Perhaps it belongs with the conv code and docs because this is really used in convolutional networks?

I guess it doesn't matter much. One thing I forgot to mention is to export the symbol `BilinearUpsample` in `src/Flux.jl`.

dhairyagandhi96:
Let's add gpu tests, and maybe we could think of performance on the GPU. It might be that we want to go with a broadcasting approach

scimas:
Linking this code written for bilinear interpolation using Interpolations.jl for reference, in case it offers any value for optimizing this pr. [comment](https://github.com/JuliaImages/Images.jl/issues/763#issuecomment-622065733)

CarloLucibello:
The approach in this PR seems neither Zygote nor Flux friendly, I don't think this is the way we go. @scimas did you test your code with differentiation and on gpu?

scimas:
@CarloLucibello I just got around to testing it today, doesn't work with zygote. I get errors about being unable to check bounds.

scimas:
So I did some more testing and Zygote calculates the gradient without any problem if I explicitly tell it to perform `forwarddiff`. Still have not tested on GPU.
```julia
using Zygote
up = Upsampler(2, 2);
x = rand(2, 2, 1, 1);
model(x) = sum(up(x));

# This works
grad = gradient(x) do p
    Zygote.forwarddiff(p) do p
        f(p)
    end
end

# But this doesn't
grad = gradient(x) do x
    f(x)
end
```
where `Upsampler` is from the previously linked code.

And frankly I don't have enough expertise to figure out why backward diff doesn't work, simply figuring out the forwarddiff part took most of the day.

ltjkoomen:
Hi all,

I have implemented my own bilinear upsampling layer that:
- is about 10 times as fast on the CPU as the code in this pull request
- Uses broadcasting, so no scalar indexing that slows down GPU operations
- allocates less
- is directly compatible with Zygote

It is located [here](https://github.com/ltjkoomen/Flux.jl/tree/add-upsample-layer)

I am not sure how to proceed, as this is my first time working on an open-source project. Should I make my own pull-request, or somehow add my code to this one?

kczimm:
Closing in favor of the superior #1180. Thanks everyone who reviewed this PR.

PR #1137, opened 2020-04-22
Fix BatchNorm Float64 grad with Float32 input

aterenin:
Re-opening a new PR for #758 because the previous one is on my fork's master branch, which I now need for a different PR. Sorry about any extra emails - GitHub doesn't seem to have a way to change the branch on the previous PR.


aterenin:
Per #757 this issue is fixed as a result of switch from Tracker to Zygote, so this PR is unnecessary.

PR #1139, opened 2020-04-25
Re: Feature : Some optimizers added

sriyash421:
Added the following function optimizers :

- Rprop
- HeavyBall

Re: #1063

### DIFF
--- a/src/optimise/Optimise.jl
+++ b/src/optimise/Optimise.jl
@@ -3,7 +3,8 @@ module Optimise
 export train!, update!,
 	Descent, ADAM, Momentum, Nesterov, RMSProp,
 	ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW,RADAM, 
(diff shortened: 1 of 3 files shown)


dhairyagandhi96:
I would appreciate doing a sanity check on whether we are doing the right thing by comparing the output from PyTorch for example.

dhairyagandhi96:
I would appreciate doing a sanity check on whether we are doing the right thing by comparing the output from PyTorch for example.

sriyash421:
> I would appreciate doing a sanity check on whether we are doing the right thing by comparing the output from PyTorch for example.

I'll try and compare the gradients across the two.

ToucheSir:
If this is still being pursued, the optimizers should probably be PR'ed against Optimisers.jl. I will say that neither has come up on GitHub, Discourse, Zulip or Slack (to my recollection) outside of this PR.

CarloLucibello:
closing for inactivity

PR #1140, opened 2020-04-25
Add Tensor Contraction

AStupidBear:


mcabbott:
This is a useful operation, although I think the name is too generic. IMO we should add this more widely, like here: https://github.com/JuliaMath/TensorCore.jl/pull/1

AStupidBear:
Yes, I found it strange to put that in Flux.jl

mcabbott:
Perhaps you can add your voice to the above PR to tell people this is useful?

Issue #189, opened 2018-03-01
Huge performance difference between sparse and dense representation on GPU


boathit:
I tried to add an embedding layer to the Char RNN example and found that the one-hot representation runs quite slower than its dense counterpart on GPU (up to 150 times). Even though we can always convert the one-hots into the dense vectors before passing them to the GPU,  I think in practice this will lead to a large amount of unnecessary GPU memory consumption in the case of large vocabulary.

```julia
using Flux
using Flux: onehot, argmax, chunk, batchseq, throttle, crossentropy
using Base.Iterators: partition
using BenchmarkTools:@btime

text = collect(readstring("input.txt"))
alphabet = [unique(text)..., '_']
text = map(ch -> onehot(ch, alphabet), text)
stop = onehot('_', alphabet)

N = length(alphabet)
seqlen = 50
nbatch = 150

Xs = collect(partition(batchseq(chunk(text, nbatch), stop), seqlen))
Ys = collect(partition(batchseq(chunk(text[2:end], nbatch), stop), seqlen));


using CuArrays

m = Chain(
  Dense(N, 128), # embedding layer
  LSTM(128, 128),
  Dense(128, N),
  softmax)

m = mapleaves(gpu, m)

opt = ADAM(params(m))

function loss(xs, ys)
  l = sum(crossentropy.(predict.(xs), gpu.(ys)))
  Flux.truncate!(m)
  return l
end
``` 
Performance difference,
```julia
# sparse representation
predict(x) = x |> gpu |> m
@btime Flux.train!(loss, zip(Xs[1:1], Ys[1:1]), opt);
# 17.930 s (16628920 allocations: 807.03 MiB)

# dense representation
predict(x) = collect(x) |> gpu |> m
@btime Flux.train!(loss, zip(Xs[1:1], Ys[1:1]), opt);
116.213 ms (102084 allocations: 7.94 MiB)
```

MikeInnes:
We should be converting to dense here; OneHot is designed to be *faster* than dense, because when you matmul you only need an indexing operation (and then everything will be identical).

So something has gone very wrong here; will have to investigate.

boathit:
Yeah, delving into the code OneHot is supposed to be faster than the dense vector and running with CPU it is indeed faster. The issue here arises from moving it into GPU, no acceleration observed but slowing down both the forward and backward computation.

boathit:
Tracing issue to OneHot matmul,

```julia
[...]

PR #1141, opened 2020-04-25
Speedup matmul of CuMatrix and OneHotMatrix

AStupidBear:
This solves #189.

```julia
julia> using Flux


julia> using Flux: CuArrays

julia> A = zeros(300, 10000) |> gpu;

julia> B = Flux.onehotbatch(rand(1:10000, 256), 1:10000) |> gpu;

julia> A * B; CuArrays.@time A * B;
└ @ GPUArrays ~/shared/.julia/packages/GPUArrays/OXvxB/src/host/indexing.jl:43
  0.002824 seconds (951 CPU allocations: 38.156 KiB) (2 GPU allocations: 301.000 KiB, 2.32% gc time of which 46.42% spent allocating)

julia> import Base: *

julia> A::AbstractMatrix * B::Flux.OneHotMatrix = @inbounds A[:, map(x->x.ix, B.data)]
* (generic function with 522 methods)

julia> A * B; CuArrays.@time A * B;
  0.000343 seconds (169 CPU allocations: 5.000 KiB) (2 GPU allocations: 301.000 KiB, 15.53% gc time of which 65.97% spent allocating)
```

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -37,10 +37,14 @@ import Adapt: adapt, adapt_structure
 
 adapt_structure(T, xs::OneHotMatrix) = OneHotMatrix(xs.height, adapt(T, xs.data))
 
-import .CuArrays: CuArray, CuArrayStyle, cudaconvert
+import .CuArrays: CuArray, CuMatrix, CuArrayStyle, cudaconvert
 import Base.Broadcast: BroadcastStyle, ArrayStyle
 BroadcastStyle(::Type{<:OneHotMatrix{<:CuArray}}) = CuArrayStyle{2}()
 cudaconvert(x::OneHotMatrix{<:CuArray}) = OneHotMatrix(x.height, cudaconvert(x.data))
+function Base.:(*)(A::CuMatrix, B::OneHotMatrix{CuArray{OneHotVector,1}})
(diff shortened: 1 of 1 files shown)


AStupidBear:
https://github.com/JuliaGPU/CuArrays.jl/issues/676

CarloLucibello:
Adding `@inbounds` is not safe, should at least check the size before. Is this a workaround for https://github.com/JuliaGPU/CuArrays.jl/issues/676? How does performance compare to moving to cpu? 
Could you add a comment like `# remove workaround when https://github.com/JuliaGPU/CuArrays.jl/issues/676 is fixed`

AStupidBear:
> Adding `@inbounds` is not safe, should at least check the size before. Is this a workaround for [JuliaGPU/CuArrays.jl#676](https://github.com/JuliaGPU/CuArrays.jl/issues/676)? How does performance compare to moving to cpu?
> Could add a comment like `# remove workaround when https://github.com/JuliaGPU/CuArrays.jl/issues/676 is fixed`

The performance of the `@inbounds` workaround is almost identical to `A[:, cpu(map(x->x.ix, B.data))]`. I will revert to the `cpu` version soon.

AStupidBear:
Any idea?

PR #1142, opened 2020-04-25
Improve padding

AStupidBear:
- remove SamePadding, use string instead
- add causal padding for 1D convolution
- use manual padding for CuArrays
- add test for same padding on gpu

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -20,13 +20,13 @@ function calc_padding(pad::AbstractString, k::NTuple{N,T}, dilation, stride) whe
     # For stride > 1 the output shape depends on the type of convolution layer.
     # In case amount of padding is odd we need to apply different amounts to each side.
     ntuple(Val(2*N)) do d
-      p = k_eff[d ÷ 2 + 1] - 1
-      isodd(d) ? ceil(Int, p / 2) : pad ÷ 2
+      p = k_eff[(d + 1) ÷ 2] - 1
+      isodd(d) ? ceil(Int, p / 2) : p ÷ 2
     end
   elseif pad == "causal"
     @assert length(k) == 1
     ntuple(Val(2*N)) do d
-      p = k_eff[d ÷ 2 + 1] - 1
+      p = k_eff[(d + 1) ÷ 2] - 1
       isodd(d) ? p : 0 
     end
   elseif pad == "valid"
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -34,6 +34,24 @@ function calc_padding(pad::AbstractString, k::NTuple{N,T}, dilation, stride) whe
   end
 end
 
+need_manual_padding(x, pad) = x isa CuArrays.CuArray && pad[1:2:end] != pad[2:2:end]
+
+function padding(x, pad::NTuple{N, Int}) where N
+  all(iszero, pad) && return x
+  if sum(pad) > 0
+    xl = fill!(similar(x, pad[1:2:end]..., size(x)[end-1:end]...), 0)
+    xr = fill!(similar(x, pad[2:2:end]..., size(x)[end-1:end]...), 0)
+    cat(xl, x, xr, dims = 1:(N ÷ 2))
+  else
+    ids = ntuple(Val(N ÷ 2)) do d
+      l, r = pad[2d - 1], pad[2d]
+      @show l, r
+      (1 - l):(size(x, d) + r)
+    end
+    x[ids..., :, :]
+  end
+end
+
 """
     Conv(size, in => out, σ = identity; init = glorot_uniform,
          stride = 1, pad = 0, dilation = 1)
@@ -70,7 +88,7 @@ end
 function Conv(w::AbstractArray{T,N}, b::AbstractVector{T}, σ = identity;
               stride = 1, pad = 0, dilation = 1) where {T,N}
   stride = expand(Val(N-2), stride)
-  dilation = expand(Val(N-2), dilation)
+  dilation = expand(Val(N-2), max.(1, dilation))
   pad = calc_padding(pad, size(w)[1:N-2], dilation, stride)
   return Conv(σ, w, b, stride, pad, dilation)
 end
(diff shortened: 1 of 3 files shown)


src/layers/conv.jl
```diff
@@ -8,25 +8,49 @@ _convtransoutdims(isize, ksize, ssize, dsize, pad) = (isize .- 1).*ssize .+ 1 .+
...
+
+need_manual_padding(x, pad) = x isa CuArrays.CuArray && pad[1:2:end] != pad[2:2:end]
+
+@nograd need_manual_padding
+
+function padding(x, pad::NTuple{N, Int}) where N
+  all(iszero, pad) && return x
+  if sum(pad) > 0
+    xl = fill!(similar(x, pad[1:2:end]..., size(x)[end-1:end]...), 0)
+    xr = fill!(similar(x, pad[2:2:end]..., size(x)[end-1:end]...), 0)
```
CarloLucibello, line 45: is this zygote friendly?

src/layers/conv.jl
```diff
@@ -95,7 +119,7 @@ Conv(weight = weight,
 function Conv(w::AbstractArray{T,N}, b::Union{Zeros, AbstractVector{T}}, σ = identity;
               stride = 1, pad = 0, dilation = 1) where {T,N}
   stride = expand(Val(N-2), stride)
-  dilation = expand(Val(N-2), dilation)
+  dilation = expand(Val(N-2), max.(1, dilation))
```
CarloLucibello, line 122: please don't insert unrelated changes in a PR. 
Also, if we want a dilsation >=1  better use an `@assert`

AStupidBear: removed now

src/layers/conv.jl
AStupidBear: already fixed

DrChainsaw:
Great to see support for asymmetric padding. Just wondering if there is a better place to add it. Why not put it where this warning message is emitted:
```julia
julia> cc = Conv((3,4), 2=>2, pad=(0,1,2,3)) |> gpu;

julia> cc(ones(4,4,2,1) |> gpu)
┌ Warning: CuDNN does not support asymmetric padding; defaulting to symmetric choice
└ @ CuArrays.CUDNN E:\Programs\julia\.julia\packages\CuArrays\A6GUx\src\dnn\conv.jl:49
```
That code does anyways not seem to work as you get a CUDNN_STATUS_BAD_PARAM right after the warning.

Btw, whats the reason for using strings as the padding alternatives? Isn't symbols more idiomatic and saves one symbol of typing? I guess the current implementation also locks out people from adding more padding types as strings, but I don't wanna be padshedding too much :)

src/layers/conv.jl
```diff
@@ -229,16 +265,16 @@ end
 
 @functor ConvTranspose
 
-function conv_transpose_dims(c::ConvTranspose, x::AbstractArray)
+function conv_transpose_dims(c::ConvTranspose, x::AbstractArray; pad = c.pad)
```
johnnychen94, line 268: We should either expose no kwargs here (keep this unchanged) or expose all the kwargs (i.e., `pad`, `dilation` and `stride`) from `DenseConvDims`.

My personal opinion is that I don't feel good to expose these keywords.

src/layers/conv.jl
```diff
@@ -8,25 +8,56 @@ _convtransoutdims(isize, ksize, ssize, dsize, pad) = (isize .- 1).*ssize .+ 1 .+
...
 
-"""
-    SamePad
-
-Padding for convolutional layers will be calculated so that outputshape == inputshape when stride = 1.
-
-For stride > 1 the output shape depends on the type of convolution layer.
-"""
-struct SamePad end
+calc_padding(pad, k::NTuple{N,T}, dilation, stride) where {T,N} = expand(Val(2*N), pad)
```
johnnychen94, line 11: Personally, I prefer the original trait-based use pattern; I don't think it makes things clearer or easier to manually write the dispatching table (if-elseif-else).

If removing the old `SamePad`, we should add a deprecation warning for it, otherwise, people get surprised. Saying this, I don't think it a good idea to change the API in this enhancement PR. A separate PR or issue would be fine.

BTW, I don't see any tests related to this, can you add some for it?

src/layers/conv.jl
```diff
@@ -8,25 +8,56 @@ _convtransoutdims(isize, ksize, ssize, dsize, pad) = (isize .- 1).*ssize .+ 1 .+
...
   k_eff = @. k + (k - 1) * (dilation - 1)
-  # How much total padding needs to be applied?
-  pad_amt = @. k_eff - 1
-  # In case amount of padding is odd we need to apply different amounts to each side.
-  return Tuple(mapfoldl(i -> [ceil(Int, i/2), floor(Int, i/2)], vcat, pad_amt))
+  if pad == "same"
+    # Padding for convolutional layers will be calculated so that outputshape == inputshape when stride = 1.
+    # For stride > 1 the output shape depends on the type of convolution layer.
+    # In case amount of padding is odd we need to apply different amounts to each side.
+    ntuple(Val(2*N)) do d
```
johnnychen94, line 22: What's the reasoning behind `Val`? My local test says it's slower.

```julia
julia> @btime ntuple(x->2x, Val(2*N))
  2.278 μs (0 allocations: 0 bytes)
(2, 4, 6, 8)

julia> @btime ntuple(x->2x, 2*N)
  28.345 ns (0 allocations: 0 bytes)
(2, 4, 6, 8)
```

src/layers/conv.jl
```diff
@@ -8,25 +8,56 @@ _convtransoutdims(isize, ksize, ssize, dsize, pad) = (isize .- 1).*ssize .+ 1 .+
...
+
+function _get_paddings(x, pad)
+  xl = fill!(similar(x, pad[1:2:end]..., size(x)[end-1:end]...), 0)
+  xr = fill!(similar(x, pad[2:2:end]..., size(x)[end-1:end]...), 0)
+  return xl, xr
+end
+
+@nograd _get_paddings
+
+function padding(x, pad::NTuple{N, Int}) where N
```
johnnychen94, line 49: This function name `padding` seems like a very common name and is very likely to be confused with kwargs `pad` and `padding`. How about renaming it to `pad_array`?

src/layers/conv.jl
```diff
@@ -249,8 +285,13 @@ end
 function (c::ConvTranspose)(x::AbstractArray)
   # ndims(x) == ndims(c.weight)-1 && return squeezebatch(c(reshape(x, size(x)..., 1)))
   σ, b = c.σ, reshape(c.bias, map(_->1, c.stride)..., :, 1)
-  cdims = conv_transpose_dims(c, x)
-  σ.(∇conv_data(x, c.weight, cdims) .+ b)
+  if need_manual_padding(x, c.pad)
+    cdims = conv_transpose_dims(c, x, pad = zero.(c.pad))
+    x = padding(σ.(∇conv_data(x, c.weight, cdims) .+ b), .-c.pad)
```
johnnychen94, line 290: unused `x` here.

src/layers/conv.jl
johnnychen94: Is this type-stable?

PR #1143, opened 2020-04-25
add GaussianNoise

AStupidBear:


src/layers/normalise.jl
```diff
@@ -414,3 +414,27 @@ function Base.show(io::IO, l::GroupNorm)
...
+  stddev::F
+  active::Union{Bool, Nothing}
+end
+
+GaussianNoise(stddev) = GaussianNoise(stddev, true)
+
+function (m::GaussianNoise)(x)
+  _isactive(a) || return x
+  a.stddev == 0 && return x
+  y = m.stddev .* randn!(similar(x))
```
johnnychen94, line 435:
```suggestion
  y = convert(eltype(x), m.stddev) .* randn!(similar(x))
```

AStupidBear: Thanks

CarloLucibello:
what's the use case for this? Is there any precedence in other DL libs?

AStupidBear:
> what's the use case for this? Is there any precedence in other DL libs?

Keras has a [GaussianNoise](https://keras.io/api/layers/regularization_layers/gaussian_noise) layer too.

AStupidBear:
Any idea?

ToucheSir:
This looks to be a decent enough implementation and the layer struct makes sense because you want to control `active`. @AStupidBear do you know where this was introduced into the literature? I think that would help with deciding whether it's worth adding to Flux since no library outside of (tf.)Keras appears to have it.

darsnack:
I think the reason we don't see this too often is because it is an easy "roll your own" normalization scheme. That being said, noise injection is an idea that comes up quite often in research (even if it isn't popular in practice). Have a convenient wrapper would be nice.

@AStupidBear how do you feel about generalizing this to arbitrary noise generation functions? I am thinking of cases where someone might reference and RNG sampler from Distributions.jl etc.

CarloLucibello:
closing as stale

PR #1146, opened 2020-04-25
Use ProgressMeter to print current loss

AStupidBear:
image


bhvieira:
That would remove the dependency on Juno, which I'm led to believe was only added for @progress. Is that so @MikeInnes ?

AStupidBear:
@bhvieira https://github.com/FluxML/Flux.jl/blob/134ac1586baf04819a7f0149f1d185d94d9bd0ee/src/data/tree.jl#L26

Juno is still used there to print trees

src/optimise/train.jl
```diff
@@ -75,30 +75,46 @@ The callback can call [`Flux.stop`](@ref) to interrupt the training loop.
...
+function train!(loss, ps, data, opt; cb = () -> (), epochs = 1, steps_per_epoch = typemax(Int), verbose = true)
   ps = Params(ps)
   cb = runall(cb)
-  @progress for d in data
-    try
-      if d isa AbstractArray{<:Number}
-        gs = gradient(ps) do
-          loss(d)
+  l̄ = 0f0
+  cb_narg = methods(cb).ms[1].nargs - 1
```
dhairyagandhi96, line 82: Do we need these kinds of checks? Maybe without checking the method table

AStupidBear: This is for backward compatibility. In this PR, a callback function should be called with current parameters and current loss. But users are still allowed to pass a zero argument function as callback just like before.

dhairyagandhi96: We should not use slow functions like `methods` in general, also they are not covered under the stability guarantee of the language.

AStupidBear: Ok, that part is removed.

MikeInnes:
I think something like this would be useful, although I don't think we can accept a PR that tries to do a number of different things at once.

I would prefer not to lose the Juno integration; does Juno work with ProgressMeters at this point?

johnnychen94:
> does Juno work with ProgressMeters at this point?

[TerminalLoggers.jl](https://github.com/c42f/TerminalLoggers.jl) might be the right tool. See also: https://junolab.org/ProgressLogging.jl/stable/#ProgressLogging.ProgressLogging

AStupidBear:
@johnnychen94 @MikeInnes I added progressive loss printing for [DiffEqFlux.sciml_train](https://github.com/SciML/DiffEqFlux.jl/blob/7bfeccc128dcdf48e68931dfb1c5db8bd7944ec8/src/train.jl#L109) before, but that's not as pretty as this PR with ProgressMeter.jl. See https://github.com/JunoLab/ProgressLogging.jl/issues/23

AStupidBear:
> I think something like this would be useful, although I don't think we can accept a PR that tries to do a number of different things at once.

@MikeInnes Now only the printing part is reserved. Can I say this PR now only do one thing?

src/optimise/train.jl
```diff
@@ -75,30 +75,44 @@ The callback can call [`Flux.stop`](@ref) to interrupt the training loop.
 
 Multiple optimisers and callbacks can be passed to `opt` and `cb` as arrays.
 """
-function train!(loss, ps, data, opt; cb = () -> ())
+function train!(loss, ps, data, opt; cb = () -> (), epochs = 1, steps_per_epoch = typemax(Int), verbose = true)
```
CarloLucibello, line 78: the new kwargs should go into the docstring

src/optimise/train.jl
```diff
@@ -75,30 +75,44 @@ The callback can call [`Flux.stop`](@ref) to interrupt the training loop.
...
   cb = runall(cb)
-  @progress for d in data
-    try
-      if d isa AbstractArray{<:Number}
-        gs = gradient(ps) do
-          loss(d)
+  l̄ = 0f0
+  ndata = min(length(data), steps_per_epoch)
+  for epoch in 1:epochs
+    printstyled("Epoch $epoch/$epochs\n", color = :yellow)
```
CarloLucibello, line 84: maybe print this only if verbose=true

src/optimise/train.jl
CarloLucibello: while `epochs` is convenient, I don't see why we should have `steps_per_epoch`

dhairyagandhi96: Epochs should with the macro instead

CarloLucibello:
This PR is also adding an extra for loop over the `epochs`. I think this is convenient to have, but certainly not related to the progress bar infrastructure

AStupidBear:
> This PR is also adding an extra for loop over the `epochs`. I think this is convenient to have, but certainly not related to the progress bar infrastructure

Now the loop over `epochs` is removed. Can this PR be merged?

CarloLucibello:
I like ProgressMeter, but we would be losing Juno support. Is there a way we can have both?

AStupidBear:
> I like ProgressMeter, but we would be losing Juno support. Is there a way we can have both?

Added back!

src/optimise/train.jl
```diff
@@ -75,21 +75,30 @@ The callback can call [`Flux.stop`](@ref) to interrupt the training loop.
 
 Multiple optimisers and callbacks can be passed to `opt` and `cb` as arrays.
 """
-function train!(loss, ps, data, opt; cb = () -> ())
+function train!(loss, ps, data, opt; cb = () -> (), verbose = true)
```
CarloLucibello, line 78: the default should be false if we want to avoid change in behavior (people may already be using callbacks for printing). Also, the argument should be documented

CarloLucibello, line 78:
```suggestion
function train!(loss, ps, data, opt; cb=() -> (), verbose=False)
```

AStupidBear: Thanks! That's added now.

MikeInnes:
To be honest, this is a fair bit of extra complexity when our `train!` function is meant to be very simple (not least so people can just copy it and do their own thing). It also seems quite specific; I'm sure there are people for whom the particular behaviour of `verbose` is useful, but anyone who wants something a bit different is back to writing the for loop again.

I suspect it would be better to experiment with more powerful training loops in a separate package, rather than adding feature after feature to Flux proper (bearing in mind we are constrained by stability). That would give you a lot of flexibility to experiment, and as the design evolves we can consider whether it's time to make it official.

Of course, I think if we could simply make `@progress` support the terminal as well as Juno without code changes here, that'd be a no brainer.

AStupidBear:
Close in favor of https://github.com/AStupidBear/DistributedFlux.jl

PR #1147, opened 2020-04-25
Add Orthogonal Initialization

AStupidBear:
- add orthogonal initialization
- add kernel_init for recurrent cells
- add more rnn variants: SGRU, MGU, SMGU


dhairyagandhi96:
I would appreciate some context about whether there are other libraries that vendor these layers by default

dhairyagandhi96:
Needs tests for the init method, and also checking on the GPU since currently it uses `Matrix` which wouldn't be GpU compatible.

Need to add to docs and probably even factor out the layers into a separate PR

PR #1149, opened 2020-04-26
Cleaner training loop

DhairyaLGandhi:


dhairyagandhi96:
@CarloLucibello it would be good if `DataLoader` did this by default

Would the necessary change be to wrap the output of `getdata` with a single index in a `tuple`?

CarloLucibello:
The current behavior is not accidental, it is what I would expect when iterating on a single data collection: 
1. `for x in DataLoader(X)`
2. `for (x, y) in DataLoader(X, Y)`

Knet and pytorch do the same. We could add

3. `for (x,) in DataLoader((X,))`
4. `for (x, y) in DataLoader((X, Y))`

and maybe deprecate 2., but I would keep 1. as it is, along with the changes in this PR

johnnychen94:
I prefer to take `DataLoader((X, Y))` as a combination of two datasets instead of an alternative to `DataLoader(X, Y)`.

dhairyagandhi96:
I would prefer it if we followed consistent semantics. Since we have so far allowed every element of a dataset being basically a minibatch and it's labels, it follows that single data collections be represented by `(x,)`

src/optimise/train.jl
```diff
@@ -56,6 +56,10 @@ function stop()
   throw(StopException())
 end
 
+@inline batchmemaybe(x::AbstractArray) = tuple(x)
+@inline batchmemaybe(x::AbstractArray{T}) where T <: AbstractArray = x
+@inline batchmemaybe(x) = x
```
CarloLucibello, line 61: Why not follow the current branching:
```julia
@inline batchmemaybe(x::AbstractArray{<:Number}) = tuple(x)
@inline batchmemaybe(x) = x
```
which is strict with respect to when bypassing splatting?

CarloLucibello:
> I would prefer it if we followed consistent semantics. Since we have so far allowed every element of a dataset being basically a minibatch and it's labels, it follows that single data collections be represented by `(x,)`

I don't see why a data iterator for the unsupervised setting should follow the semantics of the supervised one. 
In any case, I think this PR should be merged independently of what happens to DataLoader, since we want to support generic data iterators.

src/optimise/train.jl
dhairyagandhi96: Cases where we optimise colors and `RGB` values don't follow subtyping `Number`.

We shouldn't disallow `StructArray` here either

bhvieira:
Good to have multiple-dispatch on this instead of the if/else, but the naming of the function is non-informative. Perhaps a better name could be suggested?

dhairyagandhi96:
I'll take suggestions

src/optimise/train.jl
dhairyagandhi96: This might be tricky if we had `VecOfArray` inputs

dhairyagandhi96:
Assuming we require every data loader to return a tuple of arguments to the loss, we would remain consistent and generic.

In that case both supervised and unsupervised data loaders should just return a tuple of what they expect to be arguments to the loss.

CarloLucibello:
All right. So, maybe let's take a week or so to merge a few non-breaking PRs, tag a release, and then pipeline as many breaking PRs as we can for the release after that

MikeInnes:
Carlo's option of deleting the `DataLoader(X, Y)` method and documenting `DataLoader((X, Y))` would be fine. Then we can delete the branch from the training loop and encourage either `DataLoader((X,))` of `zip(DataLoader(X))` in the unsupervised case.

> I don't see why a data iterator for the unsupervised setting should follow the semantics of the supervised one.

Primarily because there *is* no unsupervised or supervised case as far as this code is concerned; just a one-arg and multi-arg case. Here's a few ways the current setup could go wrong:

* Splatting arguments; `loss(x...)` and `DataLoader(X...)`. This will throw a confusing error in the case where you happen to have only one `X`, where it should just work consistently.
* Modifying existing code: If I change `loss(x, y)` to `loss(x)` (or vice versa) it's intuitive to change `DataLoader(X, Y)` to `DataLoader(X)`. In fact the correct change is more subtle and non-obvious.
* Hitting this branch unexpectedly: If you have `data = [[1, 2, 3], [4, 5, 6], ...]` we should call `loss(1, 2, 3)` etc. Every other iterator of args works, but `Array{<:Number}` will break with an unhelpful error. We can't even assume iterators have consistent eltypes in julia, so someone could write `data = [(1, 2, 3), [4, 5, 6], (i for i in 1:3)]` and get very inconsistent behaviour.

Perhaps these seem unlikely, but if we really feel that supervised and unsupervised are fundamentally separate and need to behave differently, we need to represent that with two separate APIs and different types, not with a distinction between one and multiple arguments.

Also, note that because of the last point, the original PR #1051 was breaking. So quickly un-breaking it in a patch release seems like less of a big problem, since it's only restoring compatibility with v10.0.

CarloLucibello:
Actually I'm still quite torn. On a second thought I would use for `train` the rule  "a tuple argument will be splatted", implemented as
```
batchmemaybe(x) = tuple(x)
batchmemaybe(x::Tuple) = x

loss(batchmemaybe(x)...)
```
Would this cover any reasonable scenario?

dhairyagandhi96:
Agree that an un-breaking patch would be least disruptive

MikeInnes:
Just splatting tuples might be better, and if we do that it's easy for more advanced use cases to get reliable behaviour (just wrap everything in a tuple). There's still some potential for surprise since in general, iterators and tuples behave the same in Julia, but it's a big improvement over the current situation.

I suggest we figure that out as a follow up to the fixes here (it would also technically be breaking).

dhairyagandhi96:
I've added a simple backwards compatibility check with a warning hinting the users to the changed API. This along with checking for batching within the train loop should help with not breaking code. A quick look would be helpful @MikeInnes

src/data/dataloader.jl
```diff
@@ -27,66 +27,71 @@ The original data is preserved as a tuple in the `data` field of the DataLoader.
...
+   Xtrain = rand(10, 100)
+   Ytrain = rand(100)
+   train_loader = DataLoader((Xtrain, Ytrain), batchsize=2, shuffle=true) 
+   for epoch in 1:100
+     for (x, y) in train_loader
+       @assert size(x) == (10, 2)
+       @assert size(y) == (2,)
+       ...
+     end
+   end
```
CarloLucibello, line 49: this has to be 4 space indented, otherwise documenter will complain

src/data/dataloader.jl
```diff
@@ -27,66 +27,71 @@ The original data is preserved as a tuple in the `data` field of the DataLoader.
...
+  if d.shuffle && i == 0
+    shuffle!(d.indices)
+  end
+  nexti = min(i + d.batchsize, d.nobs)
+  ids = d.indices[i+1:nexti]
+  if length(d.data) == 1
+    batch = tuple(getdata(d.data[1], ids))
+  else
+    batch = ((getdata(x, ids) for x in d.data)...,)
+  end
```
CarloLucibello, line 90: there is no need for branching here, `((getdata(x, ids) for x in d.data)...,)` will do in both cases

CarloLucibello:
I'd like to preserve dataloader's interface  1. (and possibly also 2.) along with  3. and 4. introduced here, and without any warning. Let me do this in another PR and see if you people like it

dhairyagandhi96:
The PR as it stands is not breaking anymore, with a compat layer

CarloLucibello:
> The PR as it stands is not breaking anymore, with a compat layer

it's still deprecating the current dataloader interface, I'd like to avoid that

CarloLucibello:
also this is breaking the iteration behavior of DataLoader(X)

MikeInnes:
Let's do the following:

* Delete the `DataLoader(X, Y)` method, but support `DataLoader(X)` and `DataLoader((X, Y))`. In principle this could be generalised to any nest of tuples, and the tuple structure you iterate over reflects what you put in.
* Delete the branch in the training loop / `batchmemaybe` entirely to restore compatibility with v10.0.
* In a follow up PR we can consider adding a reverse `batchmemaybe` that explicitly splats only tuples, rather than not splatting only numeric arrays; this will have to go into v11.0.

CarloLucibello:
> ollowing:
> 
> * Delete the `DataLoader(X, Y)` method, but support `DataLoader(X)` and `DataLoader((X, Y))`. In principle this could be generalised to any nest of tuples, and the tuple structure you iterate over reflects what you put in.
> * Delete the branch in the training loop / `batchmemaybe` entirely to restore compatibility with v10.0.
> * In a follow up PR we can consider adding a reverse `batchmemaybe` that explicitly splats only tuples, rather than not splatting only numeric arrays; this will have to go into v11.0.

I've implemented the DataLoader part of this ins #1152 .
We are breaking, the DataLoader interface though. It's been around for more than a month and it is used in recent model-zoo updates. This seems likely to cause more breakage than the previous change in `train` that probably has gone unnoticed. I suggest we implement all of the changes now and tag v0.11

dhairyagandhi96:
Hmm, is this closer to what you had in mind?

CarloLucibello:
@dhairyagandhi96 bump. This should implement https://github.com/FluxML/Flux.jl/pull/1149#issuecomment-620000631 , so that we are done with the train/DataLoader overhaul

cossio:
Just so this idea is not lost. We could splat `NamedTuple` as keywords `loss(; nt...)`, which I think would fit very nicely with https://github.com/FluxML/Flux.jl/pull/1221 (@CarloLucibello's original comment https://github.com/FluxML/Flux.jl/pull/1221#issuecomment-644900982).

Something like

```
if d isa Tuple
  loss(d...)
elseif d isa NamedTuple
    loss(; d...)
end
```

dhairyagandhi96:
Ah, so the change here is to support a loss function which takes in named tuples? We should add a test here for that case as well

CarloLucibello:
Actually, I wouldn't have `train!` specialize on named tuples as well, we can't possibly support any type an iterator can throw. I suggest we have `train!` just distinguish tuples and non-tuples, as per #1149 (comment) .

In https://github.com/FluxML/Flux.jl/pull/1221#issuecomment-644900982 I was referring to somthing that could be done on the user side. That is if they want to work with named tuples and keyword arguments  they can do something like

```julia
function loss(; x, y)
....
end

loss(nt) = loss(; nt...)  # helper for train!

train_loader = DataLoader((x=rand(10,10), y=rand(10)))

train!(loss, ps, train_loader, opt)
```

CarloLucibello:
a squash  before merging would be nice

DhairyaLGandhi:
Now that I am looking at it, it means that I can no longer have a mini batch be something like `[x,y]` (ie not a tuple) which seems a little unfortunate. We are then making the API a bit more rigid to what it can expect.

src/optimise/train.jl
```diff
@@ -63,8 +63,9 @@ batchmemaybe(x::Tuple) = x
 For each datapoint `d` in `data` compute the gradient of `loss(d...)` through
 backpropagation and call the optimizer `opt`.
 
-In case datapoints `d` are of numeric array type, assume no splatting is needed
-and compute the gradient of `loss(d)`.
+Each `d` should return a collection of arguments to the `loss` function wrapped as a tuple.
+
+In case datapoints `d` are of numeric array type, assume no splatting is needed and compute the gradient of `loss(d)`.
```
CarloLucibello, line 68: Maybe the last 3 sentences could be replaced by
```
For each datapoint `d` in `data`, compute the gradient of  `loss` with respect to `params` through
backpropagation and call the optimizer `opt`. The way the batch `d` is passed to `loss` depends on the type of `d` :
- if `d` is a tuple, call  `loss(d...)`
- otherwise, call `loss(d)` 
```

src/optimise/train.jl
```diff
@@ -60,12 +60,7 @@ batchmemaybe(x::Tuple) = x
 """
     train!(loss, params, data, opt; cb)
 
-For each datapoint `d` in `data` compute the gradient of `loss(d...)` through
-backpropagation and call the optimizer `opt`.
-
-Each `d` should return a collection of arguments to the `loss` function wrapped as a tuple.
-
-In case datapoints `d` are of numeric array type, assume no splatting is needed and compute the gradient of `loss(d)`.
+For each datapoint `d` in `data`, compute the gradient of  `loss` with respect to `params` through backpropagation and call the optimizer `opt`. If `d` is a tuple of arguments to `loss`, call `loss(d...)`. Else, `loss` may handle `d` as desired.
```
CarloLucibello, line 63: line too long.
Also, the meaning of "`loss` may handle `d` as desired" is not clear
Could you squash the commits after fixing this?

DhairyaLGandhi:
oof seemingly chose some weird commits to squash. Will fix that

src/optimise/train.jl
DhairyaLGandhi: Is this clearer?

Issue #1088, opened 2020-03-17
revisit DataLoader api


CarloLucibello:
Add more consistency to the current design between the cases `DataLoader(X)`, `DataLoader(X, Y)`. As stated by @MikeInnes 
"""
It's not good that DataLoader(X) returns batches of X but DataLoader(X, Y) returns a tuples of batches. The special case in train! is not great either, since it's supposed to be simple and obvious.

There are two options here: either have DataLoader(X) return a tuple, or only allow one input but use DataLoader((X, Y)) to return tuples. Then one can write DataLoader((Xs...,)) without ambiguity about what the type will be.

The former might be a good option insofar as it makes it easy to remove the train! special case. It'd also be easy to tell people to write train!(..., zip(loader), ...) to wrap the iterator output in a tuple.
"""

cossio:
Close?

PR #1152, opened 2020-04-29
extend dataloader

CarloLucibello:
cfr discussion in #1149. Currently DataLoader interface supports

1. `for x in DataLoader(X)`
2. `for (x, y) in DataLoader(X, Y)`

This PR adds

3. `for (x,) in DataLoader((X,))`
4. `for (x, y) in DataLoader((X, Y))`

Edit:
the constructor in 2. is removed in this PR


CarloLucibello:
Removed multi-arg constructor as discussed in #1149

CarloLucibello:
fix #1088

dhairyagandhi96:
There's a few empty lines added here, good to check them

dhairyagandhi96:
There was mention on slack about some type instability, could we clear that

src/data/dataloader.jl
```diff
@@ -52,41 +58,45 @@ Example usage:
...
+    if n < batchsize
+        @warn "Number of observations less than batchsize, decreasing the batchsize to $n"
+        batchsize = n
     end
-    imax = partial ? nx : nx - batchsize + 1
-    ids = 1:min(nx, batchsize)
-    DataLoader(data, batchsize, nx, partial, imax, [1:nx;], shuffle)
+    imax = partial ? n : n - batchsize + 1
+    ids = 1:min(n, batchsize)
+    DataLoader(data, batchsize, n, partial, imax, [1:n;], shuffle)
```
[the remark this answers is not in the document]
CarloLucibello: > Do we end up creating a copy of the data that we pass into it? It will be annoying while using GPUs where VRAM is already at a premium.

the entire datasets is kept with no copying. We have copying only when indexing to produce a mini-batch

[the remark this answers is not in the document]
CarloLucibello: > Maybe it should specialize on if we can move data to the GPU in a separate thread while the training is running and add it to a buffer which can be done for like a couple batches worth, so we can eliminate time to move, retrieve data without saturating the GPU?

this should be done at some point

cossio:
See https://github.com/FluxML/Flux.jl/issues/1159

CarloLucibello:
fix #1159

src/optimise/train.jl
```diff
@@ -80,14 +83,8 @@ function train!(loss, ps, data, opt; cb = () -> ())
...
   @progress for d in data
     try
-      if d isa AbstractArray{<:Number}
-        gs = gradient(ps) do
-          loss(d)
-        end
-      else
-        gs = gradient(ps) do
-          loss(d...)
-        end
```
dhairyagandhi96, line 90: Already in #1149 ,let's not repeat it here

.travis.yml
```diff
@@ -16,7 +16,7 @@ notifications:
 jobs:
   include:
     - stage: "Documentation"
-      julia: 1
+      julia: 1.4
```
dhairyagandhi96, line 19: I think it's fixed on master

CarloLucibello:
other changes 
- bumped pkg version
- removed the Manifest (as the discussion in https://github.com/FluxML/Zygote.jl/pull/629 applies also to Flux)
-  implement also the `train` part of https://github.com/FluxML/Flux.jl/pull/1149#issuecomment-621112392 (since this PR was already breaking, let's do all of it and bump to v0.11)

src/optimise/train.jl
```diff
@@ -56,14 +56,17 @@ function stop()
   throw(StopException())
 end
 
+maketuple(x) = (x,)
+maketuple(x::Tuple) = x
```
dhairyagandhi96, line 60: We'd decided to not do this, so let's not

CarloLucibello: this is the last point of https://github.com/FluxML/Flux.jl/pull/1149#issuecomment-621112392

dhairyagandhi96:
I'd suggest keeping the manifest changes separate as there is a separate PR and discussion unrelated to Data loader here

MikeInnes:
Best to keep each PR to one logical change. For this that should be the data loaders; the training loop can happen in #1149 and we can follow up with the manifest change. As much as anything, the manifest change is going to lead to more merge conflicts.

dhairyagandhi96:
bump

src/data/dataloader.jl
```diff
@@ -52,41 +58,50 @@ Example usage:
...
+function _nobs(data::Tuple)
+    length(data) > 0 || throw(ArgumentError("Need at least one data input"))
+    n = _nobs(data[1])
+    if !all(x -> _nobs(x) == n, data[2:end])
+        throw(DimensionMismatch("All data should contain same number of observations"))
+    end
+    return n
+end
+
+function _getobs(data::A, i) where A<:AbstractArray{T,N} where {T,N}
```
cossio, line 101:
```suggestion
function _getobs(data::AbstractArray{T,N}, i) where {T,N}
```

src/data/dataloader.jl
```diff
@@ -52,41 +58,50 @@ Example usage:
...
+    getindex(data, ntuple(i->Colon(), N-1)..., i)
+end
+
+_getobs(data::Tuple, i) = ((_getobs(x, i) for x in data)...,)
```
cossio, line 105:
```suggestion
_getobs(data::Tuple, i) = map(x -> _getobs(x, i), data)
```

src/data/dataloader.jl
```diff
@@ -52,41 +58,50 @@ Example usage:
...
     end
-    if nx < batchsize
-        @warn "Number of data points less than batchsize, decreasing the batchsize to $nx"
-        batchsize = nx
-    end
-    imax = partial ? nx : nx - batchsize + 1
-    ids = 1:min(nx, batchsize)
-    DataLoader(data, batchsize, nx, partial, imax, [1:nx;], shuffle)
+    imax = partial ? n : n - batchsize + 1
+    ids = 1:min(n, batchsize)
```
cossio, line 70: Remove this line?

CarloLucibello:
reinstated the Manifest and removed the changes to `train!` (that should happen in #1149 instead). This should be ready to go

Manifest.toml
```diff
@@ -384,4 +384,4 @@ version = "0.4.20"
 deps = ["MacroTools"]
 git-tree-sha1 = "b3b4882cc9accf6731a08cc39543fbc6b669dca8"
 uuid = "700de1a5-db45-46bc-99cf-38207098b444"
-version = "0.2.0"
+version = "0.2.0"
```
MikeInnes, line 387: Trailing newline is missing here

src/optimise/train.jl
```diff
@@ -121,4 +121,4 @@ macro epochs(n, ex)
       @info "Epoch $i"
       $(esc(ex))
     end)
-end
+end
```
MikeInnes, line 124: also here

PR #1156, opened 2020-05-03
Add correct overload for apply! in docs

dhairyagandhi96:
Maybe we should considering adding a `const` name that is better than `apply!` (or rename `apply!`) and export it, so folks can just overload `descriptive_apply_my_optimiser_rule!` rather than have to go to the sub-project `Flux.Optimise`?


bhvieira:
`apply!` applies the update rule associated with an optimizer, so perhaps `apply_optim_update!` or `apply_update_rule!`

dhairyagandhi96:
Can brainstorm the name/ steps later, merging this for now.

bors r+

dhairyagandhi96:
I like `optim_{update/step}_rule!` I think

PR #1165, opened 2020-05-05
Fix docstring of logitcrossentropy

cossio:
Since `y` is a logit, there is no log (see the diff).

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -91,7 +91,7 @@ Return the crossentropy computed after a [`Flux.logsoftmax`](@ref) operation;
 calculated as `-sum(y .* logsoftmax(ŷ) .* weight) / size(y, 2)`.
 
 `logitcrossentropy(ŷ, y)` is mathematically equivalent to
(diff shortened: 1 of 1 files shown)


cossio:
Probably easier to incorporate these changes into https://github.com/FluxML/Flux.jl/pull/1150 and close this one.

src/layers/stateless.jl
```diff
@@ -91,7 +91,7 @@ Return the crossentropy computed after a [`Flux.logsoftmax`](@ref) operation;
 calculated as `-sum(y .* logsoftmax(ŷ) .* weight) / size(y, 2)`.
 
 `logitcrossentropy(ŷ, y)` is mathematically equivalent to
-[`Flux.crossentropy(softmax(log(ŷ)), y)`](@ref) but it is more numerically stable.
```
cossio, line 94: Here `ŷ` is a logit, not a probability. Therefore there is no `log`.

dhairyagandhi96:
Thanks!

PR #1166, opened 2020-05-05
Fix crossentropy when some probabilities are zero

cossio:
Use a function `xlogy(x,y) = x * log(y)` that has the correct limit at `x=0`.

Before this PR:

```julia
julia> Flux.crossentropy([0.1,0.0,0.9], [0.1,0.0,0.9])
NaN
```

After this PR:

```julia
julia> Flux.crossentropy([0.1,0.0,0.9], [0.1,0.0,0.9])
0.3250829733914482
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.10.4"
+version = "0.10.5"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -264,14 +264,19 @@ function flatten(x::AbstractArray)
 end
 
 """
-    xlogx(x::Real)
+    xlogx(x)
 Return `x * log(x)` for `x ≥ 0`, handling `x = 0` by taking the downward limit.
 """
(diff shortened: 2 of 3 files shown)


dhairyagandhi96:
Lgtm 

bors r+

CarloLucibello:
Has this been tested on gpu with `allowscalar=False`? 
Crossentropy is quite ubiquitous and performance sensitive, I'm not sure we want to introduce that branching just to cover the zero case. How does this compare to adding an eps inside the log?

cossio:
Note that `ifelse` should not branch. I have not tested, but I guess it should not make a difference.

CarloLucibello:
Looks like it's not gpu friendly https://gitlab.com/JuliaGPU/Flux.jl/-/jobs/542058740

dhairyagandhi96:
Interesting, I thought `ifelse` would be generic across devices.

cossio:
`ifelse` is used here: https://github.com/JuliaGPU/CuArrays.jl/blob/master/src/nnlib.jl, for example.

dhairyagandhi96:
Correct, I guess it's just the `cufunc` treatment

bors try

cossio:
Looks ok now.

cossio:
@dhairyagandhi96 Can you rerun bors? (I think I cannot do it myself, right?)

cossio:
bors try

dhairyagandhi96:
bors try

cossio:
Merge? Also, would it be ok to register a patch version with this?

PR #1169, opened 2020-05-08
Speed up recurrent layers by seperating broadcasting of tanh/sigmoid from other parts

AStupidBear:
Fusing broadcasting of tanh/sigmoid to other parts can block julia from vectorizing the code. 

```julia
julia> using Random, BenchmarkTools, Flux

julia> using Flux: GRUCell, gate

julia> model = GRUCell(100, 100);

julia> h = x = randn(Float32, 100, 128);

julia> function (m::GRUCell)(h, x)
           b, o = m.b, size(h, 1)
           gx, gh = m.Wi*x, m.Wh*h
           r = σ.(gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1))
           z = σ.(gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2))
           h̃ = tanh.(gate(gx, o, 3) .+ r .* gate(gh, o, 3) .+ gate(b, o, 3))
           h′ = (1 .- z).*h̃ .+ z.*h
           return h′, h′
       end

julia> @btime model($h, $x);
  1.242 ms (42 allocations: 801.44 KiB)

julia> function (m::GRUCell)(h, x)
           b, o = m.b, size(h, 1)
           gx, gh = m.Wi*x, m.Wh*h
           r = σ.((r_ = gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1);))
           z = σ.((z_ = gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2);))
           h̃ = tanh.((h_ = gate(gx, o, 3) .+ r .* gate(gh, o, 3) .+ gate(b, o, 3);))
           h′ = (1 .- z).*h̃ .+ z.*h
           return h′, h′
       end

julia> @btime model($h, $x);
  391.418 μs (34 allocations: 951.34 KiB)
```

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -122,7 +122,7 @@ function (m::LSTMCell)((h, c), x)
   cell = tanh.(gate(g, o, 3))
   output = σ.(gate(g, o, 4))
   c = forget .* c .+ input .* cell
-  h′ = output .* tanh.(c)
+  h′ = output .* (c_ = tanh.(c))
   return (h′, c), h′
 end
 
@@ -160,9 +160,9 @@ GRUCell(in, out; init = glorot_uniform) =
 function (m::GRUCell)(h, x)
   b, o = m.b, size(h, 1)
   gx, gh = m.Wi*x, m.Wh*h
-  r = σ.(gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1))
-  z = σ.(gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2))
-  h̃ = tanh.(gate(gx, o, 3) .+ r .* gate(gh, o, 3) .+ gate(b, o, 3))
+  r = σ.((r_ = gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1);))
+  z = σ.((z_ = gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2));)
+  h̃ = tanh.((h̃_ = gate(gx, o, 3) .+ r .* gate(gh, o, 3) .+ gate(b, o, 3);))
   h′ = (1 .- z).*h̃ .+ z.*h
   return h′, h′
 end


CarloLucibello:
That 2.5x speedup given by avx is something worth considering, I would be in favor, but without it the broadcast separation doesn't seem really worth doing
@dhairyagandhi96 @MikeInnes

AStupidBear:
Later we need to overload broadcasting, which is easier than using `@avx` directly in the cell main body. See https://github.com/chriselrod/LoopVectorization.jl/issues/108

AStupidBear:
The PR will be useful after https://github.com/FluxML/NNlib.jl/pull/199 is merged

MikeInnes:
Unfortunately this is type-piratical. I'm not sure that we should try to undo Julia's 'optimisations' here even though they are annoying, because (a) the type piracy could cause issues, (b) it won't help when we do AD (we split broadcasts then anyway) and (c) where performance matters we'll probably rely on ML-specific compiler passes which do this anyway.

It would be nice to have but I think this should be considered a Base Julia issue; we'll benefit automatically once Base does something like `LoopVectorization` by default.

AStupidBear:
> Unfortunately this is type-piratical. I'm not sure that we should try to undo Julia's 'optimisations' here even though they are annoying, because (a) the type piracy could cause issues, (b) it won't help when we do AD (we split broadcasts then anyway) and (c) where performance matters we'll probably rely on ML-specific compiler passes which do this anyway.
> 
> It would be nice to have but I think this should be considered a Base Julia issue; we'll benefit automatically once Base does something like `LoopVectorization` by default.

That will take a long time (maybe 1-2 years) for julia to have `LoopVectorization` like stuff because this is not an urgent issue. https://github.com/JuliaLang/julia/issues/8869 had been open for six years.

AStupidBear:
Now that FluxML/NNlib.jl#199 is merged, this PR is useful without type-piracy.

AStupidBear:
Close in favor of https://github.com/AStupidBear/FastRecurFlux.jl

PR #1170, opened 2020-05-08
Additional notes in beginner docs around getting params to optimizer.

jimmiebtlr:
I had a little trouble connecting the dots on getting params to the optimizer.

### DIFF
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -134,7 +134,7 @@ model(x) = linear2(σ.(linear1(x)))
 model(rand(5)) # => 2-element vector
 ```
 
-Another (equivalent) way is to create a struct that explicitly represents the affine layer.
+Another way is to create a struct that explicitly represents the affine layer.  This way works better for providing params to the optimizer later on.
 
 ```julia
 struct Affine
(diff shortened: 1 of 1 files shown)


docs/src/models/basics.md
```diff
@@ -134,7 +134,7 @@ model(x) = linear2(σ.(linear1(x)))
 model(rand(5)) # => 2-element vector
 ```
 
-Another (equivalent) way is to create a struct that explicitly represents the affine layer.
+Another way is to create a struct that explicitly represents the affine layer.  This way works better for providing params to the optimizer later on.
```
dhairyagandhi96, line 137: We don't necessarily provide `params` to the optimisers, so maybe that can be a bit confusing?

docs/src/models/basics.md
```diff
@@ -148,6 +148,9 @@ Affine(in::Integer, out::Integer) =
 # Overload call, so the object can be used as a function
 (m::Affine)(x) = m.W * x .+ m.b
 
+# Tell flux where params to optimize come from
+Flux.@functor Affine (W,b)
```
dhairyagandhi96, line 152: Since this syntax can be a bit tricky when moving parameters to a different type, I would mention that `at-functor` would see all the parameters by default, but that can be controlled via the tuple

ToucheSir:
I want to say #1579 covered the first note about `params`, and basics.md already mentions `@functor` later on. Is there anything else I'm missing?

CarloLucibello:
closing as stale and fixed somewhere else

PR #1171, opened 2020-05-09
Add orthogonal kernel initialization for recurrent cells

AStupidBear:


src/utils.jl
```diff
@@ -37,6 +37,24 @@ julia> Flux.glorot_normal(3, 2)
...
+julia> Flux.orthogonal(2)
+2×2 Array{Float32,2}:
+ -0.633973  -0.773356
+ -0.773356   0.633973
+```
+
+See [Exact solutions to the nonlinear dynamics of learning in deep linear neural networks](http://arxiv.org/abs/1312.6120)
+"""
+orthogonal(dim) = Matrix(qr(randn(Float32, dim, dim)).Q)
+orthogonal(dim1, dim2) = vcat([orthogonal(dim2) for n in 1:(dim1 ÷ dim2)]...)
```
findmyway, line 56: Add a warning or error when `dim1 < dim2`?

findmyway: It would be better to align with some other implementations here. For example:

> If the shape of the tensor to initialize is two-dimensional, it is initialized with an orthogonal matrix obtained from the QR decomposition of a matrix of random numbers drawn from a normal distribution. If the matrix has fewer rows than columns then the output will have orthogonal rows. Otherwise, the output will have orthogonal columns

https://www.tensorflow.org/api_docs/python/tf/keras/initializers/Orthogonal

AStupidBear: If I align with tf's implementations, then one cannot pass `orthogonal` to the kernel_init.

AStupidBear: I added `orthogonal_maxtrix` as an internal method for `orthogonal`. If `orthogonal_maxtrix` is called `orthogonal` then it will conflict with the use of `kernel_init` and I cannot think of a better name.

dhairyagandhi96:
Maybe checkout the way we have implemented Flux.convfilter for the higher level API too. Good to mirror and have a consistent API

AStupidBear:
@dhairyagandhi96 If we initialize recurrent cells just like what `Conv` does, the keyword arguments will be too long.

dhairyagandhi96:
My point is related to the API, not the keyword. The keyword can be ideated on separately

darsnack:
This has been merged in #1496.

Issue #1176, opened 2020-05-12
ExpDecay does not decay according to the description


DrChainsaw:
Maybe there is something clever about the current implementation which is desirable, but it does not seem to "Discount the learning rate `η` by the factor `decay` every `decay_step` steps till
a minimum of `clip`." Or maybe I don't understand what "Discount" mean in this context?

```julia
	@testset "Sanity Check" begin
		# Should halve the learning rate in each step
		o = ExpDecay(0.2, 0.5, 1, 1e-3)
		p = [0.0]
		steps = 1:8
		eta_expected = @. max(o.eta * 0.5 ^ steps, o.clip)
		eta_actual = [Optimise.apply!(o, p, [1.0])[1] for _ in steps]
		@test eta_actual == eta_expected
	end

  Expression: eta_actual == eta_expected
   Evaluated: [0.1, 0.07071067811865477, 0.05612310241546866, 0.04719371563408469, 0.04108451572928953, 0.036602142398640646, 0.0331514265332061, 0.030399992168952613] == [0.1, 0.05, 0.025, 0.0125, 0.00625, 0.003125, 0.0015625, 0.001]
```

DrChainsaw:
Fix is merged

PR #1177, opened 2020-05-12
Align ExpDecay implementation with documentation

DrChainsaw:
Fix for #1176

### DIFF
--- a/test/optimise.jl
+++ b/test/optimise.jl
@@ -57,6 +57,17 @@ end
 end
 
 @testset "ExpDecay" begin
+
+	@testset "Sanity Check" begin
+		o = ExpDecay(0.2, 0.5, 1, 1e-3)
+		p = [0.0]
+		steps = 1:8
+		eta_expected = @. max(o.eta * 0.5 ^ steps, o.clip)
+		eta_actual = [Optimise.apply!(o, p, [1.0])[1] for _ in steps]
+		@test eta_actual == eta_expected
+	end
+
+
     w = randn(10, 10)
     o = ExpDecay(0.1, 0.1, 1000, 1e-4)
     w1 = randn(10,10)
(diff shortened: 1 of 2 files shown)


test/optimise.jl
```diff
@@ -57,6 +57,17 @@ end
 end
 
 @testset "ExpDecay" begin
+
+	@testset "Sanity Check" begin
+		o = ExpDecay(0.2, 0.5, 1, 1e-3)
```
dhairyagandhi96, line 62: Could we align the formatting with the rest of the file? Good to go otherwise.

DrChainsaw: Better now? 

That whole testset has a different intendation compared to the others and I didn't want to obscure the actual changes made.

dhairyagandhi96:
Thoughts on keeping the current set if someone wants the closed form lr decay?

DrChainsaw:
Current set?

I was a bit tempted to rewrite so that the current decayed learning rate was not a state and instead compute it for each individual parameter. Is that what you mean? 

This would be an additional change in behaviour as the current implementation will use the decayed learning rate of the "first seen" parameter for all parameters. 

I guess there are ways this approach can cause unexpected behaviour in general cases and that might be a motivation to change the behaviour.

dhairyagandhi96:
But the lr at any step should be the same for all parameters from the classic definition of ExpDecay. That's how it's implemented in other frameworks as well.

DrChainsaw:
> But the lr at any step should be the same for all parameters from the classic definition of ExpDecay. That's how it's implemented in other frameworks as well.

I guess both the current implementation and the one I proposed above fulfills this requirement in the simple case when the optimizer sees the same set of parameters in each pass. 

In case a new parameter suddenly appears the current implementation will give it the same decayed learning rate as all the other parameters. However, I think that this new parameter will disrupt the learning rate schedule as it will have its own stepcount and eventually trigger a decay based on it which is pretty much out of sync with the other parameters (unless of course the new parameter happens to appear right at a decay step so it is accidentally in sync with the rest). This has nothing to do with the change in this PR as it is a consequence of the mechanism to detect steps which I haven't touched.

The alternative I proposed would just let the new parameter follow its own decay path independently of the existing ones, meaning it will not disrupt the decaying process. 

Again, for the normal case when the set of parameters to update never changes there is no difference.

If we care alot about new parameters entering with the same decayed learning rate as old then I think it will be a bit hard to achive without changing the optimizer API, but maybe there is some clever way to do it.

About that "keeping the current set": I guess this is not what you meant then?

dhairyagandhi96:
I think for now, let's merge this and discuss further cleaning up as a separate PR

bors r+

Issue #1178, opened 2020-05-12
Flux.Conv type instability


ltjkoomen:
Hi all,

As I was playing around with Flux I noticed that the convolutional layers I was using (`Flux.Conv`) have some un-inferable types, at least that is my understanding. 

The following is a minimal example:

```
using Flux
imgbatch = rand(Float32, 50, 50, 3, 64)
convlayer = Flux.Conv((3,3), 3=>32)
@code_warntype convlayer(imgbatch)
```

On my laptop, with Julia version 1.4.1 and Flux version 0.10.4 this results in:

```

Variables
  c::Conv{2,4,typeof(identity),Array{Float32,4},Array{Float32,1}}
  x::Array{Float32,4}
  #105::Flux.var"#105#106"
  σ::typeof(identity)
  b::Array{Float32,4}
  cdims::DenseConvDims{2,_A,_B,_C,_D,_E,_F,_G} where _G where _F where _E where _D where _C where _B where _A

Body::Any
1 ─ %1  = Base.getproperty(c, :σ)::Core.Compiler.Const(identity, false)
│   %2  = Base.getproperty(c, :bias)::Array{Float32,1}
│   %3  = Core.tuple(%2)::Tuple{Array{Float32,1}}
│         (#105 = %new(Flux.:(var"#105#106")))
│   %5  = #105::Core.Compiler.Const(Flux.var"#105#106"(), false)
│   %6  = Base.getproperty(c, :stride)::Tuple{Int64,Int64}
│   %7  = Flux.map(%5, %6)::Core.Compiler.Const((1, 1), false)
│   %8  = Core.tuple(Flux.:(:), 1)::Core.Compiler.Const((Colon(), 1), false)
│   %9  = Core._apply_iterate(Base.iterate, Flux.reshape, %3, %7, %8)::Array{Float32,4}
│         (σ = %1)
│         (b = %9)
│   %12 = (:stride, :padding, :dilation)::Core.Compiler.Const((:stride, :padding, :dilation), false)
│   %13 = Core.apply_type(Core.NamedTuple, %12)::Core.Compiler.Const(NamedTuple{(:stride, :padding, :dilation),T} where T<:Tuple, false)
│   %14 = Base.getproperty(c, :stride)::Tuple{Int64,Int64}
│   %15 = Base.getproperty(c, :pad)::NTuple{4,Int64}
│   %16 = Base.getproperty(c, :dilation)::Tuple{Int64,Int64}
[...]

PR #1180, opened 2020-05-13
Add bilinear upsample layer

ltjkoomen:
Hi,

I've implemented a bilinear upsampling layer that upsamples the first 2 dimensions of a 4-dimensional array with an integer factor. I have seen the implementation in #1136, but:
- This one is about 10 times faster on CPU on my laptop
- This one uses broadcasting instead of scalar loop indexing, which makes it usable on the GPU
- This one is directly compatible with Zygote

Some possible TODOs:

- Currently the utility functions in src/layers/upsampling.jl used by the implementation are all added to the Flux namespace when doing `using Flux`, which seems like pollution and is unnecessary. However, I am not quite sure what the correct way is in Julia to shield these "private" functions.
- With this implementation it is no problem to define an output size, instead of the quite limiting integer upscaling factors. Although usually upscaling is used to, for instance, increase the number of samples in each dimension by 2, maybe we should still expose this functionality somehow? This could be implemented through some specialized constructor perhaps.

### DIFF
--- a/src/layers/upsample.jl
+++ b/src/layers/upsample.jl
@@ -273,9 +273,9 @@ function bilinear_upsample_adjoint(arr::AbstractArray, factors::Tuple{T,T} where
             conv_extra1 = gpu(conv_extra1)
         end
 
-        arr_ds[[1],:,:,:] .+= conv_extra1(arr[1:nextras[1],:,:,:])
-        conv_extra1.weight .= conv_extra1.weight[end:-1:1,:,:,:]
-        arr_ds[[end],:,:,:] .+= conv_extra1(arr[end-nextras[1]+1:end,:,:,:])
+        arr_ds[[1],:,:,:] .+= conv_extra1(@view(arr[1:nextras[1],:,:,:]))
+        conv_extra1.weight .= @view(conv_extra1.weight[end:-1:1,:,:,:])
+        arr_ds[[end],:,:,:] .+= conv_extra1(@view(arr[end-nextras[1]+1:end,:,:,:]))
     end
 
     # Second dimension edge-effect correction
@@ -293,9 +293,9 @@ function bilinear_upsample_adjoint(arr::AbstractArray, factors::Tuple{T,T} where
             conv_extra2 = gpu(conv_extra2)
         end
 
-        arr_ds[:,[1],:,:] .+= conv_extra2(arr[:,1:nextras[2],:,:])
-        conv_extra2.weight .= conv_extra2.weight[:,end:-1:1,:,:]
-        arr_ds[:,[end],:,:] .+= conv_extra2(arr[:,end-nextras[2]+1:end,:,:])
+        arr_ds[:,[1],:,:] .+= conv_extra2(@view(arr[:,1:nextras[2],:,:]))
+        conv_extra2.weight .= @view(conv_extra2.weight[:,end:-1:1,:,:])
+        arr_ds[:,[end],:,:] .+= conv_extra2(@view(arr[:,end-nextras[2]+1:end,:,:]))
     end
 
     # Finally fix four corners if needed
--- a/src/layers/upsample.jl
+++ b/src/layers/upsample.jl
@@ -130,6 +130,10 @@ end
 Does the heavy lifting part of the bilinear upsampling operation
 """
 function bilinear_upsample_workhorse(img, ilow1, ihigh1, wdiff1, ilow2, ihigh2_r, wdiff2)
+    if typeof(img) <: CuArray
+        wdiff1 = CuArray(wdiff1)
+        wdiff2 = CuArray(wdiff2)
+    end
     imgupsampled = @view(img[ilow1,ilow2,:,:]) .* (1 .- wdiff1) .+ @view(img[ihigh1,ilow2,:,:]) .* wdiff1
     imgupsampled = imgupsampled .* (1 .- wdiff2) .+ @view(imgupsampled[:,ihigh2_r,:,:]) .* wdiff2
 end
--- a/src/layers/upsample.jl
+++ b/src/layers/upsample.jl
@@ -17,15 +17,16 @@ julia> b(rand(2, 2, 1, 1))
  0.888679  0.710044  0.352773  0.174138
  0.910357  0.7271    0.360586  0.177329```
 """
-struct BilinearUpsample2d{}
-    factors::Tuple{T,T} where T<:Integer
-    BilinearUpsample2d(factors::Tuple{R,R}) where R<:Integer = new(factors)
-    BilinearUpsample2d(factors::F) where F<:Integer = new((factors, factors))
+
+struct BilinearUpsample2d{T<:Integer}
+    factors::Tuple{T,T}
 end
+
+BilinearUpsample2d(factor::F) where F<:Integer = BilinearUpsample2d((factor, factor))
 
 @functor BilinearUpsample2d
 
-function (c::BilinearUpsample2d)(x::AbstractArray)
+function (c::T where T<:BilinearUpsample2d)(x::AbstractArray)
     bilinear_upsample2d(x, c.factors)
 end
 
@@ -81,9 +82,12 @@ Determines the adjoint of the vector of indices `idx`, based on the following as
 * `idx[1] == 1`
 * `all(d in [0,1] for d in diff(idx))`
 
-The adjoint of `idx` can be seen as an inverse operation:
-```jldoctest
-x[idx][idx_adjoint] == x
+The adjoint of `idx` can be seen as an inverse operation such that:
+```
+x = [1, 2, 3, 4, 5]
+idx = [1, 2, 2, 3, 4, 4, 5]
+idx_adjoint = adjoint_of_idx(idx)
+@assert x[idx][idx_adjoint] == x
 ```
 
 The above holds as long as `idx` contains every index in `x`.
--- a/src/layers/upsample.jl
+++ b/src/layers/upsample.jl
@@ -1,0 +1,162 @@
+"""
+    BilinearUpsample2d(factors::Tuple{Integer,Integer})
+
+Create an upsampling layer that uses bilinear interpolation to upsample the 1st and 2nd dimension of
+a 4-dimensional input array . The size of the output array will be equal to
+`(factors[1]*S1, factors[2]*S2, S3, S4)`, where `S1,S2,S3,S4 = size(input_array)`.
+
+# Examples
+```jldoctest; setup = :(using Flux: BilinearUpsample2d; using Random; Random.seed!(0))
+julia> b = Flux.BilinearUpsample2d((2, 2))
+BilinearUpsample2d(2, 2)
+julia> b(rand(2, 2, 1, 1))
+4×4×1×1 Array{Float64,4}:
+[:, :, 1, 1] =
+ 0.823648  0.658877  0.329336  0.164566
+ 0.845325  0.675933  0.337149  0.167757
+ 0.888679  0.710044  0.352773  0.174138
+ 0.910357  0.7271    0.360586  0.177329```
+"""
+struct BilinearUpsample2d{}
+    factors::Tuple{T,T} where T<:Integer
+    BilinearUpsample2d(factors::Tuple{R,R}) where R<:Integer = new(factors)
+    BilinearUpsample2d(factors::F) where F<:Integer = new((factors, factors))
+end
+
+@functor BilinearUpsample2d
+
+function (c::BilinearUpsample2d)(x::AbstractArray)
+    bilinear_upsample2d(x, c.factors)
+end
+
+function Base.show(io::IO, l::BilinearUpsample2d)
+  print(io, "BilinearUpsample2d( $(l.factors[1]), $(l.factors[2]) )")
+end
+
+"""
+    `construct_xq(n::T, m::T) where T<:Integer`
+
+Creates interpolation points for resampling, creates the same grid as used in Image.jl `imresize`.
+"""
+@nograd function construct_xq(n::T, m::T) where T<:Integer
+    typed1 = one(n)
+    typed2 = 2typed1
+    step = n // m
+    offset = (n + typed1)//typed2 - step//typed2 - step*(m//typed2 - typed1)
+    x = range(offset, step=step, length=m)
+    xq = clamp.(x, typed1//typed1, n//typed1)
+    return xq
+end
+
+"""
+    `get_inds_and_ws(xq, dim, n_dims)`
+
+Creates interpolation lower and upper indices, and broadcastable weights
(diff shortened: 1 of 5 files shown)


kczimm:
@ltjkoomen this is awesome. I hope others can help review this PR so it can be merged quickly. Thank you for this contribution. I'm happy to close #1136 in favor of this one.

dhairyagandhi96:
Could we add GPU tests, I'd like to ensure that it doesn't cause trouble there

src/layers/upsample.jl
```diff
@@ -0,0 +1,162 @@
...
+julia> b(rand(2, 2, 1, 1))
+4×4×1×1 Array{Float64,4}:
+[:, :, 1, 1] =
+ 0.823648  0.658877  0.329336  0.164566
+ 0.845325  0.675933  0.337149  0.167757
+ 0.888679  0.710044  0.352773  0.174138
+ 0.910357  0.7271    0.360586  0.177329```
+"""
+struct BilinearUpsample2d{}
+    factors::Tuple{T,T} where T<:Integer
```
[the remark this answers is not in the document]
ltjkoomen: I have noticed a problem with the gradient computation when the input is a CuArray. Although the upsampled output array is calculated correctly, the gradient is wrong. Anyone have experience getting `gradient` to work with CuArrays?

ltjkoomen:
I have noticed a problem when using `gradient` when the input is a `CuArray`. Although the upsampled output array is correct (equal to the CPU output), the output of `gradient` is wrong.

```
using Flux: gradient, BilinearUpsample2d
using CuArrays

x = Float32.([1 2; 3 4])[:,:,:,:]
x_c = CuArray(x)

c = BilinearUpsample2d((2,2))

o = c(x)
o_c = c(x_c)

g = gradient(x -> sum(c(x)), x)
g_c = gradient(x -> sum(c(x)), x_c)
```
The result is that `o` and `o_c` are equal, but `g` and `g_c` are not.
Anyone have experience getting Zygote to work with CuArrays?

mcabbott:
I haven't looked hard but I suspect that the problem is aliasing: won't you end up writing to two views of the same object, with shifted indices? When this gets done in parallel, I think that CuArrays is not careful about the order.

ltjkoomen:
Even when I remove the `@views` and do it like this:
```
v1 = img[ilow1,ilow2,:,:]
v2 = img[ihigh1,ilow2,:,:]
imgupsampled1 = v1 .* (1 .- wdiff1) .+ v2 .* wdiff1
v3 = imgupsampled1[:,ihigh2_r,:,:]
imgupsampled2 = imgupsampled1 .* (1 .- wdiff2) .+ v3 .* wdiff2
```
It still doesn't work. Wondering if it happens because the indices themselves are no CuArrays, which I did because it leads to a scalar indexing warning and slowdown. That's mentioned [here](https://github.com/JuliaGPU/CuArrays.jl/issues/1).

ltjkoomen:
I have located the problem, it seems `gradient` doesn't work with a `CuArray` generated through indexing with an array like `idx=[1, 1, 2]`, like so: `out = x[idx]`. I have included a minimal example:
```
using Flux
using CuArrays

function min_example(x, idx)
    out = x[idx]
    return out
end

x = Float32.([1, 2])
idx = [1, 1, 2]
x_c = CuArray(x)

# CPU version
fex(x) = min_example(x, idx)
f(x) = sum(fex(x))
df(x) = gradient(f, x)[1]

# CuArray version
fex_c(x_c) = min_example(x_c, idx)
f_c(x_c) = sum(fex_c(x_c))
df_c(x_c) = gradient(f, x_c)[1]

@assert fex(x) == cpu(fex_c(x_c)) # pass, forward is the same
@assert df(x) == cpu(df_c(x_c)) # fail, gradient is wrong for the CuArray version
```

Unsure whether this is a bug or expected behavior. Or is this actually the possible issue brought up by @mcabbott ?

mcabbott:
The issue I should have linked is https://github.com/JuliaGPU/CuArrays.jl/issues/684 . Was a bit confused by views etc, but the core idea is that you want to accumulate at indices like `idx=[1, 1, 2]`, which can easily go wrong. (I think the same happens for `@strided`, threaded broadcasting on CPU.) 

Surely there is some way around this, though. If you don't want to up/down-sample by factors larger than 2, can you divide up the work even/odd & be sure that there are no collisions?

ltjkoomen:
What about defining a custom adjoint function instead? The adjoint of upsampling is a downsampling operation, which could be implemented efficiently using the existing convolutional layer code. The forward part seems to work fine even with this current implementation, both the normal- and CUDA-array versions.

ltjkoomen:
I have fixed some type-instabilities, mostly by hardcoding the fact that there are 4 dimensions to the input array and the first 2 are upsampled. This has fixed all type-instability issues in the forward pass. However, the backward pass is still not type-stable because of the fact that I have used the Flux.Conv layer to implement the downsampling operation in the custom adjoint. The type of the weights of Flux.Conv cannot be inferred, an issue which I have already brought up in #1178.

andevellicus:
Is there any way to make this work for more than 2D? I work with medical imaging so a lot of that is in 3 dimensions....

ltjkoomen:
That would be trilinear upsampling. I guess the same method I use for bilinear upsampling could be used, but I haven't tried it yet.

test/runtests.jl
```diff
@@ -30,6 +30,7 @@ Random.seed!(0)
     include("layers/normalisation.jl")
     include("layers/stateless.jl")
     include("layers/conv.jl")
+include("layers/upsample.jl")
```
CarloLucibello, line 33: indentation

CarloLucibello:
I agree this should be merged and type instabilities issues addressed separately

src/layers/upsample.jl
```diff
@@ -0,0 +1,325 @@
...
+    ilow1, ihigh1, wdiff1 = get_inds_and_ws(xq1, 1)
+    ilow2, ihigh2, wdiff2 = get_inds_and_ws(xq2, 2)
+
+    # Adjust the upper interpolation indices of the second dimension
+    ihigh2_r = adjoint_of_idx(ilow2)[ihigh2]
+
+    wdiff1 = eltype(img).(wdiff1)
+    wdiff2 = eltype(img).(wdiff2)
+
+    if typeof(img) <: CuArray
```
dhairyagandhi96, line 176: Why do we need this? The kernel should handle this case generically.

src/layers/upsample.jl
```diff
@@ -0,0 +1,325 @@
...
+end
+
+"""
+    `construct_xq(n::T, m::T) where T<:Integer`
+
+Creates interpolation points for resampling, creates the same grid as used in Image.jl `imresize`.
+"""
+@nograd function construct_xq(n::T, m::T) where T<:Integer
+    typed1 = one(n)
+    typed2 = 2typed1
```
dhairyagandhi96, line 48: Use `T` and broadcasting perhaps.

src/layers/upsample.jl
```diff
@@ -0,0 +1,325 @@
...
+
+Creates interpolation lower and upper indices, and broadcastable weights
+"""
+@nograd function get_inds_and_ws(xq, dim)
+    n = length(xq)
+
+    ilow = floor.(Int, xq)
+    ihigh = ceil.(Int, xq)
+
+    wdiff = xq[:,:,:,:] .- ilow[:,:,:,:]
```
dhairyagandhi96, line 67: Do we need the colons there if it's just everything?

src/layers/upsample.jl
```diff
@@ -0,0 +1,325 @@
...
+
+# Outputs
+- `imgupsampled::AbstractArray`: the upsampled version of `img`. The size of `imgupsampled` is
+equal to `(k_upsample[1]*S1, k_upsample[2]*S2, S3, S4)`, where `S1,S2,S3,S4 = size(img)`.
+
+# Explanation
+Upsamples the first two dimensions of the 4-dimensional array `img` by the two upsample factors stored in `k_upsample`,
+using bilinear interpolation. The interpolation grid is identical to the one used by `imresize` from `Images.jl`.
+"""
+function bilinear_upsample2d(img::AbstractArray{T,4}, k_upsample::NTuple{2,<:Real}) where T
```
dhairyagandhi96, line 133: What would be needed to get us to be able to do it with 3d convs as well?

src/layers/upsample.jl
```diff
@@ -0,0 +1,325 @@
...
+x = [1, 2, 3, 4, 5]
+idx = [1, 2, 2, 3, 4, 4, 5]
+idx_adjoint = adjoint_of_idx(idx)
+@assert x[idx][idx_adjoint] == x
+```
+
+The above holds as long as `idx` contains every index in `x`.
+ """
+@nograd function adjoint_of_idx(idx::Vector{T}) where T<:Integer
+    d = trues(size(idx))
```
dhairyagandhi96, line 107: It might be type unstable to use bools with floats and that might stop julia from using blas calls more efficiently

src/layers/upsample.jl
```diff
@@ -0,0 +1,325 @@
...
+
+    # Get linear interpolation lower- and upper index, and weights
+    ilow1, ihigh1, wdiff1 = get_inds_and_ws(xq1, 1)
+    ilow2, ihigh2, wdiff2 = get_inds_and_ws(xq2, 2)
+
+    # Adjust the upper interpolation indices of the second dimension
+    ihigh2_r = adjoint_of_idx(ilow2)[ihigh2]
+
+    wdiff1 = eltype(img).(wdiff1)
+    wdiff2 = eltype(img).(wdiff2)
```
dhairyagandhi96, line 174: Use `T` from the method signature

src/layers/upsample.jl
```diff
@@ -0,0 +1,325 @@
+"""
+    BilinearUpsample2d(factors::Tuple{Integer,Integer})
```
MikeInnes, line 2: Is it necessary for this to be specific to 2D? Could it infer its dimension from `factors`, like the `Conv` layer, and be generic across dimension numbers? Also, could it not be considered `BilinearInterpolate` in general (e.g. with fractional factors)?

(It is fine if dimensions other than 2, or fractional factors, are not currently implemented and throw an error, just nice if we can add them in future.)

[If we do keep `2D` it should be capitalised that way.]

src/layers/upsample.jl
```diff
@@ -0,0 +1,325 @@
...
+function (c::T where T<:BilinearUpsample2d)(x::AbstractArray)
+    bilinear_upsample2d(x, c.factors)
+end
+
+function Base.show(io::IO, l::BilinearUpsample2d)
+  print(io, "BilinearUpsample2d( $(l.factors[1]), $(l.factors[2]) )")
+end
+
+@adjoint function (c::T where T<:BilinearUpsample2d)(x::AbstractArray)
+    (c::T where T<:BilinearUpsample2d)(x), c̄ -> (nothing, bilinear_upsample_adjoint(c̄, c.factors))
```
MikeInnes, line 38: The adjoint should probably be applied to the `bilinear_upsample` function, not the layer, so that the function can be used directly where appropriate. As with `Conv` it would be nice to support a non-layer version of this with a nice API.

The type on the second `c::T where ...` seems unnecessary.

src/layers/upsample.jl
andevellicus: +1 to this....

AlfTetzlaff:
Hi everybody, just recently I blindly translated the pytorch bilinear upsampling to julia. Unfortunately I didn't have the time to finish it. The array indexing is still messed up and honestly I don't really know how the gradient calculation works. I've put everything in a [gist](https://gist.github.com/AlfTetzlaff/fdee18829758f56a3b18c494d42c34c5), so that you can just copy it, if interested :)

maxfreu:
Hi! I would really like to give this a push. I could really need this (or nearest neighbor) for U-Net implementations. Have the reviews been addressed? If not, what would I have to do to work on them? In the mean time I have finished porting and testing the pytorch implementation core of bilinear upsampling [here](https://gist.github.com/maxfreu/815c8e0c8304bab39d7f78033289cece). It now produces the same output as pytorch for forward and backward (verified via nyan cat). Fractional upscaling factors are allowed, but still only 4D tensors - but in my opinion we dont need more for a start.

DhairyaLGandhi:
I think the pr is in a good shape. Have you been able to benchmark the run time against pytorch as well? Also, there is UNet.jl if that helps.

maxfreu:
Well, benchmarking is a science for itself... It seems that timings are on par in many cases, but there is some caching going on in pytorch. This is how I benchmark:

maxfreu:
I just checked the changes in this PR, looks good indeed. Forward and backward pass work. However for the image size given in the above post the pytorch gpu kernel is another 3x faster. So maybe it makes sense to add it as dispatch for CuArrays. The pullback should also be straightforward to implement. The question is where this should happen... merge first, then add dispatch? Or the other way round? The kernel I ported is by the way from Caffee2, there are also the newer ATen kernels. As soon as I have figured out how to port them, I could output NearestNeighbour2d/3d & trilinear as well.

CarloLucibello:
@maxfreu cpu implementations shoud go to NNlib, kernel's should go to CUDA.jl is @maleadt agrees, otherwise, we can add the to Flux. A simple layer wrapper should live in Flux.

maxfreu:
I just noted that there is an older, good attempt by @avik-pal [here](https://github.com/FluxML/NNlib.jl/pull/95) for CPU and [here](https://github.com/JuliaGPU/CuArrays.jl/pull/293) for GPU, but I don't know which method is used. He implemented UpsamplingDims along the lines of ConvDims to produce specialized code. Should we consider that or is it over the top?

@ltjkoomen are you still enganged? Some time early next year might be sufficient :wink:

CarloLucibello:
@ltjkoomen has not been responsive for some time. If the cpu implementation here is fine I will move it to NNlib. Any performance optimization can come later, the important thing is to agree on the interface

CarloLucibello:
This should be fine:
```
NNlib.bilinear_upsample2d(x::AbstractArray{<:Real,4}, k::NTuple{2,<:Real})
```
CUDA.jl will have to overload it.

maleadt:
> @maxfreu cpu implementations shoud go to NNlib, kernel's should go to CUDA.jl if @maleadt agrees, otherwise, we can add them to Flux. A simple layer wrapper should live in Flux.

GPU implementations of NNlib interfaces can go in CUDA.jl, we already have a couple.

qin-yu:
Indeed we need to finish this layer! I'm also writing a UNet (with `repeat()`, see https://github.com/FluxML/Flux.jl/pull/270) right now 😆 I guess the GPU support for `repeat()` has not even been done (see https://github.com/JuliaGPU/GPUArrays.jl/pull/126)!

PR #1187, opened 2020-05-19
LayerNorm regularization

sdobber:
This PR adds regularisation to the `LayerNorm` layer by changing the return value of the `normalise`-function to `(x .- μ′) ./ (σ′ .+ ϵ)`, with a standard value of `ϵ = sqrt(eps(Float32))`. The value of `ϵ` can be changed via a keyword argument. This avoids `NaN`s in the output.

Related discussion on Slack can be found [here](https://julialang.slack.com/archives/C7LFJTXV5/p1589714643448000). This type of regularisation is in line with other ML frameworks, see for example [PyTorch](https://pytorch.org/docs/stable/nn.html?highlight=layernorm#torch.nn.LayerNorm).

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -104,24 +104,19 @@ A [normalisation layer](https://arxiv.org/pdf/1607.06450.pdf) designed to be
 used with recurrent hidden states of size `h`. Normalises the mean and standard
 deviation of each input before applying a per-neuron gain/bias.
 """
-struct LayerNorm{T,N}
+struct LayerNorm{T}
   diag::Diagonal{T}
-  ϵ::N
 end
 
-trainable(ln::LayerNorm) = trainable(ln.diag)
-
-LayerNorm(h::Integer; ϵ = sqrt(eps(Float32))) =
-  LayerNorm(Diagonal(h), ϵ)
+LayerNorm(h::Integer) =
+  LayerNorm(Diagonal(h))
 
 @functor LayerNorm
 
-(a::LayerNorm)(x) = a.diag(normalise(x; ϵ = a.ϵ))
+(a::LayerNorm)(x) = a.diag(normalise(x))
 
 function Base.show(io::IO, l::LayerNorm)
-  print(io, "LayerNorm(", length(l.diag.α))
-  (l.ϵ == sqrt(eps(Float32))) || print(io, "; ϵ = $(l.ϵ)")
-  print(io, ")")
+  print(io, "LayerNorm(", length(l.diag.α), ")")
 end
 
 """
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -104,19 +104,24 @@ A [normalisation layer](https://arxiv.org/pdf/1607.06450.pdf) designed to be
 used with recurrent hidden states of size `h`. Normalises the mean and standard
 deviation of each input before applying a per-neuron gain/bias.
 """
-struct LayerNorm{T}
+struct LayerNorm{T,N}
   diag::Diagonal{T}
+  ϵ::N
 end
 
-LayerNorm(h::Integer) =
-  LayerNorm(Diagonal(h))
+trainable(ln::LayerNorm) = trainable(ln.diag)
+
+LayerNorm(h::Integer; ϵ = sqrt(eps(Float32))) =
+  LayerNorm(Diagonal(h), ϵ)
 
 @functor LayerNorm
 
-(a::LayerNorm)(x) = a.diag(normalise(x))
+(a::LayerNorm)(x) = a.diag(normalise(x; ϵ = a.ϵ))
 
 function Base.show(io::IO, l::LayerNorm)
-  print(io, "LayerNorm(", length(l.diag.α), ")")
+  print(io, "LayerNorm(", length(l.diag.α))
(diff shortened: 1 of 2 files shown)


src/layers/normalise.jl
```diff
@@ -104,19 +104,24 @@ A [normalisation layer](https://arxiv.org/pdf/1607.06450.pdf) designed to be
 used with recurrent hidden states of size `h`. Normalises the mean and standard
 deviation of each input before applying a per-neuron gain/bias.
 """
-struct LayerNorm{T}
+struct LayerNorm{T,N}
```
dhairyagandhi96, line 107: Do we need to store this value?

src/layers/normalise.jl
```diff
@@ -104,19 +104,24 @@ A [normalisation layer](https://arxiv.org/pdf/1607.06450.pdf) designed to be
...
-  LayerNorm(Diagonal(h))
+trainable(ln::LayerNorm) = trainable(ln.diag)
+
+LayerNorm(h::Integer; ϵ = sqrt(eps(Float32))) =
+  LayerNorm(Diagonal(h), ϵ)
 
 @functor LayerNorm
 
-(a::LayerNorm)(x) = a.diag(normalise(x))
+(a::LayerNorm)(x) = a.diag(normalise(x; ϵ = a.ϵ))
```
dhairyagandhi96, line 119: Can just be picked up from the type of `x` here

src/layers/stateless.jl
```diff
@@ -151,10 +151,10 @@ logitbinarycrossentropy(ŷ, y) = (1 - y)*ŷ - logσ(ŷ)
 CuArrays.@cufunc logitbinarycrossentropy(ŷ, y) = (1 - y)*ŷ - logσ(ŷ)
 
 """
-    normalise(x; dims=1)
+    normalise(x; dims=1, ϵ = sqrt(eps(Float32))
```
dhairyagandhi96, line 154: Should take `T` from `x`

sdobber: Are you suggesting sthg like `ϵ = sqrt(eps(eltype(x)))`? In that case, the example from the documentation will fail due to the lack of `eps(Int64)`.

dhairyagandhi96:
```suggestion
    normalise(x::AbstractArray{T}; dims=1, ϵ = sqrt(eps(T)) where T
```

dhairyagandhi96: Hmm, it should really just be 0 in that case, and if we don't have that, we will not be type stable.

sdobber: I see your point with the type stability. On the other hand, `mean` uses the right division operator and will thus always return Floats. So I guess `eps(eltype(x/1))` is sort of what we need - we will get the correct Float type for `typeof(x)` being `Float32`, `Float64` etc, and when using integers etc. we get the type that `mean` spits out. I'm just wondering if there is a more elegant solution...

dhairyagandhi96:
Can we add tests showing how the `eps` can be used?

dhairyagandhi96:
bors try

sdobber:
Sorry for the newbie question - does the error on the first build indicate a problem that needs to be fixed by my side?

DhairyaLGandhi:
Probably not.

but it looks decent to me.

bors try

CarloLucibello:
should we use pytorch's default `ϵ = 1f-5`? For comparison, the proposed default here is much larger, `sqrt(eps(Float32)) = 3.4f-4`

sdobber:
I'm fine with either default. The choice of `sqrt(eps(Float32))` is based on a discussion with Chris Rackauckas on Slack, which of course is not accessible any more by now.

CarloLucibello:
I suggest we go with pytorch's default, a lot of thought has gone in that library so we should just leverage it. Moreover, the epsilon here should not necessarily be linked to the float type used

ToucheSir:
I was curious to see why PyTorch uses `ϵ = 1f-5` and fell down a bit of a rabbit hole. TL;DR frameworks are not entirely consistent and few justify their choice.

- The [PyTorch feature request](https://github.com/pytorch/pytorch/issues/1959) includes examples with both `1e-5` and `1e-6`
- [Flax](https://github.com/google/flax/blob/ed42d0670c293d0ac205313b19fbef9832e81304/flax/nn/normalization.py#L133), [Trax](https://github.com/google/trax/blob/af3a38917bd1bc69cf5d25ce007e16185f22f050/trax/layers/normalization.py#L147) and [this Knet issue](https://github.com/denizyuret/Knet.jl/issues/492#issuecomment-535616047) use `1e-6`
- [Haiku](https://github.com/deepmind/dm-haiku/blob/d9f768106c2350c571144b4d14d28f99f52442d8/haiku/_src/layer_norm.py#L49) uses `1e-5`
- TensorFlow [Swift](https://github.com/tensorflow/swift-apis/blob/e9df616445bad8e66b48ae47674dacbc34949466/Sources/TensorFlow/Layers/Normalization.swift#L228) and [Python](https://github.com/tensorflow/tensorflow/blob/de0a617f4ef321ebf579a4539cec85690f55fb4c/tensorflow/python/keras/layers/normalization.py#L1070) use `1e-3`. The Python version used to use `1e-12` (!) but was updated (ostensibly to match batchnorm)

PR #1191, opened 2020-05-26
Pull Request Template

MikeInnes:
Hopefully makes it a little clearer what the requirements are, which will lead to easier review, and encourage things like NEWS.md that we want to be better in sync.

cc @dhairyagandhi96 and @CarloLucibello for thoughts.


CarloLucibello:
looks fine,  I like the conciseness compared to other templates you see around

MikeInnes:
We could make it even more concise if we split into "bug fix" and "feature" templates. Not sure if that's worthwhile at this stage.

dhairyagandhi96:
Seems good, easy checklist too

MikeInnes:
Alright, let's run with it for now and we can tweak as we go along.

bors r+

PR #1202, opened 2020-06-01
Fix erroneous comment in example in docs Basics

natema:
This commit fixes an erroneous comment in the doc page [Flux Basics](https://fluxml.ai/Flux.jl/stable/models/basics/).

In the example given in the [Simple Models](https://fluxml.ai/Flux.jl/stable/models/basics/#Simple-Models-1) section, it is written
```
loss(x, y) # ~ 3
```
However, averaging over $10^6$ independent runs of the given code block, I get that it should be `loss(x, y) # ~ 5.77`.

### DIFF
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -72,7 +72,7 @@ function loss(x, y)
 end
 
 x, y = rand(5), rand(2) # Dummy data
-loss(x, y) # ~ 3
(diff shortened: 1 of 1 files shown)


CarloLucibello:
i think this is an irrelevant change, if you don't mind I'll close this PR for the sake of some house cleaning

PR #1208, opened 2020-06-03
Fixing output format for `onehot`

natema:
Currently `Flux.OneHotVector` is displayed as a binary vector (0/1) rather than a boolean one (true/false). This is also shown in successive examples in the same page. 
I fixed the `onehot(:b, [:a, :b, :c])` and `onehot(:c, [:a, :b, :c])` outputs in the first example of the page accordingly.


dhairyagandhi96:
thanks!

bors r+

PR #1218, opened 2020-06-10
Require weight and bias to be AbstractArrays

### LINKED ISSUE #1199: Passing Number of hidden units as a float has unexpected behaviour
When the number of hidden units is passed as a `float` to the `Dense` function, the returned layer is always `Dense(1,1)`. Without a warning that the number of hidden has to be an integer it's easy to miss this.

```Julia
julia> using Flux

julia> num_features = 123
123

julia> num_hiddens = 45.0
45.0

julia> Dense(num_features,num_hiddens)
Dense(1, 1)

julia> Dense(num_features,Int(num_hiddens))
Dense(123, 45)
```

This is with `v0.10.5` and Julia 1.4.1

oxinabox:
closes #1199
While in theory someone could be using Dense with weights and biases that are not abstract arrays, I would be surprised.
So allowing it is just leaving a food-gun laying around.
If it is common then we can instead close #1199 by adding a special constructor for `Number` subtypes that error if they are not integers, or something a long those lines.

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

I think this is a bug-fix thus the following are not required:

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.11.0"
+version = "0.11.1"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.11.1"
+version = "0.11.0-DEV"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/test/layers/basic.jl
+++ b/test/layers/basic.jl
@@ -28,6 +28,13 @@ import Flux: activations
   end
 
   @testset "Dense" begin
+    @testset "constructors" begin
+      @test size(Dense(10, 100).W) == (100, 10)
+      @test Dense(rand(100,10), rand(10)).σ == identity
+
+      @test_throws MethodError Dense(10, 10.5)
+    end
+
     @test  length(Dense(10, 5)(randn(10))) == 5
     @test_throws DimensionMismatch Dense(10, 5)(randn(1))
     @test_throws MethodError Dense(10, 5)(1) # avoid broadcasting
@@ -37,7 +44,6 @@ import Flux: activations
     @test Dense(10, 1, identity, initW = ones, initb = zeros)(ones(10,2)) == 10*ones(1, 2)
     @test Dense(10, 2, identity, initW = ones, initb = zeros)(ones(10,1)) == 10*ones(2, 1)
     @test Dense(10, 2, identity, initW = ones, initb = zeros)([ones(10,1) 2*ones(10,1)]) == [10 20; 10 20]
-
   end
 
   @testset "Diagonal" begin
--- a/test/layers/basic.jl
+++ b/test/layers/basic.jl
@@ -33,7 +33,7 @@ import Flux: activations
       @test Dense(rand(100,10), rand(10)).σ == identity
 
       @test_throws MethodError Dense(10, 10.5)
-      @test_throws MethodError Dense(10, 10.5; σ=tanh)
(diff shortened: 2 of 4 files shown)


dhairyagandhi96:
Generally lgtm

I am fine with erroring out, better if we can give a meaningful message too. Thoughts?

dhairyagandhi96:
I have only seen instances of it in the past couple weeks, but generally we can support `Number` just fine.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.11.0"
+version = "0.11.1"
```
dhairyagandhi96, line 3: Since v0.11 isn't released yet, we don't need the bump. Moving on we should tag it with a `-dev`

oxinabox:
```suggestion
version = "0.11.0-DEV"
```

oxinabox:
> I have only seen instances of it in the past couple weeks, but generally we can support Number just fine.

Really?
I expect it works for forward, but not that anyone would ever use it.
Since you can't train it since almost all numbers are immutable.
and only works for bias for single output and for weight for single-input and single-output ford weight.
I have only ever seen 1 single input single output neural net being used ever, and I am pretty sure it had a big hidden layer.

MikeInnes:
I think this would be reasonable, you can always get the one->one case with an explicit matrix. Technically breaking but I imagine no-one is dependent on this.

I agree for general `Number` it would be helpful to either throw an error (you should pass an `Int`) or just try to convert to `Int` ourselves (which works fine in the common case eg `convert(Int, 1.0)`).

oxinabox:
> I agree for general Number it would be helpful to either throw an error (you should pass an Int) or just try to convert to Int ourselves (which works fine in the common case eg convert(Int, 1.0)).

We do thow an error.

test/layers/basic.jl
```diff
@@ -28,6 +28,13 @@ import Flux: activations
   end
 
   @testset "Dense" begin
+    @testset "constructors" begin
+      @test size(Dense(10, 100).W) == (100, 10)
+      @test Dense(rand(100,10), rand(10)).σ == identity
+
+      @test_throws MethodError Dense(10, 10.5)
```
oxinabox, line 35: lets also even test amost hitting the number constructor
```suggestion
      @test_throws MethodError Dense(10, 10.5)
      @test_throws MethodError Dense(10, 10.5; σ=tanh)
```

oxinabox:
@dhairyagandhi96 @MikeInnes  you have both looked at this PR and said you approve of it, but seem to have not hit approve.

MikeInnes:
In particular I was thinking of an error that says "you should pass integers", just to be a bit more informative.

oxinabox:
I would say the method error is better,
It says that and more.
and is a standard format for this that all julia users should be familar with.
(its not like it is a method error coming from a long way down the call stack either)

```
julia> Dense(10,10.5)
ERROR: MethodError: no method matching Dense(::Int64, ::Float64, ::typeof(identity))
Closest candidates are:
  Dense(::Integer, ::Integer, ::Any; initW, initb) at /Users/oxinabox/JuliaEnvs/Flux.jl/src/layers/basic.jl:115
  Dense(::Any, ::Any) at /Users/oxinabox/JuliaEnvs/Flux.jl/src/layers/basic.jl:111
  Dense(::S, ::T, ::F) where {F, S<:AbstractArray, T<:AbstractArray} at /Users/oxinabox/JuliaEnvs/Flux.jl/src/layers/basic.jl:106


```

test/layers/basic.jl
```diff
@@ -28,6 +28,14 @@ import Flux: activations
...
+      @test_throws MethodError Dense(10, 10.5; σ=tanh)
```
oxinabox, line 36:
```suggestion
      @test_throws MethodError Dense(10, 10.5, tanh)
```

dhairyagandhi96:
I think the intention is to add something along the lines of 

```julia
Dense(in::Number, out::Number) = error("Constructing layers with non integers not supported. Please convert $in and $out to integers, if you intended to use them to construct the layer, or wrap them in arrays if you intended to use them as weights or biases")
```

CarloLucibello:
current error seems quite clear. Merge?

Issue #1224, opened 2020-06-13
Create additional pooling layers


dnabanita7:
This issue is in for better co-ordination of the [FluxModels.jl](https://github.com/darsnack/FluxModels.jl) which can be used as a benchmark for Flux.jl. As explained by @darsnack in [this issue](https://github.com/darsnack/FluxModels.jl/issues/1), I will be creating the following pooling layers.

- [x] AdaptiveMeanPool
- [x] AdaptiveMaxPool

darsnack:
Great! A couple of notes:
- AvgPool is implemented in Flux as MeanPool
- You will want your implementation to call any [pooling primitives from NNlib.jl](https://github.com/FluxML/NNlib.jl/blob/master/src/pooling.jl) when appropriate

dnabanita7:
Oh, I should remove AvgPool then.

darsnack:
And probably want to name the layer AdaptiveMeanPool for consistency.

dnabanita7:
shall i make required changes in the NNlib.jl and create adaptivemaxpool and adaptivemeanpool backends too? I will add two functions there only.

darsnack:
I don’t think that’s necessary. Adaptive pooling does the same operation as regular pooling except that the filter window is adjusted dynamically to account for input image size.

The pooling functions in NNlib.jl accept a `PoolDims` object. Your layers in Flux should create this object based on the desired output size when `layer(x)` is called. Then pass that object to NNlib.

avik-pal:
Fixed by #1239

PR #1233, opened 2020-06-17
add adaptive pooling layers

dnabanita7:
I have added ``AdaptiveMaxPool`` and ``AdaptiveMeanPool`` so that we can do a similar [PyTorch implementation](https://github.com/darsnack/FluxModels.jl/issues/1#issue-635582192). cc @darsnack 

### PR 

- [x] Tests are added
- [ ] Entry in NEWS.md
- [ ] Final review from @MikeInnes or @dhairyagandhi96 (for API changes).

### Flux issue linking
[Flux#1224](https://github.com/FluxML/Flux.jl/issues/1224)
### MLH issue linking
[0.3.x-projects#26](https://github.com/MLH-Fellowship/0.3.x-projects/issues/26)

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -24,7 +24,7 @@ testmode!(m, mode = true) = m
     trainmode!(m, mode = true)
 
 Set a layer of model's train mode (see below).
-Symmetric to [`testmode!`](@ref) (i.e. `trainmode!(m, mode) == testmode!(m, !mode)).
+Symmetric to [`testmode!`](@ref) (i.e. `trainmode!(m, mode) == testmode!(m, !mode)`).
 
 _Note_: if you manually set a model into train mode, you need to manually place
 it into test mode during testing phase.


src/layers/conv.jl
```diff
@@ -480,6 +480,67 @@ end
...
+struct AdaptiveMaxPool{O}
+    out::NTuple{O,Int}
+end
+
+function (a::AdaptiveMaxPool)(x)
+    in_size = size(x)
+    #Output_size
+    out_size = a.out
+    #Stride
+    stride = in_size/out_size
```
darsnack, line 500: Looks like you have test errors with this line. You can look at the Travis build log for more details.

Can you fix this?

src/layers/conv.jl
```diff
@@ -480,6 +480,67 @@ end
...
+    out::NTuple{O,Int}
+end
+
+function (a::AdaptiveMeanPool)(x)
+    #Input_size
+    in_size = size(x)
+    #Output_size
+    out_size = a.out
+    #Stride
+    stride = in_size/out_size
```
darsnack, line 531: Same comment.

src/layers/conv.jl
```diff
@@ -480,6 +480,67 @@ end
...
+Adaptive max pooling layer. `out` is the size of the dimension of the output.
+The output dimension is fixed irrespective of the size of the input. The stride,
+kernel and padding is internally calculated using the input and output dimensions.
+
+"""
+struct AdaptiveMaxPool{O}
+    out::NTuple{O,Int}
+end
+
+function (a::AdaptiveMaxPool)(x)
```
avik-pal, line 495: ND Adaptive Pooling should only operate on ND Arrays. This might not be handled properly.

A simple fix would be to dispatch like
```julia
struct AdaptiveMaxPool{S, T}
     out::NTuple{S, Int}
     AdaptiveMaxPool(out::NTuple{S, Int}) where S = new{S + 2, S}(out)
end

(a::AdaptiveMaxPool{T})(x::AbstractArray{S, T}) where {S, T}
```

This will error if the uses passes an incorrect dimensional array, saying `no method matching (::AdaptiveMaxPool{...})(...)` rather than some failure in the internal of AdaptiveMaxPool

dnabanita7: I didn't get `` where S = new{S + 2, S}(out)``

dhairyagandhi96: That would simply construct a new layer object

darsnack: I am not sure `S` and `T` are being consistently used in this fix. Shouldn't it be:
```julia
struct AdaptiveMaxPool{S, O}
     out::NTuple{O, Int}
     AdaptiveMaxPool(out::NTuple{O, Int}) where O = new{O + 2, O}(out)
end

(a::AdaptiveMaxPool{S})(x::AbstractArray{T, S}) where {S, T}
```
I swapped `T` for `O` in the struct def so it isn't confused with `T` in `AbstractArray` which specifies the type. But I think it should be `out::NTuple{O, Int}` in the struct def to fix.

darsnack:
Can you fix the errors from the lines I commented originally? *(Hint: you might need broadcasting)*

darsnack:
No problem! `insize` and `outsize` are tuples, so you need to use [broadcasting](https://docs.julialang.org/en/v1/manual/arrays/#Broadcasting-1) to divide them element-wise (e.g. `insize ./ outsize`). The other issue is that `insize` will have to extra elements for the channel and batching dimensions. So you want to drop the last two elements: `insize[1:end-2]`.

darsnack:
Similar issues with other arithmetic operations involving `insize` and `outsize`. You need to add broadcasting for all those cases too.

darsnack:
Almost! The broadcasting thing applies to `outsize - 1` as well. Tuples are annoying.

dnabanita7:
> Almost! The broadcasting thing applies to `outsize - 1` as well. Tuples are annoying.

Yeah, I thought to it should be my commit message.

So, is the test ``AdaptiveMaxPool((5, 5))(x) == maxpool(2, PoolDims(2, 2))`` correct for ``x = randn(Float32, 10, 10, 3, 2)``?

darsnack:
Not sure what happened here with the bot. Yes, that test wouldn't be correct since going from (10, 10) -> (4, 2) means the stride is (2.5, 5). So to get it to work, you would need to add padding.

dnabanita7:
yeah, I am switching back to (10,10) -> (5,5) which would clear all the doubts

Issue #1164, opened 2020-05-05
Remove Manifest from master


aminya:
Using Manifest breaks backward compatibility, and it is a bad practice to add it to the master. You should allow Pkg to generate Manifest. By checking Manifest, you will bypass all of Pkg goodness across different users and Julia versions.

You should instead use `compat` entry for limiting the package bounds.
```yaml
PackageX = "0.4, 1"
```

For example by using Manifest, If a dep starts to use a Jil package, all of the Julia versions older than 1.3 will break.
```
[[IntelOpenMP_jll]]
deps = ["Libdl", "Pkg"]
git-tree-sha1 = "fb8e1c7a5594ba56f9011310790e03b5384998d6"
uuid = "1d5cc7b8-4909-519e-a0f8-d0f5ad9712d0"
version = "2018.0.3+0"
```

related https://github.com/FluxML/Zygote.jl/pull/629

aminya:
@MikeInnes @CarloLucibello @dhairyagandhi96  This is not fixed yet. Not sure why it is closed.

PR #1234, opened 2020-06-17
Delete Manifest.toml

aminya:
Using Manifest breaks backward compatibility, and it is a bad practice to add it to the master. You should allow Pkg to generate Manifest. By checking Manifest, you will bypass all of Pkg goodness across different users and Julia versions.

You should instead use `compat` entry for limiting the package bounds.
```yaml
PackageX = "0.4, 1"
```

For example by using Manifest, If a dep starts to use a Jil package, all of the Julia versions older than 1.3 will break.
```
[[Zlib_jll]]
deps = ["Libdl", "Pkg"]
git-tree-sha1 = "2f6c3e15e20e036ee0a0965879b31442b7ec50fa"
uuid = "83775a58-1f1d-513f-b197-d71354ab007a"
version = "1.2.11+9"
```

Closes https://github.com/FluxML/Flux.jl/issues/1164 for real


avik-pal:
To make the change persistent, it is best to add it to .gitignore. Otherwise, it is pretty easy to add it again by mistake.

johnnychen94:
FYI, some discussion of whether we need to check in a `Manifest.toml` can be found in https://github.com/FluxML/Zygote.jl/pull/629#issuecomment-623995315

aminya:
> FYI, some discussion of whether we need to check in a `Manifest.toml` can be found in [FluxML/Zygote.jl#629 (comment)](https://github.com/FluxML/Zygote.jl/pull/629#issuecomment-623995315)

That does not justify using Manifest:
- If there is a dependency, it should be under `[deps]` with its `compat`. If anything more complex is needed a Julia script under `deps/build.jl` should be written.

- If it is test dependency, it should go under `test/Project.toml`, and under `[deps]`, `[compat]`. You can add this simple script to the beginning of the `runtests.jl`, to be compatible with the old versions:
```julia
@static if VERSION < v"1.2.0"
  using Pkg
  Pkg.activate(".")
  Pkg.instantiate()
  Pkg.activate("..")
end
```

PR #1239, opened 2020-06-19
add adaptive pool

dnabanita7:
I have added ``AdaptiveMaxPool`` and ``AdaptiveMeanPool`` so that we can do a similar [PyTorch implementation](https://github.com/darsnack/FluxModels.jl/issues/1#issue-635582192). cc @darsnack 

### PR 

- [x] Tests are added
- [x] Entry in NEWS.md
- [x] Final review from `@MikeInnes` or `@dhairyagandhi96` (for API changes).

### Flux issue linking
[Flux#1224](https://github.com/FluxML/Flux.jl/issues/1224)
### MLH issue linking
[0.3.x-projects#26](https://github.com/MLH-Fellowship/0.3.x-projects/issues/26)

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -494,7 +494,7 @@ function (a::AdaptiveMaxPool{S})(x::AbstractArray{T, S}) where {S, T}
   insize = size(x)[1:end-2]
   outsize = a.out
   stride = insize .÷ outsize
-  k = Int.(insize .- (outsize .- 1) .* stride)
+  k = insize .- (outsize .- 1) .* stride
   pad = 0
   pdims = PoolDims(x, k; padding=pad, stride=stride)
   return maxpool(x, pdims)
@@ -518,7 +518,7 @@ function (a::AdaptiveMeanPool{S})(x::AbstractArray{T, S}) where {S, T}
   insize = size(x)[1:end-2]
   outsize = a.out
   stride = insize .÷ outsize
-  k = Int.(insize .- (outsize .- 1) .* stride)
+  k = insize .- (outsize .- 1) .* stride
   pad = 0
   pdims = PoolDims(x, k; padding=pad, stride=stride)
   return meanpool(x, pdims)
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -517,7 +517,7 @@ end
 function (a::AdaptiveMeanPool{S})(x::AbstractArray{T, S}) where {S, T}
   insize = size(x)[1:end-2]
   outsize = a.out
-  stride = Int.(insize ./ outsize)
+  stride = insize .÷ outsize
   k = Int.(insize .- (outsize .- 1) .* stride)
   pad = 0
   pdims = PoolDims(x, k; padding=pad, stride=stride)
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -493,7 +493,7 @@ end
 function (a::AdaptiveMaxPool{S})(x::AbstractArray{T, S}) where {S, T}
   insize = size(x)[1:end-2]
   outsize = a.out
-  stride = Int.(insize ./ outsize)
+  stride = insize .÷ outsize
   k = Int.(insize .- (outsize .- 1) .* stride)
   pad = 0
   pdims = PoolDims(x, k; padding=pad, stride=stride)
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -481,6 +481,54 @@ outdims(l::CrossCor, isize) =
   output_size(DenseConvDims(_paddims(isize, size(l.weight)), size(l.weight); stride = l.stride, padding = l.pad, dilation = l.dilation))
 
 """
+    AdaptiveMaxPool(out)
+
+Adaptive max pooling layer. `out` is the desired output size (batch and channel dimension excluded).
+"""
+struct AdaptiveMaxPool{S, O}
+  out::NTuple{O, Int}
+  AdaptiveMaxPool(out::NTuple{O, Int}) where O = new{O + 2, O}(out)
+end
+
+function (a::AdaptiveMaxPool{S})(x::AbstractArray{T, S}) where {S, T}
+  insize = size(x)[1:end-2]
+  outsize = a.out
+  stride = Int.(insize ./ outsize)
+  k = Int.(insize .- (outsize .- 1) .* stride)
+  pad = 0
+  pdims = PoolDims(x, k; padding=pad, stride=stride)
+  return maxpool(x, pdims)
+end
+
+function Base.show(io::IO, a::AdaptiveMaxPool)
+  print(io, "AdaptiveMaxPool(", a.out, ")")
+end
+
+"""
+    AdaptiveMeanPool(out)
+
+Adaptive mean pooling layer. `out` is the desired output size (batch and channel dimension excluded).
+"""
+struct AdaptiveMeanPool{S, O}
+  out::NTuple{O, Int}
+  AdaptiveMeanPool(out::NTuple{O, Int}) where O = new{O + 2, O}(out)
+end
+
+function (a::AdaptiveMeanPool{S})(x::AbstractArray{T, S}) where {S, T}
+  insize = size(x)[1:end-2]
+  outsize = a.out
+  stride = Int.(insize ./ outsize)
+  k = Int.(insize .- (outsize .- 1) .* stride)
+  pad = 0
+  pdims = PoolDims(x, k; padding=pad, stride=stride)
+  return meanpool(x, pdims)
+end
+
+function Base.show(io::IO, a::AdaptiveMeanPool)
+  print(io, "AdaptiveMeanPool(", a.out, ")")
+end
+
+"""
     GlobalMaxPool()
 
 Global max pooling layer.
--- a/test/layers/conv.jl
+++ b/test/layers/conv.jl
@@ -4,6 +4,15 @@ using Flux: gradient
 
 @testset "Pooling" begin
   x = randn(Float32, 10, 10, 3, 2)
+  y = randn(Float32, 20, 20, 3, 2)
+  ampx = AdaptiveMaxPool((5,5))
+  @test ampx(x) == maxpool(x, PoolDims(x, 2))
+  ampx = AdaptiveMeanPool((5,5))
+  @test ampx(x) == meanpool(x, PoolDims(x, 2))
+  ampy = AdaptiveMaxPool((10, 5))
+  @test ampy(y) == maxpool(y, PoolDims(y, (2, 4)))
+  ampy = AdaptiveMeanPool((10, 5))
+  @test ampy(y) == meanpool(y, PoolDims(y, (2, 4)))
   gmp = GlobalMaxPool()
   @test size(gmp(x)) == (1, 1, 3, 2)
   gmp = GlobalMeanPool()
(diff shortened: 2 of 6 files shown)


darsnack:
@dhairyagandhi96 do you know why no CI is being performed for this PR?

CarloLucibello:
weirrrd

darsnack:
The PR contains the CI YAML files as far as I can tell...

Could bors be used to trigger?

dnabanita7:
Shall I use GitHub Actions to add the CI workflow? I will create another PR for it.

darsnack:
@dnabanita7 can you try forcing an empty commit with no changes?

dnabanita7:
It works now.

darsnack:
Yeah thanks @CarloLucibello for restarting it!

src/layers/conv.jl
```diff
@@ -480,6 +480,65 @@ end
...
+The output dimension is fixed irrespective of the size of the input. The stride,
+kernel and padding is internally calculated using the input and output dimensions.
+"""
+struct AdaptiveMaxPool{S, O}
+     out::NTuple{O, Int}
+     AdaptiveMaxPool(out::NTuple{O, Int}) where O = new{O + 2, O}(out)
+end
+
+function (a::AdaptiveMaxPool{S})(x::AbstractArray{T, S}) where {S, T}
+    in_size = size(x)
```
darsnack, line 495: Remember you need to drop the last two dimensions. `in_size = size(x)[1:end-2]`.

src/layers/conv.jl
```diff
@@ -480,6 +480,65 @@ end
...
+kernel and padding is internally calculated using the input and output dimensions.
+"""
+struct AdaptiveMeanPool{S, O}
+     out::NTuple{O, Int}
+     AdaptiveMeanPool(out::NTuple{O, Int}) where O = new{O + 2, O}(out)
+end
+
+function (a::AdaptiveMeanPool{S})(x::AbstractArray{T, S}) where {S, T}
+    #Input_size
+    in_size = size(x)
```
darsnack, line 525: Same comment.

src/layers/conv.jl
```diff
@@ -480,6 +480,65 @@ end
 outdims(l::CrossCor, isize) =
   output_size(DenseConvDims(_paddims(isize, size(l.weight)), size(l.weight); stride = l.stride, padding = l.pad, dilation = l.dilation))
 
+"""
+    AdaptiveMaxPool(out)
```
CarloLucibello, line 484: newline here

src/layers/conv.jl
dnabanita7: done

src/layers/conv.jl
dnabanita7: done

src/layers/conv.jl
```diff
@@ -480,6 +480,65 @@ end
...
+Adaptive max pooling layer. `out` is the size of the dimension of the output.
```
CarloLucibello, line 485:
```suggestion
Adaptive max pooling layer. `out` is the desired output size (batch and channel dimension excluded).
```
Maybe it is better to add a usage example to this docstring

src/layers/conv.jl
```diff
@@ -480,6 +480,65 @@ end
...
+kernel and padding is internally calculated using the input and output dimensions.
+"""
+struct AdaptiveMaxPool{S, O}
+     out::NTuple{O, Int}
+     AdaptiveMaxPool(out::NTuple{O, Int}) where O = new{O + 2, O}(out)
+end
+
+function (a::AdaptiveMaxPool{S})(x::AbstractArray{T, S}) where {S, T}
+    in_size = size(x)
+    #Output_size
```
CarloLucibello: you can remove these comments, they don't add much explanation to the code

src/layers/conv.jl
```diff
@@ -480,6 +480,65 @@ end
 outdims(l::CrossCor, isize) =
   output_size(DenseConvDims(_paddims(isize, size(l.weight)), size(l.weight); stride = l.stride, padding = l.pad, dilation = l.dilation))
 
+"""
+    AdaptiveMaxPool(out)
+Adaptive max pooling layer. `out` is the size of the dimension of the output.
+The output dimension is fixed irrespective of the size of the input. The stride,
+kernel and padding is internally calculated using the input and output dimensions.
+"""
+struct AdaptiveMaxPool{S, O}
```
CarloLucibello, line 489: what is `S` fo here? We could use just `O`

test/layers/conv.jl
```diff
@@ -4,6 +4,10 @@ using Flux: gradient
 
 @testset "Pooling" begin
   x = randn(Float32, 10, 10, 3, 2)
+  amp = AdaptiveMaxPool((5,5))
+  @test amp(x) == maxpool(x, PoolDims(x, 2))
+  amp = AdaptiveMeanPool((5,5))
+  @test amp(x) == meanpool(x, PoolDims(x, 2))
```
CarloLucibello, line 10: it would be good to add a few more tests, e.g. with rectangular inputs and/or outputs

src/layers/conv.jl
CarloLucibello: ah now I get it, `S` is to check compatibility with the input array in the forward pass. let's leave it there then

src/layers/conv.jl
dnabanita7: ok

src/layers/conv.jl
dnabanita7: ok

src/layers/conv.jl
```diff
@@ -480,6 +480,60 @@ end
...
+     out::NTuple{O, Int}
+     AdaptiveMaxPool(out::NTuple{O, Int}) where O = new{O + 2, O}(out)
+end
+
+function (a::AdaptiveMaxPool{S})(x::AbstractArray{T, S}) where {S, T}
+    sz = size(x)
+    insize = sz[1:end-2]
+    outsize = a.out
+    stride = insize ./ outsize
+    k = insize .- (outsize .- 1) .* stride
```
darsnack, line 498: I would do the dropping of dimensions earlier when you define `insize` (i.e. `insize = size(x)[1:end-2]`). Otherwise you need to drop the dimensions every place you do math with `insize`.

src/layers/conv.jl
```diff
@@ -480,6 +480,60 @@ end
...
+
+function (a::AdaptiveMeanPool{S})(x::AbstractArray{T, S}) where {S, T}
+    sz = size(x)
+    insize = sz[1:end-2]
+    outsize = a.out
+    stride = insize ./ outsize
+    k = insize .- (outsize .- 1) .* stride
+    pad = 0
+    pdims = PoolDims(x, k; padding=pad, stride=stride)
+    return meanpool(x, pdims)
```
darsnack, line 526: Same comment.

dnabanita7:
`` PoolDims(::Tuple{Vararg{T,M}} where T, !Matched::Union{Int64, Tuple{Vararg{Int64,L}}}; stride, padding, dilation) where {M, L}`` shall I change the dispatch in struct according to this?

dnabanita7:
> Since you are removing the comments, I would get rid of the blank lines in between. Personally I would only have a blank line before the return statement.

The spaces between functions too? Well other functions don't have that in that file.

CarloLucibello:
Flux uses 2 whitespaces indent (unfortunately in my opinion), so you should correct all indents, except for docstrings (where you should leave 4)

darsnack:
> The spaces between functions too? Well other functions don't have that in that file.

No that comment is out of date. Please ignore.

darsnack:
> ``PoolDims(::Tuple{Vararg{T,M}} where T,`` or `` PoolDims(::Union{Int64, Tuple{Vararg{Int64,L}}}; stride, padding, dilation) where {M, L}`` shall I change the dispatch in struct according to this?

What exactly do you mean?

dnabanita7:
The error shown is method mismatch. Shall I change the constructor in struct to the above mentioned?

src/layers/conv.jl
```diff
@@ -480,6 +480,54 @@ end
...
+struct AdaptiveMaxPool{S, O}
+     out::NTuple{O, Int}
+     AdaptiveMaxPool(out::NTuple{O, Int}) where O = new{O + 2, O}(out)
+end
+
+function (a::AdaptiveMaxPool{S})(x::AbstractArray{T, S}) where {S, T}
+    insize = size(x)[1:end-2]
+    outsize = a.out
+    stride = insize ./ outsize
+    k = insize .- (outsize .- 1) .* stride
```
darsnack, line 497: The error is cause `k` is a Float64. Try `k = Int.(insize .- (outsize .- 1) .* stride)`.

test/layers/conv.jl
darsnack: @dnabanita7 Can you address this comment? Perhaps (20, 20) —> (10, 5) etc.?

dnabanita7:
I guess its good to get a final review from @dhairyagandhi96 or @MikeInnes

avik-pal:
@dnabanita7 An entry in the NEWS.md would be needed for this.

dnabanita7:
cc @DhairyaLGandhi

DhairyaLGandhi:
Can we also add the tests to the GPU tests. I know `maxpool` is already in the test suite, but good for making sure we are doing layer regression tests consistently.

src/layers/conv.jl
```diff
@@ -480,6 +480,54 @@ end
...
+"""
+struct AdaptiveMeanPool{S, O}
+  out::NTuple{O, Int}
+  AdaptiveMeanPool(out::NTuple{O, Int}) where O = new{O + 2, O}(out)
+end
+
+function (a::AdaptiveMeanPool{S})(x::AbstractArray{T, S}) where {S, T}
+  insize = size(x)[1:end-2]
+  outsize = a.out
+  stride = Int.(insize ./ outsize)
```
DhairyaLGandhi, line 520: You might get issues here if insize and outsize don't divide properly. Any reason to use `/` instead of `÷`?

DhairyaLGandhi: Might be good to add test cases to check against the expected output in that case

dnabanita7: > You might get issues here if insize and outsize don't divide properly. Any reason to use `/` instead of `÷`?

No, there is particularly no reason. We can use ``\div[TAB]``. The matrix values will still be of the float type and have to change to ``Int``. 

Also, I am getting ``Type error`` if I input ``Float32`` values of x and use ``Int`` instead of ``Int32`` on REPL.

dnabanita7: By using ``Int`` I mean as ``Int.(insize .÷ outsize)`` or ``Int.(insize ./ outsize)``. But the tests I have written are not showing any type error which is confusing. I have used ``Float32`` values in my tests too.

avik-pal: > The matrix values will still be of the float type and have to change to Int.

For `÷` the return type will be Int as long as the eltypes of inputs `insize` and `outsize` are integers. So you **don't** need to do the type casting by `Int.(insize .÷ outsize)`

dnabanita7: Yes, I realized that inputs I took were Floats instead of Integers as used here.

DhairyaLGandhi: insize and outsize will always be integer tuples and we always want integer tuples out, right? The input and output matrices should retain their original eltypes anyway. Why should the types of the size tuples influence that?

DhairyaLGandhi: There's a few places where it occurs, could you make sure it's changed consistently please?

DhairyaLGandhi:
There's a few other places that would need the `\div` treatment. Thanks for looking into it!

CarloLucibello:
thanks for this contribution 

bors r+

PR #1240, opened 2020-06-19
Documentation: Move logging example outside gradient block

contradict:
Placing the logging callback inside the gradient block means Zygote will try to differentiate it. That gradient information is unlikely to be useful and likely to cause errors (especially about mutating arrays).

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/training/training.md
+++ b/docs/src/training/training.md
@@ -138,7 +138,7 @@ E.g. in the places marked with comments.
 
 ```julia
 function my_custom_train!(loss, ps, data, opt)
-  # training_loss is declared local so it will be available for logging outside the gradient calculation
+  # training_loss is declared local so it will be available for logging outside the gradient calculation.
   local training_loss
   ps = Params(ps)
   for d in data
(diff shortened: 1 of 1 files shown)


docs/src/training/training.md
```diff
@@ -138,13 +138,17 @@ E.g. in the places marked with comments.
...
+  # training_loss is declared local so it will be available for logging outside the gradient calculation
+  local training_loss
   ps = Params(ps)
   for d in data
     gs = gradient(ps) do
       training_loss = loss(d...)
-      # Insert whatever code you want here that needs Training loss, e.g. logging
-      return training_loss
+      # Code inserted here will be differentiated, unless you need that gradient information
+      # it is better to do the work outside this block.
```
CarloLucibello, line 148: can we add `return training_loss` here ?

CarloLucibello:
We could also add custom loop using `pullback` in this section. Something like
```
function my_custom_train!(loss, ps, data, opt)
  ps = Params(ps)
  for d in data
    train_loss, back = pullback(() -> loss(d...), ps)
    gs = back(1f0)
    update!(opt, ps, gs)
  end
end
```

docs/src/training/training.md
contradict: Yes, it could be added, but it does not change the result and I though it was somewhat confusing since the result of this code is the gradient of loss not the loss. Is there some advantage to having the explicit return?

contradict:
This looks like an all-around nicer syntax, although it would need some more explanation (pullback is not mentioned anywhere else in the Flux documentation). If you want to switch to this syntax, I think I should close this PR and try again with this technique (after I read about it some myself).

docs/src/training/training.md
CarloLucibello: the problem is that we don't even have an implicit return, we just have a comment "insert some code here". So it would be better to make it clear what the return value is

CarloLucibello:
> This looks like an all-around nicer syntax, although it would need some more explanation (pullback is not mentioned anywhere else in the Flux documentation). If you want to switch to this syntax, I think I should close this PR and try again with this technique (after I read about it some myself).

`gradient` is implemented in terms of `pullback`, see https://github.com/FluxML/Zygote.jl/blob/257dfb139d0cc3c88a87654c4412fa5b7947bddd/src/compiler/interface.jl#L52. `pullback` just returns the output of the function and a closure that performs backpropagation.

Instead of switching syntax, we can just add another example using `pullback`

contradict:
OK, another try! How does this look?

CarloLucibello:
Looks good! I think we can add a pointer to Zygote's pullback documentation https://fluxml.ai/Zygote.jl/dev/adjoints/#Pullbacks-1 and we are done

contradict:
The link is in there, at the first mention of pullback on line 162. Thanks for looking this over!

CarloLucibello:
> The link is in there, at the first mention of pullback on line 162. Thanks for looking this over!

right. Thanks for this PR!

bors r+

PR #1242, opened 2020-06-20
Fixed Dimension Mismatch - AbtractMatrix and OneHotVector

maerory:
Fixed the multiplication function so that it only works when the column of matrix matches with the size of onehotvector.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -11,7 +11,12 @@ Base.getindex(xs::OneHotVector, i::Integer) = i == xs.ix
 
 Base.getindex(xs::OneHotVector, ::Colon) = OneHotVector(xs.ix, xs.of)
 
-A::AbstractMatrix * b::OneHotVector = A[:, b.ix]
+function Base.:*(A::AbstractMatrix, b::OneHotVector)
+  if size(A)[2] != b.of
+    throw(DimensionMismatch("Matrix column must correspond with OneHotVector size"))
+  end
+  return A[:, b.ix]
+end
 
 struct OneHotMatrix{A<:AbstractVector{OneHotVector}} <: AbstractMatrix{Bool}
   height::Int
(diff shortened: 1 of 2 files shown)


src/onehot.jl
```diff
@@ -11,7 +11,12 @@ Base.getindex(xs::OneHotVector, i::Integer) = i == xs.ix
 
 Base.getindex(xs::OneHotVector, ::Colon) = OneHotVector(xs.ix, xs.of)
 
-A::AbstractMatrix * b::OneHotVector = A[:, b.ix]
+function Base.:*(A::AbstractMatrix, b::OneHotVector)
+  if size(A)[2] != b.of
```
CarloLucibello, line 15:
```suggestion
  if size(A, 2) != b.of
```

CarloLucibello:
Can we add a `@test_throws DimensionMismatch ...`    test here?

DhairyaLGandhi:
Generally sgtm

maerory:
> Can we add a `@test_throws DimensionMismatch ...` test here?

Do you mean unit tests?

CarloLucibello:
yes, you can add a test here https://github.com/FluxML/Flux.jl/blob/master/test/onehot.jl

CarloLucibello:
@maerory  bump

maerory:
Sorry about the delay,

I have added a new tester, please check whether it is enough or I should add more.
In addition, I have discovered another error regarding OneHotVector creation.

When OneHotVector(i,  n) is called where i > n, Flux is currently returning zero vector of size n.
Should this also throw an error?
I will open another issue and fix this if that is the case.

maerory:
@CarloLucibello bump+

CarloLucibello:
Thanks!

bors r+

PR #1243, opened 2020-06-20
add kaiming initialization and relevant docstrings

### LINKED ISSUE #424: Support for Kaiming Initialization

johnnychen94:
This is an updated version of #425, `Distributions` is not used because it always generates `Array{Float64, N}` instead of `Array{Float32, N}`.

* also updates the docstring of glorot initialization
* add a method for `nfan(::Tuple)` for robustness consideration, otherwise `nfan((100, 400))` would return `(1, (100, 400))`, which isn't correct.

These methods are not exported because `glorot_*` aren't, either.

If this get merged, we could switch from `glorot_uniform` to `kaiming_uniform` for `Conv` since people nowadays use `relu` mostly, but that belongs to another PR.

closes #425 closes #424 

Co-authored-by: Aniket Das <aniketd@iitk.ac.in>

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md
- [x] Final review from `@MikeInnes` or `@dhairyagandhi96` (for API changes).

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -2,13 +2,16 @@
 nfan() = 1, 1 # fan_in, fan_out
 nfan(n) = 1, n # A vector is treated as a n×1 matrix
 nfan(n_out, n_in) = n_in, n_out # In case of Dense kernels: arranged as matrices
+nfan(dims::Tuple) = nfan(dims...)
 nfan(dims...) = prod(dims[1:end-2]) .* (dims[end-1], dims[end]) # In case of convolution kernels
 
 """
     glorot_uniform(dims...)
 
 Return an `Array` of size `dims` containing random variables taken from a uniform
-distribution in the interval ``[-x, x]``, where `x = sqrt(24 / sum(dims)) / 2`.
+distribution in the interval ``[-x, x]``, where `x = sqrt(6 / (fan_in + fan_out))`.
+
+This method is described in [1] and also known as Xavier initialization.
 
 # Examples
 ```jldoctest; setup = :(using Random; Random.seed!(0))
@@ -17,6 +20,13 @@ julia> Flux.glorot_uniform(2, 3)
  0.601094  -0.57414   -0.814925
  0.900868   0.805994   0.057514
 ```
+
+See also: [`glorot_normal`](@ref Flux.glorot_normal), [`kaiming_normal`](@ref Flux.kaiming_normal),
+[`kaiming_uniform`](@ref Flux.kaiming_uniform)
+
+# References
+
+[1] Glorot, Xavier, and Yoshua Bengio. "Understanding the difficulty of training deep feedforward neural networks." _Proceedings of the thirteenth international conference on artificial intelligence and statistics_. 2010.
 """
 glorot_uniform(dims...) = (rand(Float32, dims...) .- 0.5f0) .* sqrt(24.0f0 / sum(nfan(dims...)))
 
@@ -24,7 +34,9 @@ glorot_uniform(dims...) = (rand(Float32, dims...) .- 0.5f0) .* sqrt(24.0f0 / sum
     glorot_normal(dims...)
 
 Return an `Array` of size `dims` containing random variables taken from a normal
-distribution with mean 0 and standard deviation `sqrt(2 / sum(dims))`.
+distribution with mean 0 and standard deviation `sqrt(2 / (fan_in + fan_out))`.
+
+This method is described in [1] and also known as Xavier initialization.
 
 # Examples
 ```jldoctest; setup = :(using Random; Random.seed!(0))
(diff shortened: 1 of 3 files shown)


src/utils.jl
```diff
@@ -34,9 +46,75 @@ julia> Flux.glorot_normal(3, 2)
...
+See also: [`kaiming_normal`](@ref Flux.kaiming_normal), [`glorot_normal`](@ref Flux.glorot_normal),
+[`glorot_uniform`](@ref Flux.glorot_uniform)
+
+# References
+
+[1] He, Kaiming, et al. "Delving deep into rectifiers: Surpassing human-level performance on imagenet classification." _Proceedings of the IEEE international conference on computer vision_. 2015.
+"""
+function kaiming_uniform(dims...; gain = √2)
+  bound = Float32(√3 * gain / sqrt(first(nfan(dims...)))) # fan_in
+  return (rand(Float32, dims...) .- 0.5f0) .* 2.0f0bound
```
CarloLucibello, line 86:
```suggestion
  return (rand(Float32, dims...) .- 0.5f0) .* 2bound
```

src/utils.jl
```diff
@@ -2,13 +2,16 @@
 nfan() = 1, 1 # fan_in, fan_out
 nfan(n) = 1, n # A vector is treated as a n×1 matrix
 nfan(n_out, n_in) = n_in, n_out # In case of Dense kernels: arranged as matrices
+nfan(dims::Tuple) = nfan(dims...)
```
CarloLucibello, line 5: not needed?

src/utils.jl
```diff
@@ -2,13 +2,16 @@
...
 nfan(dims...) = prod(dims[1:end-2]) .* (dims[end-1], dims[end]) # In case of convolution kernels
 
 """
     glorot_uniform(dims...)
 
 Return an `Array` of size `dims` containing random variables taken from a uniform
-distribution in the interval ``[-x, x]``, where `x = sqrt(24 / sum(dims)) / 2`.
+distribution in the interval ``[-x, x]``, where `x = sqrt(6 / (fan_in + fan_out))`.
```
CarloLucibello, line 12: we should explain what fan_in and fan_out are

src/utils.jl
johnnychen94: fixed in a219fecd564df0b70ba63e4cbd62c815932350a0, also ensure the eltype is `Float32` in tests.

src/utils.jl
johnnychen94: Actually, pytorch doesn't explain the meaning of them, either. https://pytorch.org/docs/stable/nn.init.html#torch.nn.init.kaiming_uniform_

In c78fedf83390fb74de98ac4b4caf06f7cf843ca9, I've added a not-so-illustrative docstring to `nfan`, and give a see-also reference to it.

src/utils.jl
```diff
@@ -1,14 +1,37 @@
...
+
+Calculate the maximum numbers of input and output data each neuron weight in the layer connects to.
+
+In terms of digital electronics, these are also known as [fan-in](https://en.wikipedia.org/wiki/Fan-in)
+and [fan-out](https://en.wikipedia.org/wiki/Fan-out) of the layer.
+
+# Examples
+
+```jldoctest
+julia> Flux.nfan(100, 200) # dense kernel
```
CarloLucibello, line 15: maybe you could create a Dense and a Conv layer, and pass `size(layer.W)` to `nfan`, could add some clarity

src/utils.jl
```diff
@@ -1,14 +1,45 @@
 # Arrays
+"""
+    nfan(n_out, n_in=1) -> Tuple
+    nfan(dims...)
+    nfan(dims::Tuple)
+
+Calculate the maximum numbers of input and output data each neuron weight in the layer connects to.
```
CarloLucibello, line 7: This settence is not utterly clear, maybe something like this
```suggestion
For a layer characterized by dimensions  `dims`,  return a tuple `(fan_in, fan_out)`, where `fan_in`
is the number of input neurons connect to an output one, and `fan_out` is the number of output neurons connected to an input one.
```
?

CarloLucibello:
For future reference:

- Keras' default is  glorot uniform in [-limit, limit], where `limit = sqrt(6 / (fan_in + fan_out)` https://keras.io/api/layers/initializers/ . This implies std = sqrt(2 / (fan_in + fan_out), which should be the correct value for ReLU activations (by a variance propagation argument). 
- Pytorch's default instead is uniform in [-limit, limit], where `limit = sqrt(1 / fan_in)`
https://github.com/pytorch/pytorch/pull/19526#issuecomment-485590295.
I don't have a mathematical explanation for Pytorch's default,  I think is there for backward compatibility, maybe an initialization suggested by LeCunn paper in the '90s, I don't remember which one.

src/utils.jl
johnnychen94: Much clearer than my words, thanks

johnnychen94:
bump

DhairyaLGandhi:
Lgtm, although the references to the other functions are a bit verbose. We should also add a section for initialisation in the docs (in a future PR)

Thanks @johnnychen94

DhairyaLGandhi:
We should add a bit in the docs about how to use the initializations in regular layers

DhairyaLGandhi:
Could we consider taking `T` as user input, since although we default to f32, there are times when higher (or lower) precision may be preferred.

johnnychen94:
Rebased the commits with no content changes.

johnnychen94:
> Pytorch's default instead is uniform in [-limit, limit], where limit = sqrt(1 / fan_in)
pytorch/pytorch#19526 (comment).

https://github.com/pytorch/pytorch/blob/4f723825b48e555512813fcffd56e89e5b16eeaf/torch/nn/modules/conv.py#L85-L90

Pytorch's default is `kaiming_uniform` with `gain=sqrt(5)` for weights and `uniform` for bias.

PR #1204, opened 2020-06-01
CuArrays -> CUDA

DhairyaLGandhi:
This updates the compatibility bounds on CuArrays and moves it to CUDA.jl. CUDA.jl seems to lower bound on Julia 1.4 (@maleadt any chance we could run it on 1.3 somehow?).

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md
- [ ] Documentation, if applicable
- [ ] Final review from `@MikeInnes` or `@dhairyagandhi96` (for API changes).

PR #1248, opened 2020-06-26
Zygote 0.5 compat

MikeInnes:
Trivial, we just need to make sure bors passes with 0.5.


willtebbutt:
Scratch that, looks like some other dep is preventing 0.5 from being tests. 0.4.20 is still be installed on Julia 1.3

CarloLucibello:
i think we should bump NNlib and Zygote in #1204

Issue #33, opened 2017-05-18
tensorflow error: you must feed a value for placeholder tensor 'placeholder_8' with dtype float


ExpandingMan:
I'm probably getting a bit ahead of things here since I'm using master of `TensorFlow.jl` as I'm on 0.6 and the latest release doesn't work.  TensorFlow itself seems to be working just fine on master.

Anyway, I get the following error when attempting to train the MNIST example:
```
ERROR: LoadError: Tensorflow error: Status: You must feed a value for placeholder tensor 'placeholder_8' with dtype float
         [[Node: placeholder_8 = Placeholder[_class=[], dtype=DT_FLOAT, shape=[], _device="/job:localhost/replica:0/task:0/gpu:0"]()]]
         [[Node: placeholder_8/_17 = _Recv[client_terminated=false, recv_device="/job:localhost/replica:0/task:0/cpu:0", send_device="/job:localhost/replica:0/task:0/gpu:0", send_device_incarnation=1, tensor_name="edge_3_placeholder_8", tensor_type=DT_FLOAT, _device="/job:localhost/replica:0/task:0/cpu:0"]()]]

Stacktrace:
 [1] check_status at /home/expandingman/.julia/v0.6/TensorFlow/src/core.jl:402 [inlined]
 [2] run(::TensorFlow.Session, ::Array{TensorFlow.Port,1}, ::Array{Any,1}, ::Array{TensorFlow.Port,1}, ::Array{Ptr{Void},1}) at /home/expandingman/.julia/v0.6/TensorFlow/src/run.jl:100
 [3] run(::TensorFlow.Session, ::Array{TensorFlow.Tensor{Float32},1}, ::Dict{Any,Any}) at /home/expandingman/.julia/v0.6/TensorFlow/src/run.jl:169
 [4] run(::TensorFlow.Session, ::TensorFlow.Tensor{Float32}, ::Dict{Any,Any}) at /home/expandingman/.julia/v0.6/TensorFlow/src/run.jl:187
 [5] back!(::Flux.TF.Exec, ::Array{Float32,2}, ::Array{Float64,2}) at /home/expandingman/.julia/v0.6/Flux/src/backend/tensorflow/model.jl:44
 [6] macro expansion at /home/expandingman/.julia/v0.6/Flux/src/training.jl:36 [inlined]
 [7] macro expansion at /home/expandingman/.julia/v0.6/Juno/src/progress.jl:128 [inlined]
[...]

PR #1256, opened 2020-07-01
Updated onehot.jl

Dsantra92:
@janEbert @CarloLucibello 
Updated doctstrings and doctests  for onehot.jl and onehot.md

### DIFF
--- a/docs/src/data/onehot.md
+++ b/docs/src/data/onehot.md
@@ -2,7 +2,7 @@
 
 It's common to encode categorical variables (like `true`, `false` or `cat`, `dog`) in "one-of-k" or ["one-hot"](https://en.wikipedia.org/wiki/One-hot) form. Flux provides the `onehot` function to make this easy.
 
-```
+```jldoctest onehot
 julia> using Flux: onehot, onecold
 
 julia> onehot(:b, [:a, :b, :c])
@@ -20,7 +20,7 @@ julia> onehot(:c, [:a, :b, :c])
 
 The inverse is `onecold` (which can take a general probability distribution, as well as just booleans).
 
-```julia
+```jldoctest onehot
 julia> onecold(ans, [:a, :b, :c])
 :c
 
@@ -40,20 +40,15 @@ Flux.onecold
 
 `onehotbatch` creates a batch (matrix) of one-hot vectors, and `onecold` treats matrices as batches.
 
-```julia
+```jldoctest onehot
 julia> using Flux: onehotbatch
 
 julia> onehotbatch([:b, :a, :b], [:a, :b, :c])
-3×3 Flux.OneHotMatrix:
- false   true  false
-  true  false   true
- false  false  false
+3×3 Flux.OneHotMatrix{Array{Flux.OneHotVector,1}}:
+ 0  1  0
+ 1  0  1
+ 0  0  0
 
-julia> onecold(ans, [:a, :b, :c])
-3-element Array{Symbol,1}:
-  :b
-  :a
-  :b
 ```
 
 Note that these operations returned `OneHotVector` and `OneHotMatrix` rather than `Array`s. `OneHotVector`s behave like normal vectors but avoid any unnecessary cost compared to using an integer index directly. For example, multiplying a matrix with a one-hot vector simply slices out the relevant row of the matrix under the hood.
--- a/docs/src/data/onehot.md
+++ b/docs/src/data/onehot.md
@@ -49,6 +49,11 @@ julia> onehotbatch([:b, :a, :b], [:a, :b, :c])
  1  0  1
  0  0  0
 
+julia> onecold(ans, [:a, :b, :c])	
+3-element Array{Symbol,1}:	
+ :b	
+ :a	
+ :b   
 ```
 
 Note that these operations returned `OneHotVector` and `OneHotMatrix` rather than `Array`s. `OneHotVector`s behave like normal vectors but avoid any unnecessary cost compared to using an integer index directly. For example, multiplying a matrix with a one-hot vector simply slices out the relevant row of the matrix under the hood.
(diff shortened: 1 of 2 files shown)


src/onehot.jl
```diff
@@ -45,11 +45,16 @@ cudaconvert(x::OneHotMatrix{<:CuArray}) = OneHotMatrix(x.height, cudaconvert(x.d
...
+`onehot(unk, labels)`.
+
+Otherwise the function returns an error message.
 
-Create a `OneHotVector` with its `l`-th element `true` based on the
-possible set of `labels`.
-If `unk` is given, return `onehot(unk, labels)` if the input label `l` is not found
-in `labels`; otherwise, it will raise an error.
 
 # Examples
```
janEbert, line 59: I suggest using the imperative style below the first paragraph as well ("the function returns" -> "return").
Also, I think the way it's currently written suggests that only _one_ `k`-th index is set to 1. Though maybe that's just me.
Finally, there are some typos.

Here is a suggestion (sorry I changed so much):
```
"""
    onehot(l, labels[, unk])

Return a `OneHotVector` where each `l` in `labels` is `1`; all other elements are `0`.

If `l` is not found in `labels` and `unk` is given, return `onehot(unk, labels)`.
Otherwise, raise an error.

# Examples
[...]
"""
```

src/onehot.jl
```diff
@@ -83,6 +88,9 @@ end
 
 Create a `OneHotMatrix` with a batch of labels based on the
 possible set of `labels`.
+If every element of `ls` is present in labels it return a matrix with columns as
```
janEbert, line 91: This was accidentally duplicated and should be removed.

src/onehot.jl
```diff
@@ -83,6 +88,9 @@ end
...
+If every element of `ls` is present in labels, it return a matrix with one-hot vectors as columns. The `k`th column of the matrix has value of `onehot(ls[k], labels)`.
```
janEbert, line 92: I don't think it's necessary to explicitly state here that every element needs to be present; the next paragraph mentioning `unk` already explains this. Also, you can probably merge the sentence above your explanation ("Create a `OneHotMatrix` [...].") with yours.
Finally, the line is too long; it's suggested keeping the length limited to 92 symbols.

Here are some typo corrections and other suggestions in case you want to keep your first sentence as well:
```
If every element of `ls` is present in `labels`, return a matrix with `OneHotVector`s
as columns. The `k`-th column of the matrix is `onehot(ls[k], labels)`.
```

Taking everything into account, the final product of the first paragraph could be something like this (maybe too concise):
```
Return a `OneHotMatrix` where the `k`-th column is [`onehot(ls[k], labels)`](@ref).
```

docs/src/data/onehot.md
```diff
@@ -40,20 +40,15 @@ Flux.onecold
...
+3×3 Flux.OneHotMatrix{Array{Flux.OneHotVector,1}}:
+ 0  1  0
+ 1  0  1
+ 0  0  0
 
-julia> onecold(ans, [:a, :b, :c])
-3-element Array{Symbol,1}:
-  :b
-  :a
-  :b
```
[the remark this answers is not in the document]
Dsantra92: No the doctests were failing when I tested it on my machine.

This is the particular error I am facing.

ERROR: MethodError: no method matching |(::Int64, ::Symbol)
Closest candidates are:
  |(::Any, ::Any, ::Any, ::Any...) at operators.jl:529
  |(::T, ::T) where T<:Union{Int128, Int16, Int32, Int64, Int8, UInt128, UInt16, UInt32, UInt64, UInt8} at int.jl:300
  |(::Integer) at operators.jl:507
  ...
Stacktrace:
 [1] macro expansion at ./reducedim.jl:267 [inlined]
 [2] macro expansion at ./simdloop.jl:77 [inlined]
 [3] _mapreducedim!(::Flux.var"#33#34"{Tuple{Array{Symbol,1}}}, ::typeof(|), ::Array{Int64,1}, ::Array{Flux.OneHotVector,1}) at ./reducedim.jl:266
 [4] mapreducedim! at ./reducedim.jl:274 [inlined]
 [5] _mapreduce_dim at ./reducedim.jl:314 [inlined]
 [6] #mapreduce#580 at ./reducedim.jl:307 [inlined]
 [7] onecold(::Flux.OneHotMatrix{Array{Flux.OneHotVector,1}}, ::Array{Symbol,1}) at /home/deebyo/.julia/packages/Flux/Fj3bt/src/onehot.jl:121
 [8] top-level scope at REPL[3]:1

src/onehot.jl
Dsantra92: The `OneHot` function will return 1 for the first index where l is present and not for all index of l.
This is the function definition:
```
function onehot(l, labels)
  i = something(findfirst(isequal(l), labels), 0)
  i > 0 || error("Value $l is not in labels")
  OneHotVector(i, length(labels))
end
```
This is a sample output :

julia> onehot(3,[2,4,3,3,3])
5-element Flux.OneHotVector:
 0
 0
 1
 0
 0

Do we need to rewrite the function and write the docstring as the function is intended to work?? or do we go with the current version of `OneHot` and write docstring based on it??

src/onehot.jl
Dsantra92: Sorry I should have revised before the commit, my bad.

docs/src/data/onehot.md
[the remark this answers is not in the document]
Dsantra92: @janEbert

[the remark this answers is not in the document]
CarloLucibello: on my machine `onecold` works perfectly fine
```
julia> x = onehotbatch([:b, :a, :b], [:a, :b, :c])
3×3 Flux.OneHotMatrix{Array{Flux.OneHotVector,1}}:
 0  1  0
 1  0  1
 0  0  0

julia> onecold(x, [:a, :b, :c])
3-element Array{Symbol,1}:
 :b
 :a
 :b
```
Are you on latest Flux version?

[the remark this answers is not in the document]
Dsantra92: Got some errors with my local machine.I have re-added the test.

CarloLucibello:
thanks!

bors r+

PR #1258, opened 2020-07-02
move to CUDA.jl

CarloLucibello:
continuing #1204


DhairyaLGandhi:
I would recommend pushing to that branch directly rather than have multiple prs do the same thing with the same commits. I can give you commit bit on that

CarloLucibello:
> I would recommend pushing to that branch directly rather than have multiple prs do the same thing. I can give you commit bit on that

yeah, I wasn't sure how to do that since that branch is on your personal fork. This is on the original repo, so you should already be able to push to it

CarloLucibello:
@maleadt I observe 
```julia
julia> using Flux
┌ Warning: `haskey(::TargetIterator, name::String)` is deprecated, use `Target(; name = name) !== nothing` instead.
│   caller = llvm_compat(::VersionNumber) at compatibility.jl:181
└ @ CUDA ~/.julia/packages/CUDA/42B9G/deps/compatibility.jl:181
```
here. If `using CUDA` instead, no warning appears. Any clue?

DhairyaLGandhi:
Why remove the doctest?

PR #1261, opened 2020-07-02
reinstate curnn test

CarloLucibello:
this is passing locally on my GPU, but in #1258 bors didn't seem to like it. Let's try again 

EDIT: actually this is not passing test locally, but let's see what BORS says in any case
cfr #1262


DhairyaLGandhi:
Seems to happen trying to convert a CuArray to an Array which is odd.

PR #1263, opened 2020-07-02
dropout function always active

### LINKED ISSUE #1084: dropout function is implemented as just an identity
The forward pass of the `Dropout` layer, which gets called when you do something like`d = Dropout(p); d(x)` is defined as just an identity function [here](https://github.com/FluxML/Flux.jl/blob/df3f904f7c34f095562693b6a9ca67047319dea0/src/layers/normalise.jl#L22).

Is that a bug? Or am I going crazy?

CarloLucibello:
Fix #1084 according to the suggestions of @AzamatB and @jondeuce

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,7 +17,7 @@
 * Testing suite improvements now test for gradients of all layers along with GPU support.
 * Functors have now moved to [Functors.jl](https://github.com/FluxML/Flux.jl/pull/1174) to allow for their use outside of Flux.
 * Added [helper functions](https://github.com/FluxML/Flux.jl/pull/873) `Flux.convfilter` and `Flux.depthwiseconvfilter` to construct weight arrays for convolutions outside of layer constructors so as to not have to depend on the default layers for custom implementations.
-* `dropout` function is now [always acivte](https://github.com/FluxML/Flux.jl/pull/1263)
+* `dropout` function is now [always active](https://github.com/FluxML/Flux.jl/pull/1263)
 * and many more fixes and additions...
 
 # v0.10.1 - v0.10.4
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,7 +18,7 @@
 * Functors have now moved to [Functors.jl](https://github.com/FluxML/Flux.jl/pull/1174) to allow for their use outside of Flux.
 * Added [helper functions](https://github.com/FluxML/Flux.jl/pull/873) `Flux.convfilter` and `Flux.depthwiseconvfilter` to construct weight arrays for convolutions outside of layer constructors so as to not have to depend on the default layers for custom implementations.
 * `dropout` function now has a mandatory [active](https://github.com/FluxML/Flux.jl/pull/1263)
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi:
I don't think we should do it this way, dropout is usually omitted at test time, this way it will make it difficult to turn it off when most cases all for it to be off

CarloLucibello:
> I don't think we should do it this way, dropout is usually omitted at test time, this way it will make it difficult to turn it off when most cases all for it to be off

this is what the `Dropout` type is for. In fact, in 99% cases you should always use the `Dropout` layer and never the `dropout` function.  With this PR the implementations are now consistent with the docstrings (they weren't before). I can add some additional warning in the docstring

DhairyaLGandhi:
In that case better to just remove the starting `_` in `dropout_kernel` and document using that if anyone wants to dropout all the time. Fixing a docstring is a good idea though. Thanks!

CarloLucibello:
to avoid silent change of behavior, I introduced the mandatory keyword argument `active`, that we can possibly deprecate in the future

NEWS.md
```diff
@@ -17,6 +17,7 @@
 * Testing suite improvements now test for gradients of all layers along with GPU support.
 * Functors have now moved to [Functors.jl](https://github.com/FluxML/Flux.jl/pull/1174) to allow for their use outside of Flux.
 * Added [helper functions](https://github.com/FluxML/Flux.jl/pull/873) `Flux.convfilter` and `Flux.depthwiseconvfilter` to construct weight arrays for convolutions outside of layer constructors so as to not have to depend on the default layers for custom implementations.
+* `dropout` function is now [always active](https://github.com/FluxML/Flux.jl/pull/1263)
```
oxinabox, line 20: mention the `Dropout` layer here ?

CarloLucibello:
@DhairyaLGandhi merge?

NEWS.md
```diff
@@ -17,6 +17,8 @@
 * Testing suite improvements now test for gradients of all layers along with GPU support.
 * Functors have now moved to [Functors.jl](https://github.com/FluxML/Flux.jl/pull/1174) to allow for their use outside of Flux.
 * Added [helper functions](https://github.com/FluxML/Flux.jl/pull/873) `Flux.convfilter` and `Flux.depthwiseconvfilter` to construct weight arrays for convolutions outside of layer constructors so as to not have to depend on the default layers for custom implementations.
+* `dropout` function now has a mandatory [active](https://github.com/FluxML/Flux.jl/pull/1263)
+keyword argument. The `Dropout` struct is the recommended choice for common usage.
```
DhairyaLGandhi, line 21: Maybe say the behaviour of the layer has not changed.

CarloLucibello:
@DhairyaLGandhi done

DhairyaLGandhi:
bors r+

In general, we should try to remove the dependency on the `active` field in other structs as well, and use functional approaches instead (like we used to do earlier) and remove `trainmode!` but that's breaking and the matter of a separate discussion.

Issue #1255, opened 2020-07-01
Performance issue when calculating loss


findmyway:
I notice that many loss functions in this package are written like this:

https://github.com/FluxML/Flux.jl/blob/942d5f63eab0d3de69c54eb47be06cea8a3c2de9/src/layers/stateless.jl#L8

https://github.com/FluxML/Flux.jl/blob/942d5f63eab0d3de69c54eb47be06cea8a3c2de9/src/layers/stateless.jl#L23

https://github.com/FluxML/Flux.jl/blob/942d5f63eab0d3de69c54eb47be06cea8a3c2de9/src/layers/stateless.jl#L35

They all have a `* 1 //  ` .  Originally I thought it was redundant. Only recently I got a performance issue and found that doing so will avoid a performance issue. Can anyone explain why we need this?

CarloLucibello:
Can you post some benchmarks?

DhairyaLGandhi:
That was done to ensure type stability and prevent unnecessary type promotions that can occur by mixing `Float64` and `Float32`  for example

findmyway:
> That was done to ensure type stability and prevent unnecessary type promotions that can occur by mixing `Float64` and `Float32` for example


I'm afraid it's not related to `Float64` and `Float32` here.

> Can you post some benchmarks?

Here's the example:

```julia
m = Dense(32, 32)
x = rand(Float32, 32, 32)
y = rand(Float32, 32)
```
```julia
@benchmark gradient(Flux.params(m)) do
    sum(abs.(m(x) .- y)) / length(y) 
end
```
```
BenchmarkTools.Trial: 
  memory estimate:  114.19 KiB
  allocs estimate:  3197
  --------------
  minimum time:     136.582 μs (0.00% GC)
  median time:      140.018 μs (0.00% GC)
  mean time:        153.322 μs (7.62% GC)
  maximum time:     12.375 ms (95.82% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

```julia
@benchmark gradient(Flux.params(m)) do
    sum(abs.(m(x) .- y)) * 1 // length(y) 
end
```

```
BenchmarkTools.Trial: 
  memory estimate:  97.72 KiB
  allocs estimate:  3196
  --------------
  minimum time:     78.068 μs (0.00% GC)
  median time:      82.311 μs (0.00% GC)
  mean time:        94.679 μs (11.36% GC)
  maximum time:     12.579 ms (96.22% GC)
  --------------
[...]

PR #1264, opened 2020-07-02
create Losses module

CarloLucibello:
Continuation of #1150, grouping losses under a `Losses` module as discussed. 

An alternative for the module name could be `Loss`, but `Losses` seemed more natural.

I also renamed `bce` back to `binarycrossentropy`, since now that the function is within a module I could provide I deprecation path without changing the function name with respect to last tagged release.

Some of the function contain the  `_loss` suffix, (e.g. `hinge_loss`, `poisson_loss`). We could drop that now that we have a namespace disambiguating the meaning, but again it seems more natural to keep, closer to the way people referer to them when speaking

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


johnnychen94:
> An alternative for the module name could be Loss, but Losses seemed more natural.

How about `Metrics`?

CarloLucibello:
> How about Metrics as it can be a broader concept than Losses?

mhmh, don't really fancy that, I usually think of metrics as something you keep track of, not optimize

DhairyaLGandhi:
`Metrics` sounds better to my ear as well, the metrics that you mentioned also just sounds like the tracking information that would just be used to see training progress,

johnnychen94:
Disclaimer: I don't have a strong preference here. I proposed `Metrics` because

* it is not uncommon to track losses during training, using `Metrics` provides a place for other non-loss metrics, such as accuracy, psnr, ssim, etc.
* `Losses` has three `s` and I sometimes incorrectly type it as 😢 `loses`.
* `loss` is either a verb or a noun, while `metric` is mostly used as a noun. Using `metric`  as a module name has a clearer meaning.

CarloLucibello:
Keras has two separate modules, Losses, and Metrics
https://keras.io/api/metrics/
"A metric is a function that is used to judge the performance of your model.
Metric functions are similar to loss functions, except that the results from evaluating a metric are not used when training the model. Note that you may use any loss functions as a metric function."

pytorch instead has only losses, with metrics typically hand-coded by users. 

We currently only have only loss function here, not arbitrary metrics, so I think `Losses` would be a more appropriate name, something which conveys a clear sense of the primary purpose. If the concern is about future expandability, we could add a `Metrics` module at any point, possibly re-exporting `Losses` as well.

CarloLucibello:
@DhairyaLGandhi merge?

CarloLucibello:
CI failure likely due to https://github.com/JuliaDiff/ChainRules.jl/issues/227

src/Flux.jl
```diff
@@ -35,14 +35,16 @@ include("onehot.jl")
...
-include("layers/losses.jl")
 include("layers/basic.jl")
 include("layers/conv.jl")
 include("layers/recurrent.jl")
 include("layers/normalise.jl")
 
 include("data/Data.jl")
 
+include("losses/Losses.jl")
+using .Losses # TODO: stop importing Losses in Flux's namespace in v0.12
```
DhairyaLGandhi, line 46: Loss should export all the functions as before.

src/losses/Losses.jl
```diff
@@ -0,0 +1,271 @@
...
+mae(ŷ, y; agg=mean) = agg(abs.(ŷ .- y))
+
+"""
+    mse(ŷ, y; agg=mean)
+
+Return the loss corresponding to mean square error: 
+    
+    agg((ŷ .- y).^2)
+"""
+mse(ŷ, y; agg=mean) = agg((ŷ .- y).^2)
```
DhairyaLGandhi, line 38: Since there is a regression wrt to performance, maybe we should consider removing the aggregation for now

test/cuda/cuda.jl
```diff
@@ -4,53 +4,37 @@ using Flux: cpu, gpu
...
 using LinearAlgebra: I, cholesky, Cholesky
 
-@info "Testing GPU Support"
-
 @testset "CUDA" begin
+  x = randn(5, 5)
+  cx = gpu(x)
+  @test cx isa CuArray
 
-CUDA.allowscalar(false)
```
DhairyaLGandhi, line 11: Why is this PR moving around tests? The tests should be largely untouched and esp tests unrelated to losses should be untouched

DhairyaLGandhi:
The package shouldn't move the definition of losses to a module like so. See how Optimisers.jl handles the directory structure.

DhairyaLGandhi:
The package shouldn't move the definition of losses to a module like so. See how Optimisers.jl handles the directory structure. I would also consider calling this package Loss.jl or similar otherwise it sounds weird.

src/Flux.jl
CarloLucibello: It is, that is no loss is exported but all live in the Flux namespace, as they currently are. The comment is about keep the losses in Flux.Losses namespace, something we may want to do in v0.12

src/losses/Losses.jl
CarloLucibello: if you refer to #1255, `mean` is not affected by it

test/cuda/cuda.jl
CarloLucibello: I had to reorganize a bit the cuda tests, in order to reflect the presence of the new `Losses` module and the new test/cuda/losses.jl file

src/losses/Losses.jl
```diff
@@ -0,0 +1,35 @@
...
+    tversky_loss,
+    dice_coeff_loss,
+    poisson_loss,
+    hinge_loss, squared_hinge_loss
+
+include("utils.jl")
+include("functions.jl")
+
+# group here all losses, used in tests
+const ALL_LOSSES = [mse, mae, msle,
```
DhairyaLGandhi, line 25: Should be moved over to the tests instead.

src/losses/Losses.jl
DhairyaLGandhi: I mean specifically in the backwards pass.

src/losses/Losses.jl
CarloLucibello: I wasn't sure where to put it in tests, since it's used in two separate files

DhairyaLGandhi: Before the first stateless layer test is run.

CarloLucibello:
> I would also consider calling this package Loss.jl or similar otherwise it sounds weird to my ear

I fear using the singular here may go against julia's guidelines for module naming. Any opinion from @MikeInnes @oxinabox ?

src/losses/Losses.jl
CarloLucibello: which regression?

DhairyaLGandhi:
We should also make sure that we don't message things incorrectly. Currently it seems that we are saying that these bit of code are the loss functions which are somehow separate from being some generic transform that we can differentiate through. Continuing calling them stateless layers makes that a little bit more explicit. This is an important distinction to make for a newcomer to our ecosystem.

src/losses/Losses.jl
DhairyaLGandhi: This is with Flux@0.10.4. `cmse` is the one from master, `mymse` is a current implementation that I am working on. 

image

CarloLucibello: ok. But we can't remove aggregation now and add it back later, that's a lot of breakage. We should just improve performances under the hood. If you have a better implementation for `agg=mean` we can add a branch for that. But really for such basic functions as `mean((x .- x).^2)` we should expect Zygote to be perfomant, and track down the issue if it is not

src/losses/Losses.jl
CarloLucibello: done

DhairyaLGandhi:
bors try

DhairyaLGandhi:
Hopefully we haven't missed any tests

PR #1281, opened 2020-07-12
Ignore Zeros when gathering parameters

### LINKED ISSUE #1277: Flux.Zeros conflicts with Flux.loadparams!
Consider a model where all convolutions have their biases disabled:

```julia
m = Chain(
    Conv((3, 3), 1 => 2, bias=Flux.Zeros()),
    Conv((3, 3), 2 => 3, bias=Flux.Zeros()),
    Conv((3, 3), 3 => 4, bias=Flux.Zeros()),
)
```

If I collect its parameters (e.g. for saving them later) and print:

```julia
mp = m |> params .|> cpu
for p in mp
    println("$(typeof(p)) $(size(p))")
end
```

in the output I will have:

```
Array{Float32,4} (3, 3, 1, 2)
Array{Bool,0} () <--- disabled bias from 1st conv
Array{Float32,4} (3, 3, 2, 3)
Array{Float32,4} (3, 3, 3, 4)
```

So parameters of the model contain disabled bias, but only from the 1st layer (since parameters are stored in `Zygote.IdSet` others are ignored).
Moreover if I want to load these parameters into the model:

```julia
Flux.loadparams!(m, mp)
```

I get error `ERROR: setindex! not defined for Flux.Zeros{Bool,0}`.

Easy fix would be to define `params!` for `Flux.Zeros` that does not add it to parameters:

```julia
params!(p::Flux.Params, x::Flux.Zeros{<:Number}, seen = Flux.IdSet()) = nothing
```

I wanted to create a PR with this fix, but this solution conflicts with [this test](https://github.com/FluxML/Flux.jl/blob/master/test/layers/conv.jl#L49) which expects `Flux.Zeros` to be in parameters.
Maybe change it to instead check that `bias` is not in gradients?

pxl-th:
This fixes #1277.
IMO, it makes more sense to ignore disabled parameters when gathering model parameters, than have something that is not used and may lead to confusion.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -37,6 +37,8 @@ Possible values include:
 """
 trainmode!(m, mode = true) = mode isa Bool ? testmode!(m, !mode) : testmode!(m, mode)
 
+params!(p::Flux.Params, x::Zeros{<:Number}, seen = Flux.IdSet()) = nothing
+
 params!(p::Params, x::AbstractArray{<:Number}, seen = IdSet()) = push!(p, x)
 
 function params!(p::Params, x, seen = IdSet())
--- a/test/layers/conv.jl
+++ b/test/layers/conv.jl
@@ -46,7 +46,7 @@ end
   op = bias(ip)
   @test sum(op) ≈ 0.f0
   gs = gradient(() -> sum(bias(ip)), Flux.params(bias))
-  @test gs[bias.bias] == nothing
+  @test !haskey(gs, bias.bias)
 
   # Train w/o bias and make sure no convergence happens
   # when only bias can be converged
--- a/test/layers/conv.jl
+++ b/test/layers/conv.jl
@@ -43,6 +43,15 @@ end
   @test sum(op) == prod(size(op))
 
   bias = Conv((2,2), 1=>3, bias = Flux.Zeros())
+
+  # Test that disabled bias is not in parameters
+  # and gathered parameters can be loaded back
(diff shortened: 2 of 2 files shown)


DhairyaLGandhi:
Removing Zeros from the list params might give incorrect results for loading complete chains where it is expected to see the bias term appear.

pxl-th:
In this case there is an issue, if we have a model with all biases disabled:

```julia
m = Chain(
    Conv((3, 3), 1=>2, bias=Flux.Zeros()),
    Conv((3, 3), 2=>3, bias=Flux.Zeros()),
    Conv((3, 3), 3=>4, bias=Flux.Zeros()),
)
```

Then in parameters we have `Flux.Zeros` only from first convolution, since parameters are stored in `Zygote.Params` which wraps `IdSet{Any}` and so we effectively loose information about all other disabled biases.

```julia
for p in Flux.params(m)
    println(typeof(p))
end

>>> Output:
Array{Float32,4}
Flux.Zeros{Bool,0}
Array{Float32,4}
Array{Float32,4}
```

DhairyaLGandhi:
The crucial thing to note here is that we check for the presence of `Flux.Zeros` in the model side. That way the information is preserved. To make them different, we could make the struct mutable to create new instances of it with every call.

pxl-th:
Closing as it is not relevant anymore

Issue #337, opened 2018-07-26
CTC loss for Flux


maetshju:
Of the GSoC work I've done, I think that what's potentially most useful for inclusion in Flux proper is the CTC (connectionist temporal classification) loss function. At the moment, I'm thinking it would fit into the layers section, since that's where the other loss functions are. It seems to me that it would be better off in its own file, since the amount of code for calculating the loss function is quite large compared to your basic mse (for example). Currently, I only have the GPU version working at an acceptable speed.

Ideally, the user-facing interface would look just like the other loss functions, something like `ctc(ŷ, y)`, except that `ŷ` is a sequence, rather than one particular training example.

The GPU code will [look more or less like this](https://github.com/maetshju/gsoc2018/blob/master/speech-cnn/warp-ctc.jl). The interface function is at the bottom of the script.

Alternatively, this could exist in its own package that has easy interop with Flux. Thoughts on whether this would be useful for inclusion in Flux itself?

jekbradbury:
I think it's fine for it to be its own package (CTC.jl, if that's not taken) as long as it works directly with the same user-facing interface as Flux loss functions do (so you'll depend on Flux.jl and use Tracker to wrap the gradient)

CarloLucibello:
I'm no expert of recurrent network, but googling around it seems that CTC is famous enough to deserve inclusion in Flux itself

maetshju:
It has been used by some important speech and handwriting recognition networks for sure, but I do wonder whether it would fit that well with the rest of the Flux code, since there are bare GPU kernels, which are generally rougher to understand than the rest of Flux, in my opinion.

MikeInnes:
Let's just put it in Flux for now, and we can always split it out if it's easier for maintenance or whatever. But otherwise the AD churn is going to be a pain.

We'll need a CPU version if only for testing, it's fine to have something that's not crazy fast. It will also send a clear message of "here's the simple version, here's the fast version".

maetshju:
Opened a PR for CTC, so I'll close this issue.

PR #1287, opened 2020-07-21
Add CTC loss to new Losses module

maetshju:
This is a redux of adding the connectionist temporal classification loss from https://github.com/FluxML/Flux.jl/pull/342, now that the Losses module has been merged in https://github.com/FluxML/Flux.jl/pull/1264. Discussion in https://github.com/FluxML/Flux.jl/pull/342 suggested that a new PR would be easier than rebasing.

Since the last commit in https://github.com/FluxML/Flux.jl/pull/342, functions and data structures from `CUDAnative.jl` and `CuArrays.jl` have been updated to work with `CUDA.jl`. This is in addition to incorporating the loss function into the Losses module.

### PR Checklist

- [X] Tests are added
- [X] Entry in NEWS.md


CarloLucibello:
has this code been tested in real applications?

maetshju:
That is a good question. I haven't used this code for an actual project since Google Summer of Code 2018 because the kind of sequence it produces is not useful for my research. So, I've been working on a couple projects this weekend to make sure it scales up to something beyond the test cases, and fixing things as necessary. They're a little slow because I hacked them together without regard for runtime, so I'll report back when I have a confirmation that it's working (hopefully tomorrow). Maybe one of them would be a good addition to the model zoo as an example of how to use the loss function too, if the runtime is acceptable enough.

maetshju:
So, I've been trying to get a sample network working on the TIMIT task, but I have been having trouble getting the network to converge in a reasonable amount of time. I'm only using the CPU implementation at the moment, and I'm reasonably confident that it's calculating the loss and gradients correctly. At least, they are similar to the outputs from [Chainer's CTC function](https://github.com/chainer/chainer) and the [Stanford CTC implementation](https://github.com/amaas/stanford-ctc).

As an additional check, I trained networks that should be equivalent in Flux and Chainer. One network in Flux used my CTC implementation, and the other used Chainer's CTC loss and gradients. The Chainer network just used its own CTC loss. Strangely, both Flux networks converged very slowly compared to the Chainer network on the same subset. I've checked on both TIMIT and a speech corpus publically available from the lab I work in, and I got similar results whether I used ADAM or SGD+Momentum as the optimizer. I'm currently looking at some runs through the entire data set my lab has, and training hasn't quite finished, but the convergence time is looking to be similar to the subsets of data, just that each epoch takes longer to run (approx. 1.5 days). I can post those results when I have them.

As far as I can tell, the issue isn't with the loss gradients or else I would expect that using the Chainer CTC loss and gradients would give similar performance in Flux as it does in Chainer. I've created [a GitHub repo](https://github.com/maetshju/flux-ctc-grad) to demonstrate the behavior I'm seeing, though it is not an exhaustive collection of everything I've tried. It uses the speech corpus my lab has available so that folks don't have to try to register for or purchase TIMIT if they don't already have it.

So, as far as the CPU implementation goes, it's looking like it will eventually successfully train a network on a non-trivial data set, though it takes substantially more epochs in Flux than other frameworks. At this point, I don't really know how to address the slow convergence. Perhaps what's wrong or what a good check to perform would be is obvious to someone who knows more about Flux and Zygote's current internals, though. And, maybe I'm just not feeding the custom gradient back correctly. Any thoughts, anyone?

maetshju:
On a whim, I saw a note about broadcasting being difficult to design the autodiff for, so I changed the model to use `map` instead of broadcasting to deal with the sequences. Suddenly, the Flux version was converging properly!

After playing with the gradients, I found that multiplying the gradients from the loss function by 10 resulted in the broadcasting version working as well as the `map` version. Multiplying the learning rate of the optimizer by 10 had a similar effect when using broadcasting.

Anyway, I'm hoping I can make a full demo model for both CPU and GPU in the next few days here, and I'll also see if I can create an MWE for the map vs. broadcasting difference and open an issue for that separately.

maetshju:
Sorry for the long delay; teaching this semester has taken up a lot of the time I could've otherwise devoted to resolving this.

But, I think this may finally be more or less ready. I have replicated the results from [a chapter in Alex Graves's (2012) book](https://www.cs.toronto.edu/~graves/preprint.pdf#subsection.7.6.1), which is itself a re-presentation of the results from the original paper. The replication works with both the CPU and GPU implementations for Flux v0.11.1. The replication code and results can be found in [this GitHub repo I've made](https://www.cs.toronto.edu/~graves/preprint.pdf#subsection.7.6.1), which have the same `ctc.jl` and `ctc-gpu.jl` files as in this commit. At one point, I had tested using Tracker for AD instead of Zygote and gotten similar results, though I have not reported on that in the repo.

I'm not sure if the v0.12.0 entries in the NEWS.md file are frozen yet for an upcoming release, but I added this to the v0.12.0 notes because I hadn't seen that version see a GitHub release yet. I am happy to make a new entry for v0.12.1 if preferred, though.

src/losses/Losses.jl
```diff
@@ -16,9 +16,12 @@ export mse, mae, msle,
     tversky_loss,
     dice_coeff_loss,
     poisson_loss,
-    hinge_loss, squared_hinge_loss
+    hinge_loss, squared_hinge_loss,
+    ctc
```
CarloLucibello, line 20: let's call it `ctc_loss` for consistency and for clarity

src/losses/ctc-gpu.jl
```diff
@@ -0,0 +1,291 @@
+# GPU impelmentation
```
CarloLucibello, line 1:
```suggestion
# GPU implementation
```

src/losses/ctc-gpu.jl
```diff
@@ -0,0 +1,291 @@
...
+# paper: https://arxiv.org/pdf/1512.02595.pdf
+
+using Flux
+using Statistics
+using CUDA
+
+const MAX_THREADS = 256
+
+function log_plus_f(p1, p2)
+
```
CarloLucibello, line 14: there a low of unnecessary white lines  throughout the PR

src/losses/ctc-gpu.jl
```diff
@@ -0,0 +1,291 @@
+# GPU impelmentation
+
+# a port of the GPU kernels from Baidu's C++ warp-ctc package
+# GitHub: https://github.com/baidu-research/warp-ctc/
```
CarloLucibello, line 4: I see that this code is under APACHE 2.0 license. Does this mean that the mention here is good enough or some additional action has to be taken?

src/losses/ctc-gpu.jl
```diff
@@ -0,0 +1,291 @@
...
+  isinf(p2) && return p1
+
+  if p1 < p2
+    p1, p2 = p2, p1
+  end
+
+  return p1 + CUDA.log(1+CUDA.exp(p2 - p1))
+end
+
+function countRepeats(A)
```
CarloLucibello, line 25: the camel case style is unjulian, change this to `count_repeats`

src/losses/ctc.jl
```diff
@@ -0,0 +1,184 @@
...
+
+  return a + log(1+exp(b-a))
+end
+
+"""
+  logsum(a)
+
+Sums the elements in `a` such that the result equals `log(sum(exp.(a)))`
+"""
+function logsum(a::AbstractArray)
```
CarloLucibello, line 33: the correct name for this is `logsumexp`. Let's adapt NNlib's softmax definition 
https://github.com/FluxML/NNlib.jl/blob/2e1fa5f330ba785b0adbaec5107ca85a8e181197/src/softmax.jl
for this:
```julia
function logsumexp(xs::AbstractArray; dims=:)
    max_ = maximum(xs, dims=dims)
    exp_ = exp.(xs .- max_)
    log_ = log.(sum(exp_, dims=dims))
    max_ .- log_
end
```
Even better would be to directly implement this in `NNlib`, I'll file a PR later

src/losses/ctc.jl
```diff
@@ -0,0 +1,184 @@
...
+using Statistics
+
+# CPU implementation
+
+"""
+  logadd(a, b)
+
+Adds log-space `a` and `b` such that the result equals `log(exp(a)+exp(b))`
+"""
+function logadd(a, b)
```
CarloLucibello, line 12: this function can be removed, see below

CarloLucibello:
> Sorry for the long delay; teaching this semester has taken up a lot of the time I could've otherwise devoted to resolving this.

I feel you, I've been through the same :sob:

Except for those mainly stylistic comments, this looks fine.

src/losses/ctc-gpu.jl
maetshju: I tend to use a lot of whitespacing, but I know not everyone likes that. I've tried to take out quite a bit of it, but I'm worried I've gone too far in that regard now.

src/losses/ctc-gpu.jl
maetshju: I've added a bit more to the header that I believe takes care of this. What I'm a bit less sure of is if it needs to be mentioned elsewhere in the project.

src/losses/ctc-gpu.jl
maetshju: And here I was thinking it was the opposite! I've changed the camel case function names to snake case.

src/losses/ctc.jl
maetshju: Should I wait for that function to be committed and then update the functions to use the `NNlib` code?

CarloLucibello: new NNlib version with `logsumexp` has been tagged, so you can use it here provided you increase the NNlib compat bound

src/losses/ctc.jl
```diff
@@ -120,20 +91,20 @@ function ctc_(ŷ, y)
...
+      β[t, u] = logsumexp(v)
     end
   end
   
 
   # Loss at each time t is taken as the sum of the product (sum in log space) of the
   # α and β coefficients for all the label classes at time t
   αβ = α + β
-  losses = -1 .* mapslices(logsum, αβ, dims=2)
+  losses = -1 .* mapslices(logsumexp, αβ, dims=2)
```
CarloLucibello, line 102: logsumexp has a dims keyword argument, so I guess you can avoid mapslices

CarloLucibello:
should update the Project.toml bound on NNlib, and update the Manifest as well

maetshju:
>should update the Project.toml bound on NNlib, and update the Manifest as well

Yes, I will get to that soon; thank you for the reminder. Before working on that, though, I'm currently encountering and trying to fix an issue in the `logsumexp` function in `NNlib`. If the array or a slice of an array has all `-Inf` values, `NaN` values are induced by doing subtraction on infinite values. For example,

```julia
using NNlib
a = [-Inf, -Inf]
logsumexp(a)
# NaN
```

It would make sense to me to return `-Inf` in this case because we are taking the log of the sum of 0s, but I am having trouble writing a patch that handles the `dims` argument nicely and catches these specific `NaN` values.

CarloLucibello:
Are we supposed to handle `Inf`s?  And even the more extreme case of all `Infs`? We don't handle that in e.g. softmax

maetshju:
Well, it happens frequently in the CTC loss calculation that we're doing a logsumexp operation on two `-Inf` values. I could keep the `logadd` function above that just accepts two scalars and only use `logsumexp` on arrays. I think this would make it unnecessary for the function in `NNlib` to handle all `Inf`s, at least as far as being useful for CTC would be.

CarloLucibello:
ok, let's keep this confined to the ctc loss

CarloLucibello (dismissed):

maetshju:
I didn't realize that adding a new commit would dismiss a review. I had tied some of the whitespace changes into the commit that I reverted, so this new commit just fixes the whitespace again and changes the comment on the numerical gradient function. Otherwise, I think it's ready.

CarloLucibello (dismissed):

CarloLucibello:
need also @DhairyaLGandhi's approval

DhairyaLGandhi:
Is it possible for us to have a single implementation for both cpu and GPU? We usually try and have one implementation which works well for both (normally that means testing with `CUDA.allowscalar(false)`. What challenges do you see there?

I'm somewhat concerned about the large code addition too.

src/losses/Losses.jl
```diff
@@ -17,9 +17,12 @@ export mse, mae, msle,
...
     dice_coeff_loss,
     poisson_loss,
-    hinge_loss, squared_hinge_loss
+    hinge_loss, squared_hinge_loss,
+    ctc_loss
 
 include("utils.jl")
 include("functions.jl")
+include("ctc.jl")
+if CUDA.functional() include("ctc-gpu.jl") end
```
DhairyaLGandhi, line 26: This should go into Flux.jl or be removed

src/losses/ctc-gpu.jl
```diff
@@ -0,0 +1,252 @@
...
+function log_plus_f(p1, p2)
+  isinf(p1) && return p2
+  isinf(p2) && return p1
+  if p1 < p2
+    p1, p2 = p2, p1
+  end
+  return p1 + CUDA.log(1+CUDA.exp(p2 - p1))
+end
+
+function count_repeats(A)
```
DhairyaLGandhi, line 27: Seems like you should be able to use `Base.count` here

src/losses/ctc.jl
```diff
@@ -0,0 +1,159 @@
...
+  end
+  return z
+end
+
+"""
+  add_blanks(z)
+
+Adds blanks to the start and end of `z`, and between item in `z`
+"""
+function add_blanks(z, blank)
```
DhairyaLGandhi, line 50: Use `Base.repeat` here?

DhairyaLGandhi:
Can we try to find if the cudnn version of CTC loss also suits our needs? https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnCTCLoss_v8

Best to check runtime performance as a follow up

maetshju:
> Is it possible for us to have a single implementation for both cpu and GPU? We usually try and have one implementation which works well for both (normally that means testing with CUDA.allowscalar(false). What challenges do you see there?

I agree that having a single implementation would be ideal. I have thought quite a lot about how this might be done, but I haven't really been able to come up with anything reasonable. The hard part is that it's a dynamic programming algorithm with lots of index math that is hard to parallelize/vectorize. At most, we would be able to compute one row or column of the dynamic programming matrix in parallel (which is what the GPU kernel does). I don't really know how or if there's is a way to express the computation using dot broadcasting or otherwise would invoke the appropriate conversions for CUDA kernels and allow us to have just one codebase.

> I'm somewhat concerned about the large code addition too.

I share this concern because I have had a tough time determining what was causing issues when they came up. It is a big ask for this code to be included because even the algorithm itself is not that straightforward. @MikeInnes and I had talked about maybe making a new package out of the code instead in a previous Issue https://github.com/FluxML/Flux.jl/issues/337, but we ended up thinking including it in Flux was a better idea at the time. Part of the motivation was dealing with the AD system, which was Tracker at the time. Now that Zygote is a separate package, I don't know that the AD churn motivation is still valid. In the previous PR https://github.com/FluxML/Flux.jl/pull/342, we had also floated the idea of incorporating it into one of the GPU packages, but that didn't pan out as a great idea at the time.

I can take another pass through the code to see what simplifications I can make, but I don't think they would be too significant. One thing that would substantially reduce the amount of code needed would be if Zygote handled array mutation because we would only need to calculate the first dynamic programming matrix and can determine gradients based off the operations that fill the matrix in, which had been suggested to me by @rakeshvar in the repo that I was using to demonstrate that the CTC function was working properly https://github.com/maetshju/flux-ctc-grad/issues/1. Though, I don't know if this would make the GPU kernel unnecessary, either, unless Zygote reaches into the GPU kernels.

>Can we try to find if the cudnn version of CTC loss also suits our needs? https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnCTCLoss_v8

I can certainly take a look at this if we think it's worthwhile. Earlier, I think part of the idea was that including the GPU kernel would demonstrate that arbitrary GPU kernels can be tied into Flux relatively easily, but I understand that it's very possible that opinions and priorities have changed since then. From my understanding of the CuDNN stuff, we would still need the CPU implementation if we want to let people use the code with CPU networks.

ToucheSir:
I think this is a good argument for eventually moving Flux's loss module into an external package like https://github.com/JuliaML/LossFunctions.jl. Not now, but worth a revisit soon.

rakeshvar:
I have a CPU implementation that is like 10-30 times faster than @maetshju 's on the CPU for typical problem sizes. 
The code is also smaller and cleaner. 
I have a repo, that explore all angles of CTC implementation for Julia. 
I am new to this PR business, also never worked with GPU in Julia (did so in Python AD packages).
[Explore CTC Loss Repo.](https://github.com/rakeshvar/Explore-CTC-Loss.jl)

src/losses/ctc.jl
maetshju: I wasn't entirely clear how `Base.repeat` might be used here, so I ended up pre-allocating the final array with `fill` and set values as needed.

maetshju:
Thank you, @rakeshvar for showing the inefficiencies of the current CPU implementation! There had been a discussion about having the CPU function be rather simple mostly for testing purposes and to encourage people to use the GPU version instead. So, I had not really tried to optimize it. I've now drastically reduced the number of allocations done during the execution such that my implementation's runtime is on par with the `ctc_log_manual` function you reference, while also allocating very little memory.

I will review the GPU kernel and connecting with `CuDNN` after the Christmas holiday is over.

DhairyaLGandhi:
We can incorporate the remaining performance fixes on the cpu as well, if not done already.

One possibility is that we write an adjoint for a kernel that does the mutation which is usually simple and I'd be happy to help with that too. I'm curious about vectorisation/ fusion opportunities that you can see on the GPU still so we have a good outlook on what future work we can pick up here and have an implementation part of the next release, with performance changes following it.

maetshju:
One option I've played around with is moving the second dynamic programming matrix calculation into the adjoint since the gradients are not necessary if all someone is interested in is the loss value. It helps with readability some by separating out the functionality to some degree. The most recent commit shows this.

What I have been struggling with is getting an appropriate adjoint for the kernel doing mutation. I had interpreted this to mean writing an adjoint for the mutation operations, though perhaps I am mistaken. In any case, I can't seem to get the gradients to reach all the way from the loss value back through the assignments into the array and on back to the pre-softmax network activations, even for dummy gradients just to see if things are working. I'm not sure if anyone has any ideas on what the appropriate adjoint might look like.

I also attempted to try CuDNN's CTC function through `CUDA.jl` and couldn't get it running after a few days' work. The Nvidia documentation is vague enough that I can't tell what the API is asking for in terms of tensor shape and dimension order, and the setup prior to running the actual function is rather intricate. I don't know whether the reduction in code and potential speedup is worth it at the moment. Though, their function would handle batches in parallel as one of the dimensions of an array/tensor, while the current implementations here don't (yet) handle batches.

As regards fusion/vectorization, there is a data dependence where calculating the values in later columns requires prior columns to be calculated. So, just about the only thing I think would work is to have a function that was broadcast or vectorized over one dimension of a DP matrix at a time and took previous columns or rows as additional arguments as needed. I expect that the performance would suffer from additional overhead, especially on the GPU when there would be as many calls onto it as there are time steps in the neural net output.

src/losses/ctc.jl
```diff
@@ -0,0 +1,152 @@
...
+		  α[u,t] = logaddexp(α[u,t], α[u-2,t-1])
+		end
+	  end
+	  α[u,t] += ŷ[z′[u], t]
+    end
+  end
+  return (loss=-1 * logaddexp(α[end,T], α[end-1, T]), alpha=α, zprime=z′, logsoftyhat=ŷ)
+end
+  
+@adjoint function ctc_(ŷ, y)
```
CarloLucibello, line 87: a separate adjoint should be written for CuArray input, right?

src/losses/ctc-gpu.jl
CarloLucibello: this is counting repetitions in adjacent elements, cannot be implemented with count

src/losses/Losses.jl
CarloLucibello: Losses is its own module, this should stay here

CarloLucibello:
Since this has been in the working since August 2018 (#342) let's leave aside any further performance considerations (those can be eventually be addressed in later PRs), interface with cudnn, and so on...., let's just focus on correctness.

I would be glad to have only the cpu implementation if the gpu one poses any issues, but this doesn't seem the case. 

>  In any case, I can't seem to get the gradients to reach all the way from the loss value back through the assignments into the array and on back to the pre-softmax network activations, even for dummy gradients just to see if things are working. I'm not sure if anyone has any ideas on what the appropriate adjoint might look like.

So the current adjoint implementation is broken?

src/losses/ctc.jl
maetshju: With the split between the loss calculation and the gradients with the adjoint for the CPU kernel, yes, this would now be necessary. With the omnibus function from before, it was not necessary to have a separate adjoint for the GPU because both the CPU and GPU hidden functions returned both loss and gradients. If we would prefer to merge the PR faster (which I am all in favor of), I can revert the commit that did the split, making the extra adjoint unnecessary.

maetshju:
> So the current adjoint implementation is broken?

No, the current adjoint implementation is working fine. I had been thinking through how a small adjoint for a mutation operation might be handled so that Zygote could handle the gradients and use less code overall, but I didn't make much progress there, and I didn't commit any code where I was toying with that.

> let's leave aside any further performance considerations (those can be eventually be addressed in later PRs), interface with cudnn, and so on...., let's just focus on correctness.

I am happy to just focus on correctness for now. I am more confident in the unsplit kernels, so I could revert the last commit and make sure the tests still pass (and they very well should), after which I believe both implementations would be correct. I have been running the tests for the GPU functions on my own machine as I've been making these changes, so the GPU function should be fine too. Some of the tests compare the GPU output with the CPU output, so they should be equivalent as long as the tests are passed.

CarloLucibello:
I common pattern we have for the adjoint definitions would look like this:
```
ctc_loss(yhat::AbstractArray, y) = ... 
ctc_loss(yhat::CuArray, y) = ... 

∇ctc_loss(Δ, yhat::AbstractArray, y, out) = ... 
∇ctc_loss(Δ, yhat::CuArray, y, out) = ...
 
@adjoint function ctc_loss(yhat, y)
     out = ctc_loss(yhat, y)
     ctc_loss_pullback(Δ) = (∇ctc_loss(Δ, yhat::CuArray, y, out), nothing)
     return out, ctc_loss_pullback
end
```
If more calculations are shared between `ctc_loss` and `∇ctc_loss`  they could be factorized and used in the `adjoint` definition.

maetshju:
I think that most recent commit should take care of the interface and adjoint definitions.

I have another one I haven't pushed yet because I'm not sure if it would be better to wait and open a new PR for it or not. But, it would change the way the target argument is handled so that it would accept a onecold vector instead of a onehot matrix. This makes it easier to handle input that involves repeated labels, such as the repeated "m" in "programming". The current implementation can handle it, but it requires the user to put a blank symbol between the repeated characters so they don't get collapsed (e.g., the onehot matrix equivalent of "program_ming").

src/losses/ctc.jl
```diff
@@ -0,0 +1,157 @@
...
+end
+
+"""
+  F(A, blank)
+
+Removes blanks and repetitions in the sequence `A`
+
+This is the function `F` as defined in Graves (2012)
+"""
+function F(A, blank)
```
CarloLucibello, line 33: can we give a longer name to this? introducing a one-letter function in the Losses namespace is not ideal

CarloLucibello:
It is better to choose know whether to commit to one of the 2 interfaces or to both.
I am no expert of NLP so I'm not the best judge here, but the pros I see for the onecold only implementation are 
+ the ones you mentioned 
+ the fact the also pytorch's CTCLoss takes as input the onecold representation https://pytorch.org/docs/stable/generated/torch.nn.CTCLoss.html
+ one can directly feed to the loss the dataset's label, without going through the onehot representation

Are there any cons? Why  should we support the onehot representation at all?

src/losses/ctc.jl
maetshju: I think that's a good idea. Changing to the onecold representation makes this unnecessary for the CTC loss calculation since the user is expected to do it themselves. It makes me wonder if perhaps this function should be moved to utils or something because this function would end up being used more often while interpreting the output of a network trained with the CTC loss.

maetshju:
I would prefer to have the onecold representation, and I don't think there are really any cons. The onehot expectation was a holdover from when I first impelemented the function during GSoC '18 and was still coming to grips with how CTC was implemented. The only sort of plus for having the onehot representation is that it is parallel to the standard format for cross entropy calculations, but I don't think it's a particularly compelling reason to keep the onehot representation.

I'm currently re-running my test networks to verify that the loss still works with the onecold representation and can push those changes once I see that the networks still train.

maetshju:
My test networks have now trained to a similar level of accuracy as when we had the onehot representation. This commit also removes the `F` function from the CTC code because it isn't used anymore in calculating the loss. If we would like to provide such a function, I think it might fit in utils, but it's really more of a function that's used to interpret the network outputs, so I'm not sure that we necessarily need to provide that here.

test/ctc.jl
```diff
@@ -0,0 +1,61 @@
...
+  @test all(isapprox.(g1, g2, rtol=1e-5, atol=1e-5))
+  
+  # tests using hand-calculated values
+  x = [1. 2. 3.; 2. 1. 1.; 3. 3. 2.]
+  y = [1, 2]
+  @test ctc_loss(x, y) ≈ 3.6990738275138035
+
+  g = [-0.317671 -0.427729 0.665241; 0.244728 -0.0196172 -0.829811; 0.0729422 0.447346 0.16457]
+  ghat = gradient(ctc_loss, x, y)[1]
+  @test all(isapprox.(g, ghat, rtol=1e-5, atol=1e-5))
```
CarloLucibello, line 52:
```suggestion
  @test g ≈ ghat  rtol=1e-5 atol=1e-5
```

test/ctc.jl
```diff
@@ -0,0 +1,61 @@
...
+
+  x = [-3. 12. 8. 15.; 4. 20. -2. 20.; 8. -33. 6. 5.]
+  y = [1, 2]
+  @test ctc_loss(x, y) ≈ 8.02519869363453
+
+  g = [-2.29294774655333e-06 -0.999662657278862 1.75500863563993e-06 0.00669284889063; 0.017985914969696 0.999662657278861 -1.9907078755387e-06 -0.006693150917307; -0.01798362202195 -2.52019580677916e-20 2.35699239251042e-07 3.02026677058789e-07]
+  ghat = gradient(ctc_loss, x, y)[1]
+  @test all(isapprox.(g, ghat, rtol=1e-5, atol=1e-5))
```
CarloLucibello, line 60:
```suggestion
  @test g ≈ ghat   rtol=1e-5 atol=1e-5
```

test/ctc.jl
```diff
@@ -0,0 +1,61 @@
...
+  filter!(x -> x != blank, z)
+  return z
+end
+
+@testset "ctc_loss" begin
+  x = rand(10, 50)
+  y = F(rand(1:9, 30), 10)
+  g1 = gradient(ctc_loss, x, y)[1]
+  g2 = ctc_ngradient(x, y)
+  @test all(isapprox.(g1, g2, rtol=1e-5, atol=1e-5))
```
CarloLucibello, line 43:
```suggestion
  @test g1 ≈ g2  rtol=1e-5 atol=1e-5
```

src/losses/ctc.jl
```diff
@@ -0,0 +1,138 @@
...
+
+Computes the connectionist temporal classification loss between `ŷ`
+and `y`.
+
+`ŷ` must be a classes-by-time matrices, i.e., each row
+represents a class and each column represents a time step.
+Additionally, the `logsoftmax` function will be applied to `ŷ`, so
+`ŷ` must be the raw activation values from the neural network and
+not, for example, the activations after being passed through a
+`softmax` activation function. `y` must be a 1D `Array` of the labels
```
CarloLucibello, line 121:
```suggestion
`softmax` activation function. `y` must be a 1D array of the labels
```
as we accept generic AbstractArrays

test/ctc.jl
```diff
@@ -0,0 +1,61 @@
...
+    end
+    prev = curr
+  end
+  filter!(x -> x != blank, z)
+  return z
+end
+
+@testset "ctc_loss" begin
+  x = rand(10, 50)
+  y = F(rand(1:9, 30), 10)
```
CarloLucibello, line 40: do we still need the F function here for testing? Can't we just test passing the labels?

CarloLucibello (changes requested):

test/ctc.jl
maetshju: We do not need it anymore, no. I must have overlooked it. I'll remove it shortly.

maetshju: It has been removed now.

src/losses/ctc.jl
```diff
@@ -0,0 +1,138 @@
...
+	  if u == 1
+		α[u,t] = α[u, t-1]
+	  else
+		α[u,t] = logaddexp(α[u, t-1], α[u-1, t-1])
+		
+		# array bounds check and f(u) function from Eq. 7.9
+		if u > 2 && !(z′[u] == blank || z′[u-2] == z′[u])
+		  α[u,t] = logaddexp(α[u,t], α[u-2,t-1])
+		end
+	  end
```
CarloLucibello, line 60: indentation here is weird

src/losses/ctc.jl
```diff
@@ -0,0 +1,138 @@
...
+    for u=bound:-1:1
+	  if u == U′
+		β[u,t] = ŷ[z′[u], t+1] + β[u, t+1]
+	  else
+		β[u,t] = logaddexp(ŷ[z′[u], t+1] + β[u, t+1], ŷ[z′[u+1], t+1] + β[u+1,t+1])
+
+		# array bounds check and g(u) function from Eq. 7.16
+		if u+2 <= U′ && z′[u] != blank && z′[u] != z′[u+2]
+		  β[u,t] = logaddexp(β[u,t], ŷ[z′[u+2], t+1] + β[u+2, t+1])
+	    end
```
CarloLucibello, line 93: indentation

maetshju:
Sorry about that. My text editor inserted some tab characters for some reason but didn't display the file any differently on my end. I have replaced them with spaces and checked the other files in the PR for any errant tabs (no other files seemed to have been affected though).

CarloLucibello:
It's been a long journey, thanks for your effort!

bors r+

CarloLucibello (approved):

DhairyaLGandhi:
@CarloLucibello this does count as an api change and was merged errantly. We still don't want different implementations for cpu and GPU, and this shouldn't change that.

DhairyaLGandhi:
Thanks for all the work @maetshju, hopefully we would be able to see the two I'm implementations converge soon. We might need to tweak some things in cuda.jl, so it's good to have a reference to compare against

maetshju:
Thank you @CarloLucibello and @DhairyaLGandhi for all your attention and reviews on this code. Please don't hesitate to tag me in related future discussions or bug reports if you would like my input.

PR #1290, opened 2020-07-21
Corrections in the Optimisers section of documents

coldinjection:
`update!(x, x̄)` updates the array x according to `x .-= x̄`, so the minus sign is not needed when passing the argument
`update!(p, -η * grads[p])` will actually increase the loss, this is just wrong
also, `Params`  is not imported in the context, should use `params` instead


CarloLucibello:
Thanks

bors r+

PR #1292, opened 2020-07-24
Expose RNG in initializers

### LINKED ISSUE #1274: Expose the RNGs used in initialization to the user
It seems that there is currently no interface point for users to specify the random number generators used in Flux, unless they write their own initialisers. For reproducibility this would be good to add.

Maybe something like `glorot_uniform(rng::AbstractRNG, dims...)=...`, etc

findmyway:
This fixes #1274 . Beyond that, a handy function (`initializer(rng)` where initializer may be `glorot_uniform`, `glorot_normal`, etc) is added to create an initializer with specific RNG.

### PR Checklist

- [x] Tests are added. (This pr is simple enough. I can add if required)
- [ ] Entry in NEWS.md
- [x] Documentation, if applicable.
- [ ] Final review from @DhairyaLGandhi  (for API changes).

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -131,8 +131,8 @@ function kaiming_uniform(rng::AbstractRNG, dims...; gain = √2)
   return (rand(rng, Float32, dims...) .- 0.5f0) .* 2bound
 end
 
-kaiming_uniform(dims...;kwargs...) = kaiming_uniform(Random.GLOBAL_RNG, dims...;kwargs...)
-kaiming_uniform(rng::AbstractRNG;kwargs...) = (dims...;kwargs...) -> kaiming_uniform(rng, dims...;kwargs...)
+kaiming_uniform(dims...; kwargs...) = kaiming_uniform(Random.GLOBAL_RNG, dims...; kwargs...)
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -122,13 +126,16 @@ julia> Flux.kaiming_uniform(3, 2)
...
 """
-function kaiming_uniform(dims...; gain = √2)
+function kaiming_uniform(rng::AbstractRNG, dims...; gain = √2)
   bound = Float32(√3 * gain / sqrt(first(nfan(dims...)))) # fan_in
-  return (rand(Float32, dims...) .- 0.5f0) .* 2bound
+  return (rand(rng, Float32, dims...) .- 0.5f0) .* 2bound
 end
 
+kaiming_uniform(dims...;kwargs...) = kaiming_uniform(Random.GLOBAL_RNG, dims...;kwargs...)
+kaiming_uniform(rng::AbstractRNG;kwargs...) = (dims...;kwargs...) -> kaiming_uniform(rng, dims...;kwargs...)
```
CarloLucibello, line 135: whitespace after `;`

findmyway: Fixed. Thanks.

PR #1297, opened 2020-07-31
Change CuArrays to CUDA on docs homepage

scimas:
Spent too much time figuring out why my environment wouldn't update from v0.10.4 to v0.11.0. Even trying to force Pkg with `up Flux@0.11` wouldn't work. Had to remove the CuArrays package from project. Perhaps there needs to be a note somewhere about upgrading from <=v0.10.4 to >=v0.11.0?

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -12,6 +12,8 @@ Download [Julia 1.0](https://julialang.org/) or later, if you haven't already. Y
 
 If you have CUDA you can also run `] add CUDA` to get GPU support; see [here](gpu.md) for more details.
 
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
Thank you for pointing out the remaining CuArrays reference.

We could add a note right there which says something to that effect in one line.

scimas:
I have added a note to that effect.

Off topic: Your @ mention seems to have changed since the PR template was created, perhaps update that too.

PR #1298, opened 2020-08-01
[WIP] Per-parameter optimiser options

scimas:
Makes all options (learning rate, momentum etc) for all optimisers per-parameter specifiable. [PyTorch supports this](https://pytorch.org/docs/stable/optim.html#per-parameter-options) already. It is useful for transfer learning, where you may want to tune different layers at different rates.

The per-parameter options are specified using `IdDict`s, these are in addition to the default option values. If a parameter doesn't have its own value for some option, the default value will be used instead.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/training/optimisers.md
+++ b/docs/src/training/optimisers.md
@@ -41,6 +41,20 @@ end
 
 An optimiser `update!` accepts a parameter and a gradient, and updates the parameter according to the chosen rule. We can also pass `opt` to our [training loop](training.md), which will update all parameters of the model in a loop. However, we can now easily replace `Descent` with a more advanced optimiser such as `ADAM`.
 
+All optimisers also support per-parameter options, specified using `IdDict`s. For example, if we wanted to learn `b` at a slower rate than `W`, we could achieve it like this.
+
+```julia
+eta_dict = IdDict()
+eta_dict[b] = 0.01
+opt = Descent(0.1, eta_dict)
+
+for p in (W, b)
+  update!(opt, p, grads[p])
+end
+```
+
+This will use the default value 0.1 for `W` and 0.01 for `b` through the `eta_dict`.
+
 ## Optimiser Reference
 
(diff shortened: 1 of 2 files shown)


DrChainsaw:
Could a more scalable and generic approach be to have a new type which uses optimiser A for parameters in the dict and B for parameters not in the dict, or just map params to optimizers in the dict and have a convenience constructor to map all given params to same optimiser?

scimas:
> a new type which uses optimiser A for parameters in the dict and B for parameters not in the dict

Then you end up with potentially N different optimisers for N parameters.

DrChainsaw:
> Then you end up with potentially N different optimisers for N parameters.
Yes, but not necessarily. Depends on what the user wants, right?

Is there a problem with allowing it? One can just make the design so that it is not allowed, but I don't see why that would be needed.

DhairyaLGandhi:
The API for that is already scalable. We can only send in the params and gears we want updated to the `update!` function and have multiple optimisers. It feels a bit low level, tbh, but having this makes the whole system a lot more bloated, especially since optimisers are cheap to make. It's a paradigm we support (and is quite prevalent in research) but could do with a higher level documentation around the design pattern, rather than changing the API like so.

```julia
opt1 = Descent(0.1)
opt2 = Descent(0.2)

m = Chain(Dense(3,3), Dense(3,1))
r = rand(Float32, 3,10)
ps = Flux.params(m)
ps1 = Flux.params(m[1]) # convenience for demo
ps2 = Flux.params(m[2])

gs = gradient(()->sum(m(r)), ps)
Flux.Optimise.update!(opt1, ps1, gs[ps1])
Flux.Optimise.update!(opt1, ps2, gs[ps2])
```

scimas:
That functionality does exist, doesn't it? I don't know what I was thinking when I started making these changes, why I was thinking it is a good idea.

PR #1299, opened 2020-08-02
Fix ADADelta calculations and broken tests not catching the problems

### LINKED ISSUE #1158: ADADelta not training parameters
I tried training (using `train!`) a model (simple `Chain` of `Dense` layers) with `ADADelta` in Flux v0.10.4, and the parameters did not update. Same model trained just fine with `ADAGrad` and `Descent`. I also tried explicitly setting rho both high (0.95) and low (0.01) without seeing any updates.

I'm not familiar enough with the package to give more detained info, just starting out with both Julia and Flux.

I'm running Julia 1.4.1 and training on CPU only (no GPU).

scimas:
29832aca92748721594ea18067de2ba2ad5a077f broke ADADelta, simply reversing its change and adding comments to not move around the epsilons. Fixes #1158 

The testset probably needs more work than specifically for this case, so I'm not adding it here.
### PR Checklist

- [x] Tests are ~added~ fixed
- [ ] Entry in NEWS.md

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -283,7 +283,7 @@ mutable struct OADAM
   state::IdDict
 end
 
-OADAM(η = 0.0001, β = (0.5, 0.9)) = OADAM(η, β, IdDict())
+OADAM(η = 0.001, β = (0.5, 0.9)) = OADAM(η, β, IdDict())
 
 function apply!(o::OADAM, x, Δ)
   η, β = o.eta, o.beta
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -357,7 +357,9 @@ function apply!(o::ADADelta, x, Δ)
   ρ = o.rho
   acc, Δacc = get!(o.state, x, (zero(x), zero(x)))
   @. acc = ρ * acc + (1 - ρ) * Δ^2
-  @. Δ *= √Δacc/ (√acc + ϵ)
+  # DON'T remove epsilon from numerator
+  # or even out of the square roots
+  @. Δ *= √(Δacc + ϵ) / √(acc + ϵ)
   @. Δacc = ρ * Δacc + (1 - ρ) * Δ^2
   return Δ
 end
--- a/test/optimise.jl
+++ b/test/optimise.jl
@@ -2,12 +2,17 @@ using Flux.Optimise
 using Flux.Optimise: runall
 using Flux: Params, gradient
 using Test
+using Random
 
 @testset "Optimise" begin
+  # Ensure rng has different state inside and outside the inner @testset
+  # so that w and w' are different
+  Random.seed!(84)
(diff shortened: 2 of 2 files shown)


CarloLucibello:
thanks! can we fix tests so that we would have catched this?

scimas:
I kind of found why the test isn't working, but I have no idea how it is happening.
https://github.com/FluxML/Flux.jl/blob/1dc90238e8b44a976792e2a36141881a03971a82/test/optimise.jl#L6-L13

If you add an `@info loss(rand(10))` right after the loss function definition and before the for loop, it prints a loss of 0.0. Which means it hasn't been testing anything for who knows how long.

If you remove the `@testset` from `@testset for opt in [...]` it works correctly and reports non zero loss and that ADADelta is failing. I haven't checked other optimisers yet. Also, I don't know whether removing the `@testset` is the right way to go forward.

scimas:
Alright, figured it out. Every `@testset` execution resets the global RNG's state. Which means `w == w′` at all times. I will change that and see whether the tests pass.

DhairyaLGandhi:
Locally did you verify if there are other optimisers affected by this?

scimas:
Are you talking about the tests not doing anything or any other optimisers being affected by the epsilon like changes?

The tests should be fixed, I think. All of the optimisers are being tested in the same `@testset`, so the fix should apply to all of them.

I haven't checked other optimisers for algorithmic correctness. That would be next.

scimas:
Looks like OADAM might be broken, the test failed for only that one.

scimas:
Calculations of OADAM look fine to me. It does seem to have reduced the loss to a certain extent, just not as much as the `@test` is expecting (OADAM is taking it to ~0.025, test expects 0.01). Even giving it 10 times more iterations doesn't seem to show much improvement. Maybe OADAM is just not good at "optimising" random numbers (wild guess, not an optimisation theory expert)?

Perhaps changing the test to just check that the loss has decreased would be a good idea (as opposed to checking it has reduced below a hardcoded value of 0.01). Please provide other suggestions too.

Also should that be a separate PR? This isn't specific to fixing ADADelta anymore, it is already fixed.

DhairyaLGandhi:
It might mean that the default parameters of the optimiser aren't ideal.

scimas:
That is a good possibility. The paper itself does not suggest any default parameters, only tells which ones they used when conducting their experiments. On top of that, neither PyTorch nor Tensorflow have implemented OADAM, so that avenue of finding established defaults is closed too.

scimas:
Increasing the learning rate of OADAM by a factor of 10 is enough to make it pass the test. Should we change the default parameter value or only increase in the test?

DhairyaLGandhi:
The default one would be best, thanks!

scimas:
Ready for review

DhairyaLGandhi:
Thanks! Lgtm

bors r+

PR #1253, opened 2020-06-27
outdims: possible enhancements

hhaensel:
This PR is meant to discuss and develop further improvements to `outdims` as discussed in #1086 

I covered @HamletWantToCode 's proposal for a fallback (still without testing) and modified the dense version to keep all dimensions, as proposed by @darsnack .

Please feel free to also contribute commits.
Good night ;-)

Issue #1266, opened 2020-07-03
Single-argument `Flux.outdims` for layers with fixed output dimension?


lassepe:
As of https://github.com/FluxML/Flux.jl/pull/1252 calls to `Flux.outdims(l::Dense, isize)` check that `isize` machtes the input dimension of the dense layer. I appreciate this check and think it is the right thing to do if `isize` is provided. 

However, as the `Dense` layer has fixed input and output dimensions it might be convenient to also provide a single-argument `outdims` method (there are probably also other layers that would benefit from this). Of course, users could also directly check for the dimensions of the weights `l.W` to figure this out but field access feels like a rather unsafe way to achieve this since the field names are implementation details that may change.

While we are at it, it may also be useful to add `Flux.indims` for those fixed-dimension layers.

If this sounds like a useful addition, I am happy to create a PR.

AliMalik9599:
Hi @CarloLucibello, I am interested in working on enhancing this feature. I'm relatively new to Julia, any advice on getting started? Thanks!

CarloLucibello:
`Flux.outdims` no longer exists — it was superseded by `Flux.outputsize` and the `@autosize` macro. For a layer with fixed dimensions you can also just call `outputsize` with the known input size. Closing.

PR #1305, opened 2020-08-05
Updates to outdims

darsnack:
Since #1253 stalled, I tried committing to the author's branch, but I have not received a response. So, I am creating a new PR with the following changes from the previous one:
- `outdims` for generic functions
- Size checking for `outdims(::Dense, isize)`

I also added the following additional changes
- Removed type signature restrictions on `outdims` for generic functions
- Added `outdims` for normalization layers
    - This is helpful since `BatchNorm` etc. show up in a chain or array of layers frequently when model building
    - Right now there is a method error
    - Generic functions would address this, but I think we should avoid actually evaluating the function as much as possible
- Updated docs for `outdims` changes

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md
- [x] Final review from `@dhairyagandhi96` (for API changes).


darsnack:
Wanted to get the ball rolling on this. As you can see, the tests for the generic `outdims` fails. I think we should decide what API we want to support across all layers and functions as discussed in #1086.

I think the API should expect `isize` to be same number of dimensions as the input to the function (with the exception that the batch dimension can be left out). If the batch dimension is left out, I think we should assume it is 1 for generic function evaluation. The returned tuple should match the dimensions of `isize` (so if someone is ignoring batch, then we ignore it too).

DhairyaLGandhi:
Thank you for looking into this!

The implementation should ideally not care whether the batch size has been provided, but should return the correct batch size if one is provided. It's a bit tricky to ignore the last dimension, since for different future layers like for example 3D data, it might be that we have more dimensions than 4, so ignoring the last dimension needs to be threaded carefully through previous transforming layers.

darsnack:
>It's a bit tricky to ignore the last dimension, since for different future layers like for example 3D data, it might be that we have more dimensions than 4, so ignoring the last dimension needs to be threaded carefully through previous transforming layers.

My thoughts on implementing this would be a `_handle_batch(f, isize, ndims)` function where `f` is the standard output dimension calculation as a function of `isize` and `ndims` is the expected number of dimensions. For example, `Dense` might look like
```julia
function outdims(l::Dense, isize)
  calc_dims = isize -> begin
    first(isize) == size(l.W, 2) || throw(DimensionMismatch("input size should equal to ($(size(l.W, 2)), ...), got $isize"))
    return (size(l.W, 1), Base.tail(isize)...)
  end

  return _handle_batch(calc_dims, isize, 2)
end
```
`_handle_batch` would then know that `Dense` expects a 2D input (including batch) and accordingly pad `isize` with a 1 for the batch dimension and drop the batch (if necessary) after `calc_dims` returns.

DhairyaLGandhi:
Possibly something like this would work. This would just amount to a `ntuple`  call for `handle_batch`

darsnack:
Yeah though I guess `ndims` is unknown for generic functions. In this case, I think we can enforce that `isize` passed to generic functions includes the batch dimension. I don't foresee that being a pain point, since people should only really be relying on the generic case when it's nested inside a chain. We can modify the chain implementation to pad the batch at the start of the chain and drop at the end but maintain through the middle.

darsnack:
What conv-style layers can work on >3D arrays? I'll need to adjust the expected dimensions based on the `weight` array. Currently some layers are hardcoded to expect 4 dimensions (including batch).

darsnack:
@DhairyaLGandhi I think this is ready for review then merge. Summary of changes:
- Added graceful handling of batch dimension
- Added outdims for adaptive and global pooling
- Added outdims for SkipConnection
- Updated generic function outdims to allow functions with many arguments
- Updated Chain outdims to also work on vectors/tuples of layers
- Added more thorough testing
- Updated docs with example

darsnack:
Bump

DhairyaLGandhi:
what do you think can be done to remove the dependency on having methods defined for every new layer?

darsnack:
Referencing #1086, I don't think that this is avoidable for a certain subset of primitives. For example, I don't think we can avoid manually specifying this for `Conv`. There is of course the fallback method of evaluating the function on a sample input which exists in the PR.

We could mix these two versions of `outdims` to automatically generate the output size for higher-order layers like `SkipConnection`. Right now, it is manually specified how to evaluate the sub-layers within `SkipConnection`. But this is essentially similar to AD, in that we evaluate the trace until we hit a primitive rule. So we could have "check and evaluate" mode which replaces all the function calls in the AST with calls to `outdims`. Then when we evaluate the transformed AST, we will hit the primitive `outdims` definitions for stuff like `Conv` and the fallback definition for higher-order methods like `SkipConnection`. Similar to how Zygote effectively replaces the entire tree with calls to `pullback` in place of the normal function call.

Personally, I don't think such an effort is worth it right now. `outdims` needs a usability update to be functional, so we should merge this PR and focus on an automated `outdims` for another PR.

src/outdims.jl
```diff
@@ -0,0 +1,195 @@
...
+# Arguments:
+- `f`: a function of `isize` (including batch) that computes the output size
+- `isize`: the input size as specified by the user
+- `dimsize`: the expected number of dimensions for this layer (including batch)
+- `preserve_batch`: set to `true` to always retain the batch dimension
+"""
+function _handle_batch(f, isize, dimsize; preserve_batch = false)
+  indims = length(isize)
+  if indims == dimsize
+    return f(isize)
```
DhairyaLGandhi, line 15: Can we drop `f` from the method signature? It can easily check for `dimsize` and `preserve_batch`. That should make the further code simpler, since you don't need to add an additional closure here. I'm not clear what the benefit of having that is.

darsnack: Don't we need `f` to be able to call it for the "preserve batch" functionality?

DhairyaLGandhi: Couldn't that simply be checked after the fact? Also the throw should be an assert at the start.

darsnack: Yeah it can be. To avoid code duplication, it would involve two helper functions: `_handle_batch` and `_preserve_batch`. I'll push a commit that does this.

lorenzoh:
See [this Zulip thread](https://julialang.zulipchat.com/#narrow/stream/237432-ml-ecosystem-coordination/topic/Model.20output.20sizes/near/211098619). @darsnack said it could replace the default case for `outdims`.

darsnack:
@CarloLucibello I think this is ready to merge and should fix many issues related to `outdims`. I will submit a separate PR with @lorenzoh 's changes that will give us a more flexible implementation of `outdims`. The API will remain the same so I see no issue with multiple PRs.

CarloLucibello:
this need a rebase

src/outdims.jl
```diff
@@ -0,0 +1,225 @@
...
+"""
+    outdims(l::Conv, isize; preserve_batch = false)
+
+Calculate the output dimensions given the input dimensions `isize`.
+Set `preserve_batch` to `true` to always return with the batch dimension included.
+
+```julia
+m = Conv((3, 3), 3 => 16)
+outdims(m, (10, 10)) == (8, 8)
+outdims(m, (10, 10, 1, 3)) == (8, 8)
```
mcabbott, line 121:
```suggestion
```jldoctest
julia> c = Conv((3, 3), 3 => 16);

julia> Flux.outdims(c, (10, 10))
(8, 8)

julia> Flux.outdims(c, (10, 10, 3, 50))
(8, 8)

julia> Flux.outdims(c, (10, 10, 3, 50), preserve_batch=true) 
(8, 8, 16, 32)

julia> try Flux.outdims(c, (10, 10, 1, 50),) catch e println(e) end
DimensionMismatch("Input channels must match! (1 vs. 3)")
```

src/outdims.jl
```diff
@@ -0,0 +1,225 @@
...
+"""
+outdims(l::Dense, isize; preserve_batch = false)
+
+Calculate the output dimensions given the input dimensions, `isize`.
+Set `preserve_batch` to `true` to always return with the batch dimension included.
+
+```julia
+m = Dense(10, 5)
+outdims(m, (10,)) == (5,)
+outdims(m, (10, 2)) == (5, 2)
```
mcabbott, line 78:
```suggestion
```jldoctest
julia> d = Dense(10, 5);

julia> Flux.outdims(d, (10,))
(5,)

julia> Flux.outdims(d, (10, 32))
(5,)

julia> Flux.outdims(d, (10, 32); preserve_batch=true)
(5, 32)

julia> d(randn(Float32, 10, 32)) |> size
(5, 32)
```

mcabbott (commented):
I tried to write a clearer docstring, and added some doctests. 

Not quite sure what you intended to do in which PR, no real comment on the changes.

darsnack:
Thanks @mcabbott, I think the "spatial output" phrasing makes things more clear. There were some inconsistencies with the doctests expectation and what this PR actually does. So I added your changes w/ some modifications.

The intent of the PR is to refactor the `outdims` implementation to squash some long-standing bugs and make it more intuitive to use. Now the `isize` argument includes every dimension a user would pass if they were actually calling the model. The only exception is the batch dimension which can be optionally left off. The implementation will detect when the batch dimension is specified by the user and keep/drop it accordingly on the returned result.

@CarloLucibello If the rebase was the only change, then I think this is good to go once the CI passes (passed locally for me).

src/outdims.jl
```diff
@@ -0,0 +1,260 @@
...
+"""
+_handle_batchout(outsize, ispadded; preserve_batch = false) =
+  (ispadded && !preserve_batch) ? outsize[1:(end - 1)] : outsize
+
+# fallback for arbitrary functions/layers
+# ideally, users should only rely on this for flatten, etc. inside Chains
+"""
+    outdims(f, isize...)
+
+Calculates the output dimensions of `f(x)` where `size(x) == isize`.
```
CarloLucibello, line 37: this docstring is not consistent with what the method does, since `isize` ends up being a tuple of tuples. Maybe we want something like
"""
    outdims(f, isize1, isize2, ...)

Calculates the output dimensions of `f(x1, x2, ...)` where `size(x1) == isize1`, size(x2) == isize2`, ....
""

src/outdims.jl
```diff
@@ -0,0 +1,260 @@
...
+The batch dimension **must** be included.
+*Warning: this may be slow depending on `f`*
+"""
+outdims(f, isize...; preserve_batch = false) = size(f([ones(Float32, s) for s in isize]...))
```
CarloLucibello, line 41:
```suggestion
outdims(f, isize...; preserve_batch=false) = size(f([ones(Float32, s) for s in isize]...))
```

src/outdims.jl
CarloLucibello: to keep things simple, we could also just assume a single input, just as we assume a single output

darsnack: Yeah I am not sure the `...` is necessary.

src/outdims.jl
darsnack: Do you want me to remove the spacing around `=` for every kwarg in the changes?

CarloLucibello: yes please

src/outdims.jl
darsnack: Actually the vararg definition is required to handle the binary reduce operation in `SkipConnection`. I have updated the docstring instead to match the function.

src/outdims.jl
```diff
@@ -0,0 +1,264 @@
...
+julia> d = Dense(10, 5);
+
+julia> Flux.outdims(d, (10,))
+(5,)
+
+julia> Flux.outdims(d, (10, 32))
+(5, 32)
+
+julia> Flux.outdims(d, (10,); preserve_batch=true)
+(5, 1)
```
mcabbott, line 102: The function of this keyword seems more like it optionally *adds* a batch dimension, while the default is to preserve whatever the input does. Is that right? If so, `preserve_batch` seems like a confusing name. 

But might it be simpler just to delete this? Adding `,1)` by hand to your input might be easier & more consistent than learning about the keyword.

src/outdims.jl
```diff
@@ -0,0 +1,264 @@
...
+*Warning: this may be slow depending on `f`*
+"""
+outdims(f, isize...; preserve_batch=false) = size(f([ones(Float32, s) for s in isize]...))
+
+### start basic ###
+"""
+    outdims(c::Chain, isize; preserve_batch=false)
+    outdims(layers::Union{Tuple, AbstractVector}, isize; preserve_batch=false)
+
+Calculate the size of the spatial output dimensions, given the input size.
```
mcabbott, line 51: Apologies for slightly drive-by editing the other day. May I now vote against "spatial dimensions"? To me that implies just the first two (or as many as the dimensions of the conv. kernel) while now the function seems to handle all dimensions. Which is (I think) a better & simpler rule.

src/outdims.jl
darsnack: Yeah that's right. Currently, I use the `preserve_batch` keyword to detect whether `isize` contains the batch dimension or not for `Chain`s though. If you have a different solution, then I can safely remove this.

darsnack: One solution is to make a `expected_ndim(layer)` interface, but then that's two things the user needs to define to prevent their custom layer from hitting the fallback case.

mcabbott: But what goes wrong you don't do that? It looks like the code for Conv & Dense handle with or without, and return the same.

mcabbott: Maybe if your network won't run without the batch dim, then you shouldn't expect `outdims` to run it either?

darsnack: Something like that. The fallback requires the batch to work, and since a `Chain` could contain fallback cases, we need to evaluate the `Chain` with the batch included. But I can't know whether `isize` for a `Chain` contains the batch or not without some kind of indication from the first layer. Currently, I use the `preserve_batch` trick to get that indication.

src/outdims.jl
mcabbott: It is confusing to use `isize` to mean either the tuple `size(x)`, or the tuple `(size(x), size(y),)`, on different methods. Maybe something like  `outdims(f, size_x, [size_y, ...])` for the vararg case?

src/outdims.jl
darsnack: For example, `flatten` will not work if the input doesn't include the batch dimension.

mcabbott: OK. But how essential is it to allow as input a tuple of integers which lacks the batch dimension, when the network requires it? 

That seems like a bit of a do-what-I-mean feature, which is driving this complication.

darsnack: A typical model might be something like `m = Chain(Conv((3, 3), 3 => 16), flatten, Dense(1024, 10))`. Ideally, I would call `outdims(m, (224, 224, 3))` during the model building stage where `outdims` is most useful and batch size is unknown. This was the motivating case for handling the batch gracefully.

> That seems like a bit of a do-what-I-mean feature, which is driving this complication.

Yeah I agree with that. I guess just writing in the docstring that "include batch as 1 if unknown" is easier that supporting such a feature.

darsnack: So we could either drop handling the batch as something special or change the keyword to be something better than `preserve_batch`. I'm okay with either option. Any thoughts @DhairyaLGandhi or @CarloLucibello? If not, I will drop support for handling batch.

src/outdims.jl
darsnack: You're suggesting that change for the docstring right? And in the code I can do `isize` => `isizes` or something?

src/outdims.jl
mcabbott: Cool, if we aren't somehow bound to support missing batches, then `outdims(f, sz) = size(f(zeros(sz))) # but done faster` seems attractively simple & easy to remember.

src/outdims.jl
mcabbott: Yes here I was thinking of the docstrings. Although pluralising the code wouldn't hurt either but not a big deal.

src/outdims.jl
darsnack: Actually...if I can get the new implementation from @lorenzoh working, it will be fast for all cases (i.e. no fallback), handle the batch, and drop the need for `preserve_batch`. Though we can still drop batch handing. I'm going to try that now.

darsnack:
Okay so under a separate branch (`darsnack/better-outdims`), I have implement what Lorenz and I discussed in that Zulip thread. The basic idea is to create a `Nil` number type that voids all scalar operations. The advantage to this method is that every custom layer is automatically supported, since there is only one `outdims` definition that does `size(m(fill(nil, isize)))`. It is probably easier to maintain. The downside is that there are some issues with the warnings from NNlib for fallback convolution that requires a dependency to Logging.jl. It also includes the batch handling as a special case, and any custom layer can support batch handling by defining `dimhint(layer)` (this does not need to be defined for `outdims` to work, but you will need to include the batch w/ `isize`).

Here is a performance comparison:
```julia
julia> m = Chain(Conv((3, 3), 3 => 16), flatten, Dense(30*30*16, 10))
Chain(Conv((3, 3), 3=>16), flatten, Dense(14400, 10))

# outdims w/ current implementation
julia> @benchmark outdims($m, (32, 32, 3))
BenchmarkTools.Trial: 
  memory estimate:  59.88 KiB
  allocs estimate:  80
  --------------
  minimum time:     12.662 μs (0.00% GC)
  median time:      41.310 μs (0.00% GC)
  mean time:        53.773 μs (17.89% GC)
  maximum time:     8.441 ms (98.75% GC)
  --------------
  samples:          10000
  evals/sample:     1

# outdims w/ nil implementation
julia> @benchmark outdims($m, (32, 32, 3))
BenchmarkTools.Trial: 
  memory estimate:  3.41 KiB
  allocs estimate:  66
  --------------
  minimum time:     200.629 μs (0.00% GC)
  median time:      236.349 μs (0.00% GC)
  mean time:        258.500 μs (0.00% GC)
  maximum time:     7.229 ms (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

Using the `nil` implementation is slower, but it is still fast and much faster than actually evaluating the model.

mcabbott:
That's a neat idea. Any chance `missing` would work in place of `nil`? `@btime fill(missing, 1000,1000) * fill(missing, 1000);` is pretty quick, and `tanh(missing) === missing` while `tanh(nil)` would have to be written... sadly `fill(missing, 10,10) * fill(missing, 10,10)` is an error, but perhaps could be fixed?

darsnack:
Yeah looks like `Missing` is an option:
```julia
julia> @benchmark outdims($m, (32, 32, 3))
BenchmarkTools.Trial: 
  memory estimate:  3.41 KiB
  allocs estimate:  66
  --------------
  minimum time:     194.253 μs (0.00% GC)
  median time:      201.545 μs (0.00% GC)
  mean time:        210.417 μs (0.00% GC)
  maximum time:     505.264 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
Still need to check the test cases, but it looks like an option! @lorenzoh just wanted to double check that you didn't try `missing` and eliminate it for some reason.

lorenzoh:
The reason it doesn't use `missing` is that it is not a subtype of `Number` and some functions may restrict the input arguments to some `AbstractArray{<:Number,}`.

It also avoids type piracy for those edge cases where you have to overwrite a function.

mcabbott:
I guess that's a concern, Flux's own layers don't restrict, but others might. Are there examples? Transformers.jl doesn't seem to restrict. 

If you do go with `nil`, then perhaps it should be a special zero `<:Integer` perhaps better `<:AbstractFloat`, since `Real` would also be a plausible type restriction. 

It probably wouldn't be hard to get a method for `*` of matrices of missings added to Julia, to avoid piracy there.

darsnack:
After trying `Nil <: AbstractFloat`, I think it might be better to do `<: Number`. There are a lot of primate number related functions (e.g. `*(::Bool, ::AbstractFloat)` and `<(::AbstractFloat, AbstractFloat)`) that are specialized for floats. In contrast, most primitive functions on `Number` fallback to a smaller set of Base functions. There are also layers in Flux that specialize on `AbstractFloat` inputs to hit BLAS.

It might be easier to override a method early specifically for `Nil` inputs when the default signature has `AbstractFloat` restrictions. Currently, there are lots of little cases that need to be overridden for `AbstractFloat`, and I don't think the tests based on Flux layers will capture all the little cases users will encounter.

darsnack:
Any last comments on the `nil` approach? If not, I am going to update `darsnack#master` to use it.

mcabbott:
No strong feelings, using `missing` still seems attractively simple. 

I have no examples of people bounding things to `Number` (hence excluding missing), but no reason to think they wouldn't then bound to `Real` (shorter to type), even if you want to skip `AbstractFloat` for reasons described. Would also be tempted to make things like `nil < x  =  x>=0` i.e. to treat `nil` as a singleton zero, rather than another `missing`, although not because I've thought hard about it.

darsnack:
The only thing is getting `missing` to work requires defining `Base.::Type{T}(::Missing) where T<:Number = missing`. That seems like it could have unintended (and bad) side effects for users.

CarloLucibello:
> The only thing is getting missing to work requires defining Base.::Type{T}(::Missing) where T<:Number = missing. That seems like it could have unintended (and bad) side effects for users.

this is definitely a no-go

Is the `nil` approach compatible with models on gpu?

mcabbott:
> Base.::Type{T}(::Missing) where T<:Number = missing

I agree you can't define that! But where does this come up?

CarloLucibello:
And how does the speed compare to plain forward evaluation? Maybe we don't want to overengineer this, AFAIK you may use `outdims` when designing new models but normally you would get rid of this when done.

Another observation is about layers such as batchnorm accumulating statistics, we don't want to mess that. Probably we'll have to special case those

darsnack:
> And how does the speed compare to plain forward evaluation? Maybe we don't want to overengineer this, AFAIK you may use outdims when designing new models but normally you would get rid of it when done.

Based on the same discussion as [the previous PR](https://github.com/FluxML/Flux.jl/pull/1253), we want to avoid plain forward evaluation. I can double-check the speed, but I'm think the nil approach will still be noticeably faster.

> Another observation is about layers such as batchnorm accumulating statistics, we don't want to mess that

Good point. Maybe we can place them in test mode then restore them to the original mode?

> But where does this come up?

During normalization when the statistic update is recast to the
underlying type off the statistic.

Switching to test mode will circumvent the issue.

mcabbott:
Ah OK. I guess for layers defined here, it's easy to add missing methods if necessary.

darsnack:
I would say the original approach in this PR was over-engineered, whereas the `nil` approach just seems clever￼.

darsnack:
How do we feel about `Missing(::T) where T<:Number = missing`? It is needed for the pooling function in NNlib. Sorry I missed this in my previous comment.

darsnack:
Performance by just passing stuff through:
```julia
julia> @benchmark size($m(ones(32, 32, 3, 1)))
┌ Warning: Slow fallback implementation invoked for conv!  You probably don't want this; check your datatypes.
│   yT = Float64
│   T1 = Float64
│   T2 = Float32
└ @ NNlib ~/.julia/packages/NNlib/Z9qZP/src/conv.jl:206
BenchmarkTools.Trial: 
  memory estimate:  307.36 KiB
  allocs estimate:  45
  --------------
  minimum time:     527.285 μs (0.00% GC)
  median time:      657.550 μs (0.00% GC)
  mean time:        702.721 μs (4.87% GC)
  maximum time:     6.246 ms (86.14% GC)
  --------------
  samples:          7105
  evals/sample:     1
```

I want to do a test with bigger models, because I don't think the standard forward pass will scale as well. Also need to look into the GPU. I'll do that later tonight or tomorrow.

lorenzoh:
> Another observation is about layers such as batchnorm accumulating statistics, we don't want to mess that. Probably we'll have to special case those

I though the batch norm updates only happen inside a `gradient` call?

I would in any case suggest `nil` over `Missing`, as it is simple enough, and there are fewer hidden downsides from possible type piracy. On the user-side, people will probably use `outdims` anyway.

darsnack:
> I though the batch norm updates only happen inside a gradient call?

Yeah, I'll have to investigate this. It's possible to update the statistics without a gradient call by using `trainmode!(model)`, but that shouldn't be the default. Either way, we should make `outdims` work in all modes.

DhairyaLGandhi:
I'm a bit uncomfortable with the code bloat for something of an orthogonal utility, something that doesn't need a special type would be nicer maybe?

CarloLucibello:
yes guys, this seems largely overengineered for something that is just used a few times in the model building phase. I would just replace the whole outdims story with
```julia
outdims(f, xsize) = f(rand(xsize)) |> size
```
and never worry about outdims again.

darsnack:
It may only be used during model building, but it is a function almost everyone who build models programmatically writes. That was the original reason it was added to Flux — it came up several times on Slack, and people kept posting a code snippet of how to do it. If you only build one model, then I agree that it seems like a lot for a small thing, but if you evaluate algorithms on many models, then it is a crucial utility.

Here's why just doing `outdims(f, isize) = f(rand(isize)) |> size` won't work:
```julia
julia> m = densenet_121();

julia> length(m)
74

julia> @benchmark outdims($m, (256, 256, 3))
BenchmarkTools.Trial:
  memory estimate:  408.06 KiB
  allocs estimate:  8313
  --------------
  minimum time:     9.947 ms (0.00% GC)
  median time:      10.991 ms (0.00% GC)
  mean time:        11.132 ms (0.56% GC)
  maximum time:     22.058 ms (41.70% GC)
  --------------
  samples:          450
  evals/sample:     1

julia> @benchmark size($m(rand(256, 256, 3, 1)))
┌ Warning: Slow fallback implementation invoked for conv!  You probably don't want this; check your datatypes.
│   yT = Float64
│   T1 = Float64
│   T2 = Float32
└ @ NNlib ~/.julia/packages/NNlib/2Wxlq/src/conv.jl:206
BenchmarkTools.Trial:
  memory estimate:  556.94 MiB
  allocs estimate:  11703
  --------------
  minimum time:     7.605 s (0.54% GC)
  median time:      7.605 s (0.54% GC)
  mean time:        7.605 s (0.54% GC)
  maximum time:     7.605 s (0.54% GC)
  --------------
  samples:          1
  evals/sample:     1

julia> @benchmark size($m($(rand(256, 256, 3, 1))))
BenchmarkTools.Trial:
  memory estimate:  555.44 MiB
  allocs estimate:  11700
  --------------
  minimum time:     7.588 s (0.52% GC)
  median time:      7.588 s (0.52% GC)
  mean time:        7.588 s (0.52% GC)
  maximum time:     7.588 s (0.52% GC)
  --------------
  samples:          1
  evals/sample:     1
```
With real models, evaluating the model (or even part of it) is too slow.

FYI, there are no issues with updating normalization statistics anymore @CarloLucibello (no convoluted solution; I just bypass evaluating normalization layers).

@DhairyaLGandhi I think the point about code bloat is well taken. That's why I chose to put everything in a separate file, so it doesn't pollute the rest of the codebase in terms of readability. I also agree that having this new `Nil` type introduces a lot of code, but I would point out that it is mostly standard Base definitions for a new number type. In other words, we should need to expand or maintain the code too much, since the basic number type operations don't really change. *I would also point out that the extra code for `Nil` is worth it*, because it completely eliminates maintenance of `outdims` itself by reducing the definition to a single method. It also makes the behavior of `outdims` consistent between arbitrary functions and standard layers. In all, I think this approach is more intuitive for the user and lower maintenance overhead.

From an API perspective, there is no difference between the `nil` and non-`nil` versions, so we can always revert back without making a breaking change.

darsnack:
In the latest commit, I removed the heavy dependency to Logging. So at this point, I no longer have any gripes with the `Nil` approach.

It's worth pointing out that ~50% of the lines in this implementation are actually docstrings and examples.
Screen Shot 2020-12-01 at 8 49 57 AM

DhairyaLGandhi:
Shifting the burden of maintainence isn't unfortunately the same as alleviating it. I appreciate the outdims utility for what it is, and I'd like to see cases of this special type offering benefits over a fallback true forward pass. Maintaining a special type for this seems overkill otherwise.

As an aside, I like the functionality of a handy `dimhint`, but I'd expect it to return a tuple with acceptable sizes.

darsnack:
Here is an example use-case that I do quite often, because I mix-and-match models and datasets. So my input width and height change across each evaluation. Here's an actual function from a project (I replaced my custom size calculation function with `outdims` for clarity):
```julia
function complexcnn(width, height, ichannels, nclasses; lsizes, ksize = (3, 3), ssize = 1)    
    convs = []
    for i in 2:length(lsizes)
        if (i - 1) % 2 == 0
            push!(convs, Conv(ksize, lsizes[i - 1] => lsizes[i], stride = ssize, pad = 1))
            push!(convs, BatchNorm(lsizes[i], relu))
            push!(convs, MaxPool((2, 2)))
        else
            push!(convs, Conv(ksize, lsizes[i - 1] => lsizes[i], relu, stride = ssize, pad = 1))
        end
    end
    
    ffwidth, ffheight, _ = outdims(convs, (width, height, ichannels))
    ffsize1 = ffwidth * ffheight * lsizes[end]
    ffsize2 = (ffsize1 / 2 < nclasses) ? ffsize1 : Int(ffsize1 / 2)
    Chain(
        convs...,
        x -> reshape(x, :, size(x, 4)),
        Dense(ffsize1, ffsize2, relu),
        Dense(ffsize2, nclasses)
    )
end
```

When `length(lsizes)` gets large (e.g. ~50), then just evaluating a `rand` input would cost the same as the DenseNet example I gave above.

> As an aside, I like the functionality of a handy dimhint, but I'd expect it to return a tuple with acceptable sizes.

Sorry I am a little confused. `dimhint(m)` is supposed to return the output of the number of dimensions, so it should be a scalar quantity?

src/outdims.jl
```diff
@@ -0,0 +1,219 @@
...
+dimhint(::MaxPool) = 4
+dimhint(::MeanPool) = 4
+dimhint(::AdaptiveMaxPool) = 4
+dimhint(::AdaptiveMeanPool) = 4
+dimhint(::GlobalMaxPool) = 4
+dimhint(::GlobalMeanPool) = 4
+
+## make tuples and vectors be like Chains
+
+(m::Tuple)(x::AbstractArray{Nil}) = applychain(m, x)
```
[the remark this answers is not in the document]
darsnack: Sounds good

darsnack:
> Shifting the burden of maintainence isn't unfortunately the same as alleviating it

Yeah that's fair. FWIW, I think this does reduce the maintenance burden. In the non-`nil` version, there is a separate `outdims` for each layer, and it would need to be checked/updated every time changes are made to a layer. If a new layer is added, a new `outdims` needs to be defined. There is also the possibility of implementation errors that are missed by testing (like my original mistake with `foldr` and `Chain`). On the other hand, there might be some initial functions that need to be defined for `Nil` that we missed, but after the first release, that should stabilize. I'd say it's quite unlikely that we'll need to modify or add to any of the `Nil` functions. And as layer implementations change or new layers are added, the `nil` approach will work without any maintenance. I also think it is easier for a user to extend with `dimhint`.

darsnack:
Any more thoughts on this?

mcabbott:
I still wonder if all the dimension guessing couldn't be dropped. It seems hard to know whether this is going to work or not, and it means you can't rely on `outdims` as a test of whether the model actually accepts the given dimensions. Wouldn't it be simpler if the promise was just `outdims(m, sz) == size(m(rand(sz)))`, else error, nothing more? 

If handling size tuples which don't include a batch dimension which the model does require is necessary, perhaps this could  be something you explicitly ask for? Like `outdims(m, sz; add=1) == size(m(rand(sz..., 1)))[1:end-1]`.

CarloLucibello:
Is relu supported? probably that requires defining a  `max` method

darsnack:
Yes I agree. At this point that seems more intuitive.

darsnack:
Will double check relu.

src/outdims.jl
```diff
@@ -0,0 +1,219 @@
...
+# TODO: can this be shortened?
+Base.promote(x::Nil, y::Nil) = (nil, nil)
+Base.promote(x::Nil, y) = (nil, nil)
+Base.promote(x, y::Nil) = (nil, nil)
+Base.promote(x::Nil, y, z) = (nil, nil, nil)
+Base.promote(x, y::Nil, z) = (nil, nil, nil)
+Base.promote(x, y, z::Nil) = (nil, nil, nil)
+Base.promote(x::Nil, y, z::Nil) = (nil, nil, nil)
+Base.promote(x::Nil, y::Nil, z::Nil) = (nil, nil, nil)
+Base.promote(x::Nil, y::Nil, z) = (nil, nil, nil)
```
gxyd, line 69: We probably have `Base.promote(x, y::Nil, z::Nil) = (nil, nil, nil)` missing?

mcabbott: I haven't tried, but isn't the usual method to overload `promote_rule`? Perhaps one line could cover all cases?

lorenzoh: Wasn't aware of that so we probably can 👍

src/outdims.jl
```diff
@@ -0,0 +1,219 @@
...
+Base.:-(::Nil, ::Number) = nil
+Base.:-(::Number, ::Nil) = nil
+
+Base.:*(::Nil, ::Nil) = nil
+Base.:*(::Nil, ::Number) = nil
+Base.:*(::Number, ::Nil) = nil
+
+Base.:/(::Nil, ::Nil) = nil
+Base.:/(::Nil, ::Number) = nil
+Base.:/(::Number, ::Nil) = nil
```
mcabbott, line 43: With `Base.promote_rule(::Type{Nil}, ::Type{<:Number}) = Nil`, I think only the same-type version is needed:
```suggestion
for f in [:+, :-, :*, :/]
    @eval Base.$f(::Nil, ::Nil) = nil
end
```

src/outdims.jl
```diff
@@ -0,0 +1,219 @@
...
+
+Base.float(::Type{Nil}) = Nil
+Base.copy(::Nil) = nil
+Base.abs2(::Nil) = nil
+Base.sqrt(::Nil) = nil
+Base.zero(::Type{Nil}) = nil
+Base.one(::Type{Nil}) = nil
+
+Base.:+(::Nil) = nil
+Base.:-(::Nil) = nil
```
mcabbott, line 27:
```suggestion
for f in [copy, zero, one, oneunit, :+, :-, :abs, :abs2, :inv, :exp, :log]
    @eval Base.$f(::Nil) = nil
```

src/outdims.jl
```diff
@@ -0,0 +1,219 @@
...
+
+Nil is a singleton type with a single instance `nil`.
+Unlike `Nothing` and `Missing` it subtypes `Number`.
+"""
+struct Nil <: Number end
+
+const nil = Nil()
+
+Nil(::T) where T<:Number = nil
+(::Type{T})(::Nil) where T<:Number = nil
```
mcabbott, line 17: Should these be `Base.convert(::Type{Nil}, ::Number) = nil`? I'm not certain what the standard thing to do is.

darsnack: I think `convert` is separate from invoking a type's constructor based on my reading of the docs. I added these in to deal specifically with layers (e.g. `BatchNorm`) where something like `layer.state = Float32.(x)` is called. Probably these calls should be using `convert` and not the constructor.

We technically don't need this anymore, since I bypass normalization layers to avoid affecting the statistics. But I figured I'd keep it there in case someone does the same thing in a custom layer.

src/outdims.jl
```diff
@@ -0,0 +1,208 @@
...
+Base.promote(x::Nil, y, z) = (nil, nil, nil)
+Base.promote(x, y::Nil, z) = (nil, nil, nil)
+Base.promote(x, y, z::Nil) = (nil, nil, nil)
+Base.promote(x::Nil, y, z::Nil) = (nil, nil, nil)
+Base.promote(x::Nil, y::Nil, z::Nil) = (nil, nil, nil)
+Base.promote(x::Nil, y::Nil, z) = (nil, nil, nil)
+
+
+LinearAlgebra.adjoint(::Nil) = nil
+LinearAlgebra.transpose(::Nil) = nil
```
mcabbott, line 62: I think these can both be replaced by just overloading `Base.conj` in the unary list.

darsnack:
Okay I think the latest commit resolves all the comments. I also added test cases for every activation function too.

gxyd:
The following doesn't seem to fully I guess:
```jl
julia> using Flux

julia> outdims(Dense(10, 4), (10,))

julia> m = Chain(
          Conv((3, 3, 3), 3 => 16),
          Conv((3, 3, 3), 16 => 32),
          Conv((3, 3, 3), 32 => 64),
          flatten,
          Dense(4*4*4*64, 10)
      )

julia> input = randn(Float32, 10, 10, 10, 3, 64);

julia> m(input) |> size
(10, 64)

julia> outdims(m, input)
MethodError: no method matching fill(::Flux.NilNumber.Nil, ::Tuple{Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,Float32,....

```

I'll look around why that might be happening.

src/outdims.jl
gxyd: Sorry, I probably should've read the documentation first I guess. Though `promote_rule` does look good.

darsnack:
Yeah I can specialize `isize` on tuples so it throws a `MethodError` for `outdims` and not something downstream.

gxyd:
> Yeah I can specialize isize on tuples so it throws a MethodError for outdims and not something downstream.

For someone like me who is much of a beginner, I think that would be convenient.

Also since we're looking at `outdims`, I was thinking if we can think about https://github.com/FluxML/Flux.jl/issues/1266 in this PR itself as well, WDYT?

darsnack:
Personally, I think `indims` or a single argument `outdims` will be confusing, since they can only be defined for a subset of layers. The batch handling was defined well on all Flux layers, and even with our attempts to minimize the complexity, we found it was nicer to just have consistent behavior across every layer. I would say the same logic applies to #1266.

On the implementation side of things, #1266 came at a time when `outdims` was specialized by the type of the layer. One of the improvements in this PR is to remove that specialization, so I don't think re-introducing it is nice for maintainability.

If we still want #1266, then I'd be happy to review/write a PR, but I'd prefer it be different from this PR. I kind of think we should close this one out soon, so that model improvements for #1431 can rely on a stable, working `outdims` API.

docs/src/utilities.md
```diff
@@ -35,6 +35,34 @@ Flux.glorot_uniform
...
+the model that preserves size information without running any computation.
+`outdims(f, isize)` works for all layers (including custom layers) out of the box.
+By default, `isize` excludes the batch dimension (assuming it is one),
+but you can set a specific batch size with `outdims(f, isize; padbatch = false)`.
+
+Using this utility function lets you automate model building for various inputs like so:
+```julia
+function make_model(width, height, nchannels, nclasses)
+  # returns 1D array (vector) of conv layers
+  conv_layers = make_conv(width, height, nchannels)
```
CarloLucibello, line 53: Since `make_conv` is not explicitly defined maybe we should avoid using it. It could mislead people into thinking it is a part of Flux

docs/src/utilities.md
```diff
@@ -35,6 +35,34 @@ Flux.glorot_uniform
...
+  conv_outsize = outdims(conv_layers, (width, height, nchannels))
```
CarloLucibello, line 54: Is conv_layers supposed to be a chain? this may be a bit confusing because conv_layers is then splatted into a Chain. Generally I think this whole example should be more clear

src/Flux.jl
```diff
@@ -15,7 +15,8 @@ export Chain, Dense, Maxout, RNN, LSTM, GRU, SamePad, Conv, CrossCor, ConvTransp
        AdaptiveMaxPool, AdaptiveMeanPool, GlobalMaxPool, GlobalMeanPool, MaxPool,
        MeanPool, flatten, DepthwiseConv, Dropout, AlphaDropout, LayerNorm, BatchNorm,
        InstanceNorm, GroupNorm, SkipConnection, params, fmap, cpu, gpu, f32, f64,
-       testmode!, trainmode!
+       testmode!, trainmode!,
+       outdims
```
CarloLucibello, line 19: do we want to export it? seems a pretty common name, maybe it is better to encourage qualified usage

src/layers/conv.jl
```diff
@@ -189,22 +187,10 @@ end
...
-"""
-outdims(l::Conv, isize) =
-  output_size(DenseConvDims(_paddims(isize, size(l.weight)), size(l.weight); stride = l.stride, padding = l.pad, dilation = l.dilation))
-
-"""
-    ConvTranspose(filter, in => out, σ=identity; stride=1, pad=0, dilation=1)
+    ConvTranspose(filter, in=>out)
+    ConvTranspose(filter, in=>out, activation)
+    ConvTranspose(filter, in => out, σ = identity; init = glorot_uniform,
+                  stride = 1, pad = 0, dilation = 1)
```
CarloLucibello, line 193: unrelated change

CarloLucibello: this is issue #1422 , better to remove the change here and solve it separately

CarloLucibello (changes requested):

CarloLucibello:
@darsnack so you think `padbatch=true` is the most convenient default? Alternatives are defaulting to `padbatch=true` or always removing the `padbatch` option and always assume batch dimension is present

CarloLucibello:
Consider also that when `padbatch=true` we cannot apply composition as in 
```julia
outdims(m2, outdims(m1, sz))
```

darsnack:
> so you think padbatch=true is the most convenient default? Alternatives are defaulting to padbatch=true or removing the padbatch option and always assume batch dimension is present

I have no strong feelings for `padbatch=true`/`padbatch=false`/no `padbatch`. I chose `true` because in all the contexts that I use `outdims`, the batch dimension is not yet set. But I think the composition case is compelling enough to make `false` the default. From a user perspective, I think `padbatch=true` is cleaner to write than `(isize..., 1)`, so I do think a keyword arg is a nice feature.

CarloLucibello:
ok, let's change the default to `false` then

mcabbott:
One vote for `padbatch=false`, seems easier to explain.

Should this be called `outdims`, BTW? Normally `dims` means the keyword in Julia now, and goes with `ndims`, while `size` is the concept this deals with. Perhaps it should be renamed to `mapsize` or `sizemap` while we're fixing things?

darsnack:
Hmm or maybe `outsize` or `sizeout`? "Map" is a little confusing to me because I don't know what's being mapped immediately.

mcabbott:
Yes indeed `map` isn't right, `outdims(f, (1,2,3))` applies `f` to (something like) the 2nd arg, not to each element. But `outsize` or `outputsize` seems great. It can have a depwarn and then I think it shouldn't break anything.

darsnack:
I'll probably go with `outputsize` for clarity.

CarloLucibello:
you can add a deprecation to `deprecations.jl`. Next release is going to contain a few deprecations and breaking changes, so this is a good time for deprecating more stuff

gxyd:
`outputsize` sounds good to me. May be change `isize` -> `inputsize` as well? (explicit better than implicit)
I hope I'm not commenting out of turn here.

darsnack:
@CarloLucibello I think all comments are addressed now. Tried to make the example more clear.

src/outputsize.jl
```diff
@@ -0,0 +1,124 @@
...
+using .NilNumber: Nil, nil
+
+"""
+    outputsize(m, inputsize::Tuple; padbatch=false)
+
+Calculate the output size of model/function `m` given an input of size `inputsize` (w/o computing results).
+`inputsize` should include all dimensions (except the batch dimension can be excluded when `padbatch == true`).
+If `m` is a `Tuple` or `Vector`, `outputsize` treats `m` like a `Chain`.
+
+*Note*: this method should work out of the box for custom layers.
```
mcabbott, line 54:
```suggestion
Calculate the output size of model `m` given the input size. 
Obeys `outputsize(m, size(x)) == size(m(x))` for valid input `x`.
Keyword `padbatch=true` is equivalent to using `(inputsize..., 1)`, and 
returns the final size including this extra batch dimension.

This should be faster than calling `m(x)`. It uses a trivial number type, 
and thus should work out of the box for custom layers.

If `m` is a `Tuple` or `Vector`, its elements are applied in sequence, like `Chain(m...)`.
```

src/deprecations.jl
```diff
@@ -3,3 +3,4 @@
 @deprecate InstanceNorm(λ, β, γ, μ, σ², ϵ, momentum) InstanceNorm(λ, β, γ, μ, σ², ϵ, momentum, nothing)
 @deprecate BatchNorm(λ, β, γ, μ, σ², ϵ, momentum) BatchNorm(λ, β, γ, μ, σ², ϵ, momentum, nothing)
 @deprecate GroupNorm(G, λ, β, γ, μ, σ², ϵ, momentum) GroupNorm(G, λ, β, γ, μ, σ², ϵ, momentum, nothing)
+@deprecate outdims(f, inputsize; padbatch) outputsize(f, inputsize; padbatch=padbatch)
```
gxyd, line 6: Doing something like:
```jl
julia> layer = Dense(10, 100)

julia> outdims(layer, (10,))
UndefKeywordError: keyword argument padbatch not assigned

Stacktrace:
 [1] outdims(::Dense{typeof(identity),Array{Float32,2},Array{Float32,1}}, ::Tuple{Int64}) at ./deprecated.jl:71
 [2] top-level scope at In[4]:1
 [3] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091
```

Do we need make it `outdims(f, inputsize) outputsize(f, inputsize; padbatch=false)`?

src/deprecations.jl
```diff
@@ -3,3 +3,4 @@
 @deprecate InstanceNorm(λ, β, γ, μ, σ², ϵ, momentum) InstanceNorm(λ, β, γ, μ, σ², ϵ, momentum, nothing)
 @deprecate BatchNorm(λ, β, γ, μ, σ², ϵ, momentum) BatchNorm(λ, β, γ, μ, σ², ϵ, momentum, nothing)
 @deprecate GroupNorm(G, λ, β, γ, μ, σ², ϵ, momentum) GroupNorm(G, λ, β, γ, μ, σ², ϵ, momentum, nothing)
+@deprecate outdims(f, inputsize; padbatch=false) outputsize(f, inputsize; padbatch=padbatch)
```
CarloLucibello, line 6:
```suggestion
@deprecate outdims(f, inputsize) outputsize(f, inputsize)
```
since `padbatch` was not defined before this PR.  The deprecation is not 100% correct, since it is difficult to replicate the behavior of the old `outdims` with the new version, but this is good enough

CarloLucibello:
I'll wait for green CI then merge, thanks @darsnack for the patience of seeing this through

gxyd:
Since that'll be a deprecation of `outdims`, shouldn't depreciation raise warning (when deprecated function is used for the first time) for the user to move to the use of corresponding function?

darsnack:
> Since that'll be a deprecation of outdims, shouldn't depreciation raise warning (when deprecated function is used for the first time) for the user to move to the use of corresponding function?

I think the issue is that we can't replicate the old batch handling behavior.

Sorry there appear to be some remaining test failures. Fixed them and testing locally before pushing so we get green CI 🤞🏾

darsnack:
Okay all tests passed locally, so hopefully we will be GTM

CarloLucibello (approved):

CarloLucibello:
@DhairyaLGandhi @maleadt is bors down or just busy?

bors r+

DhairyaLGandhi:
I haven't followed this closely in the past week, but seems that the new type was merged recently? I am uncomfortable about the extra burden this puts on code bloat still, so I'd definitely expect that to be addressed

src/outputsize.jl
```diff
@@ -0,0 +1,128 @@
...
+Nil(::T) where T<:Number = nil
+(::Type{T})(::Nil) where T<:Number = nil
+Base.convert(::Type{Nil}, ::Number) = nil
+
+Base.float(::Type{Nil}) = Nil
+
+for f in [:copy, :zero, :one, :oneunit,
+          :+, :-, :abs, :abs2, :inv,
+          :exp, :log, :log1p, :log2, :log10,
+          :sqrt, :tanh, :conj]
```
DhairyaLGandhi, line 24: What happens to new layers and new functions that layers would need. That would need a wider catchall than adding it to a list like here.

darsnack: New layers don't any modifications to work. New functions might, but most functions are built using primitives such as the ones in this list. There might be some Base primitive functions we need to add (e.g. `mod`), but most functions shouldn't need changes.

darsnack: I guess what do you mean by wider catch-all? If it's the ability to adapt to functions that aren't defined here, then I think we would need to resort to meta-programming for that.

DhairyaLGandhi: Yeah I mean the functions and the correct answer would be to say operations on numbers would need to be forwarded with a macro call

darsnack: To me, that kind of dispatch sounds like Cassette.jl.

PR #1314, opened 2020-08-13
make bias optional in dense layer

AliaaRassem:
This is to make the bias in dense layer as optional parameters as in the convolution networks by using Flux.Zeros()

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -92,6 +92,8 @@ Create a traditional `Dense` layer with parameters `W` and `b`.
 The input `x` must be a vector of length `in`, or a batch of vectors represented
 as an `in × N` matrix. The out `y` will be a vector or batch of length `out`.
 
+Setting `initb` to `Flux.Zeros()` would switch `bias` off for the layer.
+
 # Example
 ```
 julia> d = Dense(5, 2)
@@ -103,7 +105,8 @@ julia> d(rand(5))
   0.123119034
 ```
 """
-struct Dense{F,S<:AbstractArray,T<:AbstractArray}
+
+struct Dense{F,S<:AbstractArray,T<: Union{Zeros, AbstractArray}}
   W::S
   b::T
   σ::F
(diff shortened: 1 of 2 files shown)


DhairyaLGandhi:
Since `Flux.Zeros <: AbstractArray`, we don't need to add the union. Rather maybe just add a constructor that takes a weight and bias kwarg and let the rest of the machinery take care of the rest of the computation?

AliaaRassem:
changed bias type back to AbstractArray

src/layers/basic.jl
```diff
@@ -112,15 +115,15 @@ end
...
+               initW = glorot_uniform, initb = zeros(out))
+  return Dense(initW(out, in), initb, σ)
 end
 
 @functor Dense
 
 function (a::Dense)(x::AbstractArray)
   W, b, σ = a.W, a.b, a.σ
-  σ.(W*x .+ b)
+  typeof(b) <: Zeros ? σ.(W*x ) : σ.(W*x .+ b)
```
DhairyaLGandhi, line 126: I would prefer not to change the forward pass here, Zeros would be handled appropriately

src/layers/basic.jl
```diff
@@ -112,15 +115,15 @@ end
 Dense(W, b) = Dense(W, b, identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
-               initW = glorot_uniform, initb = zeros)
-  return Dense(initW(out, in), initb(out), σ)
+               initW = glorot_uniform, initb = zeros(out))
```
DhairyaLGandhi, line 118: It's odd that here we have a function for weight and a vector for bias.

The more appropriate constructor would accept both weight and bias as kwargs ( refer the conv layers)

DhairyaLGandhi (commented):
Thanks! I've left a couple thoughts up, which should make the API easier to use.

DhairyaLGandhi:
bump

mcabbott:
Replaced by #1379 I think. 

Perhaps we should still kill the `initW` and `initb` keywords, as suggested by Dhairya here: https://github.com/FluxML/Flux.jl/pull/1314#discussion_r482692743.

PR #1315, opened 2020-08-15
added nll, softplus, logcosh loss

SubhadityaMukherjee:
I saw Pytorch/TF having some extra loss functions. Thought I would try to add as many as possible to Flux. (This PR has just 3 of them).
They are direct conversions from Python -> Julia. I added the source function in existing libraries as well for comparison. 

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -22,10 +22,10 @@ nll(y; agg=sum) = -agg(log.(y))
 
 Return the output of the softplus function with β being a parameter of the softplus
     
-    (1/β).*(log.(exp.(β .* x).+1))
+    (1/β).*(log.(exp.(β .* y).+1))
 """
 
-softplus(y,β=1) = (1/β).*(log.(exp.(β .* x).+1))
+softplus(y,β=1) = (1/β).*(log.(exp.(β .* y).+1))
 
 """
     logcosh(y,ŷ) [Tensorflow 2.0 source](https://www.tensorflow.org/api_docs/python/tf/keras/losses/log_cosh)
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -8,12 +8,47 @@ Return the loss corresponding to mean absolute error:
 mae(ŷ, y; agg=mean) = agg(abs.(ŷ .- y))
 
 """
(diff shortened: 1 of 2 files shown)


DhairyaLGandhi:
Thanks for this! Have you seen `softplus` and `logcosh` [here](https://github.com/FluxML/NNlib.jl/blob/d524be7dc1a44c96793c0e9b420469204a85e9e7/src/activation.jl#L1)

DhairyaLGandhi:
It is also already available to Flux after the re-export of NNlib, so it should be in the correct user namespace.

SubhadityaMukherjee:
:)
Ohh. I did a search on this repo and didn’t find it. Haha okay next time I’ll look at the related libraries too. 
On the plus side. It seems like nll wasn’t there. 😄 

Just as a side. If I find something in pytorch/tf that flux doesn’t have yet(IF). I can go ahead and drop a PR if I can make it work right? I’m not sure about the policy here so just confirming. 

Thanks for pointing it out though!

DhairyaLGandhi:
Yes that would be fine. Also, I forget if `nll` was added at some point, worth a quick check.

src/losses/functions.jl
```diff
@@ -7,13 +7,48 @@ Return the loss corresponding to mean absolute error:
...
+
+"""
+    softplus(y,β=1) [Pytorch source](https://pytorch.org/docs/stable/generated/torch.nn.Softplus.html#torch.nn.Softplus)
+
+Return the output of the softplus function with β being a parameter of the softplus
+    
+    (1/β).*(log.(exp.(β .* x).+1))
+"""
+
+softplus(y,β=1) = (1/β).*(log.(exp.(β .* x).+1))
```
DoktorMike, line 28: This function defines an input y but refers to an unknown variable x in the exp.

SubhadityaMukherjee: Oh no.. my bad. I fixed it.

CarloLucibello:
this has been stale for a while and I don't think it's worth having those functions, they are rarely used and simple to implement if needed. Closing this as I'm going through some cleanup of the repo

Issue #1319, opened 2020-08-21
ConvTranspose same padding and outdims errors


holongate:
Hi,

I tried to (manually) convert a Keras Sequence model that uses a ConvTranspose layer but get into some errors:

#sample 
julia> Using FLux
julia> t = ones(Float32, **(32,32,32,1,1)**);
julia> size(ConvTranspose((4,4,4), 1=>1, pad=**SamePad()**, stride=**2**)(t))
julia> **(63, 63, 63, 1, 1)**

Shouldn't this be **(64,64,64,1,1)** ?

Also, outdims raises an error:

julia> Flux.outdims(ConvTranspose((3,3,3), 1=>1, pad=SamePad(), stride=2), (32,32,32,1,1))

ERROR: DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 2 and 3")
Stacktrace:
 [1] _bcs1 at ./broadcast.jl:501 [inlined]
 [2] _bcs at ./broadcast.jl:495 [inlined]
 [3] broadcast_shape(::Tuple{Base.OneTo{Int64}}, ::Tuple{Base.OneTo{Int64}}) at ./broadcast.jl:489
 [4] combine_axes at ./broadcast.jl:484 [inlined]
 [5] _axes at ./broadcast.jl:209 [inlined]
 [6] axes at ./broadcast.jl:207 [inlined]
 [7] combine_axes at ./broadcast.jl:484 [inlined]
 [8] _axes at ./broadcast.jl:209 [inlined]
 [9] axes at ./broadcast.jl:207 [inlined]
 [10] combine_axes at ./broadcast.jl:484 [inlined]
 [11] _axes at ./broadcast.jl:209 [inlined]
 [12] axes at ./broadcast.jl:207 [inlined]
 [13] combine_axes at ./broadcast.jl:484 [inlined]
 [14] _axes at ./broadcast.jl:209 [inlined]
 [15] axes at ./broadcast.jl:207 [inlined]
[...]

PR #1320, opened 2020-08-22
Change ConvTranspose with SamePad to have outsize = stride * insize

DrChainsaw:
This is a partial fix for #1319. 

It changes the signature for calc_padding to include the layertype since ConvTranspose needs slightly different treatment.

### PR Checklist

- [X] Tests are added (updated)
- [ ] Entry in NEWS.md
- [X] Documentation, if applicable (docstring only)


DhairyaLGandhi:
Hmm, this would break the API somewhat, but I doubt many people are using `calc_padding` directly in their code. Implementation looks good otherwise.

DrChainsaw:
Yes, but `calc_padding` is not exported so its still not a breaking change, right?

I guess that the new behaviour of ConvTranspose is a bit of a grey zone as the old behaviour was the documented behaviour. I doubt anyone found it useful but it could still break someones code in theory.

I don't think this is news.md worthy, but let me know otherwise or if there is anything else I can do.

DhairyaLGandhi:
Thanks!

bors r+

PR #1325, opened 2020-08-31
Add AbstractOptimiser type

DhairyaLGandhi:
This adds an abstract type over all of Flux's optimisers, which serves to disambiguate an optimiser defined by Flux when there are multiple optimisation libraries being used together. Does this need news?

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -1,6 +1,8 @@
 using Flux
 using MacroTools: @forward
 
+abstract type AbstractOptimiser end
+
 const ϵ = 1e-8
 
 # TODO: should use weak refs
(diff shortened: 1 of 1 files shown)


ToucheSir:
Quick Q: is this a precursor to integrating Optimisers.jl or an unrelated change?

DhairyaLGandhi:
For the PR, I'm seeing it as an orthogonal utility. It will come in handy with the functional optimisers too, since we can define some methods on abstract types now.

ChrisRackauckas:
Bump? You're making us do bad things 😆 https://github.com/SciML/GalacticOptim.jl/pull/135/files#diff-540e217120afa5df2f98a3967f3f310bcae228d239cea1c8d3f1f2ef2065254eR1

DhairyaLGandhi:
So I take it you're in favour :p ?

DhairyaLGandhi:
Let's merge this and see where we want to go with this.

PR #1333, opened 2020-09-11
change nadam cite

JeffFessler:
the openreview site is likely to be more archival than some individual's web site
and it can be helpful for readers to see the forum comments.


DhairyaLGandhi:
Looks good, thanks!

bors r+

PR #1351, opened 2020-10-08
deprecated @treelike for @functor with warning

### LINKED ISSUE #1339: deprecate treelike
```julia
using Flux
using Flux: @treelike
struct Cat{T}
    layers::T
end
@treelike Cat
```

> LoadError: UndefVarError: functorm not defined
> in expression starting at In[17]:6
> 
> Stacktrace:
>  [1] @treelike(::LineNumberNode, ::Module, ::Vararg{Any,N} where N) at /home/cossio/.julia/packages/Flux/05b38/src/functor.jl:58
>  [2] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091
>

mungu42:
closes #1339

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -55,7 +55,8 @@ end
 
 # Deprecated stuff
 macro treelike(args...)
-  functorm(args...)
+  @warn "@treelike has been deprecate please use @functor instead"
+  functor(args...)
 end
 mapleaves(f, x) = fmap(f, x)
 


src/functor.jl
```diff
@@ -55,7 +55,8 @@ end
 
 # Deprecated stuff
 macro treelike(args...)
-  functorm(args...)
+  @warn "@treelike has been deprecate please use @functor instead"
+  functor(args...)
```
CarloLucibello, line 59: `functorm` is gone as well? is `functor` the correct replacement here?

mungu42: seems like it, couldn't find it using global search

mungu42: Currently, #1339 also mentions that `@functor` should be used

DhairyaLGandhi: `functnorm` has been moved to Functors.jl, so using it from there would be an option

CarloLucibello:
I just removed `@treelike` in #1377, so I will close this PR

PR #1358, opened 2020-10-15
Fix BPTT by overriding stateful broadcast adjoint

### LINKED ISSUE #1209: Backprop through time is truncated to only 1 time step
Unless my understanding of Backpropagation Through Time (BPTT) and Flux/Zygote is off, it seems like BPTT isn't working as intended with Flux/Zygote.

Currently, the gradient being calculated with respect to Wh does not look back in time. In other words, if we have a sequence of length 3,

```
rnn = Flux.RNN(2, 3)
seq = [rand(2) for i = 1:3]
```

then the following two gradients (`grads_seq` and `grads_2`) are the same.

```
Flux.reset!(rnn);
grads_seq = gradient(Flux.params(rnn)) do
    sum(rnn.(seq)[3])
end
```
and
```
Flux.reset!(rnn);
rnn(seq[1])
rnn(seq[2])
grads_2 = gradient(Wh->sum(tanh.(rnn.cell.Wi * seq[3] +
                                 Wh * rnn.state + rnn.cell.b)), rnn.cell.Wh)

```

Whereas, the gradient for BPTT should be as follows.

```
Flux.reset!(rnn);
bptt = gradient(Wh->sum(tanh.(rnn.cell.Wi * seq[3] + Wh *
                                tanh.(rnn.cell.Wi * seq[2] + Wh *
                                      tanh.(rnn.cell.Wi * seq[1] +
                                            Wh * rnn.init + rnn.cell.b)
                                      + rnn.cell.b)
                                + rnn.cell.b)),
                rnn.cell.Wh)

```

Issue #1168 is possibly related, and here is a [gist summarizing this](https://gist.github.com/AlexLewandowski/0fe0222ffb8394946f83b479637563ab).

DhairyaLGandhi:
Fixes #1209 

In this PR, we replace the regular broadcasting adjoint with that of the `map` equivalent which is better tested in terms of stateful cases. We ultimately will revert back to the broadacasting adjoint via https://github.com/FluxML/Zygote.jl/pull/807 but this specialises the case for recurrent layers 

@oxinabox @ToucheSir Comments?
### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -184,3 +184,7 @@ See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
 """
 GRU(a...; ka...) = Recur(GRUCell(a...; ka...))
+
+@adjoint function Broadcast.broadcasted(f::Recur, args...)
+  Zygote.∇map(__context__, f, args...)
+end
--- a/test/layers/recurrent.jl
+++ b/test/layers/recurrent.jl
@@ -1,0 +1,20 @@
+Ref FluxML/Flux.jl#1209
+@testset "BPTT" begin
+  seq = [rand(2) for i = 1:3]
+  for rnn ∈ [RNN, LSTM, GRU]
+    Flux.reset!(rnn)
+    grads_seq = gradient(Flux.params(rnn)) do
+      sum(rnn.(seq)[3])
+    end
+    Flux.reset!(rnn);
+    bptt = gradient(Wh->sum(tanh.(rnn.cell.Wi * seq[3] + Wh *
+                                  tanh.(rnn.cell.Wi * seq[2] + Wh *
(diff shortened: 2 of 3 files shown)


ToucheSir:
I don't have the expertise to test this personally, but cc'ing @AzamatB, @AStupidBear, @opus111 et al. since they have RNNs working on Tracker.

oxinabox (approved):

oxinabox:
If tests pass, then it should be good.
I assume those tests come straight from the MWE?

DhairyaLGandhi:
That's right, yes

We need the lstm and gru tests too to be complete, but I want the functionality in there so we can close a long pending issue

bors r+

bhvieira:
The most prudent thing to do at the moment, hopefully it won't be long until broadcast adjoints.

DhairyaLGandhi:
Not long at all 🙂

PR #1366, opened 2020-10-20
Update functions.jl

okaerin:
Huber loss should be squared -> https://en.wikipedia.org/wiki/Huber_loss

Apparently the documentation needs to be updated for the Huber loss, as it does not match the definition.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


CarloLucibello:
thanks!

bors r+

PR #1374, opened 2020-10-28
Fix some issues with Zeros

### LINKED ISSUE #1332: GPU error when using Zeros() as bias in Conv layer

Minimal example:

```julia
using Flux
using CUDA
using Flux: onehotbatch
using Flux: crossentropy
using Flux: convfilter, Zeros

CUDA.allowscalar(false)

opt = ADAM()
ps = params(model);
gs = gradient(ps) do
	return training_loss
end
```

```julia
Stacktrace:
 [1] Bool at float.jl:73
 [2] convert at number.jl:7
Stacktrace:
adcast.jl:845
```

DrChainsaw:
Fixes #1332, #1277 

Two things where causing trouble:
1) `adapt` (used to change type of parameters) turning `Zeros()` into `T`s
2) `reshape` (used in conv-layers) turning `Zeros` into `Base.ReshapedArray`s

Both are adressed by specializations.

I'm not sure what is the right way to overload reshape. Leaving the type out for the second argument leads to ambiguity, so there are still more than a handful ways to reshape Zeros which leads to issue 2). Any packages which does it I can look at? 

I was too lazy to come up with a nice way to check dimensions of parameters when creating a layer.

Furthermore, a number of adjoints are defined to ensure that gradients of any `Zeros` is nothing.

Also fixes a few issues with `loadparams!` and `destructure` in the presence of `Zeros`. 

### PR Checklist

- [X] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -1,4 +1,4 @@
-import Adapt: adapt, adapt_storage
+import Adapt: adapt, adapt_storage, adapt_structure
 using  LinearAlgebra: Cholesky
 using Zygote: IdSet
 import Functors: @functor, functor, fmap
@@ -78,7 +78,7 @@ gpu(x) = use_cuda[] ? fmap(CUDA.cu, x) : x
 adapt_storage(T::Type{<:Real}, xs::AbstractArray{<:Real}) = convert.(T, xs)
 
 paramtype(T::Type{<:Real}, m) = fmap(x -> adapt(T, x), m)
-adapt(::Any, x::Zeros) = x # So that we for example don't accidentally end up with layers having a scalar for bias
+adapt_structure(to, x::Flux.Zeros) = x # So that we for example don't accidentally end up with layers having a scalar for bias
 
 f32(m) = paramtype(Float32, m)
 f64(m) = paramtype(Float64, m)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -78,6 +78,7 @@ gpu(x) = use_cuda[] ? fmap(CUDA.cu, x) : x
 adapt_storage(T::Type{<:Real}, xs::AbstractArray{<:Real}) = convert.(T, xs)
 
 paramtype(T::Type{<:Real}, m) = fmap(x -> adapt(T, x), m)
+adapt(::Any, x::Zeros) = x # So that we for example don't accidentally end up with layers having a scalar for bias
 
 f32(m) = paramtype(Float32, m)
 f64(m) = paramtype(Float64, m)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -61,7 +61,7 @@ mapleaves(f, x) = fmap(f, x)
 
 function loadparams!(m, xs)
   for (p, x) in zip(params(m), xs)
-    size(p) == size(x) ||
+    (any(y -> y isa Zeros, (p, x)) || size(p) == size(x)) ||
       error("Expected param size $(size(p)), got $(size(x))")
     copyto!(p, x)
   end
--- a/test/layers/conv.jl
+++ b/test/layers/conv.jl
@@ -42,11 +42,13 @@ end
   op = bias(ip)
   @test sum(op) == prod(size(op))
 
-  bias = Conv((2,2), 1=>3, bias = Flux.Zeros())
-  op = bias(ip)
-  @test sum(op) ≈ 0.f0
-  gs = gradient(() -> sum(bias(ip)), Flux.params(bias))
-  @test gs[bias.bias] == nothing
+  @testset "Zeros mapped through $lmap" for lmap in (identity, cpu, f32)
+    bias = Conv((2,2), 1=>3, bias = Flux.Zeros())
+    op = bias(ip)
+    @test sum(op) ≈ 0.f0
+    gs = gradient(() -> sum(bias(ip)), Flux.params(bias))
+    @test gs[bias.bias] === nothing
+  end
 
   # Train w/o bias and make sure no convergence happens
   # when only bias can be converged
--- a/test/layers/conv.jl
+++ b/test/layers/conv.jl
@@ -43,7 +43,7 @@ end
   @test sum(op) == prod(size(op))
 
   @testset "Zeros mapped through $lmap" for lmap in (identity, cpu, f32)
-    bias = Conv((2,2), 1=>3, bias = Flux.Zeros())
+    bias = Conv((2,2), 1=>3, bias = Flux.Zeros()) |> lmap
     op = bias(ip)
     @test sum(op) ≈ 0.f0
     gs = gradient(() -> sum(bias(ip)), Flux.params(bias))
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -131,8 +131,8 @@ end
 
 @testset "Zeros" begin
   m = Dense(randn(2,3), Zeros())
-  @test f64(m).b === m.b
-  @test f32(m).b === m.b
+  @test f64(m).b === m.b === Zeros()
+  @test f32(m).b === m.b === Zeros()
 end
 
 @testset "Stacking" begin
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -190,3 +190,43 @@ end
   @test stack(unstacked_array, 2) == stacked_array
   @test stack(unstack(stacked_array, 1), 1) == stacked_array
 end
+
+@testset "Param remapping" begin
+  ls(dims...) = reshape(collect(Float32, 1:prod(dims)), dims...)
+  dl(nin, nout, bias) = Dense(ls(nin, nout), bias(nout)) 
+  dm(bias) = Chain(
+    dl(3, 5, bias),
+    dl(5, 4, bias),
+    dl(4, 3, bias)
+  )
+
+  testdense(m, bt) = @testset "Check layer $i" for (i, (l1, l2)) in enumerate(zip(m, dm(bt)))
+  @test l1.W == l2.W
+  @test l1.b == l2.b
+  @test typeof(l1.b) === typeof(l2.b)
+  end
+
+  ZerosNoShape(::Integer) = Zeros()
+
+  pararray(m) = mapreduce(l -> collect(params(l).order), vcat, m)
+
+  @testset "Dense bias type $bt" for bt in (zeros, Zeros, ZerosNoShape)
+    m = dm(bt)
+    Flux.loadparams!(m, params(m))
+    testdense(m, bt)
+  end
+
+
+  @testset "$b1 to $b2" for (b1, b2, be) in (
+    (zeros, ones, ones),          # Load ones as bias to a model with zeros as bias -> model gets ones as bias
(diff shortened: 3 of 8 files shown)


src/functor.jl
```diff
@@ -78,6 +78,7 @@ gpu(x) = use_cuda[] ? fmap(CUDA.cu, x) : x
 adapt_storage(T::Type{<:Real}, xs::AbstractArray{<:Real}) = convert.(T, xs)
 
 paramtype(T::Type{<:Real}, m) = fmap(x -> adapt(T, x), m)
+adapt(::Any, x::Zeros) = x # So that we for example don't accidentally end up with layers having a scalar for bias
```
DhairyaLGandhi, line 81: Is the `Any` correct here?

DhairyaLGandhi (commented):
Love seeing the green CI!

test/layers/conv.jl
```diff
@@ -42,11 +42,13 @@ end
...
-  op = bias(ip)
-  @test sum(op) ≈ 0.f0
-  gs = gradient(() -> sum(bias(ip)), Flux.params(bias))
-  @test gs[bias.bias] == nothing
+  @testset "Zeros mapped through $lmap" for lmap in (identity, cpu, f32)
+    bias = Conv((2,2), 1=>3, bias = Flux.Zeros())
+    op = bias(ip)
+    @test sum(op) ≈ 0.f0
+    gs = gradient(() -> sum(bias(ip)), Flux.params(bias))
+    @test gs[bias.bias] === nothing
```
DhairyaLGandhi, line 50: Need to test this with CUDA as well

src/functor.jl
DrChainsaw: Yes will change to `adapt_structure` following Tims advice!

`::Any` instead of just `T` or `x` is basically just because vs code marks unused variables. Should I change to `T` or did you mean that there a more constrained type to use here?

test/utils.jl
```diff
@@ -129,6 +129,12 @@ end
   @test eltype(f32(f64(m))[1].W) == Float32
 end
 
+@testset "Zeros" begin
+  m = Dense(randn(2,3), Zeros())
+  @test f64(m).b === m.b
```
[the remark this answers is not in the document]
DrChainsaw: Not sure I follow. 

It is a test that `Zeros` do not materialize from `f64`. I'll add `=== Zeros()` in the end for now to make the intention clear.

test/layers/conv.jl
DrChainsaw: I guess you are referring to the test that gradient w.r.t Zeros is nothing, right? Will add it right away!

DrChainsaw:
Btw, I had a look at how [FillArrays.jl](https://github.com/JuliaArrays/FillArrays.jl/blob/d88461b9da0200fba12829225cdad2921ff8d7b2/src/FillArrays.jl#L210-L225) does reshape. Quite a few methods, but nothing insurmountable and maybe not all of them are needed just for Zeros.

Was there ever a discussion to use (or recommend users to use) FillArrays.Zeros instead of rolling an own version?

DrChainsaw:
Another thought: Should e.g. `f32(Zeros())` return a `Zeros{Float32}`? I guess something like this: 

```julia
adapt_structure(::Type{<:AbstractArray{T}}, x::Flux.Zeros) where T = adapt_structure(T, x)
adapt_structure(::Type{T}, x::Flux.Zeros{T}) where T = x 
adapt_structure(T, x::Flux.Zeros) = Flux.Zeros(T, size(x)...)
```
Drawback is that there will then be some inconsistency between layers created through e.g `Conv(...;bias=Zeros())` v.s. the ones which are mapped. So far everything seems to work fine with `Zeros` being of type `Bool` so maybe not touch it now.

DhairyaLGandhi:
The reason we have our own `Zeros` type is that Flux needs to own a type that it can define a bunch of these routines for, otherwise it's very unexpected to see why FillArrays' types should not have gradients.

DrChainsaw:
> The reason we have our own Zeros type is that Flux needs to own a type that it can define a bunch of these routines for

Yeah, though it had something to do with type piracy-ish stuff.

I'm almost done, but I'm trying to squeeze in some testing with optimizers as well and running into some issues with `zero`, so I might need to add that too.

DrChainsaw:
Hmmm, somehow it is only Conv layers which gets no gradient for Zeros, all other ways they are involved seem to result in gradients being created:

```julia
julia> ll = Dense(ones(Float32, 1,1), Flux.Zeros())
Dense(1, 1)

julia> gs = gradient(() -> sum(ll(ones(Float32, 1,1))), params(ll))
Grads(...)

julia> gs[ll.b]
0-dimensional FillArrays.Fill{Float32,0,Tuple{}} = 1.0

julia> ll.b
0-dimensional Flux.Zeros{Bool,0}:
0
``` 

Any ideas?

DrChainsaw:
Ugh, avoiding ambiguity while still catching all permutations of Zeros turned out to be a fair bit more painful than I though. I could not come up with a simpler way than more or less implementing every permutation.

I still need to implement the non-broadcasted OPs and test the / case, but I'm too tired for that now.

CarloLucibello:
we should move this discussion elsewhere but I think it would be quite natural for FillArray.Zeros and FillArray.Ones to have zero gradients since they are meant to be constants. We should really consider switching to FillArray types given the amount of extra bloat having our own Zeros produces.

mcabbott:
Another thought is: might it be simpler not to do this at all? Just store `bias=false` (or `bias=False()`) instead of a fake array. This behaves correctly under broadcast already, and I think scalars are ignored by `params`.

DrChainsaw:
@CarloLucibello I guess Zygote could do the adjoints for it without it being piracy. I can't wrap my head around if it will cause problems to someone somewhere if there are never gradients for FillArrays. I can think of cases when there is inconvenience, but I'm not sure there are any real showstoppers. If new AD means Zygote is reaching eol then perhaps it can be done as an experiment and see what complaints come in :)

@mcabbott I was thinking along similar lines. It might be easier to just make something which only tries to solve the bias on/off case. Another option is to have some `Off` struct which is not an AbstractArray but using `false` is perhaps a bit more elegant as alot of things are already implemented for it. I think one issue is that the conv-layers for some reason need to reshape the bias. An own type could just short-circuit `reshape` though, just like `Zeros` does in this commit.

I'm not sure if it is desirable to have a turned off bias in params or not. Otherwise the same thing can perhaps be done more explicitly with Zeros as well. I remember being a bit surprised to see it there, but have in the back of my head that there was a good reason for it. If it needs to be in params then I think Zygote will return gradients for false too (see below).

About the adjoint overloading in the last commit: The core issue afaik is that Zygote has defined adjoints for broadcasting with numerical arrays (and scalars) so even if the primal op does not use a variable the gradient for it will be created. Could this be fixed in Zygote with reasonable effort? A hack like checking if the result is identical to one of the arguments is one way ofc...

Also, after getting some sleep on it there is a chance I can get rid of the adjoints and instead specialize on `Zygote.unbroadcast` so it returns nothing if one of the arguments is a `Zeros`. For multiplication this would then only work `conj(::Zeros)` is sacrificed so it returns a real array of zeros. Not sure if this will work out with e.g. CuArrays though. The solution would then also become reliant on what I guess should be an implementation detail in Zygote.

mcabbott:
> I think one issue is that the conv-layers for some reason need to reshape the bias. An own type could just short-circuit reshape though, just like Zeros does in this commit.

My foggy recollection was that the original reason for Zeros was to avoid needing a special case in one or two `conv` methods. But that doesn't sound so bad compared to all this.

Have only skimmed the issues on loading/saving & params, but it sounds like the goal is to ignore absent biases, but `Zeros <: AbstractArray` means it needs all kinds of special-casing there.

DrChainsaw:
> But that doesn't sound so bad compared to all this.

No disagreement on my side there. This endeavour turned out to be quite the rabbit hole :)

One option is that I just open another PR where try to do the `bias=false` approach and see where that takes us.

DhairyaLGandhi:
Definitely shouldn't change the behaviour of FillArrays types, and we should keep the forward passes untouched since this is orthogonal to how those are supposed to function.

Seems like this needs the correct adapt rules (check out `CUDA.Adaptor`) and a way to tell cuda not to materialize it till there's a kernel call to CuDNN perhaps.

DhairyaLGandhi:
We had considered the `false` approach while originally discussing the bias switch, I think we need to be careful not to do something which isn't generic enough. I agree that adjoints with `Zeros` can help here

DrChainsaw:
@DhairyaLGandhi I will look into the CUDA.Adaptor. I could not find it when browsing the code on github, but I will search once I'm back in front of the computer. My thinking right now is that the specialized `adapt_structure` prevents them from turning into CuArrays and the specialized broadcasts will be used even if one argument is a CuArray. Is this about more generic things (like `cu([1]) .+ 0 .+ Flux.Zeros()` which currently fails)?

I looked through the issue for the param loading case: #1277

It seems that at least the issue with Zeros turning to Arrays is solved by this PR. 

I will look at tests for `loadparams!` and `destructure` and drop some `Zeros` in there and see what happens now. It seems as those methods rely on params being `AbstractArrays`. Right now I can't wrap my head around if things will just work if non-existent biases are just not added to the set of params or not.

One obvious case is if someone tries to add a bias to a non-bias layer with `loadparams!`. This should throw an error as it is outside the scope of just modifying the parameters, right? Restructure otoh should perhaps work...

DrChainsaw:
@DhairyaLGandhi I looked up CUDA.Adaptor in the code but I could still not determine if it had any other purpose than moving stuff to the GPU which is what we want to avoid altogether here, right?

Is this sufficient to tell if it is handled correctly?

```julia
julia> cu(Flux.Zeros())
0-dimensional Flux.Zeros{Bool,0}:
0

julia> cu(Flux.Zeros(2))
2-element Flux.Zeros{Bool,1}:
 0
 0
```

DrChainsaw:
Adding fixes and tests for `loadparams!` and `destructure` made it seem slightly more attractive in my eyes to explore the path where bias is not an AbstractArray. 

There are two ways one can view the act of disabling bias:

1) It is immutable (and zero)
2) It is simply just not there

I think the current approach with `Zeros` being an `AbstractArray` tries to do a little bit of both, or at least it sends that message. 

For `destructure` having params as an `AbstractArray` was actively unhelpful as the contract is that the returned array is something which can be changed. I guess there are packages for exotic arrays which have different types for different parts of the array but I doubt that we want to go there. Whats on the branch now just makes the assumption that `bias=Zeros()` shall be treated as if the bias just does not exist, meaning they are not included in the destructured array. 

`loadparams!` otoh is a bit of a mixed bag where immutable structs being singletons makes it awkward to e.g load between models where one has bias and the other has not. I'm not sure this is how it is supposed to be used though and for cases when the params to replace with are created manually I can't see how it matters.

DrChainsaw:
Ok, I think functionality wise this PR is now ready as far as I can tell.

I will start drafting another version where Zeros is not an AbstractArray and submit that as a separate PR.

DrChainsaw:
Redundant after #1379

PR #1377, opened 2020-10-31
remove Datasets + additional deprecations

CarloLucibello:
Remove Flux's datasets in favor of data providers in the julia ecosystem.


CarloLucibello:
@DhairyaLGandhi are you ok with this change?

DhairyaLGandhi:
Best to deprecate no?

CarloLucibello:
problem is each dataset is a module, I searched for a while but I didn't find how to deprecate a module

DhairyaLGandhi:
Hmm, not sure but we might end up with add depwarns to the methods that people would call for the datasets?

CarloLucibello:
bors try

DhairyaLGandhi:
Many thanks to @KristofferC !!
```julia
 julia> module Foo
           module Bar
           end
           
           Base.@deprecate_binding Bazz Bar
        end
 Main.Foo
 
 julia> Foo.Bazz
 WARNING: Foo.Bazz is deprecated, use Bar instead.
   likely near REPL[2]:1
 Main.Foo.Bar
```

docs/src/ecosystem.md
```diff
@@ -17,5 +17,13 @@ machine learning and deep learning workflows:
 - [ProgressMeters.jl](https://github.com/timholy/ProgressMeter.jl): progress meters for long-running computations
 - [TensorBoardLogger.jl](https://github.com/PhilipVinc/TensorBoardLogger.jl): easy peasy logging to [tensorboard](https://www.tensorflow.org/tensorboard) in Julia
 
+This tight integration among Julia pakages is shown in some of the examples in the [model-zoo](https://github.com/FluxML/model-zoo) repository.
 
-This tight integration among Julia pakages is shown in some of the examples in the [model-zoo](https://github.com/FluxML/model-zoo) repository.
\ No newline at end of file
+List of packages built on top of Flux:
```
DhairyaLGandhi, line 22: We should avoid having this list here for now and it should be the synced with the list in the website. And in general we should also do a once over the packages as well

CarloLucibello: > and it should be the synced with the list in the website. And in general we should also do a once over the packages as well

I don't understand

DhairyaLGandhi:
Why did this not follow the deprecations route? It makes breaking API changes without the deprecations themselves allowing users to continue using the functionality.

CarloLucibello:
I did introduce informative errors pointing users to MLDatasets, it should be very easy for them to switch. Also, recent examples in model-zoo have been relying on MLDatasets for a long time. 
I can make a new PR with a softer deprecation path by importing MLDatasets here and rerouting its function to match the previous Flux api. I prefer not to though, it is a lot of work to do here, while users can overcome the change pretty easily

DhairyaLGandhi:
I'd prefer a warning about them being removed and using the datasets we had for now, and removing them in a couple of releases.

Agree on not adding MLDatasets as a dep

PR #1386, opened 2020-11-06
Add GitHub Actions to CI

DhairyaLGandhi:
Travis migration basically

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.julia-version == 'nightly' }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
             Pkg.instantiate()'
       - run: |
           julia --project=. -e '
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
Is there a good way to say pass when nightly fails?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,67 @@
+name: CI
+on:
+  - push
+  - pull_request
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.julia-version == 'nightly' }}
```
CarloLucibello, line 9: looks like `continue-on-error` has to be used after each step. Saw that in some PR from @oxinabox. See also https://github.com/FluxML/Zygote.jl/pull/834

DhairyaLGandhi: Seems to still show a Red Cross. Also, need to set up the deploy keys again maybe.

CarloLucibello: If you look at the commented out lines involving 'continue-on-error' in that PR, that's how you get rid of red cross.
I'm not sure that's the correct thing to do though: ideally, I would like to see nightly fails, so that I'm aware there is something to fix before the next julia release, but the overall check for the PR should report "tests were successful". I don't think there is any way to have that though

DhairyaLGandhi: I mean, we are already know the issues with IRTools, and then Zygote on Julia master, so it just seems unnecessary to monitor something we know will fail. Once we have it passing, we can monitor that metric but until then, best to see the results from the contents of the PR.

PR #1388, opened 2020-11-07
eliminate most allocations from get! in optimisers

simeonschaub:
For my model, this takes allocations in `apply!` down from 2 GB(!) to just 300 kB. Especially allocating `zero(x)` two times in every step added up very quickly.

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -68,7 +68,7 @@ Momentum(η = 0.01, ρ = 0.9) = Momentum(η, ρ, IdDict())
 
 function apply!(o::Momentum, x, Δ)
   η, ρ = o.eta, o.rho
-  v = get!(o.velocity, x, zero(x))::typeof(x)
+  v = get!(() -> zero(x), o.velocity, x)::typeof(x)
   @. v = ρ * v - η * Δ
   @. Δ = -v
 end
@@ -101,7 +101,7 @@ Nesterov(η = 0.001, ρ = 0.9) = Nesterov(η, ρ, IdDict())
 
 function apply!(o::Nesterov, x, Δ)
   η, ρ = o.eta, o.rho
-  v = get!(o.velocity, x, zero(x))::typeof(x)
+  v = get!(() -> zero(x), o.velocity, x)::typeof(x)
   d = @. ρ^2 * v - (1+ρ) * η * Δ
   @. v = ρ*v - η*Δ
   @. Δ = -d
@@ -138,7 +138,7 @@ RMSProp(η = 0.001, ρ = 0.9) = RMSProp(η, ρ, IdDict())
 
 function apply!(o::RMSProp, x, Δ)
   η, ρ = o.eta, o.rho
-  acc = get!(o.acc, x, zero(x))::typeof(x)
+  acc = get!(() -> zero(x), o.acc, x)::typeof(x)
   @. acc = ρ * acc + (1 - ρ) * Δ^2
   @. Δ *= η / (√acc + ϵ)
 end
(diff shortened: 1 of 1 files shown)


CarloLucibello:
This is nice for cpu but it's not gpu friendly, I think we can only keep the get! improvement

simeonschaub:
That's definitely the biggest one here, but can't we use dispatch to take care of the `GPUArrays` case? I can double-check though how much broadcasting still allocates with the type annotations.

CarloLucibello:
btw, I remember I read somewhere that `@.` may fail to fuse the operations. You may try to use dotted operators instead.

DhairyaLGandhi:
Is that enough to avoid the memory allocation?

simeonschaub:
> btw, I remember I read somewhere that `@.` may fail to fuse the operations. You may try to use dotted operators instead.

I doubt that's the problem here, I suspect it's the broadcasting machinery failing to fully inline, that's why I am wondering, whether this might even be fixed by just the type annotations.

> Is that enough to avoid the memory allocation?

The line with `get!` alone showed 1.9GB of allocations, because it was allocating 2x `zero(x)` for every training step. Now this only allocates if there is no entry with key `x`, which should only be at the start of training.

simeonschaub:
Ok, apparently the broadcast calls don't allocate anymore, now that `mt`, `vt` and `βp` have type annotations, so I just reverted that part.

CarloLucibello:
- is type annotation really necessary?
- could you change the other optimizers as well?

simeonschaub:
> is type annotation really necessary?

Yes, I just checked, without them I get a couple 100 kBs of allocations for the broadcast calls.

> could you change the other optimizers as well?

Ok, I can do that.

CarloLucibello:
> Yes, I just checked, without them I get a couple 100 kBs of allocations for the broadcast calls.

we could live with those, considering that we were at 2GB before this PR :) I think it's better to have a few irrelevant extra allocations than some code bloat

DhairyaLGandhi:
Are there before and after benchmarks for Adam?

simeonschaub:
I don't know if I would call this code bloat. Base has moved to adding a few annotations in the right places, not just to improve performance, but also to avoid method invalidations, which are responsible for increased latency, so I do think, we are going to see more of this in idiomatic Julia code. It feels a bit wrong to me, to leave performance on the table here, if it's just one annotation. Because `state` is not typed, the compiler will never be able to infer this otherwise, so it does seem worth it to me.

simeonschaub:
> Are there before and after benchmarks for Adam?

I have mostly just looked at the allocations, as running the model takes a while and performance is a bit noisy. Those are the allocations with 6 batches before the change:
```
        - function apply!(o::ADAM, x, Δ)
        0   η, β = o.eta, o.beta
1912684416   mt, vt, βp = get!(o.state, x, (zero(x), zero(x), β))
   450432   @. mt = β[1] * mt + (1 - β[1]) * Δ
   750720   @. vt = β[2] * vt + (1 - β[2]) * Δ^2
   450432   @. Δ =  mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + ϵ) * η
  1051008   o.state[x] = (mt, vt, βp .* β)
        0   return Δ
        - end
```
This is after, but only one batch, so these times should be multiplied by 6:
```
        - function apply!(o::ADAM, x, Δ)
        0   η, β = o.eta, o.beta
    50048   mt, vt, βp = get!(() -> (zero(x), zero(x), Float64[β[1], β[2]]), o.state, x)::Tuple{typeof(x),typeof(x),Vector{Float64}}
        0   @. mt = β[1] * mt + (1 - β[1]) * Δ
        0   @. vt = β[2] * vt + (1 - β[2]) * Δ^2
        0   @. Δ =  mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + ϵ) * η
        0   βp .= βp .* β
        0   return Δ
        - end
```

src/optimise/optimisers.jl
```diff
@@ -171,11 +171,16 @@ ADAM(η = 0.001, β = (0.9, 0.999)) = ADAM(η, β, IdDict())
 
 function apply!(o::ADAM, x, Δ)
   η, β = o.eta, o.beta
-  mt, vt, βp = get!(o.state, x, (zero(x), zero(x), β))
```
CarloLucibello, line 174: did we have a bug here? seems like mutating βp affects also β ...

CarloLucibello: ah, we didn't, cause we were storing each time a new βp in state

CarloLucibello (approved):

src/optimise/optimisers.jl
simeonschaub: Yes, I restructured this a little bit. This way, we don't have to access `o.state` twice.

PR #1391, opened 2020-11-09
Improve docs for `Conv` etc.

mcabbott:
This re-writes the docstrings for the convolutional layers, as I thought they were confusing and occasionally misleading. (I guess they grew organically over time.) It adds many examples as these often seem clearer than words; hopefully they will be run as doctests but I'm not entirely sure. 

No functional changes.

One small printing change, that `MaxPool((5, 5), pad = 2, stride = (5, 5))` prints like this, not `pad = (2, 2, 2, 2)`. Perhaps that should be made still less verbose. Or `Conv` more verbose?


DrChainsaw:
Worth pointing out that `ConvTranspose` is the odd bird in the conv-gang w.r.t SamePad as it has the relationship outsize = insize * stride.

src/layers/conv.jl
```diff
@@ -77,37 +97,43 @@ struct Conv{N,M,F,A,V}
 end
 
 """
-    Conv(weight::AbstractArray, bias::AbstractArray)
-    Conv(weight::AbstractArray, bias::AbstractArray, activation)
+    Conv(weight::AbstractArray, bias, [activation; stride, pad, dilation])
```
CarloLucibello, line 100:
```suggestion
    Conv(weight, bias, activation=identity; stride=1, pad=0, dilation=1)
```

src/layers/conv.jl
```diff
@@ -30,41 +32,59 @@ function calc_padding(lt, ::SamePad, k::NTuple{N,T}, dilation, stride) where {N,
 end
 
 """
-    Conv(filter, in => out, σ = identity; init = glorot_uniform,
-         stride = 1, pad = 0, dilation = 1)
+    Conv(filter, in => out, σ = identity; stride = 1, pad = 0, dilation = 1)
```
CarloLucibello, line 35:
```suggestion
    Conv(filter, in => out, σ=identity; stride=1, pad=0, dilation=1)
```

src/layers/conv.jl
mcabbott: My goal was to keep these secondary docstrings brief, and push all the explanation of what things mean & what their defaults are up to the first docstring. At least at the REPL, they are printed together. Perhaps in addition to saying "Accepts the same keywords" it should say "with the same defaults"?

src/layers/conv.jl
mcabbott: I'd vote against all these spaces too, great! (I guess I left them in case it was the preferred style around here.)

mcabbott: Killed some unnecessary square brackets too.

CarloLucibello:
This PR is a great improvement, thanks! It's ready to be merged?

mcabbott:
I have one more commit, to simplify printing of `MaxPool` by hiding strides & padding when they have their default values. May as well, right?

CarloLucibello:
sure

CarloLucibello:
> I have one more commit, to simplify printing of `MaxPool` by hiding strides & padding when they have their default values. May as well, right?

actually I'm not sure about this, but I think it's ok since the defaults are so commonly used and intuitive for this layer

mcabbott:
Can un-do if you think it's better. 

At the moment `Conv` doesn't print these things set by keywords at all.

CarloLucibello:
look's fine like this, thanks

bors r+

mcabbott:
Note to self: Only the docstring for the first method of `Conv` seems to show up here:

https://fluxml.ai/Flux.jl/dev/models/layers/#Flux.Conv

According to https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@docs-block one can give signatures, but they must be exact. So we should fix that once the signature has settled down.

PR #1393, opened 2020-11-14
remove implicit conversions

CarloLucibello:
For some layers, we currently downcast the input type from Float64 to Float32 if the weights are Float32. 
I think we should follow julia's promotion rules, users should provide Float32 inputs if they want Float32 outputs. 
This also simplifies layers' definitions.

This change may have a performance impact on some people code that may go unnoticed, so while we want to promote good practice, I can understand if someone opposes this change. 

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md
- [x] Final review from `@dhairyagandhi96` (for API changes).

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -119,10 +119,8 @@ end
 
 @functor Dense
 
-function (a::Dense)(x::AbstractArray)
-  if eltype(a.W) != eltype(x)
-    @warn "Element types of input and weights differ." W=eltype(a.W) x=eltype(x) maxlog=1
-  end
+function (a::Dense)(x::AbstractVecOrMat)
+  eltype(a.W) == eltype(x) || _dense_typewarn(a, x)
   W, b, σ = a.W, a.b, a.σ
   # reshape to handle dims > 1 as batch dimensions
   sz = size(x)
@@ -130,6 +128,9 @@ function (a::Dense)(x::AbstractArray)
   x = σ.(W*x .+ b)
   return reshape(x, :, sz[2:end]...)
 end
+
+_dense_typewarn(d, x) = @warn "Element types don't match for layer $d, this will be slow." typeof(d.W) typeof(x) maxlog=1
+Zygote.@nograd _dense_typewarn
 
 function Base.show(io::IO, l::Dense)
   print(io, "Dense(", size(l.W, 2), ", ", size(l.W, 1))
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -120,7 +120,6 @@ end
 @functor Dense
 
 function (a::Dense)(x::AbstractVecOrMat)
-  eltype(a.W) == eltype(x) || _dense_typewarn(a, x)
   W, b, σ = a.W, a.b, a.σ
   # reshape to handle dims > 1 as batch dimensions
   sz = size(x)
@@ -128,9 +127,6 @@ function (a::Dense)(x::AbstractVecOrMat)
   x = σ.(W*x .+ b)
   return reshape(x, :, sz[2:end]...)
 end
-
-_dense_typewarn(d, x) = @warn "Element types don't match for layer $d, this will be slow." typeof(d.W) typeof(x) maxlog=1
-Zygote.@nograd _dense_typewarn
 
 function Base.show(io::IO, l::Dense)
   print(io, "Dense(", size(l.W, 2), ", ", size(l.W, 1))
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -119,11 +119,10 @@ end
 
 @functor Dense
 
-function (a::Dense)(x::AbstractVecOrMat)
+function (a::Dense)(x::AbstractArray)
   W, b, σ = a.W, a.b, a.σ
-  # reshape to handle dims > 1 as batch dimensions
   sz = size(x)
-  x = reshape(x, sz[1], :) 
+  x = reshape(x, sz[1], :) # reshape to handle dims > 1 as batch dimensions 
   x = σ.(W*x .+ b)
   return reshape(x, :, sz[2:end]...)
 end
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -120,6 +120,9 @@ end
 @functor Dense
 
 function (a::Dense)(x::AbstractArray)
+  if eltype(a.W) != eltype(x)
+    @warn "Element types of input and weights differ." W=eltype(a.W) x=eltype(x) maxlog=1
+  end
   W, b, σ = a.W, a.b, a.σ
   # reshape to handle dims > 1 as batch dimensions
   sz = size(x)
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -134,14 +134,6 @@ function Base.show(io::IO, l::Dense)
   print(io, ")")
 end
 
-# Try to avoid hitting generic matmul in some simple cases
-# Base's matmul is so slow that it's worth the extra conversion to hit BLAS
-(a::Dense{<:Any,W})(x::AbstractArray{T}) where {T <: Union{Float32,Float64}, W <: AbstractArray{T}} =
-  invoke(a, Tuple{AbstractArray}, x)
-
-(a::Dense{<:Any,W})(x::AbstractArray{<:AbstractFloat}) where {T <: Union{Float32,Float64}, W <: AbstractArray{T}} =
-  a(T.(x))
-
 """
     Diagonal(in::Integer)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,8 @@
 
 ## v0.12.0
 
-* The Dense layer now supports inputs with [multiple batch dimensions](https://github.com/FluxML/Flux.jl/pull/1405)
+* The Dense layer now supports inputs with [multiple batch dimensions](https://github.com/FluxML/Flux.jl/pull/1405).
+* Dense and Conv layers no longer perform  [implicit type conversion](https://github.com/FluxML/Flux.jl/pull/1394).
 * Excise datasets in favour of other providers in the julia ecosystem.
 * Added option to set `bias` to [false](https://github.com/FluxML/Flux.jl/pull/1379) to eliminating `bias` from being trained.
 * Removed kwarg only constructors for [`convolutional layers`](https://github.com/FluxML/Flux.jl/pull/1379)).
--- a/docs/src/performance.md
+++ b/docs/src/performance.md
@@ -13,7 +13,6 @@ not because the operations are faster, but because the memory usage is halved.
 Which means allocations occur much faster.
 And you use less memory.
 
-
 ## Preserve inputs' types
 
 Not only should your activation and loss functions be [type-stable](https://docs.julialang.org/en/v1/manual/performance-tips/#Write-%22type-stable%22-functions-1),
@@ -21,8 +20,8 @@ they should also preserve the type of their inputs.
 
 A very artificial example using an activation function like
 
-```
-    my_tanh(x) = Float64(tanh(x))
+```julia
+my_tanh(x) = Float64(tanh(x))
 ```
 
 will result in performance on `Float32` input orders of magnitude slower than the normal `tanh` would,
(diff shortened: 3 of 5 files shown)


mcabbott:
I'm not sure these conversions for `Conv` get applied at all, since the wrong floats seem to get passed to NNlib:
```julia
julia> Conv((3,3), 5=>7)(rand(10,10,5,1));
┌ Warning: Slow fallback implementation invoked for conv!  You probably don't want this; check your datatypes.
│   yT = Float64
│   T1 = Float64
│   T2 = Float32
└ @ NNlib ~/.julia/packages/NNlib/Z9qZP/src/conv.jl:206
```
I'm all for scrapping this too-fancy dispatch. And for printing warnings.

CarloLucibello:
In fact, I believe this was used only by the Dense layer. I will add a warning also there

CarloLucibello:
this should be ready to merge @DhairyaLGandhi

mcabbott:
+1 for the explicit warning. But I think you will need to hide `@warn` from Zygote. One way is:
```
    warn_string("Input of type ", typeof(x)," does not match layer ", d)

warn_string(arg...) = @warn string(arg...) maxlog=3 _id=hash(arg)
Zygote.@nograd warn_string
```

src/layers/basic.jl
```diff
@@ -134,14 +137,6 @@ function Base.show(io::IO, l::Dense)
   print(io, ")")
 end
```
mcabbott, line 139:
```suggestion
function (a::Dense)(x::AbstractVecOrMat)
  eltype(a.W) == eltype(x) || _dense_typewarn(a, x)
  W, b, σ = a.W, a.b, a.σ
  σ.(W*x .+ b)
end
 
_dense_typewarn(d, x) = @warn "Element types don't match for layer $d, this will be slow." typeof(d.W) typeof(x) maxlog=3

Zygote.@nograd _dense_typewarn
```

DhairyaLGandhi:
I'm worried about the type instability in the dense case, which really is a shame to have in the first place. How does that match up with our current workaround?

CarloLucibello:
> I'm worried about the type instability in the dense case, which really is a shame to have in the first place. How does that match up with our current workaround?

type instability is not an issue here, the problem is that if types inferred at runtime are homogeneous the multiplication is fast, otherwise is slow. Previously we were silently converting the input to the weight's type, but the real workaround is just doing the right thing and providing coherent  input type

mcabbott:
It's not quite as slow as it was, BTW, as Julia now seems to promote matrix types before multiplying (rather than calling its generic_matmul). But still far from optimal. I think printing warnings is the right thing to do here.

CarloLucibello:
@DhairyaLGandhi do you approve this change?

mcabbott:
I have some tests from an old PR, which perhaps it's easiest to add after this is merged.

CarloLucibello:
@DhairyaLGandhi bump

CarloLucibello:
@DhairyaLGandhi maybe we should revisit the merge process if you don't have enough bandwidth to go through the PRs. We could have a policy like: people with write access should wait a week for some feedback, then can merge their own PRs. Or we add more people to the list of those who can give the final go for api changes. Right now, we have too much latency, even for simple PRs like this one...

DhairyaLGandhi:
Its unfortunately not the bandwidth, its the breakages which cause me to take pause. We *have* to support mixed precision training, esp for low precision cases with fp16 and CUDA *will* accumulate to fp32. Which is why this is something that shouldn't land until the mixed precision is ironed out.

CarloLucibello:
how is this worse than before with respect to multiple precision? (I think it is actually better to not have uncontrolled casts)

DhairyaLGandhi:
using fp16 operations accumulates into fp32 containers on Turing and Ampere, and that is true also for operations that are received downstream. So you'd have `f(<:Union{Float32,Float16}, <:Union{Float16,Float32})::Float32`. Now here, it means we wouldn't be able to use the tensor cores directly if we cast fp16 to fp32.

Also, we should avoid branching in forward passes like so imho.

DhairyaLGandhi:
Personally I'd like to see the invokes go as well, so I am in favour of moving ahead with that direction

mcabbott:
> Also, we should avoid branching in forward passes like so imho

Branches like this get compiled away, since they only depend on types, right? It's just more readable than hacking dispatch to do it, and much less prone to producing ambiguities.

> using fp16 operations accumulates into fp32 containers on Turing and Ampere, and that is true also for operations that are received downstream.

Is this tracked somewhere? If there isn't an issue, would you mind making one, with examples of what code you believe ought to work, which bits are currently supported by CUDA.jl, etc? IIRC there are quite a few flavours of mixed precision stuff, accumulating either gradients or within matmul, in different precision... it would be helpful to lay out somewhere what you think worth supporting. Otherwise it's hard to guess how it intersects things like this issue.

For instance `Conv` layers complain in exactly the way proposed here, which (absent a more careful write-up) seems like the obvious precedent. And demanding explicit type conversions (else warning) seems to me like a sensible policy, for knowing that your code is doing what you think it's doing.

CarloLucibello:
in mixed precision training, you keep a master network in fp32 and create a fp16 copy for the forward and backward. So the whole computation is in fp16, provided you pass f16 inputs (which you should). We should be fine

DhairyaLGandhi:
> in mixed precision training, you keep a master network in fp32 and create a fp16 copy for the forward and backward.

I don't think this is correct, besides if we are, we should avoid that since copying would push our memory issues even further

Cc @maleadt if that's the correct behavior

CarloLucibello:
I got that from here https://arxiv.org/abs/1710.03740

mcabbott:
OK, that's very clear. For that scenario, only `update!` and perhaps `loadparams!` would ever see mis-matched precisions. (And this warning would correctly tell you that you screwed up.)

DhairyaLGandhi:
Putting extra gc pressure seems like it should need opt in behavior

mcabbott:
> Putting extra gc pressure seems like it should need opt in behavior

Yes, that's the point of this. it shouldn't silently copy your data.

CarloLucibello:
wow, that's a lot of CI in action. Why we have (push) and (pull_request)? I don't see that in Zygote's CI.
And we could test only the linux platform on nightly, no need to waste all those resources

src/layers/basic.jl
```diff
@@ -128,7 +128,8 @@ end
 
 @functor Dense
 
-function (a::Dense)(x::AbstractArray)
+function (a::Dense)(x::AbstractVecOrMat)
```
CarloLucibello, line 131:
```suggestion
function (a::Dense)(x::AbstractArray)
```

CarloLucibello:
Fixed a problem with the rebase

CarloLucibello:
Pfff, there is CI error in the documentation step, but should be fixed in https://github.com/FluxML/Flux.jl/pull/1412. @DhairyaLGandhi could you approve this?

src/layers/basic.jl
```diff
@@ -137,20 +138,15 @@ function (a::Dense)(x::AbstractArray)
   return reshape(x, :, sz[2:end]...)
 end
 
+_dense_typewarn(d, x) = @warn "Element types don't match for layer $d, this will be slow." typeof(d.W) typeof(x) maxlog=1
```
DhairyaLGandhi, line 141: Seems orthogonal to what the pr is intended for, I'd forgo this in this pr

CarloLucibello: this is the concern that originally motivated the implicit cast we are removing here

DhairyaLGandhi:
Doing some sanity checks seems to suggest that just having the logging logic is enough to add some overhead, suggesting the compilation is still including that logging logic.

```julia
julia> r = rand(Float32, 3,10);

julia> r2 = rand(Float64, 3,10);

# get rid of warning
julia> d(r2);
┌ Warning: Element types don't match for layer Dense(3, 4), this will be slow.
│   typeof(d.W) = Array{Float32,2}
│   typeof(x) = Array{Float64,2}
└ @ Flux ~/Downloads/new_clones/temp/Flux.jl/src/layers/basic.jl:141

julia> @btime $d($r2);
  821.675 ns (4 allocations: 1.02 KiB)

julia> @btime $d($r);
  272.449 ns (2 allocations: 320 bytes)

# after commenting out branch
julia> @btime $d($r2);
  386.564 ns (3 allocations: 976 bytes)
```

Seems to suffer a penalty just for having the branching present. This can also be seen in the gradient calls, even with the correct types present

```julia

# commenting out the branching using f32 input
julia> @btime gradient((m,x) -> sum(m(x)), $d,$r);
  1.648 μs (31 allocations: 2.38 KiB)

julia> @btime gradient((m,x) -> sum(m(x)), $d,$r);
  14.345 μs (78 allocations: 4.05 KiB)

julia> @btime gradient((m,x) -> sum(m(x)), $d,$r);
  14.444 μs (78 allocations: 4.05 KiB)
```

mcabbott:
That's concerning about the gradient. Does the same thing happen with `Conv`? If not then perhaps we can use dispatch to achieve this and keep Zygote happy. 

No idea why it should care, but notice `y::Any; back::Zygote.var"#41#42"{_A} where _A` here:
```
julia> @code_warntype gradient((m,x) -> sum(m(x)), d, r)
Variables
  #self#::Core.Compiler.Const(Zygote.gradient, false)
  f::Core.Compiler.Const(var"#17#18"(), false)
  args::Tuple{Dense{typeof(identity),Array{Float32,2},Array{Float32,1}},Array{Float32,2}}
  @_4::Int64
  y::Any
  back::Zygote.var"#41#42"{_A} where _A

Body::Union{Nothing, Tuple}
1 ─ %1 = Core.tuple(f)::Core.Compiler.Const((var"#17#18"(),), false)
│   %2 = Core._apply_iterate(Base.iterate, Zygote.pullback, %1, args)::Tuple{Any,Zygote.var"#41#42"{_A} where _A}
│   %3 = Base.indexed_iterate(%2, 1)::Core.Compiler.PartialStruct(Tuple{Any,Int64}, Any[Any, Core.Compiler.Const(2, false)])
│        (y = Core.getfield(%3, 1))
│        (@_4 = Core.getfield(%3, 2))
│   %6 = Base.indexed_iterate(%2, 2, @_4::Core.Compiler.Const(2, false))::Core.Compiler.PartialStruct(Tuple{Zygote.var"#41#42"{_A} where _A,Int64}, Any[Zygote.var"#41#42"{_A} where _A, Core.Compiler.Const(3, false)])
│        (back = Core.getfield(%6, 1))
│   %8 = Zygote.sensitivity(y)::Any
│   %9 = (back)(%8)::Union{Nothing, Tuple}
└──      return %9
```
vs tagged
```
julia> @code_warntype gradient((m,x) -> sum(m(x)), d, r)
Variables
  #self#::Core.Compiler.Const(Zygote.gradient, false)
  f::Core.Compiler.Const(var"#13#14"(), false)
  args::Tuple{Dense{typeof(identity),Array{Float32,2},Array{Float32,1}},Array{Float32,2}}
  @_4::Int64
  y::Float32
  back::Zygote.var"#41#42"{typeof(∂(#13))}

Body::Tuple{Any,Array{Float32,2}}
1 ─ %1 = Core.tuple(f)::Core.Compiler.Const((var"#13#14"(),), false)
│   %2 = Core._apply_iterate(Base.iterate, Zygote.pullback, %1, args)::Tuple{Float32,Zygote.var"#41#42"{typeof(∂(#13))}}
│   %3 = Base.indexed_iterate(%2, 1)::Core.Compiler.PartialStruct(Tuple{Float32,Int64}, Any[Float32, Core.Compiler.Const(2, false)])
│        (y = Core.getfield(%3, 1))
│        (@_4 = Core.getfield(%3, 2))
│   %6 = Base.indexed_iterate(%2, 2, @_4::Core.Compiler.Const(2, false))::Core.Compiler.PartialStruct(Tuple{Zygote.var"#41#42"{typeof(∂(#13))},Int64}, Any[Zygote.var"#41#42"{typeof(∂(#13))}, Core.Compiler.Const(3, false)])
│        (back = Core.getfield(%6, 1))
│   %8 = Zygote.sensitivity(y)::Core.Compiler.Const(1.0f0, false)
│   %9 = (back)(%8)::Tuple{Any,Array{Float32,2}}
└──      return %9
```
The forward-only case is simpler, Julia is not confused by the branch. It's that `@warn ... maxlog=1` still runs logging code to check this counter. And that's fine, the point is that you should't do that. (The only reason not to make it an error is just to make hacking together the first version which runs at all less painful.)

DhairyaLGandhi:
Anyway this warning should be omitted and the performance tips can speak about type stability, even though that already happens.

CarloLucibello:
> Anyway this warning should be omitted and the performance tips can speak about type stability, even though that already happens.

I agree. Even if this specific problem it's not relevant in deep learning, where matrices involved are much bigger, we want to be performant in more general situation. For instance, SciML had to come up with its own Dense layers to squeeze out some performance with small sizes. 

It's a pity Zygote is not smart enough to optimize this stuff. I'll try to introduce a function barrier and see if it work, otherwise, I'll jest remove the warning

CarloLucibello:
Docs already discuss type instability, I wouldn't know what to add, so I should finished here.

CarloLucibello:
@DhairyaLGandhi this should be ready

CarloLucibello:
@DhairyaLGandhi you mentioned you wanted feedback from someone else here, do you have someone specific in mind?

DhairyaLGandhi:
@maleadt would you be able to clarify whether having removing the type checks could get in the way of cuda kernels, and if it would impact mixed precision that he has been working on.

maleadt:
What kind of GPU operations are we talking about? GPU kernels shouldn't be affected much; performance could be affected, but I agree with the being more explicit about input/output types (as this will be a requirement for mixed-precision support anyway). But other operations, e.g. from CUDNN, might be unavailable using Float64 inputs. I guess if GPU CI passes, all should be good?

CarloLucibello:
let's merge this

mcabbott (approved):

CarloLucibello:
@DhairyaLGandhi I think @mcabbott right's should be raised a bit here otherwise they are kind of useless

DhairyaLGandhi:
Thanks for the bump, let me take a final look at the changes here, I think mcabbott's permissions allow merging but let me verify

CarloLucibello:
His approval doesn't allow merging for me

DhairyaLGandhi (approved):

PR #1395, opened 2020-11-14
fix bias transpose conv

CarloLucibello:
Fix wrong dimension of conv transpose bias

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


mcabbott:
Some of the `bias=false`s in the docstring can be deleted -- they added in of #1391  just to avoid this bug.

PR #1397, opened 2020-11-16
Rework normalization layers

### LINKED ISSUE #802: "no method matching ∇batchnorm" on GPU
```Julia
using Flux
using Plots
using Random
using CuArrays
import NPZ; npz= NPZ

model = Chain(BatchNorm(69),Dense(69,512,NNlib.leakyrelu),
                BatchNorm(512),Dense(512,256,NNlib.leakyrelu),
                BatchNorm(256),Dense(256,168,NNlib.leakyrelu),
                BatchNorm(168),Dense(168,64,NNlib.leakyrelu),
                BatchNorm(64),Dense(64,32,NNlib.leakyrelu)) |> gpu

ps1 = [model.layers[i].W for i=2:2:length(model)]
ps2 = [model.layers[i].β for i=1:2:length(model)]
ps3 = [model.layers[i].γ for i=1:2:length(model)]
ps = vcat(ps1,ps2,ps3)

x = permutedims(npz.npzread("data.npy")[:,4:end]);
x = convert.(Float32,x)
c = npz.npzread("tempc.npy") |> gpu

println("data loaded")

loss(xs) = (sum(model(xs).-c).^2)/size(xs,2) +
                        sum(sum(.^(ps[i],2)) for i = 1:length(ps))

op = ADAM()

println("begin training")
in_gpu = cu(x)
Flux.train!(loss,ps,zip([in_gpu]),op)
```

Output:
```
ERROR: LoadError: MethodError: no method matching ∇batchnorm(::CuArray{Float32,1}, ::CuArray{Fl
oat32,1}, ::CuArray{Float32,2}, ::CuArray{Float64,2}, ::CuArray{Float32,1}, ::CuArray{Float32,1
}, ::Float32; cache=nothing, alpha=1, beta=0, eps=1.0f-5, training=true)                       
Closest candidates are:                                                                        
  ∇batchnorm(::CuArray{T<:Union{Float32, Float64},N} where N, ::CuArray{T<:Union{Float32, Float
64},N} where N, ::CuArray{T<:Union{Float32, Float64},2}, ::CuArray{T<:Union{Float32, Float64},2
}, ::CuArray{T<:Union{Float32, Float64},N} where N, ::CuArray{T<:Union{Float32, Float64},N} whe
re N, ::Any; cache, eps, alpha, beta, training) where T<:Union{Float32, Float64} at /home/nsain
i/.julia/packages/Flux/qXNjB/src/cuda/cudnn.jl:129                                             
  ∇batchnorm(::CuArray{T<:Union{Float32, Float64},N} where N, ::CuArray{T<:Union{Float32, Float
64},N} where N, ::CuArray{T<:Union{Float32, Float64},N} where N, ::CuArray{T<:Union{Float32, Fl
oat64},N} where N, ::CuArray{T<:Union{Float32, Float64},N} where N, ::CuArray{T<:Union{Float32,
 Float64},N} where N, ::Any; cache, eps, alpha, beta, training) where T<:Union{Float32, Float64
} at /home/nsaini/.julia/packages/Flux/qXNjB/src/cuda/cudnn.jl:139
```

This code is working fine on the CPU but not on the GPU. Apart from this, when I remove the batchnorm layers from the model, it gives another error `ERROR: LoadError: ArgumentError: cannot take the CPU address of a CuArray{Float32,2} `
The code for model without batchnorm is:
```Julia
using Flux
using Plots
using Random
using CuArrays
import NPZ; npz= NPZ

model = Chain(Dense(69,512,NNlib.leakyrelu),
                 Dense(512,256,NNlib.leakyrelu),
                 Dense(256,168,NNlib.leakyrelu),
                 Dense(168,64,NNlib.leakyrelu),
                 Dense(64,32,NNlib.leakyrelu)) |> gpu

ps1 = [model.layers[i].W for i=1:length(model)]
ps = ps1

x = permutedims(npz.npzread("data.npy")[:,4:end]);
x = convert.(Float32,x)
c = npz.npzread("tempc.npy") |> gpu

println("data loaded")

loss(xs) = (sum(model(xs).-c).^2)/size(xs,2) +
                        sum(sum(.^(ps[i],2)) for i = 1:length(ps))

op = ADAM()

println("begin training")
in_gpu = cu(x)
Flux.train!(loss,ps,zip([in_gpu]),op)
```

### LINKED ISSUE #1195: InstanceNorm throws a scalar getindex disallowed error on GPU
When using InstanceNorm on a 3D CNN, on the GPU with 

```
    CuArrays.allowscalar(false)
```
I get the following stack trace (abbreviated)

```
ERROR: LoadError: scalar getindex is disallowed
Stacktrace:
 [1] assertscalar(::String) at /home/###/.julia/packages/GPUArrays/HGtNV/src/host/indexing.jl:41
 [2] getindex at /home/###/.julia/packages/GPUArrays/HGtNV/src/host/indexing.jl:96 [inlined]
 [3] _getindex at ./abstractarray.jl:1003 [inlined]
 [4] getindex at ./abstractarray.jl:980 [inlined]
 [5] _broadcast_getindex at ./broadcast.jl:597 [inlined]
 [6] _getindex at ./broadcast.jl:628 [inlined]
 [7] _broadcast_getindex at ./broadcast.jl:603 [inlined]
 [8] getindex at ./broadcast.jl:564 [inlined]
 [9] copy at ./broadcast.jl:854 [inlined]
 [10] materialize(::Base.Broadcast.Broadcasted{CuArrays.CuArrayStyle{2},Nothing,Zygote.var"#1774#1777"{CuArrays.var"#68#69"{Float32}},Tuple{CuArrays.CuArray{Float32,2,CuArrays.CuArray{Float32,5,Nothing}}}}) at ./broadcast.jl:820
 [11] broadcast_forward(::Function, ::CuArrays.CuArray{Float32,2,CuArrays.CuArray{Float32,5,Nothing}}) at /home/andevellicus/.julia/packages/Zygote/YeCEW/src/lib/broadcast.jl:181
 [12] adjoint at /home/###/.julia/packages/Zygote/YeCEW/src/lib/broadcast.jl:197 [inlined]
 [13] _pullback at /home/###/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:47 [inlined]
 [14] adjoint at /home/###/.julia/packages/Zygote/YeCEW/src/lib/lib.jl:179 [inlined]
 [15] _pullback at /home/###/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:47 [inlined]
 [16] broadcasted at ./broadcast.jl:1232 [inlined]
 [17] InstanceNorm at /home/###/.julia/packages/Flux/Fj3bt/src/layers/normalise.jl:292 [inlined]
```
BatchNorm works without a problem.

Flux 0.10.4
Julia 1.4.2
CuArrays 2.2.0

### LINKED ISSUE #1247: Error with GroupNorm on GPU
When I try this:

```julia
using Statistics
using Flux
using Flux: flatten, binarycrossentropy
using CuArrays
CuArrays.allowscalar(false)

imsize = (32,32)
group = 4

model = Chain(Conv((3,3), 3=>32, pad=(1,1)),
              GroupNorm(32, group, relu),
              MaxPool((2,2)),
              flatten,
              Dense(Int(prod(imsize) / 4 * 32), 128, relu),
              Dense(128, 1, sigmoid))

x1 = Float32.(randn(32, 32, 3, 8));
y1 = Float32.([1 0 0 0 0 0 0 0]);

bce(ŷ, y) = mean(binarycrossentropy.(ŷ, y))
loss(x, y) = bce(model(x), y)

opt = Descent();
ps = params(model);
Flux.train!(loss, ps, [(x1,y1)], opt)
````
It seems OK on CPU, but when I move data and model on GPU:

```julia
model = gpu(model);
opt = Descent();
ps = params(model);
Flux.train!(loss, ps, gpu.([(x1,y1)]), opt)

ERROR: scalar getindex is disallowed
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] assertscalar(::String) at /home/afertin/.julia/packages/GPUArrays/JqOUg/src/host/indexing.jl:41
 [3] getindex at /home/afertin/.julia/packages/GPUArrays/JqOUg/src/host/indexing.jl:96 [inlined]
 [4] _getindex at ./abstractarray.jl:1003 [inlined]
 [5] getindex at ./abstractarray.jl:980 [inlined]
 [6] _broadcast_getindex at ./broadcast.jl:597 [inlined]
 [7] _getindex at ./broadcast.jl:628 [inlined]
 [8] _broadcast_getindex at ./broadcast.jl:603 [inlined]
 [9] getindex at ./broadcast.jl:564 [inlined]
 [10] copy at ./broadcast.jl:854 [inlined]
 [11] materialize(::Base.Broadcast.Broadcasted{CuArrays.CuArrayStyle{2},Nothing,Zygote.var"#1161#1164"{CuArrays.var"#68#69"{Float32}},Tuple{CuArray{Float32,2,CuArray{Float32,5,Nothing}}}}) at ./broadcast.jl:820
 [12] broadcast_forward(::Function, ::CuArray{Float32,2,CuArray{Float32,5,Nothing}}) at /home/afertin/.julia/packages/Zygote/1GXzF/src/lib/broadcast.jl:181
 [13] adjoint at /home/afertin/.julia/packages/Zygote/1GXzF/src/lib/broadcast.jl:197 [inlined]
 [14] _pullback at /home/afertin/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:47 [inlined]
 [15] adjoint at /home/afertin/.julia/packages/Zygote/1GXzF/src/lib/lib.jl:179 [inlined]
 [16] _pullback at /home/afertin/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:47 [inlined]
 [17] broadcasted at ./broadcast.jl:1232 [inlined]
 [18] _pullback(::Zygote.Context, ::typeof(Base.Broadcast.broadcasted), ::Type{Float32}, ::CuArray{Float32,2,CuArray{Float32,5,Nothing}}) at /home/afertin/.julia/packages/Zygote/1GXzF/src/compiler/interface2.jl:0
 [19] GroupNorm at /home/afertin/.julia/packages/Flux/Fj3bt/src/layers/normalise.jl:394 [inlined]
 [20] _pullback(::Zygote.Context, ::GroupNorm{typeof(relu),CuArray{Float32,1,Nothing},CuArray{Float32,2,Nothing},Float32,Int64}, ::CuArray{Float32,4,Nothing}) at /home/afertin/.julia/packages/Zygote/1GXzF/src/compiler/interface2.jl:0
 [21] applychain at /home/afertin/.julia/packages/Flux/Fj3bt/src/layers/basic.jl:36 [inlined]
 [22] _pullback(::Zygote.Context, ::typeof(Flux.applychain), ::Tuple{GroupNorm{typeof(relu),CuArray{Float32,1,Nothing},CuArray{Float32,2,Nothing},Float32,Int64},MaxPool{2,4},typeof(flatten),Dense{typeof(relu),CuArray{Float32,2,Nothing},CuArray{Float32,1,Nothing}},Dense{typeof(σ),CuArray{Float32,2,Nothing},CuArray{Float32,1,Nothing}}}, ::CuArray{Float32,4,Nothing}) at /home/afertin/.julia/packages/Zygote/1GXzF/src/compiler/interface2.jl:0
 [23] applychain at /home/afertin/.julia/packages/Flux/Fj3bt/src/layers/basic.jl:36 [inlined]
 [24] _pullback(::Zygote.Context, ::typeof(Flux.applychain), ::Tuple{Conv{2,2,typeof(identity),CuArray{Float32,4,Nothing},CuArray{Float32,1,Nothing}},GroupNorm{typeof(relu),CuArray{Float32,1,Nothing},CuArray{Float32,2,Nothing},Float32,Int64},MaxPool{2,4},typeof(flatten),Dense{typeof(relu),CuArray{Float32,2,Nothing},CuArray{Float32,1,Nothing}},Dense{typeof(σ),CuArray{Float32,2,Nothing},CuArray{Float32,1,Nothing}}}, ::CuArray{Float32,4,Nothing}) at /home/afertin/.julia/packages/Zygote/1GXzF/src/compiler/interface2.jl:0
 [25] Chain at /home/afertin/.julia/packages/Flux/Fj3bt/src/layers/basic.jl:38 [inlined]
 [26] _pullback(::Zygote.Context, ::Chain{Tuple{Conv{2,2,typeof(identity),CuArray{Float32,4,Nothing},CuArray{Float32,1,Nothing}},GroupNorm{typeof(relu),CuArray{Float32,1,Nothing},CuArray{Float32,2,Nothing},Float32,Int64},MaxPool{2,4},typeof(flatten),Dense{typeof(relu),CuArray{Float32,2,Nothing},CuArray{Float32,1,Nothing}},Dense{typeof(σ),CuArray{Float32,2,Nothing},CuArray{Float32,1,Nothing}}}}, ::CuArray{Float32,4,Nothing}) at /home/afertin/.julia/packages/Zygote/1GXzF/src/compiler/interface2.jl:0
 [27] loss at ./REPL[12]:1 [inlined]
 [28] _pullback(::Zygote.Context, ::typeof(loss), ::CuArray{Float32,4,Nothing}, ::CuArray{Float32,2,Nothing}) at /home/afertin/.julia/packages/Zygote/1GXzF/src/compiler/interface2.jl:0
 [29] adjoint at /home/afertin/.julia/packages/Zygote/1GXzF/src/lib/lib.jl:179 [inlined]
 [30] _pullback at /home/afertin/.julia/packages/ZygoteRules/6nssF/src/adjoint.jl:47 [inlined]
 [31] #17 at /home/afertin/.julia/packages/Flux/Fj3bt/src/optimise/train.jl:89 [inlined]
 [32] _pullback(::Zygote.Context, ::Flux.Optimise.var"#17#25"{typeof(loss),Tuple{CuArray{Float32,4,Nothing},CuArray{Float32,2,Nothing}}}) at /home/afertin/.julia/packages/Zygote/1GXzF/src/compiler/interface2.jl:0
 [33] pullback(::Function, ::Zygote.Params) at /home/afertin/.julia/packages/Zygote/1GXzF/src/compiler/interface.jl:172
 [34] gradient(::Function, ::Zygote.Params) at /home/afertin/.julia/packages/Zygote/1GXzF/src/compiler/interface.jl:53
 [35] macro expansion at /home/afertin/.julia/packages/Flux/Fj3bt/src/optimise/train.jl:88 [inlined]
 [36] macro expansion at /home/afertin/.julia/packages/Juno/tLMZd/src/progress.jl:134 [inlined]
 [37] train!(::typeof(loss), ::Zygote.Params, ::Array{Tuple{CuArray{Float32,4,Nothing},CuArray{Float32,2,Nothing}},1}, ::Descent; cb::Flux.Optimise.var"#18#26") at /home/afertin/.julia/packages/Flux/Fj3bt/src/optimise/train.jl:81
 [38] train!(::Function, ::Zygote.Params, ::Array{Tuple{CuArray{Float32,4,Nothing},CuArray{Float32,2,Nothing}},1}, ::Descent) at /home/afertin/.julia/packages/Flux/Fj3bt/src/optimise/train.jl:79
 [39] top-level scope at REPL[19]:1
```

### LINKED ISSUE #1280: Error with BatchNorm/InstanceNorm after Conv1D on GPU
I tried to use BatchNorm and InstanceNorm after Conv1D. It works on cpu but fails on gpu.

(BatchNorm works fine on gpu after Conv2D)

Here is a minimal example reproducing the bug.

```julia
using Flux

device = gpu
xin = rand(Float32, 10, 2, 4) |> device

model = Chain(
    Conv((3,), 2=>2, pad=1),
    BatchNorm(2),
) |> device

loss(x) = Flux.mse(model(x), x)

ps = params(model)

loss, back = Flux.pullback(ps) do
    loss(xin)
end
```

Here is the error message

```julia
MethodError: no method matching Float32(::ForwardDiff.Dual{Nothing,Float32,1})
Closest candidates are:
  Float32(::Real, !Matched::RoundingMode) where T<:AbstractFloat at rounding.jl:200
  Float32(::T) where T<:Number at boot.jl:715
  Float32(!Matched::Int8) at float.jl:60
  ...
in top-level scope at Conv1DNorm.jl:19
in pullback at Zygote\iFibI\src\compiler\interface.jl:172
in _pullback at Zygote\iFibI\src\compiler\interface2.jl
in #9 at Conv1DNorm.jl:20 
in _pullback at Zygote\iFibI\src\compiler\interface2.jl
in loss at Conv1DNorm.jl:13 
in _pullback at Zygote\iFibI\src\compiler\interface2.jl
in Chain at Flux\IjMZL\src\layers\basic.jl:38 
in _pullback at Zygote\iFibI\src\compiler\interface2.jl
in applychain at Flux\IjMZL\src\layers\basic.jl:36 
in _pullback at Zygote\iFibI\src\compiler\interface2.jl
in applychain at Flux\IjMZL\src\layers\basic.jl:36 
in _pullback at Zygote\iFibI\src\compiler\interface2.jl
in BatchNorm at Flux\IjMZL\src\layers\normalise.jl:214 
in broadcasted at base\broadcast.jl:1232 
in _pullback at ZygoteRules\6nssF\src\adjoint.jl:47 
in adjoint at Zygote\iFibI\src\lib\lib.jl:175 
in _pullback at ZygoteRules\6nssF\src\adjoint.jl:47 
in adjoint at Zygote\iFibI\src\lib\broadcast.jl:193 
in broadcast_forward at Zygote\iFibI\src\lib\broadcast.jl:181
in materialize at base\broadcast.jl:820
in copy at base\broadcast.jl:854 
in getindex at base\broadcast.jl:564 
in _broadcast_getindex at base\broadcast.jl:604 
in _broadcast_getindex_evalf at base\broadcast.jl:631 
in  at Zygote\iFibI\src\lib\broadcast.jl:175
in  at CUDA\sjcZt\src\broadcast.jl:21
```

### LINKED ISSUE #1308: Misleading InstanceNorm documentation?
`InstanceNorm` [documentation](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.InstanceNorm) says:

> Given an array with `N` dimensions, call the `N-1`th the channel dimension. (For
> a batch of feature vectors this is just the data dimension, for `WHCN` images
> it's the usual channel dimension.)

following with an example:

```julia
m = Chain(
  Dense(28^2, 64),
  InstanceNorm(64, relu),
  Dense(64, 10),
  InstanceNorm(10),
  softmax)
```

as far as I understand, the docs say that the layer can in principle be used on a batch of feature vectors (with shape `DxN`). However, the implementation actually does not support this behavior and throws an exception if it encounters 2D data: 

```julia
 ndims(x) > 2 ||
    error("InstanceNorm requires at least 3 dimensions. With 2 dimensions an array of zeros would be returned")
```

This makes it impossible to reproduce the example. So if I interpret it corretly, the docs and the implementation disagree.

### LINKED ISSUE #1429: BatchNorm alters its sliding mean/standard deviation parameters even in testmode if Zygote is called
After training a classification model, I am attempting to perform Google Deepdream-like techniques with it, i.e. altering a new input `x` to maximize outputs from selected layers of the model.

I am using Zygote in the following manner to achieve the latter part of this:

```
ps = Flux.params(x)

loss, back = @sync Zygote.pullback(() -> mod_obj(m[1:layer](x)), ps)

gs = back(one(loss))

x = x .+ gs[x]
```

I noticed, however, that performing this process seemed to lead to instability in the classifier, despite having set the model to testmode. The classifier would begin to produce wildly inaccurate results after running this code a couple thousand iterations, but could be fixed by training the network for a very short duration once again. I began to suspect that the presence of `BatchNorm` in my model was partially responsible for this.

I have verified that the `BatchNorm` layers in my model have their `active` parameter set to off before and after the autodiff. Despite that, however, `μ` and `σ²` change every iteration. This seems to be the cause of my classifier becoming inaccurate after running the above code.

On the other hand, when I don't call autodiff via Zygote or Flux, this process does not trigger if I simply call `m(x)`.

### LINKED ISSUE #1430: BatchNorm on CUDA accepts improper channel size argument and "works" in a possibly ill-defined way. Proper errors on CPU
This is partially user error, partially a failure(?) to warn/error on user error. This however leads to a case where a model "can be" trained on the GPU but not be brought back properly to a CPU. (Fixing this could likewise cause backwards compatibility issues with improperly-defined models).

It also might be the type of problem that might need to be addressed somewhere else like `CUDA.jl`.

- Windows 10
- CUDA toolkit Version 11.1
- CUDA.jl v2.3.0
- Flux.jl v0.11.2

I was training a CNN model on audio:

```
m = Chain(
            Conv((15,1), 2 => 16, identity, dilation = (9, 1), pad = SamePad()),
            Dropout(.25),
            MeanPool((32, 1)),
            BatchNorm(128, swish),
            Conv((15,1), 16 => 64, identity, dilation = (9, 1), pad = SamePad()),
            Dropout(.35),
            MeanPool((8, 1)),
            BatchNorm(128, swish),
            Conv((9,1), 64 => 128, identity, dilation = (7, 1), pad = SamePad()),
            Dropout(.45),
            MeanPool((4, 1)),
            BatchNorm(128, swish),
            Conv((1,1), 128 => 3, identity, dilation = (5, 1)),
            Dropout(.5),
            AdaptiveMeanPool((1, 1)), # guaranteed dims: (1, 1, 3, B)
            flatten,
            logsoftmax
        ) |> gpu
```

I had mistakenly set the first two BatchNorms to 128 channels, when they should have been at 16 and 64 respectively.

I started noticing that my model performed differently based on whether `trainmode` or `testmode` was on, in a way that could not be well-explained by factors like Dropout layers. I'm not yet sure that these issues are caused by this, although it seems to be the case.

I eventually tried switching the model to my CPU to see if it had the same issues. On the CPU, I received errors due to the channel mismatch on the first BatchNorm.

I'm not entirely sure yet that this actually leads to strangely-defined behavior; I do recall that the latter channels' parameters were actually all `0` on the first BatchNorm. I simply know that this is an example of a model that functions on CUDA, but not on the CPU. I will do more testing later.

CarloLucibello:
This goes in the direction of creating a powerful and consistent experience for the normalization layers. 
- add an activation function to LayerNorm, consistently with other normalization layers
- generalize LayerNorm and Diagonal to handle multiple dimensions
- add the bool keyword `affine` to LayerNorm, GroupNorm, InstanceNorm, and BatchNorm to activate/deactivate a learnable shift and scaling
- add the keyword `track_stats` to InstanceNorm and BatchNorm to activate/deactivate running time mean and variance tracking, to be used in the evaluation phase.
- use pytorch's defaults for `track_stats`  and `affine`, since AFAIK they correspond to common usage and best practice:
   - For InstanceNorm, `affine=false` and `track_stats=false`. This breaks the current behaviour (both `true`)
   - For BatchNorm, `affine=true` and `track_stats=true`, as we are currently doing
   - For GroupNorm, `affine=true` and `track_stats=false` 
- unifies InstanceNorm, GroupNorm, and BatchNorm logics
- perf improvement for InstanceNorm that was previously using repetition instead of broadcasting
- Fix InstanceNorm previously not working on gpu (Fix #1195)
- Fix GroupNorm previously not working on gpu (Fix #1247)
- Fix #1280, fix #1308, fix #1429, fix #1430, fix #802

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


CarloLucibello:
@DhairyaLGandhi bump

src/layers/normalise.jl
```diff
@@ -110,46 +110,111 @@ function (a::AlphaDropout)(x)
...
+
+If `affine=true` also applies a learnable shift and rescaling
+as in the [`Diagonal`](@ref) layer.
+
+
+Se also [`BatchNorm`](@ref), [`InstanceNorm`](@ref), [`GroupNorm`](@ref), and [`normalise`](@ref).
 """
-struct LayerNorm{T}
-  diag::Diagonal{T}
+struct LayerNorm
```
DhairyaLGandhi, line 134: Should be parameterised properly

src/layers/normalise.jl
```diff
@@ -110,46 +110,111 @@ function (a::AlphaDropout)(x)
...
+    σ² = mean((x .- μ).^2; dims)
+    if l.track_stats
+      # update moving mean/std
+      mtm = l.momentum
+      m = prod(size(x)) ÷ size(x, N-1)  # needed for computing corrected var
+      l.μ = (1-mtm) .* l.μ .+ mtm .* vec(N ∈ dims ? μ : mean(μ, dims=N))
+      l.σ² = (1-mtm) .* l.σ² .+ mtm .* (m/(m-1f0)) .* vec(N ∈ dims ? σ² : mean(σ², dims=N))
+    end
+  end
+  if hasaffine(l)
```
DhairyaLGandhi, line 187: We should try to avoid having these placeholder kwargs, seems like bad design to have the if statements. Why do we do it like so?

src/layers/stateless.jl
```diff
@@ -10,13 +10,13 @@ function flatten(x::AbstractArray)
...
 """
-    normalise(x; dims=ndims(x), ϵ=1e-5)
+    normalise(x; dims=ndims(x), ϵ=1f-5)
 
 Normalise `x` to mean 0 and standard deviation 1 across the dimension(s) given by `dims`.
 Per default, `dims` is the last dimension. 
 `ϵ` is a small additive factor added to the denominator for numerical stability.
 """
-function normalise(x::AbstractArray; dims=ndims(x), ϵ=ofeltype(x, 1e-5))
+function normalise(x::AbstractArray; dims=ndims(x), ϵ=ofeltype(x, 1f-5))
```
DhairyaLGandhi, line 19: Shouldn't matter since we are converting anyway?

DhairyaLGandhi (commented):
In general it's alright to have extra steps available for different norm layers but not too have to check everything everytime for every layer I think. Otherwise we end up with one large function which should be multiple smaller functions which are combined appropriately for the layers

src/layers/normalise.jl
CarloLucibello: how should we do it?

CarloLucibello:
> In general it's alright to have extra steps available for different norm layers but not too have to check everything everytime for every layer I think. Otherwise, we end up with one large function which should be multiple smaller functions which are combined appropriately for the layers

Every layer has `affine`, `track_stats`, and `isactive` options. While the `isactive` check has to be performed at runtime, we could maybe rely on type parametrization and dispatch for `affine` and `track_stats` (if this is what you mean, and assuming we don't want to change e.g. track_stats at some point in the execution). Honestly, I don't think removing a couple of `if` is worth the extra complexity involved, nor I believe there would be any performance benefit.   Can you mock up the alternative so I can benchmark it?

johnnychen94:
> fix InstanceNorm previously not working on gpu

This will fix #1195 if I get it right?

src/cuda/cudnn.jl
```diff
@@ -1,7 +1,20 @@
...
-  BN.λ.(batchnorm(BN.γ, BN.β, x, BN.μ, BN.σ², BN.momentum; cache = cache, alpha = 1, beta = 0, eps = BN.ϵ, training = Flux.istraining()))
+function (BN::Flux.BatchNorm)(x::Union{CuArray{T,2},CuArray{T,4},CuArray{T,5}},
+                              cache=nothing) where T<:Union{Float32, Float64}
+  
+  @assert BN.affine "BatchNorm: only affine=true supported on gpu"
+  @assert BN.track_stats "BatchNorm: only track_stats=true supported on gpu"
+  
+  return BN.σ.(batchnorm(BN.γ, BN.β, x, BN.μ, BN.σ², BN.momentum; 
+                  cache=cache, alpha=1, beta=0, eps=BN.ϵ, 
+                  training=_isactive(BN)))
```
ToucheSir, line 11: While working through https://github.com/FluxML/Flux.jl/issues/1429, I found that this path is not tested at all. Should we have an equivalent to https://github.com/FluxML/Flux.jl/blob/master/test/layers/normalisation.jl#L16 for the CUDA layer implementations? I guess ideally this functionality would be in NNlib so that the high level layer call isn't being overriden.

CarloLucibello:
> This will fix #1195 if I get it right?

yes, this is tested in test/cuda/layers.jl now that InstanceNorm in not on the list of broken layers

test/cuda/layers.jl
```diff
@@ -15,47 +15,20 @@ end
...
-        gs_cpu = back_cpu(1f0)
-
-        x_gpu = gpu(x_cpu)
-        l_gpu = l_cpu |> gpu
-        ps_gpu = Flux.params(l_gpu)
-
-        if l_gpu isa BROKEN_LAYERS
-          @test_broken gradient(() -> sum(l_gpu(x_gpu)), ps_gpu) isa Flux.Zygote.Grads
+        if l_cpu isa BROKEN_LAYERS
+          @test_broken gpu_autodiff_test(l_cpu, x_cpu,
```
DhairyaLGandhi, line 27: Best to do the tests explicitly here, a little extra code won't hurt us here

src/cuda/cudnn.jl
CarloLucibello: I added a test now. If I try to compute the gradient  on gpu when testmode!(m, true) I get an error, but I think it is a CUDA.jl issue

test/cuda/layers.jl
CarloLucibello: i use this function in multiple places

CarloLucibello:
I've also updated GroupNorm and I've done an extensive search of the issues fixed by this PR

test/cuda/layers.jl
DhairyaLGandhi: I don't see how that addresses the need to be explicit here.

DhairyaLGandhi (changes requested):
I'm still a bit uncomfortable with having `affine` switches in the layers, do we have a better way around this? Perhaps the users are better suited to add that as a layer in the model

test/cuda/layers.jl
CarloLucibello: made it explicit

CarloLucibello:
> I'm still a bit uncomfortable with having affine switches in the layers, do we have a better way around this? Perhaps the users are better suited to add that as a layer in the model

I don't see how to do it in a way that doesn't involve a branch somewhere. And we really don't want the user to have to add manually an Affine layer after BatchNorm

Pytorch has the same interface, it can't be that bad if we do the same.

CarloLucibello:
I did some benchmarking using the script below.
The script was executed 3 times on the current PR
and another 3 times on similar code where I removed the track_stat and affine branching
(so I only keep the path where both are true). 

Fortunately, there is no appreciable performance difference for any system size. 

SCRIPT:
```
using Flux
using BenchmarkTools
using Random 

function test_BN(shape)
    Random.seed!(17)
    x = rand(Float32, shape)
    bn = BatchNorm(shape[end-1])
    @btime $bn($x)
end

for shape in [  (10,10),
                (100,100),
                (1000,1000),]
    test_BN(shape)
end
```
OUTPUT¨:
```julia
# WITHOUT track_stats and affine branching

  4.427 μs (41 allocations: 2.19 KiB)
  38.264 μs (42 allocations: 40.84 KiB)
  3.552 ms (42 allocations: 3.82 MiB)

  4.402 μs (41 allocations: 2.19 KiB)
  38.231 μs (42 allocations: 40.84 KiB)
  3.336 ms (42 allocations: 3.82 MiB)

  4.407 μs (41 allocations: 2.19 KiB)
  37.572 μs (42 allocations: 40.84 KiB)
  3.476 ms (42 allocations: 3.82 MiB)


# WITH track_stats and affine branching

  4.538 μs (41 allocations: 2.19 KiB)
  38.196 μs (42 allocations: 40.84 KiB)
  3.354 ms (42 allocations: 3.82 MiB)

  4.472 μs (41 allocations: 2.19 KiB)
  40.386 μs (42 allocations: 40.84 KiB)
  3.549 ms (42 allocations: 3.82 MiB)

  4.357 μs (41 allocations: 2.19 KiB)
  38.216 μs (42 allocations: 40.84 KiB)
  3.580 ms (42 allocations: 3.82 MiB)
```

CarloLucibello:
@DhairyaLGandhi bump

src/layers/normalise.jl
```diff
@@ -104,52 +104,117 @@ function (a::AlphaDropout)(x)
...
+  else  # trainmode or testmode without tracked stats
+    μ = mean(x; dims=reduce_dims)
+    σ² = mean((x .- μ).^2; dims=reduce_dims)
+    if l.track_stats
+      ## update moving mean/std
+      Zygote.ignore() do
+        mtm = l.momentum
+        m = prod(size(x, i) for i in reduce_dims)  # needed for computing corrected var
+        μnew = vec(N ∈ reduce_dims ? μ : mean(μ, dims=N))
+        σ²new = vec(N ∈ reduce_dims ? σ² : mean(σ², dims=N))
```
DhairyaLGandhi, line 181: Is the condition check here because of unifying the different norm layers in one kernel?

src/layers/normalise.jl
```diff
@@ -104,52 +104,117 @@ function (a::AlphaDropout)(x)
...
+  hasaffine(l) || print(io, ", affine=false")
+  print(io, ")")
+end
+
+# For InstanceNorm, GroupNorm, and BatchNorm.
+# Compute the statistics on the slices specified by reduce_dims.
+# reduce_dims=[1,...,N-2,N] for BatchNorm
+# reduce_dims=[1,...,N-2] for InstanceNorm and GroupNorm
+function _norm_layer_forward(l, x; reduce_dims, affine_shape)
+  N = ndims(x)
```
DhairyaLGandhi, line 167: Take from the type signatures?

DhairyaLGandhi:
Could you also check the backwards pass?

Having an affine layer seems fine to me, but I'm open to thoughts around construction

As in other test cases on the gpu, I'd much prefer if we didn't have a separate function to go to. Using it to test layers is an extra function to track errors into. We should have the tests be very clear about what they are testing, and best to have that be clear at the site where the test is performed.

src/layers/normalise.jl
CarloLucibello: will do

CarloLucibello:
> Could you also check the backwards pass?

will do

> Having an affine layer seems fine to me, but I'm open to thoughts around construction

I'm strongly against having a separate affine layer, people don't have to insert an Affine every time they insert a BatchNorm, it's extremely annoying, this is why the default for BatchNorm is `affine=true`. Let's just follow pytorch's approach here, no harm to it

> As in other test cases on the gpu, I'd much prefer if we didn't have a separate function to go to. Using it to test layers is an extra function to track errors into. We should have the tests be very clear about what they are testing, and best to have that be clear at the site where the test is performed.

having a centralized testing function assures exhaustive and consistent testing, and makes very clear what is being tested, since it is the same thing everywhere. Having heterogeneous and likely less extensive "manual testing" is strictly worse I think, for each set of tests you have to think about what you are capturing and what you are  missing. Going through an extra function call doesn't seem a problem at all. ChainRules provides a whole package of rule testing functions, ChainRulesTestUtils, we can have a couple of testing functions ourselves (in fact we do in Zygote and now  in NNlib as well).

DhairyaLGandhi:
Where you have to test consistent output, sure, but when you have to generic tests like we do with gpu layers, clarity is better imo. Thinking about the missing bits is good when you have to test for changes you have made to the core kernels like here.

src/layers/normalise.jl
CarloLucibello: yes

CarloLucibello:
I tested the pullback as well. The results is the following:
- master, this PR, and this PR w/o branching have same performance for `n=1000`
- master's pullback is the slowest for `n=10` and `n=100`
- removing branching we gain ~8% at `n=10` and ~3% at `n=100` in the pullback compared to this PR
- master, PR, and PR w/o branching's pullbacks do a very large number of allocations, something we should investigate (in the future)

My opinion is that
```julia
using Flux
using BenchmarkTools
using Random 
using Zygote: pullback

function test_BN(shape)
    Random.seed!(17)
    println("# Shape $shape")
    x = rand(Float32, shape)
    bn = BatchNorm(shape[end-1])
    # println("### forward")
    # @btime $bn($x)
    println("forward in gradient context")
    @btime pullback(x -> sum(sin.($bn(x))), $x)
    y, back = pullback(x -> sum(sin.(bn(x))), x)
    println("pullback")
    @btime $back(1f0)
end

for shape in [  (10,10),
                (100,100),
                (1000,1000),]
    test_BN(shape)
end
```
```julia
##### Master 

# Shape (10, 10)
forward gradient context
  35.807 μs (508 allocations: 32.97 KiB)
pullback
  250.860 μs (714 allocations: 23.14 KiB)
# Shape (100, 100)
forward gradient context
  207.208 μs (877 allocations: 431.64 KiB)
pullback
  715.362 μs (30697 allocations: 999.34 KiB)
# Shape (1000, 1000)
forward gradient context
  19.116 ms (5457 allocations: 38.32 MiB)
pullback
  54.512 ms (3003397 allocations: 95.48 MiB)

##### This PR

 # Shape (10, 10)
forward in gradient context
  27.025 μs (400 allocations: 29.41 KiB)
pullback
  216.878 μs (708 allocations: 23.17 KiB)
# Shape (100, 100)
forward in gradient context
  192.794 μs (409 allocations: 419.22 KiB)
pullback
  681.668 μs (30692 allocations: 1.01 MiB)
# Shape (1000, 1000)
forward in gradient context
  19.810 ms (409 allocations: 38.21 MiB)
pullback
  53.283 ms (3003392 allocations: 99.29 MiB)

### this PR w/o affine and track_stats branching

# Shape (10, 10)
forward in gradient context
  27.070 μs (365 allocations: 27.00 KiB)
pullback
  198.775 μs (693 allocations: 22.67 KiB)
# Shape (100, 100)
forward in gradient context
  194.250 μs (374 allocations: 416.81 KiB)
pullback
  661.045 μs (30677 allocations: 1.01 MiB)
# Shape (1000, 1000)
forward in gradient context
  20.032 ms (374 allocations: 38.20 MiB)
pullback
  54.227 ms (3003377 allocations: 99.29 MiB)
```

src/layers/normalise.jl
DhairyaLGandhi: This should become a kernel dispatch based on the type of layer then. That way it can compile away and inline. Currently it's a runtime check so the compiler can't eliminate a branch

DhairyaLGandhi:
I don't think zygote is much of a culprit here. Besides, we should add an adjoint for norm, and that would make it much faster

CarloLucibello:
> I don't think zygote is much of a culprit here. Besides, we should add an adjoint for norm, and that would make it much faster

What do you mean by an adjoint for `norm`? The whole `_norm_layer_forward` function? That seems quite painful to do

src/layers/normalise.jl
CarloLucibello: since I'm passing reduce_dims explicitly to the function and this condition depends on the layer only through reduce_dims, for semantic consistency I prefer not to dispatch on the type. In any case, it seems to be an irrelevant optimization to me

DhairyaLGandhi:
This is an implementation that we can use, and improve upon, it was written for the existing case, so it would need some modification

```julia
function _bn(x, λ, β, γ, μ, σ², ϵ)
  x̂ = (x .- μ) ./ sqrt.(σ² .+ ϵ)
  λ.(γ .* x̂ .+ β), x̂
end

Flux.Zygote.@adjoint function Flux.batchnorm(x::AbstractArray{T,M}, λ, β, γ, μ, σ², ϵ) where {T,M}
  # x̂ = (x .- μ) ./ sqrt.(σ² .+ ϵ)
  # y = λ.(γ .* x̂ .+ β)
  y, x̂ = _bn(x, λ, β, γ, μ, σ², ϵ)
  function batchnorm_adjoint(Δ)
    N = convert(Float32, size(Δ, 1))
    dx̂ = @avx Δ .* x
    ds = ntuple(i -> i == M - 1 ? i + 1 : i, M - 1) # (1,2,4)
    dβ = @avx reshape(sum(Δ, dims = ds), size(β, ndims(β) - 1)) # sum(Δ .* bn.β, dims = 2)
    dγ = @avx reshape(sum(x̂ .* Δ, dims = ds), size(γ, ndims(γ) - 1))
    dx = @avx (1.f0 / N) .* σ² .* (N .* dx̂ .- sum(dx̂, dims = 2) .- y .* sum(dx̂ .* y, dims = 2))
    (dx, nothing, dβ, dγ, nothing, nothing, -2.f0)
  end
  y, batchnorm_adjoint
end
```

DhairyaLGandhi:
But doing it like this was almost an order of magnitude faster, which seems worth it

CarloLucibello:
Definitely worth it. It is a super intense time for me at work though, maybe I will be able to work more on this in February. The optimization you suggested takes some time to implement for all layers and also requires separate methods for cuarrays if we want to use avx (which I think we should). So I suggest we just finish discussing the API (which I think is good as it is). This PR already has better performance than master, and most importantly it fixes many issues and adds many features. I can attempt to squeeze out more performance in a later PR if someone doesn't beat me to it.  I don't want this to hold v0.12 release.

CarloLucibello:
@DhairyaLGandhi bump, let's do this

DhairyaLGandhi:
Let me do a once over, I still strongly feel the conditionals should be made generic, at least so Julia can produce better code, and have the kwargs only as a convenience constructor over top.

CarloLucibello:
> Let me do a once over, I still strongly feel the conditionals should be made generic, at least so Julia can produce better code, and have the kwargs only as a convenience constructor over top.

I don't oppose this (although I think it will make things much more complicated than needed), I'm just saying that devising composable internals is something that can be done later and shouldn't stop us from merging this PR. Again, this PR fixes a lot of issues, adds a ton of features, is faster than current master. It's been ready for more than a month, there is no point in not letting people benefit from work that has already been done. This is a very general concern of mine about Flux: it lacks a ton of features, it has many bugs. Those are show-stoppers for many people, and we should prioritize them over any performance, consistency, API considerations if needed. And be fast in doing it.

src/cuda/cudnn.jl
```diff
@@ -1,7 +1,20 @@
 import CUDA.CUDNN: batchnorm, ∇batchnorm
 
-(BN::Flux.BatchNorm)(x::Union{CuArray{T,2},CuArray{T,4},CuArray{T,5}}, cache = nothing) where T<:Union{Float32, Float64} =
-  BN.λ.(batchnorm(BN.γ, BN.β, x, BN.μ, BN.σ², BN.momentum; cache = cache, alpha = 1, beta = 0, eps = BN.ϵ, training = Flux.istraining()))
+function (BN::Flux.BatchNorm)(x::Union{CuArray{T,2},CuArray{T,4},CuArray{T,5}},
+                              cache=nothing) where T<:Union{Float32, Float64}
+  
+  @assert BN.affine "BatchNorm: only affine=true supported on gpu"
+  @assert BN.track_stats "BatchNorm: only track_stats=true supported on gpu"
+  @assert length(BN.β) == size(x, ndims(x)-1) "BatchNorm: input has wronng number of channels"
```
mcabbott, line 8: I think these are checks of input, not sanity checks on the program, so ideally would not use `@assert`
```suggestion
  BN.affine || throw(ArgumentError("BatchNorm: only affine=true supported on gpu"))
  BN.track_stats || throw(ArgumentError("BatchNorm: only track_stats=true supported on gpu"))
  length(BN.β) == size(x, ndims(x)-1) || throw(ArgumentError("BatchNorm: input has wronng number of channels"))
```

DhairyaLGandhi:
> devising composable internals is something that can be done later and shouldn't stop us from merging this PR.

Design decisions such as this cause issues down the line. While I appreciate the process of quick movement, it is my experience that doing it later either doesn't happen or is made harder or introduces a bunch of hacks.

CarloLucibello:
another rebase to be done and another 10 days wasted for no reason. @DhairyaLGandhi could you please respect the work I've done here and stop blocking this PR?

DhairyaLGandhi:
It's been a busy week, I'm sure you have seen me sufficiently active on this project. I Haven't intentionally done anything to disrespect your work or block any pr, and I'd apologise if I did.

I can do the rebase myself if you're not up to it. Rebasing is a part of a moving library, if your pr needs to be rebased (in the manifest and news of all things) it meant other things have gone in.

CarloLucibello:
Squashed and rebased. I don't see anything in #1495 that should block this PR. Can we merge this now and merge #1495 whenever is ready?

DhairyaLGandhi:
Well part of the motivation to PR instead of commit/ suggest was that we have to make it easier to remove much of the kwarg handling and lean on the compiler and our composition to take over much of the heavy lifting, which this currently does not. So I would want that side of the design to be addressed before merging.

CarloLucibello:
I don't see any interface change in #1495, just tiny changes to the implementation, nor has there been any sensible objection to the interface proposed in the 3 months this PR has been open. What I see instead, is several bugs and missing features that could have been addressed 2 months ago by merging this. There is no reason why #1495 should block this PR, please let's merge it. I will be happy to review #1495 and merge to master whenever is ready

DhairyaLGandhi:
The change is precisely having a way away from the kwarg syntax. Constant breakages to support this for 2 months only to remove it again is the change I have proposed and have been ignored on. I agree the bugs are annoying, but this would have been breaking either way only to be broken again.

DhairyaLGandhi:
Please don't misunderstand me, I am not trying to block this for no reason, but expect that code quality standards be put in place. You agree that composable interfaces and internals are important, right?

Let's do this, let's merge this to master, and focus on #1495 immediately and steps forward.

DhairyaLGandhi (approved):

CarloLucibello:
The interface proposed here is
```julia
bn = BatchNorm(64, affine=true, track_stats=true)
bn(x)
```
and this is a nice and retro-compatible interface. The only technically breaking change here is a change in the default behavior of `InstanceNorm` which is a fix in practice, because it now becomes equivalent to what other DL frameworks do and the paper itself suggests.
I don't know what you mean by *composable interface*, but for sure we don't want something
like
```
bn = Affine(TrackStats(BatchNorm(64)))
```
which would be wildly breaking and also quite ugly and inconvenient. 
The whole discussion in this PR was about having less branching and more composition in the inner implementation, and I'm totally fine with it, it just can be done somewhere else because it's not breaking and this PR already does a lot. In fact, #1495 is not breaking with respect to this PR. I like the direction of #1495 btw

DhairyaLGandhi:
I don't think that was what was proposed at all, what *was* is that we are able to determine the correct forward pass at compile time, with little to no branching, and a clear structure to add new ones.

```julia
function (sn::SomeNorm)(x)
  calc_shapes(...)
  track_stats(...)
  affine(...)
end
```

In this case, the API for a bunch of these layers would break (using `Val` for bools for ex)

CarloLucibello:
> In this case, the API for a bunch of these layers would break (using Val for bools for ex)

no need to break, the exposed interface should be bool (users shouldn't have to deal with Val), the constructor can convert and parameterize with Val. Anyways, finishing #1495 before v0.12 is certainly desirable

DhairyaLGandhi:
That's just one example, we might want to not have the types grow larger since that can cause compile time regressions or cache misses or fail inlining by mistake. These are tbd, but that work has to be released together to avoid constant breakages.

PR #1412, opened 2020-12-01
update ci

CarloLucibello:
Align github CI to the one suggested here
https://discourse.julialang.org/t/easy-workflow-file-for-setting-up-github-actions-ci-for-your-julia-package/49765
Also removing some platforms to save some trees


DhairyaLGandhi:
Generally a good idea.

Couple notes though, for Flux we really need to be testing on multiple OSes since we have a generally large user base and while I like to think we practice fairly generic code, it's best to catch errors sooner rather than later.

Also, codecov isn't a great metric for our use case, I would remove that

CarloLucibello:
added more platforms
removed travis.yml
let's wait and see if the codecov step brings a failure

DhairyaLGandhi:
We should test Linux with latest Julia as well

CarloLucibello:
> We should test Linux with latest Julia as well

we do, right now `1.5` and `1` are the same thing. When `1.6` comes out we will uncomment the `1` line

DhairyaLGandhi:
Best to do it now, maintenance wise it's more manageable

CarloLucibello:
> Best to do it now, maintenance wise it's more manageable

I'll take care of doing it as soon as 1.6 comes out, I don't want to test the same configuration twice for the next months

DhairyaLGandhi (approved):

PR #1413, opened 2020-12-01
Add buildkite pipeline

DhairyaLGandhi:


.buildkite/pipeline.yml
```diff
@@ -0,0 +1,15 @@
+steps:
+  - label: "GPU Integeration"
+    plugins:
+      - JuliaCI/julia#v0.5:
+          version: '1.5'
+      - JuliaCI/julia-test#v0.3:
+          test_args: "--thorough"
```
maleadt, line 7: that's CUDA.jl-specific

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,15 @@
...
+    branches: "!gh-pages"
+    agents:
+      queue: "juliagpu"
+      cuda: "11.1"
```
maleadt, line 11: Do you require this? Pleasae don't specify it if not

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,15 @@
...
+    env:
+      JULIA_CUDA_VERSION: '11.1'
+      JULIA_CUDA_USE_BINARYBUILDER: 'true'
```
maleadt, line 14: Same here, please don't specify these env vars unless you really need them.

.buildkite/pipeline.yml
DhairyaLGandhi: Done

.buildkite/pipeline.yml
DhairyaLGandhi: removed the extra variables
Thanks!

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,10 @@
+steps:
+  - label: "GPU Integeration"
+    plugins:
+      - JuliaCI/julia#v0.5:
+          version: '1.5'
```
CarloLucibello, line 5:
```suggestion
          version: '1'
```
?

CarloLucibello:
Is this going to fix the timeouts we are seeing lately in Zygote (e.g. https://github.com/FluxML/Zygote.jl/pull/841)?

.buildkite/pipeline.yml
maleadt: That's not supported by https://github.com/JuliaCI/julia-buildkite-plugin

maleadt:
You don't need Bors with Buildkite, so I assume it'll fix that issue.

DhairyaLGandhi:
Yeah, the resources for the GPU CI are essentially being shifted over

.buildkite/pipeline.yml
johnnychen94: This can be supported if using https://github.com/johnnychen94/jill.py

PR #1415, opened 2020-12-04
Update basics.md

osarjeant:
Added input/output mapping at the beginning (parameters example) so that it is easier for beginners to relate the code to the math.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -25,9 +25,9 @@ When a function has many parameters, we can get gradients of each one at the sam
 ```jldoctest basics
 julia> f(x, y) = sum((x .- y).^2) # sum squared error (element-wise)
 
-julia> x = [2,1,3]
+julia> x = [2,1,3];
 
-julia> y = [2,0,2]
+julia> y = [2,0,2];
 
 julia> gradient(f, x, y) # returns the vectors ([δfx],[δfy])
 ([0, 2, 2], [0, -2, -2])
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -23,32 +23,34 @@ julia> d2f(2)
 When a function has many parameters, we can get gradients of each one at the same time:
 
 ```jldoctest basics
-julia> f(x, y) = sum((x .- y).^2); ∑(x-y)^2 for some range of n
+julia> f(x, y) = sum((x .- y).^2) # sum squared error (element-wise)
 
-julia> gradient(f, [2, 1], [2, 0]) # gradient(function,[x1,x2], [y1,y2]) = ([δf/δx(x1,y1),δf/δx(x2,y2)], [δf/δy(x1,y1),δf/δy(x2,y2)])
-([0, 2], [0, -2])
+julia> x = [2,1,3]
+
+julia> y = [2,0,2]
+
+julia> gradient(f, x, y) # returns the vectors ([δfx],[δfy])
+([0, 2, 2], [0, -2, -2])
 ```
 
 But machine learning models can have *hundreds* of parameters! To handle this, Flux lets you work with collections of parameters, via `params`. You can get the gradient of all parameters used in a program without explicitly passing them in.
 
 ```jldoctest basics
-julia> x = [2, 1];
-
-julia> y = [2, 0];
-
 julia> gs = gradient(params(x, y)) do
          f(x, y)
        end
 Grads(...)
 
 julia> gs[x]
-2-element Array{Int64,1}:
+3-element Array{Int64,1}:
  0
+ 2
  2
 
 julia> gs[y]
-2-element Array{Int64,1}:
+3-element Array{Int64,1}:
   0
+ -2
  -2
 ```
 
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -23,9 +23,9 @@ julia> d2f(2)
 When a function has many parameters, we can get gradients of each one at the same time:
 
 ```jldoctest basics
-julia> f(x, y) = sum((x .- y).^2);
+julia> f(x, y) = sum((x .- y).^2); ∑(x-y)^2 for some range of n
 
-julia> gradient(f, [2, 1], [2, 0])
+julia> gradient(f, [2, 1], [2, 0]) # gradient(function,[x1,x2], [y1,y2]) = ([δf/δx(x1,y1),δf/δx(x2,y2)], [δf/δy(x1,y1),δf/δy(x2,y2)])
 ([0, 2], [0, -2])
 ```
 


docs/src/models/basics.md
```diff
@@ -23,9 +23,9 @@ julia> d2f(2)
 When a function has many parameters, we can get gradients of each one at the same time:
 
 ```jldoctest basics
-julia> f(x, y) = sum((x .- y).^2);
+julia> f(x, y) = sum((x .- y).^2); ∑(x-y)^2 for some range of n
 
-julia> gradient(f, [2, 1], [2, 0])
+julia> gradient(f, [2, 1], [2, 0]) # gradient(function,[x1,x2], [y1,y2]) = ([δf/δx(x1,y1),δf/δx(x2,y2)], [δf/δy(x1,y1),δf/δy(x2,y2)])
```
ToucheSir, line 28: There appear to be some off-by-one errors in the reporting of these partial derivatives :slightly_smiling_face:.

docs/src/models/basics.md
```diff
@@ -23,9 +23,9 @@ julia> d2f(2)
 When a function has many parameters, we can get gradients of each one at the same time:
 
 ```jldoctest basics
-julia> f(x, y) = sum((x .- y).^2);
+julia> f(x, y) = sum((x .- y).^2); ∑(x-y)^2 for some range of n
```
ToucheSir, line 26: This is a syntax error because of a missing `#`. I'm also dubious on the value of repeating essentially the same thing minus 2 dots and with a sigma instead of `sum`. We have LaTeX support for gnarlier equations, but in this case even a comment like `# mean squared error` (or no comment at all) would be better IMO.

ToucheSir (changes requested):

docs/src/models/basics.md
darsnack: Would prefer is this is something like
```julia
julia> gradient(f, [2, 1], [2, 0]) # gradient of f evaluated as [x1, x2] == [2, 1] and [y1, y2] == [2, 0]
```
Even better would just be defining `x = [2, 1]` and `y = [2, 0]` in the REPL code itself then switching this to
```julia
gradient(f, x, y) # gradient of f evaluated at x and y
```

darsnack (changes requested):
Thanks for these changes @osarjeant! Do you mind updating to reflect the suggestions?

docs/src/models/basics.md
osarjeant: I agree it is repetitive. I like `#mean square error`. Would it help beginners to explain the dots as element-wise operations? Maybe something like `#MSE calculated elementwise on x and y`. This would allow them to do keyword searches if they are teaching themselves.

darsnack: Sounds good to me.

ToucheSir: Sorry, that should be sum squared error instead of MSE. I also think explaining dot as broadcasting is better handled by Julia's docs or something like https://fluxml.ai/tutorials/2020/09/15/deep-learning-flux.html. We're more focused on explaining ML/AD concepts and how to express them.

darsnack: Yeah I think `# sum of squared element-wise error` is short and fine to add if it ends up helping someone understand without needing to learn broadcasting.

osarjeant: I like that.

docs/src/models/basics.md
osarjeant: I just found it unclear what the output mapping was. I like @darsnack's suggestion.  I think there is ambiguity of the output between possibly being this `([δf1/δx,δf2/δx],[δf1/δy,δf2/δy])`  or `([δf1/δx,δf1/δy],[δf2/δx,δf2/δy]) `

Maybe `gradient(f, x, y) # returns the vectors ([δfx],[δfy])` or use x, y vectors that have 3 values to remove the ambiguity as the output is clearly not the latter.

`julia> gradient(f, [2,1,3],[2,0,2])`
`([0, 2, 2], [0, -2, -2])`

`x = [2,1,3]`
`y = [2,0,2]`
`julia> gradient(f, x, y) # returns the vectors ([δfx],[δfy])`
`([0, 2, 2], [0, -2, -2])`

darsnack: Your second snippet looks good to me

docs/src/models/basics.md
```diff
@@ -23,33 +23,35 @@ julia> d2f(2)
...
 ```jldoctest basics
-julia> f(x, y) = sum((x .- y).^2);
+julia> f(x, y) = sum((x .- y).^2) # sum squared error (element-wise)
 
-julia> gradient(f, [2, 1], [2, 0])
-([0, 2], [0, -2])
+julia> x = [2,1,3]
+
+julia> y = [2,0,2]
+
```
darsnack, line 31:
```suggestion
julia> x = [2,1,3];

julia> y = [2,0,2];

```

Semi-colons are required if we don't want to print out the arrays.

darsnack (changes requested):
Just some small doctest errors that need fixing

docs/src/models/basics.md
```diff
@@ -23,33 +23,35 @@ julia> d2f(2)
...
-julia> f(x, y) = sum((x .- y).^2);
+julia> f(x, y) = sum((x .- y).^2) # sum squared error (element-wise)
 
-julia> gradient(f, [2, 1], [2, 0])
-([0, 2], [0, -2])
+julia> x = [2,1,3];
+
+julia> y = [2,0,2];
+
+julia> gradient(f, x, y) # returns the vectors ([δfx],[δfy])
```
CarloLucibello, line 32: `δfx` doesn't have a clear meaning.  Maybe
```suggestion
julia> gradient(f, x1, x2) # returns (∇ₓ₁f(x1, x2), ∇ₓ₂f(x1, x2))
```
Had to remove `y` since  julia doesn't have `\_y`

osarjeant: Okay, I think I see what you are saying.

How about this 
```julia
julia> gradient(f, x1, x2) # returns ∇f(p) as (dfₓ₁(p),dfₓ₂(p))
```
to avoid so many x1's and x2's:
********************************************************************************
When a function has many parameters, p = (x1, x2), we can get gradients of each one at the same time:
```julia
julia> f(x1, x2) = sum((x1 .- x2).^2); # sum squared error (element-wise)
    
julia> x1 = [2,1,3];
    
julia> x2 = [2,0,2];
    
julia> gradient(f, x1, x2) # returns ∇f(p) as (dfₓ₁(p),dfₓ₂(p))
([0, 2, 2], [0, -2, -2])
```

But machine learning models can have hundreds of parameters! To handle this, Flux lets you work with collections of parameters, via params. You can get the gradient of all parameters used in a program without explicitly passing them in.
```julia
julia> gs = gradient(params(x1, x2)) do
         f(x1, x2)
       end
Grads(...)

julia> gs[x1]
3-element Array{Int64,1}:
 0
 2
 2

julia> gs[x2]
3-element Array{Int64,1}:
  0
 -2
 -2
```

ToucheSir:
This would need a rebase since `basics.md` has changed a bit since, but I'm still not 100% sold on the notation. e.g. it took me a second to realize `dfₓ₁` was referring to the partial derivative.

CarloLucibello:

PR #1423, opened 2020-12-12
fix Dense's docstring

### LINKED ISSUE #1422: Layers' docstrings and constructors inconsistencies
We have a pretty bad situation with the docstrings of Dense and Conv layers (and maybe others) not matching the corresponding method definitions.  

In Conv's docstrings we have
```julia
Conv(weight::AbstractArray, bias, [activation; stride, pad, dilation])
Conv(; weight, bias, [activation, stride, pad, dilation])
```
not perfectly aligned with the actual definitions:
```julia
Conv(k::NTuple{N,Integer}, ch::Pair{<:Integer,<:Integer}, σ = identity;
            init = glorot_uniform,  stride = 1, pad = 0, dilation = 1,
            weight = convfilter(k, ch, init = init), bias = true)
Conv(w::AbstractArray{T,N}, b::Union{Bool, Zeros, AbstractVector{T}}, σ = identity;
Conv(; weight::AbstractArray{T,N}, bias::Union{Bool, Zeros, AbstractVector{T}},
              activation = identity, stride = 1, pad = 0, dilation = 1) where {T,N}
```

The situation is even worse for Dense. The docstring contains only
```julia
Dense(in::Integer, out::Integer, σ = identity; bias=true)
```
while the actual constructors are
```julia
Dense(W, b)
Dense(in::Integer, out::Integer, σ = identity; initW=glorot_uniform, initb=zeros, bias=true)
```

We should also deprecate `init` in Conv and introduce `initW` and `initb` as in the Dense layer.

CarloLucibello:
partial fix for #1422


src/layers/basic.jl
```diff
@@ -99,18 +100,19 @@ dimension is `out` and the remaining dimensions are the same as in the input.
 
 Setting `bias` to `false` will switch the bias  off for the layer.
 
-# Example
-```
+`initW` and `initb` are callables used to initialize weights and biases respectively,
+trhough the calls `initW(out, in)` and `initb(out)`.
```
CarloLucibello, line 104:
```suggestion
through the calls `initW(out, in)` and `initb(out)`.
```

src/layers/basic.jl
```diff
@@ -83,7 +83,8 @@ extraChain(::Tuple{}, x) = ()
 
 
 """
-    Dense(in::Integer, out::Integer, σ = identity; bias=true)
+    Dense(in, out, σ=identity; initW=glorot_uniform, initb=zeros, bias=true)
```
gxyd, line 86: Any specific reason to explicitly remove `in::Integer` and `out::Integer`?

CarloLucibello: just to avoid having an extra-long line, I hope that the expected type is clear from the docstring

CarloLucibello:
why bors didn't merge? maybe an approval is needed?

DhairyaLGandhi (approved):

DhairyaLGandhi:
Let's try again

bors r+

PR #1427, opened 2020-12-15
(Complete) Implementation of label smoothing with crossentropy

### LINKED ISSUE #1016: crossentropy should have label smoothing

gxyd:
Trying to complete the PR #1025 from @sambitdash (Thanks Sambit). Closes #1016 

A few changes compared to the original code
- Throwing an error when `label_smoothing` isn't between 0 and 1.
- Label smoothing is applied as a dispatch.


src/losses/functions.jl
```diff
@@ -82,41 +102,46 @@ calculated as
...
+`label_smoothing` can provide better generalization and continuity to the loss function.
+
 Typically, the prediction `ŷ` is given by the output of a [`sigmoid`](@ref) activation.
 
 Use of `logitbinarycrossentropy` is recomended over `binarycrossentropy` for numerical stability.
 
 See also: [`Flux.Losses.crossentropy`](@ref), [`Flux.Losses.logitcrossentropy`](@ref), [`Flux.Losses.logitbinarycrossentropy`](@ref)
 """
-function binarycrossentropy(ŷ, y; agg=mean, ϵ=epseltype(ŷ))
+function binarycrossentropy(ŷ, y; agg=mean, ϵ=epseltype(ŷ), label_smoothing=zero(eltype(y)))
```
gxyd, line 127: In Sambit's PR here, the default was `label_smoothing=zero(y)`, which I don't think should be the case?

src/losses/functions.jl
```diff
@@ -82,41 +102,46 @@ calculated as
...
+    logitbinarycrossentropy(ŷ, y; agg=mean, label_smoothing=zero(eltype(y)))
 
 Mathematically equivalent to
 [`Flux.binarycrossentropy(σ(ŷ), y)`](@ref) but is more numerically stable.
 
 See also: [`Flux.Losses.crossentropy`](@ref), [`Flux.Losses.logitcrossentropy`](@ref), [`Flux.Losses.binarycrossentropy`](@ref)
 ```
 """
-function logitbinarycrossentropy(ŷ, y; agg=mean)
+function logitbinarycrossentropy(ŷ, y; agg=mean, label_smoothing=zero(eltype(y)))
```
gxyd, line 143: Here also it was `label_smoothing=zero(y)`.

test/losses.jl
```diff
@@ -60,24 +60,49 @@ y = onehotbatch([1, 1, 0, 0], 0:1)
...
+y_dis = copy(y_sim)
+y_dis[1,:], y_dis[2,:] = y_dis[2,:], y_dis[1,:]
 
 @testset "crossentropy" begin
   @test crossentropy([0.1,0.0,0.9], [0.1,0.0,0.9]) ≈ crossentropy([0.1,0.9], [0.1,0.9])
   @test crossentropy(ŷ, y) ≈ lossvalue
+  @test crossentropy(ylp, yl, label_smoothing=2*ls) ≈ -sum(yls.*log.(ylp))
+  @test crossentropy(ylp, yl) ≈ -sum(yl.*log.(ylp))
+  @test iszero(crossentropy(y_same, ya, ϵ=0))
+  @test iszero(crossentropy(ya, ya, ϵ=0))
```
gxyd, line 84: Compared to Sambit's PR, I've explicitly added `ϵ=0` in this line and the one above it.

gxyd:
Bugging @oxinabox and @DhairyaLGandhi

src/losses/functions.jl
gxyd: My bad, I'm wrong about that. `zero(y)` makes more sense. I'll fix that.

src/losses/functions.jl
```diff
@@ -46,31 +46,51 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
    agg(((abs_error.^2) .* temp) .* x .+ δ*(abs_error .- x*δ) .* (1 .- temp))
 end
 
+function apply_label_smoothing(y::AbstractArray, label_smoothing)
+    if (label_smoothing < 0 || label_smoothing > 1)
```
oxinabox, line 50:
```suggestion
    if !(0<label_smoothing1)
```

src/losses/functions.jl
```diff
@@ -46,31 +46,51 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+        throw(ArgumentError("`label_smoothing` must be between 0 and 1"))
```
oxinabox, line 51: I am always tempted by:
```suggestion
        throw(DomainError(label_smoothing, "`label_smoothing` must be between 0 and 1"))
```
but don't super-mind

src/losses/functions.jl
```diff
@@ -46,31 +46,51 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+    elseif !iszero(label_smoothing)
+        y = y .* (1 - label_smoothing) .+ label_smoothing* 1 // size(y, 1)
+    end
+    return y
+end
+
+function apply_label_smoothing(y::Number, label_smoothing)
+    if (label_smoothing < 0 || label_smoothing > 1)
```
oxinabox, line 59:
```suggestion
    if !(0<label_smoothing < 1)
```

src/losses/functions.jl
```diff
@@ -46,31 +46,51 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+        throw(ArgumentError("`label_smoothing` must be between 0 and 1"))
```
oxinabox, line 60: I am always tempted by:
```suggestion
        throw(DomainError(label_smoothing, "`label_smoothing` must be between 0 and 1"))
```
but don't super-mind

src/losses/functions.jl
ToucheSir: `0 < label_smoothing < 1`?

src/losses/functions.jl
```diff
@@ -46,31 +46,51 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+function apply_label_smoothing(y::AbstractArray, label_smoothing)
+    if (label_smoothing < 0 || label_smoothing > 1)
+        throw(ArgumentError("`label_smoothing` must be between 0 and 1"))
+    elseif !iszero(label_smoothing)
+        y = y .* (1 - label_smoothing) .+ label_smoothing* 1 // size(y, 1)
+    end
+    return y
+end
+
+function apply_label_smoothing(y::Number, label_smoothing)
```
oxinabox, line 58:
```suggestion
# The bernoulli case. `y` is a bool or a probability of a bernoulli event
function apply_label_smoothing(y::Number, label_smoothing)
```
If i have this right, then i do wonder if we should check that `0<y<1`.
Since smoothing labels if your labels are integer class IDs does not do what is wanted here.

src/losses/functions.jl
```diff
@@ -46,31 +46,51 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
 as would be the case with the output of a [`softmax`](@ref) operation.
+`label_smoothing` can provide better generalization and continuity to the loss function.
 
 Use of [`logitcrossentropy`](@ref) is recomended over `crossentropy` for
 numerical stability.
 
 See also: [`Flux.logitcrossentropy`](@ref), [`Flux.binarycrossentropy`](@ref), [`Flux.logitbinarycrossentropy`](@ref)
 """
-function crossentropy(ŷ, y; dims=1, agg=mean, ϵ=epseltype(ŷ))
+function crossentropy(ŷ, y; dims=1, agg=mean, ϵ=epseltype(ŷ), label_smoothing=zero(eltype(y)))
```
oxinabox, line 87: coming back to this now, I wonder if we want it as a kwarg to the main function
or if we want to expose it in some other way.

since this can be written as the composition of the normal `crossentropy` and the `apply_label_smoothing` function.
Maybe we should document and instruct user to use this to preprocess their labels (esp since it has to be done again and again each epoch); and it only depends on that data.
and cross reference it in the see also in the docstring.

src/losses/functions.jl
gxyd: Perhaps, it should be `if !(0 ≤ label_smoothing ≤ 1)` as we don't want to raise an error in case of `label_smoothing` being a zero type (which is actually a default), because in that case we should return `y` without making any changes to it.

src/losses/functions.jl
gxyd: I'm somewhat more inclined towards what you think... will fix that.

src/losses/functions.jl
gxyd: Or may be perhaps, have an implementation in both the ways? `kwargs` would be something users of Tensorflow would be more familiar with.

Just like we've for dropout regularization technique in Flux, `Dropout` layer is to be preferred over `dropout`. Though I'm not too familiar with deep learning techniques usage with regards to Flux.

CarloLucibello: I think there is no need to have the keyword argument version, let's keep `crossentropy` clean, the interface should just be

```julia
y = label_smoothing(y, 0.1)
loss = crossentropy(yhat, y)
```

src/losses/functions.jl
```diff
@@ -46,6 +46,30 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
    agg(((abs_error.^2) .* temp) .* x .+ δ*(abs_error .- x*δ) .* (1 .- temp))
 end
 
+"""
+    label_smoothing(y::Union{Number, AbstractArray}, smoothing_factor)
```
gxyd, line 50: I thought of using `ϵ` instead of `smoothing_factor` but since this is more explicit, so I used this. Also `ϵ` might create confusion when applied in combination with `crossentropy` (it has `ϵ` as kwarg).

gxyd:
@oxinabox  ready for another review.

src/losses/functions.jl
CarloLucibello: what about `α`? epsilon usually denotes small quantities. Then it would be useful to have the  mathematical definition of the operation performed in the docstring

gxyd: > Then it would be useful to have the mathematical definition of the operation performed in the docstring

Yes, I was thinking of having it, but do we've a separate docstring for each method of `label_smoothing` or just one at the before starting both the methods?

src/losses/functions.jl
```diff
@@ -111,9 +147,12 @@ end
...
 
 Mathematically equivalent to
-[`Flux.binarycrossentropy(σ(ŷ), y)`](@ref) but is more numerically stable.
+[`binarycrossentropy(σ(ŷ), y)`](@ref) but is more numerically stable.
+
+Use [`label_smoothing`](@ref) to smooth the `y` value as preprocessing before
+computing the loss.
 
-See also: [`Flux.Losses.crossentropy`](@ref), [`Flux.Losses.logitcrossentropy`](@ref), [`Flux.Losses.binarycrossentropy`](@ref)
+See also: [`crossentropy`](@ref), [`logitcrossentropy`](@ref), [`binarycrossentropy`](@ref), [`label_smoothing`](@ref)
```
CarloLucibello, line 155: I'm not sure if removing the specifier Flux.Losses  could cause some referencing issue in Flux's documentation pages

gxyd: I think currently the reference doesn't work in "See Also" of https://fluxml.ai/Flux.jl/stable/models/losses/#Flux.Losses.binarycrossentropy , though I'm probably not looking at the dev-docs.

src/losses/functions.jl
CarloLucibello: It's a bit tricky to cover all use cases:
- a single one-hot vector
- a batch of one-hot
- a single number for binary classification
- a batch of numbers for binary classification

src/losses/functions.jl
gxyd: One example would be: https://github.com/FluxML/Flux.jl/blob/master/src/layers/conv.jl#L18 , here the See reference link works just fine (as it is an internal linking), so we probably don't need `Flux.Losses` in this case.

gxyd: We can refer: https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Cross-Referencing

CarloLucibello: but Conv doesn't leave in a submodule, crossentropy does

oxinabox:
@CarloLucibello will you or I be keeping track of this to see it merged?
If you are looking at it, then I am happy to leave it to you.
But if you think I am watching this closely, then I had better be watching it closely

DhairyaLGandhi:
I prefer this approach over a kwarg as well, thanks!

src/losses/functions.jl
```diff
@@ -46,6 +46,30 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+"""
+function label_smoothing(y::AbstractArray, smoothing_factor::Number)
+    if !(0 < smoothing_factor < 1)
+        throw(ArgumentError("`smoothing_factor` must be between 0 and 1"))
+    end
+    smoothed_y = y .* (1 - smoothing_factor) .+ smoothing_factor* 1 // size(y, 1)
+    return smoothed_y
+end
+
+function label_smoothing(y::Number, smoothing_factor::Number)
```
DhairyaLGandhi, line 65: We shouldnt need this separately I think

CarloLucibello:
> @CarloLucibello will you or I be keeping track of this to see it merged?

I can track it

src/losses/functions.jl
gxyd: (Correct me if I am wrong) As all the functions `crossentropy`, `logitcrossentropy`, `binarycrossentropy`, etc. all live in the same file, they'll need to have internal links to refer to one another, so we'll just want something like `[crossentropy](@ref])`, instead of `[Flux.Losses.Crossentropy](@ref)` to refer to it.

I built the docs on my machine locally and tried visiting the links, it doesn't work with `[Flux.Losses.Crossentropy]` but does work with simply `[crossentropy]`.

src/losses/functions.jl
gxyd: I've tried the approach of using `dims` in the newer commit.

src/losses/functions.jl
```diff
@@ -46,6 +46,41 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
    agg(((abs_error.^2) .* temp) .* x .+ δ*(abs_error .- x*δ) .* (1 .- temp))
 end
 
+"""
+    label_smoothing(y::Union{Number, AbstractArray}, α; dims=1)
+
+Returns smoothed labels, meaning the confidence on label values are relaxed.
+
+when `y` given as one-hot vector or batch of one-hot, its calculated as
```
CarloLucibello, line 54:
```suggestion
When `y` given as one-hot vector or batch of one-hot, its calculated as
```

src/losses/functions.jl
```diff
@@ -46,6 +46,41 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+
+    Label smoothing = y .* (1 - α) .+ α * 1 // size(y, 1)
```
CarloLucibello, line 56:
```suggestion
    y .* (1 - α) .+ α / size(y, dims)
```

src/losses/functions.jl
```diff
@@ -46,6 +46,41 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+its calculated as
+
+    Label smoothing = y .* (1 - α) .+ α*1//2
+
+in which case the labels are squeezed towards `0.5`.
+
+α is a number in interval (0, 1) called the smoothing factor. Higher the
+value of α larger the smoothing of `y`.
+
+`dims` denotes the one-hot dimension (default), unless `dims=0` which denotes
```
CarloLucibello, line 68:
```suggestion
`dims` denotes the one-hot dimension, unless `dims=0` which denotes
```

src/losses/functions.jl
```diff
@@ -46,6 +46,41 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+Returns smoothed labels, meaning the confidence on label values are relaxed.
+
+when `y` given as one-hot vector or batch of one-hot, its calculated as
+
+    Label smoothing = y .* (1 - α) .+ α * 1 // size(y, 1)
+
+when `y` given as a number or batch of numbers for binary classification,
+its calculated as
+
+    Label smoothing = y .* (1 - α) .+ α*1//2
```
CarloLucibello, line 61:
```suggestion
    y .* (1 - α) .+ α / 2
```
Let's not be too pedantic with type problems in the docstring

src/losses/functions.jl
```diff
@@ -46,6 +46,41 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
    agg(((abs_error.^2) .* temp) .* x .+ δ*(abs_error .- x*δ) .* (1 .- temp))
 end
 
+"""
+    label_smoothing(y::Union{Number, AbstractArray}, α; dims=1)
```
CarloLucibello, line 50:
```suggestion
    label_smoothing(y::Union{Number, AbstractArray}, α; dims::Int=1)
```

src/losses/functions.jl
CarloLucibello: ... which denotes the application of label smoothing to binary distributions encoded in a single number.

src/losses/functions.jl
```diff
@@ -46,6 +46,41 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+in which case the labels are squeezed towards `0.5`.
+
+α is a number in interval (0, 1) called the smoothing factor. Higher the
+value of α larger the smoothing of `y`.
+
+`dims` denotes the one-hot dimension (default), unless `dims=0` which denotes
+binary distributions
+
+"""
+function label_smoothing(y::Union{AbstractArray,Number}, α::Number; dims=1)
```
CarloLucibello, line 72:
```suggestion
function label_smoothing(y::Union{AbstractArray,Number}, α::Number; dims::Int=1)
```

src/losses/functions.jl
```diff
@@ -46,6 +46,41 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+    if !(0 < α < 1)
+        throw(ArgumentError("α must be between 0 and 1"))
+    end
+    if (iszero(dims) || isnothing(dims))
+        y_smoothed = y .* (1 - α) .+ α*1//2
+    else
+        y_smoothed = y .* (1 - α) .+ α* 1 // size(y, 1)
```
CarloLucibello, line 79: n = size(y, dims)
a = fill(α, n)
a = reshape(a, [ones(Int, dims-1); n; ones(Int, ndims(y)-dims)])
y_smoothed  = y .* (1 .- α) .+ α ./ n

CarloLucibello:
I made some suggestions in order to accept arbitrary one-hot dimensions. This may well be an overkill, in which case we should warn the user that `dims` should be either 0 or 1.

test/losses.jl
```diff
@@ -56,31 +56,69 @@ end
...
 logŷ = [v 0.0; 0.0 v; 0.0 v; v 0.0]'
 lossvalue = 1.203972804325936
+lossvalue_smoothed = 1.2039728043259348
+yl = onehotbatch([1], 0:1)
+sf = 0.1
+yls = [sf (1-sf)]'  # Effective y after label smoothing
+ylp = [0.9 0.1]'
+logylp = [0.0 v]'
+
+# Construct `sim`ilar and `dis`imilar versions of the dataset so we can test effect of smoothing
```
gxyd, line 71: Would we want to have a working example showing this in the docs somewhere?

src/losses/functions.jl
```diff
@@ -46,6 +46,41 @@ function huber_loss(ŷ, y; agg=mean, δ=ofeltype(ŷ, 1))
...
+    Label smoothing = y .* (1 - α) .+ α*1//2
+
+in which case the labels are squeezed towards `0.5`.
+
+α is a number in interval (0, 1) called the smoothing factor. Higher the
+value of α larger the smoothing of `y`.
+
+`dims` denotes the one-hot dimension (default), unless `dims=0` which denotes
+binary distributions
+
```
CarloLucibello, line 70: full stop at the end of sentence, remove empty line

test/losses.jl
CarloLucibello: you can add an example to the docstring, or somewhere in the docs but I'm not sure where

gxyd:
I've made the changes. Though for
> I made some suggestions in order to accept arbitrary one-hot dimensions. This may well be an overkill, in which case we should just warn the user that dims should be either 0 or 1 and add the corresponding assert

I'm a little skeptical of adding support for `dims` other than `0` and `1`, as to whether that is necessary. But if you believe we'll need that, we'll add that?

CarloLucibello:
> I'm a little skeptical of adding support for dims other than 0 and 1, as to whether that is necessary. But if you believe we'll need that, we'll add that?

since we error out when dims is not 0 or 1, we can safely add it later if the needs arise. 
Thanks  for this contribution, I'm going to merge it now

CarloLucibello (approved):

DhairyaLGandhi:
Adding the condition there is prolly bad for the gradient call. I would think a follow up pr that separates how the range is handled would be welcome @gxyd

gxyd:
Thanks @CarloLucibello and @DhairyaLGandhi . Just wondering though, if we're supposed to update the release notes, or it wasn't necessary as such in this case?

DhairyaLGandhi:
That would be good. Maybe you can add that in? Are you planning on doing the follow-up based on my comment  above?

gxyd:
Yes, will do that. Will probably update in that itself then I guess.

PR #1432, opened 2020-12-19
Generalize train/testmode! to all Functors

ToucheSir:
Addresses https://github.com/FluxML/Flux.jl/pull/1044#issuecomment-599538512. See also https://discourse.julialang.org/t/do-i-have-to-implement-flux-testmode-for-my-own-models/52038.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -19,7 +19,7 @@ Possible values include:
 - `true` for testing
 - `:auto` or `nothing` for Flux to detect the mode automatically
 """
-testmode!(m, mode = true) = m
+testmode!(m, mode = true) = (foreach(x -> testmode!(x, mode), trainable(m)); m)
 
 """
     trainmode!(m, mode = true)
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -39,8 +39,6 @@ applychain(fs::Tuple, x) = applychain(tail(fs), first(fs)(x))
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i]...)
 
-testmode!(m::Chain, mode = true) = (map(x -> testmode!(x, mode), m.layers); m)
-
 function Base.show(io::IO, c::Chain)
   print(io, "Chain(")
   join(io, c.layers, ", ")
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -272,4 +272,17 @@ end
       testdense(re(p), bt)
     end
   end
-end 
+end
+
+@testset "Train and test mode" begin
+  mutable struct DummyLayer
+    testing::Bool
+  end
+  Flux.testmode!(m::DummyLayer, testing=true) = (m.testing = testing; m)
+
+  c = Chain(DummyLayer(true))
+  testmode!(c)
+  @test c[1].testing
+  trainmode!(c)
+  @test !c[1].testing
+end


DhairyaLGandhi:
I generally like this.

Should use fmap if we are to apply it to functors more properly

maxfreu:
Oh, wow! Cool to see this as a PR so quickly! Certainly better than any half-hearted attempt by myself.

ToucheSir:
@DhairyaLGandhi fmap drills al the way down to parameters and passes only those to the callback, so unless there's a structural equivalent I'm not sure it works here?

ToucheSir:
For a perhaps clearer illustration of why `fmap` doesn't work for this use-case, I've whipped up a couple of basic diagrams. Fig. 1 is with `fmap` and Fig. 2 is the implementation in this PR. Arrows are colour-coded for clarity.
image
image

Because `fmap` only passes down leaf nodes to its callback, said callback will have no information about the intermediate structure and no opportunity to trigger overloads on layer structs as is required by `testmode!`. That is demonstrated in the figure by the lack of edges in/out of `testmode!(bn::BatchNorm)`. Only by invoking `testmode!` at every level in the tree can we guarantee those specialized methods are called.

DhairyaLGandhi:
So this where the thinner walk over structs would come in handy. Correctly so too, we should add that to Functors.jl, although conceptually you can only freeze what you could train so `testmode!(trainable)` is consistent that way.

ToucheSir:
I tried extracting out a structural iteration util for Functors, but couldn't think of how to write it without a cache. This direct implementation has the advantage of not needing a cache because the mutual recursion is inlined and every layer is guaranteed to be visited at most once (assuming no cycles).

CarloLucibello:
This is doing a lot of extra work:
```julia
julia> d = Dense(2,2)
Dense(2, 2)

julia> m = Chain(Chain(d,d), Chain(d,d))
Chain(Chain(Dense(2, 2), Dense(2, 2)), Chain(Dense(2, 2), Dense(2, 2)))

julia> ftest(m) = (foreach(x -> (println("ciao"); ftest(x)), Flux.trainable(m)); m)
ftest (generic function with 1 method)

julia> ftest(m)
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
ciao
Chain(Chain(Dense(2, 2), Dense(2, 2)), Chain(Dense(2, 2), Dense(2, 2)))
```
Can we maybe augment `fmap` in Functors.jl with a `leafsonly` keyword?
```julia
function fmap(f, x; cache = IdDict(), leafsonly=true)
  haskey(cache, x) && return cache[x]
  if leafsonly
     cache[x] = isleaf(x) ? f(x) : fmap1(x -> fmap(f, x, cache = cache, leafsonly=true), x)
  else
    cache[x] = f(x)
    fmap1(x -> fmap(f, x, cache = cache, leafsonly=false), x)
  end
end
```

ToucheSir:
Avoiding duplicated work in the presence of tied weights/subnetworks makes sense, I'm a little unclear as to how you'd implement `testmode!` in terms of this updated `fmap` though. The only way I can think of right now is by having 2 separate functions, otherwise it's still mutually recursive (and unless `testmode!` maintains its own cache, no cache will be shared because of this). That would take away the nice property of overloading one method, though.

Also, `isleaf(::BatchNorm) == false`.

CarloLucibello:
Right, I wasn't thinking clearly, this PR's approach is much better

CarloLucibello:
An alternative is 
```julia
testmode!(m, mode) = Functors.fmap1(m -> testmode!(m, mode), m)
```
where fmap1 goes just 1 step down, same as trainable. The foreach-trainable approach is to be preferred though for any mutating function

CarloLucibello (approved):

Issue #670, opened 2019-03-08
Better initialization support


domluna:
This is very nice, thanks!

It would be useful to open an issue to discuss the need for the `Linear` layer here. Hopefully we can make the builtins more flexible so this kind of thing is less necessary.

_Originally posted by @MikeInnes in https://github.com/FluxML/model-zoo/pull/115#issuecomment-470900633_

The primary need for making a new type `Linear`, was the bias initializer only takes in the output dimension, which is intuitive but is problematic when considering some bias initialization rely on more than the output dimension. For example, the default [`nn.Linear`](https://pytorch.org/docs/stable/_modules/torch/nn/modules/linear.html#Linear) layer in PyTorch scales the initialization of the bias by the input dimension. Relevant code:

```python
def reset_parameters(self):
        init.kaiming_uniform_(self.weight, a=math.sqrt(5))
        if self.bias is not None:
            fan_in, _ = init._calculate_fan_in_and_fan_out(self.weight)
            bound = 1 / math.sqrt(fan_in)
            init.uniform_(self.bias, -bound, bound)
```

MikeInnes:
I see, so I think there are two issues we can solve here. Firstly we could use better initialisation for the Dense layers anyway; that would be an easy patch.

Secondly in terms of making initialisation be more flexible, would the syntax `Dense(W, b)` be sufficient here? If so I think this is largely a documentation issue which I'm recording in #671.

domluna:
That should definitely be documented. 

It just occurred to me how multiple dispatch allows for a much more flexible API, whereas Python TensorFlow or PyTorch is limited to a single constructor. For people coming from those ecosystems its nice to have a familiar API but we should investigate further how we can make initialization more Julian.

Naba7:
It looks interesting! Is anybody working on this? I would like to take this up.

darsnack:
The constructor issues here are already addressed. #1243 fixes the rescaling issue, #1423 fixes the missing constructor from the docstring, and #1440 will simplify any issues with `initW`/`initb` and the docstrings. I think this can be safely closed.

mcabbott:
Would it be worth including in the docs something like this? 

"For more complicated initialisation, it's recommended to just write a function. For example, this should match Pytorch's `nn.Linear` layer defaults:
```julia
function pydense(in, out, σ=identity)
  W = Flux.kaiming_uniform(out, in, gain=sqrt(2/5))
  fan_in, _ = Flux.nfan(out, in)
[...]

PR #1440, opened 2020-12-26
`Dense` keyword handling, and docstring

### LINKED ISSUE #1421: Conv with non homogenous array eltypes gives confusing error message
```julia
ERROR: LoadError: MethodError: no method matching Flux.Conv(::Array{Float64, 4}, ::Vector{Float32}, ::typeof(identity); stride=1, pad=0, dilation=1)
Closest candidates are:
  Flux.Conv(::F, ::A, ::V, ::Tuple{Vararg{Int64, N}}, ::Tuple{Vararg{Int64, M}}, ::Tuple{Vararg{Int64, N}}) where {N, M, F, A, V} at /home/mccoy/.julia/packages/Flux/q3zeA/src/layers/conv.jl:71 got unsupported keyword arguments "stride", "pad", "dilation"
  Flux.Conv(::AbstractArray{T, N}, ::Union{AbstractVector{T}, Flux.Zeros}, ::Any; stride, pad, dilation) where {T, N} at /home/mccoy/.julia/packages/Flux/q3zeA/src/layers/conv.jl:105
  Flux.Conv(::Tuple{Vararg{Integer, N}}, ::Pair{var"#s150", var"#s149"} where var"#s149"<:Integer where var"#s150"<:Integer, ::Any; init, stride, pad, dilation, weight, bias) where N at /home/mccoy/.julia/packages/Flux/q3zeA/src/layers/conv.jl:132
```
when using

```julia
Conv((5, 5), 10 => 20, identity; init = glorot_uniform64),
```

### LINKED ISSUE #1422: Layers' docstrings and constructors inconsistencies

mcabbott:
Closes #1422, by killing the `initW` keyword, in favour of `init` as used by the Conv layers. 

Also fixes "in×out weight matrix" which was incorrect. 

And makes `Dense(rand(2,3), bias)` work like `Dense(3,2; bias)`, which again is like the Conv layers.

Edit -- also closes #1421 now, ensuring that the bias vectors of both Conv and Dense layers match the eltype of the weights. 

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -7,3 +7,14 @@
 @deprecate Conv(; weight,  bias, activation=identity, kws...) Conv(weight, bias, activation; kws...) 
 @deprecate ConvTranspose(; weight, bias, activation=identity, kws...) ConvTranspose(weight, bias, activation; kws...) 
 @deprecate DepthwiseConv(; weight, bias, activation=identity, kws...) DepthwiseConv(weight, bias, activation; kws...) 
+
+function Base.getproperty(a::Dense, s::Symbol)
+  if s === :W
+    Base.depwarn("field name dense.W is deprecated in favour of dense.weight", :Dense)
+    return getfield(a, :weight)
+  elseif s === :b
+    Base.depwarn("field name dense.b is deprecated in favour of dense.bias", :Dense)
+    return getfield(a, :bias)
+  end
+  return getfield(a, s)
+end
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -69,25 +69,22 @@ extraChain(::Tuple{}, x) = ()
 
 
 """
-    Dense(in, out, σ=identity; initW=glorot_uniform, initb=zeros, bias=true)
-    Dense(W, b, σ=identity)
+    Dense(in, out, σ=identity; bias=true)
+    Dense(W::AbstractMatrix, bias, [σ])
 
-Create a traditional `Dense` layer with in×out weight matrix `W` and 
-bias vector  `b` of length `out`. The forward pass is given by:
+Create a traditional `Dense` layer, whose forward pass is given by:
 
     y = σ.(W * x .+ b)
 
-The input `x` must be a vector of length `in`, a batch of vectors represented
-as an `in × N` matrix, or a higher order tensor where all dimensions
-after the first one will be treated as batch dimensions.
+The input `x` should be a vector of length `in`, or batch of vectors represented
+as an `in × N` matrix, or any array with `size(x,1) == in`.
+The out `y` will be a vector  of length `out`, or a batch with
+`size(y) == (out, size(x)[2:end]...)`
 
-The out `y` will be a vector  of length `out` or a batch whose first
-dimension is `out` and the remaining dimensions are the same as in the input.
-
-Setting `bias` to `false` will switch the bias  off for the layer.
-
-`initW` and `initb` are callables used to initialize weights and biases respectively,
-through the calls `initW(out, in)` and `initb(out)`.
+Keyword `bias=false` will switch off trainable bias for the layer.
+The initialisation of the weight matrix is `W = init(out, in)`, controlled by
+another keyword, with default `init=glorot_uniform`. The weight matrix and the
+weight vector may also be provided explicitly.
 
 # Examples
 
@@ -113,8 +110,24 @@ end
 Dense(W, b) = Dense(W, b, identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
-               initW = glorot_uniform, initb = zeros, bias=true)
-  return Dense(initW(out, in), create_bias(bias, initb, out), σ)
+               initW = nothing, initb = nothing,
+               init = glorot_uniform, bias=true)
+
+  W = if initW !== nothing
+    @warn "keyword initW is deprecated, please use init" maxlog=1 _id=hash(initW)
+    initW(out, in)
+  else
+    init(out, in)
+  end
+
+  b = if bias === true && initb !== nothing
+    @warn "keyword initb is deprecated, please simply supply the " maxlog=1 _id=hash(initW)
+    create_bias(bias, initb, out)
+  else
+    create_bias(bias, zeros, out)
+  end
+
+  return Dense(W, b, σ)
 end
 
 @functor Dense
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -69,12 +69,12 @@ extraChain(::Tuple{}, x) = ()
 
 
 """
-    Dense(in, out, σ=identity; bias=true)
-    Dense(W::AbstractMatrix, bias, [σ])
+    Dense(in, out, σ=identity; bias=true, init=glorot_uniform)
+    Dense(W::AbstractMatrix, [bias, σ])
 
 Create a traditional `Dense` layer, whose forward pass is given by:
 
-    y = σ.(W * x .+ b)
+    y = σ.(W * x .+ bias)
 
 The input `x` should be a vector of length `in`, or batch of vectors represented
 as an `in × N` matrix, or any array with `size(x,1) == in`.
@@ -82,49 +82,59 @@ The out `y` will be a vector  of length `out`, or a batch with
 `size(y) == (out, size(x)[2:end]...)`
 
 Keyword `bias=false` will switch off trainable bias for the layer.
-The initialisation of the weight matrix is `W = init(out, in)`, controlled by
-another keyword, with default `init=glorot_uniform`. The weight matrix and the
-weight vector may also be provided explicitly.
+The initialisation of the weight matrix is `W = init(out, in)`, calling the function
+given to keyword `init`, with default [`glorot_uniform`](@doc Flux.glorot_uniform).
+The weight matrix and/or the bias vector (of length `out`) may also be provided explicitly.
 
 # Examples
-
-```julia-repl
+```jldoctest
 julia> d = Dense(5, 2)
 Dense(5, 2)
 
-julia> d(rand(Float32, 5))
-2-element Array{Float32,1}:
- -0.16210233
-  0.123119034
+julia> d(rand(Float32, 5, 64)) |> size
+(2, 64)
 
-julia> d = Dense(5, 2; bias=false)
-Dense(5, 2)
+julia> d(rand(Float32, 5, 1, 1, 64)) |> size
+(2, 1, 1, 64)
+
+julia> d1 = Dense(ones(2, 5), false, tanh)
+Dense(5, 2, tanh; bias=false)
+
+julia> d1(ones(5))
+2-element Array{Float64,1}:
+ 0.9999092042625951
+ 0.9999092042625951
+
+julia> Flux.params(d1)  # no trainable bias
+Params([[1.0 1.0 … 1.0 1.0; 1.0 1.0 … 1.0 1.0]])
 ```
 """
-struct Dense{F,S<:AbstractArray,T<:Union{Zeros, AbstractVector}}
-  W::S
-  b::T
+struct Dense{F, M<:AbstractMatrix, B}
+  weight::M
+  bias::B
   σ::F
+  function Dense(W::M, bias = true, σ::F = identity) where {M<:AbstractMatrix, F}
+    b = create_bias(W, bias, size(W,1))
+    new{F,M,typeof(b)}(W, b, σ)
+  end
 end
-
-Dense(W, bias) = Dense(W, create_bias(bias, zeros, size(W,1)), identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
                initW = nothing, initb = nothing,
                init = glorot_uniform, bias=true)
 
   W = if initW !== nothing
-    @warn "keyword initW is deprecated, please use init" maxlog=1 _id=hash(initW)
+    Base.depwarn("keyword initW is deprecated, please use init (which similarly accepts a funtion like randn)", :Dense)
     initW(out, in)
   else
     init(out, in)
   end
 
   b = if bias === true && initb !== nothing
-    @warn "keyword initb is deprecated, please simply supply the " maxlog=1 _id=hash(initb)
-    create_bias(bias, initb, out)
+    Base.depwarn("keyword initb is deprecated, please simply supply the bias vector, bias=initb(out)", :Dense)
+    initb(out)
   else
-    create_bias(bias, zeros, out)
+    bias
   end
 
   return Dense(W, b, σ)
@@ -133,41 +143,52 @@ end
 @functor Dense
 
 function (a::Dense)(x::AbstractArray)
-  W, b, σ = a.W, a.b, a.σ
+  W, b, σ = getfield(a, :weight), getfield(a, :bias), getfield(a, :σ)
   sz = size(x)
-  x = reshape(x, sz[1], :) # reshape to handle dims > 1 as batch dimensions 
+  x = reshape(x, sz[1], :) # reshape to handle dims > 1 as batch dimensions
   x = σ.(W*x .+ b)
   return reshape(x, :, sz[2:end]...)
 end
 
 function Base.show(io::IO, l::Dense)
-  print(io, "Dense(", size(l.W, 2), ", ", size(l.W, 1))
+  print(io, "Dense(", size(l.weight, 2), ", ", size(l.weight, 1))
   l.σ == identity || print(io, ", ", l.σ)
+  l.bias == Zeros() && print(io, "; bias=false")
   print(io, ")")
 end
 
 """
     Diagonal(α, β)
-    Diagonal(sz::Integer...; initα=ones, initβ=zeros)
+    Diagonal(size::Integer...)
 
-Create an element-wise linear layer with learnable
-arrays `α` and `β` of size `sz`. The layer performs
+Create an element-wise linear layer, which performs
 
     y = α .* x .+ β
 
-The input `x` must have size broadcast-compatible with `α` and `β`.
-The parameters will be created with the calls 
-`α = initα(sz)` and `β = initβ(sz)`.
+The learnable arrays are initialised `α = ones(Float32, size)` and
+`β = zeros(Float32, size)`.
+
+Used by [`LayerNorm`](@ref).
 """
 struct Diagonal{T}
   α::T
   β::T
 end
 
-function Diagonal(sz::Integer...; 
-      initα = i -> ones(Float32, i), 
-      initβ = i -> zeros(Float32, i))
-  Diagonal(initα(sz), initβ(sz))
+function Diagonal(sz::Integer...; initα = nothing, initβ = nothing)
+  α = if initα !== nothing
+    Base.depwarn("keyword initα is deprecated, please simply supply the desired vectors", :Diagonal)
+    initα(sz...)
+  else
+    ones(sz...)
+  end
+  β = if initβ !== nothing
+    Base.depwarn("keyword initβ is deprecated, please simply supply the desired vectors", :Diagonal)
+    initβ(sz...)
+  else
+    zeros(sz...)
+  end
+  Diagonal(α, β)
 end
 
 @functor Diagonal
@@ -175,7 +196,7 @@ end
 (a::Diagonal)(x) = a.α .* x .+ a.β
 
 function Base.show(io::IO, l::Diagonal)
-  print(io, "Diagonal(", size(l.α), ")")
+  print(io, "Diagonal(", join(size(l.α), ", "), ")")
 end
 
 """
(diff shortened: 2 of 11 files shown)


CarloLucibello:
should we add `init` to Dense or add `initW` and `initb` to Conv?

mcabbott:
I think simpler is better here. I can imagine wanting a different scaling for `initW`, but my guess is that using `initb` is vanishingly rare, and that looking up how it's meant to work is probably more effort than just passing the vector you want.

CarloLucibello:
agree, changing the bias init is very rare, so since we have to modify either Conv or Dense for consistency, better go with the simplest interface. 

One more thing should be changed though: if the bias is not explicitly passed to the constructor, it should be created with the same eltype of the weights, while currently is always Float32. This would solve #1421

src/layers/basic.jl
```diff
@@ -126,11 +123,27 @@ struct Dense{F,S<:AbstractArray,T<:Union{Zeros, AbstractVector}}
...
+Dense(W::AbstractMatrix, bias=true) = Dense(W, create_bias(bias, zeros, size(W,1)), identity)
 
 function Dense(in::Integer, out::Integer, σ = identity;
-               initW = glorot_uniform, initb = zeros, bias=true)
-  return Dense(initW(out, in), create_bias(bias, initb, out), σ)
+               initW = nothing, initb = nothing,
+               init = glorot_uniform, bias=true)
+
+  W = if initW !== nothing
+    @warn "keyword initW is deprecated, please use init (which similarly accepts a funtion like randn)" maxlog=1 _id=hash(initW)
```
CarloLucibello, line 133: Can we add a deprecation instead? Not sure this would work:

```
@deprecate Dense(args...; initW, kws...)  Dense(args...; init=initW; kws...)
```
and something similar if possible for `initb`

test/layers/basic.jl
```diff
@@ -46,11 +47,14 @@ import Flux: activations
       @test size(Dense(10, 5)(randn(10,2,3,4))) == (5,2,3,4)
     end
     @testset "zeros" begin
+      # old keywords
       @test Dense(10, 1, identity, initW = ones, initb = zeros)(ones(10,1)) == 10*ones(1, 1)
       @test Dense(10, 1, identity, initW = ones, initb = zeros)(ones(10,2)) == 10*ones(1, 2)
       @test Dense(10, 2, identity, initW = ones, initb = zeros)(ones(10,1)) == 10*ones(2, 1)
       @test Dense(10, 2, identity, initW = ones, initb = zeros)([ones(10,1) 2*ones(10,1)]) == [10 20; 10 20]
       @test Dense(10, 2, identity, initW = ones, bias = false)([ones(10,1) 2*ones(10,1)]) == [10 20; 10 20]
```
CarloLucibello, line 55: deprecated constructors should be removed from tests

src/layers/basic.jl
mcabbott: I don't think that works:
```
julia> fun(x; kw="default") = @show kw
fun (generic function with 1 method)

julia> @deprecate fun(x; oldkw) fun(x; kw)
fun (generic function with 1 method)

julia> fun(1, kw=3) # new
ERROR: UndefKeywordError: keyword argument oldkw not assigned
```

test/layers/basic.jl
mcabbott: Done

src/layers/basic.jl
mcabbott: Done.

mcabbott:
>  if the bias is not explicitly passed to the constructor, it should be created with the same eltype of the weights, while currently is always Float32. This would solve #1421

Hadn't seen the issue. It sounds like we should change `create_bias(bias, zeros, out)` to sort this out everywhere, by passing it the eltype of the weights?

CarloLucibello:
> Hadn't seen the issue. 

me neither, just found it scrolling open issues

> It sounds like we should change create_bias(bias, zeros, out) to sort this out everywhere, by passing it the eltype of the weights? Or better the whole array... It should (1) pass through false, or (2) create via similar(W,out) .= 0, or (3) convert a given vector to match eltype.

The interface could be `create_bias(bias, W, out)`  and then do (1) and (2) but not (3), I would interfere with explicitly passed biases

src/layers/basic.jl
```diff
@@ -85,39 +85,42 @@ extraChain(::Tuple{}, x) = ()
...
-Dense(5, 2)
+julia> d1 = Dense(ones(2,5), false, tanh)
+Dense(5, 2, tanh; bias=false)
+
+julia> d1(ones(5))
+2-element Array{Float64,1}:
+ 0.9999092042625951
+ 0.9999092042625951
+
+julia> params(d1)  # no trainable bias
```
CarloLucibello, line 122:
```suggestion
julia> Flux.params(d1)  # no trainable bias
```
we generally qualify it

test/layers/basic.jl
```diff
@@ -46,10 +47,10 @@ import Flux: activations
...
     @testset "zeros" begin
-      @test Dense(10, 1, identity, initW = ones, initb = zeros)(ones(10,1)) == 10*ones(1, 1)
-      @test Dense(10, 1, identity, initW = ones, initb = zeros)(ones(10,2)) == 10*ones(1, 2)
-      @test Dense(10, 2, identity, initW = ones, initb = zeros)(ones(10,1)) == 10*ones(2, 1)
-      @test Dense(10, 2, identity, initW = ones, initb = zeros)([ones(10,1) 2*ones(10,1)]) == [10 20; 10 20]
+      @test Dense(10, 1, identity, init = ones)(ones(10,1)) == 10*ones(1, 1)
+      @test Dense(10, 1, identity, init = ones)(ones(10,2)) == 10*ones(1, 2)
+      @test Dense(10, 2, identity, init = ones)(ones(10,1)) == 10*ones(2, 1)
+      @test Dense(10, 2, identity, init = ones)([ones(10,1) 2*ones(10,1)]) == [10 20; 10 20]
       @test Dense(10, 2, identity, initW = ones, bias = false)([ones(10,1) 2*ones(10,1)]) == [10 20; 10 20]
```
CarloLucibello, line 54:
```suggestion
      @test Dense(10, 2, identity, init = ones, bias = false)([ones(10,1) 2*ones(10,1)]) == [10 20; 10 20]
```

mcabbott: Thanks, fixed.

src/layers/basic.jl
mcabbott: Done.

mcabbott:
Oh I didn't look before I wrote this, sorry. I went for `create_bias(W, bias, out)`. I do think we should convert, though, as I cannot see having mixed eltypes here as being anything but a mistake. If you wish to make a layer that promotes, then I think you should explicitly write that, and not rely on the broadcasting of bias to do it for you.

mcabbott:
Adding a size check to `create_bias` found a number of mistakes in tests along the lines of `Dense(rand(2,3), rand(3))` --- it's a bit confusing that this isn't `Dense(2, 3, bias=rand(3))`. 

I am sure I've suggested this before, but what thoughts on `Dense(2 => 3, bias=rand(3))` as a better constructor? This would match how Conv writes channels. 

If we do this, I think `Dense(in, out)` should just silently work, but the printing etc, so as not to annoy everyone with old code, but the printing & docs could be updated.

src/utils.jl
```diff
@@ -177,18 +177,22 @@ ones(dims...) = Base.ones(Float32, dims...)
...
-Return a bias parameter for a layer.
+Return a bias parameter for a layer, based on the value given
+to the constructor's keyword `bias=bias`.
 
-Essentially handles the allowed input options for the `bias` keyword:
-    If `false`: Return the `Zeros` type which turns bias off.
-    If `true` : Return the result of `iftrue(dims)`.
-    If not a boolean, return self to handle the case of bias=somearray.
+* `bias == true` creates a zero vector, of the same type as weights.
+* `bias == false` returns `Zeros()`, a special struct which is a exists to encode the absence of bias.
```
CarloLucibello, line 186:
```suggestion
* `bias == false` returns `Zeros()`, a special struct which exists to encode the absence of bias.
```

CarloLucibello:
> I am sure I've suggested this before, but what thoughts on Dense(2 => 3, bias=rand(3)) as a better constructor? This would match how Conv writes channels.

I'm unsure about this: on one hand, Dense(2 => 3) is clearer, but Dense(2, 3) has been there forever and is much nicer to write. Since I can foresee an endless discussion about this, let's leave it out of this PR

mcabbott:
OK. Then perhaps this is done, I think, modulo one doc typo I just saw. 

It doesn't at present give a warning when converting bias to match the weight's eltype, I have this locally, but no strong feelings either way.

src/utils.jl
```diff
@@ -177,18 +177,22 @@ ones(dims...) = Base.ones(Float32, dims...)
...
+* `bias == false` returns `Zeros()`, a special struct which exists to encode the absence of bias.
+* `bias::AbstractArray` uses the array provided, provided it has the correct size and eltype. If the type is wrong, it will be converted.
 """
-create_bias(shallcreate::Bool, iftrue, dims...) = shallcreate ? iftrue(dims...) : Zeros()
-create_bias(x, ::Any...) = x  
+function create_bias(weights::AbstractArray{T}, bias::Union{Bool, AbstractArray}, dims::Integer...) where {T}
+  bias===true && return fill!(similar(weights, dims...), 0)
+  bias===false && return Zeros()
+  size(bias) == dims || throw(DimensionMismatch("expected bias of size $dims, but got $(size(bias))"))
+  eltype(bias) == T || return T.(bias)
```
CarloLucibello, line 193: if really want to force the type better print a warning

test/utils.jl
```diff
@@ -241,21 +241,22 @@ end
...
     import Flux: loadparams!
-    pars(w, b::Zeros) = [w, zeros(size(w,2))]
     pars(w, b) = [w, b] 
+    import Flux: loadparams!, Zeros
+    pars(w, b::Zeros) = [w, Flux.zeros(size(w,1))]
     pars(l) = pars(l.W, l.b)
     pararray(m) = mapreduce(pars, vcat, m)
     weights(m) = mapreduce(l -> [l.W], vcat, m)
-    @testset "Bias type $bt" for bt in (zeros, nobias)
+    @testset "Bias type $bt" for bt in (Flux.zeros, nobias)
```
CarloLucibello, line 250: this Flux.zeros is confusing, at some point we should redefine it as `zeros32`

mcabbott: This whole test is far too clever for its own good. I was trying to reduce the number of conversion warnings locally, but in fact this didn't help at all...

src/utils.jl
mcabbott: Done.

CarloLucibello (approved):

CarloLucibello:
also recurrent layers have `initb` ...

mcabbott:
Oh I didn't see the RNN ones. Not this PR, I think.

ToucheSir:
> thoughts on Dense(2 => 3, bias=rand(3)) as a better constructor?

Worth noting that `Conv` already uses this syntax for `in_channels => out_channels`.

mcabbott:
> Worth noting that Conv already uses this 

Yes, maybe I forgot to say that. Although I think of that as a plus, Dense treats the first dimension as channels. Is that what you meant too, or do you see a conflict of meanings?

ToucheSir:
I meant that while we're on the topic of inconsistencies between Dense, Conv and RNN constructors, might as well note that the syntax for specifying # of input and output features(/channels) is also inconsistent. e.g. PyTorch does separate `in` and `out` arguments for all 3.

mcabbott:
Ok! I am very happy to make that change here. Or in another PR, this one is done I think, and perhaps better to do things one by one.

ToucheSir:
Yeah, ultimately it's more of a cosmetic change and I'm not qualified to comment on what the ROI for breaking the interface might be. We can revisit this in another issue :)

CarloLucibello (dismissed):

CarloLucibello:
@DhairyaLGandhi merge?

CarloLucibello:
This needs a rebase now. @DhairyaLGandhi please be quicker with the reviews, it is very annoying to have to keep bumping and rebasing

DhairyaLGandhi:
Seeing as several PRs that have gone in required further changes to match the code quality expectations we have, but never got, imo it's alright to not rush.

src/layers/basic.jl
```diff
@@ -69,52 +69,84 @@ extraChain(::Tuple{}, x) = ()
...
+    create_bias(W, initb(out), out)
+  else
+    create_bias(W, bias, out)
+  end
+
+  return Dense(W, b, σ)
+end
+
+function Dense(W::M, bias::Zeros, σ::F=identity) where {F, M<:AbstractMatrix}
+  Base.depwarn("bias=Zeros() is deprecated, please use bias=false", :Dense)
```
DhairyaLGandhi, line 148: This isn't deprecated, and I don't think it's supposed to be

DhairyaLGandhi (changes requested):

CarloLucibello:
Can the 3 constructors 
```julia
Dense(W::M, bias::AbstractVector, σ::F=identity) where {F, M<:AbstractMatrix}
Dense(W::AbstractMatrix, bias::Bool=true, σ=identity)
Dense(W::M, bias::Zeros, σ::F=identity) where {F, M<:AbstractMatrix}
```  
be merged into a single one
```julia
Dense(W::M, bias, σ::F=identity) where {F, M<:AbstractMatrix}
```
and we let `create_bias` handle the bias inside? This should be doable since as @DhairyaLGandhi said the Zeros constructor should not be deprecated

mcabbott:
I tried, but the tricky bit is that the constructor which takes a bias vector needs to have a more specific type than the default constructor from the struct definition -- in order that this goes via `create_bias` which checks length & types. 

As an implementation detail Zeros persists, but as a the user-facing API the change from keyword `bias=Flux.Zeros()` to keyword  `bais=false` was some time ago.

DhairyaLGandhi:
Maybe don't over specialize the signature? The struct should be agnostic to the type of parameters it is input.

mcabbott:
Oh, you mean make the struct itself less specific? So far I only changed the constructors, but indeed can try that next time.

DhairyaLGandhi:
I don't mean change the struct, I meant don't overspecialize the constructors

```julia
Dense{..., Zeros}
````

Or similar is a valid definition, which we follow anyway.

**Comment by mcabbott** [comment]
Well,

>  the constructor which takes a bias vector needs to have a more specific type than the default constructor from the struct definition -- in order that this goes via create_bias which checks length & types.

**Comment by DhairyaLGandhi** [comment]
I have reservations making the Dense layer be restricted to a matrix as well, since we have in general taken the direction to let users decide these metrics, I would probably retain the doc fixes but avoid making any changes to the signatures.

**Comment by mcabbott** [comment]
Got an example? Of a non-AbstractMatrix weight matrix, I mean.

**Comment by DhairyaLGandhi** [comment]
https://github.com/FluxML/Flux.jl/pull/1405

**Comment by DhairyaLGandhi** [comment]
Besides not having existing examples does not mean some constructs may not be useful in other domains, so keeping our tooling generic has been important. That's why we have DiffEqFlux et al because of generic composable components and that philosophy pays dividends in the long term, and is something I wish to bring back

**Comment by DhairyaLGandhi** [comment]
Optimisations over on top are good and something we encourage (multiple dispatch on forward passes are under utilised) for making common cases fast and dealing with special structures for other cases etc, but underlying tooling and esp layers which do generic deep learning ops should have few assumptions.

**Comment by mcabbott** [comment]
> #1405

Thanks, this clears it up, now I see where you were coming from. That PR widened the type of *input* accepted, the data, and narrowing that to Matrix only would indeed be a bad idea, make things less general, as you say...

But no, that's not being proposed here.

**Comment by DhairyaLGandhi** [comment]
If you mean specifically about the weight being a matrix, I would still hold out since we don't exactly lose out on much by allowing for the use case to exist.

**Comment by CarloLucibello** [comment]
@mcabbott #1009 is being merged. I'd like to see v0.12 released soon, this is one of the few items left. What do you think is missing beside a rebase?

**Comment by CarloLucibello** [comment]
bumpity bump

File: src/deprecations.jl [hunk 9]
```diff
@@ -7,3 +7,14 @@
...
+function Base.getproperty(a::Dense, s::Symbol)
+  if s === :W
+    Base.depwarn("field name dense.W is deprecated in favour of dense.weight", :Dense)
+    return getfield(a, :weight)
+  elseif s === :b
+    Base.depwarn("field name dense.b is deprecated in favour of dense.bias", :Dense)
+    return getfield(a, :bias)
+  end
+  return getfield(a, s)
+end
```
CarloLucibello, line 20: just for context, this PR is renaming Dense's internal fields to match Conv's ones. It is slightly annoying to change, but better sooner than later

src/layers/basic.jl
```diff
@@ -69,100 +69,134 @@ extraChain(::Tuple{}, x) = ()
...
+  end
+
+  return Dense(W, b, σ)
 end
 
 @functor Dense
 
 function (a::Dense)(x::AbstractArray)
-  W, b, σ = a.W, a.b, a.σ
+  W, b, σ = getfield(a, :weight), getfield(a, :bias), getfield(a, :σ)
```
CarloLucibello, line 146: why getfield?

CarloLucibello: to bypass getproperty during the deprecation phase? you think is relevant for performance?

src/layers/basic.jl
```diff
@@ -249,55 +283,67 @@ function Base.show(io::IO, b::SkipConnection)
 end
 
 """
-    Bilinear(in1, in2, out)
+    Bilinear(in1, in2, out, σ=identity; bias=true, init=glorot_uniform)
+    Bilinear(W::AbstractArray, [bias, σ])
 
 Creates a Bilinear layer, which operates on two inputs at the same time.
-It has parameters `W` and `b`, and its output given vectors `x`, `y` is of the form 
+It its output, given vectors `x`, `y` is another vector `z` with,
```
CarloLucibello, line 290:
```suggestion
Its output, given vectors `x`, `y` is another vector `z` with,
```

src/layers/basic.jl
mcabbott: It's just avoiding the custom `getproperty`. Perhaps an over-optimisation in this context, haven't timed anything, but shouldn't hurt.

src/utils.jl
```diff
@@ -291,18 +291,23 @@ ones(dims...) = Base.ones(Float32, dims...)
...
-    If not a boolean, return self to handle the case of bias=somearray.
+* `bias == true` creates a zero vector, of the same type as weights.
+* `bias == false` returns `Zeros()`, a special struct which exists to encode the absence of bias.
+* `bias::AbstractArray` uses the array provided, provided it has the correct size and eltype. If the type is wrong, it will be converted.
 """
-create_bias(shallcreate::Bool, iftrue, dims...) = shallcreate ? iftrue(dims...) : Zeros()
-create_bias(x, ::Any...) = x
+function create_bias(weights::AbstractArray{T}, bias::Union{Bool, AbstractArray}, dims::Integer...) where {T}
+  bias===true && return fill!(similar(weights, dims...), 0)
+  bias===false && return Zeros()
```
CarloLucibello, line 305: maybe just cleaner to split into Bool and AbstractArray methods

src/utils.jl
```diff
@@ -291,18 +291,23 @@ ones(dims...) = Base.ones(Float32, dims...)
...
+  size(bias) == dims || throw(DimensionMismatch("expected bias of size $dims, but got $(size(bias))"))
+  eltype(bias) == T && return bias
+  @warn "converting bias to match element type of weights" typeof(weights) typeof(bias) maxlog=3 _id=hash(dims)
+  return T.(bias)
```
CarloLucibello, line 309: this is one thing I'm not entirely sure about, we are making it impossible to create layers where weights and bias have different eltype. Seems very unlikely to willingly do something different though, so probably casting is good and will help catch errors.

src/zeros.jl
```diff
@@ -47,3 +47,6 @@ broadcasted(::typeof(-), a::Zeros, b::AbstractArray) = -b
 @adjoint broadcasted(::typeof(*), a::AbstractArray, b::Zeros) = zero(a), _ -> (nothing, zero(a), nothing)
 @adjoint broadcasted(::typeof(*), a::Zeros, b::AbstractArray) = zero(b), _ -> (nothing, nothing, zero(b))
 @adjoint broadcasted(::typeof(/), a::Zeros, b::AbstractArray) = zero(b), _ -> (nothing, nothing, zero(b))
+
+# Pass-through for layer constructors
+create_bias(weights::AbstractArray, bias::Flux.Zeros, dims::Integer...) = bias
```
CarloLucibello, line 52: better to have this near other create_bias methods

src/utils.jl
CarloLucibello: Maybe better to just wan without converting though, so that users are alerted but they can still use mixed types if they really want to

mcabbott: Yes. I agree this narrows the scope of what you can do with `Dense`, but I think by far the most likely reason to have a mismatch is that you made a mistake, `bias=fill(1/n, 10)` or something. 

I think it's easy enough to make your own layer if you truly want it to (say) promote to Float64 before broadcasting `σ`, or to write that step as another layer.

CarloLucibello: > Yes. I agree this narrows the scope of what you can do with Dense, but I think by far the most likely reason to have a mismatch is that you made a mistake, bias=fill(1/n, 10) or something.

I agree with that, but in that case a warning without casting seems enough

src/layers/basic.jl
CarloLucibello: shouldn't hurt, but since this is the layer everyone scrutinizes when exploring flux, maybe better care a bit about aesthetics
if there are no performance concerns
```suggestion
  W, b, σ = a.weight, a.bias, a.σ
```

src/layers/basic.jl
CarloLucibello: actually, we should be talking arrays here, not vectors

src/utils.jl
mcabbott: Indeed. Done.

src/layers/basic.jl
mcabbott: Fixed, thanks.

src/zeros.jl
mcabbott: I moved it because Zeros isn't loaded yet. Could move files around. (Or we could kill Zeros and be done with it.)

src/deprecations.jl
mcabbott: And likewise the field names of Bilinear, although I think that's never been in a tagged version, so no depwarns.

src/zeros.jl
CarloLucibello: aahah, maybe one day. Let's keep this method here then.

src/utils.jl
mcabbott: I do think I prefer conversion, but could be persuaded. 

Part of my motivation was mis-reading #1421. The user-facing issue was that, if `init` made a Float64 array, you got an error. That's fixed. But in fact it's not the `Conv` type which restricts to same-eltype-bias, but just one of the other constructor functions. That still rejects e.g. `Conv(rand(4,5,6), ones(Float32, 6))`, it's a MethodError without ever calling `create_bias`. Regardless of whether `create_bias` warns or converts, that method should be wider, right?

src/layers/basic.jl
mcabbott: OK, extremely crude timing has the `a.W` taking about 8ns, but `a.weight` just 1ns. I think that's never ever going to matter when the next step is matmul. Fixed.

CarloLucibello (approved):

darsnack:
Bump?

mcabbott:
Thoughts on including https://github.com/FluxML/Flux.jl/issues/670#issuecomment-778583843, btw? Doesn't have to be now obviously but isn't a bad fit. I just didn't quite get the `W` conventions, right, I think.

darsnack:
I would say that since we support the ability to do that kind of initialization, this PR is still good to go. Things like [the snippet](https://github.com/FluxML/Flux.jl/issues/670#issuecomment-778583843) you linked would be nice for an overhauled initialization section in the docs. That snippet addresses some particular conventions for `Dense` but the ideas go beyond `Dense` to other layers as well. It makes more sense to me to summarize all that in a docs section at a later point.

No strong feelings though.

mcabbott:
OK, yes maybe best not to delay this with any further additions.

CarloLucibello:
Yes, I think we should have that snippet somewhere to help reproducibility, not material for this PR though. Maybe we should have a "moving from pytorch" or a "differences with pytorch" section in the docs at some point

CarloLucibello:
another bump for this, needs just an @DhairyaLGandhi's approval for merge, let's please prioritize the v0.12 stuff

CarloLucibello:
@DhairyaLGandhi close your eyes and hit approve, we are going to be fine :)

CarloLucibello (approved):

DhairyaLGandhi:
Yeah, not super comfortable with this change. I'm also not clear on what the exact benefit is, it seems like the field names etc change but was anyone asking them to be?

If all you wanted was the ability to call into dense with different symbols just do, `getproperty(d::Dense, s::Symbol) = s == :weight ? d.W : getfield(d,s)` etc.

Which is much more acceptable.

DhairyaLGandhi (approved):

Issue #939, opened 2019-11-19
subset


lihebi:
Hi, a related question, is it possible to filter a subset of parameters? The use case is to apply L2 decay on weights but not bias.

lihebi:
Sorry, miss-click

MikeInnes:
You should be able to use `filter` as normal if you just want to filter matrices vs. vectors. It'd be more robust if you just write a function to pull the weights out of dense layers explicitly, though.

lihebi:
Thanks for the reply. This relates to #160, and I accidentally hit Ctrl-Enter way before I wanted.

The params returned by `Flux.params()` does not seem to contain information about which layers they come from. I managed to pull weights from dense and conv layers:

```julia
function weight_params(m::Chain, ps=Flux.Params())
    map((l)->weight_params(l, ps), m.layers)
    ps
end
weight_params(m::Dense, ps=Flux.Params()) = push!(ps, m.W)
weight_params(m::Conv, ps=Flux.Params()) = push!(ps, m.weight)
weight_params(m::ConvTranspose, ps=Flux.Params()) = push!(ps, m.weight)
weight_params(m, ps=Flux.Params()) = ps

```

PR #1444, opened 2020-12-27
define modules function

### LINKED ISSUE #1284: How to apply L2 regularization to a subset of parameters?
When training a neural network with an L2 regularization, it is often advised not to regularize the bias parameters (in contrast with weight parameters).

I implemented this as follows in AlphaZero.jl:

```julia
regularized_params_(l) = []
regularized_params_(l::Flux.Dense) = [l.W]
regularized_params_(l::Flux.Conv) = [l.weight]

function foreach_flux_node(f::Function, x, seen = IdDict())
  Functors.isleaf(x) && return
  haskey(seen, x) && return
  seen[x] = true
  f(x)
  for child in Flux.trainable(x)
    foreach_flux_node(f, child, seen)
  end
end

function regularized_params(net::FluxNetwork)
  ps = Flux.Params()
  foreach_flux_node(net) do p
    for r in regularized_params_(p)
      any(x -> x === r, ps) || push!(ps, r)
    end
  end
  return ps
end

regularization_term(nn) = sum(sum(w .* w) for w in regularized_params(nn))

```

This feels a bit hackish though (and also it relies on internals and so it tends to break at every new Flux release).
Do you see any better way? Shouldn't we make this easier?

### LINKED ISSUE #1294: define `modules` function
.. returning an iterator over nonleaf  objects contained in a model. See #1284, the `foreach_flux_node` function, for a hot start

CarloLucibello:
A function returning an iterator over non-leaf parameters. 
Mainly motivated by the need to apply L2 regularization to weights only (see #1284 , #939),
but may be of more general use. 

Fix #1294, Fix #1284

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


DhairyaLGandhi:
Better done in with functors directly

CarloLucibello:
> Better done in with functors directly

how?

CarloLucibello:
@DhairyaLGandhi bump

src/utils.jl
```diff
@@ -471,3 +471,51 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+BatchNorm(64, λ = relu)
+Chain(Dense(784, 64), BatchNorm(64, λ = relu))
+Dense(784, 64)
+Chain(Chain(Dense(784, 64), BatchNorm(64, λ = relu)), Dense(64, 10))
+
+julia> L2(m) = 0f0
+
+julia> L2(m::Dense) = sum(abs2, m.W)
+
+julia> L2reg(model) = sum(L2(m) for m in Flux.modules(model))
```
CarloLucibello, line 516:
```suggestion
julia> L2reg(model) = sum(L2(m) for m in Flux.modules(model))

```

src/utils.jl
```diff
@@ -471,3 +471,52 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+"""
+  modules(m)
+
+Return an iterator over the non-leaf objects
+that can be reached from `m` thanks
+to [`@functor`](@ref) and [`trainable`](@ref).
+
+It can be used to apply a regularization
+over certain specific modules or subsets of
+the parameters (e.g. the weights but not the biases)/
```
ToucheSir, line 494:
```suggestion
the parameters (e.g. the weights but not the biases)
```

CarloLucibello:
@DhairyaLGandhi bump

mcabbott:
Is this the intended behaviour:
```
julia> modules(Chain(SkipConnection(Conv((2,3),4=>5;pad=6,stride=7),+),LayerNorm(8)))
8-element Vector{Any}:
 Diagonal(8)
 (6, 6, 6, 6)
 Conv((2, 3), 4=>5)
 (1, 1)
 LayerNorm(8)
 (7, 7)
 Chain(SkipConnection(Conv((2, 3), 4=>5), +), LayerNorm(8))
 SkipConnection(Conv((2, 3), 4=>5), +)
```

CarloLucibello:
Those tuples should be excluded I guess, so ideally it should look like:
```julia
julia> modules(Chain(SkipConnection(Conv((2,3),4=>5;pad=6,stride=7),+),LayerNorm(8)))
5-element Vector{Any}:
 Diagonal(8)
 Conv((2, 3), 4=>5)
 LayerNorm(8)
 Chain(SkipConnection(Conv((2, 3), 4=>5), +), LayerNorm(8))
 SkipConnection(Conv((2, 3), 4=>5), +)
 ```
(Diagonal looks weird there but it is fine, it is contained in LayerNorm).  
 Also I don't understand the ordering, it doesn't look depth first as I expected. Maybe it would be even better to have something breadth first? I'd have to think about the  implementation

**Comment by mcabbott** [comment]
I was wondering why include the containers `Chain`, `SkipConnection` too, though. Are there any containers which also have parameters (directly) which ought to be included?

I didn't think IdSet was sorted, but maybe this counts as depth-first:
```
function layers(x, cache=[])
    if x in cache || _leaflike(x)
        return cache
    end
    children = trainable(x)
    if all(_leaflike, children)
        push!(cache, x)
    else
        foreach(y -> layers(y, cache), children)
    end
    return cache
end

_leaflike(x) = isleaf(x)
_leaflike(::Tuple{Vararg{<:Number}}) = true
_leaflike(::Tuple{Vararg{<:AbstractArray}}) = true

layers(Chain(SkipConnection(Conv((2,3),4=>5;pad=6,stride=7),+),LayerNorm(8)))
layers(Chain(Dense(2,3), BatchNorm(3), LSTM(3,4)))
```

**Comment by CarloLucibello** [comment]
> I was wondering why include the containers Chain, SkipConnection too, though. Are there any containers which also have parameters (directly) which ought to be included?

Since this is supposed  to also work on generic functors defined by users, we cannot exclude mix of parameters and submodules. Also, the usage may be more generic of that of parameter extraction. In practice, I would like to replicate what is done by pytorch's [modules method](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.modules) here

**Comment by ToucheSir** [comment]
Shameless plug, but I think the conv tuple issue (and to some extent custom layer handling) will be ameliorated by https://github.com/FluxML/Functors.jl/pull/7. Most layers are marking too many fields as trainable or having to `override` trainable because they want non-trainable fields to survive `functor`ing.

**Comment by darsnack** [comment]
I think this is more inline with what I was thinking with [my comment](https://github.com/FluxML/Flux.jl/pull/1144#issuecomment-777900205) on #1144. Having something like this might make writing things like the L2 example easy enough for users to do without needing to know Functors.jl.

**Comment by DhairyaLGandhi** [comment]
I'm inclined to use functors approach here for clarity and it's not difficult to write l2 as it is.

**Comment by ToucheSir** [comment]
I'm not sure I follow, this uses functors (via trainable) under the hood? Were you thinking of something like an `fmap(reduce)` over leaf nodes?

**Comment by CarloLucibello** [comment]
@DhairyaLGandhi I asked you multiple times here (with an "how?" and two bumps) to be more explicit about the functor approach you suggest. And even if there are alternative approaches, I don't see why we cannot provide a convenience utility such as this one.

**Comment by DhairyaLGandhi** [comment]
So in the comment above, it even mentions a simple fmap(reduce), I was thinking of letting users mark which fields are params (which we already do, and is documented), we get the same effect. Marking fields as trainable can be done via the functor macro already, so I don't think there's any more need for this.

**Comment by CarloLucibello** [comment]
I'd like to see the code that solves the following problem: 
"single out weight params (and not biases)  from Dense and Conv layers of an arbitrary model (and arbitrarily nested) model so that L2 regularization can then be applied".
I can glimpse what `fmap(reduce)` could do, I don't understand what @DhairyaLGandhi says in the last comment.
So please, let's be very concrete and specific, show the code, so that we can have a proper discussion on whether it's reasonable or not that users should implement it by themselves and compare it with what they'll have after this PR:

```julia
apply_reg(m:) = false
apply_reg(m::Union{Dense,Conv}) = true
L2reg(model) = sum(norm(m.W)^2 for m in model if apply_reg(m))
```

Btw, everyone understands iterators, `fmap` is much more problematic instead, and I think we should avoid mentally overloading the users as much as we can.

DhairyaLGandhi:
So as in the docstring, you have to define what happens if one were to hit the Dense layer (and every other layer). This doesn't account for the differences in removing the bias from regularisation which is what #1284 needs.

Also, if we have to define the `L2` on every layer anyway, we might as well iterate over a chain or tuple. This doesn't add much benefit over that.

```julia
struct Mw{T,F}
  a::T
  b::F
end

@functor Mw (a,)
or trainable or defining L2 on every layer
```
takes care of handling layers specifically, but there isn't a general route that one can take.

darsnack:
L2 is just a simple toy example of the value in this PR. The core problem is I have some function `f` that applies to `Dense` and `Conv` (just examples) in a special way, and I want to apply `f` to an arbitrary model, recursing through layers until I hit all leaves, then aggregate the results of applying `f` to those leaves. There are three (reasonable) ways to go about this:

1. Pure dispatch. Define the following:
    ```julia
    f(m::Dense) = # ...
    f(m::Conv) = # ...
    f(m) = # default value (usually 0 or 1) ...
    f(m::Chain) = sum(f.(m.layers)) # some reduction operator, I used sum
    ```
    This is fine when you know `m`, but it doesn't scale to arbitrary `m`. For example, if there is a `Parallel` in the model, then this would break unless you define `f(m::Parallel)`.

2. Functors.jl. This has a couple ways I think. You can either define a new `Dense` or `Conv` where only the weight is functored then use `fmap`, or you can use `isleaf` to recurse all the way except the last stage to apply `f` on `Dense` instead of the parameters of `Dense`.

3. This PR. You only need to define `f` on the layers you care about. The recursion in (1) is handled for you through `modules`.

Writing code like the L2 example is _extremely_ common. I cannot think of a single project in Flux or PyTorch where I have not written code like this.

(1) is probably what your beginner Julia user might be able to do. It's fine for solo projects, but if you want your code to be used by others, then you need (2) or (3).

I agree with @CarloLucibello that (2) is ugly and no user should ever need to understand Functors.jl or `fmap` for such a basic code pattern. Also, there is no way I am going to define a new layer just to compute things like L2. And I don't want to override `Dense`'s `trainable` because I want it to the regular thing for the rest of my program.

(3) is basically (2) except all the ugly stuff is hidden behind `modules` allowing you to use iterators + dispatch to write very little code. IMO this PR brings the power of Functors.jl to the average Flux user. I would much prefer we merge some version of this over just documenting how to do it with `fmap`.

Now it's unclear to me whether the suggestion is to use Functors.jl directly or to change this PR to use `fmap` internally. You can't directly use `fmap` because the leaves in this tree are one layer higher than in Functors.jl. So, I don't really see exactly how you would use it internally, but even so, I think it makes more sense to define "leaf-ness" on `trainable` for this feature (which relies on Functors.jl anyways).

DhairyaLGandhi:
Sorry for the late reply I was thinking of something like this (its not tested code exactly, but worth talking about)

```julia
cantrav(x::T, fs = fieldnames(T)) where T = any(!isleaf, (getfield(x, f) for f in fs))

function k(f, m::T, fs = fieldnames(T)) where T
  cantrav(m) || return f(m)
  (k(f, getfield(m, f_)) for f_ in fs)
end
```

ToucheSir:
I'll preface this by saying that I don't see the harm in having `modules`. It should probably return a generator though.

Adapting the example from the PR:
```julia
L2(m) = 0f0
L2(m::Dense) = sum(abs2, m.W)
```
This is equivalent to 1) in @darsnack's example, but doesn't handle container types. If we add the constraint that it shouldn't have to handle containers (i.e. you shouldn't have to override for `Chain`), there are two options for calculating the total l2:

1. `L2reg(model) = sum(L2(m) for m in modules(model))`, which is given in the PR
2. `L2reg(model) = fmapreduce(L2, sum, m)`. Where `fmapreduce` is a stand-in name for a function that recurses to the `_isleaf` level and passes each node to `L2`.

Note how the only real difference here is in whether we use internal or external iteration. In fact, it's easy to see how one could implement `modules` in terms of `fmapreduce` or vice versa! The point of both is that a user should not be forced to create specific dispatches for types they don't care about or to perform their own traversal in the function being applied to each node. I personally find something like 2. more elegant because it resembles a [catamorphism](https://blog.sumtypeofway.com/posts/recursion-schemes-part-2.html) (which, being a recursion scheme, is literally a method for traversing arbitrary structures), but I don't find 1. inelegant either.

CarloLucibello:
ahah, ok, I don't think I can learn about catamorphisms on a Friday night but I'll try tomorrow

darsnack:
Is `_isleaf` meant to be like `Functors.isleaf` but one level higher? I'm okay with using (2) in the implementation of this PR. One thing to note between @DhairyaLGandhi's snippet and `fmapreduce` (assuming this is a catamorphism), is that `fmapreduce` will apply `f` to the non-leaf as well. Of course, this is easily fixed, I'm just pointing it out.

The difference between `fmapreduce` and `acc(f(i...) for i in itr)` is subjective (much like whether you prefer `mapreduce` or `acc(f(i...) for i in itr)`). Personally, I feel that `fmapreduce` belongs in Functors.jl and that the generator syntax is easier for most users to comprehend. At the very least because it is easy to print out the iterator elements with `modules` (you could `fmap(show, m)` but you get my point). Lists may be a special case of `fmap`/functors, but at the end of the day, everyone understands lists.

One thing that both the `fmapreduce` and `modules` doesn't do is intercepting the recursion at an intermediate level. In (1) from my comment, I could write something like `f(m::Parallel)` to block recursion into `Parallel` and special case. Again, you could modify both options to handle these cases. But I'm just bringing it up to highlight things to consider.

ToucheSir:
So this is a bit of a rabbit hole, but the reason I brought up recursion schemes is that it's possible to do exactly what you mention. [Part 5](https://blog.sumtypeofway.com/posts/recursion-schemes-part-5.html) of the series linked above goes into all of the nitty-gritty details.

Will users need that level of control in practice though? I think it's reasonable to ask anyone who does to handle their own recursion. We've already gone beyond pretty much every other DL framework at this point anyhow.

darsnack:
Good point. Yeah just pointing out random thoughts. I agree it doesn't come up often enough to support.

On this PR, my inclination is to have `fmapreduce` in Functors.jl then wrap that in a generator interface with `modules`.

CarloLucibello:
> One thing that both the fmapreduce and modules doesn't do is intercepting the recursion at an intermediate level. In (1) from my comment, I could write something like f(m::Parallel) to block recursion into Parallel and special case. Again, you could modify both options to handle these cases. But I'm just bringing it up to highlight things to consider.

@ToucheSir I agree we should have something in Functors not intercepting the recursion. I don't think it would solve our problems here though because we want to recurse on `trainable` (which fallbacks to `functor` but can be more specific). 
At least this is my first thought, maybe traversing `functor` is also ok. 
Also, thanks to https://github.com/FluxML/Functors.jl/pull/7 we could get rid of `trainable` at some point. 

> On this PR, my inclination is to have fmapreduce in Functors.jl then wrap that in a generator interface with modules.

@darsnack `fmapreduce` can be used to implement regularization but doesn't serve the more general-purpose that `modules` does. If we want to have something like modules (and I think most of us do) I suggest we merge this PR basically as it is, 
then if manage to cook something in Functors that we can leverage we simply switch the implementation.

CarloLucibello:
I convinced myself that the function `traverse_functor` I just defined should live in Functors.jl

darsnack:
Yes, I agree. The current `fmap` assumes you want to restructure the parts of the struct. What you wrote is different, because it doesn't assume that the output of `f` needs to be restructured. I would agree that this is useful functionality that belongs in Functors.jl. Would you be opposed to me writing an RFC with a slightly modified version as a PR to Functors.jl?

darsnack:
But I guess `fcollect` is like `fmapreduce` with a concatenation operator? I still think `fcollect` is convenient to have.

DhairyaLGandhi:
```julia
julia> m = Dense(3,3, bias = Zeros());

julia> traverse_functor(m)
4-element Array{Any,1}:
 Dense(3, 3)
 Float32[-0.929115 -0.36591858 0.539934; -0.6912145 0.43692905 -0.18630306; -0.933932 0.9196916 -0.110048935]
 Flux.Zeros()
 identity (generic function with 1 method)
```

Shouldn't it eliminate the leaves? I think I would understand if this returned 

```Julia
struct M1 a b end
struct M2 a b end
m2 = M2(M1(a, b), b2)

julia> traverse_functor(m2)
* M1(a, b)
* b2
```

darsnack:
Not necessarily for `traverse_functor` (the generic version of which is in https://github.com/FluxML/Functors.jl/pull/13). It does exclude the leaves for `modules`.

darsnack:
That's cause `M1` and `M2` aren't `@functor`'d. Since the implementation relies on Functors.jl, anything that isn't `@functor`'d cannot be recursed.

DhairyaLGandhi:
Not really - note you can't functor both of them, only M2.

darsnack:
Why can't I `@functor M1` and `@functor M2`?

darsnack:
With the implementation in https://github.com/FluxML/Functors.jl/pull/13 copied into the REPL:
```julia
julia> using Functors

julia> Functors.isleaf(x) = children(x) === ()

julia> children(x) = Functors.functor(x)[1]
children (generic function with 1 method)

julia> function fcollect(x; cache = [], exclude = v -> false)
         x in cache && return cache
         if !exclude(x)
           push!(cache, x)
           foreach(y -> fcollect(y; cache=cache, exclude=exclude), children(x))
         end
         return cache
       end
fcollect (generic function with 1 method)

julia> modules(m) = fcollect(m; exclude = Functors.isleaf)
modules (generic function with 1 method)

julia> struct M1
       a
       b
       end

julia> Functors.@functor M1

julia> struct M2
       a
       b
       end

julia> Functors.@functor M2

julia> m2 = M2(M1(1, 2), 3)
M2(M1(1, 2), 3)

julia> modules(m2)
2-element Vector{Any}:
 M2(M1(1, 2), 3)
 M1(1, 2)
```

We can discuss what `exclude` should be set to (I do think it will need to be a little more than `Functors.isleaf`).

DhairyaLGandhi:
https://github.com/FluxML/Flux.jl/issues/1504#issuecomment-778417086

darsnack:
I think you'll need to give a MWE of the relevance of that comment to layers/models which is what `modules(m)` is designed for.

DhairyaLGandhi:
When layers have components other than arrays as params...

darsnack:
To be honest, the generic solution to that is more a commentary on Functors.jl than this PR.

We could allow `modules(m; ignore = [])` to accept a list of `ignore` types which get excluded from the collected list in addition to what's already considered a leaf. Alternatively, the user can define `Functors.isleaf(::RGB) = true`.

darsnack:
I think the concern is very valid, but IMO this PR does the right thing under the current limitations of Functors.jl. The `exclude` field in `fcollect` and `fmap` make Functors.jl more flexible for use cases like this. How we expose that field to the user is a good question.

src/utils.jl
```diff
@@ -701,3 +701,44 @@ function throttle(f, timeout; leading=true, trailing=false)
...
   end
 end
+
+
+"""
+    modules(m)
+
+Return an iterator over non-leaf objects
+that can be reached from `m` through recursion
+on the children given by [`functor`](@ref).
```
darsnack, line 711:
```suggestion
Return an iterator over non-leaf objects
that can be reached by recursing `m` over
the children given by [`functor`](@ref).
```

src/utils.jl
```diff
@@ -701,3 +701,44 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+
+It can be used to apply a regularization
+over certain specific modules or subsets of
+the parameters (e.g. the weights but not the biases).
```
darsnack, line 715:
```suggestion
Useful for applying a function (e.g. a regularizer)
over specific modules or subsets of the parameters
(e.g. the weights but not the biases).
```

src/utils.jl
```diff
@@ -701,3 +701,44 @@ function throttle(f, timeout; leading=true, trailing=false)
...
+
+```jldoctest
+julia> m1 = Chain(Dense(28^2, 64), BatchNorm(64, relu))
+Chain(Dense(784, 64), BatchNorm(64, relu))
+
+julia> m2 = Chain(m1, Dense(64, 10))
+Chain(Chain(Dense(784, 64), BatchNorm(64, relu)), Dense(64, 10))
+
+julia> Flux.modules(m2)
+5-element Vector{Any}:
```
darsnack, line 727: Is `Vector{Any}` a Julia 1.6 thing? Should it be `Array{Any, 1}`?

darsnack (commented):
Just some small doc changes, but looks good to me. This is a useful utility for exposing functors to users without needing them to understand Functors.jl. I think it should be included (interested if @ToucheSir agrees).

The concern about parameters that aren't `AbstractArray{<:Number}` being leaves is something that we need to address generically with Functors.jl. It's out of scope for this PR IMO.

darsnack (approved):

DhairyaLGandhi:
I thought the function is still producing results like in https://github.com/FluxML/Flux.jl/pull/1444#issuecomment-785180590 instead of

```Julia
[M1(...),
3,
M2(...)]
```

DhairyaLGandhi:
I think that isn't something that this function should decide, instead return things at the "same level", once it finds a non leaf.

darsnack:
> you want to find the first instance you find a struct you can't look inside

If I have `Parallel(+, Chain(Dense(...), Dense(...)), Conv(...))`, why would you want it to stop recursing the `Chain` because it hit `Conv`?

> users responsibility to decide how far down we are allowed to recurse

Again, this is why `fcollect` exists.

PR #1446, opened 2020-12-30
Add inference hints to SkipConnection

DhairyaLGandhi:
Removed a bit of the docs that was a bit misleading and added inference hints to the type of `SkipConnection`.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -241,7 +241,7 @@ through a user-supplied 2-argument callable. The first argument to the callable
 will be propagated through the given `layer` while the second is the unchanged,
 "skipped" input.
 
-The simplest "ResNet"-type connection is just `SkipConnection(layer, +)`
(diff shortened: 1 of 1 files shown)


src/layers/basic.jl
```diff
@@ -241,8 +241,7 @@ through a user-supplied 2-argument callable. The first argument to the callable
 will be propagated through the given `layer` while the second is the unchanged,
 "skipped" input.
 
-The simplest "ResNet"-type connection is just `SkipConnection(layer, +)`,
-and requires the output of the layers to be the same shape as the input.
+The simplest "ResNet"-type connection is just `SkipConnection(layer, +)`
```
CarloLucibello, line 244:
```suggestion
The simplest "ResNet"-type connection is just `SkipConnection(layer, +)`.
```

CarloLucibello (dismissed):

Issue #1229, opened 2020-06-14
Multi-dimensional onehot


cossio:
Could we generalize onehot encoding in Flux to more than one or two dimensions? In general, we can have an `N` dimensional binary array, where the first dimension always has exactly one bit on and the rest are off. This represents a onehot encoding of an `N-1` dimensional array where each entry is a category.

An application is a sequence of length `L` where each site can be one of `q` letters. Then one sequence can be onehot encoded as a `q * L` matrix. But to encode a batch of `B` sequences, we need a multi-dimensional `q * L * B` array.

One possible implementation: https://github.com/cossio/OneHot.jl/blob/master/src/array.jl.

I can make a PR if people agree this can be useful in general.

CarloLucibello:
I think this could be useful. Maybe extending the current implementation 
https://github.com/FluxML/Flux.jl/blob/19b45b49d351fc530ab1022eab10c7c5507880f8/src/onehot.jl#L16
to something like
```julia
struct OneHotArray{A<:AbstractArray{OneHotVector, N-1}, N} <: AbstractArray{Bool, N}
...
```
?

Issue #864, opened 2019-09-19
onecold does not work on CuMatrix


AzamatB:
```Julia
julia> xs = rand(2,3)
2×3 Array{Float64,2}:
 0.0758553  0.779633    0.218497
 0.840066   0.00234058  0.402085

julia> Flux.onecold(xs, ["a","b"])
3-element Array{String,1}:
 "b"
 "a"
 "b"

julia> xs = cu(xs)
2×3 CuArray{Float32,2}:
 0.0758553  0.779633    0.218497
 0.840066   0.00234058  0.402085

julia> Flux.onecold(xs, ["a","b"])
ERROR: Type String does not have a definite size.
Stacktrace:
 [1] sizeof at ./essentials.jl:452 [inlined]
 [2] CuArray{String,1}(::UndefInitializer, ::Tuple{Int64}) at /home/azamat/.julia/packages/CuArrays/wXQp8/src/array.jl:38
 [3] similar at /home/azamat/.julia/packages/CuArrays/wXQp8/src/array.jl:69 [inlined]
 [4] similar(::CuArray{Float32,1}, ::Type{String}, ::Tuple{Base.OneTo{Int64}}) at ./abstractarray.jl:635
 [5] #mapslices#111(::Int64, ::typeof(mapslices), ::getfield(Flux, Symbol("##25#26")){Tuple{Array{String,1}}}, ::CuArray{Float32,2}) at ./abstractarray.jl:2009
 [6] #mapslices at ./none:0 [inlined]
 [7] onecold(::CuArray{Float32,2}, ::Array{String,1}) at /home/azamat/.julia/packages/Flux/dkJUV/src/onehot.jl:121
 [8] top-level scope at none:0
```

CarloLucibello:
It does work if you replace `["a","b"]` with numbers. Also it may be not too slow in some cases to perform the computation on the cpu (e.g. large network, i.e. slow forward pass, and move only last output to the cpu. This is what Knet does)

Naba7:
Since onecold only accepts numbers, Can I try fixing this?

darsnack:
I don't see how this works for strings:
```julia
onecold(y::OneHotMatrix, labels...) =
  mapreduce(x -> Flux.onecold(x, labels...), |, y.data, dims = 2, init = 0)
```
What does `|` do for strings? Why do we need to reduce at all? Isn't the following sufficient?
```julia
[...]

Issue #556, opened 2019-01-15
onecold is very slow


afternone:
```julia
julia> valY |> size
(3755, 128)

julia> @time onecold(valY)
 21.304120 seconds (2.40 M allocations: 95.372 MiB, 0.08% gc time)
128-element Array{Int64,1}:
   1
   2
   3
   ⋮
 126
 127
 128
```

colbec:
On my machine with Julia 1.0.3 I get:
```
julia> using Flux: onecold
julia> valY = rand(3755,128);
julia> valY |> size
(3755, 128)
julia> @time onecold(valY)
  0.838053 seconds (2.47 M allocations: 126.871 MiB, 2.98% gc time)
128-element Array{Int64,1}:
 2885
 2932
  755
...
 3122
 3128
 1780
```
and with Julia Master I get 
```
  0.001299 seconds (1.06 k allocations: 58.750 KiB)
128-element Array{Int64,1}:
 3659
 1292
...
```
so maybe there is something else happening on your machine?

afternone:
What happens when change `valY` to a CuArray?

colbec:
No GPU here. Please specify your issue more completely.

KristofferC:
And running `CuArrays.allowscalar(false)` we get the error `scalar setindex! is disallowed` which is why it is slow.

MikeInnes:
I've fixed the basic issue for vectors, but there's still the problem that we don't support `argmax` across a dimension in CuArrays. It's probably straightforward to implement that with `mapreduce`. We'll also need `mapslices`, I'm not sure if that works already.

appleparan:
Bump this issue. Now we could rewrite using `findmax` by CuArray's [#500](https://github.com/JuliaGPU/CuArrays.jl/pull/500)

h-spiess:
> The problem to do when the input is on the GPU and is a `TrackedArray` cf:
> 
> ```julia
> using Flux: onecold
> valY = rand(1000,128);
> @time onecold(valY)
> #   0.000369 seconds (1.00 k allocations: 36.375 KiB)
> @time onecold(gpu(valY))
> #   0.025580 seconds (12.65 k allocations: 1.499 MiB)
> @time onecold(Flux.param(valY));
> #  0.007966 seconds (641.86 k allocations: 22.523 MiB, 49.17% gc time)
> @time onecold(Flux.param(gpu(valY)));
[...]

PR #1447, opened 2020-12-31
new onehot implementation

### LINKED ISSUE #1445: Issues about OneHotVector/OneHotMatrix
Mentioned in #1431

There are some issues about the one hot implementation. I'll list them below.

1. Higher dimension support

Currently Flux only support 1 and 2 dimensional one hot array. However, we often need more than that. For example, Transformers usually require one hot array with shape (num label, sequence length, batch size) for parallelization. Beside Transformers, image task that do pixel level classification, like semantics segmentation, also need that.

2. Array interface

There're some array operation that would not only convert one hot array to Boolean array but also copy data back to cpu. For example, if you `hcat` two gpu `OneHotMatrix`, the result will be `Array{Bool}`. But the correct result should be `OneHotMatrix{CuArray{OneHotVector,1}}`.

3. Memory consumption 

We only support encoded numbers of labels up to 2^32 (max size of `UInt32`), but we use 64bit for `OneHotVector`. The problem appears from the use of `OneHotMatrix`, which is actually a container for `Vector{OneHotVector}`. Thus we actually use twice the memory than actually needed for storing the redundant number of the label size.

(4. There used to be some problem when using `OneHotVector` with custom CUDA kernel, but most of the seems to be gone after the CUDA update. I list this here just in case anyone encountered similar problems)

chengchingwen:
fix #1445, #1229 
probably also  #864, #556, #189

The new implementation support multi-dimensional onehot representation (#1229, #1445) with the new type `OneHotArray{K}` and `OneHot{K}`. Since the `height` of previous `OneHotMatrix` is now store on the type parameter, we can reduce the memory consumption by almost half of the origin (#1445). Most the operation are now GPU compatible (#864) without scalar operations, potentially solve the performance issues (#189, #556).

The API should be the same as I implemented them with the new type.

Remaining problem:
The backward of "embedding lookup" for GPU still require scalar operation. This can be fix with [the scatter operator](https://github.com/FluxML/NNlib.jl/pull/255).

some performance comparison:
1. #189
```julia
# the old one
julia> x = Flux.onehotbatch(rand(1:100, 50), 1:100);

julia> W = rand(128, 100);

julia> @btime W * x;
  5.021 μs (13 allocations: 50.86 KiB)

julia> cW, cx = cu(W), cu(x);

julia> @btime cW * cx;
  18.942 μs (52 allocations: 2.06 KiB)

# the new one
julia> x = Flux.onehotbatch(rand(1:100, 50), 1:100);

julia> @btime W * x;
  4.588 μs (3 allocations: 50.36 KiB)

julia> cW, cx = cu(W), cu(x);

julia> @btime cW * cx;
  8.007 μs (60 allocations: 1.67 KiB)
```

2. #556
```julia
julia> valY = randn(1000, 128);

# the old one
julia> @btime onecold(valY);
  442.836 μs (998 allocations: 36.08 KiB)

julia> @btime onecold($(gpu(valY)));
└ @ GPUArrays ~/.julia/packages/GPUArrays/eVYIC/src/host/indexing.jl:43
  6.372 ms (13804 allocations: 479.33 KiB)

# the new one
julia> @btime onecold(valY);
  552.794 μs (8 allocations: 4.00 KiB)

julia> @btime onecold($(gpu(valY)));
  24.778 μs (169 allocations: 5.56 KiB)
``` 

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -131,6 +131,7 @@ function OneHotArray{K, N, var"N+1"}(xs::AbstractArray{T, N}) where {K, N, var"N
   OneHotArray(K, xs)
 end
 
+const OneHotVector{K} = OneHot{K}
 const OneHotMatrix{K} = OneHotArray{K, 1}
 
 onehotsize(::OneHotArray{K}) where K = Int(K)
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -323,7 +323,7 @@ return [`onehot(unk, labels)`](@ref) ; otherwise the function will raise an erro
 # Examples
 ```jldoctest
 julia> Flux.onehotbatch([:b, :a, :b], [:a, :b, :c])
-3×3 Flux.OneHoArray{3, 2, Array{Flux.OneHot{0x00000003},1}}:
+3×3 Flux.OneHotArray{3, 2, Array{Flux.OneHot{0x00000003},1}}:
  0  1  0
  1  0  1
  0  0  0
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -69,6 +69,11 @@ Base.zero(o::O) where {O<:OneHot} = toOneHot(O, 0x00000000)
 Base.zero(::Type{<:OneHot{K}}) where K = OneHot(K, 0)
 Base.iszero(o::O) where {O<:OneHot} = iszero(convert(UInt32, o))
 
+# number
+
+Base.typemin(::Type{<:OneHot{K}}) where K = OneHot(K, 0)
+Base.typemax(::Type{<:OneHot{K}}) where K = OneHot(K, K)
+
 # bit-op
 
 function check_onehot_top_bit(::Type{OneHot{K}}, x) where {K}
@@ -217,7 +222,7 @@ function Base.reshape(parent::OneHotArray{K}, dims::Dims) where K
     Base._reshape(parent, dims)
 end
 
-function Base.reshape(parent::OneHotArray{K}, dims::Tuple{Vararg{Union{Colon, Int64}}}) where K
+function Base.reshape(parent::OneHotArray{K}, dims::Tuple{Vararg{Union{Colon, Int}}}) where K
   rdims = Base._reshape_uncolon(parent, dims)
   return isequal(K, first(rdims)) ?
     ohreshape(parent, Base.tail(rdims)) :
@@ -229,9 +234,43 @@ end
 import Adapt: adapt, adapt_structure
 adapt_structure(T, oa::OneHotArray) = OneHotArray(adapt(T, oa.onehots))
 
-import .CUDA: CuArray, CuArrayStyle
+import .CUDA
+import .CUDA: CuArray, CuArrayStyle, @allowscalar
 
 Base.BroadcastStyle(::Type{<: OneHotArray{K, N, var"N+1", A}}) where {K, N, var"N+1", A <: CuArray} = CuArrayStyle{var"N+1"}()
+
+# avoid scalar operation by broadcasting assigment
+CuArray(o::OneHotArray) = CuArray{Bool}(o)
+function CuArray{F}(o::OneHotArray{K, N, var"N+1", A}) where {F, K, N, var"N+1", A <: CuArray}
+  dest = similar(o, F)
+  dest .= o
+  return dest
+end
+
+using Base.Cartesian
+
+function Base.findmax(o::OneHotArray{K, N, var"N+1", A}; dims=:) where {K, N, var"N+1", A <: CuArray}
+  if dims == Colon()
+    return (true, argmax(o, dims=dims))
+  elseif isone(dims)
+    return (CUDA.ones(Bool, size(o.onehots)), argmax(o, dims=dims))
+  else
+    a = CuArray{Bool}(o)
+    return findmax(a, dims=dims)
+  end
+end
+
+function Base.argmax(o::OneHotArray{K, N, var"N+1", A}; dims=:) where {K, N, var"N+1", A <: CuArray}
+  if dims == Colon()
+    n = findfirst(!isequal(zero(eltype(o.onehots))), o.onehots)
+    x = @allowscalar o.onehots[n]
+    return CartesianIndex(Int(x), n)
+  elseif isone(dims)
+    return map((x, i)->CartesianIndex(Int(x), i), o.onehots, CartesianIndices(o.onehots)) |> Base.Fix2(reshape, (1, size(o.onehots)...))
+  else
+    return findmax(o, dims=dims)[2]
+  end
+end
 
 # api
 
@@ -307,20 +346,25 @@ julia> Flux.onecold([0.3, 0.2, 0.5], [:a, :b, :c])
 :c
 ```
 """
-onecold(y::AbstractVector, labels = 1:length(y)) = labels[Base.argmax(y)]
-
-onecold(y::AbstractMatrix, labels...) =
-  dropdims(mapslices(y -> onecold(y, labels...), y, dims=1), dims=1)
-
-function onecold(y::OneHotArray, labels = 1:onehotsize(y))
-  @assert onehotsize(y) == length(labels)
-  x = convert(AbstractArray{Int}, y.onehots)
-  if labels == 1:onehotsize(y)
-    return x
+function onecold(y::AbstractArray, labels = 1:size(y, 1))
+  @assert size(y, 1) == length(labels)
+  indices = _onecold(y)
+  if labels == 1:size(y, 1)
+    return indices
   else
-    return map(xi -> labels[xi], x)
+    if isbits(labels)
+      xs = indices
+    else
+      xs = collect(indices) # non-bit type cannot be handled by CUDA
+    end
+    return map(xi -> labels[xi], xs)
   end
 end
+
+onecold(y::AbstractVector, labels = 1:length(y)) = labels[Base.argmax(y)]
+
+_onecold(y::AbstractArray) = dropdims(map(x->Int32(x[1]), Base.argmax(y, dims=1)), dims=1)
+_onecold(y::OneHotArray) = convert(AbstractArray{Int32}, y.onehots)
 
 # AD
 @nograd OneHot, OneHotArray, onecold, onehot, onehotbatch
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -1,57 +1,246 @@
-import Base: *
+using Base: @_noinline_meta, @_inline_meta
+using Core: is_top_bit_set
+using Core.Intrinsics: bitcast, trunc_int, sext_int, zext_int, sle_int, eq_int, and_int
 
-struct OneHotVector <: AbstractVector{Bool}
-  ix::UInt32
-  of::UInt32
+abstract type AbstractOneHotArray{N} <: AbstractArray{Bool,  N} end
+
+# onehot erorr
+struct OneHotEncodeError <: Exception
+  K
+  val
+  OneHotEncodeError(@nospecialize(K), @nospecialize(val)) = (@_noinline_meta; new(K, val))
 end
 
-Base.size(xs::OneHotVector) = (Int64(xs.of),)
+function Base.showerror(io::IO, e::OneHotEncodeError)
+  print(io, "OneHotEncodeError: cannot encode ")
+  print(io, e.val)
+  print(io, " with OneHot{")
+  print(io, e.K)
+  print(io, '}')
+end
 
-Base.getindex(xs::OneHotVector, i::Integer) = i == xs.ix
+throw_onehotencode_error(K, val) = (@_noinline_meta; throw(OneHotEncodeError(K, val)))
 
-Base.getindex(xs::OneHotVector, ::Colon) = OneHotVector(xs.ix, xs.of)
+# onehot encode
+primitive type OneHot{K} <: AbstractOneHotArray{1} 32 end
 
-function Base.:*(A::AbstractMatrix, b::OneHotVector)
-  if size(A, 2) != b.of
-    throw(DimensionMismatch("Matrix column must correspond with OneHotVector size"))
+OneHot(k) = OneHot{UInt32(k)}
+OneHot{K}(x) where K = convert(OneHot(K), x)
+OneHot(k, x) = OneHot{k}(x)
+
+onehotsize(::OneHot{K}) where K = Int(K)
+
+# array interface
+
+Base.size(o::OneHot) = (onehotsize(o),)
+function Base.getindex(o::OneHot, i::I) where {I<:Integer}
+  @boundscheck checkbounds(o, i)
+  return convert(I, o) == i
+end
+
+Base.getindex(o::OneHot, i::Colon) = o
+
+Base.argmax(o::OneHot) = Int(o)
+
+# printing
+
+function Base.showarg(io::IO, x::OneHot, toplevel)
+  toplevel || print(io, "::")
+  join(io, ["OneHot{", onehotsize(x), '}'])
+end
+
+# convert
+
+Base.UInt32(o::OneHot) = bitcast(UInt32, o)
+Base.UInt64(o::OneHot) = zext_int(UInt64, o)
+Base.Int32(o::OneHot) = bitcast(Int32, o)
+Base.Int64(o::OneHot) = zext_int(Int64, o)
+
+Base.convert(::Type{Any}, o::OneHot) = o
+Base.convert(::Type{OneHot{K}}, o::OneHot{K}) where {K} = o
+Base.convert(::Type{UInt32},  o::OneHot) = UInt32(o)
+Base.convert(::Type{To}, o::OneHot) where {To} = convert(To, convert(UInt32, o))
+
+Base.convert(ot::Type{OneHot{K}}, x::Core.BuiltinInts) where {K} = toOneHot(ot, x)
+
+# zero
+
+Base.zero(o::O) where {O<:OneHot} = toOneHot(O, 0x00000000)
+Base.zero(::Type{<:OneHot{K}}) where K = OneHot(K, 0)
+Base.iszero(o::O) where {O<:OneHot} = iszero(convert(UInt32, o))
+
+# bit-op
+
+function check_onehot_top_bit(::Type{OneHot{K}}, x) where {K}
+  @_inline_meta
+  is_top_bit_set(x) && throw_onehotencode_error(K, x)
+  x
+end
+
+function check_onehot_encode(ot::Type{OneHot{K}}, x) where {K}
+  @_inline_meta
+  sle_int(x, K) || throw_onehotencode_error(K, x)
+  bitcast(ot, x)
+end
+
+function checked_onehot_trunc_sint(ot::Type{OneHot{K}}, x::From) where {K, From}
+  @_inline_meta
+  y = trunc_int(UInt32, x)
+  back = sext_int(From, y)
+  eq_int(x, back) || throw_onehotencode_error(K, x)
+  check_onehot_encode(ot, y)
+end
+
+function checked_onehot_trunc_uint(ot::Type{OneHot{K}}, x::From) where {K, From}
+  @_inline_meta
+  y = trunc_int(UInt32, x)
+  back = zext_int(From, y)
+  eq_int(x, back) || throw_onehotencode_error(K, x)
+  check_onehot_encode(ot, y)
+end
+
+toOneHot(ot::Type{OneHot{K}}, x::Int8) where {K} = check_onehot_encode(ot, sext_int(UInt32, check_onehot_top_bit(ot, x)))
+toOneHot(ot::Type{OneHot{K}}, x::Int16) where {K} = check_onehot_encode(ot, sext_int(UInt32, check_onehot_top_bit(ot, x)))
+toOneHot(ot::Type{OneHot{K}}, x::Int32) where {K} = check_onehot_encode(ot, bitcast(UInt32, check_onehot_top_bit(ot, x)))
+toOneHot(ot::Type{OneHot{K}}, x::Int64) where {K} = checked_onehot_trunc_sint(ot, check_onehot_top_bit(ot, x))
+toOneHot(ot::Type{OneHot{K}}, x::Int128) where {K} = checked_onehot_trunc_sint(ot, check_onehot_top_bit(ot, x))
+toOneHot(ot::Type{OneHot{K}}, x::UInt8) where {K} = check_onehot_encode(ot, zext_int(UInt32, x))
+toOneHot(ot::Type{OneHot{K}}, x::UInt16) where {K} = check_onehot_encode(ot, zext_int(UInt32, x))
+toOneHot(ot::Type{OneHot{K}}, x::UInt32) where {K} = check_onehot_encode(ot, x)
+toOneHot(ot::Type{OneHot{K}}, x::UInt64) where {K} = checked_onehot_trunc_uint(ot, x)
+toOneHot(ot::Type{OneHot{K}}, x::UInt128) where {K} = checked_onehot_trunc_uint(ot, x)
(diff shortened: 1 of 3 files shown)


src/onehot.jl
```diff
@@ -1,57 +1,285 @@
...
+function Base.showerror(io::IO, e::OneHotEncodeError)
+  print(io, "OneHotEncodeError: cannot encode ")
+  print(io, e.val)
+  print(io, " with OneHot{")
+  print(io, e.K)
+  print(io, '}')
+end
 
-Base.getindex(xs::OneHotVector, i::Integer) = i == xs.ix
+throw_onehotencode_error(K, val) = (@_noinline_meta; throw(OneHotEncodeError(K, val)))
```
CarloLucibello, line 22: out of curiosity, why the noinline_meta?

src/onehot.jl
```diff
@@ -1,57 +1,285 @@
...
+Base.UInt64(o::OneHot) = zext_int(UInt64, o)
+Base.Int32(o::OneHot) = bitcast(Int32, o)
+Base.Int64(o::OneHot) = zext_int(Int64, o)
+
+Base.convert(::Type{Any}, o::OneHot) = o
+Base.convert(::Type{OneHot{K}}, o::OneHot{K}) where {K} = o
+Base.convert(::Type{UInt32},  o::OneHot) = UInt32(o)
+Base.convert(::Type{To}, o::OneHot) where {To} = convert(To, convert(UInt32, o))
+
+Base.convert(ot::Type{OneHot{K}}, x::Core.BuiltinInts) where {K} = toOneHot(ot, x)
```
CarloLucibello, line 64: better not rely on julia internals

src/onehot.jl
```diff
@@ -1,57 +1,285 @@
...
+toOneHot(ot::Type{OneHot{K}}, x::Int16) where {K} = check_onehot_encode(ot, sext_int(UInt32, check_onehot_top_bit(ot, x)))
+toOneHot(ot::Type{OneHot{K}}, x::Int32) where {K} = check_onehot_encode(ot, bitcast(UInt32, check_onehot_top_bit(ot, x)))
+toOneHot(ot::Type{OneHot{K}}, x::Int64) where {K} = checked_onehot_trunc_sint(ot, check_onehot_top_bit(ot, x))
+toOneHot(ot::Type{OneHot{K}}, x::Int128) where {K} = checked_onehot_trunc_sint(ot, check_onehot_top_bit(ot, x))
+toOneHot(ot::Type{OneHot{K}}, x::UInt8) where {K} = check_onehot_encode(ot, zext_int(UInt32, x))
+toOneHot(ot::Type{OneHot{K}}, x::UInt16) where {K} = check_onehot_encode(ot, zext_int(UInt32, x))
+toOneHot(ot::Type{OneHot{K}}, x::UInt32) where {K} = check_onehot_encode(ot, x)
+toOneHot(ot::Type{OneHot{K}}, x::UInt64) where {K} = checked_onehot_trunc_uint(ot, x)
+toOneHot(ot::Type{OneHot{K}}, x::UInt128) where {K} = checked_onehot_trunc_uint(ot, x)
+toOneHot(ot::Type{OneHot{K}}, x::Bool) where {K} = and_int(zext_int(ot, x), toOneHot(ot, 0x1))
```
CarloLucibello, line 117: why dispatch `convert` to  `toOneHot`  instead of replacing these with  `convert` methods?

CarloLucibello:
This looks great. I see that the new `onecold` implementation is the same as #1441 (which I'll now close)

src/onehot.jl
```diff
@@ -95,16 +323,14 @@ return [`onehot(unk, labels)`](@ref) ; otherwise the function will raise an erro
 # Examples
 ```jldoctest
 julia> Flux.onehotbatch([:b, :a, :b], [:a, :b, :c])
-3×3 Flux.OneHotMatrix{Array{Flux.OneHotVector,1}}:
+3×3 Flux.OneHoArray{3, 2, Array{Flux.OneHot{0x00000003},1}}:
```
CarloLucibello, line 326:
```suggestion
3×3 Flux.OneHotArray{3, 2, Array{Flux.OneHot{0x00000003},1}}:
```
something wrong in the print

CarloLucibello:
The only breaking change I could spot is that the `onehot` and `onehotbatch` do not return a OneHotVector and OneHotMatrix as previously mentioned in the docs. I'm not sure how we can mitigate this. Maybe something like
```julia
const OneHotVector = OneHot
const OneHotMatrix = OneHotArray{K, 1, 2} where K
```
The parametrization of OneHotArray is a bit of a problem. 

In any case, given the impact of the change and since the PR is almost ready, I'll mark this as needed for v0.12

src/onehot.jl
chengchingwen: Do you mean the `Core.BuiltinInts`?

src/onehot.jl
chengchingwen: `toOneHot` convert the primitive types into `OneHot` with specific intrinsic bit operations. `convert` will then using `toOneHot` to convert them.

chengchingwen:
We do have `OneHotMatrix` [here](https://github.com/FluxML/Flux.jl/pull/1447/files#diff-1c736ea3192ebf1a92456dea05a413200e35f0575eae282e7232c51fa0a33febR134). I can also add the `OneHotVector` one.

chengchingwen:
> The parametrization of OneHotArray is a bit of a problem.

@CarloLucibello  Could you elaborate more on this?

src/onehot.jl
chengchingwen: I guess it's because we want to prevent the `throw` being inlined, but I'm not really sure. 

I follow how they do `throw_inexacterror` in `Base`:
https://github.com/JuliaLang/julia/blob/5ee2d601edcf2dcdd467041ba9f08c43bf23e1a3/base/boot.jl#L605

src/onehot.jl
CarloLucibello: yes

src/onehot.jl
CarloLucibello: yes, I was wondering why you define `toOneHot` at all, instead of 
```julia
Base.convert(ot::Type{OneHot{K}}, x::UInt32) where {K} = check_onehot_encode(ot, x)
Base.convert(ot::Type{OneHot{K}}, x::UInt64) where {K} = checked_onehot_trunc_uint(ot, x)
....
```

CarloLucibello:
> Could you elaborate more on this?

never mind, you solved the issue

src/onehot.jl
```diff
@@ -1,57 +1,286 @@
...
+OneHotArray{K, N}(xs::AbstractArray{T, N}) where {K, N, T} = OneHotArray{K, N, N+1}(xs)
+function OneHotArray{K, N, var"N+1"}(xs::AbstractArray{T, N}) where {K, N, var"N+1", T}
+  @assert N+1 == var"N+1"
+  OneHotArray(K, xs)
+end
 
-# remove workaround when https://github.com/JuliaGPU/CuArrays.jl/issues/676 is fixed
-A::AbstractMatrix * B::OneHotMatrix = A[:, cpu(map(x->x.ix, B.data))]
+const OneHotVector{K} = OneHot{K}
+const OneHotMatrix{K} = OneHotArray{K, 1}
```
CarloLucibello, line 135:
```suggestion
const OneHotMatrix{K, A} = OneHotArray{K, 1, 2, A}
```

chengchingwen: Why `A`?

src/onehot.jl
chengchingwen: The `Core.BuiltinInts` is just a alias for `Union{Bool, Int32, Int64, UInt32, UInt64, UInt8, Int128, Int16, Int8, UInt128, UInt16}`. I can replace it with the `Union` if you think that would be better

src/onehot.jl
CarloLucibello: why should we exclude it?

src/onehot.jl
darsnack: Sorry for jumping in, but it seems better to take Carlo's suggestion below and directly define `convert` for each member of the union instead of `toOneHot`.

src/onehot.jl
```diff
@@ -1,57 +1,286 @@
...
+toOneHot(ot::Type{OneHot{K}}, x::Int128) where {K} = checked_onehot_trunc_sint(ot, check_onehot_top_bit(ot, x))
+toOneHot(ot::Type{OneHot{K}}, x::UInt8) where {K} = check_onehot_encode(ot, zext_int(UInt32, x))
+toOneHot(ot::Type{OneHot{K}}, x::UInt16) where {K} = check_onehot_encode(ot, zext_int(UInt32, x))
+toOneHot(ot::Type{OneHot{K}}, x::UInt32) where {K} = check_onehot_encode(ot, x)
+toOneHot(ot::Type{OneHot{K}}, x::UInt64) where {K} = checked_onehot_trunc_uint(ot, x)
+toOneHot(ot::Type{OneHot{K}}, x::UInt128) where {K} = checked_onehot_trunc_uint(ot, x)
+toOneHot(ot::Type{OneHot{K}}, x::Bool) where {K} = and_int(zext_int(ot, x), toOneHot(ot, 0x1))
+
+# onehot array
+struct OneHotArray{K, N, var"N+1", A<:AbstractArray{OneHot{K}, N}} <: AbstractOneHotArray{var"N+1"}
```
darsnack, line 120: I'm curious what the purpose of `var` is here, since I've never encountered it before.

src/onehot.jl
chengchingwen: Because that's not usually something people have to handle. The value of `A` is handled by the `OneHotArray`'s outer constructor, excluding `A` will allow the constructor to be correctly dispatched (avoid using the default constructor).

src/onehot.jl
CarloLucibello: `Core.BuiltinInts` is undocumented, so it could be changed or moved in next julia releases (although it seems highly unlikely). Better define the union ourselves

src/onehot.jl
chengchingwen: Readability only. The `var_string` create a `Symbol` of the given string and use it as variable name, so `var"N+1"` is actually trying to use `Symbol("N+1")` as the type variable. I think it's more clear to use `var"N+1"` instead of other symbol like `N2`, etc.

src/onehot.jl
chengchingwen: I can't really recall why they look the same as `convert` but in separate method. Originally I try to follow [the code in `Base`](https://github.com/JuliaLang/julia/blob/master/base/boot.jl#L639-L748). I'll check if they can be changed without any issues. if so, then the `Core.BuiltinInts` issue could also be fixed.

src/onehot.jl
darsnack: Okay that makes sense. But is `var"N+1"` ever not actually `N+1`?

src/onehot.jl
CarloLucibello: I may be wrong, but I think that adding `A` is not going to change the constructor dispatch at all, and has the advantage of more convenient function dispatch:
```julia
f(x::OneHotMatrix{K, <:CuArray}) where K = ...
```

src/onehot.jl
```diff
@@ -1,57 +1,286 @@
...
+  print(io, e.K)
+  print(io, '}')
+end
 
-Base.getindex(xs::OneHotVector, i::Integer) = i == xs.ix
+throw_onehotencode_error(K, val) = (@_noinline_meta; throw(OneHotEncodeError(K, val)))
 
-Base.getindex(xs::OneHotVector, ::Colon) = OneHotVector(xs.ix, xs.of)
+# onehot encode
+primitive type OneHot{K} <: AbstractOneHotArray{1} 32 end
```
DhairyaLGandhi, line 25: Why do we need to define a primitive here instead of making OneHotArray general across dimensions?

src/onehot.jl
chengchingwen: It's achievable with the default constructor (manually fit something that is not actually `N+1`). But with outer constructor? no.

I think is possible to ban that by replacing the default constructor, but I'm not sure whether it's a good idea.

DhairyaLGandhi (commented):
I'm curious why not generalise the current implementation. Seems the performance was gained by avoiding the scalar indexing on the internal vector in OneHotMatrix, we could just rewrite that?

src/onehot.jl
```diff
@@ -1,57 +1,286 @@
...
+    Base._cat(Val(1), xss...)
+  else
+    predecessor(::Val{V}) where V = Val(V-1)
+    predecessor(v) = v - 1
+    sdims = predecessor(dims)
+    xidss = map(xs->xs.onehots, xss)
+    ret = cat(xidss...; dims=sdims)
+    OneHotArray(ret)
+  end
+end
```
CarloLucibello, line 209: not sure, but probably we can remove all the `Val`s and just use Ints,
I think now the compiler is smart enough to perform the same optimizations

Also, since this doesn't support multiple dims, we should restrict the signature to
```julia
 Base.cat(xss::OneHotArray{K}...; dims::Int)
```

src/onehot.jl
```diff
@@ -1,57 +1,286 @@
-import Base: *
+using Base: @_noinline_meta, @_inline_meta
+using Core: is_top_bit_set
+using Core.Intrinsics: bitcast, trunc_int, sext_int, zext_int, sle_int, eq_int, and_int
```
DhairyaLGandhi, line 3: we should avoid using internal functions

src/onehot.jl
```diff
@@ -1,57 +1,286 @@
...
+function ohreshape(parent::OneHotArray{K}, dims) where K
+  onehots = parent.onehots
+  OneHotArray(reshape(onehots, dims))
+end
+
+function Base.reshape(parent::OneHotArray{K}, dims::Dims) where K
+  isequal(prod(dims), length(parent)) || throw(DimensionMismatch("new dimensions $(dims) must be consistent with array size $(length(parent))"))
+  return isequal(K, first(dims)) ?
+    ohreshape(parent, Base.tail(dims)) :
+    Base._reshape(parent, dims)
```
CarloLucibello, line 223: also here we shouldn't use julia's internals

src/onehot.jl
darsnack: I guess the question is why store `N + 1` at all if there is no use case where it isn't literally `N + 1`.

src/onehot.jl
darsnack: I had the same question. The primitive type avoids the pointer allocation of a wrapper type is my guess.

But I think the need for a primitive type/array of one-hots comes from the design choice of having a one-hot vector as a unique type that arbitrary dimension arrays are built off. Instead of designing a type for the 1D case first, having a N-D one-hot array type as the base seems like an alternate design choice. I am currently prototyping this.

src/onehot.jl
darsnack: Never mind, I didn't realize that wasn't possible in Julia.

darsnack:
Here is a slightly different approach (which I think @DhairyaLGandhi was alluding to). Fundamentally, both this PR and the old Flux code created one-hot arrays by first designing a one-hot vector then wrapping that type in a multi-dimensional array (or matrix only for the old code). Instead, I design first for the notion of a N-dimensional one-hot array, then I make `OneHotVector` a specialized case (similar to `AbstractArray`).

```julia
const OneHotIndex{T, N} = Union{T, AbstractArray{T, N}}

struct OneHotArray{T<:Integer, L, N, var"N+1", I<:OneHotIndex{T, N}} <: AbstractArray{Bool, var"N+1"}
    indices::I
end

const OneHotVector{T, L} = OneHotArray{T, L, 0, 1, T}
const OneHotMatrix{T, L, I} = OneHotArray{T, L, 1, 2, I}

Base.size(x::OneHotArray{<:Any, L}) where L = (L, size(x.indices)...)

_onehotindex(x, i) = (x == i)

Base.getindex(x::OneHotArray, i, I...) = _onehotindex.(x.indices[I...], i)
Base.getindex(x::OneHotVector, i) = _onehotindex(x.indices, i)
```

Like this PR, the approach above should enjoy all the same advantages:
- arbitrary dimensional support
- better handling for the array interface (e.g. `hcat` on the GPU)
- easy to use Adapt.jl to map storage
- lower memory consumption
    - this implementation is actually even better, since the user can use less that 32 bits to represent a one-hot vector if they want to

But using this approach means we don't need to define a new primitive type or use any `Core` functions. I will try and submit a PR today so we can compare these approaches for performance. If we can recover the performance of this PR with the new approach, then I think we should use the new approach (it is more intuitive and easier to maintain in imo).

chengchingwen:
There are a few thoughts about some decisions, so I'll make some explanation here.

The core idea was to develop a general representation that's both semantically meaningful and GPU friendly. So why using primitive type instead of extending the origin `OneHotVector`? I do use that solution before. Actually, if you checkout to some older version of  Transformers.jl like v0.1.1, you can find the code. The problem is that CUDAnative.jl and CuArrays.jl are actually better at handling primitive type than some custom struct that looks almost like primitive one. I wrote a lot of workaround just to avoid memory copy and scalar indexing. Sometimes there were even some functions that cannot be compiled with CUDAnative.jl. That's why I started to use primitive type for the onehot representation. The use of parametric type avoid storing duplicate values. The use of primitive type allow the value to be easily handled by GPU with less operations and memory needed, and even make `reintrepret` possible. As @darsnack  said, it's somewhat possible to use a single wrapper around integer and integer array. However, I personally think that solution is not semantically meaningful  enough, especially when it comes to `OneHotVector`. It would be confusing when you write specialized code. And both 0-dimensional array and Union-parameterized struct suffer from unnecessary allocations, and the potential of type unstability. 

However, primitive type does introduce lots of Julia's internal functions into the scope. Currently we need to use `Core.Intrinsics` to successfully define a custom primitive type. I know there are some potential risk on using internal functions, but on the other hand I think the intrinsic APIs are less prone to be change greatly for 2 reasons. First, since all code that introduce new primitive type are using the same APIs, even Julia's base primitive are defined with them (see [base/boot.jl](https://github.com/JuliaLang/julia/blob/master/base/boot.jl), [base/float.jl](https://github.com/JuliaLang/julia/blob/master/base/float.jl), [etc](https://github.com/JuliaLang/julia/search?l=Julia&q=primitive).), the change of these APIs will become some breaking changes. Second, the intrinsics are basically the direct translation of llvm IR, tbh there're not much things to change. Therefore, I believe the use of the `Core.Intrinsics` and some other relative internal function are quite harmless.

src/onehot.jl
chengchingwen: `Val` is also used in Julia `cat` relative function (see [base/abstractarray.jl](https://github.com/JuliaLang/julia/blob/master/base/abstractarray.jl)). Those definition are use to be compatible with Julia abstract array interface

src/onehot.jl
chengchingwen: Avoid pointer allocation is part of the reason. See my comment on this PR for the throughout explanation

src/onehot.jl
chengchingwen: The type var `K` is handle with `UInt32` but currently Julia cannot specific the type of type var.  Therefore, I use constructor of partial specialized type (like `OneHotMatrix{K, 1}`) to check/convert the value. Calling `OneHotArray{K, 1, 2, A}` will result in calling the default constructor, which will bypass the conversion and result in type error. If we are going the use this kind of expression then I will try to replace the default constructor.

src/onehot.jl
chengchingwen: `Base._reshape` is use to avoid redefine all the reshape dim patterns and handle the reshape of `AbstractArray{Bool}`. It's quite common for defining custom array (like [FillArrays](https://github.com/JuliaArrays/FillArrays.jl/blob/master/src/FillArrays.jl#L214-L225))

darsnack:
Thanks for the explanation!

> The problem is that CUDAnative.jl and CuArrays.jl are actually better at handling primitive type than some custom struct that looks almost like primitive one.

I don't particularly understand this point, so do you mind elaborating? Why would CUDA handle an array of `UInt32`s differently from an array of `OneHot`s (where `OneHot` is just an unsigned 32-bit integer)?

> And both 0-dimensional array and Union-parameterized struct suffer from unnecessary allocations, and the potential of type unstability.

I'm also a little confused here. The difference between `OneHotArray` here and in #1448 is that the type of each integer is a `OneHot` here and in the #1448 it is a concrete subtype of `Integer`.

Also what do you mean by 0-dimensional array? For one-hot arrays, there is always at least one dimension (the first). There's nothing like a scalar one-hot vector. Unless you are referring to the dimension of the internal integer array?

----

Do you mind suggesting some more performance tests that illustrate the issues you had to solve with CUDA? I'd like to add them to #1448 for comparison. It's entirely possible that the GPUCompiler does something better with the primitive type, but fundamentally there doesn't seem like there is a reason to do so.

chengchingwen:
> I don't particularly understand this point, so do you mind elaborating? Why would CUDA handle an array of UInt32s differently from an array of OneHots (where OneHot is just an unsigned 32-bit integer)

I'm talking about the old `OneHotVector` which is a struct of two `UInt32` and can be directly treat as a `UInt64`.

> Also what do you mean by 0-dimensional array? For one-hot arrays, there is always at least one dimension (the first). There's nothing like a scalar one-hot vector. Unless you are referring to the dimension of the internal integer array?

In Julia you can define a 0-dimension array with something like `ones()` or Array{Float64, 0}(undef). It's an array of type `Array{Float64, 0}`,  which is an unpleasant but workable way. As you can define `OneHotVector` with underlying integer array of type `Array{Int32, 0}`.

> I'm also a little confused here. The difference between OneHotArray here and in #1448 is that the type of each integer is a OneHot here and in the #1448 it is a concrete subtype of Integer.

I was referring to the `const OneHotIndex{T, N} = Union{T, AbstractArray{T, N}}` in #1448 with `I<:OneHotIndex{T, N}` as the "Union-parameterized struct". Given `I` is just `T`, we still allocate the wrapper `OneHotArray` even it's just a single value and introduce a `getfield` operation which requires a few IRs.

darsnack:
Ah I understand your point. Yeah the corner case of a single one-hot vector will incur a wrapper overhead, but I'm curious where that is a problem in practice. If you have a single one-hot vector, then there is a total of two integers of memory being used, so it doesn't seem like a problem. If you have many one-hot vectors stored in an array, then it makes more sense to batch them. Is there a use-case where you don't want to batch, but you do need many one-hot vectors?

> On the other hand, since there is an parameterized with Union, it's possible that some function might introduce type unstability.

The `Union` is only used to restrict the `I` type parameter of `OneHotArray` to be a concrete subtype of `OneHotIndex`. For an instance of `OneHotArray`, `I` should be a stable type for inference (I think)? I could add some `@test_inferred` if this is a concern.

chengchingwen:
>  If you have a single one-hot vector, then there is a total of two integers of memory being used, so it doesn't seem like a problem. If you have many one-hot vectors stored in an array, then it makes more sense to batch them. Is there a use-case where you don't want to batch, but you do need many one-hot vectors?

0-dim array allocated on heap, and struct would depend on the compiler. Primitive type is special because it can always be simply passed by value after compilation, so no potential edge case. I agree that it makes more sense to batch them, but sometimes you can only iterate thought that sequence of one-hots (for some really dynamic and un-batch-able models). They are rare but exist, so I tend to rip out the potential performance issue if possible.

src/onehot.jl
CarloLucibello: it's enough to use 
```suggestion
   reshape(parent, dims)
```
then Base will handle that
https://github.com/JuliaLang/julia/blob/0f0bc294f7b18d685c4998ada9c2c33841a74f2d/base/reshapedarray.jl#L112

chengchingwen: we are overloading it, that would result in infinite recursion.

chengchingwen: That line is exactly the reason I use `Base._reshape`

PR #1448, opened 2021-01-01
Arbitrary dimension one-hot arrays

### LINKED ISSUE #1445: Issues about OneHotVector/OneHotMatrix

darsnack:
This supersedes #1447. It should address the same issues:
- fix #1445, #1229
- probably fix also #864, #556, #189

This PR introduces a new one-hot N-dimensional array type, `OneHotArray`. Like #1447, this approach avoids the pointer allocations associated with `OneHotMatrix` being an array of `OneHotVector`s. It also lifts the "height" into the type parameter to avoid unnecessary allocation. Unlike #1447, this approach does not introduce a new primitive type. Instead, a "one-hot vector" is represented with a single subtype of `Integer` that is configurable by the user. By default, the exposed API will use `UInt32`.

Fundamentally, the primitive type is necessary because wrapping a `UInt32` as a `OneHotVector` will suffer memory penalties when you create an `Array{<:OneHotVector}`. But if we begin by designing for N-dimensions, then `OneHotVector` is just the specialized 1D case (similar to how `Vector{T} = Array{T, 1}`).

## Performance

I compared against the same tests mentioned in #1447. Please suggest more if you want to.

1. #189
```jl
#master
julia> x = Flux.onehotbatch(rand(1:100, 50), 1:100);

julia> W = rand(128, 100);

julia> @btime $W * $x;
  5.095 μs (13 allocations: 50.86 KiB)

julia> cW, cx = cu(W), cu(x);

julia> @btime $cW * $cx;
  24.948 μs (86 allocations: 3.11 KiB)

#1447
julia> x = Flux.onehotbatch(rand(1:100, 50), 1:100);

julia> W = rand(128, 100);

julia> @btime $W * $x;
  5.312 μs (3 allocations: 50.36 KiB)

julia> cW, cx = cu(W), cu(x);

julia> @btime $cW * $cx;
  8.466 μs (61 allocations: 1.69 KiB)

# this PR
julia> x = Flux.onehotbatch(rand(1:100, 50), 1:100);

julia> W = rand(128, 100);

julia> @btime $W * $x;
  4.708 μs (3 allocations: 50.56 KiB)

julia> cW, cx = cu(W), cu(x);

julia> @btime $cW * $cx;
  8.576 μs (63 allocations: 1.73 KiB)
```

2. #556
```jl
#master
julia> valY = randn(1000, 128);

julia> @btime Flux.onecold($valY);
  365.712 μs (1131 allocations: 38.16 KiB)

julia> @btime Flux.onecold($(gpu(valY)));
└ @ GPUArrays ~/.julia/packages/GPUArrays/jhRU7/src/host/indexing.jl:43
  1.330 s (781248 allocations: 31.59 MiB)

#1447
julia> valY = randn(1000, 128);

julia> @btime Flux.onecold($valY);
  524.767 μs (8 allocations: 4.00 KiB)

julia> @btime Flux.onecold($(gpu(valY)));
  27.563 μs (169 allocations: 5.56 KiB)

# this PR
julia> valY = randn(1000, 128);

julia> @btime Flux.onecold($valY);
  493.017 μs (8 allocations: 4.53 KiB)

julia> @btime Flux.onecold($(gpu(valY)));
  26.702 μs (171 allocations: 5.61 KiB)
```

## Summary

This should basically be #1447 but simpler to maintain w/ fewer changes. Tests are passing, though I think we should add more tests for one-hot data (currently our test set seems pretty sparse). Performance matches #1447 where I have tested, but please suggest more performance tests. In theory, any performance difference between #1447 and this PR should be recoverable.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md
- [ ] Final review from @DhairyaLGandhi (for API changes).

cc @CarloLucibello @chengchingwen

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -41,7 +41,8 @@ Base.hcat(xs::OneHotArray...) = cat(xs...; dims = 2)
 Base.vcat(xs::OneHotArray...) = cat(xs...; dims = 1)
 
 Base.reshape(x::OneHotArray{<:Any, L}, dims...) where L =
-  (first(dims) == L) ? OneHotArray(L, reshape(x.indices, dims[2:end]...)) : reshape(x, dims...)
+  (first(dims) == L) ? OneHotArray(L, reshape(x.indices, dims[2:end]...)) :
+                       reshape(convert(_onehot_bool_type(x), x), dims...)
 
 batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(L, _indices.(xs))
 
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -1,9 +1,7 @@
 import Adapt
 import .CUDA
 
-const OneHotIndex{T, N} = Union{T, AbstractArray{T, N}}
-
-struct OneHotArray{T<:Integer, L, N, var"N+1", I<:OneHotIndex{T, N}} <: AbstractArray{Bool, var"N+1"}
+struct OneHotArray{T<:Integer, L, N, var"N+1", I<:Union{T, AbstractArray{T, N}}} <: AbstractArray{Bool, var"N+1"}
   indices::I
 end
 OneHotArray{T, L, N, I}(indices) where {T, L, N, I} = OneHotArray{T, L, N, N+1, I}(indices)
@@ -15,18 +13,22 @@ _indices(x::OneHotArray) = x.indices
 const OneHotVector{T, L} = OneHotArray{T, L, 0, 1, T}
 const OneHotMatrix{T, L, I} = OneHotArray{T, L, 1, 2, I}
 
+OneHotVector(L, idx) = OneHotArray(L, idx)
+OneHotMatrix(L, indices) = OneHotArray(L, indices)
+
 Base.size(x::OneHotArray{<:Any, L}) where L = (Int(L), size(x.indices)...)
 
 _onehotindex(x, i) = (x == i)
 
 Base.getindex(x::OneHotVector, i::Integer) = _onehotindex(x.indices, i)
-Base.getindex(x::OneHotVector{T, L}, ::Colon) where {T, L} = OneHotVector{T, L}(x.indices)
+Base.getindex(x::OneHotVector{T, L}, ::Colon) where {T, L} = x
 
 Base.getindex(x::OneHotArray, i::Integer, I...) = _onehotindex.(x.indices[I...], i)
 Base.getindex(x::OneHotArray{<:Any, L}, ::Colon, I...) where L = OneHotArray(L, x.indices[I...])
+Base.getindex(x::OneHotArray{<:Any, <:Any, <:Any, N}, ::Vararg{Colon, N}) where N = x
 Base.getindex(x::OneHotArray, I::CartesianIndex{N}) where N = x[I[1], Tuple(I)[2:N]...]
 
-_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:OneHotIndex}) where N = Array{Bool, N}
+_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:Union{Integer, AbstractArray}}) where N = Array{Bool, N}
 _onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}) where N = CuArray{Bool, N}
 
 function Base.cat(xs::OneHotArray{<:Any, L}...; dims::Int) where L
@@ -48,7 +50,7 @@ batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(L, _ind
 
 Adapt.adapt_structure(T, x::OneHotArray{<:Any, L}) where L = OneHotArray(L, adapt(T, x.indices))
 
-Base.BroadcastStyle(::Type{<:OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}}) where N = CuArrayStyle{N}()
+Base.BroadcastStyle(::Type{<:OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}}) where N = CUDA.CuArrayStyle{N}()
 
 Base.argmax(x::OneHotArray; dims = Colon()) =
   (dims == 1) ? reshape(CartesianIndex.(x.indices, CartesianIndices(x.indices)), 1, size(x.indices)...) :
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -124,11 +124,14 @@ julia> Flux.onecold([0.3, 0.2, 0.5], [:a, :b, :c])
 """
 onecold(y::AbstractVector, labels = 1:length(y)) = labels[argmax(y)]
 function onecold(y::AbstractArray, labels = 1:size(y, 1))
-  indices = dropdims(argmax(y; dims = 1); dims = 1)
+  indices = convert(Array, _fast_argmax(y))
   xs = isbits(labels) ? indices : collect(indices) # non-bit type cannot be handled by CUDA
 
   return map(xi -> labels[xi[1]], xs)
 end
+
+_fast_argmax(x::AbstractArray) = dropdims(argmax(x; dims = 1); dims = 1)
+_fast_argmax(x::OneHotArray) = x.indices
 
 @nograd OneHotArray, onecold, onehot, onehotbatch
 
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -40,9 +40,9 @@ end
 Base.hcat(xs::OneHotArray...) = cat(xs...; dims = 2)
 Base.vcat(xs::OneHotArray...) = cat(xs...; dims = 1)
 
-Base.reshape(x::OneHotArray{<:Any, L}, dims...) where L =
+Base.reshape(x::OneHotArray{<:Any, L}, dims::Dims) where L =
   (first(dims) == L) ? OneHotArray(L, reshape(x.indices, dims[2:end]...)) :
-                       reshape(convert(_onehot_bool_type(x), x), dims...)
+                       throw(ArgumentError("Cannot reshape OneHotArray if first(dims) != size(x, 1)"))
 
 batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(L, _indices.(xs))
 
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -1,52 +1,57 @@
-import Base: *
+import Adapt
+import .CUDA
 
-struct OneHotVector <: AbstractVector{Bool}
-  ix::UInt32
-  of::UInt32
+const OneHotIndex{T, N} = Union{T, AbstractArray{T, N}}
+
+struct OneHotArray{T<:Integer, L, N, var"N+1", I<:OneHotIndex{T, N}} <: AbstractArray{Bool, var"N+1"}
+  indices::I
 end
+OneHotArray{T, L, N, I}(indices) where {T, L, N, I} = OneHotArray{T, L, N, N+1, I}(indices)
+OneHotArray(L::Integer, indices::T) where {T<:Integer} = OneHotArray{T, L, 0, T}(indices)
+OneHotArray(L::Integer, indices::AbstractArray{T, N}) where {T, N} = OneHotArray{T, L, N, typeof(indices)}(indices)
 
-Base.size(xs::OneHotVector) = (Int64(xs.of),)
+_indices(x::OneHotArray) = x.indices
 
-Base.getindex(xs::OneHotVector, i::Integer) = i == xs.ix
+const OneHotVector{T, L} = OneHotArray{T, L, 0, 1, T}
+const OneHotMatrix{T, L, I} = OneHotArray{T, L, 1, 2, I}
 
-Base.getindex(xs::OneHotVector, ::Colon) = OneHotVector(xs.ix, xs.of)
+Base.size(x::OneHotArray{<:Any, L}) where L = (Int(L), size(x.indices)...)
 
-function Base.:*(A::AbstractMatrix, b::OneHotVector)
-  if size(A, 2) != b.of
-    throw(DimensionMismatch("Matrix column must correspond with OneHotVector size"))
+_onehotindex(x, i) = (x == i)
+
+Base.getindex(x::OneHotVector, i::Integer) = _onehotindex(x.indices, i)
+Base.getindex(x::OneHotVector{T, L}, ::Colon) where {T, L} = OneHotVector{T, L}(x.indices)
+
+Base.getindex(x::OneHotArray, i::Integer, I...) = _onehotindex.(x.indices[I...], i)
+Base.getindex(x::OneHotArray{<:Any, L}, ::Colon, I...) where L = OneHotArray(L, x.indices[I...])
+Base.getindex(x::OneHotArray, I::CartesianIndex{N}) where N = x[I[1], Tuple(I)[2:N]...]
+
+_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:OneHotIndex}) where N = Array{Bool, N}
+_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}) where N = CuArray{Bool, N}
+
+function Base.cat(xs::OneHotArray{<:Any, L}...; dims::Int) where L
+  if isone(dims)
+    return cat(map(x -> convert(_onehot_bool_type(x), x), xs)...; dims = 1)
+  else
+    return OneHotArray(L, cat(_indices.(xs)...; dims = dims = dims - 1))
   end
-  return A[:, b.ix]
 end
 
-struct OneHotMatrix{A<:AbstractVector{OneHotVector}} <: AbstractMatrix{Bool}
-  height::Int
-  data::A
-end
+Base.hcat(xs::OneHotArray...) = cat(xs...; dims = 2)
+Base.vcat(xs::OneHotArray...) = cat(xs...; dims = 1)
 
-Base.size(xs::OneHotMatrix) = (Int64(xs.height),length(xs.data))
+Base.reshape(x::OneHotArray{<:Any, L}, dims...) where L =
+  (first(dims) == L) ? OneHotArray(L, reshape(x.indices, dims[2:end]...)) : reshape(x, dims...)
 
-Base.getindex(xs::OneHotMatrix, i::Union{Integer, AbstractVector}, j::Integer) = xs.data[j][i]
-Base.getindex(xs::OneHotMatrix, ::Colon, i::Integer) = xs.data[i]
-Base.getindex(xs::OneHotMatrix, ::Colon, i::AbstractArray) = OneHotMatrix(xs.height, xs.data[i])
-Base.getindex(xs::OneHotMatrix, ::Colon, ::Colon) = OneHotMatrix(xs.height, copy(xs.data))
+batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(L, _indices.(xs))
 
-Base.getindex(xs::OneHotMatrix, i::Integer, ::Colon) = map(x -> x[i], xs.data)
+Adapt.adapt_structure(T, x::OneHotArray{<:Any, L}) where L = OneHotArray(L, adapt(T, x.indices))
 
-# remove workaround when https://github.com/JuliaGPU/CuArrays.jl/issues/676 is fixed
-A::AbstractMatrix * B::OneHotMatrix = A[:, cpu(map(x->x.ix, B.data))]
+Base.BroadcastStyle(::Type{<:OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}}) where N = CuArrayStyle{N}()
 
-Base.hcat(x::OneHotVector, xs::OneHotVector...) = OneHotMatrix(length(x), [x, xs...])
-
-batch(xs::AbstractArray{<:OneHotVector}) = OneHotMatrix(length(first(xs)), xs)
-
-import Adapt: adapt, adapt_structure
-
-adapt_structure(T, xs::OneHotMatrix) = OneHotMatrix(xs.height, adapt(T, xs.data))
-
-import .CUDA: CuArray, CuArrayStyle, cudaconvert
-import Base.Broadcast: BroadcastStyle, ArrayStyle
-BroadcastStyle(::Type{<:OneHotMatrix{<:CuArray}}) = CuArrayStyle{2}()
-cudaconvert(x::OneHotMatrix{<:CuArray}) = OneHotMatrix(x.height, cudaconvert(x.data))
+Base.argmax(x::OneHotArray; dims = Colon()) =
+  (dims == 1) ? reshape(CartesianIndex.(x.indices, CartesianIndices(x.indices)), 1, size(x.indices)...) :
+                argmax(convert(_onehot_bool_type(x), x); dims = dims)
 
 """
     onehot(l, labels[, unk])
@@ -60,13 +65,13 @@ If `l` is not found in labels and  `unk` is present, the function returns
 # Examples
 ```jldoctest
 julia> Flux.onehot(:b, [:a, :b, :c])
-3-element Flux.OneHotVector:
+3-element Flux.OneHotArray{UInt32,3,0,1,UInt32}:
  0
  1
  0
 
 julia> Flux.onehot(:c, [:a, :b, :c])
-3-element Flux.OneHotVector:
+3-element Flux.OneHotArray{UInt32,3,0,1,UInt32}:
  0
  0
  1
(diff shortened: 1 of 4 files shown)


src/onehot.jl
```diff
@@ -1,52 +1,57 @@
-import Base: *
+import Adapt
+import .CUDA
 
-struct OneHotVector <: AbstractVector{Bool}
-  ix::UInt32
-  of::UInt32
+const OneHotIndex{T, N} = Union{T, AbstractArray{T, N}}
+
+struct OneHotArray{T<:Integer, L, N, var"N+1", I<:OneHotIndex{T, N}} <: AbstractArray{Bool, var"N+1"}
```
DhairyaLGandhi, line 6: Can you explain what the type params are here?

src/onehot.jl
```diff
@@ -1,52 +1,57 @@
...
 
-Base.getindex(xs::OneHotMatrix, i::Union{Integer, AbstractVector}, j::Integer) = xs.data[j][i]
-Base.getindex(xs::OneHotMatrix, ::Colon, i::Integer) = xs.data[i]
-Base.getindex(xs::OneHotMatrix, ::Colon, i::AbstractArray) = OneHotMatrix(xs.height, xs.data[i])
-Base.getindex(xs::OneHotMatrix, ::Colon, ::Colon) = OneHotMatrix(xs.height, copy(xs.data))
+_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:OneHotIndex}) where N = Array{Bool, N}
+_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}) where N = CuArray{Bool, N}
 
-Base.getindex(xs::OneHotMatrix, i::Integer, ::Colon) = map(x -> x[i], xs.data)
+function Base.cat(xs::OneHotArray{<:Any, L}...; dims::Int) where L
```
DhairyaLGandhi, line 32: Why do we need to overload cat?

src/onehot.jl
```diff
@@ -1,52 +1,57 @@
...
-  return A[:, b.ix]
-end
+_onehotindex(x, i) = (x == i)
 
-struct OneHotMatrix{A<:AbstractVector{OneHotVector}} <: AbstractMatrix{Bool}
-  height::Int
-  data::A
-end
+Base.getindex(x::OneHotVector, i::Integer) = _onehotindex(x.indices, i)
+Base.getindex(x::OneHotVector{T, L}, ::Colon) where {T, L} = OneHotVector{T, L}(x.indices)
```
DhairyaLGandhi, line 23: Can return x here

src/onehot.jl
```diff
@@ -120,11 +122,20 @@ julia> Flux.onecold([0.3, 0.2, 0.5], [:a, :b, :c])
 :c
 ```
 """
-onecold(y::AbstractVector, labels = 1:length(y)) = labels[Base.argmax(y)]
+onecold(y::AbstractVector, labels = 1:length(y)) = labels[argmax(y)]
+function onecold(y::AbstractArray, labels = 1:size(y, 1))
+  indices = _fast_argmax(y)
+  xs = isbits(labels) ? indices : collect(indices) # non-bit type cannot be handled by CUDA
```
DhairyaLGandhi, line 128: When is indices not isbits? We don't index with a CuArray, and if we do we need to remove that happening

DhairyaLGandhi (commented):
I find generalising the type to be nicer in general, but I've comments around defining some of methods and the type seems to hold more information than necessary. That would induce dynamic dispatch which would be nice to avoid

src/onehot.jl
```diff
@@ -1,52 +1,57 @@
...
+  end
+end
 
-# remove workaround when https://github.com/JuliaGPU/CuArrays.jl/issues/676 is fixed
-A::AbstractMatrix * B::OneHotMatrix = A[:, cpu(map(x->x.ix, B.data))]
+Base.hcat(xs::OneHotArray...) = cat(xs...; dims = 2)
+Base.vcat(xs::OneHotArray...) = cat(xs...; dims = 1)
 
-Base.hcat(x::OneHotVector, xs::OneHotVector...) = OneHotMatrix(length(x), [x, xs...])
+Base.reshape(x::OneHotArray{<:Any, L}, dims...) where L =
```
CarloLucibello, line 43: this will also intercept  tuples dims: `reshape(x, (2,3,4))`

src/onehot.jl
```diff
@@ -120,11 +122,20 @@ julia> Flux.onecold([0.3, 0.2, 0.5], [:a, :b, :c])
...
+  indices = _fast_argmax(y)
+  xs = isbits(labels) ? indices : collect(indices) # non-bit type cannot be handled by CUDA
+
+  return map(xi -> labels[xi[1]], xs)
+end
 
-onecold(y::AbstractMatrix, labels...) =
-  dropdims(mapslices(y -> onecold(y, labels...), y, dims=1), dims=1)
+_fast_argmax(x::AbstractArray) = dropdims(argmax(x; dims = 1); dims = 1)
+_fast_argmax(x::OneHotArray) = convert(AbstractArray, x.indices)
```
DhairyaLGandhi, line 134: converting to an abstract type seems anti pattern. you could collect it but I want to know the need for this first to say how it should be handled. It seems to me that one could argmax on the underlying array directly, but isn't that just the indices? In that case, it could be eltype `CartesianIndex`.

src/onehot.jl
darsnack: `T` is the underlying integer type used to encode the one-hot dimension. `L` is the size (or "height") of the one-hot dimension. `N` is the number of dimensions of the integer array used to store all the indices. `var"N+1"` is the number of dimensions in the `Bool` array that this `OneHotArray` mimics, and `I` is the array type used to store the indices.

src/onehot.jl
darsnack: So that you can `cat` two one-hot arrays along `dim > 1` and still get a `OneHotArray` back. It's the generalized version of `hcat`ing two `OneHotMatrix`s should return a larger `OneHotMatrix`.

src/onehot.jl
darsnack: I tried to match the old Flux code which returned a new copy. I can change it if you like.

src/onehot.jl
```diff
@@ -1,52 +1,57 @@
...
-Base.getindex(xs::OneHotMatrix, ::Colon, ::Colon) = OneHotMatrix(xs.height, copy(xs.data))
+_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:OneHotIndex}) where N = Array{Bool, N}
+_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}) where N = CuArray{Bool, N}
 
-Base.getindex(xs::OneHotMatrix, i::Integer, ::Colon) = map(x -> x[i], xs.data)
+function Base.cat(xs::OneHotArray{<:Any, L}...; dims::Int) where L
+  if isone(dims)
+    return cat(map(x -> convert(_onehot_bool_type(x), x), xs)...; dims = 1)
+  else
+    return OneHotArray(L, cat(_indices.(xs)...; dims = dims = dims - 1))
```
CarloLucibello, line 36:
```suggestion
    return OneHotArray(L, cat(_indices.(xs)...; dims = dims - 1))
```

src/onehot.jl
darsnack: My original implementation used `argmax` directly in `onecold`, but I wasn't getting the same performance as #1447 because of the unnecessary conversion to `CartesianIndex`. But even using the indices directly wasn't fast enough. Then I realized that #1447 was doing this conversion to an `AbstractArray`. I tried it and it recovered all the performance lost. Seems like something to ping the CUDA folks about.

src/onehot.jl
```diff
@@ -1,52 +1,57 @@
...
+Base.getindex(x::OneHotArray, i::Integer, I...) = _onehotindex.(x.indices[I...], i)
+Base.getindex(x::OneHotArray{<:Any, L}, ::Colon, I...) where L = OneHotArray(L, x.indices[I...])
+Base.getindex(x::OneHotArray, I::CartesianIndex{N}) where N = x[I[1], Tuple(I)[2:N]...]
 
-Base.getindex(xs::OneHotMatrix, i::Union{Integer, AbstractVector}, j::Integer) = xs.data[j][i]
-Base.getindex(xs::OneHotMatrix, ::Colon, i::Integer) = xs.data[i]
-Base.getindex(xs::OneHotMatrix, ::Colon, i::AbstractArray) = OneHotMatrix(xs.height, xs.data[i])
-Base.getindex(xs::OneHotMatrix, ::Colon, ::Colon) = OneHotMatrix(xs.height, copy(xs.data))
+_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:OneHotIndex}) where N = Array{Bool, N}
+_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}) where N = CuArray{Bool, N}
```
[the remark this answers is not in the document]
darsnack: The underlying array is an integer array. This is just an internal convenience function I use when I want to convert the `OneHotArray` to an `Bool` array. I use this to decide whether to convert to a `Array{Bool}` or `CuArray{Bool}` depending on the underlying storage location.

src/onehot.jl
darsnack: Is that not intended?

src/onehot.jl
darsnack: This is checking when `labels` `isbits` (e.g. if they are strings or symbols). But it might be unnecessary because of `_fast_argmax` being used. I'll double-check this.

src/onehot.jl
chengchingwen: In #1447 which need convert to `AbstractArray{Int32}` is just to remain the array type of the input and converting the primitive type (which can actually be replace with `reinterpret` if no further modification/mutation needed). It's utilizing a constructor of `AbstractArray` which is equivalent to `copyto_axcheck!(similar(A,T), A)` (see [base/array.jl](https://github.com/JuliaLang/julia/blob/master/base/array.jl#L541)). Here you can probably just return `x.indices`

src/onehot.jl
darsnack: Yes this throws an error for non-`isbits` when `labels` is a vector of strings and `y.indices` is a `CuArray`.

src/onehot.jl
darsnack: Yes, that's what I thought, and since I don't need to `reinterpret` the underlying type, I just did `AbstractArray` instead of `AbstractArray{Int32}`. But if I remove this conversion, then the performance will be almost as bad as the old Flux. What that tells me is that most of the performance for the #189 test case was coming from this weird `AbstractArray` conversion hack and not the changes to the one-hot representation.

darsnack: Never mind, I don't know what went wrong in my testing, but `x.indices` works just fine. I must have done something else wrong to make me think otherwise.

src/onehot.jl
CarloLucibello: The code below doesn't seem to handle the tuple case properly, it would recourse infinitely

darsnack: Ah see what you mean. I'll fix this.

darsnack: In the latest commit, I switched to reshaping the `Bool` array in the else case, since the one-hot property is broken anyways. I think that fixes the recursion.

CarloLucibello: recursion is avoided but the tuple case  `reshape(x, (2,3,4))` is not handled correctly, `dims == ((2,3,4), )` inside the function call. Also the case `reshape(x, :, 3, 4)` is not handled correctly. 

I think Base hase some machinery [here](
https://github.com/JuliaLang/julia/blob/0f0bc294f7b18d685c4998ada9c2c33841a74f2d/base/reshapedarray.jl#L111)

that ultimately lead to 
```
function reshape(x, dims::Base.Dims)
...
```
so we can define just that. 
Also, maybe better to error out if `first(dims) != L`?

darsnack: Okay will do.

src/onehot.jl
darsnack: @DhairyaLGandhi any decision on this?

DhairyaLGandhi: I would imagine returning x is fine here

src/onehot.jl
DhairyaLGandhi: Is that tested? Let's add it to tests if not already there.

src/onehot.jl
```diff
@@ -1,52 +1,58 @@
...
+                       throw(ArgumentError("Cannot reshape OneHotArray if first(dims) != size(x, 1)"))
 
-batch(xs::AbstractArray{<:OneHotVector}) = OneHotMatrix(length(first(xs)), xs)
+batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(L, _indices.(xs))
 
-import Adapt: adapt, adapt_structure
+Adapt.adapt_structure(T, x::OneHotArray{<:Any, L}) where L = OneHotArray(L, adapt(T, x.indices))
 
-adapt_structure(T, xs::OneHotMatrix) = OneHotMatrix(xs.height, adapt(T, xs.data))
+Base.BroadcastStyle(::Type{<:OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}}) where N = CuArrayStyle{N}()
```
DhairyaLGandhi, line 51: Kind of a bad signature. Prolly have the more relevant information before anyway

src/onehot.jl
```diff
@@ -1,52 +1,58 @@
-import Base: *
+import Adapt
+import .CUDA
 
-struct OneHotVector <: AbstractVector{Bool}
-  ix::UInt32
-  of::UInt32
+const OneHotIndex{T, N} = Union{T, AbstractArray{T, N}}
```
DhairyaLGandhi, line 4: Can we rename this? It's not intuitive what it means looking at lnehot index, I expect it to be an index.

`HotIndexOrIndices` isn't better. Is it needed or can we get rid of it?

src/onehot.jl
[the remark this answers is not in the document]
DhairyaLGandhi: But why do we need to have this? The implementation for CuArray should fall straight out of assuming regular array

src/onehot.jl
darsnack: The underlying array type is only in the parameters for performance. This is a rare case where we specify it. I put it last because it is the least used.

src/onehot.jl
DhairyaLGandhi: Seems like you're trying to return a value when catting with some arrays on the cpu and some on the GPU?

darsnack: I'm not sure what you mean. The goal is to return a `OneHotArray` when catting on the first dimension and a `Bool` array otherwise. The CPU/GPU stuff is just to make sure we return a `Array{Bool}` vs `CuArray{Bool}`.

DhairyaLGandhi: If the arrays are on the same device, you wouldn't need to do anything here which is what should happen.

Also, why return different types here, seems like you should be able to construct a new onehotarray with the new dimensions and corrected indices

darsnack: If you concatenate along the first dimension, then it is no longer a `OneHotArray`, it is just a `Bool` array. The conversion isn't trying to put all the arrays on the same device. It is there because just calling `cat` again will result in infinite recursion. Instead, we want to treat the arrays as `Bool` arrays and call the `cat` dispatch for `Bool` arrays.

darsnack:
The last remaining issue that needs to be addressed is that Base's `reshape` logic returns a lazy `ReshapedArray` for cases like `reshape(x, 10, :)`. This means that if you collect the lazy iterator, you'll currently get an `Array{Bool}` when we want to return a `OneHotArray` (see the failing test cases).

CarloLucibello:
What Base does with reshape is just crazy. 
I think you have to intercept the following line 
https://github.com/JuliaLang/julia/blob/788b2c77c10c2160f4794a4d4b6b81a95a90940c/base/reshapedarray.jl#L118

So maybe overload
```julia
Base._reshape(x::OneHotArray, dims:::Dims)
```
?

DhairyaLGandhi:
Please don't touch internal Julia functions

darsnack:
Probably we just want to extend
```julia
reshape(parent::AbstractArray, dims::Tuple{Vararg{Union{Int,Colon}}})
```

darsnack:
Never mind, that leads to a method ambiguity that I think can only be resolved in Base. We can either overload `_reshape` like @CarloLucibello suggested, or we ship with these tests broken until this is addressed in Base.

CarloLucibello:
> Never mind, that leads to a method ambiguity that I think can only be resolved in Base. We can either overload _reshape like @CarloLucibello suggested, or we ship with these tests broken until this is addressed in Base.

I wouldn't hold on adding a very basic functionality just because the situation is a bit weird in Base. I say we add `_reshape`, if things will change in future julia versions we will change accordingly, the important thing is to test cases like `reshape(x, 10, :)`.

Most importantly though, we have to decide whether we go with this or #1447

DhairyaLGandhi:
Agreed on testing for the right cases but not on overloading internal functions. We can catch the failing case as a separate method instead

darsnack:
> I wouldn't hold on adding a very basic functionality just because the situation is a bit weird in Base.

I agree.

> We can catch the failing case as a separate method instead

That's what I tried. Due to the situation in Base, we are forced to catch the failing case with `Base._reshape`. There's no way to catch all the cases with `Base.reshape` alone.

> Most importantly though, we have to decide whether we go with this or #1447

If this is for the v0.12 milestone, then I suggest we go for this. Based on [this comment](https://github.com/FluxML/Flux.jl/pull/1447#issuecomment-753491156), I think there is a slim set of models where the performance of this PR and #1447 can't be matched. We can always adopt #1447 if we need to in the future. Going back from a new primitive type will be harder.

#1447 and this PR are composable changes. I think all that's required to move to #1447 in the future is to replace the `indices` fields with the primitive `OneHot`.

CarloLucibello (approved):

CarloLucibello:
@DhairyaLGandhi I asked a few times already, could you lift github restrictions (I get "The base branch restricts merging to authorized users. Learn more about protected branches.") and clarify if we have to use bors r+ or not now that we have buildkite?

@darsnack we have a failing onecold test on gpu.

To me the plan sounds good, the simplicity of this approach is very compelling, we can merge this and revisit later if performance issues are lamented on those corner cases. This unless @chengchingwen has strong objections against

test/cuda/cuda.jl
```diff
@@ -40,8 +40,10 @@ end
 
 @testset "onecold gpu" begin
   y = Flux.onehotbatch(ones(3), 1:10) |> gpu;
+  l = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j']
   @test Flux.onecold(y) isa CuArray
   @test y[3,:] isa CuArray
+  @test Flux.onecold(y, l) == ['a', 'a', 'a']
```
CarloLucibello, line 46: this behavior is a bit weird right? maybe we should just error out on non-isbits labels and CuArrays?

CarloLucibello:
> @DhairyaLGandhi I asked a few times already, could you lift github restrictions (I get "The base branch restricts merging to authorized users. Learn more about protected branches.") and clarify if we have to use bors r+ or not now that we have buildkite?

@DhairyaLGandhi bump

test/cuda/cuda.jl
darsnack: Yeah a little odd since `y` is a `CuArray`, but `l` is an `Array` so not too weird. I've encountered this use case before, and my preference would be to get back a result than an error. As a user, I'd expect the first thing people try is to do `l |> gpu` or `onecold(y, l) |> gpu` if they want a `CuArray`. Both of which should throw an error, so the issue should be obvious.

darsnack: The reason this fix is in there is to address stuff like #864

CarloLucibello (approved):

src/onehot.jl
DhairyaLGandhi: I would perhaps consider restricting that rather than letting users get into gotchas. For cases where that is indeed desired, we can have users explicitly opt in, collect the arrays and do as they please

test/onehot.jl
```diff
@@ -27,9 +27,83 @@ end
 
 @testset "abstractmatrix onehotvector multiplication" begin
   A = [1 3 5; 2 4 6; 3 6 9]
-  b1 = Flux.OneHotVector(1,3)
-  b2 = Flux.OneHotVector(3,5)
+  b1 = Flux.OneHotVector{eltype(A), 3}(1)
+  b2 = Flux.OneHotVector{eltype(A), 5}(3)
```
DhairyaLGandhi, line 31: Ideally we could add constructors to be backwards compatible

darsnack: I did add a constructor `OneHotVector(L, index)` but I guess the argument order isn't backwards compatible. Let me update the order for compatibility.

DhairyaLGandhi:
Thanks, I've added a thought around saving the vcat to be an error with a message pointing users to collect the onehotarray if it's no longer one hot

I'm inclined to go with this over #1447 since they address the same concerns but in a neater fashion.

We'll continue with bors for the time being.

darsnack:
Okay the constructors should be backwards compatible and `vcat` will throw an error.

DhairyaLGandhi (approved):

DhairyaLGandhi (approved):

src/onehot.jl
```diff
@@ -1,52 +1,61 @@
...
-# remove workaround when https://github.com/JuliaGPU/CuArrays.jl/issues/676 is fixed
-A::AbstractMatrix * B::OneHotMatrix = A[:, cpu(map(x->x.ix, B.data))]
+Base.hcat(xs::OneHotArray...) = cat(xs...; dims = 2)
+Base.vcat(xs::OneHotArray...) = cat(xs...; dims = 1)
 
-Base.hcat(x::OneHotVector, xs::OneHotVector...) = OneHotMatrix(length(x), [x, xs...])
+Base.reshape(x::OneHotArray{<:Any, L}, dims::Dims) where L =
+  (first(dims) == L) ? OneHotArray(reshape(x.indices, dims[2:end]...), L) :
+                       throw(ArgumentError("Cannot reshape OneHotArray if first(dims) != size(x, 1)"))
+Base._reshape(x::OneHotArray, dims::Tuple{Vararg{Int}}) = reshape(x, dims)
```
simeonschaub, line 48: What was the reason for adding this? It seems overly restrictive to require that `first(dims) == L`, in fact, this broke some of my code.

simeonschaub: The fallback worked fine for me before

darsnack: The initial implementation converted to a `Bool` array for the else case. @CarloLucibello seems like we should add that back in?

DhairyaLGandhi: I think Simeon is referring to before we merged this? Did we convert to a bool array then? I think it would be difficult to guarantee the return type of the function then. I agree the check seems pretty restrictive

darsnack: Not sure what fallback @simeonschaub is referring to, but the original implementation of this PR did not throw an error. It converted to a `Bool` array.

simeonschaub: Sorry, should have clarified: I meant falling back to the default defintion in Base for `AbstractArray`, which produces a `ReshapedArray`. I think if we overload `reshape` here, we shouldn't make it error in cases where the fallback would work, since that makes it hard to use `reshape` in generic code.

CarloLucibello: I thought that reshaping the first dimension was something never done in practice, but since we broke @simeonschaub's code maybe is not so rare. It can be handled by `reshape(collect(oh), ...)`, but i would be fine to relax the dims check if people feel that need, although, as @DhairyaLGandhi  said, this would make reshape type unstable

darsnack: I'll submit a quick fix PR

simeonschaub: Just to explain, in my use case I was adding a singleton dimension in front of the rest for the purpose of broadcasting, i.e. something like:
```julia
reshape(Flux.onehotbatch(Flux.onecold(ŷ, classes), classes), 1, 4, :) .===  _reshape(outputs_onehot, 4, 1, :);
```

cossio:
It would be nice to add docs for this. https://github.com/FluxML/Flux.jl/issues/1519

cossio:
Was this included in a released Flux version?

cossio:
I think this PR hasn't been released yet. Why?

darsnack:
Not 100% positive cause I’m still hazy on semver but I think the struct/constructor changes means it is breaking. So it will have to wait until v0.12. Even though the highest level APIs like `onehotbatch` are the same.

ToucheSir:
Copying @darsnack's comment's from the parallel Slack discussion:
> Not 100% positive cause I’m still hazy on semver but I think the struct/constructor changes means it is breaking. So it will have to wait until v0.12. Even though the highest level APIs like onehotbatch are the same: https://github.com/FluxML/Flux.jl/compare/v0.11.6...master

PR #1449, opened 2021-01-04
Update GPU CI.

maleadt:
Dummy changes to test status pushes from BuildKite. Now that it works, remove GitLab CI.


CarloLucibello (approved):

maleadt:
Bors isn't required anymore now that BuildKite can do thirt-party PRs. Also note that it was only listening for `gitlab` statuses anyway: https://github.com/FluxML/Flux.jl/blob/ef1aa4927a06a70a2efa679a29ce3615c3e846eb/bors.toml. Either it should be updated to check the BuildKite status, all statusses, or removed altogether.

DhairyaLGandhi:
Thanks so much!

CarloLucibello:
so @DhairyaLGandhi you have to lift the github's restriction, merging is blocked here

PR #1451, opened 2021-01-05
release 0.11.4

CarloLucibello:


DhairyaLGandhi:
Let's wait for a breaking release? Zygote 0.6 has some major shuffling around and touches language semantics on getfield/ property

CarloLucibello:
I am in no hurry, but probably @willtebbutt has a bit more urgency

willtebbutt:
@DhairyaLGandhi I would very much like a release if that's possible to that I can upgrade some of my packages to Zygote 0.6. Is there anything in particular that you would expect not to be picked up by Flux's tests?

DhairyaLGandhi:
The cases that the tests might miss is what concerns me. Nnlib had some rework as well which would impact the cuda side of things too

willtebbutt:
> The cases that the tests might miss is what concerns me. Nnlib had some rework as well which would impact the cuda side of things too

Is this not a little bit concerning for the 0.12 release as well? I mean, it sounds like you're concerned that changes in NNlib and Zygote will propagate through into "bugs" in Flux. Is that the case?

I guess I don't see how not upgrading until 0.12 is helpful here. Presumably it will couple "bugs" due to changes in NNlib / Zygote with breaking changes in Flux, potentially making it harder to figure out what's going wrong.

DhairyaLGandhi:
The reasons for that are because of the reshuffle. It's something I think about on the daily anyway. I'm not sure what your concerns here are exactly, but hopefully we can verify with ci and testing with a couple models from the zoo for sureity. We have to release zygote 0.6 anyway and move from there

willtebbutt:
Sorry, I'm now quite confused 😂 Are you saying that we should go ahead with this PR?

> The reasons for that are because of the reshuffle.

What reshuffle are we talking about?

Issue #1450, opened 2021-01-05
Implement sparse initialization


atiyo:
In the PyTorch parity issue (https://github.com/FluxML/Flux.jl/issues/1431#issue-771272811) it's pointed out that PyTorch supports sparse initialization where Flux doesn't. Raising this issue to avoid taking that thread too far off-topic.

I'm happy to implement sparse initialization if it would be helpful.

The PyTorch implementation currently works only for 2d tensors (https://pytorch.org/docs/stable/_modules/torch/nn/init.html#sparse_). Mimicking this implementation seems straightforward.

Would Flux want to generalise beyond 2d? If so, what is the correct way to do so?

The PyTorch implementation sets a fixed proportion of each column to zero. On multi-dimensional generalisation possibility: setting a fixed proportion of the the last dimension of an `AbstractArray` to zero. But whether this is a meaningful thing to do might depend on the context?

DhairyaLGandhi:
Would appreciate a pr!

The most straightforward thing to do would be to start with a simple implemention that mimics pytorch behavior and subsequently generalise it

atiyo:
Addressed by PR above.

Issue #1431, opened 2020-12-19
PyTorch feature parity


CarloLucibello:
A list of PyTorch 1.7 features. 
Items are checked if we have something more or less equivalent in Flux or in the julia ecosystem and supported by Flux. 
This list comes from a first rough scan of pytorch's documentation, please fill free to add anything I missed in the comments. 

**Conv Layers**
- [x] `Conv1d`, `Conv2d`, `Conv3d`.  
- [x] `ConvTranspose1d`, `ConvTranspose2d`, `ConvTranspose3d`.
- [ ]  `Fold`, `Unfold`.

**Pooling Layers**
- [x]  `MaxPool1d`, `MaxPool2d`, `MaxPool3d` 
- [ ]  `MaxUnPool1d`, `MaxUnPool2d`, `MaxUnPool3d` 
- [x]  `AvgPool1d`, `AvgPool2d`, `AvgPool3d`
- [ ] `FractionalMaxPool2d`
- [ ] `LPPool1d`, `LPPool2d` 
- [ ] `AdaptiveAvgPool1d`, `AdaptiveAvgPool1d`, `AdaptiveAvgPool3d`
- [ ] `AdaptiveMaxPool1d`, `AdaptiveMaxPool1d`, `AdaptiveMaxPool3d`

**Padding Layers**
- [ ] ....

**Activations**
- [x] ... . NNlib has an extensive collection of activation, plus we have any julia function.

**Normalization Layers**
- [x] `BatchNorm1d`, `BatchNorm2d`, `BatchNorm3d` 
- [x] `LayerNorm`
- [x] `GroupNorm`
- [x] `InstanceNorm1d`,`InstanceNorm2d`,`InstanceNorm3d`
- [ ] `SyncBatchNorm` 
- [ ] `LocalResponseNorm`

**Recurrent Layers**
- [x] `RNN`
- [x] `GRU`
- [x]  `LSTM`

**Transformer Layers**
- [ ] `Transformer`
- [ ] `TransformerEncoder`
- [ ] `TransformerDecoder`
 
**Linear Layers**
- [x] `Identity`
- [x] `Linear`
- [x] `Bilinear`

**Dropoout Layers**
- [x] `Dropout`
- [ ] `Dropout2d`, `Dropout3d`
- [x] `AlphaDropout`

**Sparse Layers**
- [ ] **Emberdding**
- [ ] **EmbeddingBag**

**Distance Functions**
[...]

PR #1454, opened 2021-01-06
Add sparse initialization

atiyo:
Add sparse initialization, documentation and tests. Trim whitespace in editted files.

This PR is intended to address one of the outstanding points in bringing Flux to parity with PyTorch's features so it partially addresses https://github.com/FluxML/Flux.jl/issues/1431 and fully addresses https://github.com/FluxML/Flux.jl/issues/1450.

The implementation follows the method given in [PyTorch implementation](https://pytorch.org/docs/stable/_modules/torch/nn/init.html#sparse_): a normally-distributed array is created, then a fixed proportion of randomly chosen row-indices is zeroed out for every column. Like the PyTorch version, it is restricted to 2-d Arrays.

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -187,9 +187,9 @@ This method is described in [1].
 ```jldoctest; setup = :(using Random; Random.seed!(0))
 julia> Flux.sparse_init(3, 2, sparsity=0.1)
 3×2 Array{Float32,2}:
- 0.00679107  0.0
- 0.00828413  0.00586617
- 0.0         0.00297336
+  0.00828413  0.0
+ -0.00353007  0.00297336
+  0.0         0.00586617
 ```
 
 # See also
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -186,7 +186,7 @@ This method is described in [1].
 # Examples
 ```jldoctest; setup = :(using Random; Random.seed!(0))
 julia> Flux.sparse(3, 2, sparsity=0.1)
-3×2 Array{Float64,2}:
+3×2 Array{Float32,2}:
  0.00679107  0.0
  0.00828413  0.00586617
  0.0         0.00297336
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -211,7 +211,7 @@ function sparse_init(rng::AbstractRNG, dims...; sparsity, std = 0.01)
   prop_zero = min(1.0, sparsity)
   num_zeros = ceil(Integer, prop_zero * rows)
   sparse_array = randn(rng, Float32, dims...) .* Float32(std)
-  sparse_array[1:num_zeros, :] .= zero(Float32)
+  sparse_array[1:num_zeros, :] .= 0f0
   return mapslices(shuffle, sparse_array, dims=1)
 end
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -56,7 +56,7 @@ julia> Flux.glorot_uniform(2, 3)
 * glorot initialization using normal distribution: [`glorot_normal`](@ref Flux.glorot_normal)
 * kaiming initialization using normal distribution: [`kaiming_normal`](@ref Flux.kaiming_normal)
 * kaiming initialization using uniform distribution: [`kaiming_uniform`](@ref Flux.kaiming_uniform)
-* sparse initialization: [`sparse`](@ref Flux.sparse)
+* sparse initialization: [`sparse_init`](@ref Flux.sparse_init)
 * calculation of `fan_in` and `fan_out`: [`nfan`](@ref Flux.nfan)
 
 # References
@@ -89,7 +89,7 @@ julia> Flux.glorot_normal(3, 2)
 * glorot initialization using uniform distribution: [`glorot_uniform`](@ref Flux.glorot_uniform)
 * kaiming initialization using normal distribution: [`kaiming_normal`](@ref Flux.kaiming_normal)
 * kaiming initialization using uniform distribution: [`kaiming_uniform`](@ref Flux.kaiming_uniform)
-* sparse initialization: [`sparse`](@ref Flux.sparse)
+* sparse initialization: [`sparse_init`](@ref Flux.sparse_init)
 * calculation of `fan_in` and `fan_out`: [`nfan`](@ref Flux.nfan)
 
 # References
@@ -122,7 +122,7 @@ julia> Flux.kaiming_uniform(3, 2)
 * kaiming initialization using normal distribution: [`kaiming_normal`](@ref Flux.kaiming_normal)
 * glorot initialization using normal distribution: [`glorot_normal`](@ref Flux.glorot_normal)
 * glorot initialization using uniform distribution: [`glorot_uniform`](@ref Flux.glorot_uniform)
-* sparse initialization: [`sparse`](@ref Flux.sparse)
+* sparse initialization: [`sparse_init`](@ref Flux.sparse_init)
 * calculation of `fan_in` and `fan_out`: [`nfan`](@ref Flux.nfan)
 
 # References
@@ -159,7 +159,7 @@ julia> Flux.kaiming_normal(3, 2)
 * kaiming initialization using uniform distribution: [`kaiming_uniform`](@ref Flux.kaiming_uniform)
 * glorot initialization using normal distribution: [`glorot_normal`](@ref Flux.glorot_normal)
 * glorot initialization using uniform distribution: [`glorot_uniform`](@ref Flux.glorot_uniform)
-* sparse initialization: [`sparse`](@ref Flux.sparse)
+* sparse initialization: [`sparse_init`](@ref Flux.sparse_init)
 * calculation of `fan_in` and `fan_out`: [`nfan`](@ref Flux.nfan)
 
 # References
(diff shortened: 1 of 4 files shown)


src/utils.jl
```diff
@@ -170,14 +174,63 @@ end
 kaiming_normal(dims...; kwargs...) = kaiming_normal(Random.GLOBAL_RNG, dims...; kwargs...)
 kaiming_normal(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> kaiming_normal(rng, dims...; kwargs...)
 
+"""
+    sparse([rng=GLOBAL_RNG], dims...; sparsity, std = 0.01)
```
atiyo, line 178: Is calling this function `sparse` ok? It seems like a very general name to be taken. This keeps consistency with the other initialization names, though.

Also, I couldn't find a specific naming convention for standard deviations as keyword arguments. Using \sigma could be an alternative, but is often used in Flux to denote activations, so I preferred to avoid it.

src/utils.jl
```diff
@@ -170,14 +174,63 @@ end
...
+  if length(dims) != 2
+    throw(ArgumentError("Only 2-dimensional outputs are supported for sparse initialization."))
+  end
+  rows, cols = dims
+  prop_zero = min(1.0, sparsity)
+  num_zeros = ceil(Integer, prop_zero * rows)
+  sparse_array = randn(rng, Float32, dims...) .* Float32(std)
+  row_indices = collect(1:rows)
+  for col in 1:cols
+    randperm!(row_indices)
```
atiyo, line 216: Using `randperm!` and then index some elements as a shortcut to sampling without replacement.

An alternative would be to use `sample` from `StatsBase`, which is already a dependency, but not used outside of the tests. I preferred to use `randperm!` from `Random` in order to avoid adding this extra import.

src/utils.jl
DhairyaLGandhi: This would cause banking conflicts with Sparse arrays, so I would pick a different name.

src/utils.jl
```diff
@@ -170,14 +174,63 @@ end
...
+
+[1] Martens, J, "Deep learning via Hessian-free optimization" _Proceedings of the 27th International Conference on International Conference on Machine Learning_. 2010.
+"""
+function sparse(rng::AbstractRNG, dims...; sparsity, std = 0.01)
+  if length(dims) != 2
+    throw(ArgumentError("Only 2-dimensional outputs are supported for sparse initialization."))
+  end
+  rows, cols = dims
+  prop_zero = min(1.0, sparsity)
+  num_zeros = ceil(Integer, prop_zero * rows)
```
DhairyaLGandhi, line 212: Use `\div` here

src/utils.jl
DhairyaLGandhi: Seems like we should be able to generate a mask and multiply too ratherv then iterating over the columns.

src/utils.jl
```diff
@@ -170,14 +174,63 @@ end
...
+  sparse_array = randn(rng, Float32, dims...) .* Float32(std)
+  row_indices = collect(1:rows)
+  for col in 1:cols
+    randperm!(row_indices)
+    zero_indices = row_indices[1:num_zeros]
+    sparse_array[zero_indices, col] .= 0.0
```
DhairyaLGandhi, line 218: Type instability possibility. If we can take the type as input, we can use `zero(T)`

DhairyaLGandhi:
Thanks for looking into this! I've left a couple of thoughts in the implementation. We would need to use a different name though since `sparse` is already a function in a stdlib

src/utils.jl
CarloLucibello: what about `sparse_init`?

src/utils.jl
CarloLucibello, line 218:
```suggestion
    sparse_array[zero_indices, col] .= 0
```
There is no risk of instability in any case, we are assigning in-place

src/utils.jl
atiyo: My first thought was also `sparse_init`, so I've run with that.

src/utils.jl
atiyo: Agreed that iterating over columns was a bit unwieldy. Manually setting zeros then shuffling rows with `mapslice(shuffle,...)` was the cleanest I could come up with.

src/utils.jl
atiyo: I assume you mean something like `div(rows, 1/prop_zero)`? This returns a float since `prop_zero` is a float, so would require further casting to an integer. I thought above was a bit easier to follow, but am happy to go with what you think is best.

src/utils.jl
atiyo: In the corresponding refactor, I use `zero(Float32)` when assigning zeros. Happy to take the type as an input and pipe it through. But this isn't the case with other initialization schemes, and I was trying to keep things as consistent as possible.

atiyo: Appreciate the suggestion. In the refactor I assign to `zeros(Float32)`, which I appreciate might be needlessly type specific. Happy to change it to an assignment to `0` based on consensus.

src/utils.jl
```diff
@@ -170,14 +174,58 @@ end
...
+"""
+function sparse_init(rng::AbstractRNG, dims...; sparsity, std = 0.01)
+  if length(dims) != 2
+    throw(ArgumentError("Only 2-dimensional outputs are supported for sparse initialization."))
+  end
+  rows, cols = dims
+  prop_zero = min(1.0, sparsity)
+  num_zeros = ceil(Integer, prop_zero * rows)
+  sparse_array = randn(rng, Float32, dims...) .* Float32(std)
+  sparse_array[1:num_zeros, :] .= zero(Float32)
```
ToucheSir, line 214: You can use Float32 constants instead of casting or calling `zero`.
```suggestion
  sparse_array[1:num_zeros, :] .= 0f0
```

CarloLucibello:
Current implementation does `  sparse_array[1:num_zeros, :] .= 0`, but it should randomize the zero positions in each row.
The pytorch code is:
```python
def sparse_(tensor, sparsity, std=0.01):
    r"""Fills the 2D input `Tensor` as a sparse matrix, where the
    non-zero elements will be drawn from the normal distribution
    :math:`\mathcal{N}(0, 0.01)`, as described in `Deep learning via
    Hessian-free optimization` - Martens, J. (2010).

    Args:
        tensor: an n-dimensional `torch.Tensor`
        sparsity: The fraction of elements in each column to be set to zero
        std: the standard deviation of the normal distribution used to generate
            the non-zero values

    Examples:
        >>> w = torch.empty(3, 5)
        >>> nn.init.sparse_(w, sparsity=0.1)
    """
    if tensor.ndimension() != 2:
        raise ValueError("Only tensors with 2 dimensions are supported")

    rows, cols = tensor.shape
    num_zeros = int(math.ceil(sparsity * rows))

    with torch.no_grad():
        tensor.normal_(0, std)
        for col_idx in range(cols):
            row_indices = torch.randperm(rows)
            zero_indices = row_indices[:num_zeros]
            tensor[zero_indices, col_idx] = 0
    return tensor
```
We should follow them, swapping cols with rows, so `sparse_array[:, zero_indices] .= 0`

CarloLucibello:
Sorry, now I see that you randomly permute with `mapslices(shuffle, x)`. So only my comment of swapping rows and cols compared to python's implementation applies

atiyo:
> Sorry, now I see that you randomly permute with `mapslices(shuffle, x)`. So only my comment on swapping rows and cols compared to python's implementation applies

I'm not clear on why we need to swap rows and cols compared to PyTorch. I understand the batch dimension is different, but as far as I could tell Flux uses similar shapes for weights.

E.g
```python
In [1]: from torch import nn
In [2]: nn.Linear(1,2).weight.shape
Out[2]: torch.Size([2, 1])
```
```julia
julia> using Flux
julia> size(Dense(1,2).W)
(2, 1)
```

CarloLucibello:
@atiyo you're right, I always thought that pytorch applies the transform `x * W`,  but now I see that it does `x * W^T` instead. So this PR looks entirely fine to me. If we want to do some more name bikeshedding, `init_sparse` is an alternative proposal maybe more discoverable by tab-completion when looking for initialization methods.

Another consideration is that maybe we can move initialization functions to a submodule, but this doesn't have to be necessarily discussed here.

src/utils.jl
DhairyaLGandhi: `\div{tab}` should return an inyt

DhairyaLGandhi:
Let's not move it to a submodule, doesn't seem worthwhile enough as a standalone to me

`init_sparse` would be better for discoverability, but somewhat inconsistent. I guess it's alright for now

src/utils.jl
atiyo: I might be missing something. I'm finding `÷` to behave the same way as `div`, i.e. it's returning a float for float values of `prop_zero`.

```julia
julia> prop_zero = 0.11; rows = 50;
julia> ÷(rows, 1/prop_zero, RoundUp)
6.0
```

Using `÷` as an infix operator still returns a float, but also doesn't allow to specify a `RoundingMode`. We need to round up to maintain consistency with PyTorch.

CarloLucibello (approved):

PR #1456, opened 2021-01-07
Update Zygote version to 0.6

CarloLucibello:


CarloLucibello:
@DhairyaLGandhi we have an error possibly due to getproperty

CarloLucibello:
@simeonschaub do you have any clue for the failures here? Could they be related to getproperty changes in https://github.com/FluxML/Zygote.jl/848? 
I don't see where those try-catch it is complaining about come from

simeonschaub:
I think deleting the `@warn` [here](https://github.com/FluxML/Flux.jl/blob/ebd37d69afd559704b3c9ededf2a3bdead710d4c/src/layers/recurrent.jl#L107) or wrapping it in `Zygote.@ignore` should fix that error.

CarloLucibello:
I added the ignore blocks and incorporated the changes from #1458

darsnack (approved):

darsnack:
Assuming you wanted me to double check the port of #1458. Will close that when this is merged.

simeonschaub (approved):
LGTM!

src/outputsize.jl
```diff
@@ -40,6 +41,8 @@ Base.typemax(::Type{Nil}) = nil
 
 Base.promote_rule(x::Type{Nil}, y::Type{<:Number}) = Nil
 
+Random.rand(rng::Random.AbstractRNG, ::Random.SamplerType{Nil}) = nil
```
DhairyaLGandhi, line 44: This is already in a separate patch, we can skip this?

Project.toml
```diff
@@ -27,18 +27,18 @@ Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
 AbstractTrees = "0.3"
-Adapt = "2.0"
+Adapt = "2.0, 3.0"
```
DhairyaLGandhi, line 30: Can we stick with the version that cuda 2.x wants? We don't meaningfully break compat and mismatched versions due to too lax compat bounds can be bad for cuda

DhairyaLGandhi (commented):
Thanks, looks good!

src/outputsize.jl
CarloLucibello: we need this to pass the tests

Project.toml
CarloLucibello: The breaking changes in Adapt were introduced in https://github.com/JuliaGPU/Adapt.jl/pull/36
I think Flux.jl is directly compatible with those changes, so we should update the bound. If some of our dependencies have not updated the bound due to incompatibility or simply lack of time, this is for the package version solver to figure out. This is how the julia ecosystem generally works

DhairyaLGandhi (approved):

PR #1458, opened 2021-01-08
Fix rand(::Type{Nil}) errors

darsnack:
Fixes failing CI in #1452 by defining [the proper Random interface](https://docs.julialang.org/en/v1/stdlib/Random/#Generating-values-from-a-type) for the `Nil` type.


darsnack:
Closing since #1456 merged these changes already.

PR #1459, opened 2021-01-09
Use fallback for reshape/cat OneHotArray

darsnack:
This falls back to reshaping a `Bool` array whenever reshaping the first dimension of a `OneHotArray`.

@DhairyaLGandhi @CarloLucibello @simeonschaub 

### PR Checklist

- [x] Tests are added
- [ ] ~~Entry in NEWS.md~~

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -9,12 +9,29 @@ OneHotArray(indices::T, L::Integer) where {T<:Integer} = OneHotArray{T, L, 0, T}
 OneHotArray(indices::AbstractArray{T, N}, L::Integer) where {T, N} = OneHotArray{T, L, N, typeof(indices)}(indices)
 
 _indices(x::OneHotArray) = x.indices
+_indices(x::Base.ReshapedArray{<:Any, <:Any, <:OneHotArray}) =
+  reshape(parent(x).indices, x.dims[2:end])
 
 const OneHotVector{T, L} = OneHotArray{T, L, 0, 1, T}
 const OneHotMatrix{T, L, I} = OneHotArray{T, L, 1, 2, I}
 
 OneHotVector(idx, L) = OneHotArray(idx, L)
 OneHotMatrix(indices, L) = OneHotArray(indices, L)
+
+# use this type so reshaped arrays hit fast paths
+# e.g. argmax
+const OneHotLike{T, L, N, var"N+1", I} =
+  Union{OneHotArray{T, L, N, var"N+1", I},
+        Base.ReshapedArray{Bool, var"N+1", <:OneHotArray{T, L}}}
+
+# when reshaping a OneHotArray and first(dims) != L
+# convert the parent array to Array{Bool}
+# so that the ReshapedArray does not hit fast paths
+function Base.ReshapedArray(parent::OneHotArray{<:Any, L}, dims::NTuple{N,Int}, mi) where {L, N}
+  parent = (first(dims) != L) ? convert(_onehot_bool_type(parent), parent) : parent
+  
+  Base.ReshapedArray{Bool,N,typeof(parent),typeof(mi)}(parent, dims, mi)
+end
 
 Base.size(x::OneHotArray{<:Any, L}) where L = (Int(L), size(x.indices)...)
 
@@ -24,14 +41,14 @@ Base.getindex(x::OneHotVector, i::Integer) = _onehotindex(x.indices, i)
 Base.getindex(x::OneHotVector{T, L}, ::Colon) where {T, L} = x
 
 Base.getindex(x::OneHotArray, i::Integer, I...) = _onehotindex.(x.indices[I...], i)
-Base.getindex(x::OneHotArray{<:Any, L}, ::Colon, I...) where L = OneHotArray(x.indices[I...], L)
+Base.getindex(x::OneHotLike{<:Any, L}, ::Colon, I...) where L = OneHotArray(_indices(x)[I...], L)
 Base.getindex(x::OneHotArray{<:Any, <:Any, <:Any, N}, ::Vararg{Colon, N}) where N = x
 Base.getindex(x::OneHotArray, I::CartesianIndex{N}) where N = x[I[1], Tuple(I)[2:N]...]
 
-_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:Union{Integer, AbstractArray}}) where N = Array{Bool, N}
-_onehot_bool_type(x::OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}) where N = CuArray{Bool, N}
+_onehot_bool_type(x::OneHotLike{<:Any, <:Any, <:Any, N, <:Union{Integer, AbstractArray}}) where N = Array{Bool, N}
+_onehot_bool_type(x::OneHotLike{<:Any, <:Any, <:Any, N, <:CuArray}) where N = CuArray{Bool, N}
 
-function Base.cat(xs::OneHotArray{<:Any, L}...; dims::Int) where L
+function Base.cat(xs::OneHotLike{<:Any, L}...; dims::Int) where L
   if isone(dims)
     return cat(map(x -> convert(_onehot_bool_type(x), x), xs)...; dims = dims)
   else
@@ -39,22 +56,17 @@ function Base.cat(xs::OneHotArray{<:Any, L}...; dims::Int) where L
   end
 end
 
-Base.hcat(xs::OneHotArray...) = cat(xs...; dims = 2)
-Base.vcat(xs::OneHotArray...) = cat(xs...; dims = 1)
-
-Base.reshape(x::OneHotArray{<:Any, L}, dims::Dims) where L =
-  (first(dims) == L) ? OneHotArray(reshape(x.indices, dims[2:end]...), L) :
-                       reshape(convert(_onehot_bool_type(x), x), dims)
-Base._reshape(x::OneHotArray, dims::Tuple{Vararg{Int}}) = reshape(x, dims)
+Base.hcat(xs::OneHotLike...) = cat(xs...; dims = 2)
+Base.vcat(xs::OneHotLike...) = cat(xs...; dims = 1)
 
 batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(_indices.(xs), L)
 
-Adapt.adapt_structure(T, x::OneHotArray{<:Any, L}) where L = OneHotArray(adapt(T, x.indices), L)
+Adapt.adapt_structure(T, x::OneHotLike{<:Any, L}) where L = OneHotArray(adapt(T, _indices(x)), L)
 
-Base.BroadcastStyle(::Type{<:OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}}) where N = CUDA.CuArrayStyle{N}()
+Base.BroadcastStyle(::Type{<:OneHotLike{<:Any, <:Any, <:Any, N, <:CuArray}}) where N = CUDA.CuArrayStyle{N}()
 
-Base.argmax(x::OneHotArray; dims = Colon()) =
-  (dims == 1) ? reshape(CartesianIndex.(x.indices, CartesianIndices(x.indices)), 1, size(x.indices)...) :
+Base.argmax(x::OneHotLike; dims = Colon()) =
+  (dims == 1) ? reshape(CartesianIndex.(_indices(x), CartesianIndices(_indices(x))), 1, size(_indices(x))...) :
                 argmax(convert(_onehot_bool_type(x), x); dims = dims)
 
 """
@@ -135,11 +147,11 @@ function onecold(y::AbstractArray, labels = 1:size(y, 1))
 end
 
 _fast_argmax(x::AbstractArray) = dropdims(argmax(x; dims = 1); dims = 1)
-_fast_argmax(x::OneHotArray) = x.indices
+_fast_argmax(x::OneHotLike) = _indices(x)
 
 @nograd OneHotArray, onecold, onehot, onehotbatch
 
-function Base.:(*)(A::AbstractMatrix, B::OneHotArray{<:Any, L}) where L
+function Base.:(*)(A::AbstractMatrix, B::OneHotLike{<:Any, L}) where L
   size(A, 2) == L || throw(DimensionMismatch("Matrix column must correspond with OneHot size: $(size(A, 2)) != $L"))
   return A[:, onecold(B)]
 end
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -62,7 +62,7 @@ Base.BroadcastStyle(::Type{<:OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}}) wh
 Base.argmax(x::OneHotLike; dims = Colon()) =
   (_isonehot(x) && dims == 1) ?
     reshape(CartesianIndex.(_indices(x), CartesianIndices(_indices(x))), 1, size(_indices(x))...) :
-    argmax(convert(_onehot_bool_type(x), x); dims = dims)
+    invoke(argmax, Tuple{AbstractArray}, x; dims = dims)
 
 """
     onehot(l, labels[, unk])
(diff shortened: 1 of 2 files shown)


simeonschaub:
Sorry, if this is a dumb question, but I didn't quite understand the reasoning in #1448 why the default behavior for AbstractArrays in Base was bad.

> This means that if you collect the lazy iterator, you'll currently get an Array{Bool} when we want to return a OneHotArray (see the failing test cases).

darsnack:
That actually [wasn't the reason](https://github.com/FluxML/Flux.jl/pull/1448#discussion_r550897754) for throwing the error. I guess the actual reason might have been that when you reshape the first dimension, the array is no longer one-hot, so we erred on the side of alerting the user instead of silently converting (the same was done for `cat`ing along the first dimension).

With respect to the part you quoted, that is referring to the non-error branch where the first dimension is the same and the other dimensions are reshaped. In this case, the array is still a `OneHotArray`, so we'd like to return that type. But, due to https://github.com/JuliaLang/julia/issues/39123, we get a `ReshapedArray` out. Collecting the `ReshapedArray` will index the parent array by element which returns a `Bool` resulting in an `Array{Bool}` out instead of the `OneHotArray` we wanted.

But I do think https://github.com/JuliaLang/julia/pull/38965 might allow us to fix this. I can test it out on this branch and commit it if that's the desired behavior.

Side note: do you have similar requests about `cat(xs...; dims = 1)`/`vcat` for `OneHotArray`s?

simeonschaub:
Ah, I see, so the goal here is just to preserve one-hotness as much as we can, to take advantage of fast-paths. Since `OnehotArray` is an `AbstractArray`, I think we should always try to stick to the standard `AbstractArray` interfaces. Having special-cased overloads for Base functions, which may also return more convenient types is completely fine, but it should just be an optimization, I don't think we should deviate from the semantics of these functions and add stricter requirements. That's also why I am a bit skeptical of the proposal in this PR, since it implicitly `collect`s in some cases, which goes against the usual semantics of `reshape` that it is lazy and will not allocate. I wonder if we maybe could adapt some of the fast paths to `ReshapedArray{T,N,<:OnehotArray}` as well and thereby avoid rewrapping in `reshape` altogether.

> Side note: do you have similar requests about cat(xs...; dims = 1)/vcat for OneHotArrays?

Yes, I think by the same reasoning that should be allowed as well. In that case, I don't think the type instabilities would be that big of a deal, since `cat` in Base is type unstable already and typically `dims` will be a constant.

darsnack:
> I wonder if we maybe could adapt some of the fast paths to ReshapedArray{T,N,<:OnehotArray} as well and thereby avoid rewrapping in reshape altogether?

That makes sense. Let me look into what that would entail for the paths we want to hit.

@DhairyaLGandhi are you okay with allowing `vcat` for `OneHotArray`s that returns a `Bool` array? Previously, you had some reservations.

CarloLucibello:
`reshape` is not part of the [abstract array interface](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-array) (and `vcat` as well) so we don't necessarily have to support all kinds of reshapes, especially if not semantically meaningful. I really wonder what was @simeonschaub's specific use case  for reshaping the first dimension, I couldn't guess

darsnack:
That's true. I am pretty sure for these use cases, `collect` before `reshape` should fix this issue for Flux#master. I'm mentioning this for anyone else's benefit that comes across this.

Also, we can should merge this PR with the simple fallback fixes so that they make it in for v0.12. I can submit an additional PR extending `ReshapedArray`.

mcabbott:
`vcat` follows from indexing, it seems pretty odd to break it, for something <: AbstractArray. Was there an argument for breaking that, instead of just leaving the default?  
```
julia> struct Mine <: AbstractVector{Bool} x end
       Base.size(x::Mine) = size(x.x)
       Base.getindex(x::Mine, i::Integer) = x.x[i]
       vcat(Mine(randn(2).>0), Mine(randn(2).>0))
4-element Array{Bool,1}:
 1
 0
 0
 1
```

darsnack:
I think the argument is the same:
> I guess the actual reason might have been that when you reshape the first dimension, the array is no longer one-hot, so we erred on the side of alerting the user instead of silently converting (the same was done for cating along the first dimension).

But I feel now that these fallback cases are never on a critical path, so it isn't so bad to silently do what's expected instead of forcing a manual `collect`.

mcabbott:
I guess the alternative would not to pretend to be an array, and then it would only ever work on the intended straight-and-narrow. But if it is an array, then it seems like it ought to work in weird things people think up. Although where it's better to collect vs. producing some multiply wrapped thing I don't know.

darsnack:
So FYI, supporting these fallback paths is not hard to add, and in both cases we just convert to a `Bool` array (so after the conversion Base supports what happens).

Most operations on a `OneHotArray` will work as a result of `getindex`. I now agree that if we specialize a few operations for speed, we should continue to support the slower fallback path.

darsnack:
For the `reshape` case, it would be nice to have a lazy iterator to avoid data movement. Supporting it would help both the fallback and non-fallback cases.

mcabbott:
Maybe this was said, but the one downside of simple fallbacks is type-stability. I guess `cat` isn't type-stable anyway but `hcat` could be; `reshape` I don't know if there's an easy solution, you could have a different function `reshape1` which does insist on making a onehotarray? Not sure how much this matters.

darsnack:
Yeah, `hcat` can easily be made type stable in this PR. I guess it depends where these operations show up in practice. The reason for pushing for `OneHotArray`s as much as possible is because they were in #1447. If that's on a critical path then forcing `collect` might be a better option. We can make the error message more informative.

simeonschaub:
> reshape is not part of the abstract array interface (nor is vcat) so we don't necessarily have to support all kinds of reshapes, especially if not semantically meaningful.

This part in the manual just describes a minimal number of functions new subtypes of AbstractArray should implement so generic functions for AbstractArrays work like they are supposed to. Interfaces in Julia currently aren't very formalized, but I definitely believe it's bad style to break (although maybe only implicitly assumed by everyone) promises of generic function. I think most users would assume `reshape(a, dims...)` to mean "lazily give me the underlying column-major data of `a`, but indexed as if it had size `dims`". If `a` has special properties, those might not necessarily be preserved under reshape, for example, we allow `reshape(::Transpose, ...)` to work, even though the result might not be representable as a `Transpose` object. For generic code, it is often critical, that these promises still hold, since at the end of the day subtypes of `AbstractArray` should just be indexable containers – any additional properties are secondary.

> I really wonder what was @simeonschaub's specific use case for reshaping the first dimension, I couldn't guess

In my case, I added a singleton dimension in front, so I could broadcast with another array. This is something that's definitely not rare in generic code and it is quite annoying to work around if it doesn't work in all cases.

CarloLucibello:
@simeonschaub we lose type stability for reshape though, is allowing to reshape the first dimension worth it?

simeonschaub:
Not necessarily, see my first comment, where I proposed not overloading `reshape` at all, but widening some fast paths to accept `ReshapedArray` as well.

DhairyaLGandhi:
I think that's the better approach

CarloLucibello:
> Not necessarily, see my first comment, where I proposed not overloading reshape at all, but widening some fast paths to accept ReshapedArray as well.

Are you suggesting to *always* wrap in a ReshapedArray when reshaping? Array wrappers are always a bit annoying to work with, and they sometimes interact badly with the cuda infrastructure AFAIK. I'm not super familiar with the topic, but there has to be some reason why reshaping an Array returns an Array, while ReshapedArray is used as a generic fallback for AbstractArrays

mcabbott:
> widening some fast paths to accept ReshapedArray

What fast paths are there, BTW? Just `*` and argmax, or have I missed some? 

I wondered whether `onehot .* matrix` ought to specialise, and `onehot .* fun.(matrix)` could too, but not sure how much you gain especially once Zygote is involved.

DhairyaLGandhi:
bumping @darsnack to widen the types and get rid of the `reshape` overload.

darsnack:
I am assuming you are referring to the `ReshapedArray` implementation. Planning on working on that today after I address the outstanding Metalhead.jl PR which is blocking other work.

darsnack:
Sorry for the delay. I swapped overloading `reshape` to allowing `ReshapedArray{T, L, <:OneHotArray}` to hit the fast paths too. This works by widening the types of the fast path functions and using the reshaped `x.indices` for reshaped one hot arrays. To prevent reshaped one hot arrays that are no longer one hot (i.e. first dim is reshaped) from hitting the fast paths, the array is converted to `Array{Bool}`/`CuArray{Bool}` before reshaping. It will still be a wrapped `ReshapedArray`, just not treated as one hot.

darsnack:
Bump

src/onehot.jl
```diff
@@ -9,13 +9,30 @@ OneHotArray(indices::T, L::Integer) where {T<:Integer} = OneHotArray{T, L, 0, T}
...
+# e.g. argmax
+const OneHotLike{T, L, N, var"N+1", I} =
+  Union{OneHotArray{T, L, N, var"N+1", I},
+        Base.ReshapedArray{Bool, var"N+1", <:OneHotArray{T, L}}}
+
+# when reshaping a OneHotArray and first(dims) != L
+# convert the parent array to Array{Bool}
+# so that the ReshapedArray does not hit fast paths
+function Base.ReshapedArray(parent::OneHotArray{<:Any, L}, dims::NTuple{N,Int}, mi) where {L, N}
+  parent = (first(dims) != L) ? convert(_onehot_bool_type(parent), parent) : parent
```
simeonschaub, line 31: I think it would be better not to make a copy in the constructor, but instead add this check to the `cat` and `argmax` overloads.

src/onehot.jl
```diff
@@ -28,34 +39,30 @@ Base.getindex(x::OneHotArray{<:Any, L}, ::Colon, I...) where L = OneHotArray(x.i
...
 
 Base.BroadcastStyle(::Type{<:OneHotArray{<:Any, <:Any, <:Any, N, <:CuArray}}) where N = CUDA.CuArrayStyle{N}()
 
-Base.argmax(x::OneHotArray; dims = Colon()) =
-  (dims == 1) ? reshape(CartesianIndex.(x.indices, CartesianIndices(x.indices)), 1, size(x.indices)...) :
-                argmax(convert(_onehot_bool_type(x), x); dims = dims)
+Base.argmax(x::OneHotLike; dims = Colon()) =
+  (_isonehot(x) && dims == 1) ?
+    reshape(CartesianIndex.(_indices(x), CartesianIndices(_indices(x))), 1, size(_indices(x))...) :
+    argmax(convert(_onehot_bool_type(x), x); dims = dims)
```
simeonschaub, line 65:
```suggestion
    invoke(argmax, Tuple{AbstractArray}, x; dims = dims)
```

src/onehot.jl
```diff
@@ -135,11 +142,17 @@ function onecold(y::AbstractArray, labels = 1:size(y, 1))
...
+  else
+    return _fast_argmax(convert(_onehot_bool_type(x), x))
+  end
+end
 
 @nograd OneHotArray, onecold, onehot, onehotbatch
 
-function Base.:(*)(A::AbstractMatrix, B::OneHotArray{<:Any, L}) where L
+function Base.:(*)(A::AbstractMatrix, B::OneHotLike{<:Any, L}) where L
   size(A, 2) == L || throw(DimensionMismatch("Matrix column must correspond with OneHot size: $(size(A, 2)) != $L"))
```
simeonschaub, line 156:
```suggestion
  _isonehot(B) || return invoke(*, Tuple{AbstractMatrix, AbstractMatrix}, A, B)
  size(A, 2) == L || throw(DimensionMismatch("Matrix column must correspond with OneHot size: $(size(A, 2)) != $L"))
```

src/onehot.jl
```diff
@@ -135,11 +142,17 @@ function onecold(y::AbstractArray, labels = 1:size(y, 1))
...
 
 _fast_argmax(x::AbstractArray) = dropdims(argmax(x; dims = 1); dims = 1)
-_fast_argmax(x::OneHotArray) = x.indices
+function _fast_argmax(x::OneHotLike)
+  if _isonehot(x)
+    return _indices(x)
+  else
+    return _fast_argmax(convert(_onehot_bool_type(x), x))
+  end
+end
```
simeonschaub, line 151: Is this function even needed anymore?

simeonschaub (commented):
Just some small suggestions, sorry for holding this up. Otherwise looks good!

src/onehot.jl
darsnack: It's needed for the fastest path (i.e. when `x isa OneHotArray`). `Base.argmax` returns cartesian indices which incur an overhead when compared using `x.indices` directly.

darsnack: We can avoid the `_isonehot` check though with your changes.

simeonschaub: Ah, I see. Makes sense

simeonschaub (approved):

darsnack:
Thanks, think I need an approval from @CarloLucibello or @DhairyaLGandhi

src/onehot.jl
```diff
@@ -9,13 +9,24 @@ OneHotArray(indices::T, L::Integer) where {T<:Integer} = OneHotArray{T, L, 0, T}
...
 OneHotMatrix(indices, L) = OneHotArray(indices, L)
 
+# use this type so reshaped arrays hit fast paths
+# e.g. argmax
+const OneHotLike{T, L, N, var"N+1", I} =
+  Union{OneHotArray{T, L, N, var"N+1", I},
+        Base.ReshapedArray{Bool, var"N+1", <:OneHotArray{T, L, <:Any, <:Any, I}}}
+
+_isonehot(x::OneHotArray) = true
+_isonehot(x::Base.ReshapedArray{<:Any, <:Any, <:OneHotArray{<:Any, L}}) where L = (size(x, 1) == L)
```
DhairyaLGandhi, line 28: Is this check correct?  The first dimension equal to the number of available features?

src/onehot.jl
```diff
@@ -9,13 +9,24 @@ OneHotArray(indices::T, L::Integer) where {T<:Integer} = OneHotArray{T, L, 0, T}
...
 const OneHotVector{T, L} = OneHotArray{T, L, 0, 1, T}
 const OneHotMatrix{T, L, I} = OneHotArray{T, L, 1, 2, I}
 
 OneHotVector(idx, L) = OneHotArray(idx, L)
 OneHotMatrix(indices, L) = OneHotArray(indices, L)
 
+# use this type so reshaped arrays hit fast paths
+# e.g. argmax
+const OneHotLike{T, L, N, var"N+1", I} =
+  Union{OneHotArray{T, L, N, var"N+1", I},
```
DhairyaLGandhi, line 24: Man this N+1 is tripping me up, I would say we need to remove this soon. Where is it used exactly?

src/onehot.jl
DhairyaLGandhi: Is there a way to avoid it?

DhairyaLGandhi (commented):
Overall I find the implementation to have more "helper" methods than I'd care for, seems like we ought to make oharray code more generic, but I guess that's going to have to wait for a different day

src/onehot.jl
DhairyaLGandhi: Do you think we could calculate var"N+1" during runtime?

DhairyaLGandhi (approved):

src/onehot.jl
darsnack: Yes, if a reshaped array has the same first dimension as the parent array, then that means that the one-hot behavior has been preserved.

src/onehot.jl
darsnack: I don't like it either! It can't be done at runtime since `N` and `var"N+1"` are used in the type specification. `N` is used to specify the type of the index array, and `var"N+1"` is used to inherit from `AbstractArray{Bool, var"N+1"}`. Neither is evaluated at runtime.

darsnack: We could change it to another variable. I don't have strong feelings, but a part of me says that at least this naming signals the intent of the type parameter.

darsnack:
Ah apparently I don't have merge rights. So you will still need to merge this.

src/onehot.jl
DhairyaLGandhi: To be fair, I did mean we would have to switch it out during construction, because I don't think it's any better for dispatch to have to do checks on ints than types. To me it suggests that it is a preknown quantity so adding it to the type doesn't win us much.

DhairyaLGandhi:
Hmm..

Let try again

bors r+

PR #1462, opened 2021-01-12
Add Parallel layer

darsnack:
Since #1289 stalled, I have added an implementation of `Parallel` with some of the changes we discussed during ML calls. This version excludes most of the structural layers in #1289 like `Join`, `Split`, and `Nop`. I also added the ability for the user to specify the reduction operator. If it is acceptable, I would like to remap `SkipConnection` to `Parallel` (not a deprecation exactly).

The reason for submitting this PR now is because I am creating pre-trained weights for the networks in https://github.com/FluxML/Metalhead.jl/pull/70, and there is a lot of code that can be replaced with a `Parallel`. So, I'd like to have `Parallel` in Flux before continuing with training to make the process easier.

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md
- [x] Final review from @DhairyaLGandhi (for API changes).

cc @CarloLucibello


darsnack:
Also, I'd like to add that `outputsize` worked out of the box with no changes when I added `Parallel`. So that's a small win to celebrate.

src/layers/basic.jl
```diff
@@ -253,3 +253,50 @@ end
...
+```
+"""
+struct Parallel{F, T}
+  connection::F
+  layers::T
+end
+
+Parallel(connection, layers...) = Parallel(connection, layers)
+
+Flux.@functor Parallel
```
atiyo, line 291:
```suggestion
@functor Parallel
```

I think this keeps better consistency with the rest of the file.

src/layers/basic.jl
```diff
@@ -253,3 +253,50 @@ end
...
+  Dense(1, 1),
+  Parallel(
+    vcat,
+    Dense(1, 1),
+    Dense(1, 3),
+    Chain(
+      Dense(1, 5),
+      Dense(5, 2),
+    )
+  ),
```
DhairyaLGandhi, line 277:
```suggestion
  Parallel(vcat, Dense(1, 1), Dense(1, 3))
```

src/layers/basic.jl
```diff
@@ -253,3 +253,50 @@ end
 function Base.show(io::IO, b::SkipConnection)
   print(io, "SkipConnection(", b.layers, ", ", b.connection, ")")
 end
+
+"""
+    Parallel(connection, layers...)
+
+Create a new 'Parallel' layer that passes a single input array each path in
+`layers`, combining the output of each path with `connection`.
+`connection` should be a reducible operator (i.e. it can be passed to `Base.reduce`).
```
DhairyaLGandhi, line 262:
```suggestion
Create a 'Parallel' layer that passes an input array to each path in
`layers`, reducing the output with `connection`.

Equivalent to calling `reduce(connection, [l(x) for l in layers]...)
```

src/layers/basic.jl
```diff
@@ -253,3 +253,50 @@ end
...
+(m::Parallel)(x::AbstractArray) = mapreduce(f -> f(x), m.connection, m.layers)
+
+Base.getindex(m::Parallel, i::Integer) = m.layers[i]
+Base.getindex(m::Parallel, i::AbstractArray) = Parallel(m.connection, m.layers[i]...)
+
+function Base.show(io::IO, m::Parallel)
+  print(io, "Parallel(", m.connection, ", ")
+  join(io, m.layers, ", ")
+  print(io, ")")
+end
```
DhairyaLGandhi: can we fix the end of page too

DhairyaLGandhi (commented):
Thanks, do we also want to add the functionality from having multiple arguments in the forward pass as in parallel execution and reduction? Seems like there's some of that in

darsnack:
Adding multiple input arguments would be like `Join` from #1289. We could add a `Vararg` version of `Parallel` that implements that behavior.

darsnack:
Actually on second thought, maybe we shouldn't have the multi-arg version? It complicates the semantics on `Chain`, and I don't think it is used in practice for anything.

darsnack:
@DhairyaLGandhi any decision on keeping vs excluding multiple input arguments for the layer? I would recommend excluding.

src/layers/basic.jl
```diff
@@ -253,3 +253,47 @@ end
...
+
+Equivalent to calling `reduce(connection, [l(x) for l in layers]...)`.
+
+# Examples
+
+```jldoctest
+julia> model = Chain(Dense(1, 1), Parallel(vcat, Dense(1, 1), Dense(1, 3), Chain(Dense(1, 5), Dense(5, 2))), Dense(6, 1));
+
+julia> size(model(rand(1)))
+(1,)
```
mcabbott, line 271: May I suggest making these numbers all distinct?
```suggestion
julia> model = Chain(Dense(3, 13, tanh), Parallel(vcat, Dense(13, 4), Chain(Dense(13, 7, tanh), Dense(7, 4))), Dense(8, 17));

julia> size(model(rand(3)))
(17,)
```

src/layers/basic.jl
```diff
@@ -253,3 +253,47 @@ end
...
   print(io, "SkipConnection(", b.layers, ", ", b.connection, ")")
 end
+
+"""
+    Parallel(connection, layers...)
+
+Create a 'Parallel' layer that passes an input array to each path in
+`layers`, reducing the output with `connection`.
+
+Equivalent to calling `reduce(connection, [l(x) for l in layers]...)`.
```
mcabbott, line 263: I think there's no `...`, and that mentioning the multiple input case might be nice:
```suggestion
Called with one input `x`, this is equivalent to `reduce(connection, [l(x) for l in layers])`.
If called with multiple inputs, they are `zip`ped with the layers, thus `Parallel(+, f, g)(x, y) = f(x) + g(y)`.
```

src/layers/basic.jl
```diff
@@ -253,3 +253,48 @@ end
...
+Create a 'Parallel' layer that passes an input array to each path in
+`layers`, reducing the output with `connection`.
+
+Called with one input `x`, this is equivalent to `reduce(connection, [l(x) for l in layers])`.
+If called with multiple inputs, they are `zip`ped with the layers, thus `Parallel(+, f, g)(x, y) = f(x) + g(y)`.
+
+# Examples
+
+```jldoctest
+julia> model = Chain(Dense(3, 13, tanh), Parallel(vcat, Dense(13, 4), Chain(Dense(13, 7, tanh), Dense(7, 4))), Dense(8, 17));
```
DhairyaLGandhi, line 269:
```suggestion
julia> model = Chain(Dense(3, 5),
                     Parallel(vcat, Dense(5, 4), Chain(Dense(5, 7), Dense(7, 4))),
                     Dense(8, 17));
```

src/layers/basic.jl
```diff
@@ -253,3 +253,48 @@ end
...
+
+Parallel(connection, layers...) = Parallel(connection, layers)
+
+@functor Parallel
+
+(m::Parallel)(x::AbstractArray) = mapreduce(f -> f(x), m.connection, m.layers)
+(m::Parallel)(xs::Vararg{<:AbstractArray}) = mapreduce((f, x) -> f(x), m.connection, m.layers, xs)
+
+Base.getindex(m::Parallel, i::Integer) = m.layers[i]
+Base.getindex(m::Parallel, i::AbstractArray) = Parallel(m.connection, m.layers[i]...)
```
DhairyaLGandhi, line 294:
```suggestion
Base.getindex(m::Parallel, i::AbstractVector) = Parallel(m.connection, m.layers[i]...)
```

DhairyaLGandhi (commented):
The vararg AbstractArray method is the one I had in mind. I would also want to see some addition to the docs that talks about implementing the `Join` and `Split` because I feel we would inevitably have questions about wanting them.

Otherwise, lgtm

src/layers/basic.jl
darsnack: Was there a reason for removing `tanh`?

DhairyaLGandhi: It's unclear what it adds to the example of the layer working and we want examples to be concise and self contained

darsnack:
What about `SkipConnection` being syntactic sugar for `Parallel` with `identity` on one of the paths?

DhairyaLGandhi:
I'm in favour of retaining the struct because it's a very common case, but rewriting it in a follow up

DhairyaLGandhi:
@darsnack I was hoping the docs would show how to implement `Join` and `Combine` in terms of `Parallel`.

darsnack:
Note that `Join` is the same as `Parallel` with variable arguments. There isn't much work required to implement `Join` in terms of `Parallel`. So I just added a subsection to the docs. This should be more instructive for users.

darsnack:
@DhairyaLGandhi bump

DhairyaLGandhi (approved):
Lgtm, thanks @darsnack

bhvieira:
Wouldn't it be better to deprecate `SkipConnection` instead of removing it outright?

darsnack:
@bhvieira Currently, it is still in Flux as it was before. My plan was not to deprecate it but to define:
```julia
SkipConnection(layers, connection) = Parallel(connection, layers, identity)
```

I think the `SkipConnection` syntax/nomenclature is still useful and meaningful.

bhvieira:
@darsnack Yeah, I deleted my comment before seeing yours actually, that's what I had in mind! Good job btw!

DhairyaLGandhi:
Can we add follow on testing on the GPU, that seems to have been missed? My bad.

PR #1468, opened 2021-01-16
add Upsample and PixelShuffle layers

CarloLucibello:
### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


src/layers/upsample.jl
```diff
@@ -0,0 +1,34 @@
...
+  scale_factor::NTuple{N, T}
+end
+
+function Upsample(; scale_factor::NTuple{N, T}, mode::Symbol) where {N,T}
+  @assert mode in [:bilinear] "Not supported mode '$mode'"
+  @assert N == 2  "Only 2d upsampling currently supported"
+  Upsample{Val{mode}, N, T}(scale_factor)
+end
+
+(m::Upsample{Val{:bilinear}})(x) = NNlib.upsample_bilinear(x, m.scale_factor)
```
mcabbott, line 20: Can I fight off the underscore? It seems unambiguous. 
```suggestion
    Upsample(; scale, mode)

An upsampling layer.
`scale`: a tuple of ints. 
`mode`: only `:bilinear` currently supported. 

See [`NNlib.upsample_bilinear`](@ref).
"""
struct Upsample{Mode, N, T} 
  scale::NTuple{N, T}
end

function Upsample(; scale::NTuple{N, T}, mode::Symbol) where {N,T}
  @assert mode in [:bilinear] "Not supported mode '$mode'"
  @assert N == 2  "Only 2d upsampling currently supported"
  Upsample{Val{mode}, N, T}(scale)
end

(m::Upsample{Val{:bilinear}})(x) = NNlib.upsample_bilinear(x, m.scale)
```

src/layers/upsample.jl
```diff
@@ -0,0 +1,48 @@
...
+  scale::NTuple{N, T}
+end
+
+function Upsample(; scale::NTuple{N, T}, mode::Symbol) where {N,T}
+  @assert mode in [:bilinear] "Unsupported mode '$mode'"
+  @assert N == 2  "Only 2d upsampling currently supported"
+  Upsample{Val{mode}, N, T}(scale)
+end
+
+(m::Upsample{Val{:bilinear}})(x) = NNlib.upsample_bilinear(x, m.scale)
```
mcabbott, line 34: Hope you don't mind the huge suggestion. My main point here is that I don't think there's much gained by storing `Val(mode)` in the type, we can just put the symbol there directly. It should have a default. 

My secondary point is that `scale=2` would be nice, like `pad=2` means "in all directions". And at that point it seemed quicker to write than to talk... 

```suggestion
struct Upsample{Mode,T}
  scale::T
end

function Upsample(; scale, mode::Symbol=:bilinear)
  mode in [:bilinear] || throw(ArgumentError("only bilinear sampling is supported"))
  scale isa Integer || scale isa NTuple{2,<:Integer} || 
    throw(ArgumentError("only two-dimensional scaling by an integer factor is supported"))
  Upsample{mode, typeof(scale)}(scale)
end

(m::Upsample{:bilinear})(x) = NNlib.upsample_bilinear(x, m.scale isa Tuple ? m.scale : (m.scale, m.scale))

function Base.show(io::IO, u::Upsample2{mode}) where {mode}
  print(io, "Upsample(scale=", u.scale)
  mode == :bilinear || print(io, ", mode=", mode)
  print(io, ")")
end
```

src/layers/upsample.jl
```diff
@@ -0,0 +1,56 @@
...
+[:, :, 1, 1] =
+ 1.0  1.0  1.0
+ 1.0  1.0  1.0
+```
+"""
+struct Upsample{Mode,T}
+  scale::T
+end
+
+function Upsample(; scale::Union{Int, NTuple{N,Int}}, mode::Symbol) where N
```
DhairyaLGandhi, line 28: `scale` et al should be positional

src/layers/upsample.jl
```diff
@@ -0,0 +1,56 @@
...
+
+
+"""
+    PixelShuffle(r::Int)
+
+Pixel shuffling layer with upscale factor `r`.
+ 
+See [`NNlib.pixel_shuffle`](@ref).
+"""
+struct PixelShuffle
```
DhairyaLGandhi, line 52: Parameterize

DhairyaLGandhi (changes requested):

CarloLucibello:
> My main point here is that I don't think there's much gained by storing Val(mode) in the type, we can just put the symbol there directly. It should have a default.

I didn't know you could do that. I'd like to be cautious and keep it as a mandatory argument, at least for some time, because the default mode for pytorch and keras is nearest neighbor, that we don't have yet.

DhairyaLGandhi:
I'm not really worked up about what other frameworks do and follow that necessarily.

src/layers/upsample.jl
CarloLucibello: since it is likely in the future we will support both `outsize` and `scale`, they will end up being keyword arguments,
so better do it now to avoid future breakage.

CarloLucibello:
> I'm not really worked up about what other frameworks do and follow that necessarily.

Reproducibility is really important in deep learning research (but also production). And current research and productions is using either pytorch or tensorflow, so we shouldn't make crazy hard for people to figure out why they are not able to reproduce published results or their own python code using flux

src/layers/upsample.jl
CarloLucibello: with what?

DhairyaLGandhi: With the type of `r` presumably

src/layers/upsample.jl
DhairyaLGandhi: I would disagree with that. Of course wrapping it in Val would be the preferred way to go, future breakages do not mean we should have non Julian API

DhairyaLGandhi:
I don't see how it is hard or intend to force other frameworks design on to ours

src/layers/upsample.jl
mcabbott: Your concern is `Int == Int32` or something? I think that's handled automatically.

src/layers/upsample.jl
mcabbott: When would the output type depend on the value of the scaling? Otherwise no need for Val, surely.

mcabbott:
> default mode for pytorch and keras is nearest neighbor, that we don't have yet

I guess if there's a default we can't later change it, although not sure linear is a bad default. It could be always printed to make this more visible. 

If it must be typed, then perhaps it should be `order=1` for linear, 0 for nearest-neighbour?

CarloLucibello:
If someone tries to translate pytorch's `Upsample(scale_factor=(2,2))` with flux's `Upsample(scale=(2,2))` as it may feel natural, they would be doing a wrong port, where the mistake in the upsampling layer would be really hard to spot in a model containing many other layers. A mandatory argument forces you to think and compare. I'm not saying we shouldn't have a default, I'm saying we should stick with conventions since most of the time they are there for a good reason, and depart only if there are other good reasons to, and I see none here. I'm just being cautious since we are introducing these layers for the first time, adding a default in the future is a non-breaking change, the other way around is breaking instead.

DhairyaLGandhi:
I don't think it's unreasonable to have both, that was exactly what had us add the multiple constructors for Conv with kwargs, and later had them removed for being not used enough (how did we determine this?)

I don't think it's 'unnatural' in any way to omit a keyword given that they are moving to a different framework with a different language with different semantics, with convenience constructors on top for the 'natural' transition and later moved to the more language/ framework specific implementation.

CarloLucibello:
> If it must be typed, then perhaps it should be order=1 for linear, 0 for nearest-neighbour?

I don't understand, why the symbol is not enough?

Btw, we could have `:linear` to represent 1d linear, :bilinear and :trilinear, let forward pass dispatch handle the appropriate NNlib call (or unify also those). 
We could still allow  passing `:bilinear` and `:trilinear` modes, since they are commonly used names. 

> I don't think it's 'unnatural' in any way to omit a keyword given that they are moving to a different framework with a different language with different semantics, with convenience constructors on top for the 'natural' transition and later moved to the more language/ framework specific implementation.

I'm not saying it is unnatural to use a positional arg instead of a keyword, I'm saying it is natural when porting to omit the `mode` argument if not present in the original code and that would lead to an incorrect port if the defaults differ.

Anyways, we can implement right now the following
```julia
Upsample(scale, mode)
# or
# Upsample(scale, mode=:linear)

Upsample(; scale, mode)
# or
# Upsample(; scale, mode=:linear)
```
Later probably change the keyword constructor to
```julia
Upsample(; scale=nothing, outsize=nothing, mode) # require inside either scale or outsize != nothing
# or
# Upsample(; scale=nothing, outsize=nothing, mode=:linear)
```

CarloLucibello:
Another option, which corresponds to the direction taken by @maxfreu  in https://github.com/FluxML/NNlib.jl/pull/266, is to have just one constructor
```julia
Upsample(scale=1; outsize=nothing, mode)
```
where `scale` is ignored if `outsize` is given.

DhairyaLGandhi:
We can consider that at a later point in time

mcabbott:
https://github.com/FluxML/NNlib.jl/pull/269 adds `upsample_nearest`. 

I guess that using identical notation and very different defaults is likely to trip people up unnecessarily. If we want different defaults, then perhaps we should pick a different name -- perhaps `Resize` or `Resample` are possibilities. 

We could allow positional `Upsample(2)` now and `Upsample(size=(100,100))` later (default to `nothing` & check in the body, or whatever).

maxfreu:
Hi, I would keep scale as positional argument, because it is most often used in DL (and most often it is 2), so writing it out every time would be too verbose. The rest should be keywords. Nearest is a reasonable default mode because it is fastest + and already default in other frameworks (pyt & tf). I would keep the spelled out symbol form and document them well. Furthermore scale should not be restricted to int here, as the implementations will support reals at some point. So error at a lower level.

src/layers/upsample.jl
DhairyaLGandhi: My concern is consistency. Also I think that conversion is for `Integer`?

DhairyaLGandhi:
Needs GPU tests

CarloLucibello:
@maxfreu what about having these 2 constructors
```julia
Upsample(scale; mode=:nearest)
Upsample(; scale=nothing, outsize=nothing, mode=:nearest)
```
?

DhairyaLGandhi:
Having all keyword constructors is fine, but if we don't support multiple modes etc, let's not make it an option and rather document it properly.

Also, we should let users tweak the outsize in the first constructor as well, and check for compatible settings

src/layers/upsample.jl
mcabbott: > My concern is consistency. Also I think that conversion is for Integer?

Consistency with what? Conversion from what? It's an `Int`, the simplest possible bits type.

src/layers/upsample.jl
```diff
@@ -0,0 +1,56 @@
...
+ 1.0  1.0  1.0
+ 1.0  1.0  1.0
+```
+"""
+struct Upsample{Mode,T}
+  scale::T
+end
+
+function Upsample(; scale::Union{Int, NTuple{N,Int}}, mode::Symbol) where N
+  mode in [:bilinear] ||
```
maxfreu: nearest should work now as well :)

CarloLucibello: I think you commented an old commit

maxfreu: True, sorry.

src/layers/upsample.jl
```diff
@@ -0,0 +1,66 @@
...
+"""
+struct Upsample{Mode,T}
+  scale::T
+end
+
+Upsample(; scale, mode::Symbol=:nearest) = Upsample(scale; mode)
+
+function Upsample(scale; mode::Symbol=:nearest)
+  mode in [:nearest, :bilinear] || 
+    throw(ArgumentError("`:$mode` mode is not supported."))
```
mcabbott, line 34:
```suggestion
    throw(ArgumentError("mode=:$mode is not supported, allowed values are :nearest, :bilinear"))
```

maxfreu:
> @maxfreu what about having these 2 constructors

Looks good to me! @mcabbott suggested to use `size` as a keyword and `Base.size` where needed. What do you think?

src/layers/upsample.jl
CarloLucibello: I'd rather not keep adding the allowed values in error message, the list is going to become pretty long pretty soon

CarloLucibello:
> Looks good to me! @mcabbott suggested to use size as a keyword and Base.size where needed. What do you think?

I think it's fine. I'll have to wait for your PR in NNllib to be merged before adapting the wrapper here though

src/layers/upsample.jl
maxfreu: When you add it to the line above (`[:xy, ...]`), you can add it to the line below to give dumb users like me a hint :D

DhairyaLGandhi:
Best not to create ambiguity when reading the code anyway

src/layers/upsample.jl
DhairyaLGandhi: Other layers

maxfreu:
> Best not to create ambiguity when reading the code anyway

So you opt for `outsize`, right?

mcabbott:
> Best not to create ambiguity when reading the code anyway

My reading was that supplying both the size and the scale was the ambiguity in question? 

Anyway maybe we should merge this & return to further methods in another PR.

My reservation about mode, one last time, is that it seems a little weird to need to spell the dimension out (in latin!), although perhaps never a problem in practice:
```
Chain(Upsample(2, method=:linear), Conv((3,), 5=>7), ...)
Chain(Upsample(2, method=:bilinear), Conv((3,3), 5=>7), ...)
Chain(Upsample(2, method=:trilinear), Conv((3,3,3), 5=>7), ...)
```

maxfreu:
For info: I changed `outsize` to `size` in my PR.

DhairyaLGandhi:
> So you opt for outsize, right?

As names go, yes I would pick this over size. Depending on the case, `dims` and `axis` works, and maybe `sz` as a shorthand

> If we do spell it out, I also wonder a bit whether it need be a keyword, since Upsample(2, :blinear) seems hard to misinterpret.

I agree, and I think that's the api that we should adopt for this layer.

mcabbott:
Yes, I think that if we allow more options than `scale, mode`, that's when keywords start to be needed. Doesn't hurt to allow them from the start I guess. 

The obvious precedent for `size` as a keyword is that Base does `range(1, 2, length=100)` and discourages things like `range(1, 2, len=100)` (partial word) or `range(1, 2, outlength=100)` (no added clarity, longer & harder to remember).

DhairyaLGandhi:
Ci isn't happy; do we need to have a nnlib and cuda branch in the manifest for this? There's also an undefined variable currently

CarloLucibello:
update reflecting latest NNlib updates. I'm not totally satisfied with the interface:
```julia
Upsample(scale; mode=:nearest)  
Upsample(; size, mode=:bilinear)
```
Since the `nearest` method doesn't support `size`, I had to use 2 different defaults for `mode`.

mcabbott:
Yes if that simplifies things here it seems fine. We could plausibly add an `upsample_nearest` which does non-integer scaling, at some point, and it would be neat if no changes here were required for that.

DhairyaLGandhi:
Two defaults seems like it would lead to confusion

maxfreu:
Maybe keep nearest as default and just error if reals are given? Later, just the error has to be removed.

src/outputsize.jl
```diff
@@ -35,7 +35,7 @@ Base.isless(::Nil, ::Number) = true
 Base.isless(::Number, ::Nil) = true
 
 Base.isnan(::Nil) = false
-
+Base.isfinite(::Nil) = true
```
CarloLucibello, line 38: this is need by latest softmax implementation

CarloLucibello:
I think this is ready. 

A couple of seemingly unrelated oddities:
1) CI/Documentation is not able to fetch the correct NNlib version

2) @darsnack `outputsize` tests print an error that is outside of any testset, I could not figure out where it is coming from
```julia
Test Summary: | Pass  Total
Layers        |  251    251
┌ Error: layer Dense(5, 2), index 3 in Chain, gave an error with input of size (4,)
└ @ Flux ~/.julia/dev/Flux/src/outputsize.jl:107
┌ Warning: Slow fallback implementation invoked for ∇conv_data!  You probably don't want this; check your datatypes.
│   yT = Flux.NilNumber.Nil
│   T1 = Flux.NilNumber.Nil
│   T2 = Float32
└ @ NNlib ~/.julia/packages/NNlib/Pcmaw/src/conv.jl:206
Test Summary: | Pass  Total
outputsize    |   83     83
```

mcabbott (dismissed):
Looks fine. 

I presume doc CI will catch up in time? This PR adds no new doctests, so I don't think it's likely to break anything. Your call whether to wait or not.

I think the `outputsize` noise is from `@test_throws`, it logs an informative message before throwing. Could be made `@warn` not `@error` if you think this is too disturbing.

darsnack:
The error message was introduced in #1486. We can probably leave it as `@error` since NNlib/Flux will throw a dimension mismatch anyways (so it isn't like we have the option of warning and continuing).

darsnack:
@CarloLucibello what's the call on the doctest CI? #1147 is facing the same `softmax` issue. Do we want to merge this can make #1147 rebase?

DhairyaLGandhi:
Did you mean the orthogonal initiatization? If there's nothing else remaining I can do a quick once over and we can merge that first. The author has done a significant amount of work on it already there

DhairyaLGandhi:
Seems like an issue if the error isn't getting caught, can you check what is happening there @darsnack ?

I'm the meanwhile, @CarloLucibello can add the latest version to the environment locally, or set `ENV[JULIA_PACKAGE_SERVER]=""` and rerun ci

darsnack:
No the error is getting caught, but since that PR makes docstring changes, we should make sure the doctest CI passes. Unless the doctest doesn't halt on failures.

CarloLucibello:
I would merge as it is, I was just waiting for @DhairyaLGandhi's approval, problems seem unrelated. 

I don't fancy outputsize throwing an error, even if controlled, people can be confused when testing locally (actually I was the very first to be confused and wasted some time investigating it with no success). In any case, not an issue for this PR. 

> I'm the meanwhile, @CarloLucibello can add the latest version to the environment locally, or set ENV[JULIA_PKG_SERVER]="" and rerun ci

I don't know what that means, should I modify something in `ci.yml`? I don't understand why the ci documentationstep can't solve the environment, the Manifest.yml is fine. I'll just rerun the CI and see if it fixed itself

darsnack:
> I don't know what that means, should I modify something in ci.yml? I don't understand why the ci documentationstep can't solve the environment, the Manifest.yml is fine. I'll just rerun the CI and see if it fixed itself

Yeah there was a recent issue with resolving packages when using the Julia package servers. @giordano suggested on Slack to disable the package server for CI. So, yeah modifying the `ci.yml` to set the environment variable.

mcabbott:
> I don't fancy outputsize throwing an error, even if controlled, people can be confused when testing locally (actually I was the very first to be confused and wasted some time investigating it with no success).

To be clear what's going on here, it is throwing an error in the test, which is specifically testing that it does throw. But before it throws, it adds a logging message via `@error`. In real use, this would always immediately precede the actual error (and has different information to the actual error message). In this test it, like the other logging messages, appears in the CI logs.

This can obviously be changed if we decide it's too confusing to have in the logs, for instance by downgrading to `@warn`. (It's not really a warning, as it happens only when there is a real error... but possibly more eyeballs will see it in the CI logs than in real use!) I'm sorry about the wasted time.

src/layers/upsample.jl
```diff
@@ -0,0 +1,84 @@
...
+ 1.0  1.0  1.0
+ 1.0  1.0  1.0
+```
+"""
+struct Upsample{Mode,S,T}
+  scale::S
+  size::T
+end
+
+function Upsample(scale::S; mode::Symbol=:nearest) where S
```
DhairyaLGandhi, line 38: Still in favour of making mode positional

DhairyaLGandhi (commented):
Also needs GPU tests

CarloLucibello:
> Also needs GPU tests

gpu tests are already there

src/layers/upsample.jl
CarloLucibello: can do. I slightly prefer keyword because a symbol as positional looks weird to me, but I couldn't even say why. @darsnack has any preference here?

darsnack: I prefer keyword for things like this, because the expected positional ordering is ambiguous (especially when we have `size` in the other constructor). If there is an expectation, it's historical IMO.

Side note: I did not give this PR a full review since there were enough people engaged, but now that I have looked over the full discussion/code, I feel that mode should be the function used instead of a symbol.

CarloLucibello: the thumb down is for the mode proposal :)

CarloLucibello: The positional interface would read
```julia
Upsample((2,2), :nearest)
Upsample(:bilinear, size=(32,32))
```
Actually, I may not dislike it. @DhairyaLGandhi your call, I'm fine either way

darsnack: Now that I thought about it, I too 👎🏾 myself

darsnack: I'll retract both my statements. Also in favor of positional because it sidelines `size` well.

mcabbott: Another option would be to make `scale` a keyword, on equal footing with `size`. Which would then go well with positional `mode`.

darsnack: I think the argument that scale is the most likely parameter to be adjusted is a strong case to use it as the first positional.

darsnack: That being said the only API that looks clean to me is
```julia
Upsample(:linear, scale=(2,2))
Upsample(:nearest, size=(32,32))
```

Is it really that bad to have to type `scale`? Probably if you have that many of a certain layer, you would want to programmatically generate that model anyways.

DhairyaLGandhi: It's odd that the mode were to come first. `Upsample(scale, mode)` reads better to me, I would prefer to distinguish it from `Upsample(size, mode)` as well, but unsure of how without it being massively confusing.

that said, while I agree the scale/size keyword api is the most symmetric, (I think scale is the more predominant type in the wild, can someone verify? And could we make it more ergonomic). I guess forcing the keyword in this case does reduce ambiguity about what is happening, so I would be inclined to work with that as well.

We could have forced floats to represent scale, and assume size with ints otherwise, but that's confusing as well, so I'm not keen.

darsnack: Yes, very unfortunate that we can't just use dispatch here. We could do all keyword arguments too. Though unlike `scale` and `size`, the value of the mode (e.g. `:nearest`) easily implies what we're talking about. So if anything is positional for brevity, mode seems like the best candidate.

mcabbott: Writing `Upsample(scale=2)` doesn't seem too awful. 

You could allow also e.g. `Upsample(2, :bilinear)` and `Upsample(:bilinear, 2)`, since you can dispatch on the types, and they look very different to humans too. But seems less confusing to have just one true way. I think dispatching `2.0` to mean something very different to `2` would be a serious footgun.

CarloLucibello: So the possibilities are
1)
```julia
Upsample(mode::Symbol=:nearest; size=nothing, scale=nothing)
```
and we have to enforce in the constructor either size or scale to be != nothing. 
2)
```julia
Upsample(scale, mode::Symbol=:nearest)
Upsample(mode::Symbol=:nearest; size)
```
(hopefully it shouldn't cause an ambiguity error)
3)
Current proposal
```julia
Upsample(scale, mode::Symbol=:nearest)
Upsample(mode::Symbol=:nearest; size)
```

At this point the consensus seems to be geared toward 1) or 2).  Shall we go with 1) and call it a day? Or a mix of 1 and 2 like 
4)
```julia
Upsample(mode::Symbol=:nearest; size=nothing, scale=nothing)
Upsample(scale, mode::Symbol=:nearest)
```
I think any of these options are fine, let's just choose one and move on

darsnack: 1 and 4 look good to me. I think @DhairyaLGandhi can just pick one, and we go with that.

NEWS.md
```diff
@@ -12,6 +12,8 @@
 * Moved GPU CI to use buildkite instead of GitLab
 * New [`Parallel` layer](https://github.com/FluxML/Flux.jl/pull/1462) adds inception module-like building blocks.
 * Feature additions and bug fixes for BatchNorm, LayerNorm, InstanceNorm, and GroupNorm [normalization layers](https://github.com/FluxML/Flux.jl/pull/1397)
+* Added [Upsample and PixelShuffle layers](https://github.com/FluxML/Flux.jl/pull/1468)
+* Other new features and bug fixes (see GitHub releases page)
```
DhairyaLGandhi, line 16: can remove this

src/layers/upsample.jl
```diff
@@ -0,0 +1,84 @@
...
+function Upsample(; size::T, mode::Symbol=:nearest) where T
+  mode in [:nearest, :bilinear] || 
+    throw(ArgumentError("mode=:$mode is not supported."))
+  return Upsample{mode, Nothing, T}(nothing, size)
+end
+
+(m::Upsample{:nearest})(x::AbstractArray) =
+  NNlib.upsample_nearest(x, m.scale)
+function (m::Upsample{:nearest, Int})(x::AbstractArray{T, N}) where {T, N} 
+  NNlib.upsample_nearest(x, ntuple(i -> m.scale, N-2))
```
DhairyaLGandhi, line 53: `upasmple_nearest` should be able to handle given an Int directly right? This forces the input type to be specified. We should remove that from the other forward passes too, we don't seem to be using that information elsewhere anyway

DhairyaLGandhi: The forward pass should be nothing more than

```julia
(a::Upsample)(x) = a.mode(x, a.scale, a.sz)
```

test/cuda/layers.jl
```diff
@@ -79,6 +79,15 @@ gpu_gradtest("GroupNorm 3d", groupnorm, rand(Float32, 8, 8, 8, 12, 4), 12, 3, se
...
 
+upsample = [Upsample]
+gpu_gradtest("Upsample 2d", upsample, rand(Float32, 3, 4, 2, 3), (2,2))
+gpu_gradtest("Upsample 1d", upsample, rand(Float32, 3, 4, 2, 3), (2,))
+
+pixelshuffle = [PixelShuffle]
+gpu_gradtest("PixelShuffle 2d", pixelshuffle, rand(Float32, 3, 4, 18, 3), 3)
+gpu_gradtest("PixelShuffle 1d", pixelshuffle, rand(Float32, 3, 18, 3), 3)
+
+
```
DhairyaLGandhi, line 90: extra line

src/outputsize.jl
```diff
@@ -100,7 +100,7 @@ nil_input(pad::Bool, tup::Tuple{Vararg{Tuple}}) = nil_input(pad, tup...)
 function outputsize(m::Chain, inputsizes::Tuple{Vararg{Integer}}...; padbatch=false)
   x = nil_input(padbatch, inputsizes...)
   for (i,lay) in enumerate(m.layers)
-    try
+    try
```
DhairyaLGandhi, line 103: seems extra

test/cuda/layers.jl
```diff
@@ -79,6 +79,15 @@ gpu_gradtest("GroupNorm 3d", groupnorm, rand(Float32, 8, 8, 8, 12, 4), 12, 3, se
 gpu_gradtest("GroupNorm 2d", groupnorm, rand(Float32, 8, 8, 12, 4), 12, 3, setmode=true)
 gpu_gradtest("GroupNorm 1d", groupnorm, rand(Float32, 8, 3, 12, 4), 12, 3, setmode=true)
 
+upsample = [Upsample]
+gpu_gradtest("Upsample 2d", upsample, rand(Float32, 3, 4, 2, 3), (2,2))
+gpu_gradtest("Upsample 1d", upsample, rand(Float32, 3, 4, 2, 3), (2,))
```
DhairyaLGandhi, line 84: Doesn't the 1D case support giving it reals? We would need to fix that before we can release.

src/layers/upsample.jl
DhairyaLGandhi: 1 seems fine for the user facing API, and in that case I'd expect the type annotations (and possibly default mode) be removed. Why do we not want to give the sampling function as part of the API? Seems pretty reasonable to me, it makes debugging a lot easier, code clearer and doesn't rely on dynamic dispatch as here (need to wrap things into `Val` to avoid that here).

src/layers/upsample.jl
CarloLucibello: NNlib's doesn't support integer scale yet

test/cuda/layers.jl
CarloLucibello: upsampling_nearest doesn't support non-integer scale. The extension is non-trivial and shouldn't block merging

src/layers/upsample.jl
CarloLucibello: This implementation is using multiple dispatch, which is fine as always has been. Also, passing as mode the NNlib functions is a bad idea, more verbose and kinda gives the impression that you can pass arbitrary functions, while `mode` is essentially an enum, which is naturally mapped into a symbol. 
Finally, I don't understand what's the point of removing type annotation (Symbol?) and default mode (for what, for mode or for the keyword args?). We should provide a default for mode, for people who don't want to delve into the details of the different modes and use the layer brainlessly

test/cuda/layers.jl
CarloLucibello: this was meant

src/layers/upsample.jl
DhairyaLGandhi: Well you would be able to support the current behavior and default just fine with having mode be a function, including giving the option of not having to type it. and you can pass more functions to it if you're handling them otherwise. Different implementations like trilinear etc are possible, and the names of the functions can themselves act as the symbols, so no difference if you're changing the mode anyway.

CarloLucibello: or ... you can just use symbols, and type Upsample(:bilinear) instead of Upsample(upsample_bilinear)

maxfreu: Well, in any case one can use the function directly and circumvent all this.

src/layers/upsample.jl
```diff
@@ -0,0 +1,82 @@
...
+
+julia> m(ones(2, 2, 1, 1)) |> size
+(4, 5, 1, 1)
+"""
+struct Upsample{Mode,S,T}
+  scale::S
+  size::T
+end
+
+function Upsample(mode::Symbol=:nearest; scale=nothing, size=nothing)
```
maxfreu, line 39: Sorry to be nagging again; I just would love to see `Upsample(2)` possible :)

CarloLucibello: I think we should have that, but I'm a bit tired of revisiting this pr over and over :sweat_smile: . Adding that convenience constructor is not a problem at all and could be done in a little separate pr

CarloLucibello:
this is ready, I need an approval for merging

src/layers/upsample.jl
maxfreu: I know that feel, bro. Just give it a go then :)

src/layers/upsample.jl
DhairyaLGandhi: Well symbols are opaque identifiers and don't lend themselves to transparently telling users what is happening underneath. I agree the naming of those functions is dirty too, and that should be resolved later.

src/layers/upsample.jl
```diff
@@ -0,0 +1,82 @@
...
+function (m::Upsample{:nearest, Int})(x::AbstractArray{T, N}) where {T, N} 
+  NNlib.upsample_nearest(x, ntuple(i -> m.scale, N-2))
+end
+(m::Upsample{:nearest, Nothing})(x::AbstractArray) =
+  NNlib.upsample_nearest(x; size=m.size)
+
+(m::Upsample{:bilinear})(x::AbstractArray) =
+  NNlib.upsample_bilinear(x, m.scale)
+(m::Upsample{:bilinear, Nothing})(x::AbstractArray) = 
+  NNlib.upsample_bilinear(x; size=m.size)
```
DhairyaLGandhi, line 59:
```suggestion
(m::Upsample{Val{:nearest}()})(x) = NNlib.upsample_nearest(x, m.scale)
(m::Upsample{Val{:nearest}(), Nothing})(x) = NNlib.upsample_nearest(x, size = m.size)

function (m::Upsample{Val{:nearest}(), Int})(x::AbstractArray{T, N}) where {T, N} 
  NNlib.upsample_nearest(x, ntuple(i -> m.scale, N-2))
end

(m::Upsample{Val{:bilinear}()})(x) = NNlib.upsample_bilinear(x, m.scale)
(m::Upsample{Val{:bilinear}(), Nothing})(x) = NNlib.upsample_bilinear(x, size = m.size)
```

src/layers/upsample.jl
```diff
@@ -0,0 +1,82 @@
...
+
+function Base.show(io::IO, u::Upsample{mode}) where {mode}
+  print(io, "Upsample(")
+  print(io, ":", mode)
+  u.scale !== nothing && print(io, ", scale=$(u.scale)")
+  u.size !== nothing && print(io, ", size=$(u.size)")
```
DhairyaLGandhi, line 65:
```suggestion
  u.scale !== nothing && print(io, ", scale = $(u.scale)")
  u.size !== nothing && print(io, ", size = $(u.size)")
```

DhairyaLGandhi (commented):
Just some last minute cleanups.

DhairyaLGandhi (dismissed):
I think we can use the `Val` moniker here, to avoid dynamic dispatch but its not the end of the world either, it shouldn't matter to runtime performance.

DhairyaLGandhi:
I think if there is enough consensus and need, we should also add a `Upsample(scale, mode)` option, but that can happen later.

src/layers/upsample.jl
```diff
@@ -0,0 +1,82 @@
...
+  size::T
+end
+
+function Upsample(mode::Symbol=:nearest; scale=nothing, size=nothing)
+  mode in [:nearest, :bilinear] || 
+    throw(ArgumentError("mode=:$mode is not supported."))
+  if ~((scale === nothing) ⊻ (size === nothing))
+    throw(ArgumentError("Either scale or size should be specified."))
+  end
+  return Upsample{mode,typeof(scale),typeof(size)}(scale, size)
```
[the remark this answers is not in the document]
CarloLucibello: this suggestion is not correct, should act as a xnor

src/layers/upsample.jl
CarloLucibello: I don't think this is needed (I'm not even sure it's correct). A few iterations ago I was using Val in the parameterization, then @mcabbott observe is not needed, looks like you can symbols and string in parametrizing, same as you do with integers (maybe any isbits immutable? )

DhairyaLGandhi: You can parametrize with just about anything (notice DenseConvDims for eg), the reason for Val here is that its guaranteed to be that value, and you don't have to do that checking during runtime. This should be correct, you could check with something like print statements that it forwards properly.

src/layers/upsample.jl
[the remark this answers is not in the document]
DhairyaLGandhi: It has the same behaviour

```Julia
julia> ~((1 === nothing) ⊻ ((3,4) === nothing))
true
```

CarloLucibello:
> I think we can use the Val moniker here, to avoid dynamic dispatch but its not the end of the world either, it shouldn't matter to runtime performance.

I think it's equivalent and dispatch is the same, so we should use just symbols since they are nicer,  but honestly I don't know very well about this stuff, maybe @mcabbott can chime in

src/layers/upsample.jl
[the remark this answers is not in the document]
DhairyaLGandhi: Either way, prefer mentioning explicitly that we cannot specify both arguments.

DhairyaLGandhi:
They aren't the same, unless Julia wraps it in a Val internally.

DhairyaLGandhi:
I don't think it's complicating the show function, since we can definitely show `typeof(fun)` which is pretty clear. Add to that the show function is basically sugar here, and there aren't many parameters, so I'd much rather take the clearer code.

DhairyaLGandhi:
> It's known when the struct is constructed, and might simplify the code as then you only handle the symbol once (on construction), and on calling you could do something like (u::Upsample)(x) = u.fun(x, u.scale) the same for any function. 

+1 for this

src/layers/upsample.jl
```diff
@@ -0,0 +1,81 @@
...
+
+An upsampling layer. 
+
+`scale` is a number or a tuple of numbers 
+representing  the output rescaling factor along each spatial dimension.
+For integer `scale`, all but the last 2 dimensions (channel and batch)
+will be rescaled by the same factor. 
+
+It is also possible to directly specify the output spatial `size`,
+as an alternative to using `scale`.
```
mcabbott, line 12: The `nothing`s here are an implementation detail, 
```suggestion
  Upsample(mode = :nearest; [scale, size])  

An upsampling layer. One of two keywords must be given:

If `scale` is a number, this applies to all but the last two dimensions (channel and batch) of the input. 
It may also be a tuple, to control dimensions individually. Alternatively, keyword 
`size` accepts a tuple, to directly specify the leading dimensions of the output.
```

DhairyaLGandhi: Maybe we should also link to what the upsampling layer does.

src/layers/upsample.jl
darsnack: I suggested the functions above, and I ended up changing my mind. You can't pass arbitrary functions into this layer (it's not higher order in that sense). There is no way to check if the function passed is valid (e.g. if someone passed `log` it makes no sense). Allowing functions is clean, but it is not beginner friendly IMO.

This is just a bad situation for APIs. There is no good answer but (1) is pretty decent.

src/layers/upsample.jl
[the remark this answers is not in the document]
darsnack: It's not the same behavior. The suggestion doesn't work for the case where both `scale` and `size` are not `nothing`. The original code checks that only one argument is `nothing`.

Though the original code should use `isnothing` instead of `===`. And the error can be more informative.
```julia
if !(isnothing(scale) ⊻ isnothing(size))
  throw(ArgumentError("Either scale or size should be specified (but not both)."))
end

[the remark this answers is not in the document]
Comment (DhairyaLGandhi) [reply src/layers/upsample.jl:45]: Yup I'd be fine with this. Cc @CarloLucibello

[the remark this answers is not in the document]
Comment (DhairyaLGandhi) [reply src/layers/upsample.jl:45]: Ideally we wouldn't use unicode if we can avoid it, but seems cleaner here

File: src/layers/upsample.jl (hunk 18 above)
Comment (darsnack) [reply src/layers/upsample.jl:12]: You mean to the NNlib functions? Those are already linked below.

Comment (DhairyaLGandhi) [reply src/layers/upsample.jl:12]: No, I mean describing what upsampling is. The nnlib functions are implementations of specific interpolation routines.

Link to super resolution gan or reducing the memory bandwidth of a network.

**Comment by DhairyaLGandhi** [comment]
> which includes printing something copy-pastable.

Right but if we use the function there, there's presumably an existing constructor so I think thats doable.

Ultimately I feel relying on Interpolations.jl is the better longer term approach, rather than redoing work, so hopefully we can bring that into our ecosystem.

**Comment by darsnack** [comment]
Yeah +1 for storing the interpolating function. It simplifies the code significantly, and I think any issues with show can be resolved since these aren't arbitrary functions. The API should still be symbol based.

**Comment by CarloLucibello** [comment]
I'll change the implementation to store internally the function. For the record, I think this change is just exploring the "who cares?" design manifold, adding zero value to users and to us developers and wasting my time, but I'll soldier on and do it

**Review by mcabbott — APPROVED** [review]
Despite having suggested that, I agree it's not a big deal either way. And doesn't change the user interface, so could be done later by anyone who does care.

**Comment by CarloLucibello** [comment]
Also, is not clear how I should change the show method
```julia
function Base.show(io::IO, u::Upsample{mode}) where {mode}
  print(io, "Upsample(")
  print(io, ":", mode)
  u.scale !== nothing && print(io, ", scale = $(u.scale)")
  u.size !== nothing && print(io, ", size = $(u.size)")
  println(io, ")")
end
```
If `mode` now is the function type is the function type instead of the symbol, when printing I should convert it back to the symbol, which is quite ugly

darsnack:
You could do:
```julia
function Base.show(io::IO, u::Upsample)
  print(io, "Upsample(")
  print(io, ":", nameof(u.mode))
  u.scale !== nothing && print(io, ", scale = $(u.scale)")
  u.size !== nothing && print(io, ", size = $(u.size)")
  println(io, ")")
end
```

I also think that this change can easily be done in a separate PR. It doesn't change the desired API in any way. I'd be fine merging as is.

DhairyaLGandhi:
It's alright to show `typeof(func)` but I agree you don't have to do it if you don't care to do it.

I've already approved too

DhairyaLGandhi (approved):

DhairyaLGandhi:
I'm slightly confused, are we adding the positional scale constructor here already?

CarloLucibello:
`typeof(func)` or related would yield `upsample_nearest`, but we want `:nearest` here to obtain a string that represents valid code.  

> I'm slightly confused, are we adding the positional scale constructor here already?

yes I added it since I add to add a few changes from comments in any case

CarloLucibello:
need another approval

src/layers/upsample.jl
```diff
@@ -0,0 +1,79 @@
...
+  NNlib.upsample_bilinear(x, m.scale)
+(m::Upsample{:bilinear, Nothing})(x::AbstractArray) = 
+  NNlib.upsample_bilinear(x; size=m.size)
+
+function Base.show(io::IO, u::Upsample{mode}) where {mode}
+  print(io, "Upsample(")
+  print(io, ":", mode)
+  u.scale !== nothing && print(io, ", scale = $(u.scale)")
+  u.size !== nothing && print(io, ", size = $(u.size)")
+  println(io, ")")
```
mcabbott, line 65:
```suggestion
  print(io, ")")
```

mcabbott (approved):

DhairyaLGandhi (approved):

PR #1469, opened 2021-01-18
One-arg unsqueeze method

mcabbott:
This makes `unsqeeze(3)` return a function.

In fact a `Base.Fix2` so that it has a predictable type, although overloading `show` turns out to be a pain for `<:Function` (perhaps it would be simpler to give it its own `struct`?), right now this is pretty within `Chain` but not when used by itself.

The only tests of the existing method seem to be doctests, so this adds a few more. And cross-links to `flatten`.

Plus a test for `outputsize`, now.

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -1,9 +1,26 @@
 """
     flatten(x::AbstractArray)
 
-Reshape arbitrarly-shaped input into a matrix-shaped output
-preserving the last dimension size.
-Equivalent to `reshape(x, :, size(x)[end])`.
+Reshape arbitrarly-shaped input into a matrix-shaped output,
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi:
We would typically expect users to write functions of this form

mcabbott:
Yes exactly.

CarloLucibello:
unrelated datasets test are falling, It looks like the download links on cache.julialang.org are not reliable

CarloLucibello:
LGTM. 
We should also support multiple dims with `unsqueeze` at some point. Maybe also move to a keyword arg `dims`,
an unsqueeze/unflatten layer would read `unesqueeze(dims=(2,3))`.

CarloLucibello (approved):

PR #1470, opened 2021-01-19
remove dataset tests

CarloLucibello:
since the deprecation warnings are annoying, but most importantly because the download links have become unreliable and occasionally fail (as in #1469).


CarloLucibello:
@mcabbott let's test if your role has been updated

DhairyaLGandhi (approved):

Issue #1350, opened 2020-10-08
ConvTranspose on GPU fails with certain activation functions


mossr:
Running `ConvTranspose` with certain activation functions other than the `identity` will seg. fault (here's an example with `tanh`):

```julia
using Flux, CUDA
x = rand(1,1,1,1) |> gpu
m = ConvTranspose((1,1), 1=>1, tanh) |> gpu
m(x)
```
> FATAL ERROR: Symbol "__nv_tanhf"not found
signal (22): SIGABRT
...

Reading through https://github.com/JuliaGPU/CUDAnative.jl/issues/632 tells me that the activation function is not being run on the GPU. I tried a few activation functions that I know are part of CUDA (`tanh`, `exp`, `tanhshrink`, `softplus`) and those all failed, but other activation function like `σ` and `relu` worked 🤷‍♂️ (`σ` is odd to me because `exp` and `softplus` fail due to "__nv_expf" not found).

If I run without the `tanh` activation (which defaults to `indentity`) this works. I also tried `Conv` and it *does* work with other activation functions—thus this might be an issue with `conv_transpose_dims` or `∇conv_data`, but thought I'd get advice before digging any deeper.

This also may be related to https://github.com/JuliaGPU/CUDA.jl/issues/228 if I followed that correctly.

Lastly, `ConvTranspose` does not have any explicit activation functions passed in the `test/conv.jl` tests, so this may have gone under the radar.

arvindmohan:
I ran into exactly the same issue - glad to see its not an isolated problem. Have you managed a workaround thus far?

mossr:
@arvindmohan I haven't found a workaround yet. But it's good to see this issue is gaining *some* traction.

arvindmohan:
Yes, I'm a new user and its surprising no one else has mentioned this (considering how popular autoencoders are). I may be willing to take a stab at fixing it if I get some guidance...

CarloLucibello:
This is a puzzling error, I cannot get my head around it. I did the following experiments:
```julia
using  Flux


#  no-bias no-act 
function forw0(c::ConvTranspose, x::AbstractArray)
         # ndims(x) == ndims(c.weight)-1 && return squeezebatch(c(reshape(x, size(x)..., 1)))
         σ, b = c.σ, reshape(c.bias, map(_->1, c.stride)..., :, 1)
         cdims = Flux.conv_transpose_dims(c, x)
         ∇conv_data(x, c.weight, cdims)             
 end

[...]

PR #1472, opened 2021-01-22
Add activation tests for GPU layers

DhairyaLGandhi:
This is in response to #1350, we should be routinely testing our layers with multiple activation functions to catch these errors early.


DhairyaLGandhi:
Why are so many RNN tests marked broken @CarloLucibello? Was that introduced in https://github.com/FluxML/Flux.jl/pull/1367

CarloLucibello:
> Was that introduced in #1367

Is this a question? If yes, I don't know,  #1367 fixed many issues and actually unbroke one test

I don't remember when tests were marked as broken, could try with git blame

DhairyaLGandhi:
Well this pr does show up some disagreements between cpu and GPU implementations which shouldn't be.

#1367 was also a major overhaul and to the best of my knowledge isn't accurate in its statement of beating cudnn. With @denizyuret cudnn PR in CUDA we would be best served to benchmark the two and use the cudnn api properly.

DhairyaLGandhi:
Test:

@ModelZookeeper commands

DhairyaLGandhi:
Here are the commands: build feed commands

darsnack:
Does the bot only respond to the person who asked? I don't see any comments from the bot at all.

DhairyaLGandhi:
The one I emoji'd on was made by the bot. I have to generate some new tokens so its not using my account for any of this.

darsnack:
Ah cool!!

DhairyaLGandhi:
Test 2:

@ModelZookeeper commands

ModelZookeeper:
Here are the commands: build feed commands

CarloLucibello:
@DhairyaLGandhi I wanted to review this, could you ask for reviews before merging PRs?

test/cuda/layers.jl
```diff
@@ -13,71 +13,98 @@ end
...
+  r = rand(Float32, 28, 28, 1, 1)
+  conv_layers = [Conv, ConvNoBias,
+                 ConvTranspose, ConvTransposeNoBias,
+                 CrossCor, CrossCorNoBias,
+                 DepthwiseConv, DepthwiseConvNoBias]
+  gpu_gradtest("Convolution with $act", conv_layers, r, (2,2), 1=>3, act, test_cpu = false)
+  
+  batch_norm = [BatchNorm]
+  gpu_gradtest("BatchNorm 1 with $act", batch_norm, rand(Float32, 28,28,3,4), 3, act, test_cpu = false) #TODO fix errors
+  gpu_gradtest("BatchNorm 2 with $act", batch_norm, rand(Float32, 5,4), 5, act, test_cpu = false)
```
CarloLucibello, line 85: test_cpu used to be `true, why are all false now? I think that checking that results are the same on cpu and gpu is very important

test/cuda/layers.jl
```diff
@@ -149,36 +197,28 @@ end
...
-  m_cpu(x_cpu)
-  gradient(() -> sum(m_cpu(x_cpu)), Flux.params(m_cpu))
-  m_gpu(x_gpu)
-  gradient(() -> sum(m_gpu(x_gpu)), Flux.params(m_gpu))
+  # x_cpu = rand(Float64, 3, 2, 2)
+  # x_gpu = x_cpu |> gpu
+  # m_cpu(x_cpu)
+  # gradient(() -> sum(m_cpu(x_cpu)), Flux.params(m_cpu))
+  # m_gpu(x_gpu)
+  # gradient(() -> sum(m_gpu(x_gpu)), Flux.params(m_gpu))
```
CarloLucibello, line 205: Why this have been commented out?

DhairyaLGandhi:
I have been waiting for any objections on this for a while actually. Pretty sure I wanted this in either way.

test/cuda/layers.jl
```diff
@@ -13,71 +13,98 @@ end
...
+                 CrossCor, CrossCorNoBias,
+                 DepthwiseConv, DepthwiseConvNoBias]
+  gpu_gradtest("Convolution with $act", conv_layers, r, (2,2), 1=>3, act, test_cpu = false)
+  
+  batch_norm = [BatchNorm]
+  gpu_gradtest("BatchNorm 1 with $act", batch_norm, rand(Float32, 28,28,3,4), 3, act, test_cpu = false) #TODO fix errors
+  gpu_gradtest("BatchNorm 2 with $act", batch_norm, rand(Float32, 5,4), 5, act, test_cpu = false)
+  
+  instancenorm = [InstanceNorm]
+  gpu_gradtest("InstanceNorm with $act", instancenorm, r, 1, act, test_cpu = false)
```
CarloLucibello, line 88: some of these layers were checked in both testmode and trainmode, not true anymore

CarloLucibello:
> I have been waiting for any objections on this for a while actually. Pretty sure I wanted this in either way.

There is a button on the top right corner where you can request a review, or you can just ping someone as I do with all the time. 

I'm pretty sure we don't want to remove a lot of test coverage as this PR does.

test/cuda/layers.jl
DhairyaLGandhi: This pr doesn't change the src 

There's a comment explaining this.

test/cuda/layers.jl
DhairyaLGandhi: But those are tested in the specific tests for the specific layers in normalisation.jl. if you want to add those back, I can add a follow on. It would be easier to review those separately anyway

CarloLucibello:
Since you have a lot of stuff lying around incomplete, it's hard to tell what is ready for review. So please do ask explicitly for some feedback

test/cuda/layers.jl
CarloLucibello: This should be tested here as well since we currently have bugs like #1542 that are gpu only

test/cuda/layers.jl
CarloLucibello: how does setting `test_cpu=false` when it used to be `true` doesn't change things?
I don't know which comment you refer to.
Also,  there are some `#TODO fix errors`, why?

PR #1473, opened 2021-01-22
Fix RNN tests on GPU

jeremiedb:
Fix for RNN on CUDA, as discussed in #1367 .


jeremiedb:
@DhairyaLGandhi Is it good for review?

DhairyaLGandhi:
I'm slightly confused by this, if the tests were passing all this while, then why did we not see the `unexpected pass ...` , and if they were indeed working we wouldn't have marked them broken.

bors try

jeremiedb:
They were not previously passing given the `m̄[].cell[].Wi` was to be changed to `m̄[].cell.Wi`.

DhairyaLGandhi:
Ah cool, sorry I missed that

DhairyaLGandhi:
bors try-

PR #1481, opened 2021-01-27
Use Optimisers.jl

### LINKED ISSUE #637: New New Optimisers

### LINKED ISSUE #823: Flux Optimizers should define equality
I was writing some tests, and I wanted to be sure that the hyper-parameters of two models were the same.
So I was checking equality on an object that contained all my hyperparameters,
including the optimizer (which has its own hyper-parameters).

Turns out that fails as
```
julia> Flux.Optimise.NADAM(0.003, (0.9, 0.9), IdDict{Any,Any}()) == Flux.Optimise.NADAM(0.003, (0.9, 0.9), IdDict{Any,Any}())
false
```

DhairyaLGandhi:
Fixes #637, fixes #823.


DhairyaLGandhi:
@ModelZookeeper commands

src/optimise/train.jl
```diff
@@ -116,6 +117,43 @@ function train!(loss, ps, data, opt; cb = () -> ())
   end
 end
 
+function train!(m, loss, data, opt; cb = (x...) -> ()
+                                    prehook = (x...) -> (),
+                                    posthook = (x...) -> ())
+  st = [Optimisers.init(opt, p) for p in Flux.params(m)]
```
ToucheSir, line 123: I think this can be reduced to `st = Optimisers.state(opt, m)`.

darsnack: I think this should be defined as a function in the `optimiser.jl` file. One for `state(opt, p::Params)` and another for `update(opt, x::Params, dx, state)`.

darsnack:
I would prefer we remove the hooks from `Flux.train!` in this PR and save it for something else. This PR can simply remove the optimizers in favor of Optimisers.jl.

DhairyaLGandhi:
Done a bunch of cleanup, including removing the train loop changes, but I feel like we would circle back to it soon

src/optimise/train.jl
```diff
@@ -97,12 +102,13 @@ Multiple optimisers and callbacks can be passed to `opt` and `cb` as arrays.
 function train!(loss, ps, data, opt; cb = () -> ())
   ps = Params(ps)
   cb = runall(cb)
+  st = [Optimisers.init(opt, p) for p in ps]
```
darsnack, line 105: Could we define `init` for `Params` like we do `update`?

DhairyaLGandhi: I'm thinking if this should be wholesale replaced with `Optimisers.state` instead

DhairyaLGandhi: I'm only worried about corner cases that this might hit

darsnack: I think `state` makes sense. For every `p` in `ps`, we should be able to independently optimize each one by calling the user facing `state` and `update`. I think its safe to do that here.

DhairyaLGandhi: Right, so that works in cases where we have arrays as params, but not arbitrary structs, which is what we want.

The user facing API is something like (pseudocode; although it might work in this MWE)
```julia
m = Dense(3,3)
opt = ADAM()
st = Optimisers.state(opt, m) # `m` could contain arbitrary structs which shouldn't be functor'd
loss(m, x, y) = Flux.mse(m(x), y)
for i = 1:1000
  gs, = gradient(m) do m
    @show loss(m, w, w′)
  end
  m, st = opt(m, gs, st)
end
```

darsnack: That's exactly what I'm suggesting. To be more specific, change the code to:
```suggestion
  st = state(opt, ps)
```

And elsewhere, define:
```julia
init(o, ps::Params) = [init(o, p) for p in ps]
```

Should allow for the future case where `m` is not a `Params` but also the current case where we need to support `Params`.

docs/src/training/optimisers.md
```diff
@@ -97,9 +98,9 @@ v = ρ * v - η * Δ
 w = w - v
 ```
 
-The `apply!` defines the update rules for an optimiser `opt`, given the parameters and gradients. It returns the updated gradients. Here, every parameter `x` is retrieved from the running state `v` and subsequently updates the state of the optimiser.
+The `apply` defines the update rules for an optimiser `opt`, given the parameters and gradients. It returns the updated gradients. Here, every parameter `x` is retrieved from the running state `v` and returns the new state of the optimizer.
```
darsnack, line 101:
```suggestion
The `apply` defines the update rules for an optimiser `opt`, given the parameters and gradients. It returns the updated gradients and optimizer state.
```

src/optimise/train.jl
```diff
@@ -19,15 +19,20 @@ according to optimizer `opt`  and the gradients `gs` (the gradient `g`).
 
 As a result, the parameters are mutated and the optimizer's internal state may change.
 """
-function update!(opt, x, x̄)
-  x .-= apply!(opt, x, x̄)
+function update!(opt, x, x̄, st)
+  x̄, st = apply(opt, x, x̄, st)
+  x .-= x̄
```
darsnack, line 24: Should we use `Optimisers.patch` here for consistency?

src/optimise/train.jl
```diff
@@ -19,15 +19,20 @@ according to optimizer `opt`  and the gradients `gs` (the gradient `g`).
...
-  for x in xs
+function update!(opt, xs::Params, gs, st)
+  st_ = []
+  for (x,s) in zip(xs,st)
     gs[x] == nothing && continue
-    update!(opt, x, gs[x])
+    _, s = update!(opt, x, gs[x], s)
+    append!(st_, s)
   end
+  nothing, st_
```
darsnack, line 35:
```suggestion
  xs, st_
```

src/optimise/train.jl
```diff
@@ -97,12 +104,13 @@ Multiple optimisers and callbacks can be passed to `opt` and `cb` as arrays.
 function train!(loss, ps, data, opt; cb = () -> ())
   ps = Params(ps)
   cb = runall(cb)
+  st = Optimisers.init(opt, ps)
```
darsnack, line 107:
```suggestion
  st = Optimisers.state(opt, ps)
```

darsnack (commented):
Looking pretty good. Two ML calls ago, we talked about having `apply!` also defined in Optimisers.jl as a temporary measure to avoid potential performance regressions. Is that still the plan?

DhairyaLGandhi:
See https://github.com/FluxML/Optimisers.jl/pull/13

We would also want #1402 i think when walking through the tests

darsnack:
Cool, looks good. Should we be calling `Optimisers.update` here then? Calling `apply` directly bypasses the mutability check.

darsnack:
Ref https://github.com/FluxML/Flux.jl/pull/1613#issuecomment-863274729

In my mind, this PR is basically good to go. What I would suggest is establishing a `optimizers.jl` branch off `master` (similar to how we had a `zygote` branch for that transition). Then we can safely merge this into the `optimisers.jl` branch without much concern, and everyone can start the process of benchmarking/validating the transition.

DhairyaLGandhi:
That branch exists already - it's this pr! I think we're kind of tied to merge the zeros first, then this and boom.

darsnack:
Why does this depend on `Zeros`?

If this is _the_ branch, then I guess let's rebase it?

DhairyaLGandhi:
We can rebase. Without the zeros, state initialisation and optimisation doesn't work. In order to define methods in init, one would have to do it manually.

darsnack:
I'm not sure what you mean. Why is this not enough
```julia
Optimisers.init(o, ps::Params) = [Optimisers.state(o, p) for p in ps]
```

DhairyaLGandhi:
We also have a route that does not require the use of `Params`.

darsnack:
You mean where we allow passing in `[W, b]`? Wouldn't that route be handled by Optimisers.jl already?

CarloLucibello:

PR #1484, opened 2021-01-28
add FluxBot

DhairyaLGandhi:
Companion PR to https://github.com/FluxML/model-zoo/pull/278

This would allow us to test changes here taking effect in the actual model implementations. We should also add Metalhead to downstream testing.

### DIFF
--- a/.github/workflows/fluxbot.yml
+++ b/.github/workflows/fluxbot.yml
@@ -1,0 +1,28 @@
+name: FluxBot
+
+on:
+  issue_comment:
+    types: [created, edited]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        julia-version: [1.5.0]
+        julia-arch: [x86]
+        os: [ubuntu-latest]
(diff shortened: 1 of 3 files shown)


DhairyaLGandhi:
@ModelZookeeper commands

.github/workflows/fluxbot.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.julia-version }}
+      - name: Install dependencies
+        run: julia --project=.fluxbot/ -e 'using Pkg; Pkg.instantiate()'
+      - name: FluxBot.respond
+        env:
+          FLUXBOT_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BOT_SECRET: ${{ secrets.BOT_SECRET }}
+          MODELZOO_TOKEN: ${{ secrets.MODELZOO_TOKEN }}
```
DhairyaLGandhi, line 26:
```suggestion
```

PR #1486, opened 2021-01-28
Make `outputsize` understand multiple inputs

### LINKED ISSUE #1466: Better handling for layers with multiple inputs w/ outputsize
Given the variable argument version of `Parallel` (#1462) and #1009, it seems like we need better support for multiple arguments as layer inputs. Currently, `outputsize` only understands a single `Tuple` for `inputsize`.

mcabbott:
Closes #1466

### DIFF
--- a/src/outputsize.jl
+++ b/src/outputsize.jl
@@ -52,11 +52,12 @@ using .NilNumber: Nil, nil
 
 Calculate the size of the output from model `m`, given the size of the input.
 Obeys `outputsize(m, size(x)) == size(m(x))` for valid input `x`.
+
 Keyword `padbatch=true` is equivalent to using `(inputsize..., 1)`, and
 returns the final size including this extra batch dimension.
 
 This should be faster than calling `size(m(x))`. It uses a trivial number type,
-and thus should work out of the box for custom layers.
+which should work out of the box for custom layers.
 
 If `m` is a `Tuple` or `Vector`, its elements are applied in sequence, like `Chain(m...)`.
 
@@ -83,7 +84,7 @@ julia> try outputsize(m, (10, 10, 7, 64)) catch e println(e) end
 └ @ Flux ~/.julia/dev/Flux/src/outputsize.jl:114
 DimensionMismatch("Input channels must match! (7 vs. 3)")
 
-julia> outputsize([Dense(10, 4), Dense(4, 2)], (10, 1))
+julia> outputsize([Dense(10, 4), Dense(4, 2)], (10, 1)) # Vector of layers becomes a Chain
 (2, 1)
 
 julia> using LinearAlgebra: norm
@@ -145,7 +146,7 @@ true
 ```
 Notice that `Chain` only accepts multiple arrays as a tuple,
 while `Parallel` also accepts them as multiple arguments;
-`outsize` always supplies the tuple.
+`outputsize` always supplies the tuple.
 """
 outputsize
 
--- a/src/outputsize.jl
+++ b/src/outputsize.jl
@@ -86,16 +86,6 @@ DimensionMismatch("Input channels must match! (7 vs. 3)")
 
 julia> outputsize([Dense(10, 4), Dense(4, 2)], (10, 1)) # Vector of layers becomes a Chain
 (2, 1)
-
-julia> using LinearAlgebra: norm
-
-julia> f(x) = x ./ norm.(eachcol(x));
-
-julia> outputsize(f, (10, 1)) # manually specify batch size as 1
-(10, 1)
-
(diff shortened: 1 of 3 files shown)


src/outputsize.jl
```diff
@@ -79,32 +80,70 @@ julia> outputsize(m, (10, 10, 3, 64))
...
+For model or layer `m` accepting multiple arrays as input,
+this returns `size(m((x, y, ...)))` given `size_x = size(x)`, etc.
+
+# Examples
+```jldoctest
+julia> x, y = rand(Float32, 5, 64), rand(Float32, 7, 64);
+
+julia> par = Parallel(vcat, Dense(5, 9), Dense(7, 11));
 
-julia> f(x) = x ./ norm.(eachcol(x));
```
mcabbott, line 89: I removed this norm example BTW, because I thought things were getting a bit long, and because it doesn't actually work with a batch dimension:
```
julia> f(rand(10,1)) |> size
(10, 1)

julia> f(rand(10,2)) |> size
ERROR: DimensionMismatch("arrays could not be broadcast to a common size; got a dimension with lengths 10 and 2")
```

darsnack (dismissed):
No complaints from me. I like the more informative error message. We'll have to make sure it stays in sync with the actual implementation of `Chain`, but I think it is worth it.

mcabbott:
Great. As long as it still seems fast enough for your applications, that's one possible concern. It is only the outermost Chain, with no attempt to recurse into try/catch-ing sub-sub-layers.

darsnack:
Good point, I'll double check the speed like we did on the original PR. But I think for its use case, a minor slowdown is fine for convenient errors. It doesn't need to be super fast, just faster than evaluating ResNet-101 (which this should be).

darsnack (approved):
Sorry for the delay. Here are the benchmarks on a large model:
```julia
julia> m = densenet_121();

# this PR
julia> @benchmark Flux.outputsize($m, (256, 256, 3, 50))
BenchmarkTools.Trial:
  memory estimate:  201.89 KiB
  allocs estimate:  4546
  --------------
  minimum time:     402.313 ms (0.00% GC)
  median time:      410.185 ms (0.00% GC)
  mean time:        423.873 ms (0.00% GC)
  maximum time:     480.239 ms (0.00% GC)
  --------------
  samples:          12
  evals/sample:     1

# Flux#master
julia> @benchmark Flux.outputsize($m, (256, 256, 3, 50))
BenchmarkTools.Trial:
  memory estimate:  278.48 KiB
  allocs estimate:  6634
  --------------
  minimum time:     416.605 ms (0.00% GC)
  median time:      417.849 ms (0.00% GC)
  mean time:        419.333 ms (0.00% GC)
  maximum time:     431.575 ms (0.00% GC)
  --------------
  samples:          12
  evals/sample:     1
```

I think based on that, it's good to merge.

darsnack:
Think we still need someone with merge rights like @CarloLucibello

mcabbott:
Excellent, thanks for checking. 

I'm not super-sure how rights work, or whether merging is blocked to anyone not called Bors?

darsnack:
Yeah I had the same confusion. All I know is Github says I don't have merge rights, but I don't know if I'm allowed to call `bors` or not.

CarloLucibello:
I now see "1 approving review by reviewers with write access", so Kyle's rights have been fixed and hopefully he should also be able to call bors. want to try calling bors yourself @mcabbott ?

CarloLucibello:
there you go

darsnack:
Thanks Carlo!

CarloLucibello:
yeah merging is blocked also for me here, can go only through bors

darsnack:
And my reading of ColPrac says that since Michael has write access only he is allowed to call `bors` after approval from someone else?

CarloLucibello:
> And my reading of ColPrac says that since Michael has write access only he is allowed to call bors after approval from someone else? (since he is the PR author)

yes, and this is enforced by current github's settings. Those settings are actually stricter than ColPrac since for colprac also an approval from someone without write access is enough

mcabbott:
OK, great. The direct merge button is still disabled for me, so I guess the enforcement is that Bors would not have listened to me, before the approval?

CarloLucibello:
> OK, great. The direct merge button is still disabled for me, so I guess the enforcement is that Bors would not have listened to me, before the approval?

that or it does its run then fails to merge, can't remember which one

mcabbott:
If I'm reading the rules correctly, I may also merge PRs by other people, without another member's approval?

Such as #1009, which just celebrated its first birthday...

CarloLucibello:
yes. approve then call bors

mcabbott:
Ok, I tried. It says I need to be made a reviewer?

PR #1488, opened 2021-01-29
Add training loop to docs

DhairyaLGandhi:
This also links to the correct point in the docs to read about custom training loop behaviour.

### DIFF
--- a/docs/src/training/training.md
+++ b/docs/src/training/training.md
@@ -7,13 +7,36 @@ To actually train a model we need four things:
 * A collection of data points that will be provided to the objective function.
 * An [optimiser](optimisers.md) that will update the model parameters appropriately.
 
-With these we can call `train!`:
+Training a model is typically an iterative process, where we go over the data set,
+calculate the objective function over the datapoints, and optimise that.
+This can be visualised in the form of a simple loop.
+
+```julia
+for d in datapoints
+
+  # `d` should produce a collection of arguments
+  # to the loss function
+
+  # Calculate the gradients of the parameters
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
cc @ToucheSir

ToucheSir (approved):
Looks great, thanks for adding this!

docs/src/training/training.md
```diff
@@ -7,13 +7,36 @@ To actually train a model we need four things:
...
+```
+
+To make it easy, Flux defines `train!`:
 
 ```@docs
 Flux.Optimise.train!
 ```
 
-There are plenty of examples in the [model zoo](https://github.com/FluxML/model-zoo).
+There are plenty of examples in the [model zoo](https://github.com/FluxML/model-zoo), and
```
darsnack, line 38:
```suggestion
Like the loop above, `train!` expects the data to be [iterable](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration). Most default collections Julia, like `Array`s, are iterable.

There are plenty more examples in the [model zoo](https://github.com/FluxML/model-zoo), and
```

DhairyaLGandhi: We already mention and demonstrate the iteration twice in the couple lines above, do we need it again here?

darsnack: People unfamiliar to Julia may not realize that the syntax `for x in xs` requires the iteration interface. The suggestion was made to make understanding the resulting method error easier for new users.

PR #1489, opened 2021-01-30
Implementation of Focal loss

shikhargoswami:
Focal loss was introduced in the RetinaNet paper (https://arxiv.org/pdf/1708.02002.pdf). 

Focal loss is useful for classification when you we highly imbalanced classes. It down-weights well-classified examples and focuses on hard examples. The loss value is much high for a sample which is misclassified by the classifier as compared to the loss value corresponding to a well-classified example. 

Used in single-shot object detection where the imbalance between the background class and other classes is extremely high.

Here's it's tensorflow implementation (https://github.com/tensorflow/addons/blob/v0.12.0/tensorflow_addons/losses/focal_loss.py#L26-L81)
### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/src/losses/Losses.jl
+++ b/src/losses/Losses.jl
@@ -19,8 +19,7 @@ export mse, mae, msle,
     poisson_loss,
     hinge_loss, squared_hinge_loss,
     ctc_loss,
-    binary_focal_loss,
-    categorical_focal_loss
+    binary_focal_loss, focal_loss
 
 include("utils.jl")
 include("functions.jl")
--- a/src/losses/Losses.jl
+++ b/src/losses/Losses.jl
@@ -19,7 +19,8 @@ export mse, mae, msle,
     poisson_loss,
     hinge_loss, squared_hinge_loss,
     ctc_loss,
-    focal_loss
+    binary_focal_loss,
+    categorical_focal_loss
 
 include("utils.jl")
 include("functions.jl")
--- a/src/losses/Losses.jl
+++ b/src/losses/Losses.jl
@@ -18,7 +18,8 @@ export mse, mae, msle,
     dice_coeff_loss,
     poisson_loss,
     hinge_loss, squared_hinge_loss,
-    ctc_loss
+    ctc_loss,
+    focal_loss
 
 include("utils.jl")
 include("functions.jl")
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -430,16 +430,17 @@ function tversky_loss(ŷ, y; β = ofeltype(ŷ, 0.7))
 end
 
 """
-    binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, ϵ=epseltype(ŷ))
+    binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
 
 Return the [binary_focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
-Can be used in classification tasks in the presence of highly imbalanced classes. 
+which can be used in classification tasks with highly imbalanced classes.
 It down-weights well-classified examples and focuses on hard examples.
+The input, 'ŷ', is expected to be unnormalized.
 
-γ(default=2.0) is a number called modulating factor. 
-For γ=0, the loss is mathematically equivalent to binarycrossentropy(ŷ, y)
+The modulating factor, `γ`, controls the down-weighting strength.
+For `γ == 0`, the loss is mathematically equivalent to [`Losses.binarycrossentropy`](@ref).
 
-See also: ['focal_loss'](@ref)
+See also: [`Losses.focal_loss`](@ref) for multi-class setting
 
 """
 function binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
@@ -452,12 +453,17 @@ function binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=e
 end
 
 """
-    focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, ϵ=epseltype(ŷ))
+    focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
 
 Return the [focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
-For γ=0, the loss is mathematically equivalent to crossentropy(ŷ, y)
+which can be used in classification tasks with highly imbalanced classes.
+It down-weights well-classified examples and focuses on hard examples.
+The input, `ŷ`, is expected to be unnormalized.
 
-See also: [`binary_focal_loss`](@ref)
+The modulating factor, `γ`, controls the down-weighting strength.
+For `γ == 0`, the loss is mathematically equivalent to [`Losses.crossentropy`](@ref).
+
+See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
 
 """
 function focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -430,7 +430,7 @@ function tversky_loss(ŷ, y; β = ofeltype(ŷ, 0.7))
 end
 
 """
-    binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
+    binary_focal_loss(ŷ, y; agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
 
 Return the [binary_focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
 which can be used in classification tasks with highly imbalanced classes.
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -430,20 +430,17 @@ function tversky_loss(ŷ, y; β = ofeltype(ŷ, 0.7))
 end
 
 """
-    binary_focal_loss(ŷ, y; agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
+    binary_focal_loss(ŷ, y; agg=mean, γ=2, ϵ=epseltype(ŷ))
 
 Return the [binary_focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
-which can be used in classification tasks with highly imbalanced classes.
-It down-weights well-classified examples and focuses on hard examples.
 The input, 'ŷ', is expected to be normalized (i.e. [`softmax`](@ref) output).
 
-The modulating factor, `γ`, controls the down-weighting strength.
 For `γ == 0`, the loss is mathematically equivalent to [`Losses.binarycrossentropy`](@ref).
 
 See also: [`Losses.focal_loss`](@ref) for multi-class setting
 
 """
-function binary_focal_loss(ŷ, y; agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
+function binary_focal_loss(ŷ, y; agg=mean, γ=2, ϵ=epseltype(ŷ))
     ŷ = ŷ .+ ϵ
     p_t = y .* ŷ  + (1 .- y) .* (1 .- ŷ)
     ce = -log.(p_t)
@@ -453,7 +450,7 @@ function binary_focal_loss(ŷ, y; agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype
 end
 
 """
-    focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
+    focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
 
 Return the [focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
 which can be used in classification tasks with highly imbalanced classes.
@@ -466,7 +463,7 @@ For `γ == 0`, the loss is mathematically equivalent to [`Losses.crossentropy`](
 See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
 
 """
-function focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
+function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
     ŷ = ŷ .+ ϵ
     agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=dims))
 end
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -458,7 +458,7 @@ true
 See also: [`Losses.focal_loss`](@ref) for multi-class setting
 
 """
-function binary_focal_loss(ŷ, y; agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
+function binary_focal_loss(ŷ, y; agg=mean, γ=2, ϵ=epseltype(ŷ))
     ŷ = ŷ .+ ϵ
     p_t = y .* ŷ  + (1 .- y) .* (1 .- ŷ)
     ce = -log.(p_t)
@@ -501,11 +501,10 @@ true
 See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
 
 """
-function focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
+function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
     ŷ = ŷ .+ ϵ
     agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=dims))
 end
 ```@meta
 DocTestFilters = nothing
 ```
-
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -458,7 +458,7 @@ true
 See also: [`Losses.focal_loss`](@ref) for multi-class setting
 
 """
-function binary_focal_loss(ŷ, y; agg=mean, γ=2, ϵ=epseltype(ŷ))
+function binary_focal_loss(ŷ, y; agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
     ŷ = ŷ .+ ϵ
     p_t = y .* ŷ  + (1 .- y) .* (1 .- ŷ)
     ce = -log.(p_t)
@@ -501,7 +501,7 @@ true
 See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
 
 """
-function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
+function focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
     ŷ = ŷ .+ ϵ
     agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=dims))
 end
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -430,23 +430,34 @@ function tversky_loss(ŷ, y; β = ofeltype(ŷ, 0.7))
 end
 
 """
-    focal_loss(yhat, y; dims=1, agg=mean, gamma=2.0, eps = eps(eltype(yhat))
+    binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
 
-         𝛄: modulating factor.
+    γ: modulating factor.
 
-Return the [focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
-Extremely useful for classification when you have highly imbalanced classes. It down-weights
-well-classified examples and focuses on hard examples. Loss is much high for misclassified points as compared to well-classified points. Used in single-shot detectors.
 """
-function focal_loss(ŷ, y; dims=1, agg=mean, 𝛄=2.0, eps = eps(eltype(ŷ)))
+function binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
     ŷ = ŷ .+ eps
-    p_t = [y==1 ? ŷ : 1-ŷ for (ŷ, y) in zip(ŷ, y)]
+    p_t = y .*ŷ  + (1 .- y) .* (1 .- ŷ)
     ce = -log.(p_t)
-    weight = (1 .- p_t) .^ 𝛄
+    weight = (1 .- p_t) .^ γ
     loss = weight .* ce
     agg(sum(loss, dims=dims))
 end
 
+"""
+    categorical_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
+    Softmax version of Focal Loss
+    γ: modulating factor.
+
+"""
+function categorical_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
+    ŷ = softmax(ŷ; dims=dims)
+    ŷ = ŷ .+ eps
+    ce = -y .* log.(ŷ)
+    weight = (1 .- ŷ) .^ γ
+    loss = weight .* ce
+    agg(sum(loss, dims=dims))
+end
 ```@meta
 DocTestFilters = nothing
 ```
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -430,12 +430,30 @@ function tversky_loss(ŷ, y; β = ofeltype(ŷ, 0.7))
 end
 
 """
-    binary_focal_loss(ŷ, y; agg=mean, γ=2, ϵ=epseltype(ŷ))
+    binary_focal_loss(ŷ, y; agg=mean, γ=2, ϵ=eps(ŷ))
 
 Return the [binary_focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
 The input, 'ŷ', is expected to be normalized (i.e. [`softmax`](@ref) output).
 
 For `γ == 0`, the loss is mathematically equivalent to [`Losses.binarycrossentropy`](@ref).
+
+# Example
+```jldoctest
+julia> y = [0  1  0
+            1  0  1]
+2×3 Array{Int64,2}:
+ 0  1  0
+ 1  0  1
+
+julia> ŷ = [0.268941  0.5  0.268941
+            0.731059  0.5  0.731059]
+2×3 Array{Float64,2}:
+ 0.268941  0.5  0.268941
+ 0.731059  0.5  0.731059
+
+julia> Flux.binary_focal_loss(ŷ, y) ≈ 0.0728675615927385
+true
+```
 
 See also: [`Losses.focal_loss`](@ref) for multi-class setting
 
@@ -450,7 +468,7 @@ function binary_focal_loss(ŷ, y; agg=mean, γ=2, ϵ=epseltype(ŷ))
 end
 
 """
-    focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
+    focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=eps(ŷ))
 
 Return the [focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
 which can be used in classification tasks with highly imbalanced classes.
@@ -459,6 +477,26 @@ The input, 'ŷ', is expected to be normalized (i.e. [`softmax`](@ref) output).
 
 The modulating factor, `γ`, controls the down-weighting strength.
 For `γ == 0`, the loss is mathematically equivalent to [`Losses.crossentropy`](@ref).
+
+# Example
+```jldoctest
+julia> y = [1  0  0  0  1
+            0  1  0  1  0
+            0  0  1  0  0]
+3×5 Array{Int64,2}:
+ 1  0  0  0  1
+ 0  1  0  1  0
+ 0  0  1  0  0
+
+julia> ŷ = softmax(reshape(-7:7, 3, 5) .* 1f0)
+3×5 Array{Float32,2}:
+ 0.0900306  0.0900306  0.0900306  0.0900306  0.0900306
+ 0.244728   0.244728   0.244728   0.244728   0.244728
+ 0.665241   0.665241   0.665241   0.665241   0.665241
+
+julia> Flux.focal_loss(ŷ, y) ≈ 1.1277571935622628
+true
+```
 
 See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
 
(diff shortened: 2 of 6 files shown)


src/losses/functions.jl
```diff
@@ -425,6 +425,23 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
     1 - num / den
 end
 
+"""
+    focal_loss(yhat, y; dims=1, agg=mean, gamma=2.0, eps = eps(eltype(yhat))
```
darsnack, line 429:
```suggestion
    focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps=epseltype(yhat))
```

src/losses/functions.jl
```diff
@@ -425,6 +425,23 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+
+         𝛄: modulating factor.
```
darsnack, line 431: I don't know how you entered "gamma" but it's not the right unicode character (some other variant than the standard). You can enter the correct character in Julia REPL or VS Code with `\gamma[TAB]`.

src/losses/functions.jl
```diff
@@ -425,6 +425,23 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+
+Return the [focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
+Extremely useful for classification when you have highly imbalanced classes. It down-weights
+well-classified examples and focuses on hard examples. Loss is much high for misclassified points as compared to well-classified points. Used in single-shot detectors.
```
darsnack, line 435:
```suggestion
well-classified examples and focuses on hard examples.
Loss is much high for misclassified points as compared to well-classified points.
Used in single-shot detectors.
```

src/losses/functions.jl
```diff
@@ -425,6 +425,23 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+"""
+function focal_loss(ŷ, y; dims=1, agg=mean, 𝛄=2.0, eps = eps(eltype(ŷ)))
+    ŷ = ŷ .+ eps
+    p_t = [y==1 ? ŷ : 1-ŷ for (ŷ, y) in zip(ŷ, y)]
```
darsnack, line 439:
```suggestion
    p_t = [isone(y) ? ŷ : 1 - ŷ for (ŷ, y) in zip(ŷ, y)]
```

darsnack (changes requested):
Thanks for this contribution! Left some small comments, but the approach looks good to me.

src/losses/functions.jl
DhairyaLGandhi: The iterating variables should be different from the names of the arrays

src/losses/functions.jl
shikhargoswami: I searched the unicode alternative in this Julia Unicode input doc(https://docs.julialang.org/en/v1/manual/unicode-input/).
It comes from \bfgamma[TAB]

DhairyaLGandhi:
Rather than loop iteration, it might be faster to do array operations. That might be faster for when you actually have to materialize the result with bigger arrays

src/losses/functions.jl
shikhargoswami: Committing all the asked changes in a bit.

src/losses/functions.jl
darsnack: I would just use `\gamma`.

shikhargoswami: Okay. Changing now

CarloLucibello: the docstring should be homogenized to that of other losses and report the definition

src/losses/functions.jl
CarloLucibello: comprehension is not the right way to do it, makes the loss not gpu-friendly

src/losses/functions.jl
CarloLucibello: although in this case maybe is a bit too long and can avoid it. In any case that gamma shouldn't stand there in the middle of nothing

CarloLucibello:
That paper is cited enough that it may be worth having this loss in flux, although pytorch doesn't have it and tensorflow has it only as an addon

src/losses/functions.jl
```diff
@@ -425,6 +425,25 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
     1 - num / den
 end
 
+"""
+    focal_loss(yhat, y; dims=1, agg=mean, gamma=2.0, eps = eps(eltype(yhat))
+
+         𝛄: modulating factor.
+
+Return the [focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
+Extremely useful for classification when you have highly imbalanced classes. It down-weights
```
CarloLucibello, line 434:
```suggestion
Can be used in classification tasks in the presence of highly imbalanced classes. It down-weights
```
let's not overstate

darsnack:
Yeah having never heard of this loss before, I checked the paper. It's well-cited, so good enough for me to include in Flux.

shikhargoswami:
@darsnack @CarloLucibello  Need help on converting the list compreshension to array operations. Any useful leads?

darsnack:
@shikhargoswami You might take a look at how `binarycrossentropy` is written to see how to address @DhairyaLGandhi's comment.

I didn't see a performance difference when testing, but I didn't use large array sizes. Even if there isn't a gap in performance, it would be good to use the same style as the other loss functions.

CarloLucibello:
assuming 0/1 you can write
```julia 
p_t = y .* ŷ  + (1 .- y) .* (1 .- ŷ)  
```
So this is used only for binary classification? This should be mentioned

src/losses/functions.jl
DhairyaLGandhi: Agreed

src/losses/functions.jl
DhairyaLGandhi: Ref https://github.com/FluxML/Flux.jl/pull/1489#issuecomment-770243087

src/losses/functions.jl
```diff
@@ -425,6 +425,25 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+well-classified examples and focuses on hard examples.
+Loss is much high for misclassified points as compared to well-classified points.
```
CarloLucibello, line 436: this sentence can be deleted, states something quite trivial

test/losses.jl
```diff
@@ -174,3 +174,16 @@ end
...
 end
+
+y = [0 1 1]
+ŷ = [0.1 0.7 0.9]
+y1 = [1 0
+      0 0
+      0 1]
+ŷ1 = [0.6 0.3
+      0.3 0.1
+      0.1 0.6]
```
CarloLucibello, line 185: all these variables should be defined inside the testset

shikhargoswami:
I guess I only implemented Binary classification. Here are the changes made:
- [x] List comprehension -> Array operation
- [x] Implemented categorical_focal_loss and added it's test
- [x] Minor changes in docstring

Check if it needs any other change

src/losses/functions.jl
```diff
@@ -425,7 +425,35 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
 
+"""
+    binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
+
+    γ: modulating factor.
+
+"""
+function binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
+    ŷ = ŷ .+ eps
+    p_t = y .*ŷ  + (1 .- y) .* (1 .- ŷ)
```
darsnack, line 436:
```suggestion
    p_t = y .* ŷ  + (1 .- y) .* (1 .- ŷ)
```

src/losses/functions.jl
```diff
@@ -425,7 +425,35 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+    categorical_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
+    Softmax version of Focal Loss
+    γ: modulating factor.
+
+"""
+function categorical_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
+    ŷ = softmax(ŷ; dims=dims)
+    ŷ = ŷ .+ eps
+    ce = -y .* log.(ŷ)
+    weight = (1 .- ŷ) .^ γ
```
darsnack, line 453: I'm not familiar with the magnitude of the numerical stability issue, but I wonder if `logsoftmax` is better to use here even if it requires more work.

Maybe someone with a better understanding can comment on the trade-off. @CarloLucibello ?

src/losses/Losses.jl
```diff
@@ -18,7 +18,8 @@ export mse, mae, msle,
     dice_coeff_loss,
     poisson_loss,
     hinge_loss, squared_hinge_loss,
-    ctc_loss
+    ctc_loss,
+    focal_loss
```
darsnack: This needs to be updated for the two variants.

darsnack (changes requested):
Personally, I would prefer the names `focal_loss`/`binary_focal_loss` to match `crossentropy`/`binarycrossentropy`.

src/losses/functions.jl
shikhargoswami: @darsnack yhat is being used at two places. In computing ce (with log) and in weight (without log)

darsnack: Yeah that's why I suggested there is a trade-off. It would require computing the same thing twice but potentially be more numerically stable. I don't know enough about the stability issue to know what's the right call though.

src/losses/functions.jl
```diff
@@ -425,7 +425,35 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
     1 - num / den
 end
 
+"""
+    binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
+
+    γ: modulating factor.
+
+"""
+function binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
```
ToucheSir, line 434: `γ` also needs to be converted to `ŷ`'s eltype.

src/losses/functions.jl
DhairyaLGandhi: We can do the efficient thing by default and switch if it's unstable

src/losses/functions.jl
```diff
@@ -425,7 +425,35 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+    ce = -log.(p_t)
+    weight = (1 .- p_t) .^ γ
+    loss = weight .* ce
+    agg(sum(loss, dims=dims))
+end
 
+"""
+    categorical_focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, eps = epseltype(ŷ))
+    Softmax version of Focal Loss
+    γ: modulating factor.
```
CarloLucibello, line 446: this doesn't look like a proper docstring. These 2 lines shouldn't be indented, the paper should be referenced, it should be wordier. Please take your time to observe how other docstrings are implemented in flux so that we don't have to go over and over with review comments. There is no hurry here.

src/losses/functions.jl
CarloLucibello: The numerical stability issue comes around quite often. We could try with 
```julia
ŷ = logsoftmax(ŷ; dims=dims)
loss = @. -y  * (1 - exp(ŷ))^γ * ŷ 
```
Hopefully won't lose much in performance and it's stable

src/losses/functions.jl
shikhargoswami: @CarloLucibello I guess for class imbalance purpose, only the binary classification loss is implemented in the paper. Do I put the reference of the paper?

DhairyaLGandhi: Binary is a special case of Multiclass, so do we gain much from two separate definitions at all

shikhargoswami: @DhairyaLGandhi I think two separate definitions are usually implemented to achieve numerical efficiency and stability as in `crossentropy/binarycrossentropy`. The `ce` in `binarycrossentropy` is a dense array where all the elements are equal in `dims=1` so just taking the `mean` would suffice. But in `crossentropy` case, `ce` is a sparse(same as one-hot) matrix where we first `sum` in `dims=1` and take `mean`.

It just made me realise that my implementation is for logit outputs in categorical case. I guess maybe I need to implement all the four cases binary/categorical/logitbinary/logitcategorical. 

Are the logit ones really necessary(again, might be useful for numerical stability)?

darsnack: I think having binary and multi-class makes sense, because the paper has binary. People using the loss will refer to the paper and expect binary (i.e. not one hot) inputs.

I would not implement the logit variants and make the docstring clear that the input should be￼ unnormalized. No need to have name-explosion over a relatively rare loss function.

shikhargoswami: Agreed.

src/losses/functions.jl
DhairyaLGandhi: Do you have references to issues with log? I'd rather not lose performance in the average case here.

src/losses/functions.jl
```diff
@@ -425,7 +425,41 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+    agg(loss)
+end
 
+"""
+    focal_loss(ŷ, y; dims=1, agg=mean, γ=2.0, ϵ=epseltype(ŷ))
+
+Return the [focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
+For γ=0, the loss is mathematically equivalent to crossentropy(ŷ, y)
+
+See also: [`binary_focal_loss`](@ref)
```
darsnack, line 456:
```suggestion
See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
```

darsnack (changes requested):
Some suggestions on the docstrings.

src/losses/functions.jl
shikhargoswami: @darsnack Done. Can I commit the final changes now?

darsnack:
I would commit the changes that address the comments above. I think this is close to being ready for approval, but we'll need to come to a consensus on the numerical stability vs. performance issue. So probably we won't be able to merge today.

src/losses/functions.jl
shikhargoswami: > The numerical stability issue comes around quite often. We could try with
> 
> ```julia
> ŷ = logsoftmax(ŷ; dims=dims)
> loss = @. -y  * (1 - exp(ŷ))^γ * ŷ 
> ```
> 
> Hopefully won't lose much in performance and it's stable

This would be for the logit variant of the loss but based on the suggestions of all of you, I guess we are not implementing the variant so it is anyway scraped.

DhairyaLGandhi:
I'll also need to review the api before we make a final call

src/losses/functions.jl
```diff
@@ -425,7 +425,47 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+It down-weights well-classified examples and focuses on hard examples.
+The input, 'ŷ', is expected to be unnormalized.
+
+The modulating factor, `γ`, controls the down-weighting strength.
+For `γ == 0`, the loss is mathematically equivalent to [`Losses.binarycrossentropy`](@ref).
+
+See also: [`Losses.focal_loss`](@ref) for multi-class setting
+
+"""
+function binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
```
darsnack, line 442: I don't think this should have the `dims` keyword if it isn't used anywhere.

src/losses/functions.jl
```diff
@@ -425,7 +425,47 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+It down-weights well-classified examples and focuses on hard examples.
+The input, `ŷ`, is expected to be unnormalized.
+
+The modulating factor, `γ`, controls the down-weighting strength.
+For `γ == 0`, the loss is mathematically equivalent to [`Losses.crossentropy`](@ref).
+
+See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
+
+"""
+function focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
```
darsnack, line 465: `binary_focal_loss` implements the softmax for the user (accepting unnormalized input), but `focal_loss` doesn't. I think you want to add a `softmax` operation here.

src/losses/functions.jl
```diff
@@ -425,7 +425,47 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+    ŷ = ŷ .+ ϵ
+    agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=1))
```
darsnack, line 467: In addition to the comments about softmax.
```suggestion
    agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=dims))
```

test/losses.jl
```diff
@@ -174,3 +174,37 @@ end
...
+    y1 = [1 0
+          0 0
+          0 1]
+    ŷ1 = [0.4 0.2
+          0.5 0.5
+          0.1 0.3]
+    @test Flux.focal_loss(ŷ, y) ≈ 1.1277571935622628
+    @test Flux.focal_loss(ŷ1, y1) ≈ 0.45990566879720157
+    @test Flux.focal_loss(ŷ, y; γ=0.0) == Flux.crossentropy(ŷ, y)
+end
```
darsnack, line 210: These test cases all use normalized versions of `ŷ` which remains unmodified post-softmax. Instead, we should also have test cases where the sum along the rows isn't 1.

darsnack (changes requested):

src/losses/functions.jl
shikhargoswami: Both the functions don't implement softmax, they take normalized ŷs(which has to be the output of the last softmax layer of the model). It is in the logit loss function that we need to normalize(the user may not have used softmax layer).

test/losses.jl
shikhargoswami: Ref [#1489(comment)](https://github.com/FluxML/Flux.jl/pull/1489#discussion_r568355858)

src/losses/functions.jl
darsnack: You're right! No clue what I was looking at in reference to my comment.

src/losses/functions.jl
```diff
@@ -425,7 +425,47 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
     1 - num / den
 end
 
+"""
+    binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
+
+Return the [binary_focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
+which can be used in classification tasks with highly imbalanced classes.
+It down-weights well-classified examples and focuses on hard examples.
+The input, 'ŷ', is expected to be unnormalized.
```
darsnack, line 434:
```suggestion
The input, 'ŷ', is expected to be normalized (i.e. [`softmax`](@ref) output).
```

src/losses/functions.jl
```diff
@@ -425,7 +425,47 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+    agg(loss)
+end
 
+"""
+    focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
+
+Return the [focal_loss](https://arxiv.org/pdf/1708.02002.pdf)
+which can be used in classification tasks with highly imbalanced classes.
+It down-weights well-classified examples and focuses on hard examples.
+The input, `ŷ`, is expected to be unnormalized.
```
darsnack, line 457:
```suggestion
The input, 'ŷ', is expected to be normalized (i.e. [`softmax`](@ref) output).
```

darsnack (changes requested):

src/losses/functions.jl
```diff
@@ -425,7 +425,47 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
     1 - num / den
 end
 
+"""
+    binary_focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
```
darsnack, line 429: Need to remove `dims` from the docstring too

darsnack (commented):
This looks good to go for me, but @DhairyaLGandhi will need to approve the final API.

src/losses/functions.jl
shikhargoswami: oops sorry! Doing it now

shikhargoswami: Everybody pardon me for doing silly mistakes!

src/losses/functions.jl
```diff
@@ -425,7 +425,47 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+It down-weights well-classified examples and focuses on hard examples.
+The input, 'ŷ', is expected to be normalized (i.e. [`softmax`](@ref) output).
+
+The modulating factor, `γ`, controls the down-weighting strength.
+For `γ == 0`, the loss is mathematically equivalent to [`Losses.binarycrossentropy`](@ref).
+
+See also: [`Losses.focal_loss`](@ref) for multi-class setting
+
+"""
+function binary_focal_loss(ŷ, y; agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
```
DhairyaLGandhi, line 442: Is this the standard gamma or is it just the fonts tripping up because we don't want to make unicode too hard for users. Please also refer to current master about style guides around whitespace

test/losses.jl
```diff
@@ -174,3 +174,37 @@ end
...
+        0.665241   0.665241   0.665241   0.665241   0.665241]
+    y1 = [1 0
+          0 0
+          0 1]
+    ŷ1 = [0.4 0.2
+          0.5 0.5
+          0.1 0.3]
+    @test Flux.focal_loss(ŷ, y) ≈ 1.1277571935622628
+    @test Flux.focal_loss(ŷ1, y1) ≈ 0.45990566879720157
+    @test Flux.focal_loss(ŷ, y; γ=0.0) ≈ Flux.crossentropy(ŷ, y)
```
DhairyaLGandhi, line 209: Should also add gradient tests and GPU tests

test/losses.jl
```diff
@@ -174,3 +174,37 @@ end
...
+end
+
+@testset "focal_loss" begin
+    y = [1  0  0  0  1
+         0  1  0  1  0
+         0  0  1  0  0]
+
+    ŷ = [0.0900306  0.0900306  0.0900306  0.0900306  0.0900306
+        0.244728   0.244728   0.244728   0.244728   0.244728
+        0.665241   0.665241   0.665241   0.665241   0.665241]
```
DhairyaLGandhi, line 200: Where did you find these values?

DhairyaLGandhi (commented):
Thanks, I've added a few last thoughts, but this is looking good. We might want to review how we write our losses but that would be a more general shift, so keeping it consistent with the current seems sensible to me. Thanks again for the good work @shikhargoswami

test/losses.jl
shikhargoswami: `ŷ = softmax(reshape(-7:7, 3, 5) .* 1f0)`

src/losses/functions.jl
shikhargoswami: Yes. It is given in here: [Link](https://docs.julialang.org/en/v1/manual/unicode-input/). It comes from `\gamma[TAB]`

test/losses.jl
shikhargoswami: Any leads on how to write tests for these?

test/losses.jl
DhairyaLGandhi: We should have a grads test in the test/cuda directory

test/losses.jl
DhairyaLGandhi: https://github.com/FluxML/Flux.jl/pull/1489#discussion_r569580943

test/losses.jl
DhairyaLGandhi: Best to use that instead

CarloLucibello:
I think it would be better to assume logits as inputs instead of probabilities. In model-zoo we use `logitcrossentropy` everyqhere instead of crossentropy for numerical stability

darsnack:
This goes back to the numerical stability question. We would need to rework the implementation to make the logits version stable. If we do switch to logits, maybe the naming should be `logitfocalloss` and `logitbinaryfocalloss`. It's tough on the eyes but consistent with the cross entropy convention.

CarloLucibello:
the fact that we have both `crossentropy` and `logitcrossentropy` is quite unfortunate, when I revisited the loss functions I thought about having a single definition `function crossentropy(yhat, y; logits=true)` but couldn't figure out a nice deprecation path. Working with unnormalized log-probabilities is what one should want in most cases for numerical stability. Should we experiment with this keyword arg approach here?  Or we can just leave this as it is, not worth too much overthinking

darsnack:
I like the possibility of making "logit" a kwarg, but I think it is better in a separate PR. I think for here, let's just decide which version (logit or not) we want, and we can keep it consistent with what we already have.

DhairyaLGandhi:
I think the less magical the function, the better, and this makes certain contracts very implicit, like which function softmax we use. It goes against the self documenting code we expect from Flux. I'd need to see some very compelling reasons to do opaque looking keywords in Flux.

It's one reason I feel like we should remove the `agg`, it doesn't justify being in every loss function when it clearly doesn't add much to most of them.

CarloLucibello:
let's keep it as it is then, it's already done and `logitbinaryfocalloss` is quite horrifying :smile:

DhairyaLGandhi:
I agree the names... Need work

src/losses/functions.jl
```diff
@@ -425,7 +425,47 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
     1 - num / den
 end
 
+"""
+    binary_focal_loss(ŷ, y; agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
```
CarloLucibello, line 429:
```suggestion
    binary_focal_loss(ŷ, y; agg=mean, γ=2, ϵ=eps(ŷ))
```
Even if not the exact signature, is informative enough

src/losses/functions.jl
```diff
@@ -425,7 +425,47 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+    ŷ = ŷ .+ ϵ
+    p_t = y .* ŷ  + (1 .- y) .* (1 .- ŷ)
+    ce = -log.(p_t)
+    weight = (1 .- p_t) .^ γ
+    loss = weight .* ce
+    agg(loss)
+end
 
+"""
+    focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
```
CarloLucibello, line 452:
```suggestion
    focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=eps(ŷ))
```
Even if not the exact signature, is informative enough

src/losses/functions.jl
shikhargoswami: Correct me if I'm wrong(and this is also a doubt). `γ=2` will get promoted to floating point when multiplied later. Is it more numerical efficient to declare `γ=2` than `γ=ofeltype(ŷ, 2.0)`?

shikhargoswami: Does promotion bear an overhead cost?

darsnack: No I think @CarloLucibello wants you to leave out the `ofeltype` just in the docstring. The code can remain the same.

shikhargoswami: Okay! My bad :)

CarloLucibello: Yes I meant just the docstring, but now that think about it `γ=2` is actually a better default since integer exponentiation takes a different and faster path

test/cuda/losses.jl
```diff
@@ -14,13 +14,24 @@ y = [1, 1, 0.]
...
+     0  0  1  0  0]
+@test focal_loss(x, y) ≈ focal_loss(gpu(x), gpu(y))
 
 @testset "GPU grad tests" begin
   x = rand(Float32, 3,3)
   y = rand(Float32, 3,3)
 
   for loss in ALL_LOSSES
-    gpu_gradtest(loss, x, y)
+    gpu_autodiff_test(loss, x, y)
```
darsnack, line 34: Why did you change this line? It is causing errors for the GPU tests.

darsnack (changes requested):

test/cuda/losses.jl
shikhargoswami: I didn't change. I pulled the recent changes from repo. I might have been changed idk.

shikhargoswami: I didn't change it. I pulled the recent changes from repo. It might have been changed idk.

shikhargoswami: I didn't change it. I pulled the recent changes from repo. It might have been changed idk.

darsnack: Ah gotcha. @CarloLucibello do you know where this change is from?

darsnack: Okay I think you just don't have all the changes from the repo. How did you pull these changes? Can you rebase?

src/losses/functions.jl
```diff
@@ -425,7 +425,83 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+ 0.731059  0.5  0.731059
+
+julia> Flux.binary_focal_loss(ŷ, y) ≈ 0.0728675615927385
+true
+```
+
+See also: [`Losses.focal_loss`](@ref) for multi-class setting
+
+"""
+function binary_focal_loss(ŷ, y; agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
```
CarloLucibello, line 457:
```suggestion
function binary_focal_loss(ŷ, y; agg=mean, γ=2, ϵ=epseltype(ŷ))
```

src/losses/functions.jl
```diff
@@ -425,7 +425,83 @@ function tversky_loss(ŷ, y; β=ofeltype(ŷ, 0.7))
...
+ 0.665241   0.665241   0.665241   0.665241   0.665241
+
+julia> Flux.focal_loss(ŷ, y) ≈ 1.1277571935622628
+true
+```
+
+See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
+
+"""
+function focal_loss(ŷ, y; dims=1, agg=mean, γ=ofeltype(ŷ, 2.0), ϵ=epseltype(ŷ))
```
CarloLucibello, line 500:
```suggestion
function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
```

shikhargoswami:
I think I got issues with the runtime of my os. Can i close this PR and open a new one?

darsnack:
Sure, but preferably don't delete your fork so that we don't lose all the work that's been put in. Or you can give @CarloLucibello or me write access to the fork. Seems like a simple rebase should fix things.

darsnack:
@shikhargoswami that should fix it. Can you address the remaining comments?

darsnack (approved):
Let's wait for CI to pass then I think this can merge.

shikhargoswami:
@darsnack Thanks a lot for the help!

darsnack:
No problem! Thank you for your contribution!

darsnack:
bors cancel

darsnack (changes requested):
@shikhargoswami sorry to delay, but I realized you didn't add the focal loss docstring to the actual docs in `docs/models/losses.md`.

darsnack (approved):
Thanks!

DhairyaLGandhi:
Kyle, for the future, this is an API related change and the process does require a final approval from me.

darsnack:
Sorry, my mistake, I thought you had approved of the API. Next time I will leave the final merge to you on API changes.

PR #1492, opened 2021-02-01
Add sparsecrossentropy and logitsparsecrossentropy

findmyway:
The `sparsecrossentropy` and `logitsparsecrossentropy` provide similar functionality like [SparseCategoricalCrossentropy](https://www.tensorflow.org/api_docs/python/tf/keras/losses/SparseCategoricalCrossentropy) in tensorflow.

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


darsnack:
Is the idea here that `y` is just indices and not the full one-hot vector? If so, could we define a fast path for `OneHotArray` instead of a different loss function?

DhairyaLGandhi:
Yeah, could you provide an MWE? Is it slow? Ideally we wouldn't have this function and simply make regular crossentropy work. Is this using a specific array type? Then we can just use dispatch too.

darsnack:
For example, in `logitcrossentropy`, we do `y .* logsoftmax(ŷ; dims=dims)`, but if we didn't broadcast the matmul in the 2D case, we could hit the fast path for `OneHotArray` which does exactly what this loss function does.

Maybe what we should is generalize the [matmul index trick](https://github.com/FluxML/Flux.jl/blob/8f7ef92ca80dbcfd15016581bad31781ebb5063e/src/onehot.jl#L155) to more than 2D. The only reason we can't just remove the broadcast right now is that the losses can operate on ND arrays.

DhairyaLGandhi:
Definitely prefer to use broadcasting. I think it was a mistake to change that at all. Just arrays doesn't communicate what is happening under the hood well.

darsnack:
I'm sorry I don't understand what you mean (my comment was just explaining my previous one, not a response to you specifically).

DhairyaLGandhi:
I mean we should not assume the shape of the inputs as much as we can and make use of broadcasting for higher dimensional arrays. Yeah it was a bit of a tangent thought about our api design for loss functions and also the `agg` keyword.

findmyway:
Ah, yes. I forgot the `OneHotMatrix`. It should just work. Thanks for your remind.

PR #1495, opened 2021-02-02
Norm Layer fixes

DhairyaLGandhi:
I am also going through the `track_stats` call since it seems like it needs to be broken up in to smaller chunks that can individually address the differences in the forward passes.

ref #1397


DhairyaLGandhi:
see @CarloLucibello

DhairyaLGandhi:
Todo:

- [ ] Go through the tests, we don't want to use the autodiff test file as it is, doing more straightforward checks is.better.
- [ ] Remove `hasaffine`
- [ ] remove `testmode!` - use zygote instead and have an equivalent forward pass function available a la `dropout`
- [ ] determine forward pass at compile time

CarloLucibello:
this should be rebased and targeted to master

CarloLucibello:
commit history is still screwed, maybe better start with a new PR

src/layers/normalise.jl
```diff
@@ -423,35 +432,36 @@ function GroupNorm(chs::Int, G::Int, λ=identity;
             affine, track_stats, nothing)
 end
 
-function (gn::GroupNorm)(x)
+function (gn::GroupNorm)(x::AbstractArray{T,N}) where {T,N}
   @assert ndims(x) > 2
```
DhairyaLGandhi, line 436: shouldn't this check for number of channels != 1 and number of datapoints in a batch > 1

darsnack: Why does the batch size > 1? Doesn't GN compute within a sample? W.r.t. the # of channels, there appears to be an `@assert` that is commented out that checks for that?

CarloLucibello:
Can you get this in a state where git history is clean and tests are passing?

DhairyaLGandhi:
I've cleaned up the history, as we discussed. Getting the tests to pass is fine, but need to attend to some pressing work first.

DhairyaLGandhi:
Moved to 1509

PR #1496, opened 2021-02-03
Add Orthogonal initialization feature.

SomTambe:
As per issue #1431 I have added the Orthogonal matrix initialization feature. 

I will add the tests gradually. Just wondering what they can be.

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


src/utils.jl
```diff
@@ -174,6 +174,54 @@ end
...
+ -0.208255  -0.494864    0.553493   0.56362
+```
+
+# References
+[1] Saxe, A. et al. "Exact solutions to the nonlinear dynamics of learning in deep linear neural networks" * International Conference on Learning Representations* 2014
+
+"""
+function orthogonal(rng::AbstractRNG, dims...; gain = 1)
+  if length(dims) < 2
+    throw(error("Only Arrays with 2 or more dimensions are supported"))
```
johnnychen94, line 201: It should be either `error("Only Arrays with 2 or more dimensions are supported")` or `throw(ArgumentError("Only Arrays with 2 or more dimensions are supported"))`

`error(msg)` is equivalent to `throw(Error(msg))`

src/utils.jl
```diff
@@ -174,6 +174,54 @@ end
...
+  end
+
+  rows = dims[1]
+  cols = mapreduce(x->x,*, dims; init=1) ÷ rows
+  flattened = randn(rows,cols)
```
johnnychen94, line 206: I believe `Flux` doesn't support mixed-precision very nicely yet, so for consistency, it should return `Float32` array by default.

```suggestion
  flattened = randn(Float32, rows, cols)
```

src/utils.jl
```diff
@@ -174,6 +174,54 @@ end
...
+[1] Saxe, A. et al. "Exact solutions to the nonlinear dynamics of learning in deep linear neural networks" * International Conference on Learning Representations* 2014
+
+"""
+function orthogonal(rng::AbstractRNG, dims...; gain = 1)
+  if length(dims) < 2
+    throw(error("Only Arrays with 2 or more dimensions are supported"))
+  end
+
+  rows = dims[1]
+  cols = mapreduce(x->x,*, dims; init=1) ÷ rows
```
johnnychen94, line 205: It would be nice to format the codes before merging

```suggestion
  cols = mapreduce(x->x, *, dims; init=1) ÷ rows
```

src/utils.jl
```diff
@@ -174,6 +174,54 @@ end
...
+  flattened = randn(rows,cols)
+  if rows < cols
+    flattened = transpose(flattened)
+  end
+
+  Q,R = LinearAlgebra.qr(flattened)
```
johnnychen94, line 211:
```suggestion
Q, R = LinearAlgebra.qr(flattened)
```

src/utils.jl
```diff
@@ -174,6 +174,54 @@ end
...
+  d = LinearAlgebra.Diagonal(R)
+  ph = @. sign(d)
+  Q = (Array(Q)) * ph
+  if rows < cols
+    Q = transpose(Q)
+  end
```
johnnychen94, line 217: Another small one-liner trick and feel free to take any of it, or just ignore it.

```suggestion
Q = rows < cols ? transpose(Q) : Q
```

src/utils.jl
```diff
@@ -174,6 +174,54 @@ end
...
+  cols = mapreduce(x->x,*, dims; init=1) ÷ rows
+  flattened = randn(rows,cols)
+  if rows < cols
+    flattened = transpose(flattened)
+  end
+
+  Q,R = LinearAlgebra.qr(flattened)
+  d = LinearAlgebra.Diagonal(R)
+  ph = @. sign(d)
+  Q = (Array(Q)) * ph
```
johnnychen94, line 214:
```suggestion
  Q = Array(Q) * sign.(Diagonal(R))
```

Is this `Array` necessary?

src/utils.jl
```diff
@@ -174,6 +174,54 @@ end
...
+  if rows < cols
+    Q = transpose(Q)
+  end
+
+  return gain*reshape(Q,dims)
```
johnnychen94, line 219:
```suggestion
  return gain*reshape(Q, dims)
```

src/utils.jl
```diff
@@ -174,6 +174,54 @@ end
...
+end
+
+orthogonal(dims...; kwargs...) = orthogonal(Random.GLOBAL_RNG, dims...; kwargs...)
+orthogonal(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) ->orthogonal(rng, dims...; kwargs...)
```
johnnychen94, line 223:
```suggestion
orthogonal(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> orthogonal(rng, dims...; kwargs...)
```

src/utils.jl
```diff
@@ -174,6 +174,54 @@ end
...
+  if length(dims) < 2
+    throw(error("Only Arrays with 2 or more dimensions are supported"))
+  end
+
+  rows = dims[1]
+  cols = mapreduce(x->x,*, dims; init=1) ÷ rows
+  flattened = randn(rows,cols)
+  if rows < cols
+    flattened = transpose(flattened)
+  end
```
johnnychen94, line 209: Just a small trick:

```suggestion
rows < cols && (flattened = transpose(flattened))
```

johnnychen94 (commented):
I don't know the details of this initialization algorithm, so I have no idea whether this is coded correctly.

I made some format and style suggestions to it, hope you don't mind. Feel free to ignore it if you think it doesn't make sense.

src/utils.jl
SomTambe: I think I should keep my thing, looks more elegant 😄

src/utils.jl
SomTambe: Yes, `Array` is necessary. The QR decomposition of an array creates two resultant Arrays `Q` and `R`, out of which `Q` is of `LinearAlgebra.QRCompactWYQ` datatype, which causes a problem when the input matrix is rectangular. Refer to [this](https://github.com/JuliaLang/julia/issues/27397#issuecomment-394186670).

src/utils.jl
SomTambe: Wow! Nice and slick 😃

src/utils.jl
```diff
@@ -174,6 +174,51 @@ end
...
+
+# Examples
+```jldoctest; setup = :(using Random; Random.seed!(0))
+julia> Flux.orthogonal(5,4)
+5×4 Array{Float64,2}:
+ -0.205376  -0.0212576   0.304391   0.0927362
+ -0.684365   0.403613   -0.394419   0.457953
+  0.283074  -0.550965   -0.649911   0.377888
+ -0.60495   -0.536835   -0.151796  -0.566756
+ -0.208255  -0.494864    0.553493   0.56362
```
mcabbott, line 192: Might be better to show what this obeys, not just the numbers:
```suggestion
julia> W = Flux.orthogonal(5, 7);

julia> summary(W)
"5×7 Array{Float32,2}"

julia> W * W'
5×5 Array{Float32,2}:
  1.0         -2.42898f-8  6.32759f-8  -1.37195f-7  -2.19659f-8
 -2.42898f-8   1.0         4.03295f-8  -1.34284f-7   1.06978f-7
  6.32759f-8   4.03295f-8  1.0          7.93047f-8   2.6339f-7
 -1.37195f-7  -1.34284f-7  7.93047f-8   1.0          6.60169f-8
 -2.19659f-8   1.06978f-7  2.6339f-7    6.60169f-8   1.0
 
julia> W * W' ≈ I(5)
true
```

src/utils.jl
```diff
@@ -174,6 +174,51 @@ end
...
+```
+
+# References
+[1] Saxe, A. et al. "Exact solutions to the nonlinear dynamics of learning in deep linear neural networks" * International Conference on Learning Representations* 2014
```
mcabbott, line 196:
```suggestion
[1] Saxe, McClelland, Ganguli. "Exact solutions to the nonlinear dynamics of learning in deep linear neural networks", ICLR 2014, https://arxiv.org/abs/1312.6120
```

src/utils.jl
```diff
@@ -174,6 +174,51 @@ end
...
+
+"""
+function orthogonal(rng::AbstractRNG, dims...; gain = 1)
+  if length(dims) < 2
+    throw(ArgumentError("Only Arrays with 2 or more dimensions are supported"))
+  end
+
+  rows = dims[1]
+  cols = mapreduce(x->x, *, dims; init=1) ÷ rows
```
mcabbott, line 205: This is just `div(prod(dims), rows)` right? 

But why does reshaping this orthogonal matrix make sense? When would that be desirable -- instead of (say) orthogonal per slice, or just an error?

src/utils.jl
```diff
@@ -174,6 +174,51 @@ end
...
+  flattened = randn(Float32, rows, cols)
+  
+  rows < cols && (flattened = transpose(flattened))
+
+  Q, R = LinearAlgebra.qr(flattened)
```
mcabbott, line 210: This is a minor style point, but I think not re-using the same variable for different things is a good idea, both for human readers and for the computer. (Not that speed matters here.)
```suggestion
  mat = rows > cols ? randn(Float32, rows, cols) : randn(Float32, cols, rows)
  
  Q, R = LinearAlgebra.qr(mat)
```

src/utils.jl
SomTambe: I think I can apply the same thing for tests for this thing. Thanks for the wonderful docstring!

src/utils.jl
SomTambe: Orthogonal per slice could be done. That is a nice idea indeed.
In the PyTorch implementation, they flatten the matrix if the dimensions are greater than 2. I thought of implementing a similar thing. 
What do you think would be better?

src/utils.jl
SomTambe: Wouldn't it consume unwanted memory?

mcabbott: No, that shouldn't be affected. Re-using the label Q does not cause it to re-use the object which Q pointed at. For that you would need something like `Q .= ...`, but I think that won't work here, and anyway is an optimisation which we need not care about.

src/utils.jl
mcabbott: I met this idea 5 minutes ago, and haven't read the paper. It seems plausible that orthogonal matrices might be somehow interesting in Dense layers. That only needs ndims=2, and the simplest option would be just to restrict to that.

For a Conv layer, my guess would be that orthogonal in the channel dimensions (i.e. the last 2 dims of W). But I've no idea if that's something anyone does. Or possibly they do other things. 

I wonder why they reshape things in Pytorch, I would have thought the result didn't have any special properties.

src/utils.jl
```diff
@@ -174,6 +174,67 @@ end
...
+julia> summary(W)
+"5×7 Array{Float32,2}"
+
+julia> W * W'
+5×5 Array{Float32,2}:
+  1.0         -2.42898f-8  6.32759f-8  -1.37195f-7  -2.19659f-8
+ -2.42898f-8   1.0         4.03295f-8  -1.34284f-7   1.06978f-7
+  6.32759f-8   4.03295f-8  1.0          7.93047f-8   2.6339f-7
+ -1.37195f-7  -1.34284f-7  7.93047f-8   1.0          6.60169f-8
+ -2.19659f-8   1.06978f-7  2.6339f-7    6.60169f-8   1.0
```
CarloLucibello, line 199: we can delete this example, the one below is explicative enough

src/utils.jl
```diff
@@ -174,6 +174,67 @@ end
 kaiming_normal(dims...; kwargs...) = kaiming_normal(Random.GLOBAL_RNG, dims...; kwargs...)
 kaiming_normal(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> kaiming_normal(rng, dims...; kwargs...)
 
+"""
+    orthogonal([rng=GLOBAL_RNG], dims...; gain = 1)
+
+Return an `Array` of size `dims` which is a (semi) orthogonal matrix, as described in *Exact solutions to the nonlinear dynamics of learning in deep linear neural networks - Saxe, A. et al. (2013)*.
```
CarloLucibello, line 180:
```suggestion
Return an `Array` of size `dims` which is a (semi) orthogonal matrix, as described in Ref. [1]. 
```

CarloLucibello:
rename to `orthogonal_init` or `init_orthogonal`?

src/utils.jl
SomTambe: Done. Thanks for the suggestion!

SomTambe:
@CarloLucibello Since the other initialization techniques do not have any init keyword in their function name, I think it would be better to keep it `orthogonal` just to be consistent.

src/utils.jl
```diff
@@ -174,6 +174,59 @@ end
...
+  Q = Array(Q) * sign.(LinearAlgebra.Diagonal(R))
+  if rows < cols
+    Q = transpose(Q)
+  end
+
+  return gain * reshape(Q, dims)
+end
+
+orthogonal(dims...; kwargs...) = orthogonal(Random.GLOBAL_RNG, dims...; kwargs...)
+orthogonal(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> orthogonal(rng, dims...; kwargs...)
```
mcabbott, line 228: Should probably be `dims::Integer...`, if only for less-confusing error messages. 

Currying versions probably need to look like this (and need to be tested):
```
orthogonal(; kwargs...) = (dims...,) -> orthogonal(dims...; kwargs...)
orthogonal(rng::AbstractRNG; kwargs...) = (dims...,) -> orthogonal(rng, dims...; kwargs...)
```

SomTambe:
Could someone help me find where the docstring is going wrong? I am consistently getting the error `Base.Meta.ParseError("unexpected \",\"")` here in my docstring. Thank you!

darsnack:
I also thing the name should be more than `orthogonal`. The other init functions like "glorot" or "kaiming" are unique to ML parameter initialization but "orthogonal" is used in many contexts.

mcabbott:
Since it behaves a lot like `randn`, perhaps something like `rand_orthogonal` or `orthogonal_rand`?

src/utils.jl
SomTambe: I agree with you on the Conv layer point. I believe it should be orthogonal channel-wise. Let me open a new issue on it at the PyTorch repo.

SomTambe:
It is orthogonal **initialization** using `randn`, so can we try something like `init_orthogonal` or even `rand_orthogonal` seems fine to me as @mcabbott suggested.

src/utils.jl
```diff
@@ -174,6 +174,59 @@ end
 kaiming_normal(dims...; kwargs...) = kaiming_normal(Random.GLOBAL_RNG, dims...; kwargs...)
 kaiming_normal(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> kaiming_normal(rng, dims...; kwargs...)
 
+"""
+    orthogonal([rng=GLOBAL_RNG], dims...; gain = 1)
+
+Return an `Array` of size `dims` which is a (semi) orthogonal matrix, as described in Ref. [1]. 
+
+The input tensor must have at least 2 dimensions.
```
CarloLucibello, line 182: this should be rephrased, there is no input tensor here

CarloLucibello: also the role of `gain` should be explained

CarloLucibello:
`orthogonal_init` would be consistent with `sparse_init`

src/utils.jl
SomTambe: Done.

src/utils.jl
mcabbott: OK. For now, perhaps this should be restricted to matrices? Just`orthogonal_rang(rng, rows, cols; gain)`. This method is probably what you would want to call repeatedly to make higher-dim ones, in any case.

SomTambe: Let's rely on the PyTorch implementation since a large number of people use it. We can always change it later as we feel.

SomTambe:
Again, can anyone explain why is the doctesting failing?

src/utils.jl
mcabbott: We should not blindly copy them. But if it turns out they have a reason, that may be fine.

src/utils.jl
```diff
@@ -174,6 +174,57 @@ end
 kaiming_normal(dims...; kwargs...) = kaiming_normal(Random.GLOBAL_RNG, dims...; kwargs...)
 kaiming_normal(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> kaiming_normal(rng, dims...; kwargs...)
 
+"""
+    orthogonal_init([rng=GLOBAL_RNG], dims...; gain = 1)
+
+Return an `Array` of size `dims` which is a (semi) orthogonal_init matrix, as described in [1].
```
darsnack, line 180: Think you had a find-and-replace error here. Should be "orthogonal" not "orthogonal_init".

src/utils.jl
```diff
@@ -174,6 +174,57 @@ end
...
 
+"""
+    orthogonal_init([rng=GLOBAL_RNG], dims...; gain = 1)
+
+Return an `Array` of size `dims` which is a (semi) orthogonal_init matrix, as described in [1]. 
+
+The input must have at least 2 dimensions. 
+
+# Examples
+```jldoctest;
```
darsnack, line 185: I think this is causing the docs error.
```suggestion
```jldoctest
```

darsnack (changes requested):

src/utils.jl
SomTambe: Let me try, but the kaiming initialization docs have a semicolon there. Let me see once.

darsnack: Hmm okay maybe that's not it then. I'll take a closer look.

darsnack: No I think that should be it. The other cases have the `setup` after the semi-colon, and since your test case doesn't, I think it causes some weird parsing bug for Documenter.jl.

SomTambe: I had removed the setup thing just in one of these previous commits to check if to works, but unfortunately it didn't. waiting for travis to work up these tests.

SomTambe: Whoa docs passed!!!!

darsnack:
My only remaining concern is the point @mcabbott raised about reshaping. This approach breaks intuition, since I don't see how orthogonality is preserved. In this case, I would prefer not to just copy PyTorch. If someone is using the > 2D case, then I'd rather wait and see why/how they want to use it. Maybe that will shed light on what's the right thing to do.

SomTambe:
> My only remaining concern is the point @mcabbott raised about reshaping. This approach breaks intuition, since I don't see how orthogonality is preserved. In this case, I would prefer not to just copy PyTorch. If someone is using the > 2D case, then I'd rather wait and see why/how they want to use it. Maybe that will shed light on what's the right thing to do.

@darsnack I am currently in the progress of reading an article about it [here](https://hjweide.github.io/orthogonal-initialization-in-convolutional-layers). I am a bit busy, let me try understanding it till tomorrow.

darsnack:
No problem, thanks for the article. Okay, based on that explanation, being orthogonal in the output channels makes sense. In which case, your code will need to change slightly for the > 2D case. For conv kernels in Flux, the dimension that needs to be orthogonal is `dims[end]`. This is unfortunate, because for `Dense` it will be `dims[1]`.

I think what will make most sense is to have two dispatch cases. One for the 2D case, where you are orthogonal with respect to the smallest dimension. For more than two dimension arguments, we reshape w.r.t. the last dimension and apply the 2D case then reshape back.

We might be able to have a kwarg >2D case but AFAICT it must be `1` or `end`, so it doesn't seem worth it to me.

DhairyaLGandhi:
Can we retain just `orthoginal`, the case with `sparse` was to avoid a conflict, which is not a concern here and it would be nicer to not conflate the name

src/utils.jl
```diff
@@ -177,12 +177,14 @@ kaiming_normal(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> kaiming_no
...
 
-Return an `Array` of size `dims` which is a (semi) orthogonal_init matrix, as described in [1]. 
+Return an `Array` of size `dims` which is a (semi) orthogonal matrix, as described in [1]. 
 
 The input must have at least 2 dimensions. 
 
 # Examples
-```jldoctest;
+```jldoctest
+julia> using LinearAlgebra
```
DhairyaLGandhi, line 186: This should go into `setup` for the doctest

SomTambe:
> No problem, thanks for the article. Okay, based on that explanation, being orthogonal in the output channels makes sense. In which case, your code will need to change slightly for the > 2D case. For conv kernels in Flux, the dimension that needs to be orthogonal is `dims[end]`. This is unfortunate, because for `Dense` it will be `dims[1]`.
> 
> I think what will make most sense is to have two dispatch cases. One for the 2D case, where you are orthogonal with respect to the smallest dimension. For more than two dimension arguments, we reshape w.r.t. the last dimension and apply the 2D case then reshape back.
> 
> We might be able to have a kwarg >2D case but AFAICT it must be `1` or `end`, so it doesn't seem worth it to me.

@darsnack I interpreted something else from the article. 
Ideally, each kernel for a convolutional layer we have should be orthogonal to the others in its vector form.
So it would mean, suppose I choose kernels of size `3 x 3` for an RGB image, I will have to orthogonalize a `3 x 9` which I was already doing.
Apparently, my implementation looks correct to me. Kindly clarify.

darsnack:
> If the first dimension in Flux is not the number of channels, I can change the code accordingly.

PyTorch stores its conv kernel in OIWH format and Flux stores it in WHIO format (W = width, H = height, I = input channels, O = output channels). Hopefully, you will be convinced by my interpretation of the article below, but even if you are not, we would still need to change the code for this PR. In Flux, the first dimension is neither the input nor the output channel.

> Ideally, each kernel for a convolutional layer we have should be orthogonal to the others in its vector form.

You have this part correct, but it seems you are missing a dimension in your following statements.

Let's look at the dense case. Like the article, we have a layer `y = Wx + b`. According to the article, "we want the rows of `W` to be orthogonal to each other." Looking at what this means for the output:

y_=_left_y_1_y_2.pdf

You can see that making the rows orthogonal means that each output neuron is "orthogonal" to another. What would it mean to do the same thing for convolution layers? The output of a standard image convolution is 3D. Suppose we start the same way as the dense layer and treat each output pixel as separate, and we want to make them "orthogonal." This would be impossible, since the pixels within and output channel slice all share the same weight kernel, so there aren't multiple vectors to make orthogonal like we did with rows in the dense case.

Another interpretation, is that each output channel in a convolution is like an output neuron in a dense layer, since all the pixels in the output channel share the same weight which is distinct from other channels. In this case, it is possible to make each output channel slice orthogonal to the other channel slices, since each one is the result of a separate 3D weight kernel. Suppose you take a WHO output, `y`, and rearrange it to be O(WxH) (I am re-ordering so that it lines up better with the dense). Then the first column of `y` corresponds to the first pixel for each output channel. As you move across the columns, you are looking at the output as you slide the convolution kernel. `y_11` is the first output channel, first slide position. `y_12` is the first output channel, second slide position. Let's focus on the first column alone (so each element in the vector corresponds to a different output channel). If we write out the matrix multiply that results in this column vector:

y_=_left_y_1_y_2.pdf

So, hopefully, you can see that the analogue to "orthogonalizing" the rows of the dense layer corresponds to the output channels of the convolution layer. If the kernel is `Conv((k, k), i => o)` then you want to apply your code to the `(k*k*i) x o` matrix.

SomTambe:
@darsnack Nice! That was my only doubt. 

I shall update the code accordingly.

darsnack:
As I was saying before, we probably want to handle this with two methods. One that does the 2D case (only two dimension arguments — row and column), and then one that takes the >2D variable argument case and calls the 2D case with `prod(dims[1:end-1]), dims[end]` then reshapes the return.

SomTambe:
@darsnack Yup got it! Let me get back with some new code.

SomTambe:
@darsnack I have modified it for `dims > 2`. Kindly have a look.

src/utils.jl
```diff
@@ -174,6 +174,60 @@ end
...
+  Q = Array(Q) * sign.(LinearAlgebra.Diagonal(R))
+  if rows < cols
+    Q = transpose(Q)
+  end
+
+  return gain * reshape(Q, dims)
+end
+
+orthogonal_init(dims::Integer...; kwargs...) = orthogonal_init(Random.GLOBAL_RNG, dims...; kwargs...)
+orthogonal_init(rng::AbstractRNG; kwargs...) = (dims::Integer...; kwargs...) -> orthogonal_init(rng, dims...; kwargs...)
```
darsnack, line 229: This a bug fix we merged recently that you'll need to adopt.

```suggestion
orthogonal_init(rng::AbstractRNG; init_kwargs...) = (dims::Integer...; kwargs...) -> orthogonal_init(rng, dims...; init_kwargs..., kwargs...)
```

src/utils.jl
```diff
@@ -174,6 +174,60 @@ end
...
+"""
+function orthogonal_init(rng::AbstractRNG, dims...; gain = 1)
+  if length(dims) < 2
+    throw(ArgumentError("Only Arrays with 2 or more dimensions are supported"))
+  end
+  if length(dims) == 2
+    rows = dims[1]
+  else
+    rows = dims[end]
+  end
```
darsnack, line 215: Personally, I think this would be less opaque if written as two methods. The second method makes it clear that for N-dimensions, we treat it as a matrix with the last dimension forming one side.
```julia
function orthogonal_init(rng::AbstractRNG, rows, cols; gain = 1)
  cols = div(prod(dims),rows)
  mat = rows > cols ? randn(Float32, rows, cols) : randn(Float32, cols, rows)

  Q, R = LinearAlgebra.qr(mat)
  Q = Array(Q) * sign.(LinearAlgebra.Diagonal(R))
  if rows < cols
    Q = transpose(Q)
  end

  return gain * reshape(Q, rows, cols)
end

function orthogonal_init(rng::AbstractRNG, dims...; kwargs...)
  @assert length(dims) > 2
  rows, cols = prod(dims[1:(end - 1)]), dims[end]
  
  return reshape(orthogonal_init(rng, rows, cols; kwargs...), dims)
end
```

darsnack (changes requested):

src/utils.jl
SomTambe: So that will have to change for both the positions where there is `kwargs` right?

darsnack: No only for the one I listed, since it returns a closure. Previously the arguments to the anonymous function were overwriting the "initial" arguments when constructing the closure.

src/utils.jl
SomTambe: Is this necessary? Can't we simply add the details in the docstring saying what we do for `dims > 2`?

darsnack: It does the same thing, yes. And we'd have to explain the behavior in the docstring regardless of how we do it. But I would argue that this approach is more self-documenting which is something we strive for.

darsnack: Generally, I would say if there is a function with multiple behavioral cases, it is more idiomatic in Julia to use dispatch to address those cases than if-statements.

mcabbott: Two methods seems much clearer to me. The simple one need not call `div`, and not reshape, can just return `gain .* Q`. The complex one could also be `orthogonal_init(rng::AbstractRNG, a::Integer, bs::Integer...; kwargs...)` to only match > 1 dims.

DhairyaLGandhi:
I think the `init` moniker isn't well defined, this is a utility to generate matrices and arrays that have the property of being orthogonal. Being used to initialise NNs is just happenstance. Therefore, as long as we don't export it, `Flux.orthogonal` should remain unambiguous about its intent, so should be fine to go ahead with.

darsnack:
In the N-dimensional case, I feel that we are being NN-biased for the behavior, so I wouldn't treat this as a generic orthogonal matrix generating function. I still feel it is more clear for users what the intent is with "init," but as long as it isn't exported then it's fine.

SomTambe:
Okay I am not used to this, what should I do now about the `init` thing 🤣 ? Whom should I listen to - @DhairyaLGandhi or @darsnack ?

darsnack:
@DhairyaLGandhi gets final say!

SomTambe:
Honestly speaking, I like `orthogonal` as I had previously kept it, so if @DhairyaLGandhi says so, I should continue keeping that! Waiting for your opinion @DhairyaLGandhi ...

DhairyaLGandhi:
Mimicking pytorch behavior isn't ideal imo. Lots of respect for what they do, but fear that it shouldn't let us keep ourselves from making different judgement calls.

In this case, it's something that should (tm) be the least surprising for new users, but in general I agree doing the correct thing is a safer long term bet, so I would be happy to consider changing its behavior to be more generic. But for now, dropping the init should work.

darsnack:
This is just a comment, not feedback for the PR.

> Mimicking pytorch behavior isn't ideal imo. Lots of respect for what they do, but fear that it shouldn't let us keep ourselves from making different judgement calls.

To be fair, in this case the PyTorch behavior was the "generic" thing to do w.r.t. our goal of initializing NN weights. There is a more generic definition of orthogonality for N > 2 (in Einstein notation for tensors), but I don't think that the most generic behavior mathematically will produce our desired goal.

DhairyaLGandhi:
Are you referring to the channel and batch dimensions here?

src/utils.jl
SomTambe: Tests are failing. Could you figure out what is going wrong? Or was it `orthogonal_init(rng::AbstractRNG; init_kwargs..., kwargs...)` ?

src/utils.jl
```diff
@@ -174,6 +174,65 @@ end
...
+julia> W * W' ≈ I(5)
+true
+
+julia> W2 = Flux.orthogonal(7, 5);
+
+julia> W2 * W2' ≈ I(7)
+false
+
+julia> W2' * W2 ≈ I(5)
+true
```
darsnack, line 202:
```suggestion
true

julia> W3 = Flux.orthogonal(3, 3, 2, 4);

julia> transpose(reshape(W3, :, 4)) * reshape(W3, :, 4) ≈ I(4)
true
```

src/utils.jl
```diff
@@ -174,6 +174,65 @@ end
...
 kaiming_normal(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> kaiming_normal(rng, dims...; kwargs...)
 
+"""
+    orthogonal([rng=GLOBAL_RNG], dims...; gain = 1)
+
+Return an `Array` of size `dims` which is a (semi) orthogonal matrix, as described in [1]. 
+
+The input must have at least 2 dimensions. 
+
+For input dimensions > 2, the array is flattened with respect to the last dimension, and then an orthogonal matrix is computed before reshaping it to original dimensions.
```
darsnack, line 184:
```suggestion
The input must have at least 2 dimensions.
For `length(dims) > 2`, a `prod(dims[1:(end - 1)])` by `dims[end]` orthogonal matrix 
is computed before reshaping it to the original dimensions.
```

src/utils.jl
```diff
@@ -174,6 +174,65 @@ end
...
+function orthogonal(rng::AbstractRNG, dims...; kwargs)
+  if length(dims) < 2
+    throw(ArgumentError("Only Arrays with 2 or more dimensions are supported"))
+  end
+
+  @assert length(dims) > 2
+  rows = dims[end]
+  cols = div(prod(dims),rows)
+  return reshape(orthogonal(rng, rows, cols; kwargs), dims)
+end
```
darsnack, line 231:
```suggestion
function orthogonal(rng::AbstractRNG, d1, ds...; kwargs...)
  dims = (d1, ds...)
  rows = dims[end]
  cols = div(prod(dims), rows)
  return reshape(orthogonal(rng, rows, cols; kwargs...), dims)
end
```

src/utils.jl
```diff
@@ -174,6 +174,65 @@ end
...
+julia> W2' * W2 ≈ I(5)
+true
+```
+
+# References
+[1] Saxe, McClelland, Ganguli. "Exact solutions to the nonlinear dynamics of learning in deep linear neural networks", ICLR 2014, https://arxiv.org/abs/1312.6120
+
+"""
+function orthogonal(rng::AbstractRNG, rows, cols; gain = 1)
+  cols = div(rows*cols,rows)
```
darsnack, line 210:
```suggestion
```

src/utils.jl
```diff
@@ -174,6 +174,65 @@ end
...
+  mat = rows > cols ? randn(Float32, rows, cols) : randn(Float32, cols, rows)
+
+  Q, R = LinearAlgebra.qr(mat)
+  Q = Array(Q) * sign.(LinearAlgebra.Diagonal(R))
+  if rows < cols
+    Q = transpose(Q)
+  end
+
+  return gain * reshape(Q, (rows, cols))
```
darsnack, line 219:
```suggestion
  return gain * Q
```

src/utils.jl
```diff
@@ -174,6 +174,65 @@ end
...
+
+julia> W2 = Flux.orthogonal(7, 5);
+
+julia> W2 * W2' ≈ I(7)
+false
+
+julia> W2' * W2 ≈ I(5)
+true
+```
+
```
darsnack, line 204:
```suggestion

# See also
* kaiming initialization using normal distribution: [`kaiming_normal`](@ref Flux.kaiming_normal)
* kaiming initialization using uniform distribution: [`kaiming_uniform`](@ref Flux.kaiming_uniform)
* glorot initialization using normal distribution: [`glorot_normal`](@ref Flux.glorot_normal)
* glorot initialization using uniform distribution: [`glorot_uniform`](@ref Flux.glorot_uniform)
* sparse initialization: [`sparse_init`](@ref Flux.sparse_init)

```

darsnack (changes requested):
Probably also need to add `orthogonal` as a cross-reference to the other initialization docstrings.

darsnack:
> Are you referring to the channel and batch dimensions here?

I mean that for an `Array{T, N}`, there is definition of what it means to be orthogonal when `N > 2`. We don't use that definition here, and we treat the weight kernel as 2D. It makes sense to do that for weight kernels in NNs given that our goal is for each output neuron/channel to be produced by orthogonal weights.

Thus, my comment that this function is not a generic orthogonal array initializer, and my comment that we match PyTorch here because it is the right thing to do (not just cause we don't know better).

src/utils.jl
darsnack: Can you rebase? Probably you just need the updated tests from that PR.

SomTambe:
@darsnack @DhairyaLGandhi I have made the required changes. All builds passing, only doctest failing which also seems to be from some other set of tests.

darsnack:
@SomTambe No problem, thanks for putting the effort in to see this through. The only remaining changes are the docstring changes. Do you mind making those?

SomTambe:
> @SomTambe No problem, thanks for putting the effort in to see this through. The only remaining changes are the docstring changes. Do you mind making those?

Let me try figuring out what has gone wrong. If you get time, can you mark out in brief what needs to be changed? If you are occupied (which I expect), the general drill will be that I will take some time (I guess till tomorrow morning, since it is night here in India) and I will make the changes 😃 .

SomTambe:
> @SomTambe No problem, thanks for putting the effort in to see this through. The only remaining changes are the docstring changes. Do you mind making those?

Let me try figuring out what has gone wrong. If you get time, can you mark out in brief what needs to be changed? If you are occupied (which I expect), the general drill will be that I will take some time (I guess till tomorrow morning, since it is night here in India) and I will make the changes 😃 .

darsnack:
They should be marked out in my last review above ☝🏾

darsnack:
Don't worry about the doctest errors btw. The issue is addressed in another PR, we'll either just need to rebase, or we'll merge this PR with the failing test (knowing that it is already fixed in another PR).

darsnack:
You'll also need to add `orthogonal` as a reference in all the other init function's docstrings.

src/utils.jl
mcabbott: Might it be possible just to link to the heading of the right section of the docs? That's more likely to stay upt o date than attempting to cross-reference all-to-all. 

And possibly "see also [`glorot_uniform`](@ref Flux.glorot_uniform), the default initialser" or something.

darsnack: Yeah I agree this list is becoming too cumbersome. @SomTambe can you update all the init docstrings to reflect this suggestion?

darsnack:
Sorry, one more thing! You need to add `orthogonal` to `docs/src/utilities.md`.

src/utils.jl
SomTambe: > Might it be possible just to link to the heading of the right section of the docs? That's more likely to stay up to date than attempting to cross-reference all-to-all.

Honestly speaking, I do not know how to do this. Right now, I will commit the `# See also` section as you have suggested. Maybe, I shall open another PR for making things tidy 😅.

src/utils.jl
SomTambe: @darsnack I don't understand why these are failing. I tried both the possibilities, `w' * w` as well as `w * w'`, none of them gives an Identity matrix. Surprising why it does not work. Any clues?

mcabbott: Your reshaping function isn't right, I think.

src/utils.jl
```diff
@@ -174,6 +174,72 @@ end
...
+* glorot initialization using normal distribution: [`glorot_normal`](@ref Flux.glorot_normal)
+* glorot initialization using uniform distribution: [`glorot_uniform`](@ref Flux.glorot_uniform)
+* sparse initialization: [`sparse_init`](@ref Flux.sparse_init)
+
+# References
+[1] Saxe, McClelland, Ganguli. "Exact solutions to the nonlinear dynamics of learning in deep linear neural networks", ICLR 2014, https://arxiv.org/abs/1312.6120
+
+"""
+function orthogonal(rng::AbstractRNG, rows, cols; gain = 1)
+  mat = rows > cols ? randn(Float32, rows, cols) : randn(Float32, cols, rows)
```
mcabbott, line 222: Note that you aren't using `rng` at all. 
```suggestion
function orthogonal(rng::AbstractRNG, rows::Integer, cols::Integer; gain = 1)
  mat = rows > cols ? randn(rng, Float32, rows, cols) : randn(rng, Float32, cols, rows)
```

CarloLucibello:
A very final attempt in favor of the `orthogonal_init` name, feel free to ignore. Would you ever give this PR the title "Add Orthogonal feature"? I guess not, because it would convey little information about its content. Same for `Flux.orthogonal`, someone looking at the code wouldn't guess what would be the purpose of this function without looking at the docs

src/utils.jl
mcabbott: I think the reason this strikes several of us as weird is partly that it's not type-stable to re-use Q, not just for different things, but for different types depending on the values of `rows, cols`. This isn't performance-critical code but that's where everyone's taste was honed. 

Again, I would write 
```
return rows > cols ? gain .* M : gain .* transpose(M)
```
where `M` is some name for the thing which isn't `Q` anymore, and the two branches match the branches which generate the random numbers above.

src/utils.jl
darsnack: Isn't `reshape(W3, :, 4)` correspond to `prod(dims[1:end-1]), dims[end]`?

mcabbott: Sorry, what I meant is that the vararg function which makes the reshaped array, has a mistake in its row/col defn.

src/utils.jl
darsnack: Yeah the mirrored if-else clause are a bit confusing. Should change that if nothing else.

src/utils.jl
darsnack: Ah I see, good catch

src/utils.jl
```diff
@@ -174,6 +174,72 @@ end
...
+    Q = transpose(Q)
+  end
+
+  return gain * Q
+end
+
+function orthogonal(rng::AbstractRNG, d1, ds...; kwargs...)
+  dims = (d1, ds...)
+  rows = dims[end]
+  cols = div(prod(dims), rows)
```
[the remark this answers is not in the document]
darsnack: @SomTambe this should resolve the doctest error

SomTambe:
@darsnack Made the required changes. I think we should go ahead with `orthogonal`. All tests passed finally!

darsnack (approved):
We still want to switch the docstrings to cross-reference the initialization sections of the docs instead of all-to-all with the other initialization functions.

But this PR is good to go, and we can make that change in a separate PR. Final review from @DhairyaLGandhi needed.

SomTambe:
@DhairyaLGandhi Could you review the request? Maybe merge it?

DhairyaLGandhi (approved):
Thanks! This looks great!

bors r+

PR #1497, opened 2021-02-04
Fix docs syntax for Join/ Split layers

DhairyaLGandhi:


DhairyaLGandhi (commented):
should be more readable

darsnack (approved):

Issue #1498, opened 2021-02-05
Does gain do anything in kaiming_uniform?


DevJac:
Consider:

```
julia> using Flux

julia> Flux.kaiming_uniform(gain=0.001)(2, 3)
2×3 Array{Float32,2}:
 -0.0064778  -0.899992  0.464747
 -0.850806   -0.913933  1.05438

julia> Flux.kaiming_uniform(gain=10000000)(2, 3)
2×3 Array{Float32,2}:
 -1.17412  -0.519475  -0.141687
 -1.24499  -0.468959  -0.174382

julia> Flux.kaiming_uniform(2, 3, gain=0.001)
2×3 Array{Float32,2}:
 0.000955879   0.000643053  -0.000561294
 0.00062913   -9.48081f-5    0.000849087

julia> Flux.kaiming_uniform(2, 3, gain=1000000)
2×3 Array{Float32,2}:
 -6.40772f5  5216.84       -1.86908f5
 -6.44296f5     4.93956f5  -6.97983f5
```

It seems when calling `Flux.kaiming_uniform(gain=0.001)(2, 3)` that `gain` does nothing.

I believe the intention is that you could do something like `Dense(2, 3, initW=Flux.kaiming_uniform(gain=10))`, but again, in this form `gain` does nothing.

The relevant code is here: https://github.com/FluxML/Flux.jl/blob/7e9a1802153196d562e54382a811fafb0e900a13/src/utils.jl#L132

DevJac:
The output of
```
using Random
using Flux: nfan

function kaiming_uniform(rng::AbstractRNG, dims...; gain = √2)
  @info("in main", gain)
  bound = Float32(√3 * gain / sqrt(first(nfan(dims...)))) # fan_in
  return (rand(rng, Float32, dims...) .- 0.5f0) .* 2bound
end

function kaiming_uniform(dims...; kwargs...)
    @info("in 2", kwargs)
    kaiming_uniform(Random.GLOBAL_RNG, dims...; kwargs...)
end

function kaiming_uniform(rng::AbstractRNG; kwargs...)
    @info("in 3", kwargs)
    (dims...; kwargs...) -> kaiming_uniform(rng, dims...; kwargs...)
end

@info("start, partial call")
kaiming_uniform(gain=1234)
@info("mid, full call")
kaiming_uniform(gain=1234)(2, 3)
@info("end")
```
is
```
[ Info: start, partial call
┌ Info: in 2
│   kwargs =
[...]

PR #1499, opened 2021-02-05
Fix layer init functions kwargs getting overwritten

DevJac:
There were a few functions of the form:
foo(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> foo(rng, dims...; kwargs...)

The intention was that you could do `foo(my_kw=42)` to get a callable, and then later
call that callable with the kwarg already set. However, the `kwargs` variable of the
lambda shadowed the original `kwargs`. This is a fix.

Fixes: https://github.com/FluxML/Flux.jl/issues/1498

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -135,7 +135,7 @@ function kaiming_uniform(rng::AbstractRNG, dims...; gain = √2)
 end
 
 kaiming_uniform(dims...; kwargs...) = kaiming_uniform(Random.GLOBAL_RNG, dims...; kwargs...)
-kaiming_uniform(rng::AbstractRNG; kwargs...) = (dims...; kwargs...) -> kaiming_uniform(rng, dims...; kwargs...)
+kaiming_uniform(rng::AbstractRNG; init_kwargs...) = (dims...; kwargs...) -> kaiming_uniform(rng, dims...; init_kwargs..., kwargs...)
 
 """
     kaiming_normal([rng=GLOBAL_RNG], dims...; gain = √2)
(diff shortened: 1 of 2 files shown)


ToucheSir (commented):
Nice catch! If you'll allow some bikeshedding, I think it may be clearer to rename the first `kwargs` to `init_args` or `init_kwargs`. Ff there are no objections to `kwargs_λ` though, I'm happy to approve this as-is.

DevJac:
I agree and will make the change you suggested.

DevJac:
> I think it may be clearer to rename the first kwargs to init_kwargs

Good suggestion, I have made that change.

darsnack (changes requested):
Can we add some tests to check for the bug you encountered?

This will allow the `init_kwargs` to be overwritten by `kwargs`. I wonder if a warning is worth it for that case. Anyone have any thoughts?

darsnack:
Actually I would suggest changing `merge(init_kwargs, kwargs)...` into `init_kwargs..., kwargs...`.

DevJac:
> Actually I would suggest changing merge(init_kwargs, kwargs)... into init_kwargs..., kwargs....

I don't understand what you mean by this. Do you mean make init_kwargs and kwargs into a pair or a tuple? Can a function take multiple kwargs?

DevJac:
I amended the commit to include some tests. They are not sophisticated, because I am not very familiar with the math behind these functions, but they do demonstrate the partial application of the `gain` or other kwargs, and ensures they have an effect.

CarloLucibello (approved):

darsnack:
What's the deal with `bors` here?

DhairyaLGandhi (approved):
bors r+

CarloLucibello:
c'mon

bors r+

CarloLucibello:
@DhairyaLGandhi what's happening with bors?

PR #1503, opened 2021-02-11
Add Julia 1.6 doc changes to CI

DhairyaLGandhi:
Test against 1.6


docs/src/data/onehot.md
```diff
@@ -44,13 +44,13 @@ Flux.onecold
 julia> using Flux: onehotbatch
 
 julia> onehotbatch([:b, :a, :b], [:a, :b, :c])
-3×3 Flux.OneHotArray{3,2,Array{UInt32,1}}:
+3×3 Flux.OneHotArray{UInt32,3,1,2,Vector{UInt32}}:
```
darsnack, line 47:
```suggestion
3×3 Flux.OneHotArray{3,2,Vector{UInt32}}:
```

DhairyaLGandhi:
The docs failing don't make much sense, possible that is an artefact of Documenter needing to update to 1.6.

Oh well, the state of this PR is better than master, so anyone want to approve this/ have suggestions for changes? @darsnack ?

darsnack (approved):
Documenter fails anytime Base (or any package) changes what `show` prints out. Unfortunately, Julia 1.6 changed the `Vector`/`Matrix` printing which really screwed up pretty much every print out. For future reference, you can use https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Fixing-Outdated-Doctests to automate replacing the REPL output when it happens.

But looks like docs are passing now, thanks!

DhairyaLGandhi:
I did do that locally, but the refs were screwed up there. Not sure if that was related to the state of machine which is a shared one.

bors r+

DhairyaLGandhi:
Bk suggests there is scalar indexing happening. Why is that in the ci? That undermines the value of the ci imo.

darsnack:
It's disabled for the GPU integration tests. Being triggered somewhere in the losses tests? I would guess it might be related to #1534. I can add an explicit GPU test for that case.

DhairyaLGandhi:
Not sure why bors is unhappy. Merging for now.

PR #1508, opened 2021-02-15
Add RNG state to shuffling data in DataLoader

DhairyaLGandhi:
This is similar in spirit to #1501, and allows users to define the RNG they want to use while shuffling, but helps us to not store the RNG in the struct. Removing the type restriction shouldn't hurt performance in this case either, its a simple function.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/data/dataloader.jl
+++ b/src/data/dataloader.jl
@@ -1,18 +1,18 @@
 # Adapted from Knet's src/data.jl (author: Deniz Yuret)
 using Random
 
-struct DataLoader{D}
+struct DataLoader{D, S}
     data::D
     batchsize::Int
     nobs::Int
     partial::Bool
     imax::Int
     indices::Vector{Int}
-    shuffle
+    shuffle::S
 end
 
 """
-    DataLoader(data; batchsize = 1, shuffle = x -> Random.shuffle(Random.GLOBAL_RNG, x), partial = true)
+    DataLoader(data; batchsize = 1, shuffle = x -> Random.shuffle!(Random.GLOBAL_RNG, x), partial = true)
 
 An object that iterates over mini-batches of `data`, each mini-batch containing `batchsize` observations
 (except possibly the last one).
@@ -21,7 +21,7 @@ Takes as input a single data tensor, or a tuple (or a named tuple) of tensors.
 The last dimension in each tensor is considered to be the observation dimension.
 
 By default, the dataloader shuffles the observations each time iterations are re-started.
-To not shuffle the data, pass `shuffle = identity`.
+To not shuffle the data, pass `shuffle = identity` or shuffle = false.
 
 If `partial = false`, drops the last mini-batch if it is smaller than the batchsize.
 
@@ -68,7 +68,7 @@ Usage example:
         @assert size(datum.labels) == (2,)
     end
 """
-function DataLoader(data; batchsize = 1, shuffle = identity, partial = true)
+function DataLoader(data; batchsize = 1, shuffle = x -> Random.shuffle!(Random.GLOBAL_RNG, x), partial = true)
     batchsize > 0 || throw(ArgumentError("Need positive batchsize"))
 
     n = _nobs(data)
@@ -77,6 +77,9 @@ function DataLoader(data; batchsize = 1, shuffle = identity, partial = true)
         batchsize = n
     end
     imax = partial ? n : n - batchsize + 1
+    shuffle = if isa(shuffle, Bool)
+      shuffle ? x -> Random.shuffle!(Random.GLOBAL_RNG, x) : identity
+    end
     DataLoader(data, batchsize, n, partial, imax, [1:n;], shuffle)
 end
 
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
This is an admittedly late-isa response to https://github.com/FluxML/Flux.jl/pull/1501#issuecomment-774051647

CarloLucibello:
I think the interface in #1501 is much more convenient than the one proposed in this PR. I don't see any downside to #1501 (storing internally an rng is perfectly fine), has a better interface and is also backward-compatible.

btw, this or is switching current default to shuffle=false, so it is breaking

darsnack:
I like that this PR allows an alternate "shuffling" iterator, but I agree that the interface is more cumbersome than #1501. `shuffle!` itself provides a pretty powerful interface to alternate shuffling mechanisms, so I feel you can gain most of the flexibility from this design just by specifying the RNG correctly.

Of course, there is non-shuffle-like iteration that you can do with this PR that you can't with #1501, but that feels like abusing the `shuffle` keyword for alternate purposes. To a certain extent, I agree with the comments that the eventual data loader should have a more extensible interface built on composable iterators. I would suggest that effort is better spent refactoring MLDataPattern.jl than trying to design the same package within Flux.

darsnack:
If it wasn't clear, I also think it makes more sense to have #1501. Taking a look at Dataloaders.jl, we can probably keep the current `Flux.Dataloader` interfaces and use MLDataPattern.jl under the hood eventually. This will make for a smooth transition (probably even non-breaking). If we move to a non-Boolean `shuffle` keyword, then it would make that change breaking.

DhairyaLGandhi:
The current default can be retained. The pr already also allows for the Booleans, so the API concern should be largely addressed. I don't think it's too much of an abuse because varying shuffling policies technically only needs to happen once at the start, and all permutations are shuffles too

darsnack:
> I don't think it's too much of an abuse because varying shuffling policies technically only needs to happen once at the start, and all permutations are shuffles too

Right, that's my point. Most of the time you could just use `shuffle!` anyways, so there isn't a lot of value in specifying a closure over it. The potential value you could get is if you do something that isn't a permutation, in which case I would consider that abuse.

> The pr already also allows for the Booleans, so the API concern should be largely addressed

I was saying that by allowing a function in addition to booleans, we are creating a new interface that will be more difficult to make backwards compatible if/when we transition to MLDataPattern.jl. Keeping it restricted to booleans means a less complex interface to support through the transition.

DhairyaLGandhi:
> if/when we transition to MLDataPattern.jl.

I am uncomfortable to think that that should be a reason to hold back on this. there's no such transition planned.

darsnack:
I agree that it shouldn't hold us back at all. But in this case, there is #1501 that gets us almost everything here. IMO the additional functionality gained with this PR isn't worth it even if MLData* isn't officially planned. It doesn't even have to be w.r.t. MLData*, the boolean-only interface is much easier to keep backwards compatible than arbitrary functions no matter what we move forward with in the future.

CarloLucibello:
We should hold back on this just because it's an ugly interface. It is actually already quite easy to do custom shuffling (the absence of which no one ever lamented btw). 

```julia
train_loader = DataLoader((X, y))   # shuffle=false by default
for epoch in 1:epochs
  customshuffle!(train_loader.indices)
  ...
end
```
We can document this pattern if we really feel there is such need. I prefer to commit to the presence of a `indices` field rather than the interface proposed in this pr.

DhairyaLGandhi:
> Most of the time you could just use shuffle! anyways, so there isn't a lot of value in specifying a closure over it

You can't, because the current system expects a bool. The closure is to let you add the RNG without storing it explicitly. It would be good not to keep increasing the list of fields with orthogonal functionality to what the struct is intended to be used for.

> because it's an ugly interface.

It's the same as can be seen from the reference. You can extend it but it's the same otherwise as was done originally.

> absence of which no one ever lamented

Right, but #1501 is an example.

darsnack:
> You can't, because the current system expects a bool.

Yes, but with #1501 you could just use the RNG that give you the permutation specification you want. Or you can do @CarloLucibello's snippet.

> It would be good not to keep increasing the list of fields with orthogonal functionality to what the struct is intended to be used for.

I agree, but fields are cheaper than keywords in the interface. Ideally, most people shouldn't be accessing a struct's fields directly. It's not really so expensive to store the RNG. The global RNG is a reference anyways, and if someone is closing over `shuffle!(custom_rng, ...)`, then it is being allocated and stored somewhere. What's so bad about storing it in the struct's fields? Passing in an RNG is intuitive in Julia and other languages.

> Right, but #1501 is an example.

It's an example of wanting a custom RNG. A little bit different wanting to use the custom RNG with a function other than `shuffle!`.

CarloLucibello:
can we just merge #1501?

CarloLucibello:
can close this since #1501 was mergerd

CarloLucibello:

Issue #1027, opened 2020-02-08
`trainable` for BatchNorm stops parameters from being saved and loaded


xukai92:
Below is defined to only take gradient of the `β` and `γ` in batch norm layers.
```julia
trainable(bn::BatchNorm) = (bn.β, bn.γ)
```
However, this stops us from using `params` and `loadparams!` to save and load parameters as the other two fields, `μ` and `σ²`, which are updated during training as well, to be saved and loaded.

Maybe it's just fine to not define `trainable(bn::BatchNorm) = (bn.β, bn.γ)` as `μ` and `σ²` doesn't seems to have gradient?

CarloLucibello:
Maybe is fine to add  μ and σ² to the trainable parameters, since they are used only at test time and they have zero gradient during training.
A more general solution would be to add a flag to `params` to also return non-trainable fields.

xukai92:
> Maybe is fine to add μ and σ² to the trainable parameters, since they are used only at test time and they have zero gradient during training.

Yes that's what I mean ... don't know why I typed "no" even I said "doesn't` in the end of the setence...

CarloLucibello:
@xukai92 what's your specific use case? can you work around it with bson saving and loading?

nantonel:
Apparently, if you save your model weights:
```julia
julia> weights = params(model);

julia> using BSON: @save

julia> @save "mymodel.bson" weights
```
the batchnorm parameters are lost.

ToucheSir:
We should probably advocate against saving implicit `Params` in favour of saving explicit params or the model itself. Flat params are just too brittle and are easily invalidated with architectural changes.

darsnack:
I suppose we should think more comprehensively about how to do this. It has come up in https://github.com/FluxML/MetalheadWeights/pull/2. On one hand, Metalhead should just save the model itself and return that instead copying over the weights. On the other hand, this makes it possible to silently return a different model with `pretrain = true` or `pretrain = false`. Hopefully, we would never make a release that allows that to happen, but it has the code smell of a poor design.

Ideally, the saving mechanism should allow users to specify state that isn't trainable to be saved. And the loading mechanism should error when the architectures don't match. e.g. maybe we should be doing a structural walk of the model instead of a flat iteration over the parameters.

ToucheSir:
That all makes sense to me. I think it (mostly the structural walk) might even be possible to do with what's currently in Optimisers.jl.

[...]

PR #1509, opened 2021-02-16
Fix Norm Layers, Again

DhairyaLGandhi:
This is a successor to #1495 

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

Todo:

- [x] Go through the tests, we don't want to use the autodiff test file as it is, doing more straightforward checks is.better.
- [x] Remove `hasaffine`
- [ ] remove `testmode!` - use zygote instead and have an equivalent forward pass function available a la `dropout`
- [x] determine forward pass at compile time


test/cuda/layers.jl
```diff
@@ -16,23 +16,20 @@ end
...
-          if setmode
-            testmode!(l_cpu)
-            gpu_autodiff_test(l_cpu, x_cpu, 
-              test_equal=test_cpu, rtol=rtol, atol=atol)
-          end  
+          ps_gpu = Flux.params(l_gpu)
+          ps_cpu = Flux.params(l_cpu)
+          y_gpu, back_gpu = pullback(() -> sum(l_gpu(x_gpu)), ps_gpu)
+          gs_gpu = back_gpu(1.f0)
+          @test gs isa Flux.Zygote.Grads
```
CarloLucibello, line 32: I really don't understand why you put effort in undoing other people's work, lowering the test coverage and adding code duplication, while having a centralized and exhaustive testing function is clearly a better choice.

CarloLucibello:
>  remove testmode! - use zygote instead and have an equivalent forward pass function available a la dropout
 this TODO has nothing to do with this PR.

darsnack:
And the [last time it was dicussed](https://github.com/FluxML/Flux.jl/pull/1301) there was not a complete consensus. I think it is better not to put this TODO on the v0.12 path.

DhairyaLGandhi:
Still not perfect, but better than master. We would need to figure removing the unions as well.

This:
```Julia
julia> bn = BatchNorm(10)
BatchNorm(10, identity, affine = )

julia> x = rand(Float32, 3,3,3,10);

julia> @code_warntype bn(x)
Variables
  BN::BatchNorm{typeof(identity),Array{Float32,1},Float32,Array{Float32,1}}
  x::Array{Float32,4}
  #334::Flux.var"#334#335"{Array{Float32,4},Int64}
  @_4::Int64
  N::Int64
  reduce_dims::Array{Int64,1}
  affine_shape::Union{Nothing, NTuple{4,Int64}}
  ts::Union{Nothing, Bool}
  μ::Array{Float32,4}
  σ²::Array{Float32,4}
  @_11::Union{Nothing, NTuple{4,Int64}}
  @_12::Union{Nothing, Bool}

Body::Array{Float32,4}
1 ──       Core.NewvarNode(:(#334))
│          Core.NewvarNode(:(@_4))
│          Core.NewvarNode(:(reduce_dims))
│          Core.NewvarNode(:(affine_shape))
│          Core.NewvarNode(:(ts))
│          Core.NewvarNode(:(μ))
│          Core.NewvarNode(:(σ²))
│          (N = Flux.ndims(x))
[...]
```

Master:

```julia
julia> bn = BatchNorm(10)
BatchNorm(10)

julia> x = rand(Float32, 3,3,10,10);

julia> bn(x);

julia> @code_warntype bn(x)
Variables
  BN::BatchNorm{typeof(identity),Array{Float32,1},Float32,Array{Float32,1}}
  x::Array{Float32,4}
  #239::Flux.var"#239#240"{Array{Float32,4},Int64}
  N::Int64
  reduce_dims::Array{Int64,1}
  affine_shape::NTuple{4,Int64}

Body::Any
1 ─       Core.NewvarNode(:(#239))
│         Core.NewvarNode(:(N))
│         Core.NewvarNode(:(reduce_dims))
│         Core.NewvarNode(:(affine_shape))
│   %5  = Flux.ndims(x)::Core.Compiler.Const(4, false)
│   %6  = (%5 - 1)::Core.Compiler.Const(3, false)
│   %7  = Flux.size(x, %6)::Int64
│   %8  = Base.getproperty(BN, :chs)::Int64
│   %9  = (%7 == %8)::Bool
```

darsnack:
Based on our discussion yesterday, this PR looks ready except the plans for `active`. Can we agree to save those for another release, since master currently makes no changes to the `active`/`trainmode`/`testmode` API?

Other than that looks like we only need tests and docs.

src/layers/normalise.jl
```diff
@@ -245,40 +278,42 @@ mutable struct BatchNorm{F,V,N,W}
...
+                   initγ = i -> ones(Float32, i), 
+                   affine = true, track_stats = true,
+                   ϵ = 1f-5, momentum = 0.1f0)
 
-  β = affine ? initβ(chs) : nothing
-  γ = affine ? initγ(chs) : nothing
-  μ = track_stats ? zeros(Float32, chs) : nothing
-  σ² = track_stats ? ones(Float32, chs) : nothing
+  β = initβ(chs)
+  γ = initγ(chs)
```
CarloLucibello, line 288: why we should have params that we don't use in practice?

src/layers/normalise.jl
```diff
@@ -245,40 +278,42 @@ mutable struct BatchNorm{F,V,N,W}
...
 
   return BatchNorm(chs, λ, β, γ,
-            μ, σ², ϵ, momentum, 
+            μ, σ², ϵ, momentum,
             affine, track_stats, nothing)
 end
 
 @functor BatchNorm
-trainable(bn::BatchNorm) = hasaffine(bn) ? (bn.β, bn.γ) : ()
+# trainable(bn::BatchNorm) = hasaffine(bn) ? (bn.β, bn.γ) : ()
```
CarloLucibello, line 298: if you remove this μ, σ² will end up in the params

src/layers/normalise.jl
```diff
@@ -245,40 +278,42 @@ mutable struct BatchNorm{F,V,N,W}
...
+                   ϵ = 1f-5, momentum = 0.1f0)
 
-  β = affine ? initβ(chs) : nothing
-  γ = affine ? initγ(chs) : nothing
-  μ = track_stats ? zeros(Float32, chs) : nothing
-  σ² = track_stats ? ones(Float32, chs) : nothing
+  β = initβ(chs)
+  γ = initγ(chs)
+  μ = zeros(Float32, chs)
+  σ² = ones(Float32, chs)
```
CarloLucibello, line 290: same as above, when `track_stats=false` these are not used, shouldn't be set to nothing?

src/layers/normalise.jl
```diff
@@ -245,40 +278,42 @@ mutable struct BatchNorm{F,V,N,W}
...
 function (BN::BatchNorm)(x)
-  @assert size(x, ndims(x)-1) == BN.chs
-  N = ndims(x)
+  N = ndims(x)::Int
+  @assert size(x, N - 1) == BN.chs
   reduce_dims = [1:N-2; N]
-  affine_shape = ntuple(i -> i == N-1 ? size(x, N-1) : 1, N)
-  return _norm_layer_forward(BN, x; reduce_dims, affine_shape)
+  nc = NormConfig(BN.affine, BN.track_stats, reduce_dims)
+  @show nc
```
CarloLucibello, line 305:
```suggestion
```

test/cuda/layers.jl
CarloLucibello, line 32: the `gpu_autodiff_test` that you removed was testing that cpu and gpu matched on: forward pass, gradient with respect to params, gradients with respect to inputs. You are testing none of that anymore. Could you just leave `gpu_autodiff_test` there?

test/cuda/layers.jl
```diff
@@ -80,19 +77,18 @@ gpu_gradtest("GroupNorm 2d", groupnorm, rand(Float32, 8, 8, 12, 4), 12, 3, setmo
...
-
-pixelshuffle = [PixelShuffle]
-gpu_gradtest("PixelShuffle 2d", pixelshuffle, rand(Float32, 3, 4, 18, 3), 3)
-gpu_gradtest("PixelShuffle 1d", pixelshuffle, rand(Float32, 3, 18, 3), 3)
+ gpu_gradtest("Upsample 2d", upsample, rand(Float32, 3, 4, 2, 3), (2,2))
+ gpu_gradtest("Upsample 1d", upsample, rand(Float32, 3, 4, 2, 3), (2,))
 
+ pixelshuffle = [PixelShuffle]
+ gpu_gradtest("PixelShuffle 2d", pixelshuffle, rand(Float32, 3, 4, 18, 3), 3)
+ gpu_gradtest("PixelShuffle 1d", pixelshuffle, rand(Float32, 3, 18, 3), 3)
```
CarloLucibello, line 85: there is a weird one space indent here

CarloLucibello:
I'll recap what this PR when ready should **not** change with respect to master:
- the constructor interface `BatchNorm(3, affine=true, track_stats=false)` and its default values
- the behaviour in the forward pass
- the layers' parameters:
```julia
julia> params(BatchNorm(2))

julia> params(BatchNorm(2, affine=false))
Params([])
```

So, since there won't be any API changes with respect to master (unless I'm missing something), we can consider this non-blocking for v0.12 release

DhairyaLGandhi:
How does one change the behaviour of a layer at different points of training? Say I don't want to perform `affine` transforms initially, but want to later. This params change would hinder that without adding extra boilerplate code on the user end and potentially extra copies slowing down the entire process unnecessarily. This breaks the contract of eager execution.

How do I also restart training from a given point if I don't serialise the state of the tracking variables at such a time? They would not be trained either way, so the user choice is respected.

CarloLucibello:
> How does one change the behaviour of a layer at different points of training? Say I don't want to perform `affine` transforms initially, but want to later. This params change would hinder that without adding extra boilerplate code on the user end and potentially extra copies slowing down the entire process unnecessarily. This breaks the contract of eager execution.

It is not clear what you say in the last two sentences, but the thing is that `params` function should return a meaningful result, that is whatever are meant to be a layer's trainable parameters. If you declare  `m = BatchNorm(2, affine=false)` then `params(m)` should be empty, anything else would be surprising and would make `params` meaningless.    

The use case you mention can be simply achieved by creating the layer `affine=true` and then freezing the parameters in any
part of the training you want the same way you would do it with any set of parameters.  

> How do I also restart training from a given point if I don't serialise the state of the tracking variables at such a time? They would not be trained either way, so the user choice is respected.

Again, `params(m)` should return trainable parameters not anything in a module. Serialization has nothing to do with `params`. 

PyTorch handles the separation between trainable parameters and the rest of the internal state very neatly:

```python
In [15]: import torch.nn as nn

In [16]: m = nn.BatchNorm2d(2, affine=True)

In [17]: list(m.parameters())
Out[17]: 
[Parameter containing:
 tensor([1., 1.], requires_grad=True),
 Parameter containing:
 tensor([0., 0.], requires_grad=True)]

In [18]: m = nn.BatchNorm2d(2, affine=False)

In [19]: list(m.parameters())
Out[19]: []

In [20]: m = nn.BatchNorm2d(2, track_running_stats=True)

In [21]: list(m.buffers())
Out[21]: [tensor([0., 0.]), tensor([1., 1.]), tensor(0)]

In [22]: m = nn.BatchNorm2d(2, track_running_stats=False)

In [23]: list(m.buffers())
Out[23]: []
```

darsnack:
> If you declare m = BatchNorm(2, affine=false) then params(m) should be empty, anything else would be surprising and would make params meaningless.

Just to clear up confusion, did you mean `trainable` here?

DhairyaLGandhi:
```julia
# if I want affine = false and then true
m = Chain(Conv(...), BatchNorm(N, affine = true))

# train
```
How do I now make this train as if `affine = false`? Say there is a `setaffine!` function.

```julia
function setaffine!(model::Chain, flag = true)
  # requirement -> does not make copies to the original model
  
end
```

while something like this is far more in line with how the API reads
```Julia

model = BatchNorm(N, affine = false)

# train a bit with affine = false

model.affine = true # setaffine! could just do this now - although its ugly to have such a frontend because its not obvious what it does with any other kind of layer.  TLDR `setaffine!(::Conv, ...)` should error, not return the Conv back.

# train some more with affine = true
```

> The use case you mention can be simply achieved by creating the layer affine=true and then freezing the parameters in any part of the training you want the same way you would do it with any set of parameters.

This is the kind of boilerplate that I was referring to.

What about `params` doesn't have to do with Serialization? I was under the impression that saving the state of the model was done via this method, because it guarantees reproducibility. How is that handled now?

CarloLucibello:
> Just to clear up confusion, did you mean `trainable` here?

same, `params` relies on `trainable` or falls back to `functor`

CarloLucibello:
> while something like this is far more in line with how the API reads

if you define a model with a batchnorm layer with `affine=false` and then you want to switch on affine, you are just changing the model so you have to insert the new layer (no need to copy weights or stats, just create with the old ones). This is no different than creating a dense with `bias=false` and then wanting to switch on the layer, you don't just do `dense.bias = true`. 

Your use case is no more common, and actually a specific case, of the generic scenario where you would use parameter freezing. If you think the current interface for that is too clumsy (on which I could agree) we should think of a general solution,
not the specific one you suggest that also has some undesirable side effects. 
In fact, I can think of very common use cases such as
```julia
num_params = sum(length, params(m))
L2norm = sum(p -> sum(abs2, p), params(m))
```
where the changes you suggest, throwing in things that are not supposed to be trainable params, would lead to erroneous 
results accounting for spurious stuff. 

> What about `params` doesn't have to do with Serialization? I was under the impression that saving the state of the model was done via this method, because it guarantees reproducibility.

I was referring to bson saving the entire model, which is the only reliable way we have right now to serialize models.
The `saveparams` and `loadparams` interface is just buggy instead. We have issue #1027 on v0.11 and master, 
but contrary to what I suggested in that issue, we should not let spurious stuff into params, but introduce `savestate` and `loadstate` accounting for the whole state of the model, not only trainable parameters.
See also the `state_dict` approach in pytorch
https://pytorch.org/tutorials/beginner/saving_loading_models.html

DhairyaLGandhi:
I would thank you to stay pertinent to this pr. I still haven't been answered.

The regularisation function should pick up `trainable` instead. So that's incorrect. You want to normalise learnable entities.

> This is no different than creating a dense with bias=false and then wanting to switch on the layer, you don't just do dense.bias = true.

It is in fact different. Because doing that to dense is not a permissible operation (at least last I checked). It's these kinds of clean looking but opaque API decisions that cause confusion, and lead to hacky workarounds. Ultimately hurting the usability.

The loadparams is doing precisely the thing it was designed for. What has changed is the underlying representation of the data, but without accounting for the consistency that was part of the system prior.

Also this isn't a typical subset of parameter freezing since there are extra operations involved, something that cannot be accounted for in parameter freezing.

DhairyaLGandhi:
Also chains are immutable

CarloLucibello:
> I would thank you to stay pertinent to this pr. I still haven't been answered.

??? all of my previous comment was about answering your questions

>  The regularisation function should pick up trainable instead. So that's incorrect. You want to normalise learnable entities.

which is exactly what it does since it uses `params`, and that's why you shouldn't remove 
```julia
trainable(bn::BatchNorm) = hasaffine(bn) ? (bn.β, bn.γ) : ()
```
as you did here. As this PR stands, `params(bn)` returns affine parameters and tracking statistics which is very wrong to start with, and especially wrong since it does it also when you select `affine=false` and  `track_stats=false`.

> > This is no different than creating a dense with bias=false and then wanting to switch on the layer, you don't just do dense.bias = true.
> It is in fact different. Because doing that to dense is not a permissible operation (at least last I checked). It's these kinds of clean looking but opaque API decisions that cause confusion, and lead to hacky workarounds. Ultimately hurting the usability.

It is no different at all, same way once you do `Dense(..., bias=false)` you cannot change it later, you do `BatchNorm(..., affine=false)` and you cannot change it later. How are they different? It's not a permitted operation for both.

What makes people confused, actually even worse, what is a source of bugs, is what you are suggesting here,  removing
`trainable(bn::BatchNorm) = hasaffine(bn) ? (bn.β, bn.γ) : ()`. Essentially with this PR people applying L2reg on all params will be regularizing the tracked statistics (very bad) and some parameters that are not even supposed to exists if `affine=false` (mildly bad). 

Freezing parameters is not a workaround, is the thing to do. 

I think I gave plenty of arguments, in the end it boils down to what `params` should do. I think it should return only trainable parameters, and honestly seems unreasonable to argue it should do anything else since it we tell people to do
`gradient(..., params(m))`. I also gave two very common usage cases, parameter counting and regularization, where
doing anything else would introduce bugs.   

> The loadparams is doing precisely the thing it was designed for. What has changed is the underlying representation of the data, but without accounting for the consistency that was part of the system prior.

`loadparams` is just not well thought of, or at least just a part of the story, because the `params` interface is about trainable parameters, while to properly restore the state of a generic model you need more than that (i.e. the full state in fact).

> Also this isn't a typical subset of parameter freezing since there are extra operations involved, something that cannot be accounted for in parameter freezing.

Again, same as `Dense(..., bias=false)`. So if you think this is a relevant issue (which I think is not) open another PR where you address the issue generally and without hurting the behavior of `params`.

> Also chains are immutable

immutability doesn't mean you can't do `chain.layers[i] = ...`

ToucheSir:
> See also the state_dict approach in pytorch
> https://pytorch.org/tutorials/beginner/saving_loading_models.html

Though I'm not convinced their design is anywhere close to optimal, separating the concept of trainable parameters from auxiliary state (as handled by `register_parameter` and `register_buffer` respectively) is a good idea. From the Flux side of things, this would probably require decoupling `trainable` from `params`. We'd also want some mechanism for handling this in Functors as well to prevent overwriting `trainable` for every layer where trainable params != state, but I'm not clear on what that would look like.

darsnack:
`trainable` exists to allow specifying the trainable leaves separately from all the leaves (as recognized by `@functor`). `params` is a user facing function for building a vector of parameters via calls to `trainable`. In this way, `trainable` is more an interface for writers of layers, and `params` is the function for users of layers.

I don't know if we need to separate `trainable` from `params` as much as we need a totally new function to return the other subset of leaves that aren't trainable (i.e. `setdiff(functor(m)[1], trainable(m)`).

darsnack:
Or I guess maybe your point is that `trainable` is unnecessarily coupled to `functor` right now?

ToucheSir:
Kind of a mix of the two. `params` is named too generally and should probably be called `trainable_params` or something if we're trying to distinguish between trainable params and layer state (the former being a subset of the latter). 90% of the headaches around this (e.g. `loadparams!`) would probably disappear if everyone used explicit params, but realistically implicit params aren't going anywhere any time soon.

DhairyaLGandhi:
They are conceptually similar, but regularisation is just a case where trainable is more appropriate.

The state dict for us is saving the model directly, which encapsulates all the various config for a model. It does so generically, and I'm not sure there's much benefit to adding other things. But in many cases, we also want to make sure that some config is overwritten, like ignore the bias config of the model if the params suggest that it needs a Zeros; ignore affine if the loading params suggest there are trainable params etc.

ToucheSir:
>  It does so generically, and I'm not sure there's much benefit to adding other things

I agree for the most part, but there's definitely some utility in being able to load weights without having the model code loaded as well (much like the namedtuple you get when taking explicit gradients). That's probably a discussion for another thread through.

The bigger issue here is that `loadparams` and `params` do not and should not be working with the same thing. The former expects a `state_dict`-like structure, while the latter is fine with just `trainable`. The only way I can see to resolve this now is a) add a function `actually_all_params` that does what it says on the tin, b) use that function in `loadparams!` instead of https://github.com/FluxML/Flux.jl/blob/e2b73e142067a99e88cc393c3fdd51b1e2029cf2/src/functor.jl#L58, and c) manually define `trainable(l::Layer)` for any layer where `trainable(l) != functor(l)[1]`. Suggestions for less janky approaches that accomplish the same goal would be much appreciated :)

test/cuda/layers.jl
DhairyaLGandhi: Handled already in #1472

src/layers/normalise.jl
```diff
@@ -321,41 +354,40 @@ mutable struct InstanceNorm{F,V,N,W}
...
+  μ = zeros(Float32, chs)
+  σ² = ones(Float32, chs)
+  InstanceNorm(chs, λ, β, γ,
             μ, σ², ϵ, momentum, 
             affine, track_stats, nothing)
 end
 
 @functor InstanceNorm
-trainable(in::InstanceNorm) = hasaffine(in) ? (in.β, in.γ) : ()
+trainable(in::InstanceNorm) = (in.β, in.γ)
```
CarloLucibello, line 373: This has been said in any possible way: if a layer is created with `affine=false`, the affine parameters shouldn't exist at all and for sure they should be collected in `params`. The case of `bias=false` for `Dense` is  a good example of that:
```julia
julia> Dense(2,3,bias=false)
Dense(2, 3; bias=false)

julia> params(m) # no bias here
Params([Float32[-0.37812215 -0.3036801; 0.8392983 0.26071376; 0.65674675 -0.4557534]])
``

Comment (darsnack) [reply src/layers/normalise.jl:373]: Instead of entering this argument again, can we just remove this change and agree to punt it to another PR? It is not necessary for the other changes in this PR, there is clear disagreement on what's right, so it seems like a good idea that this change deserves its own PR.

Comment (DhairyaLGandhi) [reply src/layers/normalise.jl]: The idea behind Zeros is to have something more widely applicable than shown here. In its proper implementation, it explicitly says we are willingly eliminating any computation with this field (as opposed to not having any way to back track how the model was intended to be used) and must be part of the params. That is the long term solution imo.

Comment (darsnack) [reply src/layers/normalise.jl]: Yes, I could see generic notion of Zeros being an option for "turn-off-able" parameters like the affine parameters or the bias in dense. I would be interested in discussing/iterating it in a future PR whenever one of us has time to put it together.

Comment (DhairyaLGandhi) [reply src/layers/normalise.jl]: I'll do it.

File: test/cuda/layers.jl (hunk 6 above)
Comment (darsnack) [reply test/cuda/layers.jl:85]: I think the indent is still here?

File: test/layers/normalisation.jl [hunk 8]
```diff
@@ -11,32 +11,34 @@ evalwgrad(f, x...) = pullback(f, x...)[1]
...
   m = Dropout(0.9)
-  y = evalwgrad(m, x)
-  @test count(a->a==0, y) > 50
+  y = m(x)
+  # By default no dropout is performed outside training
+  # @test count(a -> a == 0, y) > 50
   testmode!(m, true)
-  y = evalwgrad(m, x) # should override istraining
-  @test count(a->a==0, y) == 0
+  y = m(x) # should override istraining
```
darsnack, line 18: These changes don't seem right. `evalwgrad` tests whether the `active` keyword is working correctly. We should keep these tests in until we change `active` in another PR.

test/layers/normalisation.jl
```diff
@@ -83,149 +81,147 @@ end
...
     @test isapprox(x′[1], (1 .- 0.3) / sqrt(1.3), atol = 1.0e-5)
   end
 
   # with activation function
-  let m = BatchNorm(2, sigmoid), x = [1.0 3.0 5.0;
-                                      2.0 4.0 6.0]
+  let m = trainmode!(BatchNorm(3, sigmoid)), x = reshape(1:6, 1,1,3,2)
     y = m(x)
-    @test isapprox(y, sigmoid.((x .- m.μ) ./ sqrt.(m.σ² .+ m.ϵ)), atol = 1.0e-7)
+    @test_broken isapprox(y, mean(sigmoid.((x .- m.μ) ./ sqrt.(m.σ² .+ m.ϵ)), dims = 1), atol = 1.0e-7)
```
darsnack, line 102: Why the change to `@test_broken`?

test/layers/normalisation.jl
```diff
@@ -83,149 +81,147 @@ end
...
+  #   m(x)
+  #   @test (@allocated m(x)) <  100_000_000
+  # end
 
-  @test length(Flux.params(BatchNorm(10))) == 2
-  @test length(Flux.params(BatchNorm(10, affine=true))) == 2
-  @test length(Flux.params(BatchNorm(10, affine=false))) == 0
+  # @test length(Flux.params(BatchNorm(10))) == 2
+  # @test length(Flux.params(BatchNorm(10, affine=true))) == 2
+  # @test length(Flux.params(BatchNorm(10, affine=false))) == 0
```
darsnack, line 130: These tests can probably be put back in since we're punting on `trainable` for another PR.

darsnack (commented):
Looks pretty good for the implementation side of things. There are some stylistic changes around `=` for keywords that I don't think are inline with the rest of the Flux code, but no technical comments from me.

I do have concerns about the tests. I think there are a couple tests that got changed, but I don't see why. Left some comments on the specific line numbers.

darsnack:
Thanks @DhairyaLGandhi for seeing this through. I think we might need a rebase too?

DrChainsaw:
At the risk of sticking my hand in a hornet nest, would this be a good time to add the possibility for different types on the betas and gammas so one can use `Zeros`?

DhairyaLGandhi:
Yeah, that's a good idea, thanks for bringing it up!

darsnack:
Isn't that already possible with this PR? Something like `Zeros` as a consistent way to have "turn-off-able" parameters seems like the right long term solution for the affine components. But I'd prefer to leave that for a later PR given that we've never come to a complete consensus on `Zeros`.

darsnack:
Btw I think a rebase and un-commenting a few tests are all that are all that is needed to merge this.

darsnack:
Bump

src/layers/normalise.jl
darsnack: Other thought on this: I know we rejected `Affine(BatchNorm(...))` because it is not ergonomically nice. But what about having a field in `BatchNorm` called `affine` that is a function. When the affine kwarg is true, then `affine = Dense(...)` but when it is false, `affine = identity`? In a `Chain`, the user still see just `BatchNorm` and `bn.affine` will show a `Dense` which the user will intuitively understand. We also don't end up storing data we don't need, and `trainable` will automatically be empty.

ToucheSir: Kind of like you'd have for an activation function? That does sound pretty appealing, would it make sense to call that field `transform` or something then?

darsnack: Yeah exactly. `transform` would certainly capture the field accurately, but `affine` might be better here cause people will understand it.

DhairyaLGandhi:
The fix for CI is probably in NNlibCUDA, unless we do a manual reshape here.

src/cuda/cudnn.jl
```diff
@@ -1,14 +1,13 @@
 import NNlibCUDA: batchnorm, ∇batchnorm
 
-function (BN::Flux.BatchNorm)(x::Union{CuArray{T,2},CuArray{T,4},CuArray{T,5}},
-                              cache=nothing) where T<:Union{Float32, Float64}
+function (BN::Flux.BatchNorm)(x::CuArray{T},
+                              cache = nothing) where T<:Union{Float32, Float64}
```
darsnack, line 4: While I like the space around `=`, this isn't in-line with the rest of the code style in Flux, so it should be changed back.

src/layers/normalise.jl
```diff
@@ -129,77 +130,109 @@ The input is normalised along the first `length(sz)` dimensions
...
-  x = normalise(x, dims=1:length(a.size), ϵ=a.ϵ)
-  a.diag === nothing ? a.λ.(x) : a.λ.(a.diag(x))
+  x = normalise(x, dims = 1:length(a.sz), ϵ = a.ϵ)
+  a.λ.(a.diag(x))
 end
 
 function Base.show(io::IO, l::LayerNorm)
   print(io, "LayerNorm($(l.size)")
-  l.λ == identity || print(io, ", $(l.λ)")
-  hasaffine(l) || print(io, ", affine=false")
```
darsnack, line 162: Feel like it would nice to show something like `"affine = $(l.diag)"` so a user can easily see what affine transformation is being applied.

src/layers/normalise.jl
```diff
@@ -129,77 +130,109 @@ The input is normalised along the first `length(sz)` dimensions
...
-        μnew = vec(N ∈ reduce_dims ? μ : mean(μ, dims=N))
-        σ²new = vec(N ∈ reduce_dims ? σ² : mean(σ², dims=N))
-        l.μ = (1-mtm) .* l.μ .+ mtm .* μnew
-        l.σ² = (1-mtm) .* l.σ² .+ mtm .* (m / (m - one(eltype(l.σ²)))) .* σ²new
-      end
+    μ = mean(x; dims = nc.dims)
+    σ² = mean((x .- μ) .^ 2; dims = nc.dims) # ./ l.chs
+
+    μnew, σ²new = track_stats(x, (l.μ, l.σ²), (μ,σ²),
+                        l.momentum, reduce_dims = nc.dims)
```
darsnack, line 189: Tab over to line up args?

src/layers/normalise.jl
```diff
@@ -245,50 +278,50 @@ mutable struct BatchNorm{F,V,N,W}
...
+  affine(BN, x, μ, σ², nc)
 end
 
 testmode!(m::BatchNorm, mode=true) =
   (m.active = (isnothing(mode) || mode == :auto) ? nothing : !mode; m)
 
 function Base.show(io::IO, l::BatchNorm)
   print(io, "BatchNorm($(l.chs)")
-  l.λ == identity || print(io, ", $(l.λ)")
-  hasaffine(l) || print(io,  ", affine=false")
```
darsnack, line 282: Same as other comment

src/layers/normalise.jl
```diff
@@ -403,20 +435,20 @@ mutable struct GroupNorm{F,V,N,W}
...
-function GroupNorm(chs::Int, G::Int, λ=identity;
-              initβ = (i) -> zeros(Float32, i), 
-              initγ = (i) -> ones(Float32, i), 
-              affine=true, track_stats=false,
-              ϵ=1f-5, momentum=0.1f0) 
+function GroupNorm(chs::Int, G::Int, λ = identity;
+              initβ = i -> zeros(Float32, i), 
+              initγ = i -> ones(Float32, i), 
+              affine = true, track_stats = false,
+              ϵ = 1f-5, momentum = 0.1f0)
```
darsnack, line 444: Tab over

test/layers/normalisation.jl
darsnack: This still holds

test/layers/normalisation.jl
darsnack: Bump here

test/layers/normalisation.jl
darsnack: Bump here

test/layers/normalisation.jl
```diff
@@ -83,149 +81,147 @@ end
...
   # with activation function
-  let m = InstanceNorm(2, sigmoid; affine=true, track_stats=true), sizes = (3, 2, 2),
+  let m = InstanceNorm(2, sigmoid; affine = true, track_stats = true), sizes = (3, 2, 2),
       x = reshape(collect(1:prod(sizes)), sizes)
     x = Float64.(x)
     affine_shape = collect(sizes)
     affine_shape[[1,3]] .= 1
 
-    y = evalwgrad(m, x)
+    y = m(x)
```
darsnack, line 185: Should be changed back

test/layers/normalisation.jl
```diff
@@ -237,37 +233,28 @@ end
...
-    m(x)
-    @test (@allocated m(x)) <  100_000_000
-  end
-
-  @test length(Flux.params(InstanceNorm(10))) == 0
-  @test length(Flux.params(InstanceNorm(10, affine=true))) == 2
-  @test length(Flux.params(InstanceNorm(10, affine=false))) == 0
+  # @test length(Flux.params(InstanceNorm(10))) == 0
+  # @test length(Flux.params(InstanceNorm(10, affine = true))) == 2
+  # @test length(Flux.params(InstanceNorm(10, affine =false))) == 0
```
darsnack, line 238: Should be changed back

darsnack (changes requested):
There's still some issues related to changes to the tests which aren't clear.

src/cuda/cudnn.jl
DhairyaLGandhi: Well, Flux does implement the spaces, I think better to clean up the places that don't.

test/layers/normalisation.jl
DhairyaLGandhi: I am not sure this makes sense. We want to test it as it is for the most part.

test/layers/normalisation.jl
DhairyaLGandhi: We aren't guaranteeing this.

src/layers/normalise.jl
DhairyaLGandhi: its a shame that we don't just have an clearer constructor rather than the `affine` as a kwarg. its much more informative to show the users what's actually inside their layer.

src/cuda/cudnn.jl
darsnack: A current search on master reveals the opposite. Either way, I think this PR should stay consistent with what's already there. Another PR can clean up the `=` across the repo.

image

src/layers/normalise.jl
darsnack: Agreed, but let's just save that for another PR. This PR already takes us a step closer to being better on the norm layers, so it would be nice not to prolong it.

test/layers/normalisation.jl
DhairyaLGandhi: I think this is a tolerance thing, this PR doesn't change the behaviour though.

test/layers/normalisation.jl
DhairyaLGandhi: We don't want to make guarantees around this.

test/layers/normalisation.jl
darsnack: This test was specifically written to test the behavior of the layer when it is used in a gradient context. By changing to `m(x)` we are no longer testing what the test was designed for.

test/layers/normalisation.jl
darsnack: A test isn't a guarantee of behavior. But this is the current behavior so it should be tested. If two weeks from now, someone borks `trainable`, then this test would accurately identify the issue which is why it exists.

test/layers/normalisation.jl
darsnack: Is there no tolerance at which it passes? It was passing before with some tolerance.

src/layers/normalise.jl
```diff
@@ -129,77 +130,111 @@ The input is normalised along the first `length(sz)` dimensions
...
-        μnew = vec(N ∈ reduce_dims ? μ : mean(μ, dims=N))
-        σ²new = vec(N ∈ reduce_dims ? σ² : mean(σ², dims=N))
-        l.μ = (1-mtm) .* l.μ .+ mtm .* μnew
-        l.σ² = (1-mtm) .* l.σ² .+ mtm .* (m / (m - one(eltype(l.σ²)))) .* σ²new
-      end
+    μ = mean(x; dims = nc.dims)
+    σ² = mean((x .- μ) .^ 2; dims = nc.dims) # ./ l.chs
+
+    μnew, σ²new = track_stats(x, (l.μ, l.σ²), (μ,σ²),
+                              l.momentum, reduce_dims = nc.dims)
```
CarloLucibello, line 191: these can go in the zygote block below

darsnack: I think this is okay since `track_stats` has `@nograd`

CarloLucibello: in which case the `@nograd` on track_stats can be removed

test/layers/normalisation.jl
darsnack: Never mind, I thought this was a different line. Here I'm not even sure what the purpose of calling `m(x)` twice is?

test/layers/normalisation.jl
```diff
@@ -83,149 +81,147 @@ end
...
   # with activation function
-  let m = BatchNorm(2, sigmoid), x = [1.0 3.0 5.0;
-                                      2.0 4.0 6.0]
+  let m = trainmode!(BatchNorm(3, sigmoid)), x = reshape(1:6, 1,1,3,2)
     y = m(x)
-    @test isapprox(y, sigmoid.((x .- m.μ) ./ sqrt.(m.σ² .+ m.ϵ)), atol = 1.0e-7)
+    @test_broken isapprox(y, mean(sigmoid.((x .- m.μ) ./ sqrt.(m.σ² .+ m.ϵ)), dims = 1), atol = 1.0e-7)
   end
 
-  let m = trainmode!(BatchNorm(2)), x = reshape(Float32.(1:6), 3, 2, 1)
```
CarloLucibello, line 105: these tests shouldn't be changed

test/layers/normalisation.jl
```diff
@@ -83,149 +81,147 @@ end
...
   end
 
-  let m = BatchNorm(32), x = randn(Float32, 416, 416, 32, 1);
-    m(x)
-    @test (@allocated m(x)) <  100_000_000
-  end
+  # let m = BatchNorm(32), x = randn(Float32, 416, 416, 32, 1);
+  #   m(x)
+  #   @test (@allocated m(x)) <  100_000_000
+  # end
```
CarloLucibello, line 126: probably this was here for a reason, could try to check with git blame when it was introduced

test/layers/normalisation.jl
```diff
@@ -83,149 +81,147 @@ end
...
     μ = mean(x, dims=1)
-    σ² = var(x, dims=1, corrected=false)
+    σ² = var(x, dims=1, corrected = false)
     @test y ≈ sigmoid.((x .- μ) ./ sqrt.(σ² .+ m.ϵ))   atol=1.0e-7
   end
 
   let m = InstanceNorm(2, sigmoid), sizes = (3, 2, 2),
       x = reshape(collect(1:prod(sizes)), sizes)
-    @test Flux.hasaffine(m) == false
-    @test length(params(m)) == 0
```
CarloLucibello, line 208: this should stay

test/layers/normalisation.jl
```diff
@@ -83,149 +81,147 @@ end
...
     
     x = Float64.(x)
-    y = m(x)
-    μ = mean(x, dims=1)
-    σ² = var(x, dims=1, corrected=false)
+    y, back = pullback((m,x) -> m(x), m, x)
```
CarloLucibello, line 207: when internal changes to a layer previous test shouldn't be modified so that one can be sure that previous behavior is preserved. Normalization layer behave differently in gradient contexts, this test was testing the behaviour in a non-gradient context

CarloLucibello:
there are too many changes in tests, they should be mostly reverted, otherwise it's hard to check that the revised internals don't introduce new (or old) bugs

src/layers/normalise.jl
```diff
@@ -129,77 +130,111 @@ The input is normalised along the first `length(sz)` dimensions
...
+  σ²new = vec((N in reduce_dims) ? σ² : mean(σ², dims = N))
+  μ_ = (1 - mtm) .* μprev .+ mtm .* μnew
+  σ²_ = (1 - mtm) .* σ²prev .+ mtm .* (m / (m - one(T))) .* σ²new
+  μ_, σ²_
+end
+@nograd track_stats
+
+function affine(l, x::AbstractArray{T,N}, μ, σ², nc::NormConfig{true}; dims = N - 1) where {T,N}
+  affine_shape = getaffine(nc, size(x), dims = dims)
+  γ = reshape(l.γ, affine_shape)
```
DhairyaLGandhi, line 219: > 

- 

-

src/layers/normalise.jl
DhairyaLGandhi: well that had been a consideration. Its a bit awkward to also do `γ = reshape(l.γ, affine_shape)` but I think we can have broadcasting take care of the shape as needed by `l.λ.(γ .* x̂ .+ β)`

src/layers/normalise.jl
```diff
@@ -129,77 +130,111 @@ The input is normalised along the first `length(sz)` dimensions
...
+    μ = mean(x; dims = nc.dims)
+    σ² = mean((x .- μ) .^ 2; dims = nc.dims) # ./ l.chs
+
+    μnew, σ²new = track_stats(x, (l.μ, l.σ²), (μ,σ²),
+                              l.momentum, reduce_dims = nc.dims)
+
+    Zygote.ignore() do
+      l.μ = reshape(μnew, :)
+      l.σ² = reshape(σ²new, :)
     end
```
darsnack, line 188: Why not put this in `track_stats`?

src/layers/normalise.jl
```diff
@@ -129,77 +130,111 @@ The input is normalised along the first `length(sz)` dimensions
...
-  size::NTuple{N,Int}
-  affine::Bool
+  sz::S
 end
 
-function LayerNorm(sz, λ=identity; affine=true, ϵ=1f-5)
-  sz = sz isa Integer ? (sz,) : sz
-  diag = affine ? Diagonal(sz...) : nothing
-  return LayerNorm(λ, diag, ϵ, sz, affine)
+function LayerNorm(sz, λ = identity; affine = Diagonal(sz...), ϵ = 1f-5)
```
darsnack, line 146: This isn't quite what I was suggesting. I didn't mean change the `affine` keyword to be non-boolean. `LayerNorm` was fine as is. I meant storing `affine` as a function in `BatchNorm`, etc. similar to `LayerNorm`. The API would still be a boolean.

DhairyaLGandhi: I get that, but for the future, we would likely not want to limit kwargs to Boolean, since while it is symmetric with other libraries, it is somewhat limiting what can be expressed. Specifically a boolean allows to express two states, while an anonymous function can do more than that. Since our implementation isn't limiting what can be done with the layer, we should pass those freedoms to the end user as well. This means, we can have a compatibility layer to support booleans, but in times when it isn't worth it to write ones own layer or tweak with certain settings that aren't directly expressed with the Boolean, it's handy to pass in an anonymous function instead.

We wouldn't want the Boolean to be the only api that users can interact with. In fact, over use of it can add bloat to the code base.

darsnack: I agree completely, but the default constructor provides that flexibility? I also agree on the overuse of booleans, but for the norm layers 99.99% of users want an affine transformation (not some other function) and they just want to turn it on or off. Here, a boolean makes sense for the sake of convenience. We can still allow the advanced use case with the default constructor.

DhairyaLGandhi: The default constructor is much less ergonomic and harder to use. Not to mention requiring the user to know way more details about Flux's innards that it's untenable and a poor use of time, not to mention making the user code fragile to updates to Flux.

The fact that other libraries push boolean kwargs is why I proposed having some kind of hybrid constructor that can accept either.

darsnack: Sure, a hybrid constructor is fine by me

test/layers/normalisation.jl
DhairyaLGandhi: we've removed `hasaffine`

DhairyaLGandhi:
bumping

src/layers/normalise.jl
```diff
@@ -129,77 +130,111 @@ The input is normalised along the first `length(sz)` dimensions
...
+  # diag = affine ? Diagonal(sz...) : identity
```
ToucheSir, line 147: Per the discussion above, I believe this line needs to be brought back.

src/layers/normalise.jl
```diff
@@ -129,77 +130,111 @@ The input is normalised along the first `length(sz)` dimensions
...
+
+getaffine(nc::NormConfig{false}, args...; kwargs...) = ()
+
 # For InstanceNorm, GroupNorm, and BatchNorm.
 # Compute the statistics on the slices specified by reduce_dims.
 # reduce_dims=[1,...,N-2,N] for BatchNorm
 # reduce_dims=[1,...,N-2] for InstanceNorm and GroupNorm
-function _norm_layer_forward(l, x::AbstractArray{T,N}; reduce_dims, affine_shape) where {T, N}
-  if !_isactive(l) && l.track_stats # testmode with tracked stats
+function norm_forward(l, x::AbstractArray{T,N}, nc::NormConfig{A, true}) where {A, T, N}
```
ToucheSir, line 181: Now that all this functionality has been pulled out, we should consider moving it into NNlib next. Getting rid of https://github.com/FluxML/Flux.jl/tree/master/src/cuda (which only contains a batchnorm override) would be excellent and save us quite a few headaches.

ToucheSir (commented):
Seems like there are still a handful of comments here, so resolving those should clear up any blockers.

test/layers/normalisation.jl
ToucheSir: Blame leads back to https://github.com/FluxML/Flux.jl/pull/586. Personally I think it's a little weird to just test allocations for BatchNorm, but on the other hand this is 4 lines that run quickly and don't add any real maintenance overhead. My vote is either uncomment it or get rid of it entirely. Leaving stuff around but commented without a little explainer comment raises more questions than it answers :P

CarloLucibello:
closing as stale

PR #1511, opened 2021-02-17
Add ParameterSchedulers.jl to docs

darsnack:
As per the discussion in #1506, this adds a section to the docs that briefly describes scheduling with ParameterSchedulers.jl. Only concerns before merging are the naming conventions in ParameterSchedulers.jl. If I could get some feedback on those, then I can submit a minor release before officially merging this into the Flux docs.

### PR Checklist

- [ ] ~~Tests are added~~
- [ ] ~~Entry in NEWS.md~~
- [ ] ~~Final review from `@dhairyagandhi96` (for API changes).~~

### DIFF
--- a/docs/src/training/optimisers.md
+++ b/docs/src/training/optimisers.md
@@ -145,21 +145,24 @@ First, we import ParameterSchedulers.jl and initalize a cosine annealing schedul
 ```julia
 using ParameterSchedulers
 
-schedule = Cos(λ0 = 1e-4, λ1 = 1e-2, period = 10)
+schedule = ScheduleIterator(Cos(λ0 = 1e-4, λ1 = 1e-2, period = 10))
 opt = Momentum()
 ```
 
-Next, you can use your schedule directly in a `for`-loop like any iterator:
+Next, you can use your schedule directly in a `for`-loop:
+```julia
+for epoch in 1:100
+  opt.eta = next!(schedule)
+  # your training code here
+end
+```
+
+`schedule` can also be indexed (e.g. `schedule[100]`) or iterated like any iterator in Julia:
 ```julia
 for (eta, epoch) in zip(schedule, 1:100)
   opt.eta = eta
   # your training code here
 end
-```
-
-Alternatively, use `ScheduledOptim` from ParameterSchedulers.jl to wrap the optimiser and schedule into a single object that behaves like any Flux optimiser.
-```julia
-@epochs 100 Flux.train!(loss, ps, data, ScheduledOptim(schedule, opt))
 ```
 
 ParameterSchedulers.jl allows for many more scheduling policies including arbitrary functions, looping any function with a given period, or sequences of many schedules. See the ParameterSchedulers.jl documentation for more info.
--- a/docs/src/training/optimisers.md
+++ b/docs/src/training/optimisers.md
@@ -137,6 +137,33 @@ In this manner it is possible to compose optimisers for some added flexibility.
 Flux.Optimise.Optimiser
 ```
 
+## Scheduling Optimisers
+
+In practice, it is fairly common to schedule the learning rate of an optimiser to obtain faster convergence. There are a variety of popular scheduling policies, and you can find implementations of them in [ParameterSchedulers.jl](https://darsnack.github.io/ParameterSchedulers.jl/dev/README.html). The documentation for ParameterSchedulers.jl provides a more detailed overview of the different scheduling policies, and how to use them with Flux optimizers. Below, we provide a brief snippet illustrating a [cosine annealing](https://arxiv.org/pdf/1608.03983.pdf) schedule with a momentum optimiser.
+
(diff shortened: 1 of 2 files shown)


CarloLucibello:
Do we need to document and mention direct usage of ParameterSchedulers.jl in Flux? I thought that one of the points of #1506 was to hide ParameterSchedulers.jl usage behind a Flux.Schedule module (which is something I like)

docs/src/training/optimisers.md
```diff
@@ -137,6 +137,33 @@ In this manner it is possible to compose optimisers for some added flexibility.
...
+```julia
+for (eta, epoch) in zip(schedule, 1:100)
+  opt.eta = eta
+  # your training code here
+end
+```
+
+Alternatively, use `ScheduledOptim` from ParameterSchedulers.jl to wrap the optimiser and schedule into a single object that behaves like any Flux optimiser.
+```julia
+@epochs 100 Flux.train!(loss, ps, data, ScheduledOptim(schedule, opt))
```
CarloLucibello, line 162: This is not part of  #1434, why are we documenting it?

docs/src/training/optimisers.md
```diff
@@ -137,6 +137,33 @@ In this manner it is possible to compose optimisers for some added flexibility.
...
+```julia
+using ParameterSchedulers
+
+schedule = Cos(λ0 = 1e-4, λ1 = 1e-2, period = 10)
+opt = Momentum()
+```
+
+Next, you can use your schedule directly in a `for`-loop like any iterator:
+```julia
+for (eta, epoch) in zip(schedule, 1:100)
```
CarloLucibello, line 154: more than this interface, which I find of little convenience, the next! interface needs to be documented

darsnack: Sure will do

docs/src/training/optimisers.md
darsnack: Yeah I guess I'll leave this out. Personally, I've found this interface most convenient for my work. But we don't want to make any implications about what the interface would look like in Flux. I'll remove.

darsnack:
I also want to hide behind `Schedule`, but I can't do that without #1506. At the same time, it seemed like #1506 wasn't getting merged because of some doubts about ParameterSchedulers.jl usage with Flux code? The option presented was to document it instead of depend on it which I felt was reasonable. It would direct people towards the package, and we could get feedback without committing to it in Flux.

darsnack:
I'm okay just mentioning ParameterSchedulers.jl in the "ecosystem" section as just a bullet and removing the rest too. My impression is that if it is not explicitly mentioned with an example, then people will not know to use it. That would defeat the purpose of "vetting" the package.

docs/src/training/optimisers.md
```diff
@@ -137,6 +137,36 @@ In this manner it is possible to compose optimisers for some added flexibility.
...
+
+schedule = ScheduleIterator(Cos(λ0 = 1e-4, λ1 = 1e-2, period = 10))
+opt = Momentum()
+```
+
+Next, you can use your schedule directly in a `for`-loop:
+```julia
+for epoch in 1:100
+  opt.eta = next!(schedule)
+  # your training code here
```
CarloLucibello, line 156:
```suggestion
  # your training code here
```
```suggestion
  # your training code here
  opt.eta = next!(schedule)
```

CarloLucibello, line 156:
```suggestion
  # your training code here
  opt.eta = next!(schedule)
```

docs/src/training/optimisers.md
```diff
@@ -137,6 +137,36 @@ In this manner it is possible to compose optimisers for some added flexibility.
...
+end
+```
+
+`schedule` can also be indexed (e.g. `schedule[100]`) or iterated like any iterator in Julia:
+```julia
+for (eta, epoch) in zip(schedule, 1:100)
+  opt.eta = eta
+  # your training code here
```
CarloLucibello, line 164:
```suggestion
  # your training code here
  opt.eta = eta
```

CarloLucibello (approved):

CarloLucibello:
I took a quick look at ParametersSchedulers.jl and it looks, add just a minor naming concern for which I filed an issue. 

> I also want to hide behind Schedule, but I can't do that without #1506. At the same time, it seemed like #1506 wasn't getting merged because of some doubts about ParameterSchedulers.jl usage with Flux code?

I would prefer to merge #1506 but don't want to spend months discussing it, so let's merge this and #1506 will come later.

docs/src/training/optimisers.md
darsnack: The original is actually the expected behavior. The first call to `next!(schedule)` will return the very first parameter value. I think that makes sense given that the `opt.eta` value when you construct `opt` can be out of sync with the schedule policy. This style ensures that the schedule policy sets is the authoritative `eta` on every iteration.

darsnack:
@CarloLucibello if you are OK with my response to the review, then we can merge

darsnack:
Never mind missed the approve

bors r+

PR #1515, opened 2021-02-22
tweaks to onecold

CarloLucibello:


src/onehot.jl
```diff
@@ -131,24 +132,33 @@ julia> Flux.onecold([true, false, false], [:a, :b, :c])
...
+ 3
+ 2
+ 1
+ 3
 ```
 """
-onecold(y::AbstractVector, labels = 1:length(y)) = labels[argmax(y)]
-function onecold(y::AbstractArray, labels = 1:size(y, 1))
-  indices = _fast_argmax(y)
-  xs = isbits(labels) ? indices : collect(indices) # non-bit type cannot be handled by CUDA
```
CarloLucibello, line 139: this doesn't seem necessary, I think it is better to error out than convert in those situations, force users to handle things with more care

darsnack: This is to handle the case `onecold(y, ["dog", "cat", ...])`. If `y` is on the GPU, the indices is a `CuArray` and indexing with a `CuArray` fails. I think if we allow non-integer labels, then it should work on the CPU/GPU. There are several issues that related to this that were closed by adding this line.

darsnack:
I would approve other than that I think the `isbits` check should be kept

src/onehot.jl
CarloLucibello: but we are not fixing the issue, we are converting the indexes to cpu arrays, which is kind of surprising, as the fact that this function returns gpu arrays for some labels and cpu ones for others. 
I think that if people want to use non isbits labels they just don't have to call onecold on gpu arrays. 
What if error out nicely with 
```julia
indices isa CuArray && !isbits(labels) && throw(ArgumentError("Only `isbits(labels)` allowed on gpu")). 
```
?

darsnack: (Error looks good to me, just trying to explain a bit more)

The function always returns the type of `labels`. I agree that converting `indices` is surprising, but that's because this is the bug:
```julia
labels = ["a", "b", "c"]
idx = [2, 3]
labels[idx] # works fine
labels[cu(idx)] # throws error
```
The first thing users try to do (myself included back in the day) is `gpu(labels)` which isn't allowed.

darsnack: My thought is that this is a simple case where we can catch what the user intended to do and safely do that. So, if we weren't to do that, then what's the reason for not doing it? I don't see any major performance issues.

Your call though.

CarloLucibello: > The function always returns the type of labels.

not really, it returns 
`map(xi -> labels[xi[1]], xs)`
where `xs` are the indexes.  So it returns an array with same `eltype` as `labels`, but the array type is essentially given by the input `y`. So I prefer to just error out, also because that line could cause type instability

darsnack: Ah true. Okay can we just add the nice error message then?

CarloLucibello:
I ended up with an indexing-based solution, seems to catch more cases on gpu. I also hit #1517.
Need to do some benchmarking before merging.

src/onehot.jl
```diff
@@ -131,25 +131,46 @@ julia> Flux.onecold([true, false, false], [:a, :b, :c])
...
-    return _fast_argmax(convert(_onehot_bool_type(x), x))
-  end
+onecold(y::OneHotArray) = _indices(y)
+onecold(y::OneHotLike) = _isonehot(y) ? _indices(y) : 
+                            onecold(convert(_onehot_bool_type(y), y))
+
+onecold(y::AbstractVector, labels) = labels[onecold(y)]
+onecold(y::AbstractArray, labels) = _onecold(onecold(y), labels)
+
+function _onecold(indices::CuArray, labels)
```
darsnack, line 164: Rename to `_onecold_index`?

PR #1516, opened 2021-02-22
add Embedding layer

CarloLucibello:
Basic implementation. 

Maybe could be improved when https://github.com/FluxML/NNlib.jl/pull/255 lands

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


src/layers/basic.jl
```diff
@@ -371,3 +373,55 @@ function Base.show(io::IO, m::Parallel)
...
+end
+
+@functor Embedding
+
+function Embedding(in::Integer, out::Integer;
+               init = (i...) -> randn(Float32, i...))
+  return Embedding(init(out, in))
+end
+
+(m::Embedding)(x::OneHotMatrix) = m.weight * x # equivalent to m.weight[:, onecold(x)]
```
DhairyaLGandhi, line 420: Please use the correct indexing instead

src/layers/basic.jl
```diff
@@ -371,3 +373,55 @@ function Base.show(io::IO, m::Parallel)
...
+(m::Embedding)(x::OneHotVector) = m.weight * x
+(m::Embedding)(x::AbstractVector) = m.weight[:, x]
+(m::Embedding)(x::Int) = m.weight[:, x]
```
DhairyaLGandhi, line 423: Seems like we should also generalise it to higher dimensions and do the indexing consistently. We can remove the extra type annotations.

DhairyaLGandhi (changes requested):

DhairyaLGandhi:
Needs GPU tests

DhairyaLGandhi:
For such a straightforward routine, do we need a layer? I guess people are used to seeing them elsewhere, but why should that make it a necessary thing.

Might just make it a necessary evil. Oh well.

src/layers/basic.jl
CarloLucibello: multiplication is better because it also performs a consistency check for the sizes, which is one of the very few reasons why one would use a onehot matrix here instead of a plain index vector

src/layers/basic.jl
```diff
@@ -119,14 +120,13 @@ end
...
+function (a::Dense)(x::Union{AbstractVector, AbstractMatrix})
   W, b, σ = a.W, a.b, a.σ
-  sz = size(x)
-  x = reshape(x, sz[1], :) # reshape to handle dims > 1 as batch dimensions 
-  x = σ.(W*x .+ b)
-  return reshape(x, :, sz[2:end]...)
+  return σ.(W*x .+ b)
 end
 
+(a::Dense)(x::AbstractArray) = reshape(a(mat(x)), :, size(x)[2:end]...)
```
darsnack, line 128: I like this clean up. Can we move it to another PR so it doesn't suffer the slowness of new API PRs? Happy to do it (though would be better to preserve authorship).

src/layers/basic.jl
```diff
@@ -371,3 +371,54 @@ function Base.show(io::IO, m::Parallel)
...
+
+"""
+    Embedding(in, out; init=randn)
+
+A lookup table that stores embeddings of dimension `out` 
+for a vocabulary of size `in`. 
+
+This layers is often used to store word embeddings and retrieve them using indices. 
+The input to the layer can be either a vector of indexes
+or the corresponding onehot encoding.
```
darsnack, line 383:
```suggestion
or the corresponding [onehot encoding](@ref Flux.OneHotArray). 
```

src/layers/basic.jl
```diff
@@ -371,3 +371,54 @@ function Base.show(io::IO, m::Parallel)
...
+
+julia> model(x)
+4×5 Matrix{Float32}:
+  0.91139    0.670462    0.463217   0.670462    0.110932
+  0.247225  -0.0823874   0.698694  -0.0823874   0.945958
+ -0.393626  -0.590136   -0.545422  -0.590136    0.77743
+ -0.497621   0.87595    -0.870251   0.87595    -0.772696
+```
+
+julia> model(vocab_idxs) # same as above
```
darsnack, line 405:
```suggestion
julia> model(vocab_idxs) == model(x)
true
```

darsnack (commented):
Nice work

src/layers/basic.jl
CarloLucibello: let's see if this PR gets through quickly (is a simple addition in the end), otherwise I'll factor this change out

darsnack (approved):
Seems simple, makes sense to add.

src/layers/basic.jl
DhairyaLGandhi: No it would forward to our method for it which would index.

DhairyaLGandhi:
Yeah, I don't think we want to take the charge on `mat` here. That, the hard typing for the same logic, and changes to dense aside, it's looking neat.

CarloLucibello:
Since I need to define and use `mat` here for the Embedding layer it is needlessly painful to factor out the change to the dense layer, so please let me keep that change to dense here. We don't need to document and export `mat`

darsnack:
`mat` seems useful for writing N-dimensional behavior for layers. I am fine leaving as is and undocumented/unexported. If there is protest, let's just rename it to something else (it's basically a sibling to `flatten`). Could just have a kwarg in `flatten`.

DhairyaLGandhi:
please take care of the rest of the comments in the meantime?

I don't see how mat was something needed to be defined. The reshape api is pretty clever so as to be self documenting, but I can deal with that later.

src/layers/basic.jl
CarloLucibello: it size checks then indexes

CarloLucibello:
> please take care of the rest of the comments in the meantime?

don't see anything to take care of

> I don't see how mat was something needed to be defined. The reshape api is pretty clever so as to be self documenting, but I can deal with that later.

it is defined because it is convenient, same as flatten, when a code pattern is used in multiple spots it is typically factorized into a function, this how people write programs

src/layers/basic.jl
DhairyaLGandhi: `getindex` checks bounds anyway. Plus that is not relevant to the operation, else we would need that on every other type as well.

DhairyaLGandhi:
> hard typing for the same logic, and changes to dense aside

These ones.

src/layers/basic.jl
CarloLucibello: getindex cannot capture the same errors as this https://github.com/FluxML/Flux.jl/blob/ea41ea64e09c44775f91ee14a253ee9b800ac71a/src/onehot.jl#L157
Plus, let me repeat my first comment (to which you responded erroneously twice)
> multiplication is better because it also performs a consistency check for the sizes, which is one of the very few reasons why one would use a onehot matrix here instead of a plain index vector

CarloLucibello:
> hard typing for the same logic,

I don't know what you refer to here

> and changes to dense aside

I asked to keep the change here

test/layers/basic.jl
```diff
@@ -153,4 +153,25 @@ import Flux: activations
       @test size(Parallel(+, Dense(10, 2), Dense(5, 2), Dense(4, 2))(inputs)) == (2,)
     end
   end
+
+  @testset "Embedding" begin
+    vocab_size, embed_size = 10, 4
+    m = Embedding(vocab_size, embed_size)
+    @test size(m.weight) == (embed_size, vocab_size)
+    
+    x = rand(1:vocab_size, 3)
```
DhairyaLGandhi, line 162: Should test with matrices as well, since we have the method defined specifically for AbstractArray

DhairyaLGandhi: It would `StackOverflow`, so worth having better errors in there.

test/layers/basic.jl
```diff
@@ -153,4 +153,25 @@ import Flux: activations
...
+    y = m(x)
+    @test y isa Matrix{Float32}
+    @test y ≈ m.weight[:,x]
+
+    x2 = OneHotMatrix(x, vocab_size)
+    y2 = m(x2)
+    @test y2 isa Matrix{Float32}
+    @test y2 ≈ y
+    @test_throws DimensionMismatch m(OneHotMatrix(x, 1000))
```
DhairyaLGandhi, line 171: Its surprising to get `DimensionMismtach`, where on regular cases, it should be a bounds error

src/layers/basic.jl
mcabbott: +1 for the cleanup. There is `AbstractVecOrMat` for this union. 

Am not sure I like `mat` as a name, I guess it matches `vec` but also seems like a name often used for variables. `flatten` is a verb... don't have a great suggestion.

src/layers/basic.jl
```diff
@@ -371,3 +371,55 @@ function Base.show(io::IO, m::Parallel)
...
+
+@functor Embedding
+
+function Embedding(in::Integer, out::Integer;
+               init = (i...) -> randn(Float32, i...))
+  return Embedding(init(out, in))
+end
+
+(m::Embedding)(x::Union{OneHotVector, OneHotMatrix}) = m.weight * x # equivalent to m.weight[:,onecold(x)]
+(m::Embedding)(x::Union{Int,AbstractVector}) = m.weight[:, x]
```
mcabbott, line 420: Won't this run into aliasing problems in the gradient? If there are repeats in `x` then I think broadcasting will overwrite itself.

DhairyaLGandhi: I think this also has the potential to limit us by not letting Base figure out the shape of the resultant indexing operation. Ideally, you wouldn't need to convert your input down to index, but rather use the dimensions to figure out the correct output dims. Thoughts?

src/layers/basic.jl
CarloLucibello: I removed `mat` in favor of plain reshape

test/layers/basic.jl
CarloLucibello: right, there was a bug, fixed

src/layers/basic.jl
CarloLucibello: > Won't this run into aliasing problems in the gradient? If there are repeats in x then I think broadcasting will overwrite itself.

I didn't know there was such problem, do we have a corresponding issue in Zygote?

CarloLucibello: It seems aliasing is not a problem (right?)
```julia
julia> w = [1,2,3,4];

julia> gradient(w -> sum(w[[1,1,2,2,2]]), w)
([2, 3, 0, 0],)
```

CarloLucibello: > I think this also has the potential to limit us by not letting Base figure out the shape of the resultant indexing operation.

julia needs to infer the type (which should figure out will be a matrix), not the shape

> Ideally, you wouldn't need to convert your input down to index, but rather use the dimensions to figure out the correct output dims. Thoughts?

I don't understand what you are suggesting, but i don't think the purpose of this layer can be achieved without indexing

test/layers/basic.jl
CarloLucibello: it gets the error from the mul, for which is the appropriate error. Maybe outofbounds would be better in this context, but it doesn't really matter, we just have to error out, a DimensionMismatch error will be hardly confusing

src/layers/basic.jl
mcabbott: Re aliasing, see https://github.com/FluxML/Zygote.jl/issues/821 etc.

CarloLucibello: > Re aliasing, see [FluxML/Zygote.jl#821](https://github.com/FluxML/Zygote.jl/issues/821) etc.

sh*t, this is a big problem for this functionality, when embedding a sentence multiple word occurrences are totally common.
So due to FluxML/Zygote.jl#821  it would be broken on gpu, even worse than that actually, it would give incorrect gradients

mcabbott: Yes. I think ScatterNNlib.jl had GPU code to avoid this, and the NNlib PR was the first step to hooking that up.

CarloLucibello: I see. That PR is taking quite some time, maybe I'll just implement scatter_add here for the time being.

DhairyaLGandhi: > julia needs to infer the type (which should figure out will be a matrix), not the shape

Right, I meant Base specifically, and not the compiler. The `getindex` operation can also determine the resultant shape

```Julia
julia> r = reshape(1:10, 2,:)
2×5 reshape(::UnitRange{Int64}, 2, 5) with eltype Int64:
 1  3  5  7   9
 2  4  6  8  10

julia> r[rand(1:10, 3,3,3)]
3×3×3 Array{Int64,3}:
[:, :, 1] =
 5  6  7
 2  8  1
 3  1  1

[:, :, 2] =
  8  6   4
  5  5  10
 10  2   9

[:, :, 3] =
 2  7  9
 5  9  3
 4  7  9
```

darsnack: I'm not sure I understand the relevance to the PR? Embedding layers accept indices as input. Where is the "conversion to indices" happening?

DhairyaLGandhi: Where do we convert? I think I should have said reshape instead of "convert down" the shape.

darsnack: I'm sorry, I just don't see where the reshaping or "converting down" is happening. We have a matrix whose columns are word embeddings. We want to select a subset of those embeddings. What's preventing the shape from being inferred?

CarloLucibello: @DhairyaLGandhi are you aware of the fact that the line of code we are commenting restricts the indexes to `Union{Int,AbstractVector}`?

CarloLucibello:
@mcabbott given the aliasing problem, I think that the thing to do here is to restrict the input to cpu only (i.e. error out on CuArrays) for the time being so that we can merge this safely. When we have the gather/scatter infrastructure in place in NNlib.jl and CUDA.jl we switch the backend here.

jeremiedb:
Is there a reason to favor the reliance on ScatterNNLib for the embedding operations rather than having a fix for the incorrect gradient when indices have repeats? 
I understand that ScaterNNLib presents the benefit of doing the correct calculation, but it seems to leave in place a nasty silent bug. Having improper gradient if indexing an array with repeating indices looks like the elephant in room. 

It's my understanding that https://github.com/JuliaLang/julia/pull/31407 was to address it, but it looks stalled. Any idea how to raise awareness on this issue would be welcome!

src/layers/basic.jl
DhairyaLGandhi: I'm specifically trying to avoid having that type in there.

darsnack: Oh I think I finally see what you were trying to say. You're concerned about the case where `x` > 1D if the type restrictions are lifted?

CarloLucibello:
Yes, it would be good to fix https://github.com/FluxML/Zygote.jl/issues/821. 
Since the cpu gradient is working correctly, maybe it is enough to address https://github.com/JuliaGPU/CUDA.jl/issues/89

src/layers/basic.jl
darsnack: What's the reason for removing the hard-typing? To be more generic? I'm not sure being able to index with non-vectors is useful here, and I'm more worried about silent errors.

CarloLucibello:
this can now be implemented on top of the scatter/gather functions implemented in NNlib

darsnack:
This layer needs to special case (like the normalization layers) in `Flux.outputsize` since indexing with `Nil` is non-sensical. We ran into this issue trying to use the layer in FastAI.jl.

CarloLucibello:
gpu failure seems unrelated

CarloLucibello:
I added `gather` here to have tests passing and added @manikyabard as a coauthor, but I guess #1656 is still necessary for a few improvements

CarloLucibello:
Looks like the gpu version of `NNlib.gather` is stll buggy
```julia
julia> x = [1, 3, 5, 3]; # repeated indexes

julia> W = rand(2,5);

julia> gradient(W -> sum(NNlib.gather(W,x)), W)[1]
2×5 Matrix{Float64}:
 1.0  0.0  2.0  0.0  1.0
 1.0  0.0  2.0  0.0  1.0

julia> gradient(W -> sum(NNlib.gather(W,x)), W |> cu)[1]
2×5 CuArray{Float32, 2}:
 1.0  0.0  1.0  0.0  1.0
 1.0  0.0  1.0  0.0  1.0
```

CarloLucibello:
It's weird, Zygote it's not picking up the correct rule, which is defined here
https://github.com/FluxML/NNlib.jl/blob/master/src/gather.jl#L77
and works properly:
```julia
julia> NNlib.scatter!(+, zeros(size(W)) |> cu, ones(2,4) |> cu, x |> cu)
2×5 CuArray{Float32, 2}:
 1.0  0.0  2.0  0.0  1.0
 1.0  0.0  2.0  0.0  1.0
```

CarloLucibello:
I fixed the issue in `NNlib.jl`, tests should pass here once the new version is released (within an hour I guess)

CarloLucibello:
pfff, buildkite is not picking up the last NNlib release. Maybe it is periodically caching the Registry?

docs/src/models/overview.md
```diff
@@ -42,38 +42,38 @@ Normally, your training and test data come from real world observations, but thi
...
+```julia
 julia> model = Dense(1, 1)
 Dense(1, 1)
 
-julia> model.W
-1-element Array{Float64,1}:
- -0.99009055
+julia> model.weight
+1×1 Matrix{Float32}:
+ -1.4925033
```
darsnack, line 51: Are these changes not included in a rebase?

src/layers/basic.jl
```diff
@@ -428,3 +429,58 @@ function Base.show(io::IO, m::Parallel)
...
   print(io, ")")
 end
+
+"""
+    Embedding(in, out; init=randn)
+
+A lookup table that stores embeddings of dimension `out` 
+for a vocabulary of size `in`. 
+
+This layers is often used to store word embeddings and retrieve them using indices.
```
darsnack, line 439:
```suggestion
This layer is often used to store word embeddings and retrieve them using indices. 
```

src/layers/basic.jl
```diff
@@ -428,3 +429,58 @@ function Base.show(io::IO, m::Parallel)
...
+
+"""
+    Embedding(in, out; init=randn)
+
+A lookup table that stores embeddings of dimension `out` 
+for a vocabulary of size `in`. 
+
+This layers is often used to store word embeddings and retrieve them using indices. 
+The input to the layer can be either a vector of indexes
+or the corresponding [onehot encoding](@ref Flux.OneHotArray).
```
darsnack, line 441:
```suggestion
The input to the layer can be either a vector of indices
or the corresponding [onehot encoding](@ref Flux.OneHotArray). 
```

src/layers/basic.jl
```diff
@@ -428,3 +429,58 @@ function Base.show(io::IO, m::Parallel)
...
+A lookup table that stores embeddings of dimension `out` 
+for a vocabulary of size `in`. 
+
+This layers is often used to store word embeddings and retrieve them using indices. 
+The input to the layer can be either a vector of indexes
+or the corresponding [onehot encoding](@ref Flux.OneHotArray). 
+
+# Examples
+
+```julia-repl
```
darsnack, line 445:
````suggestion
```jldoctest
````

File: src/layers/basic.jl [hunk 13]
```diff
@@ -428,3 +429,58 @@ function Base.show(io::IO, m::Parallel)
...
+struct Embedding{W}
+  weight::W
+end
+
+@functor Embedding
+
+function Embedding(in::Integer, out::Integer;
+               init = (i...) -> randn(Float32, i...))
+  return Embedding(init(out, in))
+end
```
darsnack, line 477:
```suggestion
Embedding(in::Integer, out::Integer;
          init = (i...) -> randn(Float32, i...)) = Embedding(init(out, in))
```

mcabbott: Might be clearer to define `randn32(s...) = randn(Float32, s...)` somewhere, instead of a closure.

src/layers/basic.jl
mcabbott: Is the plan to merge #1656 on top of this, or to copy things across from it to here? 

It has a nicer jldoctest example, without random numbers.

src/layers/basic.jl
mcabbott: Or, should this even be the default? It could be something like `unitvector_init(r,c) = begin m = randn(Float, r, c); m ./ sum(abs, m, dims=1)`.

Looks like pytorch does just use `randn`.

darsnack: There is `Flux.ones`, etc. for `Float32` defaults. We could add `Flux.rand` and `Flux.randn`.

mcabbott: IMO `Flux.zeros` being special is a footgun, and we should try to remove it. If it were called `zeros32` that would be self-explanatory.

CarloLucibello:
How do I retrigger buildkite?

src/layers/basic.jl
CarloLucibello: #1656 will be merged after merging this, so I'll leave changed to that docstring to tht PR

docs/src/models/overview.md
CarloLucibello: these are doc issues I found while working at this PR and took the chance to fix

src/layers/basic.jl
CarloLucibello: I'll file a separate PR with the zeros32, ones32, ... change 

> Or, should this even be the default?

I think it is standard initialization

mcabbott: Too slow, #1661 is the PR

CarloLucibello:
ok, need an approval for merge. refinements will go into #1656

darsnack (approved):

CarloLucibello:
@DhairyaLGandhi can you dismiss your change request?

src/layers/basic.jl
```diff
@@ -428,3 +429,55 @@ function Base.show(io::IO, m::Parallel)
...
+"""
+struct Embedding{W}
+  weight::W
+end
+
+@functor Embedding
+
+Embedding(in::Integer, out::Integer; init = randn32) = Embedding(init(out, in))
+  
+(m::Embedding)(x::Union{OneHotVector, OneHotMatrix}) = m.weight * x # equivalent to m.weight[:,onecold(x)]
```
DhairyaLGandhi, line 476: This should use the definition in the comment.

src/layers/basic.jl
```diff
@@ -428,3 +429,55 @@ function Base.show(io::IO, m::Parallel)
...
+(m::Embedding)(x::Integer) = m([x])
```
DhairyaLGandhi, line 477:
```suggestion
(m::Embedding)(x::Int) = m([x])
```

Safer for 32-bit systems

src/layers/basic.jl
```diff
@@ -428,3 +429,55 @@ function Base.show(io::IO, m::Parallel)
...
+julia> model(vocab_idxs) == model(x)
+true
+"""
+struct Embedding{W}
+  weight::W
+end
+
+@functor Embedding
+
+Embedding(in::Integer, out::Integer; init = randn32) = Embedding(init(out, in))
```
DhairyaLGandhi, line 474:
```suggestion
Embedding(in::Int, out::Int; init = randn32) = Embedding(init(out, in))
```

test/cuda/layers.jl
```diff
@@ -259,3 +259,39 @@ end
...
+  
+  gs = gradient(() -> sum(m(x)), params(m))
+  gs_g = gradient(() -> sum(m_g(x_g)), params(m_g))
+  @test collect(gs_g[m_g.weight]) ≈ gs[m.weight]
+
+  gs = gradient(() -> sum(tanh.(m(x))), params(m))
+  gs_g = gradient(() -> sum(tanh.(m_g(x_g))), params(m_g))
+  @test collect(gs_g[m_g.weight]) ≈ gs[m.weight]
+
+  @testset "repeated indexes" begin
```
DhairyaLGandhi, line 282:
```suggestion
  @testset "repeated indices" begin
```

test/cuda/layers.jl
```diff
@@ -259,3 +259,39 @@ end
...
+    vocab_size, embed_size = 5, 2
+    m = Flux.Embedding(vocab_size, embed_size)
+
+    x = [1, 3, 5, 3] # repeated indexes
+    y = m(x)
+    m_g = m |> gpu
+    x_g = x |> gpu
+    y_g = m_g(x_g)
+    @test collect(y_g) == y
```
DhairyaLGandhi, line 291:
```suggestion
    @test Array(y_g) == y
```

test/cuda/layers.jl
```diff
@@ -259,3 +259,39 @@ end
...
+    gs = gradient(() -> sum(m(x)), params(m))
+    gs_g = gradient(() -> sum(m_g(x_g)), params(m_g))
+    @test collect(gs_g[m_g.weight]) ≈ gs[m.weight]
```
DhairyaLGandhi, line 294:
```suggestion
    @test Array(gs_g[m_g.weight]) ≈ gs[m.weight]
```

test/layers/basic.jl
```diff
@@ -191,4 +191,29 @@ import Flux: activations
...
+
+    x = rand(1:vocab_size, 3, 4)
+    y = m(x)
+    @test y isa Array{Float32, 3}
+    @test size(y) ==  (embed_size, 3, 4)
```
DhairyaLGandhi, line 213:
```suggestion
    @test size(y) == (embed_size, 3, 4)
```

test/cuda/layers.jl
```diff
@@ -259,3 +259,39 @@ end
...
+  x = [1, 3, 5]
+  y = m(x)
+  m_g = m |> gpu
+  x_g = x |> gpu
+  y_g = m_g(x_g)
+  @test collect(y_g) == y
+  
+  gs = gradient(() -> sum(m(x)), params(m))
+  gs_g = gradient(() -> sum(m_g(x_g)), params(m_g))
+  @test collect(gs_g[m_g.weight]) ≈ gs[m.weight]
```
DhairyaLGandhi, line 276:
```suggestion
  @test Array(gs_g[m_g.weight]) ≈ gs[m.weight]
```

Should this be added to `gpu_gradtest` ? We don't have any activations here, but we can run it through `identity`.

NEWS.md
```diff
@@ -1,5 +1,14 @@
 # Flux Release Notes
 
+## v0.12.4
+* Implemented an [`Embedding layer`](https://github.com/FluxML/Flux.jl/pull/1516) 
+  based on recently added  `NNlib.gather` and `NNlib.scatter`.
```
DhairyaLGandhi, line 5:
```suggestion
  based on `NNlib.gather` and `NNlib.scatter`.
```

Project.toml
```diff
@@ -1,12 +1,14 @@
...
-version = "0.12.4"
+version = "0.12.5"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
```
DhairyaLGandhi, line 11: These deps don't look right.

DhairyaLGandhi (commented):
I've left a couple thoughts.

There seem to be a lot of changes around code not intended to be changed in this PR (around the field names of `Dense`) which should go in a different PR.

src/layers/basic.jl
CarloLucibello: the comment mentions a semantically equivalent implementation but not necessarily the actual one. 
Since we have a definition of onehot multiplication, I'd rather trust it to do the best thing than do it on my own.

src/layers/basic.jl
CarloLucibello: I'd rather not constrain types, since people could use unsigned integers or other stuff in applications

src/layers/basic.jl
CarloLucibello: as above

src/layers/basic.jl
DhairyaLGandhi: Well, embedding is conceptually an indexing operation, so the correct implementation should come from indexing. In fact, this method should ideally be avoided entirely in favour of using the one with abstract array.

DhairyaLGandhi:
Also needs GPU tests with one-hot and integers.

mcabbott:
If the plan is to make this perfect, then #1656 should be closed and all comments there addressed here. Alternatively, if the plan is to merge this as a stepping stone to #1656, as mentioned above, then it's a waste of time trying to make sure it doesn't have a hair out of place.

DhairyaLGandhi:
My comment pointed out adding tests for the features this PR wants to bring in.

#1656 aims to build on top of this, and has its own opinions about API, not implementation.

DhairyaLGandhi (approved):

src/layers/basic.jl
CarloLucibello: getindex cannot be used due to the gpu problems discussed above

CarloLucibello:
need an approval again

DhairyaLGandhi (approved):
bors r+

Issue #1483, opened 2021-01-27
RNN and GRU give mutation error; LSTM gives ArgumentError about number of fields


maetshju:
I couldn't find another issue mentioning this, so my apologies if this has already come up. On the current master branch, 08e79c4, RNN layers give a mutation error from Zygote during the backward pass.

```julia
using Flux
m = RNN(5, 5)
loss(x) = sum(Flux.stack(m.(x), 2))
x = [rand(5) for i in 1:2]
Flux.train!(loss, Flux.params(m), x, ADAM())
```
<details>
  <summary>RNN stacktrace</summary>

```julia
ERROR: Mutating arrays is not supported
Stacktrace:
 [1] error(::String) at .\error.jl:33
 [2] (::Zygote.var"#375#376")(::Nothing) at C:\Users\freez\.julia\packages\Zygote\Iz3wR\src\lib\array.jl:61
 [3] (::Zygote.var"#2270#back#377"{Zygote.var"#375#376"})(::Nothing) at C:\Users\freez\.julia\packages\ZygoteRules\OjfTt\src\adjoint.jl:59
 [4] copyto_axcheck! at .\abstractarray.jl:946 [inlined]
 [5] (::typeof(∂(copyto_axcheck!)))(::Nothing) at C:\Users\freez\.julia\packages\Zygote\Iz3wR\src\compiler\interface2.jl:0
 [6] Array at .\array.jl:562 [inlined]
 [7] (::typeof(∂(Array{Float32,2})))(::Nothing) at C:\Users\freez\.julia\packages\Zygote\Iz3wR\src\compiler\interface2.jl:0
 [8] convert at .\array.jl:554 [inlined]
 [9] (::typeof(∂(convert)))(::Nothing) at C:\Users\freez\.julia\packages\Zygote\Iz3wR\src\compiler\interface2.jl:0
 [10] setproperty! at .\Base.jl:34 [inlined]
 [11] (::typeof(∂(setproperty!)))(::Nothing) at C:\Users\freez\.julia\packages\Zygote\Iz3wR\src\compiler\interface2.jl:0
 [12] Recur at C:\Users\freez\.julia\dev\Flux\src\layers\recurrent.jl:34 [inlined]
[...]

PR #1521, opened 2021-02-28
Fixes to Recurrent models for informative type mismatch error & output Vector for Vector input

jeremiedb:
Minor fix to Recurrent to return `Vector` with `Vector` input, returns an indicative error relative to type incompatibility where eltype of input doesn't match with eltype of state, as well as some typos in associated docs. 
As discussed in #1483.

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -27,11 +27,11 @@ rnn.state   # 60
 """
 mutable struct Recur{T,S}
   cell::T
-  state::S
+  state::Tuple{Vararg{S}}
 end
 
-function (m::Recur)(xs...)
-  m.state, y = m.cell(m.state, xs...)
+function (m::Recur)(x)
+  m.state, y = m.cell(m.state, x)
   return y
 end
 
@@ -74,16 +74,16 @@ struct RNNCell{F,A,V,S}
   Wi::A
   Wh::A
   b::V
-  state0::S
+  state0::Tuple{Vararg{S}}
 end
 
 RNNCell(in::Integer, out::Integer, σ=tanh; init=Flux.glorot_uniform, initb=zeros, init_state=zeros) = 
-  RNNCell(σ, init(out, in), init(out, out), initb(out), init_state(out,1))
+  RNNCell(σ, init(out, in), init(out, out), initb(out), (init_state(out,1),))
 
-function (m::RNNCell)(h, x)
+function (m::RNNCell{F,A,V,S})((h,), x) where {F,A,V,S}
   σ, Wi, Wh, b = m.σ, m.Wi, m.Wh, m.b
   h = σ.(Wi*x .+ Wh*h .+ b)
-  return h, h
+  return (h,), h
 end
 
 @functor RNNCell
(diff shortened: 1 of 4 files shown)


jeremiedb:
Tests error on OneHotInput. I missed that case, which brakes the design of the informative error on inconsistent eltype between input and state. Would dropping the parametric type in Cells' arguments be acceptable? That would revert to current case where the input/state eltype mismatch results in a more obscure `mutation nor supported` error.

darsnack:
Why not just let `x::S where S<:Union{<:AbstractArray{S}, OneHotArray}`?

jeremiedb:
> Why not just let `x::S where S<:Union{<:AbstractArray{S}, OneHotArray}`?

I think that's a good suggestion :) My understanding is that Union are to be avoided if possible, but in this case, it seems the best course of action. Tests are back to green, thanks!

CarloLucibello:
looks good! it only needs a `@test_throw` check for the type mismatch and tests for the vector input.

src/layers/recurrent.jl
```diff
@@ -80,10 +80,11 @@ end
 RNNCell(in::Integer, out::Integer, σ=tanh; init=Flux.glorot_uniform, initb=zeros, init_state=zeros) = 
   RNNCell(σ, init(out, in), init(out, out), initb(out), init_state(out,1))
 
-function (m::RNNCell)(h, x)
+function (m::RNNCell{F,A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {F,A,V,T}
```
DhairyaLGandhi, line 83: Doesn't one hot array being an abstract array hold already? We don't seem to be doing anything specific to one hot arrays.

DhairyaLGandhi:
We are also considering using cudnn instead, it's unclear whether we have actually seen a performance advantage.

CarloLucibello:
whether we do it through cudnn or not, we can add support for batched 3d sequences at any point in addition to the current interface. I think @jeremiedb already mentioned that a few months ago

src/layers/recurrent.jl
darsnack: `OneHotArray` won't type match the `<:AbstractMatrix{T}` for the hidden state. In general, the point of this PR is to match the `eltype`s, but in the case of `OneHotArray`, we don't need that guarantee because we know it will promote to the `eltype` of the hidden state. So no type instabilities.

ToucheSir:
Given how other frameworks have to bend over backwards and add a bunch of nasty conditionals/documentation caveats to integrate with cuDNN RNNs, I'd say we should only consider adding it if we decide to support 3d sequences.

darsnack (commented):
I think we want to avoid unions for struct fields, but it should be fine for use in dispatch like this.

jeremiedb:
@CarloLucibello Tests on type mismatch and vector vs matrix input have been added. 
@DhairyaLGandhi From the previously held discussion around CUDNN's RNN, point was following what @ToucheSir mentioned, that is that CUDNN's would makes sense in the context of 3d input. I still don't see benefit of introducing a machinery designed for 3D input whose optimizations get lost in the process of adapting it back into the Flux 2D design. I still see adding such feature as desirable, but of a different nature than the current Flux Recurrent approach.

CarloLucibello (approved):

PR #1524, opened 2021-03-02
Add identity_init

DrChainsaw:
As discussed in https://github.com/FluxML/Flux.jl/issues/1431#issuecomment-788836031

Since the only information available is the dimensions, it makes the following assumptions:

*  1D: A `Vector` of `zeros` (assumes bias)
*  2D: An identity matrix (assumes matrix multiplication)
*  2+D: A diagnoal matrix of identity kernels (assumes convolution) 

Not sure if there is a better description of the last one (thats just how I envision it).

I don't think it is possible to create identity mappings for all layer types and if it is, then doing so probably requires that weight init gets to know the layer type. The ones which will not have an ID mapping with this weight init are the RNNs, normalization layers (don't need it) and DepthwiseConv.

### PR Checklist

- [X ] Tests are added
- [X ] Entry in NEWS.md

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -285,16 +285,28 @@ sparse_init(dims...; kwargs...) = sparse_init(Random.GLOBAL_RNG, dims...; kwargs
 sparse_init(rng::AbstractRNG; init_kwargs...) = (dims...; kwargs...) -> sparse_init(rng, dims...; init_kwargs..., kwargs...)
 
 """
-    identity_init([rng=GLOBAL_RNG], dims...)
+    identity_init([rng=GLOBAL_RNG], dims...;gain=1)
 
-Return an `Array` of size `dims` which yields an identity mapping when used as parameters in most Flux layers.
+Return an `Array` of size `dims` which yields an identity mapping when used as parameters in 
+most Flux layers.
 
-Often useful in the context of transfer learning, i.e when one wants to add more capacity to a model but start from the same mapping.
+Often useful in the context of transfer learning, i.e when one wants to add more capacity to
+a model but start from the same mapping.
+
+Some caveats: Not all layers will be identity mapping when used with this init. Exceptions
+include recurrent layers, `DepthwiseConv` and normalization layers.
+
+Also note that layers must have `input_size == output_size` for identity mapping to be 
+possible.
+
+For convolutional layers, in addition to the above, the kernel sizes must also be odd and 
+padding must be applied so that output feature maps have the same size as input feature maps,
+e.g by using `SamePad`.
 
 Has the following behaviour
 *  1D: A `Vector` of `zeros` (useful for an identity bias)
 *  2D: An identity matrix (useful for an identity matrix multiplication)
-*  More than 2D: A diagnoal matrix of identity kernels (useful for an identity convolution) 
+*  More than 2D: A dense block array of center tap spatial filters (useful for an identity convolution)
 
 ```jldoctest
 julia> Flux.identity_init(3,3)
@@ -303,71 +315,55 @@ julia> Flux.identity_init(3,3)
  0.0  1.0  0.0
  0.0  0.0  1.0
 
- julia> Flux.identity_init(3,1,1)
- 3×1×1 Array{Float32,3}:
- [:, :, 1] =
-  0.0
-  1.0
-  0.0
+julia> Flux.identity_init(3,5)
+3×5 Array{Float32,2}:
+ 1.0  0.0  0.0  0.0  0.0
+ 0.0  1.0  0.0  0.0  0.0
+ 0.0  0.0  1.0  0.0  0.0
 
-  julia> Flux.identity_init(1,3,3,3)
-  1×3×3×3 Array{Float32,4}:
-  [:, :, 1, 1] =
-   0.0  1.0  0.0
-  
-  [:, :, 2, 1] =
-   0.0  0.0  0.0
-  
-  [:, :, 3, 1] =
-   0.0  0.0  0.0
-  
-  [:, :, 1, 2] =
-   0.0  0.0  0.0
-  
-  [:, :, 2, 2] =
-   0.0  1.0  0.0
-  
-  [:, :, 3, 2] =
-   0.0  0.0  0.0
-  
-  [:, :, 1, 3] =
-   0.0  0.0  0.0
-  
-  [:, :, 2, 3] =
-   0.0  0.0  0.0
-  
-  [:, :, 3, 3] =
-   0.0  1.0  0.0
+julia> Flux.identity_init(3,3,2,2)
+3×3×2×2 Array{Float32,4}:
+[:, :, 1, 1] =
+ 0.0  0.0  0.0
+ 0.0  1.0  0.0
+ 0.0  0.0  0.0
+
+[:, :, 2, 1] =
+ 0.0  0.0  0.0
+ 0.0  0.0  0.0
+ 0.0  0.0  0.0
+
+[:, :, 1, 2] =
+ 0.0  0.0  0.0
+ 0.0  0.0  0.0
+ 0.0  0.0  0.0
+
+[:, :, 2, 2] =
+ 0.0  0.0  0.0
+ 0.0  1.0  0.0
+ 0.0  0.0  0.0
 ```
 """
-identity_init(::AbstractRNG, dims...) = identity_init(dims...)
 # Assume bias
-identity_init(cols) = zeros(cols)
-# Assume matrix multiplication
-function identity_init(rows, cols)
-  rows == cols || throw(ArgumentError("Identity mapping not possible with rows != cols! Got rows=$rows, rows=$cols. Use identity_init_nocheck to bypass this check."))
-  return identity_init_nocheck(rows,cols)
-end
-identity_init_nocheck(rows, cols) = Matrix{Float32}(I, rows,cols)
-# Assume convolution
-function identity_init(dims...)
-  nin, nout = dims[end-1], dims[end]
-  nin == nout || throw(ArgumentError("Identity mapping not possible with nin != nout! Got nin=$nin, nout=$nout. Use identity_init_nocheck to bypass this check."))
-  all(isodd, dims[1:end-2]) || throw(ArgumentError("Identity mapping requires odd kernel sizes! Got $(dims[1:end-2]). Use identity_init_nocheck to bypass this check."))
-  return identity_init_nocheck(dims...)
-end
+identity_init(cols;gain=1) = zeros(cols)
 
-# This can still be useful if one e.g is creating an inception-like block which as a whole shall be an identity mapping
-# Requires a bit of circshifting but it is fully doable
-function identity_init_nocheck(dims...)
+# Assume matrix multiplication
+identity_init(rows, cols;gain=1) = Matrix{Float32}(I * gain, rows,cols)
+
+# Assume convolution
+function identity_init(dims...;gain=1)
   nin, nout = dims[end-1], dims[end]
   centers = map(d -> cld(d, 2), dims[1:end-2])
   weights = zeros(dims)
   for i in 1:min(nin,nout)
-      weights[centers..., i, i] = 1
+      weights[centers..., i, i] = gain
   end
   return weights
 end
+
+identity_init(::AbstractRNG, dims...;kwargs...) = identity_init(dims...;kwargs...)
+identity_init(;init_kwargs...) = identity_init(Random.GLOBAL_RNG; init_kwargs...)
+identity_init(rng::AbstractRNG; init_kwargs...) = (args...;kwargs...) -> identity_init(rng, args...; init_kwargs..., kwargs...)
 
 
 ones(T::Type, dims...) = Base.ones(T, dims...)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -297,7 +297,7 @@ Some caveats: Not all layers will be identity mapping when used with this init. 
 include recurrent layers, `DepthwiseConv` and normalization layers.
 
 Also note that layers must have `input_size == output_size` for identity mapping to be 
-possible.
+possible. When this is not the case, the "identity array" is padded with zeros.
 
 For convolutional layers, in addition to the above, the kernel sizes must also be odd and 
 padding must be applied so that output feature maps have the same size as input feature maps,
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -294,6 +294,7 @@ Often useful in the context of transfer learning, i.e when one wants to add more
 a model but start from the same mapping.
 
 Use `shift` (integer or tuple) to apply circular shift to the output.
+Equivalent to `Base.circshift(identity(dims...), shift)`.
 
 Some caveats: Not all layers will be identity mapping when used with this init. Exceptions
 include recurrent layers, `DepthwiseConv` and normalization layers.
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -351,7 +351,7 @@ identity_init(cols;gain=1) = zeros(cols)
 identity_init(rows, cols; gain=1) = Matrix{Float32}(I * gain, rows,cols)
 
 # Assume convolution
-function identity_init(dims...;gain=1)
+function identity_init(dims...; gain=1)
   nin, nout = dims[end-1], dims[end]
   centers = map(d -> cld(d, 2), dims[1:end-2])
   weights = zeros(dims)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -356,7 +356,7 @@ function identity_init(dims...; gain=1)
   centers = map(d -> cld(d, 2), dims[1:end-2])
   weights = zeros(dims)
   for i in 1:min(nin,nout)
-      weights[centers..., i, i] = gain
+    weights[centers..., i, i] = gain
   end
   return weights
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -285,7 +285,7 @@ sparse_init(dims...; kwargs...) = sparse_init(Random.GLOBAL_RNG, dims...; kwargs
 sparse_init(rng::AbstractRNG; init_kwargs...) = (dims...; kwargs...) -> sparse_init(rng, dims...; init_kwargs..., kwargs...)
 
 """
-    identity_init([rng=GLOBAL_RNG], dims...;gain=1)
+    identity_init([rng=GLOBAL_RNG], dims...; gain=1)
 
 Return an `Array` of size `dims` which yields an identity mapping when used as parameters in 
 most Flux layers.
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -345,21 +345,21 @@ identity_init(::AbstractRNG, dims...) = identity_init(dims...)
 identity_init(cols) = zeros(cols)
 # Assume matrix multiplication
 function identity_init(rows, cols)
-  rows == cols || @warn "Identity mapping not possible with rows != cols! Got rows=$rows, rows=$cols."
-  return identity_init_nowarn(rows,cols)
+  rows == cols || throw(ArgumentError("Identity mapping not possible with rows != cols! Got rows=$rows, rows=$cols. Use identity_init_nocheck to bypass this check."))
+  return identity_init_nocheck(rows,cols)
 end
-identity_init_nowarn(rows, cols) = Matrix{Float32}(I, rows,cols)
+identity_init_nocheck(rows, cols) = Matrix{Float32}(I, rows,cols)
 # Assume convolution
 function identity_init(dims...)
   nin, nout = dims[end-1], dims[end]
-  nin == nout || @warn "Identity mapping not possible with nin != nout! Got nin=$nin, nout=$nout."
-  all(isodd, dims[1:end-2]) || @warn "Identity mapping requires odd kernel sizes! Got $(dims[1:end-2])."
-  return identity_init_nowarn(dims...)
+  nin == nout || throw(ArgumentError("Identity mapping not possible with nin != nout! Got nin=$nin, nout=$nout. Use identity_init_nocheck to bypass this check."))
+  all(isodd, dims[1:end-2]) || throw(ArgumentError("Identity mapping requires odd kernel sizes! Got $(dims[1:end-2]). Use identity_init_nocheck to bypass this check."))
+  return identity_init_nocheck(dims...)
 end
 
 # This can still be useful if one e.g is creating an inception-like block which as a whole shall be an identity mapping
 # Requires a bit of circshifting but it is fully doable
-function identity_init_nowarn(dims...)
+function identity_init_nocheck(dims...)
   nin, nout = dims[end-1], dims[end]
   centers = map(d -> cld(d, 2), dims[1:end-2])
   weights = zeros(dims)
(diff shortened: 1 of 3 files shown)


darsnack:
Looking good. A few initial comments:
- Rename the function to `identity_init` to match `sparse_init`?
- Instead of warnings, throw argument errors (this is typically what we do for the constraints like the ones here)

DrChainsaw:
> Rename the function to identity_init to match sparse_init?
Sounds good!

> Instead of warnings, throw argument errors (this is typically what we do for the constraints like the ones here)
Yeah, this is probably safer. I'll do that and add a reference to the unchecked function in the error message.

I manged to context switch before I saw your very timely comments, so I'll do the updates tomorrow since its late here now.

DrChainsaw:
@darsnack Comments adressed now.

darsnack (changes requested):
I would remove the `*_nocheck` versions of the functions and directly put the `*_nocheck` bodies into the original functions. We don't offer the ability to bypass constraints typically. If there is a need to bypass them, then let's think about whether we want to enforce the constraint in the first place. This ties into my second comment.

I'm trying to think of what the right behavior is for the N-dimensional case. If I understand the code correctly, the current implementation "picks" the center of the portion under the sliding window of a convolution? What happens across the input channels? Do the centers get added together? If so, why is that the right thing to do? Not trying to shoot down anything, I'm genuinely uncertain about what's correct and trying to understand the motivation for the choices. I saw a comment about inception-like modules, but I didn't quite follow it.

In general though, everything here is great. It's always figuring out the generic case that gets me tripped up with these init functions.

DrChainsaw:
> I would remove the *_nocheck versions of the functions and directly put the *_nocheck bodies into the original functions.

This has to do with the inception comment. Maybe it is a bit of an edge case, but I do make use of it here: https://github.com/DrChainsaw/NaiveGAflux.jl/blob/99651d8906d71693e2f72d469ceb208c1372fc2d/src/archspace.jl#L252
which comes from here:
https://github.com/DrChainsaw/NaiveGAflux.jl/blob/99651d8906d71693e2f72d469ceb208c1372fc2d/src/archspace.jl#L479
and I was hoping to be able to use the Flux builtin when it becomes available.

Trying to explain in words instead: To insert an identity inception block between two layers, you need the output of the concatenation to have the same size as the output/input of the layers you are inserting between. This in turn means that each path must shrink the input and this in turn means that each path needs to be the identity of a subset of its input, and there is where the circshifted "partial identity" comes in. Without any extra effort, this "partial identity" is exactly what you get out from `identity_init_nocheck` if input and output size does not match and all you need to do is to circshift it so it is identity for the exact same part of its input as the part it is representing in the concatenation.

>I'm trying to think of what the right behavior is for the N-dimensional case.
I must say that I still get dizzy when trying to reason about N (>2)-dimensional arrays, but here is my thought model of how they are used in convolutional layers: Just as a normal dense layer "mangles" all inputs together with matrix multiplication, so does a conv layer, but you basically replace each scalar multiplication in the dense case with a convolution. 

The "convolutional identity" is a single tap unity filter, and in order to not shift the phase the tap must be centered (which is only possible if the kernel has an odd size). That last part is btw the reason why I think in some cases `identity_init_nocheck` can be valid; A picture convolved with e.g a 2x2 kernel with the first element being 1 will look exactly the same, but will be shifted one pixel.

Anyways, back to the "mangling" of inputs, the way I wrap my head around ND arrays for convolution is to think of the weight as a matrix with convolutional kernels instead of scalars. E.g. for a 4D weight, I think of it as a matrix where the elements are matrices. Now, if you put identities on the diagonal, the way the conv layers work, this will yield the exact same effect as when you put scalar identities in the diagonal of a normal matrix and perform matrix multiplication: Each input channel will only ever hit an identity kernel and for that same row/column all other input channels will hit a zero-kernel so that after summation only the input channel remains. This also offers some intuition on how the "partial identity" for the inception case works.

Note that this ofc only holds for convolutional layers. If someone where to invent a new layertype which uses more than 2 dimensions in the weight, there is absolutely nothing which says that the current identity_init will actually be identity for that layertype.

src/utils.jl
```diff
@@ -284,6 +284,92 @@ end
...
+*  More than 2D: A diagnoal matrix of identity kernels (assumes convolution) 
+
+```jldoctest
+julia> Flux.identity_init(3,3)
+3×3 Array{Float32,2}:
+ 1.0  0.0  0.0
+ 0.0  1.0  0.0
+ 0.0  0.0  1.0
+
+ julia> Flux.identity_init(3,1,1)
```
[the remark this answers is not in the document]
DrChainsaw: But the 3,1,1 is showcasing an identity weight for a 1D convolutional layer. Won't that be more useful to understand what it does as returning zeros is described above and probably not something the user will have any doubts about what it looks like.

darsnack:
Thanks for the explanation.

So for the checks (given the inception case you described), I am inclined to suggest no checking at all. Certainly in the 2D case, I don't think we should force `rows == cols`. We can just pad the larger dimension with zeros, since that is generally accepted as the identity for non-square matrices.

The odd kernel size constraint is tougher to justify ignoring, but maybe this is better served with a docs example.

As for generalizing the N-dimensional case, I just want to repeat the explanation to make sure I get it. If my kernel is WxHxIxO, then I think of that as a IxO block matrix where each block is a WxH matrix? Each block on the diagonal needs to be a "1" and "0". For a "1", we use a filter with a tap in the center (makes sense when you just think about 2D spatial convolution). The "0" is just `zeros`. In effect, this means that when the output channel matches the input channel, we copy the input channel and ignore the rest. Thanks for the explanation, makes sense now that I have written it out.

So, I think for the same reasons as the 2D case, we can allow the number of input channels to be different from the number of output channels. Thinking in your block matrix format, you would just pad the larger dimension with zero blocks (as with the 2D case).

src/utils.jl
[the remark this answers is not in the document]
darsnack: I want to keep both examples in there! Just adding to the list, not replacing.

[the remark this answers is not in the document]
darsnack: But I agree, you can ignore this if there are too many examples

DhairyaLGandhi:
Wouldn't using `I` (with repeats for higher dims) be a sufficient definition?

DhairyaLGandhi:
I guess I'm trying to think of cases where having this is better over just using inbuilt functions

src/utils.jl
```diff
@@ -284,6 +284,92 @@ end
...
+    identity_init([rng=GLOBAL_RNG], dims...)
+
+Return an `Array` of size `dims` which yields an identity mapping when used as parameters in most Flux layers.
+
+Often useful in the context of transfer learning, i.e when one wants to add more capacity to a model but start from the same mapping.
+
+Has the following behaviour
+*  1D: A `Vector` of `zeros` (useful for an identity bias)
+*  2D: An identity matrix (useful for an identity matrix multiplication)
+*  More than 2D: A diagnoal matrix of identity kernels (useful for an identity convolution)
```
CarloLucibello, line 297:
```suggestion
*  More than 2D: A diagonal matrix of identity kernels (useful for an identity convolution) 
```

CarloLucibello: this is not really clear to me, can we expand a bit?

CarloLucibello:
> I guess I'm trying to think of cases where having this is better over just using inbuilt functions

first of all the case D>2 is not trivial, moreover you want to produce some learnable parameters, so passing `I` would not work

DrChainsaw:
> you want to produce some learnable parameters, so passing I would not work

Hmmm, I guess I approached this from the transfer learning/NAS case originally and there you ofc want trainable params.

Otoh, for the pixes shuffle case where this was brought up you explicitly want it to be non-trainable or else it won't be a "pixel shuffle" for long (I'm not familiar with pixel shuffle but from the name this seems to be the case)? Its also somewhat inelegant that for this case you end up crunching a bunch of zeros and ones when all you want is a no-op.

Perhaps this is a bit of an accidental sidestep from the original issue which perhaps warrants some other solution. I suppose one can put `I` where the current implementation puts `1` but this does not work out of the box, and even if it did, you'll have the runtime overhead of the zeros. Array of Bools perhaps, or will Zygote produce gradients for them?

I don't want to start another Zeros-debacle here, but it sounds like the pixel shuffle case could be handled by an IdentityWeight type, but I don't want to go there now :)

src/utils.jl
DrChainsaw: Yeah, reading it now I can see how this is confusing as the output is still a ND array and not a Matrix of arrays or anything. 

Perhaps it should be contracted to just say "ND array which makes conv layers no-ops" but expressed a bit more formally. The expanded version would require something similar to what @darsnack wrote above.

CarloLucibello:
> Hmmm, I guess I approached this from the transfer learning/NAS case originally and there you ofc want trainable params.

this is perfectly fine, allo our initializers return trainable parameters.
Whatever we are going to with pixel shuffle should be of no concern for this pr.

Seems the doctests are not passing

DrChainsaw:
> Seems that the doctests are not passing

Yup, I will try to attend to that tonight. I have never used Documenter and adding in and running `doctest(Flux)` didn't work out of the box, so I winged it from what I could discern from the previous CI failure. I might need to spend some time learning it which is something I anyways wanted to do someday.

Btw, big thumbs up to everyone for quick and good feedback!

DhairyaLGandhi:
> moreover you want to produce some learnable parameters, so passing I would not work

Couldn't we simply collect it? The eltypes can control if they're trainable or not. Bools have no grads either way.

```julia
 julia> collect(cat(I(3), I(3), dims = 3))
 3×3×2 Array{Bool,3}:
 [:, :, 1] =
  1  0  0
  0  1  0
  0  0  1
 
 [:, :, 2] =
  1  0  0
  0  1  0
  0  0  1
```

I'll anyway take this as a case where zeros with dims approach works out.

DrChainsaw:
> collect(cat(I(3), I(3), dims = 3))

This is basically what is done for 2D afaik, but the same approach does unfortunately not yeld identity when used as weight for conv layers.

The way Julia prints 3D arrays makes the output look confusing and irregular for 1D convolutions (3D arrays) but for 2D conv I think the required pattern is clear:

```julia
julia> Flux.identity_init(3,3,2,2)
3×3×2×2 Array{Float32, 4}:
[:, :, 1, 1] =
 0.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  0.0

[:, :, 2, 1] =
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

[:, :, 1, 2] =
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0

[:, :, 2, 2] =
 0.0  0.0  0.0
 0.0  1.0  0.0
 0.0  0.0  0.0
``` 

This could maybe be achieved somehow with collecting `I` across the right dimensions, but perhaps it would end up being more convoluted than the current implementation.

The eltype thing is perhaps worth considering though. I skipped passing a scaling parameter as that could just as easily be achieved by scaling the output of the function. I guess the same strategy is somewhat possible for changing eltype, but it is a bit more cumbersome.

DhairyaLGandhi:
The current output is a little surprising to me, maybe I need a better mental model but I'm possibly thinking of it as consistent with:

```julia
julia> collect(cat(I(3), I(3), dims = (3,4)))
3×3×2×2 Array{Bool,4}:
[:, :, 1, 1] =
 1  0  0
 0  1  0
 0  0  1

[:, :, 2, 1] =
 0  0  0
 0  0  0
 0  0  0

[:, :, 1, 2] =
 0  0  0
 0  0  0
 0  0  0

[:, :, 2, 2] =
 1  0  0
 0  1  0
 0  0  1
```

DrChainsaw:
> The current output is a little surprising to me, maybe I need a better mental model but I'm possibly thinking of it as consistent with

A good way to think about this particular example is that those diagonal matrices are not identity kernels as they will "mix up" the pixles they pass over: Upper left, middle and lower right will be added together in each step.

This wiki page is what made the idea behind conv layers click with me and it shows the identity kernel as the first example: https://en.wikipedia.org/wiki/Kernel_(image_processing)

As a side note, at least to me, it immediately becomes clear how some kernels are better at detecting certain features, and if you just slam the kernels with AD you can just force the computer to find the ones which are useful for your data set. It also gives a smooth evolution path from traditional image processing to DL image processing with conv layers.

darsnack:
> maybe I need a better mental model

Take a look at the discussion above. For the >2D case, you are probably initializing some kind of convolution kernel. Convincing with a bunch of identity matrices is not the identity operation for convolution.

src/utils.jl
darsnack: "A dense block array of center tap spatial filters (useful for an identity convolution)"

darsnack:
> Yup, I will try to attend to that tonight.

This might be a spacing issue. I noticed the copy paste for your `jldoctest` REPL code had an extra space in front. Documenter can be finicky about parsing those tests.

darsnack:
`scale`/`gain` might be a kwarg to add here (we have similar kwargs for the other init functions). For the case where you want to initialize to a certain gain but don't necessarily want to add a constant scaling to your function (assuming training).

DrChainsaw:
> scale/gain might be a kwarg to add here

This would be easy to add, but does it really add much over just scaling the output, i.e
```julia
Conv((3,3), 1=>1; init=(dims...) -> identity_init(dims...;gain=5))
```
vs
```julia
Conv((3,3), 1=>1; init=(dims...) -> 5 .* identity_init(dims...))
```

I suppose that if one also makes it set the eltype it could offer some extra utility at the cost of having users accidentally getting integer or float64 params because they forgot to put `f0` when they just wanted to scale. 

> I also wanted to add a dims keyword

Would this be used to e.g. set which dimensions are input/output and which are kernels for conv layers? I sense this could result in an inner platform API where it is easier to just write the function yourself than figuring out how to make use of it, but perhaps I don't understand the applications for it well enough.

DrChainsaw:
> So for the checks (given the inception case you described), I am inclined to suggest no checking at all

I personally found warnings to be a good middle ground here. Anecdotal statistics say its about 50/50 for me between "oh, well thats expected, I'll just remember to use the nowarn verion next time" and "crap, I forgot to XX, abort!".

darsnack:
Re: the gain

The way our init functions are written, you don't need to manually write out the closure.
```julia
Conv((3,3), 1=>1; init=identity_init(gain=5))
```
is the syntax we would end up having. Agree that it doesn't same much for this init function, but it would be consistent with the other init functions.

>>I also wanted to add a dims keyword
>
> I don't understand the applications for it well enough.

Me neither! Probably not worth it.

> warnings to be a good middle ground here

Is there a nice way to suppress warnings? I guess it's not too important for init functions unless you have a lot of them in a chain.

For most of the checks (i.e. `rows == cols`, `nins == nouts`), I don't think the check is even worth doing, because there exists a well-established, sensible thing to do when these checks fail (zero padding the larger dim). For the `all(isodd, kernel_dims)` check, I agree that a warning would be worth it here. Sorry to make you go back and forth on this. Maybe let's see what everyone else thinks? These are just my preferences, but I wouldn't oppose any of the alternatives being merged.

DrChainsaw:
> The way our init functions are written, you don't need to manually write out the closure.

Ah ok, that makes sense. I'll add it and scrap the asserts at the same time to avoid having to define the partially applied version of the nocheck/nowarn versions at the same time.

> Sorry to make you go back and forth on this. 

No worries. Thanks for the feedback.

src/utils.jl
```diff
@@ -284,6 +284,92 @@ end
...
+   0.0  0.0  0.0
+  
+  [:, :, 2, 3] =
+   0.0  0.0  0.0
+  
+  [:, :, 3, 3] =
+   0.0  1.0  0.0
+```
+"""
+identity_init(::AbstractRNG, dims...) = identity_init(dims...)
```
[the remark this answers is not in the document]
DrChainsaw: Done!

src/utils.jl
DrChainsaw: Changed to the wording from @darsnack

src/utils.jl
```diff
@@ -284,6 +284,88 @@ end
...
+most Flux layers.
+
+Often useful in the context of transfer learning, i.e when one wants to add more capacity to
+a model but start from the same mapping.
+
+Some caveats: Not all layers will be identity mapping when used with this init. Exceptions
+include recurrent layers, `DepthwiseConv` and normalization layers.
+
+Also note that layers must have `input_size == output_size` for identity mapping to be 
+possible.
```
darsnack, line 300:
```suggestion
Also note that layers must have `input_size == output_size` for identity mapping to be 
possible. When this is not the case, the "identity array" is padded with zeros.
```

src/utils.jl
```diff
@@ -284,6 +284,88 @@ end
...
+
+For convolutional layers, in addition to the above, the kernel sizes must also be odd and 
+padding must be applied so that output feature maps have the same size as input feature maps,
+e.g by using `SamePad`.
```
darsnack, line 304:
```suggestion
e.g by using [`SamePad`](@ref).
```

src/utils.jl
```diff
@@ -284,6 +284,88 @@ end
...
+ 0.0  0.0  0.0
+
+[:, :, 2, 2] =
+ 0.0  0.0  0.0
+ 0.0  1.0  0.0
+ 0.0  0.0  0.0
+```
+"""
+# Assume bias
+identity_init(cols;gain=1) = zeros(cols)
```
darsnack, line 348:
```suggestion
identity_init(cols; gain=1) = zeros(Float32, cols)
```

src/utils.jl
```diff
@@ -284,6 +284,88 @@ end
...
+
+# Assume matrix multiplication
+identity_init(rows, cols;gain=1) = Matrix{Float32}(I * gain, rows,cols)
```
darsnack, line 351:
```suggestion
identity_init(rows, cols; gain=1) = Matrix{Float32}(I * gain, rows,cols)
```

src/utils.jl
```diff
@@ -284,6 +284,88 @@ end
...
+
+# Assume convolution
+function identity_init(dims...;gain=1)
```
darsnack, line 354:
```suggestion
function identity_init(dims...; gain=1)
```

src/utils.jl
```diff
@@ -284,6 +284,88 @@ end
...
+  nin, nout = dims[end-1], dims[end]
+  centers = map(d -> cld(d, 2), dims[1:end-2])
+  weights = zeros(dims)
+  for i in 1:min(nin,nout)
+      weights[centers..., i, i] = gain
```
darsnack, line 359:
```suggestion
    weights[centers..., i, i] = gain
```

src/utils.jl
```diff
@@ -284,6 +284,88 @@ end
...
+identity_init(cols;gain=1) = zeros(cols)
+
+# Assume matrix multiplication
+identity_init(rows, cols;gain=1) = Matrix{Float32}(I * gain, rows,cols)
+
+# Assume convolution
+function identity_init(dims...;gain=1)
+  nin, nout = dims[end-1], dims[end]
+  centers = map(d -> cld(d, 2), dims[1:end-2])
+  weights = zeros(dims)
```
darsnack, line 357:
```suggestion
  weights = zeros(Float32, dims)
```

test/utils.jl
```diff
@@ -150,6 +150,48 @@ end
     @test maximum(partial_si(8, 8)) == 0
     @test maximum(partial_si(8, 8, sparsity=0)) > 0
   end
+
+  @testset "identity_init" begin
```
darsnack, line 154: Test that padded areas are zeros?

src/utils.jl
```diff
@@ -284,6 +284,88 @@ end
 sparse_init(dims...; kwargs...) = sparse_init(Random.GLOBAL_RNG, dims...; kwargs...)
 sparse_init(rng::AbstractRNG; init_kwargs...) = (dims...; kwargs...) -> sparse_init(rng, dims...; init_kwargs..., kwargs...)
 
+"""
+    identity_init([rng=GLOBAL_RNG], dims...;gain=1)
```
darsnack, line 288:
```suggestion
    identity_init([rng=GLOBAL_RNG], dims...; gain=1)
```

darsnack (changes requested):
Looking near ready for me. Just some small changes to match the repo coding style and use `Float32` as the return type. Also would like others to review the docstring to make sure everything is as clear as possible.

test/utils.jl
DrChainsaw: Yes, why not. Will add later.

darsnack: Thanks

src/utils.jl
```diff
@@ -284,6 +284,88 @@ end
...
+  centers = map(d -> cld(d, 2), dims[1:end-2])
+  weights = zeros(Float32, dims)
+  for i in 1:min(nin,nout)
+    weights[centers..., i, i] = gain
+  end
+  return weights
+end
+
+identity_init(::AbstractRNG, dims...;kwargs...) = identity_init(dims...;kwargs...)
+identity_init(;init_kwargs...) = identity_init(Random.GLOBAL_RNG; init_kwargs...)
```
CarloLucibello, line 365:
```suggestion
identity_init(::AbstractRNG, dims...; kwargs...) = identity_init(dims...; kwargs...)
identity_init(; init_kwargs...) = identity_init(Random.GLOBAL_RNG; init_kwargs...)
```

test/utils.jl
DrChainsaw: Added

DrChainsaw:
Added kwarg `shift` which applies a circshift to the output intended to be useful with partial application (see the inception test case). Maybe a bit too much given that identity inception blocks perhaps isn't the most common use case?

src/utils.jl
```diff
@@ -284,6 +284,90 @@ end
...
+"""
+    identity_init([rng=GLOBAL_RNG], dims...; gain=1, shift=0)
+
+Return an `Array` of size `dims` which yields an identity mapping when used as parameters in 
+most Flux layers. Use `gain` to get other scalings.  
+
+Often useful in the context of transfer learning, i.e when one wants to add more capacity to
+a model but start from the same mapping.
+
+Use `shift` (integer or tuple) to apply circular shift to the output.
```
darsnack, line 296:
```suggestion
Use `shift` (integer or tuple) to apply circular shift to the output. Equivalent to `Base.circshift(identity(dims...), shift)`
```

darsnack: Seems fine to add this keyword to me. It's not like there is too much complexity to this function's behavior already.

src/utils.jl
```diff
@@ -284,6 +284,91 @@ end
 sparse_init(dims...; kwargs...) = sparse_init(Random.GLOBAL_RNG, dims...; kwargs...)
 sparse_init(rng::AbstractRNG; init_kwargs...) = (dims...; kwargs...) -> sparse_init(rng, dims...; init_kwargs..., kwargs...)
 
+"""
+    identity_init([rng=GLOBAL_RNG], dims...; gain=1, shift=0)
+
+Return an `Array` of size `dims` which yields an identity mapping when used as parameters in 
+most Flux layers. Use `gain` to get other scalings.
```
darsnack, line 291:
```suggestion
most Flux layers. Use `gain` to scale the identity by a constant.
```

darsnack (approved):
This is good for me. Since it is a new API, I'll let others weigh in.

DhairyaLGandhi (dismissed):
Seems fine to have to get the ball rolling.

darsnack (approved):

PR #1527, opened 2021-03-07
preparatory work for v0.12

CarloLucibello:


CarloLucibello:
Hard to pinpoint the perf regressions if you don't share something reproducible. I suspect any regressions is related to CUDA.jl, so little we can do here, but let's see when you share the script (or some reduced examples). 

> Objectdetector - used to be faster than darknet - fails to load - and is now slower after fixing

again, this should be pinpointed to some specific layer/function. Failure to load probably due to using stuff at the end of the deprecation cycle (e.g. losses are not exported from Flux anymore, should get them from Flux.Losses). You should open an issue/pr there. 

Model-zoo vision's examples are working properly and I could see any noticeable perf regression. 

> Torch.jl - uses Flux.batchnorm (same as CUDA until recently or even till today) and there's no replacement

There is no such thing as Flux.batchnorm, if torch Torch.jl is using CUDA.CUDNN.batchnorm as long as it keeps its compat bound it will be fine (same as flux), it won't be removed till CUDA 3.0.0 is released.

> EquivariantNetworks can't use StaticArrays in some kernels now.

You should open an issue with a MWE

> The changes made to CI are appreciated, but the Manifest change can be reserved for the time being. We use it for consistency and development anyway. We want our CI runs to be meaningful (see the Group Conv for eg)

Problem is with a Manifest I cannot test with CUDA 2.4 and 2.6 respectively on julia 1.5 and 1.6. 
CI runs are very meaningful, you test with the latest versions of dependencies. The largest part of julia's packages don't use a manifest I don't see how why we should do something different that has no pros and a few cons in our situation.  

> The change to tests is curious, since we shouldn't be passing tests with that. 

why not, `initW` has been deprecated for `init`

> Same was the case with the normalization PR which still needs to pass tests before merging for 0.12.

since that pr has been open for more than a month, it would be nice if you could wrap it up and avoid derailing it with questionable changes (read bugs) to the activation state and to the returned parameters. In any case, that shouldn't block v0.12 since it is an internal refactoring.

DhairyaLGandhi:
The for the tests is pretty standard; https://gist.github.com/DhairyaLGandhi/12d06b9389e13130cf4c03c67885be91

darsnack:
The PR seems fine to me, but let's address the regressions. Either we use `ModelZooKeeper` or need to put together a temporary repo with all the tests (so that we can all debug).

CarloLucibello:
> The tests are pretty standard and can be found here; https://gist.github.com/DhairyaLGandhi/12d06b9389e13130cf4c03c67885be91

this something that cannot be run (e.g. can't install metalhead). Also, there is a subtle bug, during the benchmarks new models are created, but gradients are computed with respect to old params, so the gradient is nothing. 

I fixed the script and cannot see any performance difference between master and latest release. 

> The PR seems fine to me, but let's address the regressions. Either we use ModelZooKeeper or need to put together a temporary repo with all the tests (so that we can all debug).

Regression should be addressed, but I cannot see any regression. If someone (@DhairyaLGandhi ?) claims to have spotted any bug or per regression, he should open a corresponding issue and provide a reproducible example to be helpful. 

I'm opening a pr adding a benchmark script to a perf/ folder here, but definitely, we should think of a longer-term solution.

CarloLucibello:
In any case, this PR can be merged, release is not supposed to be tagged right now

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.12.0-dev"
+version = "0.12.0"
```
darsnack, line 3: Do we leave this as is? Can do a commit to change it and tag that commit for release?

darsnack (approved):
If this isn't a PR to tag a release, then the changes are fine.

DhairyaLGandhi:
Thanks for pointing out the bug. I still see numbers similar to what I reported earlier after fixing it. The script does require adding the appropriate bounds to Metalhead in a local clone (and possibly removing the dev from the version number in the Flux clone). The gradients wouldn't have affected the runtime anyway.

darsnack:
Even without #1535, let's put these benchmarks in a repo? Can just be quick and dirty so that we can all commit/clone the same code for testing. Doesn't have to be a permanent solution, just something to make us all more comfortable about the release and on the same page.

DhairyaLGandhi:
Flux is also not compatible with CUDA#master on 1.6

CarloLucibello:
yeah this is known and nothing that concerns v0.12. CUDA's next release will be breaking. We have to register NNlibCUDA

PR #1529, opened 2021-03-08
Added equality comparison for layers

### LINKED ISSUE #1012: Layers and Params should support equality
For testing purposes I wanted to check if two models were identical.
(Making sure I was doing the right thing with my seeds)

Right now there is no easy way to do this.
See what I tried: (these tests are using on old version

```
julia> Random.seed!(1);

julia> x = Dense(10,10)
Dense(10, 10)

julia> Random.seed!(1);

julia> y = Dense(10,10)
Dense(10, 10)

julia> x==y
false

julia> params(x)
Params([Float32[0.052575205 0.44167045 … -0.44290894 0.115914196; 0.34975964 -0.0914253 … 0.08449556 0.14750573; … ; 0.2187416 0.2876231 … 0.108851254 0.32387242; 0.074596524 0.13426614 … -0.2106525 0.4129367], Float32[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]])

julia> params(x) == params(y)
false

julia> params(y)
Params([Float32[0.052575205 0.44167045 … -0.44290894 0.115914196; 0.34975964 -0.0914253 … 0.08449556 0.14750573; … ; 0.2187416 0.2876231 … 0.108851254 0.32387242; 0.074596524 0.13426614 … -0.2106525 0.4129367], Float32[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]])

julia> collect(params(y)) == collect(params(x))
true
```

VasanthManiVasi:
Fixes #1012 for layers where comparison is usually done. Happy to add the rest of the layers right away if it is required.

```julia
Random.seed!(1)
x = Dense(10,10)
Random.seed!(1)
y = Dense(10,10)
println(x == y) # true

y = Dense(10, 10, relu)
println(x == y) # false
```
### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -706,3 +706,8 @@ function Base.show(io::IO, m::MeanPool)
   m.stride == m.k || print(io, ", stride=", _maybetuple_string(m.stride))
   print(io, ")")
 end
+
+# define equality comparison for the layers
+function Base.:(==)(x::T, y::T) where {T <: Union{Conv, ConvTranspose, DepthwiseConv, CrossCor}}
+  params(x) == params(y)
+end
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -40,6 +40,8 @@ trainable(a::Recur) = (a.cell,)
 
 Base.show(io::IO, m::Recur) = print(io, "Recur(", m.cell, ")")
 
+Base.:(==)(x::Recur, y::Recur) = params(x) == params(y)
+
 """
     reset!(rnn)
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -422,3 +422,9 @@ function Base.show(io::IO, m::Parallel)
   join(io, m.layers, ", ")
   print(io, ")")
 end
+
+
+# define equality comparison for the layers
+function Base.:(==)(x::T, y::T) where {T <: Union{Chain, Dense}}
+  params(x) == params(y)
+end


src/layers/conv.jl
```diff
@@ -706,3 +706,8 @@ function Base.show(io::IO, m::MeanPool)
   m.stride == m.k || print(io, ", stride=", _maybetuple_string(m.stride))
   print(io, ")")
 end
+
+# define equality comparison for the layers
+function Base.:(==)(x::T, y::T) where {T <: Union{Conv, ConvTranspose, DepthwiseConv, CrossCor}}
```
ToucheSir, line 711: Is there any reason for this to be duplicated instead of defined once after all the relevant layers are?

src/layers/recurrent.jl
```diff
@@ -40,6 +40,8 @@ trainable(a::Recur) = (a.cell,)
 
 Base.show(io::IO, m::Recur) = print(io, "Recur(", m.cell, ")")
 
+Base.:(==)(x::Recur, y::Recur) = params(x) == params(y)
```
ToucheSir, line 43: Ditto here WRT duplication

ToucheSir (commented):
I'm generally in favour of the idea, but the special-casing seems brittle at best. Is the union really necessary/is there a more general way to do this?

ToucheSir:
Another thing I thought of: `params` relies on `trainable`, but some layers override it to return a subset of parameters (e.g. https://github.com/FluxML/Flux.jl/blob/master/src/layers/normalise.jl#L265). As such I'm not sure it's robust enough to be used for general layer comparisions.

VasanthManiVasi:
The layers are written in separate files, so the definition had to be duplicated for each file. @ToucheSir 
I wanted to ask if it would be fine to define the function in src/Flux.jl, since it is the common file which has all the layers in scope. 
```julia
# src/Flux.jl
function Base.:(==)(x::T, y::T) where {T <: Union{Chain, Dense Conv, ConvTranspose, DepthwiseConv, CrossCor, Recur}}
  params(x) == params(y)
end
```

The general way to do this would be to define a `layers.jl` with an abstract type `Layer` and define the `==` function for it. 
```julia
Base.:(==)(x::Layer, y::Layer) = params(x) == params(y)
```
Other layers would then subtype from `Layer` and inherit the `==` function.
It would be an efficient design in the long run since we could define various default layer functions for `Layer`. It would also make defining custom layer types quite easy.

DhairyaLGandhi:
The issue is also that params being equal may not mean the layers are equal. Consider a graph convolution layer with an adjacency matrix and a bias, or an embedding style layer. It's possible that it's params would be equal to that of another Dense, but would you say they are equal?

DhairyaLGandhi:
Thank you for the contribution and the discussion!

We may need generally need an idea of the real benefit of such definitions. Are there use cases where this is critical/ if we are to define equality on params, why not let users do it directly. From an api perspective, it's quite similar to do `params(l1) == params(l2)` as it is to say `l1 == l2` but avoids the issue with similar enough looking params.

VasanthManiVasi:
In the current implementation, we are able to compare only layers of the same types, and the comparison between a Conv and a Dense would default to `==(x, y) in Base at operators.jl` and return false.
A default layer abstract type would surely make the error you've mentioned! @DhairyaLGandhi 
It can be restricted to comparing only layers of a same type so that such errors are not made.

`==` for comparison is instinctive IMO, even though `params(l1) == params(l2)` works the same. It would be helpful to quickly compare two models or layers.

ToucheSir:
To add onto Dhairya's points, equality on layers is a difficult problem because not everything that could be used to determine it is exposed via an API. For example, `BatchNorm(1, relu) == BatchNorm(1, gelu)` if one only compares the params because the activation is not included in them.

Now, the question is what level of generality is required to implement a robust equals method for all Flux layers? Bit-for-bit equality (i.e. `memcmp`) is probably out of the question because of reference types, but some form of structural equality that can recursively follow + compare the underlying data would probably work. This can probably be made less expensive and general purpose, but I can't think of how to do so off the top of my head. There's also the question of how cycles and tied weights are handled, though I suppose that becomes less of a problem if we only implement equality for layers without children.

CarloLucibello:
Closing as  this doesn't seem something we want, feel free to reopen otherwise

PR #1531, opened 2021-03-10
Add grouped convolution

DhairyaLGandhi:
Add support for grouped convolutions. Needed for a bunch of things. Depends on an NNlib branch where the support is present in the Dims Helpers, but needs to be chained through to the conv kernel. Although we now can generate the correct shaped kernels and filters here.


lorenzoh:
Do the grouped convolutions run on the GPU as well?

DhairyaLGandhi:
This PR adds the necessary changes to the dims objects that we can use to pass the groups to CuDNN.

DhairyaLGandhi:
Seems like we are good to merge https://github.com/FluxML/NNlib.jl/pull/289

test/runtests.jl
```diff
@@ -15,9 +15,9 @@ end
   include("onehot.jl")
 end
 
-@testset "Optimise" begin
-  include("optimise.jl")
-end
+# @testset "Optimise" begin
+#   include("optimise.jl")
+# end
```
CarloLucibello, line 20: ?

CarloLucibello:
Needs some tests. Manifest can be removed

test/runtests.jl
DhairyaLGandhi:
```julia
  MethodError: no method matching push!(::Zygote.IdSet{Union{}})
  Closest candidates are:
    push!(::Zygote.IdSet{T}, ::T) where T at /home/dhairyalgandhi/.julia/packages/Zygote/JBaiY/src/tools/idset.jl:10
    push!(::Any, ::Any, ::Any) at abstractarray.jl:2387
    push!(::Any, ::Any, ::Any, ::Any...) at abstractarray.jl:2388
    ...
  Stacktrace:
    [1] Zygote.IdSet{Union{}}(xs::Vector{Union{}})
      @ Zygote ~/.julia/packages/Zygote/JBaiY/src/tools/idset.jl:14
    [2] Zygote.IdSet(xs::Vector{Union{}})
      @ Zygote ~/.julia/packages/Zygote/JBaiY/src/tools/idset.jl:16
    [3] Params(xs::Vector{Union{}})
```

Flux is currently broken. I noticed you'd set up a tag, but I believe this would need to be rectified before that so I've held it off until then. Hopefully its not a major deal to fix.

CarloLucibello: Is this a Flux problem or a Zygote one? Can you open an issue?

DhairyaLGandhi:
Going by how many releases of NNlib were needed to get the Embedding layer merged into Flux, it seemed to me that development could have been a lot faster and releases a lot safer had a Manifest been used. Releases aren't free since they affect unsuspecting users the most.

CarloLucibello:
development with a Manifest would have gone at exactly the same speed. Something that did slow down a bit development is the separation of NNlib and NNlibCUDA. Releases are free as long as they are non-breaking, they don't affect users that don't use the new features

CarloLucibello:
Anyways, Manifest should not be discussed here, this PR is meant to add groups support not to reinstate the manifest, let's not derail the PR

DhairyaLGandhi:
Agreed, that I don't want to derail the PR but:

> development with a Manifest would have gone at exactly the same speed.

A manifest would have shown the issues with every CI run, without having to make disruptive releases. Not to mention the time it takes to make them, and the buggy code that is pushed out to users

> don't affect users that don't use the new features

Features are features, new or old. Sure bug fixes will come in faster for new ones than the old ones but if it is known that something isn't quite right, then why release?

> the separation of NNlib and NNlibCUDA

The issue is the existence of the separation. The release schedule would have been exactly the same even if in the same repo, and with a Manifest, the gap of where packages live is made significantly smaller.

test/runtests.jl
DhairyaLGandhi: I will check that part out.

DhairyaLGandhi: Was this not issue not checked before a release was requested?

CarloLucibello: Hard to check something that only you observe, especially if you don't file an issue. 
We had all greens in the last PRs

CarloLucibello:
> Agreed, that I don't want to derail the PR but:

but then you do. I disagree with everything you said, but again, this is not the right place to have that discussion

test/runtests.jl
DhairyaLGandhi: https://buildkite.com/julialang/flux-dot-jl/builds/1394

CarloLucibello: this should be fixed now

CarloLucibello (approved):

CarloLucibello:
a news entry would be good

PR #1545, opened 2021-03-24
Add early stopping utils

queensferryme:
This pull request introduces a utility function `early_stopping(f; min_delta=0, patience=3)` for conveniently implementing early stopping. Its motive was discussed in #227 back in 2018. Also AFAIU, early stopping is widely adopted in training large-scale neural networks nowadays as a mechanism to prevent overfitting. Therefore, I believe a built-in utility function for early stopping would be beneficial for Flux.jl users.

The implementation is heavily based on [tf.keras.callbacks.EarlyStopping](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/EarlyStopping) and [ignite.handlers.EarlyStopping](https://github.com/pytorch/ignite/blob/master/ignite/handlers/early_stopping.py).

> This is a _draft_ implementation. Documentation and tests are not added yet. As I am new to Flux.jl, I think advice and opinions from maintainers would be helpful before I proceed.

**Example Usage**:
```julia
# test_ea.jl
using Flux

function loss()
    v = 1
    return () -> v += 1
end

ea = Flux.early_stopping(loss(), patience=5)
Flux.@epochs 10 begin
    ea() || break
end
```

**Output**:

```
➜ julia test_ea.jl
 Activating environment at `~/Developer/Flux.jl/Project.toml`
[ Info: Epoch 1
[ Info: Epoch 2
[ Info: Epoch 3
[ Info: Epoch 4
[ Info: Epoch 5
[ Info: Epoch 6
➜
```

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/utilities.md
+++ b/docs/src/utilities.md
@@ -102,5 +102,5 @@ Flux.nfan
 Flux.throttle
 Flux.stop
 Flux.skip
-Flux.patience_counter
+Flux.plateau
 ```
--- a/docs/src/utilities.md
+++ b/docs/src/utilities.md
@@ -102,4 +102,5 @@ Flux.nfan
 Flux.throttle
 Flux.stop
 Flux.skip
+Flux.early_stopping
 ```
--- a/docs/src/utilities.md
+++ b/docs/src/utilities.md
@@ -102,5 +102,5 @@ Flux.nfan
 Flux.throttle
 Flux.stop
 Flux.skip
-Flux.early_stopping
+Flux.patience_counter
 ```
--- a/docs/src/utilities.md
+++ b/docs/src/utilities.md
@@ -121,10 +121,10 @@ end
 
 # create an early stopping trigger
 # returns true when the loss increases for two consecutive steps
-es = Flux.early_stopping(loss, 2; init_score=9)
+es = early_stopping(loss, 2; init_score = 9)
 
 # this will stop at the 6th (4 decreasing + 2 increasing calls) epoch
-Flux.@epochs 10 begin
+@epochs 10 begin
   es() && break
 end
 ```
@@ -138,33 +138,33 @@ acc = let v = 0
 end
 
 # create an early stopping trigger for accuracy
-es = Flux.early_stopping(acc, 3; delta = (best_score, score) -> score - best_score)
+es = early_stopping(acc, 3; delta = (best_score, score) -> score - best_score)
 
 # this will iterate until the 10th epoch
-Flux.@epochs 10 begin
+@epochs 10 begin
   es() && break
 end
 ```
 
 `early_stopping` and `plateau` are both built on top of `patience`. You can use `patience` to build your own triggers that use a patient counter. For example, if you want to trigger when the loss is below a threshold for several consecutive iterations:
 ```julia
-threshold(f, thresh, delay) = Flux.patience(delay) do
+threshold(f, thresh, delay) = patience(delay) do
   f() < thresh
 end
 ```
 
 Both `predicate` in `patience` and `f` in `early_stopping` / `plateau` can accept extra arguments. You can pass such extra arguments to `predicate` or `f` through the returned function:
 ```julia
-trigger = Flux.patience((a; b) -> a > b, 3)
+trigger = patience((a; b) -> a > b, 3)
 
 # this will iterate until the 10th epoch
-Flux.@epochs 10 begin
-  trigger(1; b=2) && break
+@epochs 10 begin
+  trigger(1; b = 2) && break
 end
 
 # this will stop at the 3rd epoch
-Flux.@epochs 10 begin
-  trigger(3; b=2) && break
+@epochs 10 begin
+  trigger(3; b = 2) && break
 end
 ```
 
--- a/docs/src/utilities.md
+++ b/docs/src/utilities.md
@@ -106,41 +106,63 @@ Flux.skip
 
 ## Patience Helpers
 
-Flux provide utilities for scheduling your workflow according to some monitored condition and a maximum `patience`. For example, you can use `early_stopping` to stop training when the model is converging or deteriorating, or you can use `plateau` to check if the model is stagnating.
+Flux provides utilities for controlling your training procedure according to some monitored condition and a maximum `patience`. For example, you can use `early_stopping` to stop training when the model is converging or deteriorating, or you can use `plateau` to check if the model is stagnating.
 
-The keyword argument `distance` of `early_stopping` is a function of the form `distance(best_score, score)`. By default `distance` is `-`, which implies that the monitored metric `f` is expected to be decreasing and mimimized. If you use some increasing metric (e.g. accuracy), you can customize the `distance` function: `(best_score, score) -> score - best_score`.
-
+For example, below we create a pseudo-loss function that decreases, bottoms out, then increases. The early stopping trigger will break the loop before the loss increases too much.
 ```julia
-acc = let v = 0
-  () -> v = max(1, v + 0.01)
+# create a pseudo-loss that decreases for 4 calls, then starts increasing
+# we call this like loss()
+loss = let t = 0
+  () -> begin
+    t += 1
+    (t - 4) ^ 2
+  end
 end
 
-es = Flux.early_stopping(acc, 3; delta = (best_score, score) -> score - best_score)
+# create an early stopping trigger
+# returns true when the loss increases for two consecutive steps
+es = Flux.early_stopping(loss, 2; init_score=9)
 
-# This will iterate until the 10th epoch
-Flux.@epochs 10 begin
-  es() && break
-end
-
-es = Flux.early_stopping(acc, 3)
-
-# This will stop at the 3rd epoch
+# this will stop at the 6th (4 decreasing + 2 increasing calls) epoch
 Flux.@epochs 10 begin
   es() && break
 end
 ```
 
-Both `predicate` in `patience` and `f` in `early_stopping` / `plateau` can accept extra arguments. You can pass such extra arguments to `predicate` or `f` through the returned function:
+The keyword argument `distance` of `early_stopping` is a function of the form `distance(best_score, score)`. By default `distance` is `-`, which implies that the monitored metric `f` is expected to be decreasing and mimimized. If you use some increasing metric (e.g. accuracy), you can customize the `distance` function: `(best_score, score) -> score - best_score`.
+```julia
+# create a pseudo-accuracy that increases by 0.01 each time from 0 to 1
+# we call this like acc()
+acc = let v = 0
+  () -> v = max(1, v + 0.01)
+end
 
+# create an early stopping trigger for accuracy
+es = Flux.early_stopping(acc, 3; delta = (best_score, score) -> score - best_score)
+
+# this will iterate until the 10th epoch
+Flux.@epochs 10 begin
+  es() && break
+end
+```
+
+`early_stopping` and `plateau` are both built on top of `patience`. You can use `patience` to build your own triggers that use a patient counter. For example, if you want to trigger when the loss is below a threshold for several consecutive iterations:
+```julia
+threshold(f, thresh, delay) = Flux.patience(delay) do
+  f() < thresh
+end
+```
+
+Both `predicate` in `patience` and `f` in `early_stopping` / `plateau` can accept extra arguments. You can pass such extra arguments to `predicate` or `f` through the returned function:
 ```julia
 trigger = Flux.patience((a; b) -> a > b, 3)
 
-# This will iterate until the 10th epoch
+# this will iterate until the 10th epoch
 Flux.@epochs 10 begin
   trigger(1; b=2) && break
 end
 
-# This will stop at the 3rd epoch
+# this will stop at the 3rd epoch
 Flux.@epochs 10 begin
   trigger(3; b=2) && break
 end
--- a/docs/src/utilities.md
+++ b/docs/src/utilities.md
@@ -102,11 +102,13 @@ Flux.nfan
 Flux.throttle
 Flux.stop
 Flux.skip
-Flux.patience
-Flux.plateau
 ```
 
-The keyword argument `delta` of [`plateau`](@ref) is a function of the form `delta(best_score, current_score)`. By default `delta` is `-`, which implies that the metric `f` is expected to be decreasing and mimimized. If you use some increasing metric (e.g. accuracy), you can customize the `delta` function: `(best_score, score) -> score - best_score`.
+## Patience Helpers
+
+Flux provide utilities for scheduling your workflow according to some monitored condition and a maximum `patience`. For example, you can use `early_stopping` to stop training when the model is converging or deteriorating, or you can use `plateau` to check if the model is stagnating.
+
+The keyword argument `distance` of `early_stopping` is a function of the form `distance(best_score, score)`. By default `distance` is `-`, which implies that the monitored metric `f` is expected to be decreasing and mimimized. If you use some increasing metric (e.g. accuracy), you can customize the `distance` function: `(best_score, score) -> score - best_score`.
 
 ```julia
 acc = let v = 0
@@ -128,7 +130,7 @@ Flux.@epochs 10 begin
 end
 ```
 
-Both `predicate` in [`patience`](@ref) and `f` in [`plateau`](@ref) can accept extra arguments. You can pass such extra arguments to `predicate` or `f` through the returned function:
+Both `predicate` in `patience` and `f` in `early_stopping` / `plateau` can accept extra arguments. You can pass such extra arguments to `predicate` or `f` through the returned function:
 
 ```julia
 trigger = Flux.patience((a; b) -> a > b, 3)
@@ -142,4 +144,10 @@ end
 Flux.@epochs 10 begin
   trigger(3; b=2) && break
 end
+```
+
+```@docs
+Flux.patience
+Flux.early_stopping
+Flux.plateau
 ```
--- a/docs/src/utilities.md
+++ b/docs/src/utilities.md
@@ -102,5 +102,6 @@ Flux.nfan
 Flux.throttle
 Flux.stop
 Flux.skip
+Flux.patience
 Flux.plateau
 ```
--- a/docs/src/utilities.md
+++ b/docs/src/utilities.md
@@ -105,3 +105,41 @@ Flux.skip
 Flux.patience
 Flux.plateau
 ```
+
+The keyword argument `delta` of [`plateau`](@ref) is a function of the form `delta(best_score, current_score)`. By default `delta` is `-`, which implies that the metric `f` is expected to be decreasing and mimimized. If you use some increasing metric (e.g. accuracy), you can customize the `delta` function: `(best_score, score) -> score - best_score`.
+
+```julia
+acc = let v = 0
+  () -> v = max(1, v + 0.01)
+end
+
+es = Flux.early_stopping(acc, 3; delta = (best_score, score) -> score - best_score)
+
+# This will iterate until the 10th epoch
+Flux.@epochs 10 begin
+  es() && break
+end
+
+es = Flux.early_stopping(acc, 3)
+
+# This will stop at the 3rd epoch
+Flux.@epochs 10 begin
+  es() && break
+end
+```
+
+Both `predicate` in [`patience`](@ref) and `f` in [`plateau`](@ref) can accept extra arguments. You can pass such extra arguments to `predicate` or `f` through the returned function:
+
+```julia
+trigger = Flux.patience((a; b) -> a > b, 3)
+
+# This will iterate until the 10th epoch
+Flux.@epochs 10 begin
+  trigger(1; b=2) && break
+end
+
+# This will stop at the 3rd epoch
+Flux.@epochs 10 begin
+  trigger(3; b=2) && break
+end
+```
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -742,3 +742,15 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
 isleaflike(x) = Functors.isleaf(x)
 isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
+
+function early_stopping(f; min_delta=0, patience=3)
+  min_metric = Inf
+  count = 0
+
+  return function ()
+      metric = f()
+      count = min_metric - metric < min_delta ? count + 1 : 0
+      min_metric = min(min_metric, metric)
+      return count < patience
+  end
+end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -770,7 +770,7 @@ julia> function loss()
        return () -> l += 1
        end; # pseudo loss function that returns increasing values
 
-julia> es = Flux.early_stopping(loss(); patience=3);
+julia> es = Flux.early_stopping(loss(), 3);
 
 julia> Flux.@epochs 10 begin
        es() && break
@@ -780,7 +780,7 @@ julia> Flux.@epochs 10 begin
 [ Info: Epoch 3
 ```
 """
-function patience_counter(f; delta = -, init_score = 0, min_delta = 0, patience = 3)
+function patience_counter(f, patience; delta = -, init_score = 0, min_delta = 0)
   best_score = init_score
   count = 0
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -749,19 +749,18 @@ isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
 Return a function that evaluates the metric `f` and compares its value
 against its value on last invocation. When the difference has been less
 than `min_delta` at least `patience` times, `true` is returned,
-otherwise `false` is returned. The function returned by `patience_counter`
-accepts a variable number of arguments and keyword arguments, which will then
-be passed to the metric function `f`  when being evaluated.
+otherwise `false` is returned. If `f` accepts arguments,
+then the returned function respects that.
 
 The difference is measured by keyword argument, `delta`.
-By default, `early_stopping` expects the metric `f` to be minimized.
+By default, `patience_counter` expects the metric `f` to be minimized.
 If you are using some increasing metric (e.g. accuracy),
 you can customize the `delta` function:
 `(best_score, score) -> score - best_score` (for increasing metrics).
 
 A common use case of `patience_counter` is early stopping. For this,
-we have added `early_stopping` as an alias to `patience_counter`. But
-please note that you can do more generic things with `patience_counter`,
+we have added `early_stopping` as an alias to `patience_counter`.
+Note that you can do more generic things with `patience_counter`,
 for example reducing the learning rate when the training loss plateaus.
 
 # Examples
@@ -779,18 +778,13 @@ julia> Flux.@epochs 10 begin
 [ Info: Epoch 1
 [ Info: Epoch 2
 [ Info: Epoch 3
-[ Info: Epoch 4
 ```
 """
-function patience_counter(f; delta = -, min_delta = 0, patience = 3)
-  best_score = NaN
+function patience_counter(f; delta = -, init_score = 0, min_delta = 0, patience = 3)
+  best_score = init_score
   count = 0
 
   return function (args...; kwargs...)
-    if isnan(best_score)
-      best_score = f(args...; kwargs...)
-      return false
-    end
     score = f(args...; kwargs...)
     Δ = delta(best_score, score)
     count = Δ < min_delta ? count + 1 : 0
@@ -799,4 +793,4 @@ function patience_counter(f; delta = -, min_delta = 0, patience = 3)
   end
 end
 
-early_stopping = patience_counter
+early_stopping(args...; kwargs...) = patience_counter(args...; kwargs...)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -748,13 +748,14 @@ isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
 
 Return a function that evaluates the metric `f` and compares its value
 against its value on last invocation. When the difference has been less
-than `min_delta` for at least `patience` times, `true` is returned,
+than `min_delta` at least `patience` times, `true` is returned,
 otherwise `false` is returned.
 
+The difference is measured by keyword argument, `delta`.
 By default, `early_stopping` expects the metric `f` to be minimized.
-However, if you are using some increasing metric, accuracy for example,
-you can change `early_stopping`'s behaviour by customizing the `delta`
-function: `(best_score, score) -> score - best_score`.
+If you are using some increasing metric (e.g. accuracy),
+you can customize the `delta` function:
+`(best_score, score) -> score - best_score` (for increasing metrics).
 
 # Examples
 ```jldoctest
@@ -767,7 +768,7 @@ loss (generic function with 1 method)
 julia> es = Flux.early_stopping(loss(); patience=3);
 
 julia> Flux.@epochs 10 begin
-       es() || break
+       es() && break
        end
 [ Info: Epoch 1
 [ Info: Epoch 2
@@ -783,6 +784,6 @@ function early_stopping(f; delta = -, min_delta = 0, patience = 3)
     Δ = delta(best_score, score)
     count = Δ < min_delta ? count + 1 : 0
     best_score = Δ < 0 ? best_score : score
-    return count < patience
+    return count >= patience
   end
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -744,22 +744,27 @@ isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
 
 """
-    early_stopping(f; min_delta=0, patience=3)
+    early_stopping(f; delta=-, min_delta=0, patience=3)
 
 Return a function that evaluates the metric `f` and compares its value
-against that of last invokation. When the difference has been less
-than `min_delta` for at least `patience` times, `true` is returned;
+against its value on last invocation. When the difference has been less
+than `min_delta` for at least `patience` times, `true` is returned,
 otherwise `false` is returned.
+
+By default, `early_stopping` expects the metric `f` to be minimized.
+However, if you are using some increasing metric, accuracy for example,
+you can change `early_stopping`'s behaviour by customizing the `delta`
+function: `(best_score, score) -> score - best_score`.
 
 # Examples
 ```jldoctest
 julia> function loss()
-       l = 1
+       l = 0
        return () -> l += 1
-       end
+       end # pseudo loss function that returns increasing values
 loss (generic function with 1 method)
 
-julia> es = Flux.early_stopping(loss());
+julia> es = Flux.early_stopping(loss(); patience=3);
 
 julia> Flux.@epochs 10 begin
        es() || break
@@ -770,14 +775,15 @@ julia> Flux.@epochs 10 begin
 [ Info: Epoch 4
 ```
 """
-function early_stopping(f; min_delta = 0, patience = 3)
-  min_metric = Inf
+function early_stopping(f; delta = -, min_delta = 0, patience = 3)
+  best_score = f()
   count = 0
 
   return function ()
-      metric = f()
-      count = min_metric - metric < min_delta ? count + 1 : 0
-      min_metric = min(min_metric, metric)
-      return count < patience
+    score = f()
+    Δ = delta(best_score, score)
+    count = Δ < min_delta ? count + 1 : 0
+    best_score = Δ < 0 ? best_score : score
+    return count < patience
   end
 end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -747,13 +747,13 @@ isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
     patience(predicate, patience)
 
 Return a function that internally counts by one when
-`predicate(...) == true`, otherwise the count is reset
-to zero. If the count is greater than or equal to `patience`,
+`predicate(...) == true`, otherwise the count is reset to zero.
+If the count is greater than or equal to `patience`,
 the function returns `true`, otherwise it returns `false`.
 
 # Examples
 ```jldoctest
-julia> loss = () -> rand();
+julia> loss() = rand();
 
 julia> trigger = Flux.patience(() -> loss() < 1, 3);
 
@@ -775,21 +775,15 @@ function patience(predicate, patience)
   end
 end
 
-
 """
-    plateau(f, p; delta = -, init_score = 0, min_delta = 0)
+    early_stopping(f, delay; distance = -, init_score = 0, min_dist = 0)
 
 Return a function that internally counts by one when
-`delta(best_score, f(...)) <= min_delta`, where
+`distance(best_score, f(...)) <= min_dist`, where
 `best_score` is the last seen best value of `f(...)`.
-If the count is greater than or equal to the maximum patience `p`,
+If the count is greater than or equal to `delay`,
 the function returns `true`, otherwise it returns `false`.
-The count is reset when `delta(best_score, f(...)) > min_delta`.
-
-A common use case of `plateau` is early stopping. For this,
-we have added `early_stopping` as an alias to `plateau`.
-Note that you can do more generic things with `plateau`,
-for example reducing the learning rate when the training loss plateaus.
+The count is reset when `distance(best_score, f(...)) > min_dist`.
 
 # Examples
 ```jldoctest
@@ -807,18 +801,57 @@ julia> Flux.@epochs 10 begin
 [ Info: Epoch 3
 ```
 """
-function plateau(f, p; delta = -, init_score = 0, min_delta = 0)
-  is_plateau = let best_score = init_score
+function early_stopping(f, delay; distance = -, init_score = 0, min_dist = 0)
+  trigger = let best_score = init_score
     (args...; kwargs...) -> begin
       score = f(args...; kwargs...)
-      Δ = delta(best_score, score)
+      Δ = distance(best_score, score)
       best_score = Δ < 0 ? best_score : score
 
-      return Δ < min_delta
+      return Δ < min_dist
     end
   end
 
-  return patience(is_plateau, p)
+  return patience(trigger, delay)
 end
 
-const early_stopping = plateau
+"""
+    plateau(f, width; distance = -, init_score = 0, min_dist = 1f-6)
+
+Return a function that internally counts by one when
+`abs(distance(last_score, f(...))) <= min_dist`, where
+`last_score` holds the last value of `f(...)`.
+If the count is greater than or equal to `width`,
+the function returns `true`, otherwise it returns `false`.
+The count is reset when `abs(distance(last_score, f(...))) > min_dist`.
+
+# Examples
+```jldoctest
+julia> f = let v = 10
+       () -> v = v / abs(v) - v
+       end; # -9, 8, -7, 6, ...
+
+julia> trigger = Flux.plateau(f, 3; init_score=10, min_dist=18);
+
+julia> Flux.@epochs 10 begin
+       trigger() && break
+       end
+[ Info: Epoch 1
+[ Info: Epoch 2
+[ Info: Epoch 3
+[ Info: Epoch 4
+```
+"""
+function plateau(f, width; distance = -, init_score = 0, min_dist = 1f-6)
+  is_plateau = let last_score = init_score
+    (args...; kwargs...) -> begin
+      score = f(args...; kwargs...)
+      Δ = abs(distance(last_score, score))
+      last_score = score
+
+      return Δ < min_dist
+    end
+  end
+
+  return patience(is_plateau, width)
+end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -744,7 +744,7 @@ isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
 
 """
-    early_stopping(f; delta=-, min_delta=0, patience=3)
+    patience_counter(f; delta=-, min_delta=0, patience=3)
 
 Return a function that evaluates the metric `f` and compares its value
 against its value on last invocation. When the difference has been less
@@ -757,13 +757,17 @@ If you are using some increasing metric (e.g. accuracy),
 you can customize the `delta` function:
 `(best_score, score) -> score - best_score` (for increasing metrics).
 
+A common use case of `patience_counter` is early stopping. For this,
+we have added `early_stopping` as an alias to `patience_counter`. But
+please note that you can do more generic things with `patience_counter`,
+for example reducing the learning rate when the training loss plateaus.
+
 # Examples
 ```jldoctest
 julia> function loss()
        l = 0
        return () -> l += 1
-       end # pseudo loss function that returns increasing values
-loss (generic function with 1 method)
+       end; # pseudo loss function that returns increasing values
 
 julia> es = Flux.early_stopping(loss(); patience=3);
 
@@ -775,7 +779,7 @@ julia> Flux.@epochs 10 begin
 [ Info: Epoch 3
 ```
 """
-function early_stopping(f; delta = -, min_delta = 0, patience = 3)
+function patience_counter(f; delta = -, min_delta = 0, patience = 3)
   best_score = f()
   count = 0
 
@@ -787,3 +791,5 @@ function early_stopping(f; delta = -, min_delta = 0, patience = 3)
     return count >= patience
   end
 end
+
+early_stopping = patience_counter
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -744,7 +744,26 @@ isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
 
 """
-    plateau(f, patience; delta = -, min_delta = 0)
+    patience(predicate, patience)
+
+Return a function that internally counts by one when
+`predicate(...) == true`, otherwise the count is reset
+to zero. If the count is greater than or equal to `patience`,
+the function returns `true`, otherwise it returns `false`.
+"""
+function patience(predicate, patience)
+  on_trigger = let count = 0
+    (args...; kwargs...) -> begin
+      count = predicate(args...; kwargs...) ? count + 1 : 0
+
+      return count >= patience
+    end
+  end
+end
+
+
+"""
+    plateau(f, p; delta = -, init_score = 0, min_delta = 0)
 
 Return a function that internally counts by one when
 `delta(best_score, f(...)) <= min_delta` where
@@ -782,19 +801,18 @@ julia> Flux.@epochs 10 begin
 [ Info: Epoch 3
 ```
 """
-function plateau(f, patience; delta = -, init_score = 0, min_delta = 0)
-  best_score = init_score
-  count = 0
-
-  let best_score = best_score, count = count
-    function on_plateau(args...; kwargs...)
+function plateau(f, p; delta = -, init_score = 0, min_delta = 0)
+  is_plateau = let best_score = init_score
+    (args...; kwargs...) -> begin
       score = f(args...; kwargs...)
       Δ = delta(best_score, score)
-      count = Δ < min_delta ? count + 1 : 0
       best_score = Δ < 0 ? best_score : score
-      return count >= patience
+
+      return Δ < min_delta
     end
   end
+
+  return patience(is_plateau, p)
 end
 
 const early_stopping = plateau
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -758,7 +758,7 @@ julia> loss() = rand();
 julia> trigger = Flux.patience(() -> loss() < 1, 3);
 
 julia> Flux.@epochs 10 begin
-       trigger() && break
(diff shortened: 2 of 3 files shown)


DhairyaLGandhi:
Seems geberally handy, do you think it would make more sense as a callback though?

DhairyaLGandhi:
We can always do it that way later though. For now some docs and tests would be helpful.

DhairyaLGandhi:
Something like this might make sense too once we merge #1471 

```julia
struct EarlyStopping
  l::AbstractVector
  patience
  tolerance
end

EarlyStopping(memory, patience, tolerance) = EarlyStopping(CircularBuffer(memory), patience, tolerance)

function (es::EarlyStopping)(l, gs, d)
  ...
end
```

src/utils.jl
```diff
@@ -742,3 +742,15 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
 isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
+
+function early_stopping(f; min_delta=0, patience=3)
+  min_metric = Inf
+  count = 0
+
+  return function ()
+      metric = f()
+      count = min_metric - metric < min_delta ? count + 1 : 0
```
darsnack, line 752: My understanding of early stopping counts up even if the metric lowers within delta? You want to stop training as long as there isn't a rising slope?

darsnack (changes requested):
Thanks for your contribution. Can you add a doc string, entry in the documentation, and some tests?

darsnack:
No need to limit this utility by pigeon-holing it into a callback-biased implementation. I think the current approach in the PR is best. It can easily be used in a CB like:
```julia
ea = Flux.early_stopping(loss(), patience=5)
cb = () -> ea() && Flux.stop()
```

Perhaps we can have a version of `stop()` that accepts a function. i.e. `Flux.stop(f)` returns a function that calls `Flux.stop()` when `f()` returns true.

DhairyaLGandhi:
Implementing it as a callback doesn't take away anything from it being used outside of a callback context.

We should also generally want this to be decoupled from the loss function. It would hold unnecessary references alive.
 
Anyway this would be expected to be used within a train loop 99% of the times, so it's worth it to make that api first class.

darsnack:
The callback system uses `Flux.stop()` whereas this just return true when a condition is met. This implementation is more generic because it can be used in more than `Flux.train!`. There are other ways to make the `Flux.train!` API "first-class" without limiting these utilities to only be used in the stopping implementation used by `Flux.train!`.

darsnack:
Anyways, let's not open an argument. Like you said, we can do the callback (or not callback) later.

src/utils.jl
darsnack: Oh wait, I read this backwards. This is for the loss specifically.

I guess my comment is we should try to make this something that can be applied to the loss or accuracy metrics. Currently, it assumes a downward trending metric like the loss.

darsnack:
> As I am new to Flux.jl, I think advice and opinions from maintainers would be helpful before I proceed.

The tests should be added under `test/utils.jl`. You can probably do a test with a pseudo-loss function that you know will hit the early stop criterion in `n` iterations, run the loop with the early stopping criterion in place. Then test if `n` matches the number you expect.

Docs will include a docstring and an entry in `docs/src/utilities.md`.

src/utils.jl
```diff
@@ -742,3 +742,15 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
 isleaflike(x) = Functors.isleaf(x)
 isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
+
+function early_stopping(f; min_delta=0, patience=3)
```
DhairyaLGandhi, line 746:
```suggestion
function early_stopping(f; min_delta = 0, patience = 3)
```

src/utils.jl
queensferryme: Thx for your advice, I am working on that. Early stopping should work on both increasing/decreasing metrics.

DhairyaLGandhi: Correct, it could get a condition function maybe which can operate on a fixed size of vector of some quantity of interest (typically loss).

queensferryme:
> Implementing it as a callback doesn't take away anything from it being used outside of a callback context.
> 
> We should also generally want this to be decoupled from the loss function. It would hold unnecessary references alive.
> 
> Anyway this would be expected to be used within a train loop 99% of the times, so it's worth it to make that api first class.

I have considered implementing a callback function for early stopping, which seems more natural and integrated into the current Flux workflow. However, here are some considerations that drive me to this closure-based implementation:

1. In my experience, most early stopping happens between loops instead of between batches (within one train loop) because loss generally tends to **_oscillate_** greatly between batches. That means early stopping bwtween batches is likely to break the entire training process even when the model is still underfitting.

2. As far as I understand, `Flux.stop` only terminates the current train loop, whereas in early stopping we'd like that the current train loop along with any follow-up loops are **_all_** stopped. Since there aren't any inter-loop hooks for Flux right now, I think manually breaking from loops like `ea() || break` becomes the only choice.

src/utils.jl
```diff
@@ -742,3 +742,42 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
 isleaflike(x) = Functors.isleaf(x)
 isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
+
+"""
+    early_stopping(f; min_delta=0, patience=3)
+
+Return a function that evaluates the metric `f` and compares its value
+against that of last invokation. When the difference has been less
```
darsnack, line 750:
```suggestion
against its value on last invocation. When the difference has been less
```

src/utils.jl
```diff
@@ -742,3 +742,42 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+than `min_delta` for at least `patience` times, `true` is returned;
```
darsnack, line 751:
```suggestion
than `min_delta` for at least `patience` times, `true` is returned,
```

src/utils.jl
```diff
@@ -742,3 +742,42 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+
+julia> es = Flux.early_stopping(loss());
+
+julia> Flux.@epochs 10 begin
+       es() || break
+       end
+[ Info: Epoch 1
+[ Info: Epoch 2
+[ Info: Epoch 3
+[ Info: Epoch 4
```
darsnack, line 770: This example seems a little too opaque right now, since when it stops depends a lot on the dynamics we are discussing down on L779. But I think once you figure out how to do increasing/decreasing metrics, then this will become clearer.

darsnack (changes requested):
Looking good. I think once you figure out the increasing/decreasing metric part, then this will likely be ready to go.

Your feedback on early stopping is appreciated. I think it is just more proof that we should do the most generic thing in this context, which is to write a utility that returns `true` or `false`. It will make the utility more flexible to use a variety of contexts. There are several options for how we integrate a boolean utility into the rest of the callback system.

DhairyaLGandhi:
re 2. since `Flux.stop` throws, it will actually terminate all loops outside of `train` and will break the loop when used inside `train`, so it can be used either way.

darsnack:
It can be used either way if the user is willing to a) write boiler plate try/catch code or b) accept an ugly exception message. A boolean utility truly can be used however the user wants, because it makes no impositions on the user.

src/utils.jl
```diff
@@ -770,14 +774,17 @@ julia> Flux.@epochs 10 begin
...
 ```
 """
-function early_stopping(f; min_delta = 0, patience = 3)
-  min_metric = Inf
+function early_stopping(f; min_delta = 0, mode = "dec", patience = 3)
+  sign = mode == "dec" ? 1 : -1
+  best_score = sign * Inf
+  best = mode == "dec" ? min : max
   count = 0
+  delta = score -> (best_score - score) * sign
```
darsnack: If we're going to introduce a new argument, why not make `delta` a function that's passed into `early_stopping` (as opposed to a `mode` argument).

src/utils.jl
```diff
@@ -742,3 +742,49 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+[ Info: Epoch 3
+[ Info: Epoch 4
+```
+"""
+function early_stopping(f; min_delta = 0, mode = "dec", patience = 3)
+  sign = mode == "dec" ? 1 : -1
+  best_score = sign * Inf
+  best = mode == "dec" ? min : max
+  count = 0
+  delta = score -> (best_score - score) * sign
```
darsnack, line 782: Instead of a `mode` argument, we could let the user pass `delta` as a function into `early_stopping`.

queensferryme:
> re 2. since `Flux.stop` throws, it will actually terminate all loops outside of `train` and will break the loop when used inside `train`, so it can be used either way.

My personal experience tells otherwise:
Screenshot

I believe this is because `StopException` and `SkipException` is captured within `Flux.train!`
https://github.com/FluxML/Flux.jl/blob/81007079bc54edb5623e6eadc6d1e33a953f1c12/src/optimise/train.jl#L106-L114

Maybe this is an unexpected behavior?

src/utils.jl
queensferryme: Will try this.

darsnack:
No that's expected. I think what Dhairya was saying is that if you call the callback in your outer loop (over epochs) then it will throw an exception and stop the program. That's not great IMO because it means you either accept an ugly exception, or you write some boilerplate try/catch code to make it pretty. I don't see why we would want the user to write more boilerplate when `ea() || break` works just fine.

src/utils.jl
```diff
@@ -742,3 +742,48 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+
+Return a function that evaluates the metric `f` and compares its value
+against its value on last invocation. When the difference has been less
+than `min_delta` for at least `patience` times, `true` is returned,
+otherwise `false` is returned.
+
+By default, `early_stopping` expects the metric `f` to be minimized.
+However, if you are using some increasing metric, accuracy for example,
+you can change `early_stopping`'s behaviour by customizing the `delta`
+function: `(best_score, score) -> score - best_score`.
```
darsnack, line 757:
```suggestion
Return a function that evaluates the metric `f` and compares its value
against its value on last invocation. When the difference has been less
than `min_delta` at least `patience` times, `true` is returned,
otherwise `false` is returned.

The difference is measured by keyword argument, `delta`.
By default, `early_stopping` expects the metric `f` to be minimized.
If you are using some increasing metric (e.g. accuracy),
you can customize the `delta` function:
`(best_score, score) -> score - best_score` (for increasing metrics).
```

queensferryme:
> Correct, it could get a condition function maybe which can operate on a fixed size of vector of some quantity of interest (typically loss).

I am also considering @DhairyaLGandhi's suggestion. Maybe we could provide a variant of `early_stopping` that holds a fixed size `buffer` of the last n collected metrics, as well as a `condition` function that operates on the buffer to decide whether we should early stop now. This could add more flexibility to the behavior of early stopping.

I will try implement if this sounds like a good idea to you both @DhairyaLGandhi @darsnack.

darsnack (changes requested):

darsnack:
> Maybe we could provide a variant of early_stopping that holds a fixed size buffer of the last n collected metrics, as well as a condition function that operates on the buffer to decide whether we should early stop now.

My only concern here is calling this "early stopping," since it doesn't match the accepted understanding of the term. To @DhairyaLGandhi's point, the current approach doesn't stop anything, it just returns a boolean that can be used to stop something. If we were to move to this more generic and powerful approach, I would suggest we consider renaming the function.

The other alternative is to build this utility for applying a threshold on a window of values, then make `early_stopping` be syntactic sugar for passing the correct value for `condition` so that the it matches the conventional understanding.

src/utils.jl
```diff
@@ -742,3 +742,47 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+function early_stopping(f; delta = -, min_delta = 0, patience = 3)
+  best_score = f()
+  count = 0
+
+  return function ()
+    score = f()
+    Δ = delta(best_score, score)
+    count = Δ < min_delta ? count + 1 : 0
+    best_score = Δ < 0 ? best_score : score
+    return count < patience
```
darsnack, line 786: If we do go with the early stopping wording, shouldn't this return `count >= patience`? As in, if this function indicates when to stop, then it should be "active" (true) when the stopping criterion is met.

queensferryme: You are right, a fix is on the way.

queensferryme:
> The other alternative is to build this utility for applying a threshold on a window of values, then make `early_stopping` be syntactic sugar for passing the correct value for `condition` so that the it matches the conventional understanding.

Sounds somewhat troublesome and complicated to use 😂 I think I will stick with this implemention for now, which seems more intuitive.

darsnack:
Since the current design allows for more than just stopping, do we still want to consider a name change? PyTorch's [`ReduceLROnPlateau`](https://pytorch.org/docs/stable/optim.html#torch.optim.lr_scheduler.ReduceLROnPlateau) is an example that uses the same "early stopping" logic to decide when to advance the schedule. We have the opportunity to do something more generic here. `ReduceLROnPlateau` is hardcoded to a specific criterion for advancing the schedule, and a specific type of schedule. We can write a more generic version (see [here](https://github.com/darsnack/ParameterSchedulers.jl/issues/11#issuecomment-815897717)) that works with any underlying schedule and predicate for advancing.

But if we use the utility here as our predicate, then the name `early_stopping` is weird. And in general, this utility doesn't do any stopping itself.

test/utils.jl
```diff
@@ -418,3 +418,43 @@ end
...
+    l = 0
+    return () -> l += step
+  end
+
+  @testset "delta" begin
+    es = Flux.early_stopping(metric(); delta=(best_score, score) -> score - best_score)
+
+    n_iter = 0
+    while n_iter < 99
+      es() ? break : n_iter += 1
```
darsnack, line 433:
```suggestion
      es() && break
      n_iter += 1
```

test/utils.jl
```diff
@@ -418,3 +418,43 @@ end
...
+
+    @test n_iter == 99
+  end
+
+  @testset "min delta" begin
+    es = Flux.early_stopping(metric(step=-2); min_delta=1)
+
+    n_iter = 0
+    while n_iter < 99
+      es() ? break : n_iter += 1
```
darsnack, line 444:
```suggestion
      es() && break
      n_iter += 1
```

test/utils.jl
```diff
@@ -418,3 +418,43 @@ end
...
+
+    @test n_iter == 99
+  end
+
+  @testset "patience" begin
+    es = Flux.early_stopping(metric(); patience=10)
+
+    n_iter = 0
+    while n_iter < 99
+      es() ? break : n_iter += 1
```
darsnack, line 455:
```suggestion
      es() && break
      n_iter += 1
```

darsnack (changes requested):

src/utils.jl
```diff
@@ -742,3 +742,48 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+julia> Flux.@epochs 10 begin
+       es() && break
+       end
+[ Info: Epoch 1
+[ Info: Epoch 2
+[ Info: Epoch 3
+```
+"""
+function early_stopping(f; delta = -, min_delta = 0, patience = 3)
+  best_score = f()
```
DhairyaLGandhi, line 779: We might want to have a way to pass `f` some args. We also don't want to have to call `f()` twice here.

darsnack: The second call only happens when the closure is called? This call only happens once to initialize `best_score`. Since the correct initial value will depend on `f` and `delta`, maybe it would be better to pass `best_score` in as a keyword argument. The default can be `f()`.

darsnack: Also, what args were you thinking of passing to `f`? The only additional variables introduced into the scope are `patience` and `best_score`. Just trying to understand how the additional args might be used.

DhairyaLGandhi: Well the closure would be called frequently. Optimising here makes sense.

I was referring to passing arguments to `f` - be that the loss function or for delayed evaluation etc. Having the general case here is nicer.

darsnack: Oh are you saying something like
```julia
return function (args...)
    score = f(args...)
    # etc
end
```

darsnack: That's a good idea. Should we just make it args + kwargs?

darsnack: > Well the closure would be called frequently. Optimising here makes sense.

Yeah but you need the updated score from `f`. What would you optimize out? Could you share a snippet of what you mean? I am just confused what you want to eliminate.

queensferryme:
What the current implementation does under the hood is repeatedly evaluating the metric `f` until its `delta` is less than `min_delta` for at least `patience` times, so maybe we should call it a `MetricValiator` or `MetricDeltaChecker`?

I agree that the current implementation could do something more generic than just "early stopping", but I also think it is better to list one or two examples in the document so that users won't be confused by the more generic but also more abstract name.

darsnack:
`patience_counter`? I think also having `early_stopping` as syntactic sugar is fine too.

darsnack:
Can you address [this comment](https://github.com/FluxML/Flux.jl/pull/1545/files#r610792838)?

PS: thanks for having the patience (no pun intended) for seeing this through

queensferryme:
> `patience_counter`? I think also having `early_stopping` as syntactic sugar is fine too.

sounds good.

> Can you address [this comment](https://github.com/FluxML/Flux.jl/pull/1545/files#r610792838)?

Yes, actually I am currently working on this. It have been great working with you guys 😁

src/utils.jl
```diff
@@ -742,3 +742,61 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+"""
+function patience_counter(f; delta = -, min_delta = 0, patience = 3)
+  best_score = NaN
+  count = 0
+
+  return function (args...; kwargs...)
+    if isnan(best_score)
+      best_score = f(args...; kwargs...)
+      return false
+    end
```
darsnack, line 793: I guess this is the downside to including `(args...; kwargs)`. If we implement it this way, then I prefer `missing` or `nothing` to NaN.

Alternatively, we have have an `init_score` keyword argument that defaults to 0.

src/utils.jl
```diff
@@ -742,3 +742,61 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+    score = f(args...; kwargs...)
+    Δ = delta(best_score, score)
+    count = Δ < min_delta ? count + 1 : 0
+    best_score = Δ < 0 ? best_score : score
+    return count >= patience
+  end
+end
+
+early_stopping = patience_counter
```
darsnack, line 802:
```suggestion
early_stopping(args...; kwargs...) = patience_counter(args...; kwargs...)
```

src/utils.jl
```diff
@@ -742,3 +742,61 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+
+"""
+    patience_counter(f; delta=-, min_delta=0, patience=3)
+
+Return a function that evaluates the metric `f` and compares its value
+against its value on last invocation. When the difference has been less
+than `min_delta` at least `patience` times, `true` is returned,
+otherwise `false` is returned. The function returned by `patience_counter`
+accepts a variable number of arguments and keyword arguments, which will then
+be passed to the metric function `f`  when being evaluated.
```
darsnack, line 754:
```suggestion
Return a function that evaluates the metric `f` and compares its value
against its value on last invocation. When the difference has been less
than `min_delta` at least `patience` times, `true` is returned,
otherwise `false` is returned. If `f` accepts arguments,
then the returned function respects that.
```

src/utils.jl
```diff
@@ -742,3 +742,61 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+
+The difference is measured by keyword argument, `delta`.
+By default, `early_stopping` expects the metric `f` to be minimized.
```
darsnack, line 757:
```suggestion
By default, `patience_counter` expects the metric `f` to be minimized.
```

src/utils.jl
```diff
@@ -742,3 +742,61 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+If you are using some increasing metric (e.g. accuracy),
+you can customize the `delta` function:
+`(best_score, score) -> score - best_score` (for increasing metrics).
+
+A common use case of `patience_counter` is early stopping. For this,
+we have added `early_stopping` as an alias to `patience_counter`. But
+please note that you can do more generic things with `patience_counter`,
+for example reducing the learning rate when the training loss plateaus.
```
darsnack, line 765:
```suggestion
A common use case of `patience_counter` is early stopping. For this,
we have added `early_stopping` as an alias to `patience_counter`.
Note that you can do more generic things with `patience_counter`,
for example reducing the learning rate when the training loss plateaus.
```

darsnack (changes requested):

darsnack (approved):
Would want to get @DhairyaLGandhi to give a final look over anyways, but this looks good to me.

darsnack:
Seeing no objections, let's merge this. We can always have a follow-up PR. Thanks for your contribution @queensferryme!

bors r+

DhairyaLGandhi:
Sorry this seems to have moved quite a bit. Actually there are some changes that we would want here. First being that the api being passed in should simply be `f` and have an example that actually inputs multiple arguments. Wrapping a function directly seems somewhat awkward.

Why not name it either early_stopping or patience? The counter in there is throwing me off.

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+  return function (args...; kwargs...)
+    score = f(args...; kwargs...)
+    Δ = delta(best_score, score)
+    count = Δ < min_delta ? count + 1 : 0
+    best_score = Δ < 0 ? best_score : score
+    return count >= patience
+  end
+end
+
+early_stopping(args...; kwargs...) = patience_counter(args...; kwargs...)
```
DhairyaLGandhi, line 796: An alias can be made with `const`

darsnack:
See above discussion for `early_stopping`. `patience` seems a bit vague to me? The utility "counts" up based on the metric values and triggers after some "patient" waiting period.

> First being that the api being passed in should simply be f and have an example that actually inputs multiple arguments. Wrapping a function directly seems somewhat awkward.

Not sure what you mean here.

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+
+julia> Flux.@epochs 10 begin
+       es() && break
+       end
+[ Info: Epoch 1
+[ Info: Epoch 2
+[ Info: Epoch 3
+```
+"""
+function patience_counter(f; delta = -, init_score = 0, min_delta = 0, patience = 3)
```
DhairyaLGandhi, line 783: Rename to patience and make `patience` positional

DhairyaLGandhi (changes requested):

src/utils.jl
DhairyaLGandhi: @darsnack ref https://github.com/FluxML/Flux.jl/pull/1545#issuecomment-828448253

darsnack: Yeah making `patience` positional is a good idea. @queensferryme do you mind doing this?

I still think `patience` is too vague of a name for the function. I'm not tied to `patience_counter`, so open to other suggestions.

DhairyaLGandhi: This shouldn't have a default either, since there isn't a good value that can be applied generally to most use cases. Best to have users explicitly say what they expect.

DhairyaLGandhi:
I would imagine we should be able to refactor it so that we can write something like

```julia
train(..., cb = early_stopping(patience))
```

Currently, it forces users to get an awkward setup to use the model once they've attached stopping criteria, and they can't access it directly. Conceptually these are separate things so I think we should separate them in code as well.

darsnack:
With the addition of arguments to `f`, the model and stopping function are decoupled. For example,
```julia
loss(x, y, m) = Flux.logitcrossentropy(m(x), y)
trigger = early_stopping(m -> loss(x, y, m), 3)
```
allows the model to be passed in as an argument. And if we end up with a standard callback interface, then you could make the stopping criteria adhere to that.

As far as `cb = early_stopping(...)` goes, I think we should keep the callback part separate from this utility. As I mentioned before, making `early_stopping` a callback that stops training is only going to limit its potential use. For example, it would not be able to be used for the reduce on plateau style scheduling described above.

DhairyaLGandhi:
I think we can extend it to different use cases quite easily. 

Stopping shouldn't need any information about f it should only know how to compare some entities, and have a condition to signal if the condition is met.

darsnack:
`f` is not the model

DhairyaLGandhi:
I don't think it's particularly relevant what `f` is in this case. In the dP sense, the loss metric is absolutely a transform that is part of the model.

DhairyaLGandhi:
The utility should know how to compare two different outputs of `f` is my limited point there.

darsnack:
Isn't that the purpose of the `delta` function that's passed in?

darsnack:
You could certainly refactor the utility so that `f` (or its latest output) is passed in each time utility is called. But I think that just means the utility is more verbose to use without any real benefits.

darsnack:
Also maybe `plateau` is a good name for the utility

DhairyaLGandhi:
You would ideally use common ones defined for you, pass it to the stopping function and you're done. The currect default is sensible already. It's not any more or less verbose iiuc.

darsnack:
It's the difference between specifying it when creating the trigger: `trigger = early_stopping(f, patience)` and `trigger(...)` vs. passing it when calling the trigger: `trigger = early_stopping(patience)` and `trigger(f(...))`.

I don't see the advantage of moving to the second version. The first version, which uses closes over `f`, is a super-set of the second version. By passing in `identity` for `f`, you get the same syntax as the second. We could even have `early_stopping(patience) = early_stopping(identity, patience)` as syntactic sugar. The first version, however, is more powerful, because it let's you control what needs to be in scope for whatever calls the trigger. For the schedule example above, you want to be able to close over everything and just call `trigger()`.

src/utils.jl
queensferryme: done

queensferryme:
> Also maybe `plateau` is a good name for the utility

Do sound better than `patience_counter` : )

queensferryme:
> Stopping (and other such utilities) shouldn't need any information about f. It should only know how to compare some entities, and have a condition to signal if the condition is met.

I agree with @DhairyaLGandhi that stopping should only know about the entities to compare ideally. However as I imagine, it is syntactically more convenient to bind an `f` at initialization, since in most scenarios `f` does not change a lot. PyTorch ignite is doing this the same way ([link](https://github.com/pytorch/ignite/blob/130517738999cf81787a02cca6c134fa7bd7d221/ignite/handlers/early_stopping.py#L49)).

Also, now that we support passing args & kwargs (https://github.com/FluxML/Flux.jl/pull/1545/commits/4b2c690443a67ab8351e105d752fe98af8ba19a6) to the trigger, we can always do something like
```julia
trigger = plateau(g -> g(), 10)
trigger(g1)
trigger(g2)
```
if we do need binding a different `f` every time triggered.

darsnack:
I agree that not knowing `f` on construction simplifies the implementation, but this is also a fairly simple bit of code relatively speaking.

I do think not binding `f` on construction is a mistake. Almost always (99% of the time) the internal state (`best_score`) is tightly coupled to the sequence of calls to `f`. It would be incorrect behavior to call the trigger with a different `f` for nearly every use case (maybe even every case). So, it only makes sense to close over both `f` and `best_score` together, since they are linked. As you mentioned, since we allow arguments, you can always have pass `f` in as an argument or use `identity` if you want `trigger(f(...))` like syntax.

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+  return function (args...; kwargs...)
+    score = f(args...; kwargs...)
+    Δ = delta(best_score, score)
+    count = Δ < min_delta ? count + 1 : 0
+    best_score = Δ < 0 ? best_score : score
+    return count >= patience
+  end
+end
+
+const early_stopping(args...; kwargs...) = plateau(args...; kwargs...)
```
DhairyaLGandhi, line 796:
```suggestion
const early_stopping = plateau
```

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
 isleaflike(x) = Functors.isleaf(x)
 isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
+
+"""
+    patience_counter(f; delta=-, min_delta=0, patience=3)
```
DhairyaLGandhi, line 747:
```suggestion
    plateau(f, patience; delta = -, min_delta = 0)
```

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+against its value on last invocation. When the difference has been less
+than `min_delta` at least `patience` times, `true` is returned,
+otherwise `false` is returned. If `f` accepts arguments,
+then the returned function respects that.
+
+The difference is measured by keyword argument, `delta`.
+By default, `patience_counter` expects the metric `f` to be minimized.
+If you are using some increasing metric (e.g. accuracy),
+you can customize the `delta` function:
+`(best_score, score) -> score - best_score` (for increasing metrics).
```
DhairyaLGandhi, line 759: Add an example in the proper docs for this please (not in this docstring)

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+
+A common use case of `patience_counter` is early stopping. For this,
```
DhairyaLGandhi, line 761:
```suggestion
A common use case of `plateau` is early stopping. For this,
```

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+we have added `early_stopping` as an alias to `patience_counter`.
```
DhairyaLGandhi, line 762:
```suggestion
we have added `early_stopping` as an alias to `plateau`.
```

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+[ Info: Epoch 1
+[ Info: Epoch 2
+[ Info: Epoch 3
+```
+"""
+function plateau(f, patience; delta = -, init_score = 0, min_delta = 0)
+  best_score = init_score
+  count = 0
+
+  return function (args...; kwargs...)
```
DhairyaLGandhi, line 787:
```suggestion
  function on_plateau(args...; kwargs...)
```

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+A common use case of `patience_counter` is early stopping. For this,
+we have added `early_stopping` as an alias to `patience_counter`.
+Note that you can do more generic things with `patience_counter`,
+for example reducing the learning rate when the training loss plateaus.
+
+# Examples
+```jldoctest
+julia> function loss()
+       l = 0
+       return () -> l += 1
```
DhairyaLGandhi, line 770: I think this got lost. Typically, a loss wouldn't return a closure. Use a `let` block to access local variables.

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+       end; # pseudo loss function that returns increasing values
+
+julia> es = Flux.early_stopping(loss(), 3);
```
DhairyaLGandhi, line 773: See comment above. This would change to `loss` accordingly and in consistency with the intent of the api.

DhairyaLGandhi (commented):
This is turning out good! Thanks!

If `delta` is the function intended to be customised, may I suggest it needs a slightly more generic name that reflects that it represents the "condition". Maybe `by`?

I also thought we were going with `patience` instead of `plateau`? The former can be better associated with being more general purpose than `plateau` would indicate. 

See `patience(..., by = ...)` vs `plateau(, delta)`. It may not necessarily be a plateau or a delta. Just some tolerance over a condition. Conveniences like `early_stopping` can have more domain specific keywords. We would also want to link the docs to the `early_stopping` with a one liner about what early stopping is.

I'm not sure what incorrect usage is being referred to. One can imagine wanting to stop based on several sub conditions in a model and having more than one such conditions attached to one section, or stopping some part of training and not others. The simplicity is a side effect of keeping the users context clean. Currently it would result in a bunch of  `gensym` which is less elegant.

darsnack:
It will always be a plateau. `patience(..., by = ...)` does not say much about what the function is intended to do. Patience is the mechanism through which we get/measure a plateau; it isn't the key component of the utility. It could just as easily have been "delay" or "wait" instead of "patience" (but in all cases "plateau" would describe the end behavior).

Similarly, I would discourage `by` which makes sense for `sort(..., by = ...)`, but it is not as informative here. Perhaps `diff` or `dist`/`distance` would be better than `delta`.

> Currently it would result in a bunch of gensym which is less elegant.

Is there something special about closing over a function? Even without closing on `f`, you would still need to close over `best_score`.

> One can imagine wanting to stop based on several sub conditions in a model and having more than one such conditions attached to one section, or stopping some part of training and not others.

The incorrect behavior is that you could not reuse the same `trigger = plateau(patience)` for all of those multiple conditions. The value of `best_score` is tied to the specific `f` that you are calling. If you reuse the same trigger, then you comparison will be incorrect.

DhairyaLGandhi:
> you would still need to close over best_score.

Yeah but couldn't we use a `let` block?

Similarly, I am thinking if we can use the same construct to also implement pre-baked routines that aren't plateaus. Something that inspects external state or tracks metrics that cannot be inferred by loss for example.

darsnack:
Why can't you use the same let on `f` as well?

src/utils.jl
```diff
@@ -742,3 +742,55 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
+
+"""
+    patience_counter(f; delta=-, min_delta=0, patience=3)
+
+Return a function that evaluates the metric `f` and compares its value
+against its value on last invocation. When the difference has been less
+than `min_delta` at least `patience` times, `true` is returned,
+otherwise `false` is returned. If `f` accepts arguments,
+then the returned function respects that.
```
darsnack, line 753:
```suggestion
Return a function that evaluates the metric `f` and compares its value
against its value on last invocation. When the difference
(measured by `delta`) has been less than `min_delta` at least `patience` times,
`true` is returned, otherwise `false` is returned.
If `f` accepts arguments, then the returned function respects that.
```

src/utils.jl
darsnack, line 759:
```suggestion
Return a function that internally counts by one when
`delta(best_score, f(...)) <= min_delta` where
`best￼_score` is the last seen best value of `f(...)`.
If the count is greater than or equal to `patience`,
the function returns `true`, otherwise it returns `false`.
The count is reset when `delta(best_score, f(...)) > min_delta`.

The keyword argument `delta` is a function of the form
`delta(best_score, current_score)`.
If you are using some increasing metric (e.g. accuracy),
you can customize the `delta` function:
`(best_score, score) -> score - best_score`.
```

queensferryme:
> I also thought we were going with `patience` instead of `plateau`? The former can be better associated with being more general purpose than `plateau` would indicate.

> But this utility specifically will always be a "plateau." You could turn the patience off completely by setting it to 1. But it's still a plateau because the utility measures when the distance between two points is small enough (by definition a "plateau"). BTW I am not tied to plateau. I just think `patience` is too uninformative.

For the current implementation, yes it will always be a `plateau`, and `patience` may seem too general in this case. In my opinion, a general `patience` utility should have only three components: a maximum `patience`, an internal `count` as well as a `predicate` that operates over a set of variables and returns eithr `true` or `false`. When the `predicate` says `true`, the internal `count` is increased by one; otherwise it is reset to zero.

So, in our case, `plateau` is more like a specialization of `patience`, where the `predicate` is reduced to `delta < min_delta`. The current `plateau` will not be able to do something like "stop the training when the loss has been below some threshold for `n` times". We can do this more general `patience` thing in another pull request and provide some pre-baked routines maybe, if you are interested.

src/utils.jl
```diff
@@ -742,3 +742,59 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+[ Info: Epoch 1
+[ Info: Epoch 2
+[ Info: Epoch 3
+```
+"""
+function plateau(f, patience; delta = -, init_score = 0, min_delta = 0)
+  best_score = init_score
+  count = 0
+
+  let best_score = best_score, count = count
```
queensferryme, line 789: @DhairyaLGandhi I am not quite sure about "using a `let` block to access local variables". Is this what you mean?

src/utils.jl
```diff
@@ -742,3 +742,59 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+`best_score` is the last seen best value of `f(...)`.
+If the count is greater than or equal to `patience`,
+the function returns `true`, otherwise it returns `false`.
+The count is reset when `delta(best_score, f(...)) > min_delta`.
+
+The keyword argument `delta` is a function of the form
+`delta(best_score, current_score)`.
+If you are using some increasing metric (e.g. accuracy),
+you can customize the `delta` function:
+`(best_score, score) -> score - best_score`.
```
queensferryme, line 760: > Add an example in the proper docs for this please (not in this docstring)

@DhairyaLGandhi Sorry for the dumb question. Where would you suggest to put this doc? `docs/src/utilities.md`?

darsnack:
What about something like:

```julia
function patience(predicate, patience)
  count = 0

  on_trigger = let count = count
   (args...; kwargs...) -> begin
      count = predicate(args...; kwargs...) ? count + 1 : 0

      return count >= patience
    end
  end

  return on_trigger
end

function plateau(f, p; distance = -, min_dist = 0, init_score = 0)
  best_score = init_score

  on_plateau = let best_score = best_score
    (args...; kwargs...) -> begin
        score = f(args...; kwargs...)
        Δ = distance(best_score, score)
        best_score = Δ < 0 ? best_score : score

        return Δ < min_dist
     end
  end

  return patience(metric, p)
end

const early_stopping = plateau
```

Personally, I think the patience logic is fairly simple. The hard part is the score closure. If it wasn't for the best score tracking, then I would suggest completely foregoing `plateau` in the example above. But I think the logic in `plateau` is cumbersome enough that it warrants its own function.

src/utils.jl
DhairyaLGandhi: Yes that would be correct

queensferryme:
> What about something like:
> 
> ```julia
> function patience(predicate, patience)
>   on_trigger = let count = 0
>    (args...; kwargs...) -> begin
>       count = predicate(args...; kwargs...) ? count + 1 : 0
> 
>       return count >= patience
>     end
>   end
> 
>   return on_trigger
> end
> 
> function plateau(f, p; distance = -, min_dist = 0, init_score = 0)
>   on_plateau = let best_score = init_score
>     (args...; kwargs...) -> begin
>         score = f(args...; kwargs...)
>         Δ = distance(best_score, score)
>         best_score = Δ < 0 ? best_score : score
> 
>         return Δ < min_dist
>      end
>   end
> 
>   return patience(on_plateau, p)
> end
> 
> const early_stopping = plateau
> ```
> 
> Personally, I think the patience logic is fairly simple. The hard part is the score closure. If it wasn't for the best score tracking, then I would suggest completely foregoing `plateau` in the example above. But I think the logic in `plateau` is cumbersome enough that it warrants its own function.

This looks great!

docs/src/utilities.md
```diff
@@ -102,4 +102,44 @@ Flux.nfan
 Flux.throttle
 Flux.stop
 Flux.skip
+Flux.patience
+Flux.plateau
+```
+
```
darsnack, line 108: This should be in its own sub-section, separate from the other callback helpers.

Also, the docs reads as a continuation of the docstrings. They should read independently of the docstrings. I would structure the content as "tutorial/docs" then the docstring too.

src/utils.jl
```diff
@@ -742,3 +742,83 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
 isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
+
+"""
+    patience(predicate, patience)
+
+Return a function that internally counts by one when
+`predicate(...) == true`, otherwise the count is reset
+to zero. If the count is greater than or equal to `patience`,
+the function returns `true`, otherwise it returns `false`.
```
darsnack, line 752:
```suggestion
Return a function that internally counts by one when
`predicate(...) == true`, otherwise the count is reset to zero.
If the count is greater than or equal to `patience`,
the function returns `true`, otherwise it returns `false`.
```

src/utils.jl
```diff
@@ -742,3 +742,83 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+
+# Examples
+```jldoctest
+julia> loss = () -> rand();
```
darsnack, line 756:
```suggestion
julia> loss() = rand();
```

src/utils.jl
```diff
@@ -742,3 +742,83 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+
+julia> Flux.@epochs 10 begin
+       es() && break
+       end
+[ Info: Epoch 1
+[ Info: Epoch 2
+[ Info: Epoch 3
+```
+"""
+function plateau(f, p; delta = -, init_score = 0, min_delta = 0)
```
darsnack, line 810: Can we change `delta` to `distance` or `diff` to be more meaningful?

darsnack, line 810: In other frameworks, is the default equivalent to `delta = -` or `delta = (best, s) -> abs(best - s)`. The former measures improvement whereas the latter measures no significant change. (I think the latter is what's typical). Using the second option also means no one needs to change `delta` just to use an increasing metric.

darsnack (changes requested):
Looking good, thanks. I noticed a couple questions that I missed before.

src/utils.jl
queensferryme: Well, the latter do seem more suitable for `plateau` since it measures "no significant change". However, there are two things we might want to reconsider here if we are changing to the `abs` default:

1. The default of `min_delta` should be changed to some positive value (`eps()` maybe) correspondingly;
2. These new defaults of `plateau` are no longer suitable for `early_stopping`. For example, the monitored metric could still be changing significantly (thus no plateau), but to the worse direction (thus early stopping needed). I suggest something like this (to override the defaults):
    ```julia
    const early_stopping = (f, p; diff = -, min_delta = 0) -> begin
      return plateau(f, p; diff = diff, min_delta = min_delta)
    end
    ```

darsnack: Seems like that would get too confusing since the only difference would be the default values. Let's keep what we have. Maybe change the name from `plateau` again?

darsnack: Actually, perhaps the following implementations make sense:

```
function plateau(f, width; distance = -, init_score = 0, min_dist = 1f-6)
  is_plateau = let best_score = init_score
    (args...; kwargs...) -> begin
      score = f(args...; kwargs...)
      Δ = abs(distance(best_score, score))
      best_score = Δ < 0 ? best_score : score

      return Δ < min_dist
    end
  end

  return patience(is_plateau, width)
end

function early_stopping(f, delay; distance = -, init_score = 0, min_dist = 0)
  is_plateau = let best_score = init_score
    (args...; kwargs...) -> begin
      score = f(args...; kwargs...)
      Δ = distance(best_score, score)
      best_score = Δ < 0 ? best_score : score

      return Δ < min_dist
    end
  end

  return patience(is_plateau, delay)
end
```

`plateau` is two-sided and `early_stopping` is one-sided. Both convenience wrappers round `patience`.

docs/src/utilities.md
```diff
@@ -103,3 +103,51 @@ Flux.throttle
 Flux.stop
 Flux.skip
 ```
+
+## Patience Helpers
+
+Flux provide utilities for scheduling your workflow according to some monitored condition and a maximum `patience`. For example, you can use `early_stopping` to stop training when the model is converging or deteriorating, or you can use `plateau` to check if the model is stagnating.
```
darsnack, line 109:
```suggestion
Flux provides utilities for controlling your training procedure according to some monitored condition and a maximum `patience`. For example, you can use `early_stopping` to stop training when the model is converging or deteriorating, or you can use `plateau` to check if the model is stagnating.
```

docs/src/utilities.md
```diff
@@ -103,3 +103,51 @@ Flux.throttle
...
+
+The keyword argument `distance` of `early_stopping` is a function of the form `distance(best_score, score)`. By default `distance` is `-`, which implies that the monitored metric `f` is expected to be decreasing and mimimized. If you use some increasing metric (e.g. accuracy), you can customize the `distance` function: `(best_score, score) -> score - best_score`.
+
+```julia
+acc = let v = 0
+  () -> v = max(1, v + 0.01)
+end
+
+es = Flux.early_stopping(acc, 3; delta = (best_score, score) -> score - best_score)
```
darsnack, line 118:
```suggestion
# early stopping trigger for acc with patience of 3
es = Flux.early_stopping(acc, 3; delta = (best_score, score) -> score - best_score)
```

docs/src/utilities.md
```diff
@@ -103,3 +103,51 @@ Flux.throttle
 Flux.stop
 Flux.skip
 ```
+
+## Patience Helpers
+
+Flux provide utilities for scheduling your workflow according to some monitored condition and a maximum `patience`. For example, you can use `early_stopping` to stop training when the model is converging or deteriorating, or you can use `plateau` to check if the model is stagnating.
+
```
darsnack, line 110:
```suggestion

For example, below we create a pseudo-loss function that decreases, bottoms out, then increases. The early stopping trigger will break the loop before the loss increases too much.
```julia
# create a pseudo-loss that decreases for 4 calls, then stars increasing
# we call this like loss()
loss = let t = 0
  () -> begin
    (t - 4)^2
    t += 1
  end
end

# create an early stopping trigger
# returns true when the loss increases for two consecutive steps
es = Flux.early￼￼￼_stopping(loss, 2)

# this loop will break early
for iter in 1:100
  @info "Iteration $iter"
  es() && break
end
```

```

File: docs/src/utilities.md [hunk 39]
```diff
@@ -103,3 +103,51 @@ Flux.throttle
...
+Flux.@epochs 10 begin
+  es() && break
+end
+
+es = Flux.early_stopping(acc, 3)
+
+# This will stop at the 3rd epoch
+Flux.@epochs 10 begin
+  es() && break
+end
```
darsnack, line 130: Let's delete this second example in favor of my other suggestion with the loss example.

docs/src/utilities.md
```diff
@@ -103,3 +103,51 @@ Flux.throttle
...
+```
+
```
darsnack, line 132:
```suggestion

`early_stopping` and `plateau` are both built on top of `patience`. You can use `patience` to build your own triggers that use a patient counter. For example, if you want to trigger when the loss is below a threshold for several consecutive iterations:

```julia
threshold(f, thresh, delay) = Flux.patience(delay) do
  f() < thresh
end
```

```

File: src/utils.jl [hunk 41]
```diff
@@ -742,3 +742,116 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+"""
+function patience(predicate, patience)
+  on_trigger = let count = 0
+    (args...; kwargs...) -> begin
+      count = predicate(args...; kwargs...) ? count + 1 : 0
+
+      return count >= patience
+    end
+  end
+end
```
darsnack, line 776: Just so that the argument and function name aren't the same.

```suggestion
function patience(predicate, wait)
  on_trigger = let count = 0
    (args...; kwargs...) -> begin
      count = predicate(args...; kwargs...) ? count + 1 : 0

      return count >= wait
    end
  end
end
```

src/utils.jl
```diff
@@ -742,3 +742,116 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
 isleaflike(::Tuple{Vararg{<:Number}}) = true
 isleaflike(::Tuple{Vararg{<:AbstractArray{<:Number}}}) = true
+
+"""
+    patience(predicate, patience)
+
+Return a function that internally counts by one when
+`predicate(...) == true`, otherwise the count is reset to zero.
+If the count is greater than or equal to `patience`,
+the function returns `true`, otherwise it returns `false`.
```
darsnack, line 752:
```suggestion
    patience(predicate, wait)

Return a function that internally counts by one when
`predicate(...) == true`, otherwise the count is reset to zero.
If the count is greater than or equal to `wait`,
the function returns `true`, otherwise it returns `false`.
```

test/utils.jl
```diff
@@ -418,3 +418,67 @@ end
                                 LayerNorm(8)))
   @test length(modules) == 5
 end
+
+@testset "Patience" begin
```
darsnack, line 422:
```suggestion
@testset "Patience triggers" begin
```

test/utils.jl
```diff
@@ -418,3 +418,67 @@ end
...
+  @testset "patience" begin
+    trigger = Flux.patience(() -> true, 3)
+
+    @test trigger() == false
+    @test trigger() == false
+    @test trigger() == true
+  end
```
darsnack, line 429:
```suggestion
    v = [false, true, false, true, true, true]
    trigger = let v = v
      patience(i -> v[i], 3)
    end

    n_iter = 0
    for i in 1:length(v)
      n_iter += 1
      trigger(i) && break
    end
    @test n_iter == 6
  end
```

darsnack (changes requested):
Looking really good. I like the doc additions. I think we should be GTM after this last round of revisions.

cc @DhairyaLGandhi so he can look over things

darsnack (dismissed):

DhairyaLGandhi:
Cool, I'll give this a once over this weekend

src/utils.jl
```diff
@@ -742,3 +742,116 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+       trigger() && break
+       end
+[ Info: Epoch 1
+[ Info: Epoch 2
+[ Info: Epoch 3
+```
+"""
+function patience(predicate, wait)
+  on_trigger = let count = 0
+    (args...; kwargs...) -> begin
```
DhairyaLGandhi, line 770: `on_ trigger` should be the name of the function imo. Good to have it named for stacktraces

docs/src/utilities.md
```diff
@@ -103,3 +103,73 @@ Flux.throttle
...
+    (t - 4) ^ 2
+  end
+end
+
+# create an early stopping trigger
+# returns true when the loss increases for two consecutive steps
+es = Flux.early_stopping(loss, 2; init_score=9)
+
+# this will stop at the 6th (4 decreasing + 2 increasing calls) epoch
+Flux.@epochs 10 begin
```
DhairyaLGandhi, line 127: I think it makes sense to drop the `Flux.`. I'm sure we can do a setup to import Flux and the symbols separately. I'd like to focus the API docs on the functions themselves.

src/utils.jl
darsnack: How is that done with a `let`?

Like
```julia
let ...
  function name(...)
    ...
  end
end
```
?

darsnack:
@queensferryme If you can address the latest comments when you get a chance, we can finish this PR up! Thanks for all your work and patience.

docs/src/utilities.md
queensferryme: Do I also need to remove the `Flux.` in docstrings?

src/utils.jl
```diff
@@ -742,3 +742,116 @@ modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
...
+the function returns `true`, otherwise it returns `false`.
+
+# Examples
+```jldoctest
+julia> loss() = rand();
+
+julia> trigger = Flux.patience(() -> loss() < 1, 3);
+
+julia> Flux.@epochs 10 begin
+       trigger() && break
```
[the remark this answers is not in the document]
queensferryme: Do you mean adding more indentation or removing the preceding whitespaces?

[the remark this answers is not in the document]
darsnack: Adding indentation and treating the REPL prompt as a left margin.

docs/src/utilities.md
darsnack: No those are necessary for the auto-docstrings to resolve.

src/utils.jl
DhairyaLGandhi: Yeah, I believe so

darsnack (approved):

darsnack:
Great, thanks @queensferryme!

bors r+

DhairyaLGandhi:
Thanks all

queensferryme:
@DhairyaLGandhi @darsnack Thanks for your patience and time!

PR #1534, opened 2021-03-11
Fix printing of OneHotArray in REPL when CUDA.allowscalar(false)

darsnack:
Adds `Base.show` for `OneHotArray`. Fixes #1532. This turned out to be harder than I originally thought.

### PR Checklist

- [ ] ~~Tests are added~~
- [ ] ~~Entry in NEWS.md~~
- [ ] ~~Documentation, if applicable~~
- [ ] ~~API changes require approval from a committer (different from the author, if applicable)~~

PR #1550, opened 2021-03-28
Add test for show(io, ::OneHotArray) on GPU

darsnack:
Adds a test for #1534.

### PR Checklist

- [x] Tests are added
- [ ] ~~Entry in NEWS.md~~
- [ ] ~~Documentation, if applicable~~
- [ ] ~~API changes require approval from a committer (different from the author, if applicable)~~

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -25,14 +25,14 @@ function _show_elements(x::OneHotArray)
   return xrepr
 end
 
-function Base.show(io::IO, ::MIME"text/plain", x::OneHotArray{<:Any, L, <:Any, N, I}) where {L, N, I}
+function Base.show(io::IO, x::OneHotArray{<:Any, L, <:Any, N, I}) where {L, N, I}
   join(io, string.(size(x)), "×")
   print(io, " Flux.OneHotArray{")
   join(io, string.([L, N, I]), ",")
   println(io, "}:")
   print(io, _show_elements(x))
 end
-function Base.show(io::IO, ::MIME"text/plain", x::OneHotVector{T, L}) where {T, L}
+function Base.show(io::IO, x::OneHotVector{T, L}) where {T, L}
   print(io, string.(length(x)))
   print(io, "-element Flux.OneHotVector{")
   join(io, string.([L, T]), ",")
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -25,14 +25,14 @@ function _show_elements(x::OneHotArray)
   return xrepr
 end
 
-function Base.show(io::IO, x::OneHotArray{<:Any, L, <:Any, N, I}) where {L, N, I}
+function Base.show(io::IO, ::MIME"text/plain", x::OneHotArray{<:Any, L, <:Any, N, I}) where {L, N, I}
   join(io, string.(size(x)), "×")
   print(io, " Flux.OneHotArray{")
   join(io, string.([L, N, I]), ",")
   println(io, "}:")
   print(io, _show_elements(x))
 end
-function Base.show(io::IO, x::OneHotVector{T, L}) where {T, L}
+function Base.show(io::IO, ::MIME"text/plain", x::OneHotVector{T, L}) where {T, L}
   print(io, string.(length(x)))
   print(io, "-element Flux.OneHotVector{")
   join(io, string.([L, T]), ",")
--- a/test/cuda/cuda.jl
+++ b/test/cuda/cuda.jl
@@ -39,8 +39,8 @@ using LinearAlgebra: I, cholesky, Cholesky
 end
 
 @testset "onehot gpu" begin
-  y = Flux.onehotbatch(ones(3), 1:10) |> gpu;
-  @test show(stdout, y)
+  y = Flux.onehotbatch(ones(3), 1:2) |> gpu;
+  @test repr("text/plain", y) == "2×3 Flux.OneHotArray{2,2,Vector{UInt32}}:\n 1  1  1\n 0  0  0"
 end
 
 @testset "onecold gpu" begin
--- a/test/cuda/cuda.jl
+++ b/test/cuda/cuda.jl
@@ -40,7 +40,7 @@ end
 
 @testset "onehot gpu" begin
   y = Flux.onehotbatch(ones(3), 1:2) |> gpu;
-  @test repr("text/plain", y) == "2×3 Flux.OneHotArray{2,2,Vector{UInt32}}:\n 1  1  1\n 0  0  0"
+  @test (show(y); true)
 end
 
 @testset "onecold gpu" begin
--- a/test/cuda/cuda.jl
+++ b/test/cuda/cuda.jl
@@ -40,7 +40,7 @@ end
 
 @testset "onehot gpu" begin
   y = Flux.onehotbatch(ones(3), 1:2) |> gpu;
-  @test (show("text/plain", y); true)
+  @test (repr("text/plain", y); true)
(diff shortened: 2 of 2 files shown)


DhairyaLGandhi:
This seems a bit fragile to changes to the code. What is the expected "thing" to be tested?

DhairyaLGandhi:
Maybe you want to check that warnings are empty instead?

darsnack:
> Maybe you want to check that warnings are empty instead?

Good idea

darsnack:
Just want to test that scalar indexing isn't triggered. I think in the GPU integration tests this would error out not warn.

darsnack:
I could do something like `@test (show(stdout, y); true)`. It will error on `show` if it uses scalar indexing.

darsnack:
What does build kite do for the regular tests (not under CUDA)?

DhairyaLGandhi:
I would prefer to do the most obvious test. The piping to different io means there are extra indirections which involve going through different code paths than just the expression to be tested.

The warnings are also the safest behavior to test, since they go through exactly the code path we want to avoid. Are there test utilities we have in Base that can track the warnings?

> What does build kite do for the regular tests (not under CUDA)?

I'm not sure I understand the question. I'm fairly sure that the scalar indexing should be disallowed globally since it's only active on gpu backed memory, and can catch erroneous tests in other places (separate from where we disallow it now)

darsnack:
> The piping to different io means there are extra indirections which involve going through different code paths than just the expression to be tested.

Sure, I can just do `show(y)`.

> The warnings are also the safest behavior to test, since they go through exactly the code path we want to avoid.

I agree, but since the GPU build has `allowscalar(false)`, a test pass will be no errors/warnings, and a test fail will throw an error not a warning.

Re: my question about build kite

Currently, all the tests under `test/cuda` are run with `allowscalar(false)` globally. But I see from the built kite logs, that it is also running the other tests. Since the tests not under `test/cuda` don't move data to the GPU, I am wondering if these tests run on the CPU with build kite, or is there some process where build kite moves the data to the GPU even without an explicit `gpu` in the test.

DhairyaLGandhi:
Those tests are regular runs. We could force to use GPUs since clearly it seems needed, but nothing like that so far.

darsnack:
Ah, okay. I guess I am confused about the scalar indexing error we got on build kite then. From the logs it looked like the warning was during the regular tests (specifically the losses tests) and not the GPU tests.

darsnack:
So what I am trying to test here is that `show(::OneHotArray)` runs without triggering scalar indexing. Previously, the default `show` for any array would trigger calls to `getindex`.

DhairyaLGandhi:
It is possible that there are regular tests that use gpus that weren't vetted to be put in the correct place.

ToucheSir:
I did a bit of digging for this, here's a MWE:
```julia
julia> print(y)
Bool[ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore are only permitted from the REPL for prototyping purposes.
If you did intend to index this array, annotate the caller with @allowscalar.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] assertscalar(op::String)
    @ GPUArrays ~/.julia/packages/GPUArrays/ymNij/src/host/indexing.jl:51
  [3] getindex
    @ ~/.julia/packages/GPUArrays/ymNij/src/host/indexing.jl:84 [inlined]
  [4] getindex
    @ ~/.julia/packages/Flux/0c9kI/src/onehot.jl:59 [inlined]
  [5] isassigned(::Flux.OneHotArray{UInt32, 2, 1, 2, CuArray{UInt32, 1}}, ::Int64, ::Int64)
    @ Base ./abstractarray.jl:513
  [6] _show_nonempty(io::IOContext{Base.TTY}, X::Flux.OneHotArray{UInt32, 2, 1, 2, CuArray{UInt32, 1}}, prefix::String)
    @ Base ./arrayshow.jl:407
  [7] show(io::Base.TTY, X::Flux.OneHotArray{UInt32, 2, 1, 2, CuArray{UInt32, 1}})
    @ Base ./arrayshow.jl:451
  [8] print(io::Base.TTY, x::Flux.OneHotArray{UInt32, 2, 1, 2, CuArray{UInt32, 1}})
    @ Base ./strings/io.jl:35
  [9] print(xs::Flux.OneHotArray{UInt32, 2, 1, 2, CuArray{UInt32, 1}})
    @ Base ./coreio.jl:3
 [10] top-level scope
    @ REPL[46]:1
 [11] top-level scope
    @ ~/.julia/packages/CUDA/mVgLI/src/initialization.jl:52
```
`show("text/plain", y)` needs an IO argument, because otherwise it is treating the mime type as a plain string and simply invoking `print` (which does not provide a mime type to show) for each argument. This means we're missing a compact default string representation for `OneHotArray`s like other special array types have.

darsnack:
Sorry, I didn't quite catch how the test should be changed.

ToucheSir:
I should've been a little more clear with my comment. There are a couple options here:
1. re-write the test to use `show(::IO, "text/plain", y)` or `repr("text/plain", y)` (I like the latter). This will test exactly the functionality in #1534, but we'd need a follow-up PR for generic `show`/`print`/`repr` without a mime type.
2. re-write the test to use a generic `show/repr(y)` and remove the mime restriction on the `OneHotArray` method. This would save some code, but break a bit from the array convention of having a simpler default representation. e.g.
```julia
julia> ones(5)
5-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0
 1.0

julia> print(ones(5))
[1.0, 1.0, 1.0, 1.0, 1.0]

julia> ones(5)'
1×5 adjoint(::Vector{Float64}) with eltype Float64:
 1.0  1.0  1.0  1.0  1.0

julia> print(ones(5)')
[1.0 1.0 1.0 1.0 1.0]

julia> I(5)
5×5 Diagonal{Bool, Vector{Bool}}:
 1  ⋅  ⋅  ⋅  ⋅
 ⋅  1  ⋅  ⋅  ⋅
 ⋅  ⋅  1  ⋅  ⋅
 ⋅  ⋅  ⋅  1  ⋅
 ⋅  ⋅  ⋅  ⋅  1

julia> print(I(5))
Bool[1 0 0 0 0; 0 1 0 0 0; 0 0 1 0 0; 0 0 0 1 0; 0 0 0 0 1]

```

darsnack:
AFAICT, something like `[gpu(Flux.onehotbatch(rand(1:10, 4), 1:10)) for _ in 1:3]` will print compactly without failing on the GPU.

> we'd need a follow-up PR for generic show/print/repr without a mime type

Like a follow-up to add such functionality or to test it? According to the [Julia docs](https://docs.julialang.org/en/v1.6/base/io-network/#Base.show-Tuple{IO,%20Any,%20Any}), the 2-arg `show` without mime will default to the 3-arg version.

CarloLucibello (approved):

ToucheSir:
> Like a follow-up to add such functionality or to test it? According to the [Julia docs](https://docs.julialang.org/en/v1.6/base/io-network/#Base.show-Tuple%7BIO,%20Any,%20Any%7D), the 2-arg `show` without mime will default to the 3-arg version.

I thought so too, but as you can see from the stacktrace that doesn't appear to be the case. Perhaps this has something to do with being an `AbstractArray`?

PR #1555, opened 2021-03-31
add downstream.yml

DhairyaLGandhi:
Add some downstream tests

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -25,6 +25,7 @@ jobs:
           - {user: FluxML, repo: Metalhead.jl, group: All}
           - {user: FluxML, repo: AtomicGraphNets.jl, group: All}
 
+    if: contains(github.event.pull_request.labels.*.name, 'run downstream test')
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -16,8 +16,8 @@ jobs:
         julia-version: [1]
         os: [ubuntu-latest]
         package:
-          - {user: FluxML, repo: ObjectDetector.jl, group: All}
-          - {user: FluxML, repo: Transformers.jl, group: All}
+          - {user: r3tex, repo: ObjectDetector.jl, group: All}
+          - {user: chengchingwen, repo: Transformers.jl, group: All}
           - {user: FluxML, repo: GeometricFlux.jl, group: All}
           - {user: FluxML, repo: FastAI.jl, group: All}
           - {user: FluxML, repo: Flux3D.jl, group: All}
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -1,0 +1,56 @@
+name: Downstream
+
+on:
+  push:
+    branches: [master]
+    tags: [v*]
+  pull_request:
+
+jobs:
+  test:
+    name: ${{ matrix.package.repo }}/${{ matrix.package.group }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: [1]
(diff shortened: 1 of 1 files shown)


CarloLucibello:
This is cool, but having it in every PR is maybe too much? Can we trigger it only on request same as people in Base do with`@nanosoldier `runtests(["JSON", "Crayons"], vs = ":master")`?

darsnack:
Yeah something like invoking via a GH comment, and we automatically do it for release branches? Like when we create a branch for a new breaking release, this should be a sanity check that runs on that branch.

ToucheSir:
Has this functionality been rolled into the FluxBot/ModelZookeeper or is it still WIP?

.github/workflows/Downstream.yml
```diff
@@ -0,0 +1,56 @@
...
+          - {user: FluxML, repo: ObjectDetector.jl, group: All}
+          - {user: FluxML, repo: Transformers.jl, group: All}
+          - {user: FluxML, repo: GeometricFlux.jl, group: All}
+          - {user: FluxML, repo: FastAI.jl, group: All}
+          - {user: FluxML, repo: Flux3D.jl, group: All}
+          - {user: FluxML, repo: Torch.jl, group: All}
+          - {user: FluxML, repo: Metalhead.jl, group: All}
+          - {user: FluxML, repo: AtomicGraphNets.jl, group: All}
+
+    steps:
```
johnnychen94, line 28: > but having it in every PR is maybe too much? Can we trigger it only on request same as people in Base do with@nanosoldier runtests(["JSON", "Crayons"], vs = ":master")?

This can be done by adding or removing a PR label, e.g., `run downstream test`.

```suggestion
    if: contains(github.event.pull_request.labels.*.name, 'run downstream test')
    steps:
```

ChrisRackauckas:
Add your friendly neighborhood scientists to the list? 😅

ToucheSir:
Is this good to go? What's the best/fastest way to test it?

.github/workflows/Downstream.yml
```diff
@@ -0,0 +1,59 @@
...
+  test:
+    name: ${{ matrix.package.repo }}/${{ matrix.package.group }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        julia-version: [1]
+        os: [ubuntu-latest]
+        package:
+          - {user: FluxML, repo: ObjectDetector.jl, group: All}
```
johnnychen94, line 19: Just a question, what does `group` mean here?

johnnychen94:
> What's the best/fastest way to test it?

I guess is to tag this PR with "run downstream test" and then push an empty commit.

.github/workflows/Downstream.yml
ChrisRackauckas: Which test group, https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/test/runtests.jl#L18 so you can run a subset of the tests.

johnnychen94: Sounds like a simplified version of https://github.com/JuliaTesting/ReTest.jl

darsnack:
We need to label this PR as "run downstream test" which I can't seem to do on mobile. That will trigger the workflow.

darsnack (approved):
It seems like this works for most of our downstream targets. I am not sure why Transformers or ObjectDetector throw those errors. I recommend we merge this PR and solve those issues, tweak the trigger, etc. in a different PR.

Having running tests is better than nothing.

.github/workflows/Downstream.yml
```diff
@@ -0,0 +1,59 @@
...
+          - {user: FluxML, repo: Transformers.jl, group: All}
```
ToucheSir, line 20:
```suggestion
          - {user: r3tex, repo: ObjectDetector.jl, group: All}
          - {user: chengchingwen, repo: Transformers.jl, group: All}
```
@darsnack

PR #1557, opened 2021-03-31
Fix #1556

### LINKED ISSUE #1556: Restructure on Dense no longer plays nicely with alternative types
MWE:

```julia
using ReverseDiff, Flux
ann = Chain(Dense(2,10,tanh), Dense(10,1))
p1,re = Flux.destructure(ann)
function f(p)
    re(p[1:41])([1.0,2.0])
end

ReverseDiff.gradient(f,p3)
```

```julia
MethodError: ReverseDiff.TrackedReal{Float32, Float32, ReverseDiff.TrackedArray{Float32, Float32, 2, Matrix{Float32}, Matrix{Float32}}}(::ForwardDiff.Dual{ForwardDiff.Tag{ReverseDiff.var"#105#107"{DataType, Tuple{}, Val{(1,)}}, Float32}, Float32, 1}) is ambiguous. Candidates:
  (T::Type{var"#s31"} where var"#s31"<:Real)(x::ForwardDiff.Dual) in Tracker at C:\Users\accou\.julia\packages\Tracker\YNNTM\src\lib\real.jl:111
  ReverseDiff.TrackedReal{V, D, O}(value) where {V, D, O} in ReverseDiff at C:\Users\accou\.julia\packages\ReverseDiff\iHmB4\src\tracked.jl:56
Possible fix, define
  ReverseDiff.TrackedReal{V, D, O}(::ForwardDiff.Dual) where {V, D, O}
macro expansion at broadcast.jl:126 [inlined]
splatcall at broadcast.jl:111 [inlined]
(::ReverseDiff.var"#105#107"{DataType, Tuple{}, Val{(1,)}})(s::StaticArrays.SVector{1, ForwardDiff.Dual{ForwardDiff.Tag{ReverseDiff.var"#105#107"{DataType, Tuple{}, Val{(1,)}}, Float32}, Float32, 1}}) at broadcast.jl:150
static_dual_eval at apiutils.jl:32 [inlined]
vector_mode_gradient! at gradient.jl:125 [inlined]
gradient! at gradient.jl:48 [inlined]
(::ReverseDiff.var"#df#106"{DataType, DiffResults.ImmutableDiffResult{1, Float32, Tuple{StaticArrays.SVector{1, Float32}}}, Tuple{}, Val{(1,)}})(x::Float32) at broadcast.jl:148
_broadcast_getindex_evalf at broadcast.jl:648 [inlined]
_broadcast_getindex at broadcast.jl:621 [inlined]
getindex at broadcast.jl:575 [inlined]
copy at broadcast.jl:922 [inlined]
materialize at broadcast.jl:883 [inlined]
broadcast(f::ReverseDiff.var"#df#106"{DataType, DiffResults.ImmutableDiffResult{1, Float32, Tuple{StaticArrays.SVector{1, Float32}}}, Tuple{}, Val{(1,)}}, As::Vector{Float32}) at broadcast.jl:821
∇broadcast at broadcast.jl:154 [inlined]
copy(_bc::Base.Broadcast.Broadcasted{ReverseDiff.TrackedStyle, Tuple{Base.OneTo{Int64}}, Type{ReverseDiff.TrackedReal{Float32, Float32, ReverseDiff.TrackedArray{Float32, Float32, 2, Matrix{Float32}, Matrix{Float32}}}}, Tuple{ReverseDiff.TrackedArray{Float32, Float32, 1, Vector{Float32}, Vector{Float32}}}}) at broadcast.jl:94
materialize(bc::Base.Broadcast.Broadcasted{ReverseDiff.TrackedStyle, Nothing, Type{ReverseDiff.TrackedReal{Float32, Float32, ReverseDiff.TrackedArray{Float32, Float32, 2, Matrix{Float32}, Matrix{Float32}}}}, Tuple{ReverseDiff.TrackedArray{Float32, Float32, 1, Vector{Float32}, Vector{Float32}}}}) at broadcast.jl:883
broadcast at broadcast.jl:821 [inlined]
create_bias(weights::ReverseDiff.TrackedArray{Float32, Float32, 2, Matrix{Float32}, Matrix{Float32}}, bias::ReverseDiff.TrackedArray{Float32, Float32, 1, Vector{Float32}, Vector{Float32}}, dims::Int64) at utils.jl:397
Dense(W::ReverseDiff.TrackedArray{Float32, Float32, 2, Matrix{Float32}, Matrix{Float32}}, bias::ReverseDiff.TrackedArray{Float32, Float32, 1, Vector{Float32}, Vector{Float32}}, σ::typeof(tanh)) at basic.jl:117
(::Flux.var"#126#127")(y::NamedTuple{(:weight, :bias, :σ), Tuple{ReverseDiff.TrackedArray{Float32, Float32, 2, Matrix{Float32}, Matrix{Float32}}, ReverseDiff.TrackedArray{Float32, Float32, 1, Vector{Float32}, Vector{Float32}}, typeof(tanh)}}) at functor.jl:18
fmap1(f::Function, x::Dense{typeof(tanh), Matrix{Float32}, Vector{Float32}}) at functor.jl:49
fmap(f::Flux.var"#54#55"{ReverseDiff.TrackedArray{Float32, Float32, 1, Vector{Float32}, Vector{Float32}}}, x::Dense{typeof(tanh), Matrix{Float32}, Vector{Float32}}; exclude::typeof(Functors.isleaf), cache::IdDict{Any, Any}) at functor.jl:56
```

The stack trace is a bit off and gets cut off, but the issue comes from:

```julia
function create_bias(weights::AbstractArray, bias::AbstractArray, dims::Integer...)
  size(bias) == dims || throw(DimensionMismatch("expected bias of size $(dims), got size $(size(bias))"))
  if eltype(bias) == eltype(weights)
    return bias
  else
    @warn "converting bias to match element type of weights" typeof(weights) typeof(bias) maxlog=3 _id=hash(dims)
    return broadcast(eltype(weights), bias)
  end
end
```

Generally that's not the most correct way to convert types. A better way to handle this is:

```julia
function Flux.create_bias(weights::AbstractArray, bias::AbstractArray, dims::Integer...)
  size(bias) == dims || throw(DimensionMismatch("expected bias of size $(dims), got size $(size(bias))"))
  if eltype(bias) == eltype(weights)
    return bias
  else
    @warn "converting bias to match element type of weights" typeof(weights) typeof(bias) maxlog=3 _id=hash(dims)
    return convert.((eltype(weights),),bias)
  end
end
```

However, that's only a partial solution since it shouldn't be throwing a warning on a legitimate use case, so there's a bigger question as to why it ends up there in the first place (which is cut off from the stack trace).

DhairyaLGandhi:
Reverts some Dense changes since those caused issues downstream. cc @ChrisRackauckas @darsnack @CarloLucibello

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -117,7 +117,7 @@ end
 
 Dense(W, b) = Dense(W, b, identity)
 
-Dense(W, b::Bool, σ) =
+Dense(W, b::Bool, σ = identity) =
   Dense(W, create_bias(W, b, size(W,1)), σ)
 
 function Dense(in::Integer, out::Integer, σ = identity; initW = nothing,
@@ -144,7 +144,7 @@ function (a::Dense)(x::AbstractVecOrMat)
 end
 
 (a::Dense)(x) =
-   reshape(a(reshape(x, size(x,1), :)), :, size(x)[2:end]...)
+  reshape(a(reshape(x, size(x,1), :)), :, size(x)[2:end]...)
 
 function Base.show(io::IO, l::Dense)
   print(io, "Dense(", size(l.weight, 2), ", ", size(l.weight, 1))
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -138,13 +138,13 @@ end
 
 @functor Dense
 
-(a::Dense)(x) = 
-   reshape(a(reshape(x, size(x,1), :)), :, size(x)[2:end]...)
-
 function (a::Dense)(x::AbstractVecOrMat)
   W, b, σ = a.weight, a.bias, a.σ
   σ.(W * x .+ b) 
 end
+
+(a::Dense)(x) =
+   reshape(a(reshape(x, size(x,1), :)), :, size(x)[2:end]...)
 
 function Base.show(io::IO, l::Dense)
   print(io, "Dense(", size(l.weight, 2), ", ", size(l.weight, 1))
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -117,7 +117,7 @@ end
 
 Dense(W, b) = Dense(W, b, identity)
 
-Dense(W, b::Bool, σ = identity) =
+Dense(W::AbstractArray, b::Bool, σ = identity) =
   Dense(W, create_bias(W, b, size(W,1)), σ)
 
 function Dense(in::Integer, out::Integer, σ = identity; initW = nothing,
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -110,30 +110,43 @@ Params([[1.0 1.0 … 1.0 1.0; 1.0 1.0 … 1.0 1.0]])
 ```
 """
 struct Dense{F,S<:AbstractArray,T}
-  W::S
-  b::T
+  weight::S
+  bias::T
   σ::F
 end
 
 Dense(W, b) = Dense(W, b, identity)
 
-function Dense(in::Integer, out::Integer, σ = identity;
-               initW = glorot_uniform, initb = zeros, bias::Bool = true)
-  Dense(initW(out, in), bias ? initb(out) : Zeros(), σ)
+function Dense(in::Integer, out::Integer, σ = identity; initW = nothing,
+               init = glorot_uniform, initb = nothing, bias::Bool = true)
+  if initW !=== nothing
+    depwarn("initW is deprecated, please use the `init` keyword instead")
+    init = initW
+  end
+
+  if initb !=== nothing
+    depwarn("initb is deprecated, please use the array based constructors instead")
+    initb = initb
+  else
+    initb = zeros
+  end
+  Dense(init(out, in), bias ? initb(out) : Zeros(), σ)
 end
 
 @functor Dense
 
 function (a::Dense)(x)
-  W, b, σ = a.W, a.b, a.σ
+  W, b, σ = a.weight, a.bias, a.σ
   x_reshaped = reshape(x, size(x, 1), :)
   x_out = σ.(W * x_reshaped .+ b) 
   reshape(x_out, :, size(x)[2:end]...)
 end
 
 function Base.show(io::IO, l::Dense)
+  print(io, "Dense(", size(l.weight, 2), ", ", size(l.weight, 1))
   print(io, "Dense(", size(l.W, 2), ", ", size(l.W, 1))
   l.σ == identity || print(io, ", ", l.σ)
+  l.bias == Zeros() && print(io, "; bias=false")
   print(io, ")")
 end
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -69,7 +69,7 @@ extraChain(::Tuple{}, x) = ()
 
 
 """
-    Dense(in, out, σ=identity; bias=true, init=glorot_uniform)
+    Dense(in, out, σ = identity; bias = true, init = glorot_uniform)
     Dense(W::AbstractMatrix, [bias, σ])
 
 Create a traditional `Dense` layer, whose forward pass is given by:
@@ -81,7 +81,7 @@ as an `in × N` matrix, or any array with `size(x,1) == in`.
 The out `y` will be a vector  of length `out`, or a batch with
 `size(y) == (out, size(x)[2:end]...)`
 
-Keyword `bias=false` will switch off trainable bias for the layer.
+Keyword `bias = false` will switch off trainable bias for the layer.
 The initialisation of the weight matrix is `W = init(out, in)`, calling the function
 given to keyword `init`, with default [`glorot_uniform`](@doc Flux.glorot_uniform).
 The weight matrix and/or the bias vector (of length `out`) may also be provided explicitly.
@@ -144,7 +144,6 @@ end
 
 function Base.show(io::IO, l::Dense)
   print(io, "Dense(", size(l.weight, 2), ", ", size(l.weight, 1))
-  print(io, "Dense(", size(l.W, 2), ", ", size(l.W, 1))
   l.σ == identity || print(io, ", ", l.σ)
   l.bias == Zeros() && print(io, "; bias=false")
   print(io, ")")
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -117,14 +117,17 @@ end
 
 Dense(W, b) = Dense(W, b, identity)
 
+Dense(W, b::Bool, σ) =
+  Dense(W, create_bias(W, b, size(W,1)), σ)
+
 function Dense(in::Integer, out::Integer, σ = identity; initW = nothing,
                init = glorot_uniform, initb = nothing, bias::Bool = true)
-  if initW !=== nothing
+  if initW !== nothing
     depwarn("initW is deprecated, please use the `init` keyword instead")
     init = initW
   end
 
-  if initb !=== nothing
+  if initb !== nothing
     depwarn("initb is deprecated, please use the array based constructors instead")
     initb = initb
   else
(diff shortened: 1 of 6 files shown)


DhairyaLGandhi:
It needs some fixes to the tests to make sure we are getting expected results. This is a technically breaking change now.

darsnack (changes requested):
I don't think this is the correct way to fix #1556. It is not necessary at all to revert the constructor work. Beyond necessitating a breaking change, most of the stuff being reverted here is actually good work that should be kept. Let's not do drastic changes to the codebase for bug fixes if it is not necessary. If there's a desire to do a design refactor, then that should be done in a separate PR. Happy for us to have a design discussion, but in this case, we can release the appropriate fix without a design overhaul.

A proper fix for #1556 will involve just deleting the `eltype` check from `create_bias`.

darsnack:
To be clear, it isn't `create_bias` or the `Dense` changes that caused this issue. `create_bias` was added in #1379, and the erroneous `eltype` check was added in #1440. All the changes in #1440 can be kept without the unnecessary check, and `create_bias` is reverted to its prior behavior.

DhairyaLGandhi:
I'll do a more minimal change. What specific changes would you like retained?

I would also prefer for some changes to our design to simplify its implementation. Removing `create_bias` from here doesn't seem too disruptive to me. Good call on fixing the bad check too.

darsnack:
My preference would be to keep everything except delete [this check code](https://github.com/FluxML/Flux.jl/blob/81007079bc54edb5623e6eadc6d1e33a953f1c12/src/utils.jl#L393) and just `return bias`. We could remove `create_bias`, but to do it in a non-breaking way that keeps the current API means adding a `bias isa Bool` check to all the constructors with a bias. `create_bias` should really be an internal utility cause that's how it is used.

One other possible change is the specification of `W` in `Dense` to be `<:AbstractMatrix`. My gut says this constraint should be lifted and allowed to be `<:AbstractArray`. But I haven't hunted down where and why that change happened, so maybe it is necessary. Possibly @CarloLucibello or @mcabbott can shed some light.

I do agree with you that we can/should think about a better design that doesn't require an internal utility to handle the bias argument. But my feeling is that it would involve a discussion on `Flux.Zeros`, so that's why I am suggesting opening a separate issue/PR for that refactor.

darsnack:
It's possible there is more than the change I suggested to be reverted, but can we approach the PR from the opposite direction? i.e. restricting the changes to what I suggest, then slowly reverting more stuff as necessary?

mcabbott:
No time today but agree that if trying to fix something quickly, the minimal change is best:

> My preference would be to keep everything except delete [this check code](https://github.com/FluxML/Flux.jl/blob/81007079bc54edb5623e6eadc6d1e33a953f1c12/src/utils.jl#L393) and just `return bias`. 

That seems to be what's actually causing problems, right? 

> One other possible change is the specification of `W` in `Dense` to be `<:AbstractMatrix`. My gut says this constraint should be lifted and allowed to be `<:AbstractArray`. 

Dense calls `W * x`, are there other `ndims(W)` this can sensibly work with?

DhairyaLGandhi:
The thinking is thus: 
1. remove the check in `create_bias` 
2. remove the conversion

Removing the function from here basically means we don't have to worry about its behaviour in a corner case, and the package becomes more reliable for downstream users. We aren't losing features here either.

DhairyaLGandhi:
@darsnack the layer is back to where it was, the tests are back to their earlier state as well.

src/layers/basic.jl
```diff
@@ -109,46 +109,42 @@ julia> Flux.params(d1)  # no trainable bias
...
-               initW = nothing, initb = nothing,
-               init = glorot_uniform, bias=true)
+Dense(W, b) = Dense(W, b, identity)
 
-  W = if initW !== nothing
-    Base.depwarn("keyword initW is deprecated, please use init (which similarly accepts a funtion like randn)", :Dense)
-    initW(out, in)
-  else
-    init(out, in)
+Dense(W, b::Bool, σ) =
```
darsnack, line 120:
```suggestion
Dense(W, b::Bool, σ = identity) =
```

src/layers/basic.jl
```diff
@@ -109,46 +109,42 @@ julia> Flux.params(d1)  # no trainable bias
...
+  if initb !== nothing
+    depwarn("initb is deprecated, please use the array based constructors instead")
+    initb = initb
   else
-    bias
+    initb = zeros
   end
-
-  return Dense(W, b, σ)
+  Dense(init(out, in), bias ? initb(out) : Zeros(), σ)
```
darsnack, line 136: Functionally, this is the same as before, but it means repeating the `bias ? ...` check instead of just calling `Dense(W, b, σ)`. I think the old code that assigns `b` reduces the surface area of errors w.r.t. the `bias ? ...` check.

src/layers/basic.jl
```diff
@@ -109,46 +109,42 @@ julia> Flux.params(d1)  # no trainable bias
...
 function (a::Dense)(x::AbstractVecOrMat)
   W, b, σ = a.weight, a.bias, a.σ
-  return σ.(W*x .+ b)
+  σ.(W * x .+ b) 
 end
 
-(a::Dense)(x::AbstractArray) = 
-  reshape(a(reshape(x, size(x,1), :)), :, size(x)[2:end]...)
+(a::Dense)(x) =
+   reshape(a(reshape(x, size(x,1), :)), :, size(x)[2:end]...)
```
darsnack, line 147:
```suggestion
  reshape(a(reshape(x, size(x,1), :)), :, size(x)[2:end]...)
```

darsnack (changes requested):
Looking close to me. Can you test the snippet @ChrisRackauckas provided in #1556 to see if it is resolved?

> Dense calls W * x, are there other ndims(W) this can sensibly work with?

Probably not, but I guess our preference is to not pre-constrain. I think the change to `<:AbstractArray` makes sense. `W * x` will throw an error on the forward pass, or the output passed into the next layer will throw a dimension mismatch. As long as we don't have a silent error, then I think it might better to leave it open.

src/layers/basic.jl
DhairyaLGandhi: We aren't repeating the bias check in here. It should be fine.

DhairyaLGandhi:
The MWE works

test/utils.jl
```diff
@@ -168,7 +168,7 @@ end
     end
 
     @testset "Dense ID mapping" begin
-        l = Dense(3,3, init = identity_init)
+        l = Dense(3,3, initW = identity_init)
```
darsnack, line 171: Isn't this part of the identity initialization test set? Why is a change needed here?

darsnack (commented):
Looking correct to me.

DhairyaLGandhi:
bors ping

mcabbott:
IMO this should be reverted. Changes orthogonal to fixing a pressing issue don't need to be rushed in.

test/layers/basic.jl
```diff
@@ -40,16 +40,17 @@ import Flux: activations
       @test Dense(rand(Float16, 100,10), true).bias isa Vector{Float16}  # creates matching type
       @test_skip Dense(rand(Float16, 100,10), rand(100)).bias isa Vector{Float16}  # converts to match
 
-      @test Dense(3,4; init=Base.randn, bias=true).bias isa Vector{Float64}
+      @test_skip Dense(3,4; init=Base.randn, bias=true).bias isa Vector{Float64}
```
CarloLucibello, line 43: Is this now producing a float32 bias? this is no good.

DhairyaLGandhi: Since there is no indication of the type of bias, it would make sense to produce single precision bias. It is also addressed in #1559 to restore backwards compatibility with the behaviour.

PR #1560, opened 2021-03-31
Revert #1557 temporarily

darsnack:
This reverts the changes from #1557. I think some of those changes are good, but there were some reservations expressed. Since the changes aren't pressing, let's discuss them instead. We can merge this PR to revert, then open another for the changes from #1557.

### PR Checklist

- [x] ~~Tests are added~~
- [x] ~~Entry in NEWS.md~~
- [x] ~~API changes require approval from a committer (different from the author, if applicable)~~

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.12.0"
+version = "0.12.1"
 
 [deps]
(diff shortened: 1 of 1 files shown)


CarloLucibello:
Looks like the reversion wasn't committed. 

I also agree we should revert, there was a change of behavior there which didn't seem sane

darsnack:
My mistake; closing in favor of #1561

PR #1561, opened 2021-04-01
Reverts #1557

DhairyaLGandhi:
Reverts #1557

### DIFF
--- a/test/layers/basic.jl
+++ b/test/layers/basic.jl
@@ -40,17 +40,16 @@ import Flux: activations
       @test Dense(rand(Float16, 100,10), true).bias isa Vector{Float16}  # creates matching type
       @test_skip Dense(rand(Float16, 100,10), rand(100)).bias isa Vector{Float16}  # converts to match
 
-      @test_skip Dense(3,4; init=Base.randn, bias=true).bias isa Vector{Float64}
+      @test Dense(3,4; init=Base.randn, bias=true).bias isa Vector{Float64}
       @test_skip Dense(3,4; init=Base.randn, bias=[1,2,3,4]).bias isa Vector{Float64}
-
 
       @test_throws MethodError Dense(10, 10.5)
       @test_throws MethodError Dense(10, 10.5, tanh)
-      # @test_throws DimensionMismatch Dense(3,4; bias=rand(5))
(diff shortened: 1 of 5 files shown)


DhairyaLGandhi:
Closes #1560 #1559 

cc @darsnack @CarloLucibello

CarloLucibello (approved):

test/layers/basic.jl
```diff
@@ -40,17 +40,16 @@ import Flux: activations
       @test Dense(rand(Float16, 100,10), true).bias isa Vector{Float16}  # creates matching type
       @test_skip Dense(rand(Float16, 100,10), rand(100)).bias isa Vector{Float16}  # converts to match
 
-      @test_skip Dense(3,4; init=Base.randn, bias=true).bias isa Vector{Float64}
+      @test Dense(3,4; init=Base.randn, bias=true).bias isa Vector{Float64}
```
darsnack, line 43: This should be a `@test_skip` right? The correct `eltype` passing is an accident here (cause `bias = true` uses `zeros` which happens to be `Float64`).

darsnack (changes requested):
Believe this is correct other than the one test

test/layers/basic.jl
DhairyaLGandhi: No. This was in #1558, and it uses `Flux.zeros` (not `Base.zeros`) which produces `Float32` by default.

```julia
julia> Flux.zeros(4)
4-element Array{Float32,1}:
 0.0
 0.0
 0.0
 0.0
```

I agree that the test isn't very explicit about its intent.

darsnack: Ah I see, okay we'll address it separately. Thanks!

CarloLucibello: Reverting here is good, we should produce float64 bias when the weights produced are float64

darsnack:
Okay so do I need to hit approve or is `bors` okay?

test/layers/basic.jl
DhairyaLGandhi: > we should produce float64 bias when the weights produced are float64

No
We should do as we are told by the user. Flux is meant to be hackable. While it is valid to have defaults, it is required to have overrides at different levels.

PR #1572, opened 2021-04-10
Add a dropout_mask method for ComplexF64 array types

ShoofLLC:
Calling _similar_ with a type argument makes dropout work for complex valued networks as well.
Currently using dropout in a complex valued network will fail with the error:

`ERROR: MethodError: no method matching isless(::Float64, ::Complex{Float64})`

This change adds a dropout_mask method for complex arrays with a call to similar with Float64 type.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -46,6 +46,12 @@ function dropout_mask(x, p; dims=:)
   return y
 end
 
+function dropout_mask(x::Array{Complex{Float64}}, p; dims=:)
+  y = rand!(similar(x, Float64, _dropout_shape(x, dims)))
+  y .= _dropout_kernel.(y, p, 1 - p)
+  return y
+end
+
 """
     Dropout(p; dims=:)
 
--- a/test/layers/normalisation.jl
+++ b/test/layers/normalisation.jl
@@ -4,7 +4,7 @@ using Zygote: pullback
 evalwgrad(f, x...) = pullback(f, x...)[1]
 
 @testset "Dropout" begin
-  x = [1.+0im,2.+1im,3.+3im]
+  x = [1.0+0im,2.0+1im,3.0+3im]
   @test x == Dropout(0.1)(x)
   @test x == evalwgrad(Dropout(0), x)
   @test zero(x) == evalwgrad(Dropout(1), x)
--- a/test/layers/normalisation.jl
+++ b/test/layers/normalisation.jl
@@ -4,6 +4,11 @@ using Zygote: pullback
 evalwgrad(f, x...) = pullback(f, x...)[1]
 
 @testset "Dropout" begin
+  x = [1.+0im,2.+1im,3.+3im]
+  @test x == Dropout(0.1)(x)
+  @test x == evalwgrad(Dropout(0), x)
+  @test zero(x) == evalwgrad(Dropout(1), x)
+  
   x = [1.,2.,3.]
   @test x == Dropout(0.1)(x)
   @test x == evalwgrad(Dropout(0), x)


ShoofLLC:
Modifying the dropout_mask function itself to always have a float32 type when calling similar:
`function dropout_mask(x, p; dims=:)
  y = rand!(similar(x, Float32, _dropout_shape(x, dims)))
  return y
end`

would make dropout work for any input type, currently using dropout for Int inputs would fail with an inexact error, but I didn't want to modify the main method just for this use-case.

src/layers/normalise.jl
```diff
@@ -46,6 +46,12 @@ function dropout_mask(x, p; dims=:)
   return y
 end
 
+function dropout_mask(x::Array{Complex{Float64}}, p; dims=:)
```
DhairyaLGandhi, line 49: Seems a bit strict. Maybe we want to generalise the existing method? It should work the same for single precision.

DhairyaLGandhi (commented):
Thanks for looking into this! I like the idea.

src/layers/normalise.jl
ShoofLLC: I agree, I just didn't want to introduce changes that would affect any other types

ShoofLLC: I added a test with a complex array as an input, I could also just modify dropout_mask to always use a Float type when calling similar:

> Modifying the dropout_mask function itself to always have a float32 type when calling similar:
> `function dropout_mask(x, p; dims=:) y = rand!(similar(x, Float32, _dropout_shape(x, dims))) y .= _dropout_kernel.(y, p, 1 - p) return y end`
> 
> would make dropout work for any input type, currently using dropout for Int inputs would fail with an inexact error, but I didn't want to modify the main method just for this use-case.

ToucheSir:
Thanks for the PR! Is this something that could be more generally useful in NNlib, eventually?

ShoofLLC:
> 
> 
> Thanks for the PR! Is this something that could be more generally useful in NNlib, eventually?

Possibly, I'm not sure whether dropout is useful for NNlib users or not, but if it works for all input data types then it could be used as a general function.

ToucheSir:
Coming back to this, I think the same objective could be accomplished without a custom overload. Updating https://github.com/FluxML/Flux.jl/blob/d1510801499a6c2b881d45fcff4c85251665744a/src/layers/normalise.jl#L52 to:
```julia
  y = rand!(rng, similar(x, real(eltype(x)), _dropout_shape(x, dims)))
```
@ShoofLLC would you be interested in rebasing and updating the PR to this change? If not, I'd be happy to create a new one with your tests.

ShoofLLC:
> Coming back to this, I think the same objective could be accomplished without a custom overload. Updating https://github.com/FluxML/Flux.jl/blob/d1510801499a6c2b881d45fcff4c85251665744a/src/layers/normalise.jl#L52 to:
> ```julia
>   y = rand!(rng, similar(x, real(eltype(x)), _dropout_shape(x, dims)))
> ```
> Leaves `IntNN` and `FloatNN` intact, while changing `Complex{T}` to `T`.
> @ShoofLLC would you be interested in rebasing and updating the PR to this change? If not, I'd be happy to create a new one with your tests.

Sure I can rebase and push these changes instead

ToucheSir:
Good catch, though it seems `Dropout` is already failing for int types on master. I think the fix should be pretty straightforward though, instead of 
```julia
  y = rand!(rng, similar(x, real(eltype(x)), _dropout_shape(x, dims)))
```
Use
```julia
  fptype = float(real(eltype(x)))
  y = rand!(rng, similar(x, fptype, _dropout_shape(x, dims)))
```
Where `float` will be a no-op for Real and Complex floating points, but convert Integers.

ToucheSir:
I think this was superseded by #1867, but feel free to reopen if there are still bits you want to land.

PR #1584, opened 2021-04-25
fixes #1583

### LINKED ISSUE #1583: Flux overrides cat behaviour and causes stack overflow
Flux seems to override the behaviour of cat and fails in absence of arrays to concatenate. 

Versions:
```
julia 1.6.1
Flux 0.12.2
```

Correct behaviour:
```
julia> cat(dims=1)
Any[]
```

Incorrect behaviour
```
julia> using Flux

julia> cat(dims=1)
ERROR: StackOverflowError:
Stacktrace:
 [1] cat(; dims::Int64) (repeats 39991 times)
   @ Flux ~/.julia/packages/Flux/6BByF/src/onehot.jl:69
```

DhairyaLGandhi:
Is the classic `Union{}` dispatch piracy gotcha

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -64,22 +64,22 @@ Base.getindex(x::OneHotArray, I::CartesianIndex{N}) where N = x[I[1], Tuple(I)[2
 _onehot_bool_type(x::OneHotLike{<:Any, <:Any, <:Any, N, <:Union{Integer, AbstractArray}}) where N = Array{Bool, N}
 _onehot_bool_type(x::OneHotLike{<:Any, <:Any, <:Any, N, <:CuArray}) where N = CuArray{Bool, N}
 
-function Base.cat(xs::OneHotLike{<:Any, L}...; dims::Int) where L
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
Makes me think if `OneHotLike` is the correct design. It anyway goes against the design goal which expects things like these to fall out naturally. Also a good opportunity to clean up the `_onehot_bool_type` etc.

darsnack:
Probably safer to remove `OneHotLike` and explicitly repeat the dispatch for the long term.

`_onehot_bool_type` is orthogonal. I think it is more convenient than writing out the full parameter list in the dispatch to get the underlying type, but we can remove it if you want.

darsnack (approved):

DhairyaLGandhi:
Agree it's more convenient since we have a long type, but I'm thinking we can contract that. I'd like to see it go specifically because I don't think we should be needing that dispatch at all. Also agree about the `OneHotLike`, it's prone to finicky implementation details. We will need something clever to work around wrapper types causing issues. I might take a page out of how we handled it earlier. 

bors r+

Issue #1588, opened 2021-05-02
Regularization example from docs can't be differentiated


yha:
The second example from https://fluxml.ai/Flux.jl/stable/models/regularisation/ trips up Zygote if you try to compute the gradient:
```
m = Chain(
  Dense(28^2, 128, relu),
  Dense(128, 32, relu),
  Dense(32, 10))

sqnorm(x) = sum(abs2, x)

loss(x, y) = logitcrossentropy(m(x), y) + sum(sqnorm, Flux.params(m))

gradient(()->loss(rand(28^2), rand(10)), params(m))
```
```
ERROR: Mutating arrays is not supported
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:33
  [2] (::Zygote.var"#411#412")(#unused#::Nothing)
    @ Zygote ~\.julia\packages\Zygote\6HN9x\src\lib\array.jl:65
  [3] (::Zygote.var"#2279#back#413"{Zygote.var"#411#412"})(Δ::Nothing)
    @ Zygote ~\.julia\packages\ZygoteRules\OjfTt\src\adjoint.jl:59
  [4] Pullback
    @ ~\.julia\packages\Flux\6o4DQ\src\functor.jl:45 [inlined]
  [5] (::typeof(∂(params!)))(Δ::Nothing)
    @ Zygote ~\.julia\packages\Zygote\6HN9x\src\compiler\interface2.jl:0
  [6] Pullback
    @ ~\.julia\packages\Flux\6o4DQ\src\functor.jl:47 [inlined]
  [7] (::typeof(∂(params!)))(Δ::Nothing)
    @ Zygote ~\.julia\packages\Zygote\6HN9x\src\compiler\interface2.jl:0
  [8] Pullback
    @ ~\.julia\packages\Flux\6o4DQ\src\functor.jl:47 [inlined]
  [9] (::typeof(∂(params!)))(Δ::Nothing)
    @ Zygote ~\.julia\packages\Zygote\6HN9x\src\compiler\interface2.jl:0
 [10] Pullback
    @ ~\.julia\packages\Flux\6o4DQ\src\functor.jl:47 [inlined]
 [11] (::typeof(∂(params!)))(Δ::Nothing)
    @ Zygote ~\.julia\packages\Zygote\6HN9x\src\compiler\interface2.jl:0
 [12] Pullback
    @ ~\.julia\packages\Flux\6o4DQ\src\functor.jl:44 [inlined]
[...]

PR #1590, opened 2021-05-03
Add @nograd for params

ToucheSir:
Testing a quick fix for https://github.com/FluxML/Flux.jl/issues/1588 and https://github.com/FluxML/Zygote.jl/issues/941.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -54,6 +54,9 @@ function params(m...)
   return ps
 end
 
+@nograd params
+
+
 function loadparams!(m, xs)
   for (p, x) in zip(params(m), xs)
     size(p) == size(x) ||


darsnack:
Do we need a test case for the issue?

DhairyaLGandhi:
I mean, we would need to recurse through for implicit parameters and is necessary for higher order ad otherwise we can accidentally return `nothing` for parameters that would otherwise have gradients defined. See https://github.com/FluxML/Zygote.jl/pull/823 for a more a more complete implementation. We can merge that instead with adding some tests.

ToucheSir:
I figured there would be some caveat to this brute force approach. Is there a reason we couldn't bring back the old behaviour of not disallowing differentiation through the IdSet code? As-is, this bug affects the model zoo and who knows how many other models.

CarloLucibello:
We should fix this quickly, #1588 and FluxML/Zygote.jl#941 are pretty serious, breakage of widely exposed and used 
functionality shouldn't go unfixed for months. 

@DhairyaLGandhi could you prioritize  FluxML/Zygote.jl#823 or can we merge this as a temp solution?
What would me a mwe of code that this would break? 

What caused this regression, https://github.com/FluxML/Zygote.jl/pull/876?

darsnack:
> What would me a mwe of code that this would break?

It seems we must not test any regularization? At the very least, the loss in #1588 is a MWE, but we should probably add a couple regularization tests.

CarloLucibello:
> > What would me a mwe of code that this would break?
> 
> It seems we must not test any regularization? At the very least, the loss in #1588 is a MWE, but we should probably add a couple regularization tests.

sorry I meant what would `@nograd params` break? Is that second order diffs that right now doesn't work in any case due to other reasons or would `@nograd params` cause actual regressions?

CarloLucibello:
I think we should go with this, unless an alternative is proposed. @ToucheSir can you add a test for #1588?

DhairyaLGandhi:
The correct answer has been linked. This is an incorrect solution. Second order does work with params, I'm sorry if that was not clear.

CarloLucibello:
So I assume this PR doesn't break anything we currently have.

Since this PR fixes an outstanding issue introduced by https://github.com/FluxML/Zygote.jl/pull/876, so I think we should consider merging it now unless you manage to wrap up quickly https://github.com/FluxML/Zygote.jl/pull/823 (it's been stale for quite some time, so while referencing that PR is helpful, it still not a solution to the serious problem we face here unless you see it polished and merged).

DhairyaLGandhi:
https://github.com/FluxML/Zygote.jl/pull/876 was introduced after any of these concerns were brought up. Im confused about why you single it out? The second order for implicit params was fragile without https://github.com/FluxML/Zygote.jl/pull/823 anyway

It can be brought back, it's basically ready as is

CarloLucibello:
> [FluxML/Zygote.jl#876](https://github.com/FluxML/Zygote.jl/pull/876) was introduced after any of these concerns were brought up. Im confused about why you single it out? 

Because that appears to be the root cause of the problem, according to the discussion in https://github.com/FluxML/Zygote.jl/issues/941

> It can be brought back, it's basically ready as is

great, let's do it then!

DhairyaLGandhi:
That's a separate issue? As I said it was a fragile path, so it has broken up previously. The push actually catches it earlier in the call stack, so there's a chance it can be fixed for good, but no promises.

ToucheSir:
So to bring us back, this PR is evidently too heavy-handed. FluxML/Zygote.jl#823 seems to be the best option, so what is needed to get that done in a timely manner? Are there any review resources needed? If a decent about of time is required to finalize it, can we craft a more surgical PR that _only_ targets `IdSet` first? Keeping this broken is very not ideal given how much code out there (including many of our examples) uses `params` in the loss function.

DhairyaLGandhi:
Critically; `Params` was not intended to be part of a loss function explicitly, which is why it was implemented in Zygote as opposed to Flux. I would still say that for L2 and so on we use something a bit more simplified. You probbaly don't want to penalise every layer.

ToucheSir:
I didn't mean to imply that `Params` or the functionality to resolve this issue should be in Flux over Zygote. The issue right now is that we inadvertently introduced a breaking change without a breaking release, and this break effects a non-trivial amount of real-world code (including many model-zoo examples). Now I'm of the opinion that calling `params` in a `gradient` context is a code smell and shouldn't be allowed, but we can hardly retroactively issue a major release at this point.

Going to close this PR since it's obviously not the correct solution. Let's pick up in  FluxML/Zygote.jl#823.

CarloLucibello:
Actually,  FluxML/Zygote.jl#823 doesn't seem related to these issues at all, maybe @DhairyaLGandhi could clarify what's the connection

CarloLucibello:
Also, I don't see how `@nograd params` would be incorrect in any scenario. The output of `params`, the parameters themeselves, act as sinks in differentiation, and a `nothing` is backpropagated to `params`, as you can in the stacktrace of https://github.com/FluxML/Flux.jl/issues/1588#issue-874021301:
```
[14] Pullback
    @ ~\.julia\packages\Flux\6o4DQ\src\functor.jl:53 [inlined]
 [15] (::typeof(∂(params)))(Δ::Nothing)
```

CarloLucibello:
Ok, this is a mixed use of the Zygote interface where `@nograd params` causes an issue
```julia
julia> using Flux, Zygote

julia> Zygote.@nograd params

julia> struct A; x; y; end

julia> Flux.@functor a 

julia> a = A(rand(2), rand(2))
A([0.34296822283326556, 0.20719908034130974], [0.41790308062086123, 0.5659970261346654])

# CORRECT
julia> gradient(() -> sum(a.x), params(a)).grads
IdDict{Any, Any} with 2 entries:
  [0.417903, 0.565997] => nothing
  [0.342968, 0.207199] => 2-element Fill{Float64}: entries equal to 1.0

# CORRECT
julia> gradient(a -> sum(a.x), a)
((x = 2-element Fill{Float64}: entries equal to 1.0, y = nothing),)

# WRONG
julia> gradient(a -> sum(params(a)[1]), a)
(nothing,)
```

CarloLucibello:
Also this case is problematic
```
julia> gs = gradient(() -> sum(params(a)[1]), params(a)).grads
IdDict{Any, Any} with 3 entries:
  Buffer{Any, Vector{Any}}(Any[[0.920002, 0.0455781], [0.086206, 0.852121]], false) => Any[2-element Fill{Float64}: entries equal to 1.0, nothing]
  [0.086206, 0.852121]                                                              => nothing
  [0.920002, 0.0455781]                                                             => nothing

julia> gs[params(a)[1]]
```

PR #1595, opened 2021-05-11
optimized hcat of onehot vectors and matrices

### LINKED ISSUE #1594: Too slow hcat of OneHotMatrix.
On Flux 0.12.1 and also on master, I get this
```
using Flux: OneHotArray, OneHotVector, OneHotMatrix, OneHotLike
using BenchmarkTools

ov = OneHotVector(rand(1:10), 10)
om = OneHotMatrix(rand(1:10, 5), 10)

julia> @btime hcat(om, om)
  6.925 μs (58 allocations: 2.47 KiB)
10×10 Flux.OneHotArray{10,2,Vector{Int64}}:
 0  0  1  0  0  0  0  1  0  0
 0  0  0  0  0  0  0  0  0  0
 1  0  0  0  1  1  0  0  0  1
 0  0  0  0  0  0  0  0  0  0
 0  1  0  0  0  0  1  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  1  0  0  0  0  1  0
 0  0  0  0  0  0  0  0  0  0
```

when I do the `hcat` directly by `vcat`ing indices, 
```
faster_hcat(Xs::T...) where T <: OneHotMatrix = T(reduce(vcat, Flux._indices.(Xs)))
```
I get
```
julia> @btime faster_hcat(om, om)
  708.527 ns (8 allocations: 368 bytes)
10×10 Flux.OneHotArray{10,2,Vector{Int64}}:
 0  0  1  0  0  0  0  1  0  0
 0  0  0  0  0  0  0  0  0  0
 1  0  0  0  1  1  0  0  0  1
 0  0  0  0  0  0  0  0  0  0
 0  1  0  0  0  0  1  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  1  0  0  0  0  1  0
 0  0  0  0  0  0  0  0  0  0
```

which is almost order of magnitude faster and allocates much less memory.

I understand this faster `hcat` is much less generic than current `cat` implementation.

Should I make PR with this `hcat` implementation, which would make `hcat` faster?

racinmat:
### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

Fixes #1594 , adds tests for it and also optimized reduce(hcat, xs).
No new features are added, only performance optimization.

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -76,9 +76,9 @@ Base.hcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 2)
 Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 
 # optimized concatenation for matrices and vectors of same parameters
-Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} = OneHotMatrix(vcat(_indices(x), _indices.(xs)...), L)
+Base.hcat(xs::T...) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
 Base.reduce(::typeof(hcat), xs::Vector{T}) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
-Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} = OneHotVector(vcat(_indices(x), _indices.(xs)...), L)
+Base.hcat(xs::OneHotVector{T,L}...) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
 Base.reduce(::typeof(hcat), xs::Vector{OneHotVector{T,L}}) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
 
 batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(_indices.(xs), L)
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -75,6 +75,12 @@ end
 Base.hcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 2)
 Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 
+# optimized concatenation for matrices and vectors of same parameters
+Base.hcat(xs::T...) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
+Base.reduce(::typeof(hcat), xs::Vector{T}) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
+Base.hcat(xs::OneHotVector{T,L}...) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
+Base.reduce(::typeof(hcat), xs::Vector{OneHotVector{T,L}}) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
+
 batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(_indices.(xs), L)
 
 Adapt.adapt_structure(T, x::OneHotArray{<:Any, L}) where L = OneHotArray(adapt(T, _indices(x)), L)
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -5,8 +5,8 @@ struct OneHotArray{T<:Integer, L, N, var"N+1", I<:Union{T, AbstractArray{T, N}}}
   indices::I
 end
 OneHotArray{T, L, N, I}(indices) where {T, L, N, I} = OneHotArray{T, L, N, N+1, I}(indices)
-OneHotArray(indices::T, L::Integer) where {T<:Integer} = OneHotArray{T, L, 0, T}(indices)
-OneHotArray(indices::AbstractArray{T, N}, L::Integer) where {T, N} = OneHotArray{T, L, N, typeof(indices)}(indices)
+OneHotArray(indices::T, L::Integer) where {T<:Integer} = OneHotArray{T, L, 0, 1, T}(indices)
+OneHotArray(indices::I, L::Integer) where {T, N, I<:AbstractArray{T, N}} = OneHotArray{T, L, N, N+1, I}(indices)
 
 _indices(x::OneHotArray) = x.indices
 _indices(x::Base.ReshapedArray{<: Any, <: Any, <: OneHotArray}) =
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -78,10 +78,8 @@ Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 # optimized concatenation for matrices and vectors of same parameters
 Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} =
   OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
-Base.reduce(::typeof(hcat), xs::Vector{T}) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
 Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} =
   OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
-Base.reduce(::typeof(hcat), xs::Vector{OneHotVector{T,L}}) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
 
 batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(_indices.(xs), L)
 
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -76,9 +76,9 @@ Base.hcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 2)
 Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 
 # optimized concatenation for matrices and vectors of same parameters
-Base.hcat(xs::T...) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
+Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} = OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
 Base.reduce(::typeof(hcat), xs::Vector{T}) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
-Base.hcat(xs::OneHotVector{T,L}...) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
+Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} = OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
 Base.reduce(::typeof(hcat), xs::Vector{OneHotVector{T,L}}) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
 
 batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(_indices.(xs), L)
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -76,9 +76,11 @@ Base.hcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 2)
 Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 
 # optimized concatenation for matrices and vectors of same parameters
-Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} = OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
+Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} =
+  OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
 Base.reduce(::typeof(hcat), xs::Vector{T}) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
-Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} = OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
+Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} =
+  OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
 Base.reduce(::typeof(hcat), xs::Vector{OneHotVector{T,L}}) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
 
 batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(_indices.(xs), L)
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -76,9 +76,9 @@ Base.hcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 2)
 Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 
 # optimized concatenation for matrices and vectors of same parameters
-Base.hcat(xs::T...) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
+Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} = OneHotMatrix(vcat(_indices(x), _indices.(xs)...), L)
 Base.reduce(::typeof(hcat), xs::Vector{T}) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
-Base.hcat(xs::OneHotVector{T,L}...) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
+Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} = OneHotVector(vcat(_indices(x), _indices.(xs)...), L)
 Base.reduce(::typeof(hcat), xs::Vector{OneHotVector{T,L}}) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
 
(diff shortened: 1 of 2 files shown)


src/onehot.jl
```diff
@@ -75,6 +75,12 @@ end
 Base.hcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 2)
 Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 
+# optimized concatenation for matrices and vectors of same parameters
+Base.hcat(xs::T...) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
```
darsnack, line 79:
```suggestion
Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} = OneHotMatrix(vcat(_indices(x), _indices.(xs)...), L)
```

src/onehot.jl
```diff
@@ -75,6 +75,12 @@ end
...
+Base.reduce(::typeof(hcat), xs::Vector{T}) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
+Base.hcat(xs::OneHotVector{T,L}...) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
```
darsnack, line 81:
```suggestion
Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} = OneHotVector(vcat(_indices(x), _indices.(xs)...), L)
```

src/onehot.jl
```diff
@@ -75,6 +75,12 @@ end
 Base.hcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 2)
 Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 
+# optimized concatenation for matrices and vectors of same parameters
+Base.hcat(xs::T...) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
+Base.reduce(::typeof(hcat), xs::Vector{T}) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
```
darsnack, line 80: Is this necessary? I see the differences, but I'm curious about the performance benefit.

racinmat: I think this is more about type stability. In master, the reduction converts it to Matrix. See
```julia
julia> reduce(hcat, [om, om])
10×10 Matrix{Bool}:
 0  0  0  0  0  0  0  0  0  0
 0  1  0  0  0  0  1  0  0  0
 1  0  0  0  0  1  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  1  1  0  0  0  1  1
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  1  0  0  0  0  1  0  0
 0  0  0  0  0  0  0  0  0  0
```
and with this change it will keep it one-hot encoded and won't be "materializing" it to dense matrix.
Maybe it would make sense to fill the issue about inconsistency of return type of these two?
See 
```julia
julia> reduce(hcat, [om, om])
10×10 Matrix{Bool}:
 0  0  0  0  0  0  0  0  0  0
 1  0  1  0  0  1  0  1  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  1  0  0  0  0  1  0
 0  0  0  0  1  0  0  0  0  1
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  1  0  0  0  0  1  0  0  0

julia> hcat(om, om)
10×10 Flux.OneHotArray{10,2,Vector{Int64}}:
 0  0  0  0  0  0  0  0  0  0
 1  0  1  0  0  1  0  1  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  1  0  0  0  0  1  0
 0  0  0  0  1  0  0  0  0  1
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  1  0  0  0  0  1  0  0  0
```

johnnychen94:
I believe the root is is that we don't have `similar` defined for `OneHotArray`; once it's defined, it's very likely that many of these overly-verbose optimizations will be unnecessary.

src/onehot.jl
racinmat: May I mark this as resolved?

src/onehot.jl
racinmat: Wait, this is wrong, hcating vectors should not result in OneHotVector, but in OneHotMatrix. While the code works, it's not semantic

racinmat: Also, this is as slow as the original hcat, when you hit the `OneHotArray(indices::AbstractArray{T, N}, L::Integer) where {T, N} = OneHotArray{T, L, N, typeof(indices)}(indices)` constructor where typeof is part of type, you get lots of allocations and thus the slowness.

src/onehot.jl
racinmat: I'm reverting it, because this is as slow as the original. The goal is to avoid calculation of `typeof(indices)` which causes the slowness.

src/onehot.jl
darsnack: Yeah

darsnack (changes requested):
Let's go back to the previously suggested changes (I have commented them again with some updates). If the constructor is causing issues, then why not just change the constructor to:
```
OneHotArray(indices::I, L::Integer) where {T, N, I<: AbstractArray{T, N}} = OneHotArray{T, L, N, I}(indices)
```

Then much more than `hcat` can see the performance benefit.

darsnack:
> I believe the root is is that we don't have similar defined for OneHotArray; once it's defined, it's very likely that many of these overly-verbose optimizations will be unnecessary.

Good point! Though this means we would need to pair `similar` with `setindex!` which would require checks to stop someone from breaking the one-hot behavior. I think directly operating on the index arrays will still see a performance benefit.

src/onehot.jl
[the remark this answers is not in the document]
racinmat: The init is not helping unfortunately, because the problem is not in reduction, but in creation of OneHotMatrix type where `OneHotArray(indices::AbstractArray{T, N}, L::Integer) where {T, N} = OneHotArray{T, L, N, typeof(indices)}(indices)` is hit and `typeof(indices)` needs to be calculated.
I benchmarked it using the following code
```
Base.hcat(xs::T...) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
Base_hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} = OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
Base.hcat(xs::OneHotVector{T,L}...) where {T,L} = OneHotArray{T,L,1,Vector{T}}(reduce(vcat, _indices.(xs)))
Base_hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} = OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
julia> @btime hcat(ov, ov)
  78.401 ns (2 allocations: 112 bytes)
10×2 Flux.OneHotArray{10,2,Vector{Int64}}:
 0  0
 1  1
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0

julia> @btime Flux.Base_hcat(ov, ov)
  6.600 μs (40 allocations: 1.98 KiB)
10×2 Flux.OneHotArray{10,2,Vector{Int64}}:
 0  0
 1  1
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0

julia> @btime hcat(om, om)
  102.754 ns (2 allocations: 176 bytes)
10×10 Flux.OneHotArray{10,2,Vector{Int64}}:
 1  0  0  0  1  1  0  0  0  1
 0  0  0  0  0  0  0  0  0  0
 0  0  1  0  0  0  0  1  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  1  0  0  0  0  1  0
 0  0  0  0  0  0  0  0  0  0
 0  1  0  0  0  0  1  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0

julia> @btime Flux.Base_hcat(om, om)
  6.620 μs (40 allocations: 2.05 KiB)
10×10 Flux.OneHotArray{10,2,Vector{Int64}}:
 1  0  0  0  1  1  0  0  0  1
 0  0  0  0  0  0  0  0  0  0
 0  0  1  0  0  0  0  1  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  1  0  0  0  0  1  0
 0  0  0  0  0  0  0  0  0  0
 0  1  0  0  0  0  1  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
```
and you can see that if we don't pass the `I` parameter into `OneHotArray` type explicitly, it's slow.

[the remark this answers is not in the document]
darsnack: Thanks for benchmarking. Then I suggest using the `vcat(_indices(x), _indices.(xs)...)` syntax as originally suggested, since it is the most readable code. Using the suggestion in my review for the constructor should address the `typeof` slowdown.

It might even be the case that after addressing the constructor issue, that the original `cat` implementation is just as fast.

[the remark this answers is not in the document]
darsnack: It's also possible that `OneHotMatrix` vs. `OneHotLike` is causing the issue. Unfortunately, I can't benchmark this right now since my laptop is in for repair. But I will later tonight.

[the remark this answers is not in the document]
racinmat: I think the slowness is all about the `typeof` and `cat` itself, excluding the creation of type itself is just as fast.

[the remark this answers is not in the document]
darsnack: Yes, you're probably correct. But changing the constructor as suggested should fix that, no? If calling `typeof` is slowing things down, then let's fix it right in the constructor.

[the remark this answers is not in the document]
darsnack: I'm referring to my suggestion in the top level comment [here](https://github.com/FluxML/Flux.jl/pull/1595#pullrequestreview-657849592).

DhairyaLGandhi:
Ideally we would fix it generally rather than for special cases. That's the flux way.

src/onehot.jl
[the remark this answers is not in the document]
racinmat: Oh, my bad, I missed the edited comment, thanks, I'll try that.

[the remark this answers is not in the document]
racinmat: It seems I was wrong with the `typeof`, it seems to be not the typeof, but apparently something else in the construction of the struct. I tried the new constructor you proposed, but it did not help, see

```julia
julia> indices = [2,3,1,3,1]
5-element Vector{Int64}:
 2
 3
 1
 3
 1

julia> @btime OneHotArray{Int64, 5, 1, 2, Vector{Int64}}(indices)
  150.877 ns (1 allocation: 16 bytes)
5×5 Flux.OneHotArray{5,2,Vector{Int64}}:
 0  0  1  0  1
 1  0  0  0  0
 0  1  0  1  0
 0  0  0  0  0
 0  0  0  0  0

julia> @btime OneHotArray{Int64, 5, 1, 2, typeof(indices)}(indices)
  382.759 ns (1 allocation: 16 bytes)
5×5 Flux.OneHotArray{5,2,Vector{Int64}}:
 0  0  1  0  1
 1  0  0  0  0
 0  1  0  1  0
 0  0  0  0  0
 0  0  0  0  0

julia> @btime OneHotArray(indices, 5)
  5.980 μs (39 allocations: 1.89 KiB)
5×5 Flux.OneHotArray{5,2,Vector{Int64}}:
 0  0  1  0  1
 1  0  0  0  0
 0  1  0  1  0
 0  0  0  0  0
 0  0  0  0  0

julia> @which OneHotArray(indices, 5)
OneHotArray(indices::I, L::Integer) where {T, N, I<:AbstractArray{T, N}} in Flux at ...
```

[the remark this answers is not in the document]
darsnack: That's interesting. We'll have to look at `@code_lowered`. I'll look into this later tonight if you don't mind.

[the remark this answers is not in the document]
darsnack: Okay, I looked into it. If you change the constructor definitions to:
```julia
OneHotArray(indices::T, L::Integer) where {T<:Integer} = OneHotArray{T, L, 0, 1, T}(indices)
OneHotArray(indices::I, L::Integer) where {T, N, I<:AbstractArray{T, N}} =
  OneHotArray{T, L, N, N+1, I}(indices)
```
And you use the suggestions in this review, then you recover the performance:
```julia
julia> @btime hcat($om, $om)  353.043 ns (2 allocations: 160 bytes)
10×8 Flux.OneHotArray{10,2,Vector{Int64}}:
 0  0  0  0  0  0  0  0
 0  1  0  0  0  1  0  0
 0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0
 1  0  0  0  1  0  0  0
 0  0  1  0  0  0  1  0
 0  0  0  1  0  0  0  1
 0  0  0  0  0  0  0  0
```

darsnack:
Here is an implementation of `similar`:
```julia
function Base.similar(x::OneHotArray, ::Type{Bool}, dims::Dims)
  indices = similar(_indices(x), Base.tail(dims))

  return OneHotArray(indices, first(dims))
end
```

This would need to be paired with `Base.setindex!` to address the `reduce(hcat, ...)` type instability. Though I am guessing that since `setindex!` will need to perform checks to ensure the one hot property is maintained, this will not be as fast as an optimized custom `reduce(::typeof(hcat), ...)`. I think we should define `similar` no matter what though.

src/onehot.jl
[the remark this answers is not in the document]
racinmat: Perfect, thanks! So the problem was caused by calculating the +1 in the default constructor?

racinmat:
Ok, the performance is now good even for reshaped arrays. Should I now include the `Base.similar` in this PR, or not, because of pairing with `Base.setindex!`?

src/onehot.jl
DhairyaLGandhi: I'm worried about using too-strict constructors here since there is a lot of information on the type itself. It can result in some unintentional promotion

racinmat: Information stored in type should not change during hcat, right? 
Sure for vcat it would change, but I though that for chat it does not.
Or should I remove rows with `Base.reduce(::typeof(hcat)` from this PR and it will be health with later together with `similar`?

DhairyaLGandhi: Yes. In general, it's good practice to apply optimisations that can benefit multiple code paths simultaneously.

src/onehot.jl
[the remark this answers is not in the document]
darsnack: Not that, but the two layers of constructors (first that extracts all type parameters except `N+1`, then another that extracts `N+1`). I am still not sure why this results in so many accumulations, but computing `N+1` directly in the first level constructor fixes the issue.

darsnack:
I think the best move is to remove the `reduce(hcat, ...)` paths from this PR. Leave #1596 open, and we address it later with `similar` and `setindex!`.

src/onehot.jl
```diff
@@ -75,6 +75,12 @@ end
 Base.hcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 2)
 Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 
+# optimized concatenation for matrices and vectors of same parameters
+Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} = OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
```
darsnack, line 79:
```suggestion
Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} =
  OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
```

src/onehot.jl
```diff
@@ -75,6 +75,12 @@ end
...
+Base.reduce(::typeof(hcat), xs::Vector{T}) where T <: OneHotMatrix = T(reduce(vcat, _indices.(xs)))
+Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} = OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
```
darsnack, line 81:
```suggestion
Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} =
  OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
```

darsnack (changes requested):
Just some readability changes in addition to the request to remove the `reduce(hcat, ...)` paths for now.

src/onehot.jl
racinmat: Ok, removing it

test/onehot.jl
```diff
@@ -74,17 +76,23 @@ end
   @testset "Concatenating" begin
     # vector cat
     @test hcat(ov, ov) == OneHotMatrix(vcat(ov.indices, ov.indices), 10)
+    @test hcat(ov, ov) isa OneHotMatrix
```
darsnack, line 79: This tests appear redundant with the one above.

test/onehot.jl
```diff
@@ -74,17 +76,23 @@ end
...
     @test vcat(ov, ov) == vcat(collect(ov), collect(ov))
     @test cat(ov, ov; dims = 3) == OneHotArray(cat(ov.indices, ov.indices; dims = 2), 10)
 
     # matrix cat
     @test hcat(om, om) == OneHotMatrix(vcat(om.indices, om.indices), 10)
+    @test hcat(om, om) isa OneHotMatrix
```
darsnack, line 85: Same here

darsnack (changes requested):

racinmat:
Done, `reduce` and tests for it are removed.

test/onehot.jl
racinmat: I think it is not, because == compares that values are equal, but not types.
Note that after removing the reduce rule, this happens
```julia
julia> reduce(hcat, [om, om]) == OneHotMatrix(vcat(om.indices, om.indices), 10)
true
julia> reduce(hcat, [om, om]) isa OneHotMatrix
false
```
the upper test passes and the latter does not.

test/onehot.jl
racinmat: ditto, comparing values does not guarantee we have same types

test/onehot.jl
darsnack: Yes, just saw this myself in the REPL. Let's keep the test.

test/onehot.jl
```diff
@@ -74,17 +76,23 @@ end
...
     @test vcat(om, om) == vcat(collect(om), collect(om))
     @test cat(om, om; dims = 3) == OneHotArray(cat(om.indices, om.indices; dims = 2), 10)
 
     # array cat
     @test cat(oa, oa; dims = 3) == OneHotArray(cat(oa.indices, oa.indices; dims = 2), 10)
```
darsnack, line 90: Can we add a `isa OneHotArray` test here too?

darsnack (approved):
Looks great, thanks!

test/onehot.jl
racinmat: done

PR #1602, opened 2021-05-20
Added new Chain constructor

VaclavMacha:
This pull request replaces the inner constructor for the Chain type. The goal is to allow using `if` conditions in the building process of the model, as can be seen in the following example. I think adding layers based on some conditions can be helpful and is used, for example, in the PyTorch implementation of the [EfficientNet](https://github.com/lukemelas/EfficientNet-PyTorch/blob/master/efficientnet_pytorch/model.py).

```julia
julia> some_condition = false
false

julia> Chain(
           if some_condition
               Dense(4, 4)
           end,
           Dense(4,2)
       )
Chain(Dense(4, 2))

julia> some_condition = true
true

julia> Chain(
           if some_condition
               Dense(4, 4)
           end,
           Dense(4,2)
       )
Chain(Dense(4, 4), Dense(4, 2))
```

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -24,7 +24,11 @@ true
 """
 struct Chain{T<:Tuple}
   layers::T
-  Chain(xs...) = new{typeof(xs)}(xs)
+
+  function Chain(xs...)
+    layers = filter(!isnothing, xs)
+    return new{typeof(layers)}(layers)
+  end
 end
 
 @forward Chain.layers Base.getindex, Base.length, Base.first, Base.last,


darsnack:
I'm not too familiar with the use-case. Can you say a little more about what the conditions are? Is there a reason the following is not desirable?
```julia
Chain(filter!(!isnothing, [
  if condition Dense(4, 4) end,
  Dense(4, 2)])...)
```
This is not the only syntax that would work, I just tried to match the approach in the PR. In general, there are many ways in Julia to conditionally build up a vector of layers and then just call `Chain(layers...)`.

DhairyaLGandhi:
Filtering `nothing` doesn't seem like it would hurt the constructor, there isn't a `skipnothing` is there? That would make this nicer. could you say why this would be preferable to nesting chains within chains? That way one can have the conditions outside of the chain constructors and get the same result.

ToucheSir:
Moreover, there's nothing wrong with subbing in `identity` in place of a layer. e.g.
```julia
Chain(
  some_condition ? Dense(4, 4) : identity,
  Dense(4,2)
)
```
This pattern is used all the time in PyTorch with `nn.Sequential` and `nn.Identity`. It also has the benefit of maintaining stable ordinal positions for all of the sublayers (e.g. batchnorm is always at index 2 of the conv block). For scenarios where you _must_ not have a layer at all, there's always the option of constructing a list incrementally beforehand (with any amount of fancy control flow) and passing it all to `Chain` at once.

darsnack:
As of now, I can't see any reason that someone would want to keep `nothing` in the list, but I'm still wary of adding this filtering to the inner constructor. For two reasons:
1. In Flux, we err on the side of not messing with what the user passes in unless it is a common case that provides a big boost to ease of use.
2. It will just be an extra corner case to support should we decide to add new constructor options for `Chain` in the future.

I agree that the first example is less verbose, but, to me, they are both equally readable. So, I would rather keep `Chain` as flexible as possible for now.

ToucheSir:
Again, I'd argue that special-casing `nothing` in the constructor isn't much cleaner than just using a ternary:
```julia
model = Chain(
    # Expansion phase
    # Note: I inverted the condition purely for aesthetic purposes
    expand_ratio == 1 ? identity : Chain(
        Conv((1,1), inp => mid; pad, stride),
        BatchNorm(mid, swish; momentum, ϵ),
    ),

    # Depthwise convolution phase
    DepthwiseConv(kernel_size, mid => mid; pad, stride),
    BatchNorm(mid, swish; momentum, ϵ),

    # Squeeze & Excitation optimation
    SkipConnection(
        Chain(
            GlobalMeanPool(),
            Conv((1, 1), mid => squeezed, swish),
            Conv((1, 1), squeezed => mid, σ),
        ), (A,B) -> A.*B
    ),

    Conv((1, 1), mid => out),
    BatchNorm(out; momentum, ϵ),
)
```
I'd be willing to bet that a PyTorch solution using `Sequential` (which the linked file does not) would be best served by following a similar approach.

VaclavMacha:
I tried to remove the layer completely in a nice way, but replacing it with identity is a better approach. Thank you very much for your comments and suggestions.

PR #1607, opened 2021-06-02
document Join and Split error

magicly:
```julia
(m::Join)(xs::Tuple) = m.combine(map((f, x) -> f(x), m.paths, xs))
```
will get a `Vector` of `Tuple` of `Vector`.  Should use `mapreduce` as `Flux.Parallel` or just use `(m::Join)(xs::Tuple) = m.combine(map((f, x) -> f(x), m.paths, xs)...)`.

`CustomSplit` should be `Split`.


DhairyaLGandhi:
Thanks for this! LGTM

DhairyaLGandhi (approved):

Issue #1414, opened 2020-12-02
performance Dense layer on CPU


CarloLucibello:
This is just to track the performance of the Dense layer on CPU. I use the following script:

```julia
using BenchmarkTools, Flux
using Zygote: pullback

using LinearAlgebra
BLAS.set_num_threads(1)

function perf_test(n)
    r = rand(Float32, n, n) 
    d = Dense(n, n)
    println("  FORW")
    @btime sum($d($r))
    println("  GRADIENT")
    @btime gradient(() -> sum($d($r)), $(Flux.params(d)))
    @btime gradient((d) -> sum(d($r)), $d)
    println("  PULLBACK")
    y, back =  pullback((d) -> sum(d(r)), d)
    @btime pullback((d) -> sum(d($r)), $d)
    @btime $back(1f0)
end

println("SMALL NET n=2")
perf_test(2)
println("MEDIUM NET n=20")
perf_test(20)
println("LARGE NET n=200")
perf_test(200)
println("VERY LARGE NET n=2000")
perf_test(2000)
```
and on my system:
```julia
julia> versioninfo()
Julia Version 1.5.3
Commit 788b2c77c1* (2020-11-09 13:37 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-10.0.1 (ORCJIT, skylake)
```
I obtain the following output
```julia
SMALL NET n=2
  FORW
  98.204 ns (2 allocations: 192 bytes)
  GRADIENT
  1.939 μs (37 allocations: 2.75 KiB)
  991.300 ns (31 allocations: 1.67 KiB)
  PULLBACK
  121.768 ns (4 allocations: 400 bytes)
  842.190 ns (25 allocations: 944 bytes)
MEDIUM NET n=20
  FORW
  992.300 ns (2 allocations: 3.53 KiB)
  GRADIENT
  10.254 μs (41 allocations: 11.39 KiB)
  8.985 μs (35 allocations: 10.31 KiB)
  PULLBACK
  1.041 μs (4 allocations: 3.73 KiB)
  7.798 μs (29 allocations: 6.22 KiB)
LARGE NET n=200
  FORW
  203.562 μs (4 allocations: 312.66 KiB)
  GRADIENT
[...]

PR #1613, opened 2021-06-10
use ArrayInterface.restructure in update!

### LINKED ISSUE #1510: Minimizing sum fails
```julia
T = randn(10,10)
ps = Flux.params(T)
opt=ADAM()
for iter = 1:100
    gs = gradient(ps) do
        sum(T)
    end
    Flux.update!(opt, ps, gs)
end
```

Throws the following error:

> 
> ArgumentError: Cannot setindex! to 0.0009999999900000003 for an AbstractFill with value 1.0.
> 
> Stacktrace:
>  [1] setindex! at /home/cossio/.julia/packages/FillArrays/tE9Xq/src/FillArrays.jl:41 [inlined]
>  [2] _setindex! at ./abstractarray.jl:1176 [inlined]
>  [3] setindex! at ./abstractarray.jl:1153 [inlined]
>  [4] macro expansion at ./broadcast.jl:932 [inlined]
>  [5] macro expansion at ./simdloop.jl:77 [inlined]
>  [6] copyto! at ./broadcast.jl:931 [inlined]
>  [7] copyto! at ./broadcast.jl:886 [inlined]
>  [8] materialize! at ./broadcast.jl:848 [inlined]
>  [9] materialize!(::FillArrays.Fill{Float64,2,Tuple{Base.OneTo{Int64},Base.OneTo{Int64}}}, ::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2},Nothing,typeof(*),Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2},Nothing,typeof(/),Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2},Nothing,typeof(/),Tuple{Array{Float64,2},Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0},Nothing,typeof(-),Tuple{Int64,Float64}}}},Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2},Nothing,typeof(+),Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2},Nothing,typeof(sqrt),Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2},Nothing,typeof(/),Tuple{Array{Float64,2},Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0},Nothing,typeof(-),Tuple{Int64,Float64}}}}}},Float64}}}},Float64}}) at ./broadcast.jl:845
>  [10] apply!(::ADAM, ::Array{Float64,2}, ::FillArrays.Fill{Float64,2,Tuple{Base.OneTo{Int64},Base.OneTo{Int64}}}) at /home/cossio/.julia/packages/Flux/05b38/src/optimise/optimisers.jl:177
>  [11] update!(::ADAM, ::Array{Float64,2}, ::FillArrays.Fill{Float64,2,Tuple{Base.OneTo{Int64},Base.OneTo{Int64}}}) at /home/cossio/.julia/packages/Flux/05b38/src/optimise/train.jl:23
>  [12] update!(::ADAM, ::Params, ::Zygote.Grads) at /home/cossio/.julia/packages/Flux/05b38/src/optimise/train.jl:29
>  [13] top-level scope at In[89]:8
>  [14] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091
>

CarloLucibello:
Suggestion coming from @ChrisRackauckas in https://github.com/FluxML/Zygote.jl/pull/989. 
Now `update!` handles basically any gradient Zygote emits, e.g. FillArrays and Zygote.OneElement. 

Fix #1510


ChrisRackauckas:
It would be interesting to test this on some odd case like where Params contains a [ComponentArray](https://jonniedie.github.io/ComponentArrays.jl/dev/quickstart/) and the user just assumes that the gradient will be able to do `grad.a == grad[1]`.

DhairyaLGandhi:
Better handling is present in optimisers.jl

src/optimise/train.jl
```diff
@@ -18,14 +18,17 @@ Perform an update step of the parameters `ps` (or the single parameter `p`)
 according to optimizer `opt`  and the gradients `gs` (the gradient `g`).
 
 As a result, the parameters are mutated and the optimizer's internal state may change.
+The gradient could be mutated as well.
 """
 function update!(opt, x, x̄)
-  x .-= apply!(opt, x, x̄)
+  x̄r = ArrayInterface.restructure(x, x̄) # address some cases where Zygote's
```
mcabbott, line 24: Are there solutions without ArrayInterface? Since this is (I think) intended as a quick fix pending a better design.

Perhaps it could just test `ismutable`, and then use broadcasting to make a similar array (since `x` is already assumed mutable)?
```
function update!(opt, x, x̄)
  # quick fix for #1510, Zygote's output may not be mutable:
  x̄r = ismutable(x̄) ? x̄ : x̄ .+ false .* x
  x .-= apply!(opt, x, x̄r)
end
```
Or perhaps better to test  `x̄ isa DenseArray` or something, maybe there are mutable structs which aren't writeable... 
```
function update!(opt, x, x̄::DenseArray)
  x .-= apply!(opt, x, x̄r) # as before
end
function update!(opt, x, x̄)
  x .-= apply!(opt, x, x̄ .+ false .* x)  # fix for #1510, Zygote's output may not be mutable
end
```

ChrisRackauckas: `ismutable` is wrong: it doesn't tell you if an array is mutable. It tells you if a type is a mutable type, which is different. 

```julia
julia> ismutable(Adjoint(rand(4)))
false
```

Base Julia does not have the right verbiage to answer these kinds of questions in an adequately generic way, which is the reason for ArrayInterface.jl becoming a standard trait interface to cover these kinds of issues.

ChrisRackauckas: `ArrayInterface.ismutable(x)` does signify "array is mutable"

CarloLucibello: I'm not worried about carrying on the ArrayInterface dependence, but I do want to get the semantics right.
Is `ArrayInterface.ismutable(x)` more appropriate than `ArrayInterface.destructure`?

mcabbott: Yes to Adjoint, but that would fail to the safe side, of making one too many copies.

ChrisRackauckas: You mean `ArrayInterface.restructure`? `restructure` is a generalization of `reshape` or `adapt` to match the general characteristics of an array. For example, ComponentArrays.jl encodes in type information that for `x = ComponentArray(a=1.0, b=[2, 1, 4], c=c)`, `x.a == 1.0`. `ArrayInterface.restructure(x::ComponentArray,y::Vector)` would change `y` to be the correct ComponentArray (or error if it's not sized correctly), so `ArrayInterface.restructure(x::ComponentArray,y::Vector).a` would work.

Now it all comes back to what you believe is a good interface for the gradient. There is a natural relationship where the gradient has 1 value for each parameter: it's always the same size of values, and so you can match the structure of the user's input with the gradient output if you wanted to. If they say `p` is a `Tridiagonal`, the gradient will only be non-zero on the tridiagonal elements, and so in theory you could give them back a `Tridiagonal`. If they used a `ComponentArray` to name their parameters, then you could give them back an array such that `grad.b` gives back an array of 3 things, just like they wrote their model to expect. What `ArrayInterface.restructure` is doing here is encoding that assumption: I will change the output gradient to match all of the structure that the user's code expects their parameters to have. To me, given the natural 1-1 mapping of the gradient to parameters, I think that's a very nice quality to keep for when people write these models.

In fact, thinking about a rewrite of Flux to use explicit parameters, in theory you could just use one `ComponentArray` to store all of the parameters of a `Chain`, and then `grad.W1` and `grad.b2` would be endowed with a natural meaning. Not only that, but the entire parameters would be represented as a single contiguous array, which means it would be a bit more performant, linear algebra would naturally work on that whole thing (a matmul against the `grad` vector would "just work"), and that last piece means that it would be easy to implement L-BFGS on this interface (since that requires the ability to do linear algebra unlike other optimizers). That's a bit of a tangent, but IMO that's a direction Flux should be moving and it's all made compatible if you just match the user's type assumptions with the 1-1 mapping in the output parameters.

Now as I showed in the earlier example, you could instead say that's not necessarily the most efficient representation, and so other things could come out. So you could take a gradient with respect to an Array, and then Zygote can spit out a FillArray, and the user is expected to recognize that the gradient does not have the same properties, actions, etc. as their parameter array (here, it does not support mutation). It's now a user issue to change from the seemingly random but possibly more efficient representation that Zygote chose (not entirely random, but since it's dependent on the function being differentiated, to users it feels like magic) to a representation they know for their parameters. Is that better or worse? That's a personal decision. However, I would note that the memory and compute usage of reverse mode automatic differentiation far outstrips the cost of doing one `ArrayInterface.restructure` at the end, since the 1-1 mapping means this is just O(p) and reverse mode does a lot more to parameters than copy them into a new array, and so I would hesitate to say that the interface without restructuring is really better performance in the vast majority of cases.

mcabbott: > and so you can match the structure of the user's input with the gradient output

ChainRules is doing a lot of thinking about this. The gradient type need not be the same as the primal, but must be compatible, in pretty much the sense that you can perturb the primal by the gradient. Or some multiple thereof -- gradients always live in a vector space, even if the primal does not (e.g. it has some constraint like `sum(x)==1`.)

> easy to implement L-BFGS on this interface 

Another possible direction is that taken by OptimKit.jl, which defines algebraic properties the gradient must have, without needing it to be as simple as a vector. And certainly not a mutable vector.

But to be clear, this PR is much more modest. Given that Flux's optimisers at the moment demand that gradients be mutable arrays, the goal is to catch a few edge cases where they aren't. Temporarily, until the re-worked optimisers are available. That's why I wonder whether doing something safe & simple (like testing `xbar isa DenseArray`) might be preferable here to fancy traits. If the band-aid fix makes one more copy than strictly necessary, in some cases, maybe that's OK.

CarloLucibello: If we want to be a bit more liberal and allow for possible reshapes to happen we could go with `ArrayInterface.restructure`, otherwise `xbar isa DenseArray` is also fine. I don't have a strong opinion on either directions

ToucheSir: If the worry is about taking on an ArrayInterface dependency, I'd note that the plan was to have Optimisers.jl do so already. Parts of Flux have historically played fast and loose with assumptions around array types and it's bitten us more than once.

CarloLucibello:
> It would be interesting to test this on some odd case like where Params contains a ComponentArray and the user just assumes that the gradient will be able to do grad.a == grad[1].

@ChrisRackauckas is the test I added what you meant here? Although probably that was already working, the returned gradient from Zygote is a ComponentArray

ChrisRackauckas:
>@ChrisRackauckas is the test I added what you meant here? Although probably that was already working, the returned gradient from Zygote is a ComponentArray

Yup that's the test. It was not working because if you did that on the `sum` example, you'd get back a FillArray in previous versions. So it's not that it was always not working, it was Zygote gets a little fussy and chooses when it wanted to obey those rules 😆

src/optimise/train.jl
darsnack: Yeah and currently the approach in Optimisers.jl is `ArrayInterface.ismutable`, but I agree with Chris on using `ArrayInterface.restructure` long term instead. Especially since Optimisers.jl is the first step to explicit params.

CarloLucibello: @darsnack I don't get which approach you prefer for this PR

darsnack: I think for this PR, `ArrayInterface.ismutable` is my preference, but I wouldn't actively oppose the other option.

ToucheSir: Since `restructure` tries to maintain the original array type, isn't `ismutable` still required when the original array is immutable?

mcabbott: But this is called when Flux is trying to update the parameter `x` by mutating, so I think it ought to be safe to assume that's mutable here.

ToucheSir: Yes and no, IIRC the only reason it works now is because certain immutable parameter types are unceremoniously transformed to dense arrays during AD?

mcabbott: Not sure what you're hinting at. I thought that using Flux's optimisers with an immutable array `x` inside Params would simply fail, as it will try to update it. This update is after the AD and would be a problem even if the AD returned an Array.

ToucheSir: I have a vague recollection of `Fill`s not being handled universally. https://github.com/FluxML/Flux.jl/issues/1510 and/or https://github.com/FluxML/Flux.jl/issues/1414, maybe?

mcabbott: Right, #1510 is what this PR fixes. It’s an immutable gradient dx, which Flux wants to use to update a mutable parameter x.

darsnack: Ah, I think my comment muddled the conversation, since what I was referring to in Optimisers.jl is about x not dx. Sorry for the confusion.

ToucheSir: Oh no, I was referring to `x` as well. Seems like this is no longer an issue though, as to my understanding `Zeros` no longer takes part in the gradients as of https://github.com/FluxML/Flux.jl/pull/1379.

mcabbott: Ok! I guess I see this PR as a band-aid for the current structure of things. 

I'd forgotten about that weird issue caused by Zeros. However, honest parameters expressed as `x::SArray` might not be crazy, and a design that allowed them might be nice. Perhaps a smaller step would be a design that didn't need to mutate `dx` to do `x .+= ε .* dx`. If the former happened before the latter then some variants of this PR would need re-visiting... which seems OK.

ToucheSir: I don't think it's a tough fix, essentially https://github.com/FluxML/Flux.jl/blob/v0.12.4/src/optimise/train.jl#L29 needs to be rewritten to something like:
```julia
xs[x] = update!(opt, x, gs[x])
```
Where update always returns the transformed parameter and can optionally mutate it to save an allocation. This would require implementing `setindex` on `Params`, but it's the only solution I can think of that preserves both the API and semantics.

mcabbott: If you replace what object is stored in the `IdDict`, this won't be picked up by the model. I agree that you want some mutate-where-possible-but-always-return scheme, but maybe it has to take & return the whole model? IIRC this was being discussed at some point, Functors?

ToucheSir: Pretty much. It's _the_ inherent limitation of implicit params, which is why Optimisers.jl focuses on explicit params and being able to traverse through structures (à la JAX's pytrees) instead. I would love it if we could nuke implicit params overnight, but there are still a few UX things to figure out with explicit params before then.

ToucheSir (approved):
Since this has come up before and Optimisers.jl needs some more time to bake, shall we get this merged?

CarloLucibello:
@darsnack or @DhairyaLGandhi  can I get an approval here?

DhairyaLGandhi:
What needs to bake in optimisers.jl?

PR #1614, opened 2021-06-10
placeholder adjoint for params

### LINKED ISSUE #1588: Regularization example from docs can't be differentiated

CarloLucibello:
Fix #1588, Fix https://github.com/FluxML/Zygote.jl/issues/941

Conservative version of #1590, which throws an error when the pullback receives a not-nothing sensitivity. 
This is strictly better than what we currently have, fixes a relevant issue without introducing any regression. 

Unfortunately, there are some corner cases, marked as `broken_tests` that this does not address. I could not manage to trigger the `params_pullback` branch throwing the error: the pullback always receives `nothing` which seems a little odd to me.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -56,7 +56,7 @@ end
 
 @adjoint function params(m...)
   params_pullback(Δ::Nothing) = nothing
-  params_pullback(Δ) = (println(Δ); error("derivations through params not supported yet"))
+  params_pullback(Δ) = (println(Δ); error("derivation through `params` not supported yet"))
   params(m...), params_pullback
 end
 
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -54,6 +54,12 @@ function params(m...)
   return ps
 end
 
+@adjoint function params(m...)
+  params_pullback(Δ::Nothing) = nothing
+  params_pullback(Δ) = (println(Δ); error("derivations through params not supported yet"))
+  params(m...), params_pullback
+end
+
 function loadparams!(m, xs)
   for (p, x) in zip(params(m), xs)
     size(p) == size(x) ||
--- a/test/functor.jl
+++ b/test/functor.jl
@@ -1,0 +1,8 @@
+@testset "use params in gradient context"
+    m = Chain(Dense(3,2), Dense(2,2))
+    ps = Flux.params()
+    gs = gradient(() -> sum(sum(p) for p in Flux.params(m)), ps)
+    for p in ps
+        @test gs[p] ≈ ones(size(p))
+    end    
+begin
--- a/test/functor.jl
+++ b/test/functor.jl
@@ -20,4 +20,4 @@
 
     gs = gradient(() -> sum(params(m)[1]), params(m))
     @test_broken gs[params(m)[1]] !== nothing
-begin
+end
--- a/test/functor.jl
+++ b/test/functor.jl
@@ -1,4 +1,4 @@
-@testset "use params in gradient context"
+@testset "use params in gradient context" begin
     m = Chain(Dense(3,2), Dense(2,2))
     ps = Flux.params()
     gs = gradient(() -> sum(sum(p) for p in Flux.params(m)), ps)
--- a/test/functor.jl
+++ b/test/functor.jl
@@ -5,4 +5,19 @@
     for p in ps
         @test gs[p] ≈ ones(size(p))
     end    
+
+    w1, w2 =  rand(2), rand(2)
+    ps = Flux.params(w1, w2)
+    gs = gradient(() -> sum(sum(p) for p in Flux.params(w1, w2)), ps)
+    for p in ps
+        @test gs[p] ≈ ones(size(p))
+    end
+
+    # BROKEN TESTS
+    m = Chain(Dense(3,2), Dense(2,2))
+    @test_broken gradient(m -> sum(params(m)[1]), m) != (nothing, )
+    @test_broken gradient(m -> sum(params(m)[1]), m) != (nothing, )
+
+    gs = gradient(() -> sum(params(m)[1]), params(m))
+    @test_broken gs[params(m)[1]] !== nothing
 begin
--- a/test/functor.jl
+++ b/test/functor.jl
@@ -1,6 +1,6 @@
 @testset "use params in gradient context" begin
     m = Chain(Dense(3,2), Dense(2,2))
-    ps = Flux.params()
+    ps = Flux.params(m)
     gs = gradient(() -> sum(sum(p) for p in Flux.params(m)), ps)
     for p in ps
         @test gs[p] ≈ ones(size(p))
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.12.7"
+version = "0.12.8"
 
 [deps]
 AbstractTrees = "1520ce14-60c1-5f80-bbc7-55ef81b5835c"
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -45,6 +45,10 @@ end
   include("outputsize.jl")
 end
 
+@testset "functor" begin
+  include("functor.jl")
+end
+
 @testset "CUDA" begin
   if CUDA.functional()
     include("cuda/runtests.jl")


DhairyaLGandhi:
The zygote pr has tests now, did you see those? We had discussed merging it now since it seems to solve the most pressing issue. The design here is not never materialize objects from `Params` to maintain that refs in the `Params` point to the right place.

CarloLucibello:
which Zygote PR?

DhairyaLGandhi:
https://github.com/FluxML/Zygote.jl/pull/823

CarloLucibello:
FluxML/Zygote.jl#823 by your own admission doesn't solve the pressing issues, which are those fixed by this PR. 
Besides that, FluxML/Zygote.jl#823 is a mess right now: it tries to do 2 unrelated things, introduces outstanding bugs in both things, and CI is not passing.
I think FluxML/Zygote.jl#823 is totally unrelated to the issues this PR and #1590 are trying to solve and no argument has been given of the converse, so I don't see why we keep discussing it.

DhairyaLGandhi:
I'm sorry? I've added another test that iterates over params and calculates the squared sum over them to show it working. Maybe that was the one you wanted?

CarloLucibello:
Iterating over already collected `Params` is already working on master:
```
julia> using Flux

julia> m = Chain(Dense(3,2), Dense(2,2))
Chain(Dense(3, 2), Dense(2, 2))

julia> ps = Flux.params(m);

julia> gs = gradient(()->sum(sum(p) for p in ps), ps)
Grads(...)

julia> gs[ps[1]]
2×3 Fill{Float32}: entries equal to 1.0
```
so nothing FluxML/Zygote.jl#823 solves. 
The issue this PR fixes is with calls to `Flux.params` within a gradient context

DhairyaLGandhi:
Right, and we had discussed adding an explicit error for doing that cc @ToucheSir

CarloLucibello:
I don't think removing features is a good idea. Iterating over parameters can be convenient in some situations, I've done it a few times in pytorch, Flux should avoiding imposing unneeded constraints on users. Moreover, since is a feature we used to have, and has been documented here and used in model zoo, and we inadvertently broke it recently, we just have to fix it, doing anything else would be breaking. We can discuss whether to deprecate this feature or not (I think we shouldn't) but that would be a separate issue/PR

ToucheSir:
@CarloLucibello the discussion was to prevent calling `params` in a gradient context, not iterating over them in general. I feel somewhat strongly about not letting the user do that and nudging them towards implicit params. This PR accomplishes that, though I should note that it will still be a breaking change because model zoo examples will still error.

However, I was convinced that there might be valid use-cases for building up params in a gradient context. The question is whether that building needs to be differentiated through or can be sequestered in a `nograd` block. In #1590 it was mentioned that the latter could break higher-order AD, so does anyone have a concrete example I could look over?

CarloLucibello:
> @CarloLucibello the discussion was to prevent calling params in a gradient context, not iterating over them in general

right

> This PR accomplishes that, though I should note it will still be a breaking change because downstream code/model zoo examples will still error (just more informatively now).

This PR does fix those issues. It fixes those issues because for some reason not clear to me the `params_pulback(Δ)` always receives a `nothing`, I wasn't able to produce any case where `Δ != nothing`. 

> However, I was convinced that there might be valid use-cases for building up params in a gradient context. The question is whether that building needs to be differentiated through or can be sequestered in a nograd block.

One of the broken_tests I add here is a case where I think differentiation through `params` is needed. 
```
# ON THE PR
julia> m = Chain(Dense(3,2), Dense(2,2))
Chain(Dense(3, 2), Dense(2, 2))

julia> gradient(m -> sum(params(m)[1]), m)
(nothing,)
```
Since also in this case `Δ == nothing` though, I thing something is wrong in Zygote. 
This other broken test is weird, because the gradient is there but associated to a buffered param:
```
julia> gradient(() -> sum(params(m)[1]), params(m)).grads
IdDict{Any, Any} with 5 entries:
  Float32[-0.913362 -0.970178 -0.688053; -0.285252 0.748989 -0.453376]                       => nothing
  Buffer{Any, Vector{Any}}(Any[Float32[-0.913362 -0.970178 -0.688053; -0.285252 0.748989 -0… => Any[2×3 Fill{Float32}: entries equal to 1.0, nothing, nothing, nothing]
  Float32[0.0, 0.0]                                                                          => nothing
  Float32[0.0, 0.0]                                                                          => nothing
  Float32[1.07433 -0.0018764; -0.725969 0.996931]                                            => nothing
```
 > In #1590 it was mentioned that the latter could break higher-order AD, so does anyone have a concrete example I could look over?

None that I know

CarloLucibello:
merge?

test/functor.jl
```diff
@@ -0,0 +1,23 @@
+@testset "use params in gradient context" begin
+    m = Chain(Dense(3,2), Dense(2,2))
+    ps = Flux.params()
+    gs = gradient(() -> sum(sum(p) for p in Flux.params(m)), ps)
+    for p in ps
+        @test gs[p] ≈ ones(size(p))
```
DhairyaLGandhi, line 6: This shouldn't happen. It indicates that the references within `Params` have been changed during taking the gradient somewhere, which should be disallowed.

darsnack: Are you referring to `isapprox`?

CarloLucibello: I don't understand the comment. The test is fine

ToucheSir: Isn't `ps` empty as well? Would it not be easier to test if `ps` and `gs` are empty instead?

DhairyaLGandhi: Both of them should be empty, there should be no params to iterate over, there shouldn't be any grads either.

darsnack: Oh I missed that `ps` was empty. Probably this test "passes" because nothing actually runs. I think the test was meant to be `ps = Flux.params(m)`.

test/functor.jl
```diff
@@ -0,0 +1,23 @@
+@testset "use params in gradient context" begin
+    m = Chain(Dense(3,2), Dense(2,2))
+    ps = Flux.params()
```
CarloLucibello, line 3:
```suggestion
    ps = Flux.params(m)
```

DhairyaLGandhi:
Ref https://github.com/FluxML/Zygote.jl/pull/992

test/functor.jl
CarloLucibello: yes, fixed

DhairyaLGandhi:
This was never called out to be a feature. In fact, model zoo should be updated to not use it in this manner.

CarloLucibello:
Even if we don't advertise this feature, it comes very natural and people are bitten by it, e.g. the recent https://discourse.julialang.org/t/weight-regularisation-which-iterates-params-m-in-flux-mutating-arrays-is-not-supported/64314

Pytorch supports using `model.parameters()` in gradient contexts, and actually I have used it a few times. We shouldn't force people away from their programming patterns if not strictly necessary, so we should merge this

darsnack (dismissed):
https://github.com/FluxML/Zygote.jl/pull/992 never fixed these issues, and I agree that there's no reason not allow common design patterns (especially when we have an easy fix).

Before merging, can we revisit the broken tests? Are they still broken in the same way?

DhairyaLGandhi:
That is not a good reason to introduce code that can  produce incorrect results. I would rather get a clear MWE that does not depend on Flux. The tests in Zygote capture this scenario already.

darsnack:
1) the code doesn't produce incorrect results
2) this specifically deals with `Flux.params`, so the MWE has to involve Flux

DhairyaLGandhi:
Are the tests in Zygote incorrect?

darsnack:
The Zygote tests only test for `Zygote.Params` not the full route through `Flux.params`.

src/functor.jl
```diff
@@ -54,6 +54,12 @@ function params(m...)
   return ps
 end
 
+@adjoint function params(m...)
+  params_pullback(Δ::Nothing) = nothing
+  params_pullback(Δ) = (println(Δ); error("derivation through `params` not supported yet"))
```
mcabbott, line 59:
```suggestion
  params_pullback(Δ) = (@show(typeof(Δ)); error("differentiation through `params` is not supported yet"))
```

mcabbott (approved):
This looks fine to me, it's a pretty small step that fixes some things. And will give an error before it can contribute to wrong answers.

ToucheSir:
It might be what other behaviour did work prior to https://github.com/FluxML/Zygote.jl/pull/876 as well? I certainly don't remember any bug reports about broadcasting in `Params` not working, but then again I don't recall seeing any code doing so either.

mcabbott:
Was  FluxML/Zygote.jl#876 intended to address things with Params? It has no description and tests precisely the behaviours it defines. 

I don't have much sense of what used to work here. Agree that broadcasting seems weird, but you can broadcast most iterators & so someone will...

ToucheSir:
It limited the push adjoint to vectors only, which meant other data structures like IdSet no longer got a free ride and started throwing mutation errors. https://github.com/FluxML/Zygote.jl/pull/992 should've handled exactly that though, so I think we should look into it as well before trying to address things at a higher level.

DhairyaLGandhi:
If there are pending issues, those need to be addressed in Zygote. Note too, https://github.com/FluxML/Zygote.jl/pull/992 already has tests over iteration of `Params` etc. So if there's something in `params` that needs to be addressed, it can be mwe'd to be fixed at the correct place.

ToucheSir:
Right, my point was that `params` hasn't changed in literally 2 years and (at least appears to have) behaved this whole time. So if a Zygote-only change triggered the misbehaviour, then a Zygote-only change should be sufficient to resolve it.

mcabbott:
Trying the same examples I had above, on Zygote before those PRs -- one loud error (the same, broadcasting) one silent wrong answer (also wrong with this PR) and one right answer which master + this PR gets wrong:
```
julia> using Flux, Zygote

(jl_ykYRbP) pkg> st
      Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_ykYRbP/Project.toml`
  [587475ba] Flux v0.12.4
  [e88e6eb3] Zygote v0.6.3

julia> c = rand(3);

julia> gs = gradient(params(c)) do 
         sum(sum(x) for x in params(c))
       end
Grads(...)

julia> gs[c]
3-element Fill{Float64}: entries equal to 1.0

julia> gs = gradient(params(c)) do 
         sum(sum, params(c))
       end
Grads(...)

julia> gs[c]
3-element Fill{Float64}: entries equal to 1.0

julia> gs = gradient(params(c)) do 
         sum(sum.(params(c)))
       end
ERROR: Mutating arrays is not supported

julia> gs = gradient(params(c)) do 
         sum(params(c)[1])
       end
Grads(...)

julia> gs[c] 
3-element Fill{Float64}: entries equal to 1.0

julia> ps = params(c);

julia> gs2 = gradient(params(c)) do 
           sum(ps[1])
       end
Grads(...)

julia> gs2[c] == nothing
true
```
Then it does sound like this *can* be fixed within Zygote. The patch to Flux here avoids the problem in several cases, although it's unfortunate that it makes one noisy error into a silent wrong answer.

darsnack:
Keeping broadcasting aside, [these failures](https://github.com/FluxML/Zygote.jl/pull/992#issuecomment-867340277) were the original motivation for focusing on `Flux.params` in addition to `Zygote.Params`. The error seems to be something subtle because plain iteration over `Params` works fine.

darsnack:
Here is your MWE @DhairyaLGandhi.

```julia
(jl_BPLyJP) pkg> st
      Status `/private/var/folders/d_/tctk50k95_1f2jfc3pq05y_m0000gn/T/jl_BPLyJP/Project.toml`
  [587475ba] Flux v0.12.4
  [e88e6eb3] Zygote v0.6.14 `https://github.com/FluxML/Zygote.jl.git#master`

julia> using Flux

julia> c = rand(3);

julia> gs = gradient(params(c)) do
         sum(sum(x) for x in params(c))
       end
Grads(...)

julia> gs[c]
3-element Fill{Float64}, with entries equal to 2.0

julia> c = fill(2f0, 3);

julia> gs = gradient(params(c)) do
         sum(sum(x .^ 2) for x in params(c))
       end
Grads(...)

julia> gs[c]
3-element Vector{Float32}:
 5.0
 5.0
 5.0
```

I believe the error is coming from a +1 due to [this line](https://github.com/FluxML/Zygote.jl/blob/c91ff1fc3a9f85be65468fc8678022ee50a1520e/src/compiler/interface.jl#L187) from https://github.com/FluxML/Zygote.jl/pull/992. Notably, https://github.com/FluxML/Zygote.jl/pull/992 had no tests for `push!(ps::Params, x)` which missed this extra counting.

mcabbott:
Since broadcasting over `Params` seemingly never worked, we can just do `Broadcast.broadcastable(::Zygote.Params) = error("can't do that, try iterating, etc")`.

mcabbott:
> I believe the error is coming from a +1 due to this line from FluxML/Zygote.jl#992. 

The gradient for `push!` from https://github.com/FluxML/Zygote.jl/pull/876 also seems so have many bugs, on the simplest test I could think of it gets every array wrong. (Its tests use the length of the mutated list, not the argument, so don't even spot the length mismatch.)
```
julia> gradient((xs, y) -> sum(abs2, push!(xs, y)[1]), [[1,2], [3,4]], [5,6])
([[2, 4], nothing, nothing], 2-element Ones{Int64})
([[2, 4], nothing], nothing)  # correct

julia> gradient((xs, y) -> sum(abs2, push!(xs, y)[2]), [[1,2], [3,4]], [5,6])
([nothing, [6, 8], nothing], 2-element Ones{Int64})
([nothing, [6, 8]], nothing)  # correct

julia> gradient((xs, y) -> sum(abs2, push!(xs, y)[3]), [[1,2], [3,4]], [5,6])
([nothing, nothing, [10, 12]], 2-element Ones{Int64})
([nothing, nothing], [10, 12])  # correct

(jl_aA0cjy) pkg> st Zygote
      Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_aA0cjy/Project.toml`
  [e88e6eb3] Zygote v0.6.14
```

DhairyaLGandhi:
We had it behave that way in a previous commit in that pr, so I'm thinking if it's a matter of length preservation that you're pointing at?

mcabbott:
Every commit has `Ones{T}`, so no, not true.

mcabbott:
@darsnack  can you explain what you mean by "which missed this extra counting."? 

In the line you link to, your example has `typeof(Δ) = Nothing`. I don't see why the gradient for each `x` being pushed should be `Ones{T}`. Perhaps that was copied from `push!` gradient in FluxML/Zygote.jl#876 , where I am pretty certain it's wrong, in my examples above.

(This was there from the first commit, FWIW. It's easy to make mistakes but this is why we have review before merging.)

DhairyaLGandhi:
Fwiw, I was referring to the length case iirc.

darsnack:
> @darsnack can you explain what you mean by "which missed this extra counting."?

I meant there were no tests for `push!(ps::Params, x)`, and so we missed the error in the resulting gradient. By "extra counting," I specifically meant that `Ones{T}` looks like the incorrect gradient to me. You can see in my examples that a +1 is always accumulated onto the `Params` due to this error.

mcabbott:
Ok, thanks. Somehow I thought by +1 you meant an indexing mistake! But now I agree, the result of your example is wrong because the `Ones()` from there are added to the correct answer. The gradient might always be always wrong, but it's wrong here because it gets `Δ == nothing` and returns a finite answer. Restricting it to just pass no `nothing` would be easy. My WIP attempt to fix the `push!` gradients in general is #1025.

DhairyaLGandhi:
Let's go ahead with the adjoints as we had before, with a few more test cases. I don't think we'll want to worry about binary heaps in the adjoints since those can be written specifically them.

CarloLucibello:
On master
```julia
using Flux, Random


function get_grad1(m, data)
    gradient(Flux.params(m)) do
        loss(m, data)
    end
end

function get_grad2(m, data)
    gradient(Flux.params(m)) do
        ps = Flux.params(m) # just creating params without using them
        loss(m, data)
    end
end

function get_grad3(m, data)
    gradient(Flux.params(m)) do
        ps = Flux.params(m)
        loss(m, data) + sum(sum(p) for p in ps)
    end
end

function get_grad4(m, data)
    ps = Flux.params(m)
    gradient(Flux.params(m)) do
        loss(m, data) + sum(sum(p) for p in ps)
    end
end


function get_grad5(m, data)
    gradient(Flux.params(m)) do
        sum(Flux.params(m)[1]) + sum(Flux.params(m)[2])
    end
end

function get_grad6(m, data)
    ps = Flux.params(m)
    gradient(Flux.params(m)) do
        sum(ps[1]) + sum(ps[2])
    end
end

Random.seed!(17)
m = Dense(3, 2);
data = rand(Float32, 3, 5)
loss(m, x) = sum(m(x).^2)

g1 = get_grad1(m, data) 
g2 = get_grad2(m, data)
g3 = get_grad3(m, data) 
g4 = get_grad4(m, data) 

@show g1[m.weight] # correct
@show g2[m.weight] # == g1 .+ 1  wrong, should be == g1
@show g3[m.weight] # == g1 .+ 2  wrong, should be == g1 .+ 1
@show g4[m.weight] # == g1 .+ 1  correct
@show g5[m.weight] # .== 2       wrong, should be .== 1
@show g6[m.weight] # == nothing  wrong,  should be .== 1
```
This PR:
```julia
@show g1[m.weight] # correct
@show g2[m.weight] # correct
@show g3[m.weight] # correct
@show g4[m.weight] # correct
@show g5[m.weight] # == nothing  wrong, should be .== 1
@show g6[m.weight] # == nothing  wrong,  should be .== 1
```
So it fixes 2 out of 4 failures.

CarloLucibello:
Let's merge this! Need an approval

ToucheSir:
I've been testing with https://github.com/FluxML/Zygote.jl/pull/1025 and so far so good. Do you mind giving that a try?

mcabbott:
Here's what I get on master, with FluxML/Zygote.jl#1025
```
julia> g1[m.weight]
2×3 Matrix{Float32}:
 -8.90053  -8.05664  -4.45624
 -2.61655  -2.25529  -1.30547

julia> g2[m.weight] # should be == g1
2×3 Matrix{Float32}:
 -8.90053  -8.05664  -4.45624
 -2.61655  -2.25529  -1.30547

julia> g3[m.weight] # should be == g1 .+ 1
2×3 Matrix{Float32}:
 -7.90053  -7.05664  -3.45624
 -1.61655  -1.25529  -0.305468

julia> g4[m.weight] # == g1 .+ 1 
2×3 Matrix{Float32}:
 -7.90053  -7.05664  -3.45624
 -1.61655  -1.25529  -0.305468

julia> g5[m.weight] # should be .== 1
2×3 Fill{Float32}, with entries equal to 1.0

julia> g6[m.weight] # should be .== 1

julia> gradient(x -> sum(push!(x,1)), [1,2,3])  # check we've got #1025
ERROR: Mutating arrays is not supported -- called push!(::Vector{Int64}, _...)
```

CarloLucibello:
Right, that PR hasn't been tagged yet.  With FluxML/Zygote.jl#1025  we have
```julia
@show g1[m.weight] # correct
@show g2[m.weight] # correct
@show g3[m.weight] # correct
@show g4[m.weight] # correct
@show g5[m.weight] # correct
@show g6[m.weight] # == nothing  wrong,  should be .== 1
```
so one more issue addressed! I'll close this then, and file a PR later containing only the tests.

ToucheSir:
Just so we're on the same page, this would a Zygote PR? I don't think there's anything that needs Flux specifically. For example, I was testing with this function:
```julia
# Emulates Flux's params(!)
function params(xs...)
    tmp = Params()
    for x in xs
         push!(tmp, x)
     end
     tmp
end
```

Issue #1601, opened 2021-05-20
`pullback`'s `back` returns unexpected size if some parameters are not used


garibarba:
This behavior can lead to a `DimensionMismatch` error in `DiffEqSensitivity`. I don't know if this is intended behaviour, but I feel like there should be a warning.

Minimal example:
```julia
using Flux, Zygote

struct TwoDenses
    dense::Dense
    dense2::Dense
end

Flux.@functor TwoDenses

function (m::TwoDenses)(x)
    out = m.dense(x)
end

model = TwoDenses(
    Dense(3,1),
    Dense(3,2)
)
p, re = Flux.destructure(model)

x = [1., 2., 3.]
y, back = Zygote.pullback((x, p) -> re(p)(x), x, p)

dy = [4.]
dx, dp = back(dy)
@show length(p), length(dp)
```

Returns:
```julia
(length(p), length(dp)) = (12, 4)
```

Edit:
To be clear, I would have expected:
```julia
(length(p), length(dp)) = (12, 12)
```

ToucheSir:
`_restructure` is [not exactly picky](https://github.com/FluxML/Flux.jl/blob/e6629dd98f2b28c5ed42d4031ca8bd7d41dadc31/src/utils.jl#L615) about how large the input vector is as long as it's large enough. I assume there wouldn't be any harm in adding an assert or other error check in there, will put in a PR next week if someone doesn't get to it first.

Issue #1408, opened 2020-11-27
No sanity checks on `destructure` and `loadparams!`


mcabbott:
Given too many parameters, or parameters of the wrong shapes, `destructure` and `loadparams!` silently have a go. I believe it would be safer to make these errors. Or at least warnings:
```julia
using Flux

m1 = Chain(Dense(2,2), Dense(2,2))
m1[1].W .= 1
v1, re1 = Flux.destructure(m1)
m2 = re1(rand(100)) # doesn't check length
m2[1].W # new values

m3 = Chain(Dense(2,2), Dense(2,2), Dense(2,2))
Flux.loadparams!(m1, params(m3)) # doesn't check length
m1[1].W # has been overwritten

m4 = Chain(Dense(2,2), Dense(2,3), Dense(3,2))
Flux.loadparams!(m1, params(m3)) # doesn't check shape either

Flux.loadparams!(m4, params(m1)) # ERROR: Expected param size (3, 2), got (2, 2)
```

mcabbott:
> However, what if you wanted to `loadparams!` from a 3-layer MLP onto a 2-layer one or vice versa? Assuming the shape of the first two layers are the same, this should be supported in Flux. That is what `strict=False` in `load_state_dict` gives you: the ability to partially load params into a network without having to slice it up and reconstitute it afterwards (if indeed that's even possible in e.g. the presence of custom container layers).
> 
> _Originally posted by @ToucheSir in https://github.com/FluxML/Flux.jl/issues/1402#issuecomment-734959072_

The pytorch function is described for instance here (maybe note the best link?) https://pytorch.org/docs/master/generated/torch.nn.Module.html?highlight=load_state_dict#torch.nn.Module.load_state_dict

Perhaps we should follow that and make a keyword `strict=true` by default, which rejects all the size mismatches above. But what exactly is the rule for `strict=false`?

darsnack:
If we want to address the `strict = false` case in a forward-thinking manner, then I think `params` needs to evolve past the "bag of arrays" approach to store structural information à la functors. This alone would make sanity checking better (by having a more narrowly scoped error case that `length(params(m)) != length(ps)`), and it would make educated-guessing for the `strict = false` case more reliable.

mcabbott:
[...]

PR #1616, opened 2021-06-14
Warn on reconstruct length mismatch

ToucheSir:
Ref. #1601. This is kept as a plain warning for backwards compat, but perhaps we want to consider it a bugfix and error/depwarn instead?

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -610,12 +610,15 @@ end
 
 function _restructure(m, xs)
   i = 0
-  fmap(m) do x
+  m̄ = fmap(m) do x
     x isa AbstractArray || return x
     x = reshape(xs[i.+(1:length(x))], size(x))
     i += length(x)
     return x
   end
+
+  length(xs) == i || @warn "Expected $(i) params, got $(length(xs))"
+  return m̄
 end
 
 @adjoint function _restructure(m, xs)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -616,13 +616,18 @@ function _restructure(m, xs)
     i += length(x)
     return x
   end
-
   length(xs) == i || @warn "Expected $(i) params, got $(length(xs))"
   return m̄
 end
 
 @adjoint function _restructure(m, xs)
-  _restructure(m, xs), dm -> (nothing,destructure(dm)[1])
+  m̄, numel = _restructure(m, xs), length(xs)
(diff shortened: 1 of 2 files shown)


mcabbott:
xref #1408. It might be worth fixing `loadparams!` at the same time?

But also, it would be nice if the example in #1601 could also give you an error (or warning). You will already get a BoundsError if you try to use the too-short gradient in `re`, but could the `gradient` call itself complain that this is producing the wrong length?

ToucheSir:
I purposely chose not to wade into `loadparams!` because that entails a minefield of `Zeros`, parameter representation and more. This seemed like much lower hanging fruit :sweat_smile:.

> could the `gradient` call itself complain that this is producing the wrong length?

Do you mean verifying the length in the pullback function? That makes sense to me, would there be any performance impact to making the pullback a closure over the stored length of `xs`?

mcabbott:
> chose not to wade into loadparams!

Fair enough! Steps in the right direction are good.

> verifying the length in the pullback function

Yes, something like that. I haven't looked closely at quite what path that gradient takes. (Or if I did, I have forgotten again.) Nor exactly what closes over what.

ToucheSir:
Tried adding a check in the pullback as well. If there is any overhead, it's probably negligible.

```julia
function f(m, x)
  p, re = Flux.destructure(m)
  ∇m = gradient(m -> sum(m(x)), m)[1]
  gradient(θ -> sum(re(θ)(x)), p)[1]
end

m, x = Dense(3, 1), rand(Float32, 3, 1);
#  Master: 49.911 μs (367 allocations: 19.08 KiB)
# This PR: 52.329 μs (368 allocations: 19.20 KiB)
@btime f($m, $x);

m2, x2 = Chain(Dense(256, 32), Dense(32, 4)), rand(Float32, 256, 256);
#  Master: 406.237 μs (769 allocations: 933.28 KiB)
# This PR: 404.844 μs (773 allocations: 933.38 KiB)
@btime f($m2, $x2);
```
The example in #1601 now warns:
```juliarepl
...
julia> dx, dp = back(dy)
┌ Warning: Expected 12 params, got 4
└ @ Flux ~/.julia/dev/Flux.jl/src/utils.jl:627
```

CarloLucibello (approved):

Issue #1372, opened 2020-10-26
Models with dropout effect GLOBAL_RNG differently when run on GPU


ablaom:
I have been comparing some Flux models on the CPU and GPU and was surprised to find they were not giving similar predictions. It took me quite some time to nail this down to the problem with dropout given in the comment title and demonstrated below. Maybe I'm missing something, but it seems to me this is serious impediment to reproducibility using Flux. 

```julia
using Flux
import Random.seed!
seed!(123);

data = [(rand(Float32, 5, 1), rand(Float32, 5)), ]

model = Flux.Chain(Flux.Dense(5, 2, identity),
                  Flux.Dropout(0.5),
                  Flux.Dense(2, 1))

data_gpu = gpu(data);
model_gpu = gpu(model)

loss(x, y)     = Flux.mse(model(x),     y)
loss_gpu(x, y) = Flux.mse(model_gpu(x), y)

optimiser = Flux.ADAM()

# cpu training:
seed!(123);
Flux.train!(loss, Flux.params(model), data, optimiser)
rand() # 0.6739586945680673


# gpu training: 
seed!(123);
Flux.train!(loss_gpu, Flux.params(model_gpu), data_gpu, optimiser)
rand() # 0.13672511011651545  <----------------------- should be the same as a `rand()` above
```

If one removes `Dropout` from the chain, the `rand()` calls return the same value, as expected. 

```julia
julia> versioninfo()
Julia Version 1.5.2
Commit 539f3ce943 (2020-09-23 23:17 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel Xeon Processor (Skylake, IBRS)
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake-avx512)

(jl_4sVBIY) pkg> st
Status `/tmp/jl_4sVBIY/Project.toml`
[587475ba] Flux v0.11.1
```

DhairyaLGandhi:
There was a similar concern brought up earlier.
cc @maleadt who had given the explanation for it then iirc

maleadt:
I don't recall an issue with the CPU's global RNG behaving differently depending on whether the model is on the GPU or not.
@ablaom you could try redefining some inner method of e.g. MersenneTwister to have it trace where randomness is requested.

CarloLucibello:
you can use CUDA.seed! to get reproducible sequences on gpu

PR #1618, opened 2021-06-15
Add RNG to Dropout

### LINKED ISSUE #1617: Allow specification of RNG in Dropout

DhairyaLGandhi:
Closes #1617

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -87,7 +87,7 @@ end
 
 function (a::Dropout)(x)
   _isactive(a) || return x
-  return dropout(x, a.p; dims = a.dims, active = true)
+  return dropout(a.rng, x, a.p; dims = a.dims, active = true)
 end
 
 testmode!(m::Dropout, mode = true) =
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -1,3 +1,5 @@
+using Random: GLOBAL_RNG
+
 istraining() = false
 
 @adjoint istraining() = true, _ -> nothing
@@ -10,7 +12,7 @@ _dropout_shape(s, dims) = tuple((i ∉ dims ? 1 : si for (i, si) ∈ enumerate(s
 _dropout_kernel(y::T, p, q) where {T} = y > p ? T(1 / q) : T(0)
 
 """
-    dropout(x, p; dims=:, active=true)
+    dropout([rng = GLOBAL_RNG], x, p; dims=:, active=true)
 
 The dropout function. If `active` is `true`,
 for each input, either sets that input to `0` (with probability
@@ -28,15 +30,19 @@ automatically managed using the [`Dropout`](@ref) layer instead of the
 
 The [`Dropout`](@ref) layer is what you should use in most scenarios.
 """
-function dropout(rng::AbstractRNG, x, p; dims=:, active::Bool=true)
+function dropout(rng::AbstractRNG, x, p; dims = :, active::Bool = true)
   active || return x
   y = dropout_mask(rng, x, p, dims=dims)
   return x .* y
 end
 
-@adjoint function dropout(rng, x, p; dims=:, active::Bool=true)
+function dropout(x, p; dims = :, active::Bool = true)
+  dropout(GLOBAL_RNG, x, p, dims = dims, active = active)
+end
+
+@adjoint function dropout(rng, x, p; dims = :, active::Bool = true)
   active || return x, Δ -> (Δ, nothing)
-  y = dropout_mask(rng, x, p, dims=dims)
+  y = dropout_mask(rng, x, p, dims = dims)
   return x .* y, Δ -> (nothing, Δ .* y, nothing)
 end
 
@@ -66,7 +72,7 @@ end
 
 function Dropout(p; dims=:)
   @assert 0 ≤ p ≤ 1
-  Dropout(Random.GLOBAL_RNG, p; dims)
+  Dropout(GLOBAL_RNG, p; dims)
 end
 
 function Dropout(rng, p; dims = :)
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -47,7 +47,7 @@ function dropout(x::CUDA.CuArray, p; dims = :, active::Bool = true)
 end
 
 @adjoint function dropout(rng, x, p; dims = :, active::Bool = true)
-  active || return x, Δ -> (Δ, nothing)
+  active || return x, Δ -> (nothing, Δ, nothing)
   y = dropout_mask(rng, x, p, dims = dims)
   return x .* y, Δ -> (nothing, Δ .* y, nothing)
 end
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -40,6 +40,12 @@ function dropout(x, p; dims = :, active::Bool = true)
   dropout(default_rng(), x, p, dims = dims, active = active)
 end
 
+# CUDA currently needs a manual dispatch to avoid
+# calling a non-GPU RNG with a CuArray
+function dropout(x::CUDA.CuArray, p; dims = :, active::Bool = true)
+  dropout(CUDA.CURAND.default_rng(), x, p, dims = dims, active = active)
+end
+
 @adjoint function dropout(rng, x, p; dims = :, active::Bool = true)
   active || return x, Δ -> (Δ, nothing)
   y = dropout_mask(rng, x, p, dims = dims)
@@ -70,8 +76,7 @@ mutable struct Dropout{F,D}
   active::Union{Bool, Nothing}
 end
 
-function Dropout(p; dims=:)
-  @assert 0 ≤ p ≤ 1
+function Dropout(p; dims = :)
   Dropout(default_rng(), p; dims)
 end
 
@@ -82,7 +87,7 @@ end
 
 function (a::Dropout)(x)
   _isactive(a) || return x
-  return dropout(a.rng, x, a.p; dims=a.dims, active=true)
+  return dropout(x, a.p; dims = a.dims, active = true)
 end
 
 testmode!(m::Dropout, mode = true) =
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -1,4 +1,4 @@
-using Random: GLOBAL_RNG
+using Random: default_rng, GLOBAL_RNG
 
 istraining() = false
 
@@ -12,7 +12,7 @@ _dropout_shape(s, dims) = tuple((i ∉ dims ? 1 : si for (i, si) ∈ enumerate(s
 _dropout_kernel(y::T, p, q) where {T} = y > p ? T(1 / q) : T(0)
 
 """
-    dropout([rng = GLOBAL_RNG], x, p; dims=:, active=true)
+    dropout([rng = default_rng()], x, p; dims=:, active=true)
 
 The dropout function. If `active` is `true`,
 for each input, either sets that input to `0` (with probability
@@ -37,7 +37,7 @@ function dropout(rng::AbstractRNG, x, p; dims = :, active::Bool = true)
 end
 
 function dropout(x, p; dims = :, active::Bool = true)
-  dropout(GLOBAL_RNG, x, p, dims = dims, active = active)
+  dropout(default_rng(), x, p, dims = dims, active = active)
 end
 
 @adjoint function dropout(rng, x, p; dims = :, active::Bool = true)
@@ -53,7 +53,7 @@ function dropout_mask(rng::AbstractRNG, x, p; dims=:)
 end
 
 """
-    Dropout([rng = GLOBAL_RNG], p; dims=:)
+    Dropout([rng = default_rng()], p; dims=:)
 
 Dropout layer. In the forward pass, apply the [`Flux.dropout`](@ref) function on the input.
 
@@ -72,7 +72,7 @@ end
 
 function Dropout(p; dims=:)
   @assert 0 ≤ p ≤ 1
-  Dropout(GLOBAL_RNG, p; dims)
+  Dropout(default_rng(), p; dims)
 end
 
 function Dropout(rng, p; dims = :)
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -28,26 +28,26 @@ automatically managed using the [`Dropout`](@ref) layer instead of the
 
 The [`Dropout`](@ref) layer is what you should use in most scenarios.
 """
-function dropout(x, p; dims=:, active::Bool=true)
+function dropout(rng::AbstractRNG, x, p; dims=:, active::Bool=true)
   active || return x
-  y = dropout_mask(x, p, dims=dims)
+  y = dropout_mask(rng, x, p, dims=dims)
   return x .* y
 end
 
-@adjoint function dropout(x, p; dims=:, active::Bool=true)
+@adjoint function dropout(rng, x, p; dims=:, active::Bool=true)
   active || return x, Δ -> (Δ, nothing)
-  y = dropout_mask(x, p, dims=dims)
-  return x .* y, Δ -> (Δ .* y, nothing)
+  y = dropout_mask(rng, x, p, dims=dims)
+  return x .* y, Δ -> (nothing, Δ .* y, nothing)
 end
 
-function dropout_mask(x, p; dims=:)
-  y = rand!(similar(x, _dropout_shape(x, dims)))
+function dropout_mask(rng::AbstractRNG, x, p; dims=:)
+  y = rand!(rng, similar(x, _dropout_shape(x, dims)))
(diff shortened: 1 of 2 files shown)


src/layers/normalise.jl
```diff
@@ -28,26 +30,30 @@ automatically managed using the [`Dropout`](@ref) layer instead of the
...
-  y = dropout_mask(x, p, dims=dims)
-  return x .* y, Δ -> (Δ .* y, nothing)
+  y = dropout_mask(rng, x, p, dims = dims)
+  return x .* y, Δ -> (nothing, Δ .* y, nothing)
 end
 
-function dropout_mask(x, p; dims=:)
-  y = rand!(similar(x, _dropout_shape(x, dims)))
+function dropout_mask(rng::AbstractRNG, x, p; dims=:)
+  y = rand!(rng, similar(x, _dropout_shape(x, dims)))
```
darsnack, line 50: I don't think this dispatches to the GPU intrinsic when `rng` is specified. @maleadt?

maleadt: We only override `default_rng`, I haven't looked into getting GLOBAL_RNG to work (maybe it does).

DhairyaLGandhi: We can switch it for `default_rng` too of course

CarloLucibello:
LGTM, just needs tests and addressing the gpu case. 
A slightly different interface could be given by using a keyword argument for the layer constructor, `Dropout(p; rng=GLOBAL_RNG)`.

ablaom:
See also https://github.com/FluxML/Flux.jl/issues/1372

DhairyaLGandhi:
Switched out `GLOBAL_RNG` for `default_rng`, @maleadt  do we need to do something differently?

src/layers/normalise.jl
```diff
@@ -28,26 +30,30 @@ automatically managed using the [`Dropout`](@ref) layer instead of the
...
   y .= _dropout_kernel.(y, p, 1 - p)
   return y
 end
 
 """
-    Dropout(p; dims=:)
+    Dropout([rng = default_rng()], p; dims=:)
```
CarloLucibello, line 56: `DataLoader` takes `rng` as a keyword argument
https://github.com/FluxML/Flux.jl/blob/335286adf118b61ad6fffa5937bd9358477a00c9/src/data/dataloader.jl#L70
we may want to do the same here for consistency.
For the functional form `dropout` instead is ok to take `rng` as the first positional argument

src/layers/normalise.jl
```diff
@@ -1,3 +1,5 @@
+using Random: default_rng, GLOBAL_RNG
```
CarloLucibello, line 1:
```suggestion
using Random: default_rng
```

maleadt:
>  do we need to do something differently for it to pick up `CUDA.CURAND.default_rng()`

No, we have device overrides now:

https://github.com/JuliaGPU/CUDA.jl/blob/ee70b71b620edad627f7dc8aa7e3e385a63f8bb8/src/device/random.jl#L95

src/layers/normalise.jl
DhairyaLGandhi: I don't think the constructor was duplicated without reason. Flux hasn't really been kwarg forward where it doesn't add value.

src/layers/normalise.jl
maleadt: But you have to call it on the GPU. You can't call `default_rng` in the Dropout ctor and forward that RNG object to the GPU (unless Dropout objects are created in device code?).

DhairyaLGandhi: I guess it depends on who we give the responsibility to provide a valid rng to (the user or flux) - a proverbial `gpu(rng)`

maleadt: But is this code actually being executed directly on the GPU (i.e. in a kernel)? If not, device overrides and `default_rng` doesn't matter here, and you do actually need to provide a GPU RNG for use on the host, e.g., `CUDA.RNG()` or `CUDA.CURAND.RNG()`. "write your kernel once and run anywhere" doesn't really work if it isn't a kernel.

src/layers/normalise.jl
darsnack: In general, I think kwargs at clarity over positional arguments precisely because their position is irrelevant. In general, you can assume the user is not going to remember the placement of arguments > 3. That being said, having the RNG first is a standard pattern throughout Julia.

I could see `dropout([rng], ...)` following this pattern, but it seems really odd for `Dropout`. For example, in Distributions, you don't construct `Normal([rng], ...)`. I don't think I've ever seen this pattern used in a non-functional form (i.e. constructors). I would also suggest using a keyword argument for the constructor.

src/layers/normalise.jl
darsnack: Sorry, but I'm confused. If I call `rand!(x)` and `x isa CuArray`, then I'll dispatch to the GPU override. If I call `rand!(rng, x)` and `rng` is one of the many RNGs available in Julia, then this obviously cannot run on the GPU, but I would consider that a user error. But, what should the default value of `rng` be such that when the user doesn't customize it, it will call `Random.rand!` for `Array`s and `CUDA.rand!` for `CuArray`s? Or do we need to set the default conditionally based on the type of `x`?

darsnack: Alternatively to setting the default conditionally, we can make `gpu(::Dropout)` change `GLOBAL_RNG` to `CUDA.RNG()` (maybe this is what you meant @DhairyaLGandhi).

src/layers/normalise.jl
DhairyaLGandhi: But distributions also doesn't have an rng as part of the constructor, so I don't think it's comparable. Most users would not need to worry about this constructor. Anyway, it doesn't have too many args, and the same argument can be made for kwargs (what kwargs can I supply to this function?), so it is fair to say this is equivalent in those regards.

src/layers/normalise.jl
maleadt: > But, what should the default value of `rng` be such that when the user doesn't customize it, it will call `Random.rand!` for `Array`s and `CUDA.rand!` for `CuArray`s? Or do we need to set the default conditionally based on the type of `x`?

There currently isn't an API to figure out a good RNG for an array type, so this would need some additional methods. `default_rng` isn't such an API because it doesn't take arguments; I had presumed this code is executing in a GPU kernel where we use contextual dispatch (think Cassette) to replace the method and return an appropriate RNG. But that's not relevant here, since this is host code using GPU functionality instead of code directly executing on the GPU.

DhairyaLGandhi: I have added a dispatch for CUDA. If we are executing a forward pass, I had assumed the whole call graph can get compiled at once doing the method swaps as necessary, but maybe that's not the case - as Tim mentioned it explicitly needs to be part of a CUDA kernel call.

darsnack: I am not sure this is the approach I would go for. It would be weird if I created `Dropout` with `MersenneTwister` and the program silently used `CUDA.CURAND.default_rng()`. Instead, I would use the approach you initially suggested so that `gpu(m)` swaps `GLOBAL_RNG`/`Random.default_rng()` to `CUDA.CURAND.default_rng()` (and errors when `rng` is set to some other variant that cannot be run on the GPU).

My only concern is that we [`fmap(CUDA.cu, ...)`](https://github.com/FluxML/Flux.jl/blob/335286adf118b61ad6fffa5937bd9358477a00c9/src/functor.jl#L72) and defining `CUDA.cu(GLOBAL_RNG)` seems like a big no-no. Maybe we can do the following:
```julia
_isleaf(::AbstractArray{<:Number}) = true
_isleaf(::AbstractArray{T}) where T = isbitstype(T)
_isleaf(::AbstractRNG) = true
_isleaf(x) = false

to_gpu(x) = CUDA.cu(x)
to_gpu(::GLOBAL_RNG) = CUDA.CURAND.default_rng()
to_gpu(rng::AbstractRNG) = error("Cannot move $rng to GPU (only `GLOBAL_RNG` is supported).")

gpu(x) = use_cuda[] ? fmap(to_gpu, x; exclude = _isleaf) : x
```

src/layers/normalise.jl
```diff
@@ -60,20 +72,25 @@ Does nothing to the input once [`Flux.testmode!`](@ref) is `true`.
...
+function Dropout(rng, p; dims = :)
   @assert 0 ≤ p ≤ 1
-  Dropout(p, dims, nothing)
+  Dropout(p, dims, rng, nothing)
 end
 
 function (a::Dropout)(x)
   _isactive(a) || return x
-  return dropout(x, a.p; dims=a.dims, active=true)
+  return dropout(x, a.p; dims = a.dims, active = true)
```
darsnack, line 90: This ignores the RNG right now

DhairyaLGandhi: Right - I did it to check out what we can do to ensure the kernel can actually run with a correct rng - I don't intend to ignore the rng here when merging.

DhairyaLGandhi: This is more intended for having concrete code to weigh out the different approaches

src/layers/normalise.jl
DhairyaLGandhi: See https://github.com/FluxML/Flux.jl/pull/1618#discussion_r652780581

This approach also ignores a gou rng, so I'm not in favour. Either way, I don't think `to_gpu` is a good api, and the most common case of `Dropout |> gpu` with a cuda compatible rng wouldn't be handled this way

darsnack: I don't really follow the relevance of the linked comment. AFAIK there are no GPU compatible RNGs. There are simply the RNGs provided by CUDA.jl/GPUArrays.jl. `to_gpu` is similar to `trainable` — it allows us to define how to map leaf in the structure to the GPU without defining `CUDA.cu` on types that it shouldn't be defined on. It's out of necessity, but if you can think of an alternative, go for it (I'm not tied to `to_gpu` at all).

src/layers/normalise.jl
CarloLucibello: even assuming that they are equivalent (which I'm not convinced of), can we avoid moving randomly in this equivalent options space and keep doing what we do with DataLoader without introducing inconsistencies?

src/layers/normalise.jl
DhairyaLGandhi: I was referring to the exactly the ones provided by CUDA/CURAND. Sorry if that wasn't clear.

darsnack: So I don't understand your comment then. This provides a mapping from `GLOBAL_RNG` to `CUDA.CURAND.default_rng()`. And when a user does `Dropout |> gpu` with any other CPU RNG, it throws an error instead of silently ignoring it.

darsnack: I think the most sensible place to error is when the user tries `m |> gpu` on a model that contains `Dropout` with CPU RNGs that have no GPU equivalent. Either way, I don't care how we do it, but as long as we don't silently use a CUDA RNG when `x isa CuArray` and throw an error instead, then it's fine by me.

DhairyaLGandhi: Right - it will error anyway, its more clear when we can say that CUDA needs a way to dispatch to a CUDA RNG. I agree on that.

src/layers/normalise.jl
DhairyaLGandhi: I would err on what makes sense for the layer in this case. I don’t think we need to enforce the same on DataLoader where its more or less a detail. Its already odd that DataLoader takes both a shuffle and a rng - what exactly is the rng supposed to do when there isn’t shuffling? Here it is generally actionable, so it makes sense for Dropout to have the RNG as a positional.

CarloLucibello: I really don't understand your argument. Can we have some other opinions here? @ToucheSir @mcabbott

DhairyaLGandhi: The argument is simple - functions with arguments that are critical for them to function typically have them as positional. An rng is always used in the Dropout layer whether implicitly or explicitly, and it directly affects it's functionality, so it's already a positional. It is also typical for rng to be first argument `rand(rng, ...)` etc.

A data loader is fundamentally very different. I would argue an rng isn't central to its purpose. We can turn off shuffling in data loaders and not affect its use at all. Even if not, it is trivial to shuffle the minibatch after you have a minibatch in hand.

Also by extending the same logic of kwargs being inherently better, it would be more common to have kwarg only methods Everywhere (esp in Base, but also elsewhere).

ToucheSir: I'm very much in the camp of RNG as the first positional arg, but to build on Kyle's point this is an odd one. If we were to draw a comparison to distributions, why is the RNG stored in the layer at all? If we look at e.g. https://github.com/google/flax/blob/master/flax/linen/stochastic.py#L42, an alternative would be to pass in an RNG when calling the layer. For ambient overrides, my feeling is that it would be better to have a context manager that temporarily overrides the default RNG for a given scope. I would also note that with this PR, `Dropout` will be the _only_ layer with mutable non-array parameter state, and state that changes with the forward pass at that.

DhairyaLGandhi: That's what I was gravitating towards also, where we make the decision Might be in rewriting the forwards pass a little bit.

darsnack: > an alternative would be to pass in an RNG when calling the layer

Right, that's exactly why I made the distinction between the functional form and the layer. `dropout(rng, ...)` makes perfect sense to me or `(::Dropout)(rng, ...)` because it is being passed in on the call. I haven't seen the same pattern used in constructors. Flux doesn't support making calls like `(::Dropout)(rng, ...)` work within a larger model, so I erred on the side of kwarg for the constructor to limit confusion.

ToucheSir: I'm just struggling to think of many scenarios where one has enough access to model construction to thread through an RNG to a subset of layers, but not enough to define a custom forward pass that does the same.  From my understanding of the context behind https://github.com/FluxML/Flux.jl/issues/1617, something like this would be sufficient:
```julia
m = ... # has Dropout

seed!(...)
m(x)

# or 
with_seed(...) do
  m(x)
end
```
i.e. the RNG can apply to the entire model.

DhairyaLGandhi: Right, we wouldn't want that. More generally, not everything inside that do block may run on the gpu, and it's more annoying to write out one part of the problem with this block and others not just for rngs. It would make sense for off loading everything to the GPU for example. I don't think we have the seed api for CUDA? Even then that's only relevant for global effects, it may be desirable to have different rngs for Dropout vs (say) initialisation. Personally I would prefer it if we could handle it outside too.

darsnack: > one has enough access to model construction to thread through an RNG to a subset of layers

There are limits to this, but if all someone wanted is to pass a specific RNG to certain types of layers, then that can be done with `fmap` and the `exclude` keyword specified appropriately.

Personally, I prefer the solution where layer-specific arguments are possible. It would match how RNGs are passed into most Julia functions, and conceptually, the RNG/RNG state is very much an "input" to the model. I am struggling to see how we make that happen though. How does Flax handle it? Do you have to write a custom forward pass to utilize a non-default RNG?

ToucheSir: Wait, so you _are_ advocating for storing the RNG in the layer struct? My intent was to figure out what scenarios would require doing so, and thus far it doesn't seem like there are many. Using different RNGs for the forward pass vs initialization isn't one of them IMO, because those don't happen at the same time (and thus you can seed/scope the RNG for one and not the other).

darsnack: No, I also prefer not to store the RNG in the struct, but I don't see an alternative. If seeding is the only goal, then this PR is not really needed. If different RNG types are required, then I feel that you need scoping, but that can get tricky with different devices like Dhairya mentioned. Only option left seems like the ability to pass through layer-specific arguments on the forward pass. IMO this would be a nice addition if we figured out a good way to do it. It can be used for more than RNGs.

Re `fmap`: that was basically me pointing out that if we are forced to store the RNG in the struct, swapping it out doesn't require fine-grained access to the model building.

ToucheSir: What I was trying to get at (and probably muddled, apologies) is that putting the RNG in the struct isn't necessary to fix https://github.com/FluxML/Flux.jl/issues/1372. If we instead provided, say, a top-level function to seed all RNGs like PyTorch does, then all this discussion about device compatibility and conversion would be moot. This is to say nothing of edge cases that haven't been discussed: I can already foresee saving/loading behaviour and how it interacts with RNG "tying" across different layers being a headache.

TL;DR adding an RNG struct parameter feels like YAGNI in context and isn't worth the implementation complexity/headaches it'll bring IMO. If we really want to look into this, a more principled investigation (e.g. can we co-opt something like JAX's RNG key design) would be prudent.

ablaom:
@DhairyaLGandhi Right, so what is the status of #1617, then?

darsnack:
See #1849 now.

PR #1632, opened 2021-06-26
Add WIP docstrings to CPU and GPU

logankilpatrick:
We could surely use more details, but ideally we would want to document these functions and what they do + usage.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -97,7 +97,12 @@ _isbitsarray(x) = false
 """
     gpu(x)
 
-Moves a particular resource from where it is, onto the GPU, if available. It is a no-op otherwise.
+Moves `m` to the current GPU device, if available. It is a no-op otherwise.
+See the [CUDA.jl docs](https://juliagpu.github.io/CUDA.jl/stable/usage/multigpu/) 
+to help identify the current device.
+
+This works for functions and 
+any struct with [`@functor`](@ref) defined.
 
 ```julia-repl
 julia> m = Dense(1,2)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -63,12 +63,20 @@ function loadparams!(m, xs)
 end
 
 # CPU/GPU movement conveniences
-
+"""
+    cpu(m)
+Moves a paticular resource from where it is, onto the CPU.
+"""
 cpu(m) = fmap(x -> adapt(Array, x), m)
 
 _isbitsarray(::AbstractArray{<:Number}) = true
 _isbitsarray(::AbstractArray{T}) where T = isbitstype(T)
 _isbitsarray(x) = false
+
+"""
+    gpu(x)
+Moves a paticular resource from where it is, onto the GPU.
+"""
 gpu(x) = use_cuda[] ? fmap(CUDA.cu, x; exclude = _isbitsarray) : x
 
 # Precision
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -65,6 +65,7 @@ end
 # CPU/GPU movement conveniences
 """
     cpu(m)
+
 Moves a paticular resource from where it is, onto the CPU.
 """
 cpu(m) = fmap(x -> adapt(Array, x), m)
@@ -75,6 +76,7 @@ _isbitsarray(x) = false
 
 """
     gpu(x)
+
 Moves a particular resource from where it is, onto the GPU, if available. It is a no-op otherwise.
 """
 gpu(x) = use_cuda[] ? fmap(CUDA.cu, x; exclude = _isbitsarray) : x
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -68,8 +68,7 @@ end
 
 Moves `m` onto the CPU.
 
-This works for functions and 
-any struct with [`@functor`](@ref) defined.
+This utility uses [`@functor`](@ref) to properly move structures to the CPU.
 
 ```julia-repl
 julia> m = Dense(1,2)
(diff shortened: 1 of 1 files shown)


src/functor.jl
```diff
@@ -63,12 +63,20 @@ function loadparams!(m, xs)
...
+"""
 cpu(m) = fmap(x -> adapt(Array, x), m)
 
 _isbitsarray(::AbstractArray{<:Number}) = true
 _isbitsarray(::AbstractArray{T}) where T = isbitstype(T)
 _isbitsarray(x) = false
+
+"""
+    gpu(x)
+Moves a paticular resource from where it is, onto the GPU.
```
DhairyaLGandhi, line 78:
```suggestion
Moves a particular resource from where it is, onto the GPU, if available. It is a no-op otherwise.
```

DhairyaLGandhi (commented):
Lgtm

src/functor.jl
```diff
@@ -63,12 +63,20 @@ function loadparams!(m, xs)
...
+"""
 cpu(m) = fmap(x -> adapt(Array, x), m)
 
 _isbitsarray(::AbstractArray{<:Number}) = true
 _isbitsarray(::AbstractArray{T}) where T = isbitstype(T)
 _isbitsarray(x) = false
+
+"""
+    gpu(x)
+Moves a particular resource from where it is, onto the GPU, if available. It is a no-op otherwise.
```
DhairyaLGandhi, line 78:
```suggestion

Moves a particular resource from where it is, onto the GPU, if available. It is a no-op otherwise.
```

src/functor.jl
```diff
@@ -63,12 +63,20 @@ function loadparams!(m, xs)
 end
 
 # CPU/GPU movement conveniences
-
+"""
+    cpu(m)
+Moves a paticular resource from where it is, onto the CPU.
```
DhairyaLGandhi, line 68:
```suggestion

Moves a paticular resource from where it is, onto the CPU.
```

DhairyaLGandhi (approved):

DhairyaLGandhi:
bors r-

Just wanted to get an opinion on whether to include details about needing a CUDA compatible GPU. Also, good to add an example for each, ideally with a simple model.

ToucheSir:
Given https://github.com/FluxML/Flux.jl/pull/1566, I don't think it's necessary to include details about GPU compat for CUDA specifically.

logankilpatrick:
Would a simple example like: 
```Julia
julia> m = Dense(10, 5)
Dense(10, 5)

julia> cpu(m)
Dense(10, 5)
```
work?

DhairyaLGandhi:
That'd work, i would also show that the types of arrays have changed to a `CuArray` (for `gpu`)

logankilpatrick:
I don't have an Nvidia GPU, can you add that example?

DhairyaLGandhi:
Done.

src/functor.jl
```diff
@@ -63,12 +63,53 @@ function loadparams!(m, xs)
 end
 
 # CPU/GPU movement conveniences
+"""
+    cpu(m)
+
+Moves a particular resource from where it is, onto the CPU.
```
darsnack, line 69:
```suggestion
Moves `m` onto the CPU.

This works for functions and 
any struct with [`@functor`](@ref) defined.
```

src/functor.jl
```diff
@@ -63,12 +63,53 @@ function loadparams!(m, xs)
...
 cpu(m) = fmap(x -> adapt(Array, x), m)
 
 _isbitsarray(::AbstractArray{<:Number}) = true
 _isbitsarray(::AbstractArray{T}) where T = isbitstype(T)
 _isbitsarray(x) = false
+
+"""
+    gpu(x)
+
+Moves a particular resource from where it is, onto the GPU, if available. It is a no-op otherwise.
```
darsnack, line 97:
```suggestion
Moves `m` to the current GPU device, if available. It is a no-op otherwise.

This works for functions and 
any struct with [`@functor`](@ref) defined.
```

darsnack (changes requested):
Just some suggestions to make it clearer and a forward link to `@functor` so people aren't confused when they custom layer won't transfer.

src/functor.jl
DhairyaLGandhi: This isn't clear - what does current device want to imply? I don't think users would understand this clearly without knowing the whole `CUDA.device` story.

src/functor.jl
DhairyaLGandhi: What about `gpu((rand(...), zeros(...))`? That doesn't have functor defined or anything. May be better to say "it uses the `@functor` utility decide whether something may be moved or not"

DhairyaLGandhi: I think aliasing "Moves `x` onto the CPU memory" is the most directly helpful form of the docstring. The rest may just go to the proper docs.

src/functor.jl
darsnack:
```suggestion
Moves `m` to the current GPU device, if available. It is a no-op otherwise.
See the [CUDA.jl docs](https://juliagpu.github.io/CUDA.jl/stable/usage/multigpu/) to help identify the current device.

This works for functions and 
any struct with [`@functor`](@ref) defined.
```

src/functor.jl
darsnack: Yeah I would say what you wrote ("This utility uses [`@functor`](@ref) to properly move structures to the CPU").

I still think it is worth putting here. Good documentation repeats important information in multiple places. Flux is severely lacking in this regard.

logankilpatrick:
I tried to address the comments, please let me know if more needs to be updated!

ToucheSir (approved):

darsnack (approved):

PR #1633, opened 2021-06-26
Add section on Checking GPU Availability

logankilpatrick:
Follow up from this question I posted: https://stackoverflow.com/questions/68142837/how-to-check-if-flux-jl-is-using-the-gpu/68144877#68144877 which was inspired by the top PyTorch question on SO: https://stackoverflow.com/questions/48152674/how-to-check-if-pytorch-is-using-the-gpu

### DIFF
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -2,6 +2,17 @@
 
 NVIDIA GPU support should work out of the box on systems with CUDA and CUDNN installed. For more details see the [CUDA](https://github.com/JuliaGPU/CUDA.jl) readme.
 
+## Checking GPU Availability
+
+By default, Flux will run the checks on your system to see if it can support GPU functionality. You can check this explicity by typing `Flux.use_cuda[]` which is the internal flag Flux uses. You can also run:
+
+```julia
+julia> using CUDA
+
+julia> CUDA.functional()
+```
+which is part of the internal code that Flux is running to set the `Flux.use_cuda[]` flag.
+
 ## GPU Usage
 
 Support for array operations on other hardware backends, like GPUs, is provided by external packages like [CUDA](https://github.com/JuliaGPU/CUDA.jl). Flux is agnostic to array types, so we simply need to move model weights and data to the GPU and Flux will handle it.
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -10,6 +10,7 @@ By default, Flux will run the checks on your system to see if it can support GPU
 julia> using CUDA
 
 julia> CUDA.functional()
+true
 ```
 which is part of the internal code that Flux is running to set the `Flux.use_cuda[]` flag (note this flag is not stable and you should use the code above to check for GPU availability).
 
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -4,14 +4,14 @@ NVIDIA GPU support should work out of the box on systems with CUDA and CUDNN ins
 
 ## Checking GPU Availability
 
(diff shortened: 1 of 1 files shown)


ToucheSir:
Depending on how long it takes for https://github.com/FluxML/Flux.jl/pull/1566 to land, we may want to pull over the `default_gpu_converter` functionality into Flux first and reference that there and/or add a user-facing API for checking GPU availability.

logankilpatrick:
I don't have a good grasp of these things right now but I thought this would be useful for the time being.

docs/src/gpu.md
```diff
@@ -2,6 +2,17 @@
 
 NVIDIA GPU support should work out of the box on systems with CUDA and CUDNN installed. For more details see the [CUDA](https://github.com/JuliaGPU/CUDA.jl) readme.
 
+## Checking GPU Availability
+
+By default, Flux will run the checks on your system to see if it can support GPU functionality. You can check this explicity by typing `Flux.use_cuda[]` which is the internal flag Flux uses. You can also run:
```
darsnack, line 7:
```suggestion
By default, Flux will run the checks on your system to see if it can support GPU functionality. You can check if Flux identified a valid GPU setup by typing `Flux.use_cuda[]` which is the internal flag Flux uses. You can also run:
```

darsnack (commented):
I think this will need to be updated for multiple GPUs, but clearer docs never hurt anyone in the meantime.

DhairyaLGandhi:
Should we avoid talking about the `Flux.use_cuda` flag. Its not something stable at all, and is more of a detail which may change any time, in fact I believe Tim also suggested some changes to that mechanism.

darsnack:
Sure we can limit this to just `CUDA.functional`

docs/src/gpu.md
```diff
@@ -2,6 +2,17 @@
...
+## Checking GPU Availability
+
+By default, Flux will run the checks on your system to see if it can support GPU functionality. You can check if Flux identified a valid GPU setup by typing the following:
+
+```julia
+julia> using CUDA
+
+julia> CUDA.functional()
+```
+which is part of the internal code that Flux is running to set the `Flux.use_cuda[]` flag (note this flag is not stable and you should use the code above to check for GPU availability).
```
darsnack, line 14:
```suggestion
which is part of the internal code that Flux is running to set the `Flux.use_cuda[]` flag (note this flag is not part of the public API, and you should use the code above to check for GPU availability).
```

darsnack (dismissed):

docs/src/gpu.md
logankilpatrick: Should we just not mention it if it's not part of the public api?

CarloLucibello: yeah, maybe better remove this sentence entirely and stop at `CUDA.functional()`

docs/src/gpu.md
```diff
@@ -2,6 +2,17 @@
...
 NVIDIA GPU support should work out of the box on systems with CUDA and CUDNN installed. For more details see the [CUDA](https://github.com/JuliaGPU/CUDA.jl) readme.
 
+## Checking GPU Availability
+
+By default, Flux will run the checks on your system to see if it can support GPU functionality. You can check if Flux identified a valid GPU setup by typing the following:
+
+```julia
+julia> using CUDA
+
+julia> CUDA.functional()
```
CarloLucibello, line 12:
```suggestion
julia> CUDA.functional()
true
```

logankilpatrick:
Alright I think we are at the finish line here!

darsnack (approved):

PR #1635, opened 2021-06-27
Fixup `Dataloader`'s docstring

mcabbott:
This makes the Dataloader example into a `jldoctest`, and adds a friendly error if you mess up the sizes.

DataLoader was exported from Data, but accessible only as (deep breath) `Flux.Data.DataLoader` because the module was not imported. I take it that's a mistake, because of the export statement. So I fixed it. 

Note that `julia> Flux.DataL<tab>` still doesn't find it. Perhaps we can murder some sub-modules once the datasets currently residing in Data are removed. 

(Test failure on nightly ought to be fixed by https://github.com/FluxML/NNlibCUDA.jl/pull/15, but CI gets the old version?)

### DIFF
--- a/src/data/dataloader.jl
+++ b/src/data/dataloader.jl
@@ -23,8 +23,9 @@ Takes as input a single data tensor, or a tuple (or a named tuple) of tensors.
 The last dimension in each tensor is the observation dimension, i.e. the one
 divided into mini-batches.
 
-If `shuffle=true`, shuffles the observations each time iterations are re-started.
-If `partial=false`, drops the last mini-batch if it is smaller than the batchsize.
+If `shuffle=true`, it shuffles the observations each time iterations are re-started.
+If `partial=false` and the number of observations is not divisible by the batchsize, 
+then the last mini-batch is dropped.
 
 The original data is preserved in the `data` field of the DataLoader.
 
@@ -70,10 +71,10 @@ true
 julia> first(train_loader).label == Ytrain[1:5]  # because of shuffle=true
 false
 
-julia> foreach(println∘size, Flux.DataLoader(rand(10, 64), batchsize=30))  # partial=false would omit last
-(10, 30)
-(10, 30)
-(10, 4)
+julia> foreach(println∘summary, Flux.DataLoader(rand(Int8, 10, 64), batchsize=30))  # partial=false would omit last
+10×30 Matrix{Int8}
+10×30 Matrix{Int8}
+10×4 Matrix{Int8}
 ```
 """
 function DataLoader(data; batchsize=1, shuffle=false, partial=true, rng=GLOBAL_RNG)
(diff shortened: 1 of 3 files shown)


ToucheSir (approved):
LGTM, also not sure why the REPL autocomplete wouldn't work.

src/data/dataloader.jl
```diff
@@ -100,8 +110,10 @@ _nobs(data::AbstractArray) = size(data)[end]
 function _nobs(data::Union{Tuple, NamedTuple})
     length(data) > 0 || throw(ArgumentError("Need at least one data input"))
     n = _nobs(data[1])
-    if !all(x -> _nobs(x) == n, Base.tail(data))
-        throw(DimensionMismatch("All data should contain same number of observations"))
+    for i in keys(data)
+        ni = _nobs(data[i])
+        n == ni || throw(DimensionMismatch("All data inputs should have the same number of observations, i.e. size in the last dimension. " *
```
DhairyaLGandhi, line 115: Oh weird. This would give you an error if you wanted to pass reals or a single vector with the inputs or some other config that is needed only once.

src/data/dataloader.jl
```diff
@@ -13,59 +13,69 @@ struct DataLoader{D,R<:AbstractRNG}
...
-    end
+julia> Ytrain = rand('a':'z', 100);  # now make a DataLoader yielding 2-element named tuples
+
+julia> train_loader = Flux.DataLoader((data=Xtrain, label=Ytrain), batchsize=5, shuffle=true);
+
+julia> for epoch in 1:100
+         for (x, y) in train_loader  # access via tuple destructuring
+           @assert size(x) == (10, 5)
+           @assert size(y) == (5,)
+           # loss += f(x, y) # etc, runs 100 * 20 times
```
DhairyaLGandhi, line 61:
```suggestion
```

src/data/dataloader.jl
```diff
@@ -13,59 +13,69 @@ struct DataLoader{D,R<:AbstractRNG}
...
+         end
+       end
+
+julia> first(train_loader) isa NamedTuple{(:data, :label)}
```
DhairyaLGandhi, line 65:
```suggestion
julia> typeof(first(train_loader)) # minibatches respect the keys of the input data
```

Might be clearer.

src/data/dataloader.jl
mcabbott: Not sure what you're thinking, and whether you're complaining about current behaviour. MWE?

DhairyaLGandhi (approved):
bors r+

src/data/dataloader.jl
DhairyaLGandhi: Definitely an oddity of the current behavior.

DhairyaLGandhi:
By the by, the suggestion was to note that the minibatches respect the keys of the named tuple as given in the input dataset. I think it got lost in the GitHub commits.

src/data/dataloader.jl
mcabbott: OK. But, MWE? Or make an issue.

PR #1636, opened 2021-06-27
Add warnings for mismatched sizes in losses

### LINKED ISSUE #1522: Warn different size inputs in loss functions
Say I have a model output a scalar, and the batch size is 5.
`m(x)` gives me an array of size `(1, 5)`, while `y` is of size `(5, )`. 
If I pass them to `Flux.Losses.mae`, the value is different than what I expect due to broadcasting.

In PyTorch, I get a warning message.
```
dense.py:43: UserWarning: Using a target size (torch.Size([5])) that is different to the input size (torch.Size([5, 1])). This will likely lead to incorrect results due to broadcasting. Please ensure they have the same size.
```

It would be nice to have a warning message in Flux so I know I'm doing something wrong.

### LINKED ISSUE #1599: Why does `DataLoader` not throw an error when fed with a 1D vector for the target?
Suppose you have the following simple example:
```
# a data set of two training examples [0 ,0] and [1, 1], with respective output targets of 0 and 1: 
train_input = [0 1; 0 1]
train_target = [0, 1]
```
We generate a machine learning model as follows:
```
# define the model
model = Chain(Dense(2, 1, σ))

# define parameters and cost function
parameters = params(model)
loss(x, y) = mse(model(x), y)

# set optimization algorithm
η = 0.1
optimizer = Descent(η)

loader = DataLoader((train_input, train_target), batchsize=2)

@epochs 1000 train!(loss, parameters, loader, optimizer)
```
The code runs without throwing any errors, though the model actually converges to a constant output value and fails to approximate the variation of the data.

If, however, one defines `train_target` as
```
train_target = [0 1]
``` 
which is a 1*2 matrix, the model behaves as expected, and we get a good result.

The documentation of `DataLoader` says that `train_input` and `train_target` have to be tensors, with the last dimension in each of them to be the observation dimension. Are 1D vectors not included in this definition of "tensors"? If so, why doesn't `DataLoader` throw an error?

mcabbott:
Closes #1599, I think, by making loss functions give a warning if the sizes don't match:
```julia
julia> mse([1,0], [1 0 0])
┌ Error: size mismatch in loss function! In future this will be an error; in Flux 0.12 broadcasting acceps some mismatches
│   summary(ŷ) = "2-element Vector{Int64}"
│   summary(y) = "1×3 Matrix{Int64}"
└ @ Flux.Losses ~/.julia/dev/Flux/src/losses/utils.jl:29
0.5

julia> @btime gradient(sum∘mse, $(rand(10,100)), $(rand(10,100)));
  19.709 μs (130 allocations: 51.25 KiB)
  19.625 μs (130 allocations: 51.25 KiB)
```
Appears to have no effect on speed, although Zygote is weird and maybe someone has a better test of that.

Edit -- closes #1522, too.

### DIFF
--- a/src/losses/utils.jl
+++ b/src/losses/utils.jl
@@ -1,7 +1,7 @@
 """
-xlogx(x)
+    xlogx(x)
 
-Return `x * log(x)` for `x ≥ 0`, handling `x = 0` by taking the downward limit.
+Return `x * log(x)` for `x ≥ 0`, handling `x == 0` by taking the limit from above, to get zero.
 """
 function xlogx(x)
   result = x * log(x)
@@ -9,9 +9,9 @@ function xlogx(x)
 end
 
 """
-xlogy(x, y)
+    xlogy(x, y)
 
-Return `x * log(y)` for `y > 0` with correct limit at `x = 0`.
+Return `x * log(y)` for `y > 0`, and zero when `x == 0`.
 """
 function xlogy(x, y)
   result = x * log(y)
--- a/src/losses/utils.jl
+++ b/src/losses/utils.jl
@@ -24,12 +24,13 @@ end
 end
 
 # This can be made an error in Flux v0.13, for now just a warning
-function match_sizes(ŷ, y)
+function match_sizes(ŷ::AbstractArray, y::AbstractArray)
   for d in 1:max(ndims(ŷ), ndims(y)) 
     if size(ŷ,d) != size(y,d)
       @error "size mismatch in loss function! In future this will be an error; in Flux 0.12 broadcasting acceps some mismatches" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
     end
   end
 end
+match_sizes(ŷ, y) = nothing  # pass-through, for constant label e.g. y = 1
 
 Zygote.@nograd match_sizes
--- a/src/losses/utils.jl
+++ b/src/losses/utils.jl
@@ -24,13 +24,13 @@ end
 end
 
 # This can be made an error in Flux v0.13, for now just a warning
-function match_sizes(ŷ::AbstractArray, y::AbstractArray)
+function _check_sizes(ŷ::AbstractArray, y::AbstractArray)
   for d in 1:max(ndims(ŷ), ndims(y)) 
     if size(ŷ,d) != size(y,d)
       @error "size mismatch in loss function! In future this will be an error; in Flux 0.12 broadcasting acceps some mismatches" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
     end
   end
 end
-match_sizes(ŷ, y) = nothing  # pass-through, for constant label e.g. y = 1
+_check_sizes(ŷ, y) = nothing  # pass-through, for constant label e.g. y = 1
 
-Zygote.@nograd match_sizes
+Zygote.@nograd _check_sizes
--- a/src/losses/utils.jl
+++ b/src/losses/utils.jl
@@ -27,7 +27,7 @@ end
 function _check_sizes(ŷ::AbstractArray, y::AbstractArray)
   for d in 1:max(ndims(ŷ), ndims(y)) 
     if size(ŷ,d) != size(y,d)
-      @warn "size mismatch in loss function! In future this will be an error. In Flux <= 0.12 broadcasting acceps this, but may not give sensible results" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
+      @warn "Size mismatch in loss function! In future this will be an error. In Flux <= 0.12 broadcasting accepts this, but may not give sensible results" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
     end
   end
 end
--- a/src/losses/utils.jl
+++ b/src/losses/utils.jl
@@ -25,8 +25,10 @@ end
 
 # This can be made an error in Flux v0.13, for now just a warning
 function match_sizes(ŷ, y)
-  if size(ŷ) != size(y)
-    @error "size mismatch in loss function! In future this will be an error; in Flux 0.12 broadcasting acceps some mismatches" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
+  for d in 1:max(ndims(ŷ), ndims(y)) 
+    if size(ŷ,d) != size(y,d)
+      @error "size mismatch in loss function! In future this will be an error; in Flux 0.12 broadcasting acceps some mismatches" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
+    end
   end
 end
 
--- a/src/losses/utils.jl
+++ b/src/losses/utils.jl
@@ -22,3 +22,12 @@ end
   res = xlogy.(x, y)
   res, Δ -> (nothing, Zygote.unbroadcast(x, xlogy.(Δ, y)), Zygote.unbroadcast(y, Δ .* x ./ y))
 end
+
+# This can be made an error in Flux v0.13, for now just a warning
+function match_sizes(ŷ, y)
+  if size(ŷ) != size(y)
+    @error "size mismatch in loss function! In future this will be an error; in Flux 0.12 broadcasting acceps some mismatches" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
+  end
+end
+
+Zygote.@nograd match_sizes
--- a/src/losses/utils.jl
+++ b/src/losses/utils.jl
@@ -27,7 +27,7 @@ end
 function _check_sizes(ŷ::AbstractArray, y::AbstractArray)
   for d in 1:max(ndims(ŷ), ndims(y)) 
     if size(ŷ,d) != size(y,d)
-      @error "size mismatch in loss function! In future this will be an error; in Flux 0.12 broadcasting acceps some mismatches" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
+      @warn "size mismatch in loss function! In future this will be an error. In Flux <= 0.12 broadcasting acceps this, but may not give sensible results" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
     end
   end
 end
--- a/src/losses/ctc.jl
+++ b/src/losses/ctc.jl
@@ -132,6 +132,7 @@ for mathematical details.
 ctc_loss(ŷ::AbstractArray, y) = ctc_alpha(ŷ, y).loss
 
 @adjoint function ctc_loss(ŷ, y)
+  match_sizes(ŷ, y)
   out = ctc_alpha(ŷ, y)
   ctc_loss_pullback(Δ) = (Δ .* ∇ctc_loss(ŷ, y, out), nothing)
   return out.loss, ctc_loss_pullback
--- a/src/losses/ctc.jl
+++ b/src/losses/ctc.jl
@@ -132,7 +132,6 @@ for mathematical details.
 ctc_loss(ŷ::AbstractArray, y) = ctc_alpha(ŷ, y).loss
 
 @adjoint function ctc_loss(ŷ, y)
-  match_sizes(ŷ, y)
   out = ctc_alpha(ŷ, y)
   ctc_loss_pullback(Δ) = (Δ .* ∇ctc_loss(ŷ, y, out), nothing)
   return out.loss, ctc_loss_pullback
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -19,7 +19,7 @@ julia> Flux.mae(y_model, 1:3)
 ```
 """
 function mae(ŷ, y; agg = mean)
-  match_sizes(ŷ, y)
+  _check_sizes(ŷ, y)
   agg(abs.(ŷ .- y))
 end
 
@@ -43,7 +43,7 @@ julia> Flux.mse(y_model, y_true)
 ```
 """
 function mse(ŷ, y; agg = mean)
-  match_sizes(ŷ, y)
+  _check_sizes(ŷ, y)
   agg((ŷ .- y) .^ 2)
 end
 
@@ -67,7 +67,7 @@ julia> Flux.msle(Float32[0.9, 1.8, 2.7], 1:3)
 ```
 """
 function msle(ŷ, y; agg = mean, ϵ = epseltype(ŷ))
-  match_sizes(ŷ, y)
+  _check_sizes(ŷ, y)
   agg((log.((ŷ .+ ϵ) ./ (y .+ ϵ))) .^2 )
 end
 
@@ -82,7 +82,7 @@ given the prediction `ŷ` and true values `y`.
                  |  δ * (|ŷ - y| - 0.5 * δ), otherwise
 """
 function huber_loss(ŷ, y; agg = mean, δ = ofeltype(ŷ, 1))
-   match_sizes(ŷ, y)
+   _check_sizes(ŷ, y)
    abs_error = abs.(ŷ .- y)
    #TODO: remove dropgrad when Zygote can handle this function with CuArrays
    temp = Zygote.dropgrad(abs_error .<  δ)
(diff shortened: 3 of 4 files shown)


DhairyaLGandhi:
We converged on adding a documentation note for the issue. Its unhelpful to restrict the usability of a function if it has little to do with how it's written. Its also better not to complain when use cases go out of the box. 

There's also no intent to error in future versions.

mcabbott:
Not sure who "we" refers to, but 4 people in the linked issue find it surprising that `mse` silently computes something which isn't MSE.

Got a real example of this strange behaviour being useful, instead of being a footgun?

DhairyaLGandhi:
Image warps create mismatched dimensions.

darsnack:
> Image warps create mismatched dimensions.

I'm not terribly familiar with that use case, so maybe you could specify some code/share a reference? If the original image is WxH and the model produces a HxW, then the model should be transposing the output anyways. If the sizes are completely mismatched, then how would broadcasting even return the correct result?

For the PR, I think a warning with `maxlog=1` might be better than an error. Personally, I don't like to error unless the user is asking us to do something blatantly wrong.

DhairyaLGandhi:
We are working on DiffImages.jl which uses image transformations and images
in general.

darsnack:
I understand that you are using image transformations. I'm asking how that specifically affects what's being discussed in this PR. I did a search online to try and find a program where this was the case, and I couldn't. So it would be helpful if you could provide an example from your work with DiffImages.jl.

mcabbott:
> I don't like to error unless the user is asking us to do something blatantly wrong.

I guess the position taken by this PR is that this *is* wrong. The specification of (say) MSE is that it compares two arrays the same shape; the fact that this is implemented using broadcasting instead of loops is a detail. 

I'm not quite confident enough to make it an immediate error, maybe someone really does have a sensible use case, which relies on broadcasting... in which case they will tell us, and we can revisit the issue. But I'm opposed to leaving things like this a warning forever. The function either accepts some shape or it doesn't. Input the wrong shape means you're computing garbage. (Permanent warnings for more interactive things, like maybe trying to move a model to the GPU when you don't have one, that's different.)

DhairyaLGandhi:
Well, given that DiffImages is still under development, it's harder to give you a sufficiently complete mwe - but the key is that warps are not shape preserving operations, and the homography has to be learned by comparing images of mismatched dimensions. There are cases (such as augmentation) where we would be able to crop to an roi, and others (such as stitching, or pose) where the model will need to learn by itself. 

Future looking work requires some freedoms. I think you're over estimating the number of people that would speak up if their specific case isn't met.

ToucheSir:
Thanks for the clarification, but honestly at this point it would help to have _any_ example. Even just a line in a paper or other library somewhere that says something to the effect of "we apply [standard loss function] to inputs x and y [of differing sizes in dimensions a, b, and c]". There are certainly some losses where the labels and targets may not have the same dimensions (anything with labels comes to mind), but I've never heard of one where sizes _along_ the same dimension vary. 

Moreover, none of the loss functions we have currently meet this criteria! The purpose of this PR is not to force `size(ŷ) == size(y)` for all present and future loss functions forevermore, but to add helpful checks to those where we _can_ say with certainty what the expected sizes should be. It's not like `mse(rand(10, 10), rand(5, 5))` works currently anyhow.

> I think you're over estimating the number of people that would speak up if their specific case isn't met.

Fair enough, but if we apply this argument to the other side, then there must be an order of magnitude more people who have encountered incorrect behaviour with not passing correctly shaped inputs to our existing losses because some of them have taken the time to speak up. Given that nobody else in this thread has even remotely seen or heard of a counterexample, let me implore again for something, _anything_, that is concrete enough for us to look at it and say "ok, this means that [mse/mae/logitcrossentropy/etc] can not assert any relationship between the sizes of its inputs".

CarloLucibello:
The only case where I found myself using mismatched sizes is with the binarycrossentropy in the discriminator loss of a GAN. There the class is 1 for all real images and -1 for all fake ones, so I used `y = 1` instead of `y=ones(size(ŷ))`.
Not a big deal, it's perfectly fine to create an all-ones vector in this situation, but surely feeding just 1 is slightly more convenient.
Maybe we can add to each loss a flag to suppress the warning
```
function mse(....; check_sizes=true)
     check_sizes && match_sizes(ŷ, y)
     ...
end
``` 
The default should be `check_sizes=true`.

mcabbott:
OK, that's an actual example, good. (One noun trumps a thousand adjectives.) I suppose you can also write it out, but perhaps it's easier not to have to worry about the minus sign yourself...
```julia
julia> ys = rand(5, 20);

julia> Flux.binarycrossentropy(ys, 1)  # check_sizes=false
0.8859776657579094

julia> -mean(log, ys)
0.8859776657579111
```
Perhaps a keyword ends up being quite a mouthful. Another idea would be to restrict the check to `match_sizes(ŷ::AbstractArray, y::AbstractArray)`.

CarloLucibello:
Another consideration is that the argument [here](https://github.com/FluxML/Flux.jl/pull/1636#issuecomment-873284296) is quite convincing, users should not rely on an implementation detail, so we should give a warning now and throw an error later.

On the other hand, the implementation detail is exposed in the docstrings, so users could actually rely on that. I exposed it in the docstrings because it was a convenient way to express the (standard) behavior though, not because I wanted people to rely on broadcasting. 

Anyways, the status quo is not good, people are bitten by this issue, mostly with MSE. Let's go with the `match_sizes(ŷ::AbstractArray, y::AbstractArray)` checks,   other cases are more likely to happen by design, as in the GAN example, than by mistake.

mcabbott:
> detail is exposed in the docstrings, ... was a convenient way to express the (standard) behavior

That seems fine to me. 

But, BTW, are we aggregating correctly? If I understand right, MSE with `agg=sum` ought to change how the MSE from different observations in the batch are combined, but not the M in MSE. So shouldn't it be `sum(mean((ŷ .- y) .^ 2; dims=1:ndims(y)-1))`?

CarloLucibello:
The M in MSE refers to how different observations in the batch are combined, so in practice when we choose `agg != mean` we are not computing MSE anymore, but it is fine and standard to have the extra flexibility of choosing `agg`. 

One could argue that the E in MSE being an error, it is supposed to be a scalar quantity for each observation, so a sensible definition would be
```julia
agg(sum((ŷ .- y) .^ 2; dims=1:ndims(y)-1))
```

But our current choice is also fine, it just treats each element of y as an independent observation. It's also what pytorch does.

Btw, while looking into how pytorch implements MSE, I bumped into this
https://github.com/pytorch/pytorch/pull/5085 
which is related to this PR

mcabbott:
Oh right, you are correct this is what Pytorch does. I hadn't tried, and mis-read the docs https://pytorch.org/docs/stable/generated/torch.nn.MSELoss.html (note that they use `n` for two things!) where in "The division by `n` can be avoided if one sets reduction = 'sum'." they mean division by the feature dimension (i.e. removing the M in MSE), not by the batch dim `N`. Still seems weird to me, that it makes the sum over the batch of, err, Total squared error. (I'd call your formula `agg` over the batch of TSE.) But not a big deal.

Can also confirm that they do give an error on any shape mismatch. Nicely done finding the issue; their behaviour before adding a shape check was like `mean((vec(ŷ) .- vec(y)) .^ 2)` which will allow a different but equally confusing set of silent mistakes.

mcabbott:
An example from the wild today:

yhat
image-1
y
image

```
julia> y = [1,0,1,0];

julia> yhat = [0.5 0.5 0.5 0.0498733];  # I inserted a zero but did not change type

julia> Flux.binarycrossentropy(yhat, y)  # silently wrong answer
0.9010390634731245

julia> Flux.binarycrossentropy(vec(yhat), y)
0.5326503691347774
```

src/losses/utils.jl
```diff
@@ -22,3 +22,15 @@ end
   res = xlogy.(x, y)
   res, Δ -> (nothing, Zygote.unbroadcast(x, xlogy.(Δ, y)), Zygote.unbroadcast(y, Δ .* x ./ y))
 end
+
+# This can be made an error in Flux v0.13, for now just a warning
+function match_sizes(ŷ::AbstractArray, y::AbstractArray)
```
CarloLucibello, line 27: Can we call this `check_sizes_match` or something similar?

mcabbott: Sure. How about `Flux.Losses._check_sizes`?

src/losses/utils.jl
```diff
@@ -22,3 +22,15 @@ end
   res = xlogy.(x, y)
   res, Δ -> (nothing, Zygote.unbroadcast(x, xlogy.(Δ, y)), Zygote.unbroadcast(y, Δ .* x ./ y))
 end
+
+# This can be made an error in Flux v0.13, for now just a warning
+function _check_sizes(ŷ::AbstractArray, y::AbstractArray)
+  for d in 1:max(ndims(ŷ), ndims(y)) 
+    if size(ŷ,d) != size(y,d)
+      @error "size mismatch in loss function! In future this will be an error; in Flux 0.12 broadcasting acceps some mismatches" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
```
darsnack, line 30: Did we decide to error instead of warn?

mcabbott: My suggestion is to warn loudly right now, and error in the next breaking release.

darsnack: I agree, and maybe I'm just missing something, but doesn't the line above throw an error?

mcabbott: I guess it says "Error" but it carries on running -- like the top message. It's `@warn`'s louder cousin.

CarloLucibello: maybe a `@warn` is enough?

CarloLucibello (dismissed):

src/losses/utils.jl
```diff
@@ -22,3 +22,15 @@ end
   res = xlogy.(x, y)
   res, Δ -> (nothing, Zygote.unbroadcast(x, xlogy.(Δ, y)), Zygote.unbroadcast(y, Δ .* x ./ y))
 end
+
+# This can be made an error in Flux v0.13, for now just a warning
+function _check_sizes(ŷ::AbstractArray, y::AbstractArray)
+  for d in 1:max(ndims(ŷ), ndims(y)) 
+    if size(ŷ,d) != size(y,d)
+      @warn "size mismatch in loss function! In future this will be an error. In Flux <= 0.12 broadcasting acceps this, but may not give sensible results" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
```
darsnack, line 30:
```suggestion
      @warn "Size mismatch in loss function! In future this will be an error. In Flux <= 0.12 broadcasting accepts this, but may not give sensible results" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
```

darsnack (changes requested):

src/losses/utils.jl
mcabbott: Done.

darsnack (approved):

DhairyaLGandhi:
Came across https://github.com/SciML/DiffEqFlux.jl/blob/master/test/newton_neural_ode.jl which does indeed use mismatched shapes for its uses. See CI logs such as https://github.com/SciML/DiffEqFlux.jl/pull/592/checks?check_run_id=3128353069

mcabbott:
The error there is surprising: 
```
289
┌ Warning: Size mismatch in loss function! In future this will be an error. In Flux <= 0.12 broadcasting accepts this, but may not give sensible results
290
│   summary(ŷ) = "1×5 Matrix{Float64}"
291
│   summary(y) = "VectorOfArray{Float64,3}"
292
└ @ Flux.Losses ~/.julia/packages/Flux/ma5BY/src/losses/utils.jl:30
293
WARNING: both Flux and Iterators export "flatten"; uses of it in module DiffEqFlux must be qualified
294
WARNING: both Flux and Distributions export "params"; uses of it in module DiffEqFlux must be qualified
295
Newton Neural ODE Tests: Error During Test at /home/runner/.julia/packages/SafeTestsets/A83XK/src/SafeTestsets.jl:25
296
297
  LoadError: BoundsError: attempt to access Tuple{Int64, Int64} at index [3]
298
  Stacktrace:
299
    [1] getindex
300
      @ ./tuple.jl:29 [inlined]
301
    [2] size
302
      @ ./abstractarray.jl:38 [inlined]
303
    [3] _check_sizes(ŷ::Matrix{Float64}, y::ODESolution{Float64, 3, Vector{Vector{Float64}}, Nothing, Nothing, Vector{Float64}, Vector{Vector{Vector{Float64}}}, ODEProblem{Matrix{Float64}, Tuple{Float64, Float64}, false, Vector{Float32}, ODEFunction{false, DiffEqFlux.var"#dudt_#107"{NeuralODE{Chain{Tuple{Dense{typeof(tanh), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}, Vector{Float32}, Flux.var"#62#64"{Chain{Tuple{Dense{typeof(tanh), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}}, Tuple{Float64, Float64}, Tuple{OrdinaryDiffEq.ROCK4{Nothing}}, Base.Iterators.Pairs{Symbol, Any, Tuple{Symbol, Symbol}, NamedTuple{(:reltol, :saveat), Tuple{Float64, Vector{Float64}}}}}}, LinearAlgebra.UniformScaling{Bool}, Nothing, typeof(DiffEqFlux.basic_tgrad), Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED), Nothing}, Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, SciMLBase.StandardODEProblem}, OrdinaryDiffEq.ROCK4{Nothing}, OrdinaryDiffEq.InterpolationData{ODEFunction{false, DiffEqFlux.var"#dudt_#107"{NeuralODE{Chain{Tuple{Dense{typeof(tanh), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}, Vector{Float32}, Flux.var"#62#64"{Chain{Tuple{Dense{typeof(tanh), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}}, Tuple{Float64, Float64}, Tuple{OrdinaryDiffEq.ROCK4{Nothing}}, Base.Iterators.Pairs{Symbol, Any, Tuple{Symbol, Symbol}, NamedTuple{(:reltol, :saveat), Tuple{Float64, Vector{Float64}}}}}}, LinearAlgebra.UniformScaling{Bool}, Nothing, typeof(DiffEqFlux.basic_tgrad), Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, Nothing, typeof(SciMLBase.DEFAULT_OBSERVED), Nothing}, Vector{Vector{Float64}}, Vector{Float64}, Vector{Vector{Vector{Float64}}}, OrdinaryDiffEq.ROCK4ConstantCache{StaticArrays.SVector{6, Float64}, StaticArrays.SVector{4, Float64}, StaticArrays.SVector{5, Float64}, Float64, Matrix{Float64}}}, DiffEqBase.DEStats})
304
      @ Flux.Losses ~/.julia/packages/Flux/ma5BY/src/losses/utils.jl:29
305
    [4] _pullback
306
      @ ~/.julia/packages/Zygote/TaBlo/src/lib/grad.jl:8 [inlined]
```
So it looks like it comes from a too-naiive `size` implementation, which does `size(x,d) == size(x)[d]`. But neither VectorOfArray nor Matrix are so naiive:
```
julia> using RecursiveArrayTools

julia> va = VectorOfArray([rand(Int8, 2,3) for _ in 1:4])
VectorOfArray{Int8,3}:
4-element Vector{Matrix{Int8}}:
 [-20 -99 3; 1 83 56]
 [114 0 -69; 72 -100 -120]
 [-103 -77 6; 88 32 48]
 [109 52 107; -105 -87 36]

julia> summary(va)
"VectorOfArray{Int8,3}"

julia> size(va)
(2, 3, 4)

julia> size(va, 4) # correct
1

julia> size(rand(1,5), 3)
1
```
So in fact, `VectorOfArray` is causing the warning (which seems correct) but `ODESolution` might have a bug in how it defines `size`?

PR #1639, opened 2021-06-27
fix recurrence docs

### LINKED ISSUE #1638: bug in RNN docs
The documentation for recurrent neural networks
https://github.com/FluxML/Flux.jl/blob/master/docs/src/models/recurrence.md
provides examples that would lead to erroneous gradient computations. 

This is due to the fact that while `Recur` special cases broadcast and forwards it to the map machinery, which in turn reverses the execution order when computing the adjoint for vector inputs, the same does not apply to `Chain` or generic stateful functions.

Here I show explicitly the issue:
```julia
using Flux
using FiniteDifferences
using Random

m = Chain(RNN(2, 5), Dense(5, 1))

function loss(x, y)
    Flux.reset!(m)
    sum((Flux.stack(m.(x), 1) .- y) .^ 2)
end


function loss_loop(x, y)
    Flux.reset!(m)
    l = 0f0 
    for (xi, yi) in zip(x, y)
        l += sum((m(xi) .- yi).^ 2)
    end
    return l
end

Random.seed!(17)
x = [rand(Float32, 2) for i = 1:3]
y = rand(Float32, 3)

print(loss_loop(x, y)) # 3.1758256
print(loss(x, y))      # 3.1758256

println(gradient(x -> loss_loop(x, y), x)[1])
# Vector{Float32}[[-0.27259377, 0.3039751], [1.0806422, 0.79439837], [-0.6706671, 0.6848929]]

println(gradient(x -> loss(x, y), x)[1]) # WRONG GRADIENT
# Vector{Float32}[[-0.09357249, 0.2245327], [0.42562985, 0.81408167], [0.044034332, 1.131884]]

fdm = FiniteDifferences.central_fdm(5,1)
println(FiniteDifferences.grad(fdm, x -> loss(x, y), x)[1])
# Vector{Float32}[[-0.27259293, 0.3039818], [1.0806386, 0.7944023], [-0.670666, 0.68489385]]
```

I think we should stop directing users toward the use of `broadcast` and `map` when using stateful layers, the julia language doesn't give any ordering guarantees so why should we? 

Related to https://github.com/FluxML/Zygote.jl/pull/807

CarloLucibello:
fix #1638

### DIFF
--- a/docs/src/models/recurrence.md
+++ b/docs/src/models/recurrence.md
@@ -13,7 +13,7 @@ An aspect to recognize is that in such model, the recurrent cells `A` all refer 
 In the most basic RNN case, cell A could be defined by the following: 
 
 ```julia
-Wxh = randn(Float32, 5, 2)
+Wxh = randn(Float32, 5, 4)
 Whh = randn(Float32, 5, 5)
 b   = randn(Float32, 5)
 
@@ -22,7 +22,7 @@ function rnn_cell(h, x)
     return h, h
 end
 
-x = rand(Float32, 2) # dummy data
+x = rand(Float32, 4) # dummy data
 h = rand(Float32, 5)  # initial hidden state
 
 h, y = rnn_cell(h, x)
@@ -37,9 +37,9 @@ There are various recurrent cells available in Flux, notably `RNNCell`, `LSTMCel
 ```julia
 using Flux
 
-rnn = Flux.RNNCell(2, 5)
+rnn = Flux.RNNCell(4, 5)
 
-x = rand(Float32, 2) # dummy data
+x = rand(Float32, 4) # dummy data
 h = rand(Float32, 5)  # initial hidden state
 
 h, y = rnn(h, x)
@@ -50,7 +50,7 @@ h, y = rnn(h, x)
 For the most part, we don't want to manage hidden states ourselves, but to treat our models as being stateful. Flux provides the `Recur` wrapper to do this.
 
 ```julia
-x = rand(Float32, 2)
+x = rand(Float32, 4)
 h = rand(Float32, 5)
 
 m = Flux.Recur(rnn, h)
@@ -60,11 +60,11 @@ y = m(x)
 
 The `Recur` wrapper stores the state between runs in the `m.state` field.
 
-If we use the `RNN(2, 5)` constructor – as opposed to `RNNCell` – you'll see that it's simply a wrapped cell.
+If we use the `RNN(4, 5)` constructor – as opposed to `RNNCell` – you'll see that it's simply a wrapped cell.
 
 ```julia
-julia> RNN(2, 5)
-Recur(RNNCell(2, 5, tanh))
+julia> RNN(4, 5)
+Recur(RNNCell(4, 5, tanh))
 ```
 
 Equivalent to the `RNN` stateful constructor, `LSTM` and `GRU` are also available. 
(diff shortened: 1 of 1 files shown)


docs/src/models/recurrence.md
```diff
@@ -60,113 +60,130 @@ y = m(x)
...
+    since the julia language doesn't guarantee a specific execution order.
+    Therefore, avoid  
+    ```julia
+    y = m.(x)
+    # or 
+    y = map(m, x)
+    ```
+    and use explicit loops 
+    ```julia
+    y = [m(x) for x in x]
```
DhairyaLGandhi, line 122: I think `foldl` might be more appropriate to hit some optimised methods. Although, I think longer term, this isn't the cleanest of APIs, and maybe handling it in broadcasting is better.

docs/src/models/recurrence.md
```diff
@@ -60,113 +60,130 @@ y = m(x)
...
 ```julia
-x = [rand(Float32, 2) for i = 1:3]
-julia> m.(x)
-3-element Array{Array{Float32,1},1}:
- [-0.17945863]
- [-0.20863166]
- [-0.20693761]
+julia> x = [rand(Float32, 4) for i = 1:3];
+
+julia> [m(x[i]) for i = 1:3]
```
darsnack, line 104:
```suggestion
julia> [m(xi) for xi in x]
```

docs/src/models/recurrence.md
```diff
@@ -60,113 +60,130 @@ y = m(x)
...
 If for some reason one wants to exclude the first step of the RNN chain for the computation of the loss, that can be handled with:
 
 ```julia
 function loss(x, y)
-  sum((Flux.stack(m.(x)[2:end],1) .- y) .^ 2)
+  m(x[1]) # ignores the output but updates the hidden states
+  l = 0f0
+  for i in 2:length(x)
+    l += sum((m(x[i]) .- y[i-1]).^2)
+  end
```
darsnack, line 133: Some alternatives that I think read cleaner:

```suggestion
  l = mapfoldl(+, x[2:end], y) do xi, yi
    sum((m(xi) .- yi).^2)
  end
```

```suggestion
  l = sum(sum((m(xi) .- yi).^2) for (xi, yi) in zip(x[2:end], y))
```

docs/src/models/recurrence.md
```diff
@@ -60,113 +60,130 @@ y = m(x)
...
 
-y = rand(Float32, 2)
-julia> loss(x, y)
-1.7021208968648693
+y = [rand(Float32, 2) for i=1:2]
+loss(x, y)
 ```
 
-In such model, only `y2` and `y3` are used to compute the loss, hence the target `y` being of length 2. This is a strategy that can be used to easily handle a `seq-to-one` kind of structure, compared to the `seq-to-seq` assumed so far.   
+In such model, only the last two outputs are used to compute the loss, hence the target `y` being of length 2. This is a strategy that can be used to easily handle a `seq-to-one` kind of structure, compared to the `seq-to-seq` assumed so far.
```
darsnack, line 141:
```suggestion
In such a model, only the last two outputs are used to compute the loss, hence the target `y` being of length 2. This is a strategy that can be used to easily handle a `seq-to-one` kind of structure, compared to the `seq-to-seq` assumed so far.   
```

docs/src/models/recurrence.md
```diff
@@ -60,113 +60,130 @@ y = m(x)
...
 
-In such model, only `y2` and `y3` are used to compute the loss, hence the target `y` being of length 2. This is a strategy that can be used to easily handle a `seq-to-one` kind of structure, compared to the `seq-to-seq` assumed so far.   
+In such model, only the last two outputs are used to compute the loss, hence the target `y` being of length 2. This is a strategy that can be used to easily handle a `seq-to-one` kind of structure, compared to the `seq-to-seq` assumed so far.   
 
 Alternatively, if one wants to perform some warmup of the sequence, it could be performed once, followed with a regular training where all the steps of the sequence would be considered for the gradient update:
 
 ```julia
 function loss(x, y)
-  sum((Flux.stack(m.(x),1) .- y) .^ 2)
+  sum(sum((m(x[i]) .- y[i]).^2) for i=1:length(x))
```
darsnack, line 147:
```suggestion
  sum(sum((m(xi) .- yi).^2) for (xi, yi) in zip(x, y))
```

docs/src/models/recurrence.md
```diff
@@ -60,113 +60,130 @@ y = m(x)
...
 end
 
-seq_init = [rand(Float32, 2) for i = 1:1]
-seq_1 = [rand(Float32, 2) for i = 1:3]
-seq_2 = [rand(Float32, 2) for i = 1:3]
+seq_init = [rand(Float32, 4) for i = 1:1]
```
darsnack, line 150:
```suggestion
seq_init = [rand(Float32, 4)]
```

docs/src/models/recurrence.md
```diff
@@ -60,113 +60,130 @@ y = m(x)
...
+That would mean that we have 4 sentences (or samples), each with 4 features (let's say a very small embedding!) and each with a length of 3 (3 words per sentence). Computing `m(batch[1])`, would still represent `x1 -> y1` in our diagram and returns the first word output, but now for each of the 4 independent sentences (second dimension of the input matrix).
 
 In many situations, such as when dealing with a language model, each batch typically contains independent sentences, so we cannot handle the model as if each batch was the direct continuation of the previous one. To handle such situation, we need to reset the state of the model between each batch, which can be conveniently performed within the loss function:
 
 ```julia
 function loss(x, y)
   Flux.reset!(m)
-  sum((Flux.stack(m.(x),1) .- y) .^ 2)
+  sum(sum((m(x[i]) .- y[i]).^2) for i=1:length(x))
```
darsnack, line 186:
```suggestion
  sum(sum((m(xi) .- yi).^2) for (xi, yi) in zip(x, y))
```

darsnack (changes requested):
Also, do we want to replace `sum((m(x) .- y).^2)` with `mse(m(x), y; agg=sum)` everywhere? It might make things easier to read than nested sums.

DhairyaLGandhi:
Let's keep things explicit for the docs. Fine to add a note saying it's equivalent, but use the explicit transform since it's clearer to read why this section is the way it is.

darsnack:
Is there a reason the loss per batch is just added and not averaged? Cause without `agg` I think `mse` is much more explicit and readable.

DhairyaLGandhi:
Its using a function that Flux defines, so it's not more explicit by definition.

docs/src/models/recurrence.md
ToucheSir: `foldl` makes a lot of sense for a single output, but perhaps not so much for a sequence because you'd have to use `cat` as the accumulator function (much slower than materializing the generator)

docs/src/models/recurrence.md
```diff
@@ -60,113 +60,130 @@ y = m(x)
...
+In this previous example, model's state is first reset with `Flux.reset!`. Then, there's a warmup that is performed over a sequence of length 1 by feeding it with `seq_init`, resulting in a warmup state. The model can then be trained for 1 epoch, where 2 batches are provided (`seq_1` and `seq_2`) and all the timesteps outputs are considered for the loss.
 
 In this scenario, it is important to note that a single continuous sequence is considered. Since the model state is not reset between the 2 batches, the state of the model flows through the batches, which only makes sense in the context where `seq_1` is the continuation of `seq_init` and so on.
 
 Batch size would be 1 here as there's only a single sequence within each batch. If the model was to be trained on multiple independent sequences, then these sequences could be added to the input data as a second dimension. For example, in a language model, each batch would contain multiple independent sentences. In such scenario, if we set the batch size to 4, a single batch would be of the shape:
 
 ```julia
-batch = [rand(Float32, 2, 4) for i = 1:3]
+batch = [rand(Float32, 4, 4) for i = 1:3]
```
jeremiedb, line 176: Is there a reason to move the number of features from 2 to 4? Doing so appears to add an ambiguity on which dimension contains the features vs the samples given they are now both 4.

docs/src/models/recurrence.md
CarloLucibello: yes I think we'll have to generally assume and support ordered broadcast in the future

docs/src/models/recurrence.md
CarloLucibello: reverted

darsnack (approved):
LGTM!

Issue #1596, opened 2021-05-11
Reduce hcat creates dense matrix


racinmat:
I think reducing onehot matrices should result in onehot matrix otherwise we lose the advantage of onehot matrices.
See
```julia
julia>   using Flux: OneHotArray, OneHotVector, OneHotMatrix, OneHotLike

julia>   om = OneHotMatrix(rand(1:10, 5), 10)
10×5 Flux.OneHotArray{10,2,Vector{Int64}}:
 0  0  0  0  0
 1  0  1  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  1  0
 0  0  0  0  1
 0  0  0  0  0
 0  0  0  0  0
 0  1  0  0  0

julia> reduce(hcat, [om, om])
10×10 Matrix{Bool}:
 0  0  0  0  0  0  0  0  0  0
 1  0  1  0  0  1  0  1  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  1  0  0  0  0  1  0
 0  0  0  0  1  0  0  0  0  1
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  1  0  0  0  0  1  0  0  0

julia> hcat(om, om)
10×10 Flux.OneHotArray{10,2,Vector{Int64}}:
 0  0  0  0  0  0  0  0  0  0
 1  0  1  0  0  1  0  1  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  0  0  1  0  0  0  0  1  0
 0  0  0  0  1  0  0  0  0  1
 0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0
 0  1  0  0  0  0  1  0  0  0
```

racinmat:
In flux 0.11, both `reduce(hcat, [om, om])` and `hcat(om, om)` return dense array, not onehot.

DhairyaLGandhi:
`hcat` does however work as expected with v0.12 and the mwe also works as expected with using tuples instead of the array. Further discussion in the linked PR of course.

CarloLucibello:
[...]

PR #1640, opened 2021-06-28
added support for reduce(hcat, OneHotMatrix)

pevnak:
[Please delete this text and describe your change here.
For bugfixes, please detail the bug and include a test case which your patch fixes.
If you are adding a new feature, please clearly describe the design, its rationale, the possible alternatives considered.
It is easiest to merge new features when there is clear precedent in other systems; we need to know we're taking
the right direction since it can be hard to change later.]

This fixes a bug
https://github.com/FluxML/Flux.jl/issues/1596
by adding `reduce(::typeof(hcat), Vector{TV}) where {TV<:OneHotLike}`

### PR Checklist

- [x ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -76,10 +76,12 @@ Base.hcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 2)
 Base.vcat(x::OneHotLike, xs::OneHotLike...) = cat(x, xs...; dims = 1)
 
 # optimized concatenation for matrices and vectors of same parameters
-Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} =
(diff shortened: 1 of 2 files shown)


pevnak:
I have not noticed Matej fixed it here
https://github.com/FluxML/Flux.jl/pull/1595

darsnack:
Actually the `reduce(hcat, ...)` case was removed from that PR before merging. The fix that we agreed upon was to define `similar` and `setindex!` for `OneHotArray`s (which should handle the `reduce` and other fast paths automatically).

There is an implementation of `similar` that I wrote in the comments for #1595 which will work. I never got around to `setindex!`, but a PR with those two changes would be appreciated.

tantheta01:
@darsnack wanted to check if I could work on this issue or has a PR been created already?

ToucheSir:
AFAIK nobody is working on it, but fair warning that the `setindex!` piece may be a little tricky!

PR #1641, opened 2021-06-29
Fix overview trivial but annoying problem

hannibalhuang:
This has been fix in the master branch however it still looks strange for first timer like myself, thus the proposal. Issue reported at #1642 

I have a lot of results  different from the tutorial as well, but given they are mostly 32 vs 64 problems, probably not really an issue per se.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


johnnychen94:
Almost all the contents in the gh-pages branches are generated using Documenter.jl; most julia packages use this tool to generate their documentation. Hence you should instead modify https://github.com/FluxML/Flux.jl/blob/master/docs/src/models/overview.md

hannibalhuang:
> Good catch.
> 
> Almost all the contents in the gh-pages branches are generated using [Documenter.jl](https://github.com/JuliaDocs/Documenter.jl); most julia packages use this tool to generate their documentation. Hence you should instead modify https://github.com/FluxML/Flux.jl/blob/master/docs/src/models/overview.md

An thx, let me submit the correct PR :)

darsnack:
Thanks for this! This bug has already been fixed (which is why it doesn't exist on master). You can check out the `dev` docs to see the fix. When the next version is released, the changes will be in the `stable` docs as well.

PR #1643, opened 2021-06-29
Replace Juno with ProgressLogging

DhairyaLGandhi:
cc @pfitzseb @vchuravy 

@pfitzseb what would be the equivalent to replace the rendering from Juno?


pfitzseb:
You could use https://github.com/pfitzseb/TreeViews.jl instead, but I'd recommend just deleting that code, tbh.

DhairyaLGandhi:
That assumes the length of the dataset is known already iiuc. Iteration does not.

darsnack:
True, but what would you log then? There are some issues about open ended logging on ProgressLogging.jl, but AFAIK the package only supports finite length logging.

darsnack:
Maybe we don't call `@logprogress` for unknown or infinite length iterators?

ToucheSir:
@DhairyaLGandhi this basically just needs a rebase, right?

ToucheSir:
If you'd like, I can do the cleanup work for this.

DhairyaLGandhi:
This would need a bit more work than a rebase. ProgressLogging isn't exactly compatible with Juno's, but as long as we have some fallbacks in place, we can merge this.

ToucheSir:
The only API we're switching over is `Juno.@progress` to `ProgressLogging.@progress`. For the 2 places the former is used, the latter should have the same behaviour.

DhairyaLGandhi:
superseded by #1807

DhairyaLGandhi:
There's no guarantee for a known length to the `data` arg in `train!`. In some cases, the termination will happen only once some accuracy criteria is met and more samples would be generated indefinitely

Issue #1465, opened 2021-01-14
Document conv data handling, especially for 1d 


AriMKatz:
From slack:

> So I the Flux docs are not great for non-image data w.r.t. Conv. If you want to do a 1D convolution, Conv expects a 3D input. This is more explicit in the NNlib docstring and the Flux one (https://fluxml.ai/Flux.jl/stable/models/nnlib/#Convolution-1).
> :point_up:
> 1
> 
> 
> Kyle Daruwalla  3 days ago
> As for the ordering of that 3D input…I’m not sure. I would expect it to be NCovariates x Ntime x NBatch, but the docs nor the source code does a great job of explaining that. Maybe @Brian Chen or @Andrew Dinhobl who have worked with time series data can chime in.
> :+1:
> 1
> 
> 
> Kyle Daruwalla  3 days ago
> Definitely, you shouldn’t have to conform your time series data to match WHCN format. There is some other format…I just don’t know what it is.
> 
> Brian Chen  3 days ago
> Yup, that's the way to go. I like to think of it as lopping the first dimension off WHCN, so channels and batch are still 2nd last and last respectively
> 
> Brian Chen  3 days ago
> So in this case it would be NTime x NCovariates X NBatch (which incidentally is the exact opposite of PyTorch because of batch first vs batch last)
> 
> Brian Chen  3 days ago
> I'm not sure if a spectrogram makes sense for your data, but if it can be represented that way you could make use of 2d convs
> 
> Jon Alm Eriksen  3 days ago
> @Brian Chen,  am I understanding you correctly in that if I provide a 3D tensor to Chain, with a filter tuple of one element, then it convolves along the first dimension only?
> :heavy_check_mark:
> 1
> 
> 
> Jon Alm Eriksen  3 days ago
> @Brian Chen Many thanks!
> 
> Jon Alm Eriksen  3 days ago
> You can think of my data as a specrogram-ish with complex data values, but I think that the 1D convolution is the only thing that makes sense
> 
> Jon Alm Eriksen  3 days ago
> the data is not translational invariant in the “frequency” direction
> 
> Jon Alm Eriksen  3 days ago
> Thanks all for helping me out!
> 
> Andrew Dinhobl  3 days ago
> I've been struggling with this as well recently. I'm also currently working on a 1-D convolution with (#vars x #timesteps x batchsize) and I used the unsqueeze thing to add an additional dimension for "channels" but still running into some issues.
> 
> Brian Chen  3 days ago
[...]

PR #1645, opened 2021-06-30
Update conv.jl

vivekkumar7089:
Specifying the format of the data given to the convolutional layer, as mentioned in #1465 .

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -40,7 +40,8 @@ Image data should be stored in WHCN order (width, height, channels, batch).
 In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
-a 2-tuple of integers.
+a 2-tuple of integers. For different `N`, data should be `N_1 x N_2 x ... x N_n x channels x batch`,
+i.e. `spatial/temporal dimensions x channels x batch`.
 
 For `N` spatial dimensions, this layer expects as input an array
 with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of channels.


DhairyaLGandhi:
LGTM

bors r+

src/layers/conv.jl
```diff
@@ -40,7 +40,8 @@ Image data should be stored in WHCN order (width, height, channels, batch).
 In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
-a 2-tuple of integers.
+a 2-tuple of integers. For different `N`, data should be `N_1 x N_2 x ... x N_n x channels x batch`,
+i.e. `spatial/temporal dimensions x channels x batch`.
```
DhairyaLGandhi, line 44:
```suggestion
i.e. `spatial/temporal dimensions x channels x batch` (notice that the batch is the last dimension).
```

src/layers/conv.jl
```diff
@@ -40,7 +40,8 @@ Image data should be stored in WHCN order (width, height, channels, batch).
 In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
-a 2-tuple of integers.
+a 2-tuple of integers. For different `N`, data should be `N_1 x N_2 x ... x N_n x channels x batch`,
```
ToucheSir, line 43: Thanks for the PR! A couple of comments:

1. "2-tuple" is obviously no longer correct when we're talking about convolutions with arbitrary dimensions.
2. I used `monospace` text in the original comment for convenience, but the dimensions ought to be properly latexified here. See https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d for reference.
3. Likewise, I wrote "spatial/temporal dimensions" as a shorthand that everyone in the issue thread would understand, but that language should be made more precise and elaborated upon for a docstring. For example, a video may have 2 spatial dimensions + 1 temporal dimension, whereas a 3d volume has 3 spatial dimensions. Ideally neither term should be required to explain the expect format of the kernel.

DhairyaLGandhi: I think describing it too specifically might hurt it.  Giving examples is fine, and I think the video example is good, but important to note that the examples can go beyond images or whatever and be generalised to regular feature arrays.

ToucheSir: Agreed. I should clarify that these examples are meant to help _write_ the docstring, and should not appear in the docstring itself. That should aim to be as domain agnostic as possible, hence my point about not using "spatial" or "temporal" unless needed.

vivekkumar7089: I think, we can give another example for 1D like already given for 2D (WHCN), and we will also mention that channels and batch size should be the 2nd last and last arguments, respectively. Is it fine @ToucheSir @DhairyaLGandhi ?

PR #1646, opened 2021-06-30
Update docstring for `Conv` to clarify feature dimensions

vivekkumar7089:
Specifying the format of the data given to the convolutional layer, as mentioned in #1465 .

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -42,7 +42,7 @@ and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
 a 2-tuple of integers.
 
-To take convolutions along `N` dimensions, this layer expects as input an array
+To take convolutions along `N` feature dimensions, this layer expects as input an array
 with `ndims(x) == N+2`, where `size(x, N+1) == in` is the number of input channels,
 and `size(x, ndims(x))` is (as always) the number of observations in a batch.
 Then:
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -41,7 +41,7 @@ In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
 a 2-tuple of integers. For different `N`, data should be `N_1 x N_2 x ... x N_n x channels x batch`,
-i.e. `spatial/temporal dimensions x channels x batch`.
+i.e. `spatial/temporal dimensions x channels x batch` (notice that the batch is the last dimension).
 
 For `N` spatial dimensions, this layer expects as input an array
 with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of channels.
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -40,11 +40,11 @@ Image data should be stored in WHCN order (width, height, channels, batch).
 In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
-a 2-tuple of integers. For different `N`, data should be `N_1 x N_2 x ... x N_n x channels x batch`,
-i.e. `spatial/temporal dimensions x channels x batch` (notice that the batch is the last dimension).
+a 2-tuple of integers.
 
-For `N` spatial dimensions, this layer expects as input an array
-with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of channels.
+For `N` dimensional data, this layer expects as input an array
+with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of input channels,
+and the last dimension is `size(x,N+2) == batches`.
 Then:
 * `filter` should be a tuple of `N` integers.
 * Keywords `stride` and `dilation` should each be either single integer,
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -43,8 +43,8 @@ This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
 a 2-tuple of integers.
 
 To take convolutions along `N` dimensions, this layer expects as input an array
-with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of input channels,
-and the last dimension is `size(x,N+2) == batches`.
+with `ndims(x) == N+2`, where `size(x, N+1) == in` is the number of input channels,
+and `size(x, ndims(x))` is (as always) the number of observations in a batch.
 Then:
 * `filter` should be a tuple of `N` integers.
 * Keywords `stride` and `dilation` should each be either single integer,
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -42,7 +42,7 @@ and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
 a 2-tuple of integers.
 
-For `N` dimensional data, this layer expects as input an array
+To take convolutions along `N` dimensions, this layer expects as input an array
 with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of input channels,
 and the last dimension is `size(x,N+2) == batches`.
 Then:
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -40,7 +40,8 @@ Image data should be stored in WHCN order (width, height, channels, batch).
 In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
-a 2-tuple of integers.
+a 2-tuple of integers. For different `N`, data should be `N_1 x N_2 x ... x N_n x channels x batch`,
+i.e. `spatial/temporal dimensions x channels x batch`.
 
 For `N` spatial dimensions, this layer expects as input an array
 with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of channels.


src/layers/conv.jl
```diff
@@ -40,7 +40,8 @@ Image data should be stored in WHCN order (width, height, channels, batch).
 In other words, a 100×100 RGB image would be a `100×100×3×1` array,
 and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
-a 2-tuple of integers.
+a 2-tuple of integers. For different `N`, data should be `N_1 x N_2 x ... x N_n x channels x batch`,
```
ToucheSir, line 43: Thanks for the PR! A couple of comments:

1. "2-tuple" is obviously no longer correct when we're talking about convolutions with arbitrary dimensions.
2. I used `monospace` text in the original comment for convenience, but the dimensions ought to be properly latexified here. See https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d for reference.
3. Likewise, I wrote "spatial/temporal dimensions" as a shorthand that everyone in the issue thread would understand, but that language should be made more precise and elaborated upon for a docstring. For example, a video may have 2 spatial dimensions + 1 temporal dimension, whereas a 3d volume has 3 spatial dimensions. Ideally neither term should be required to explain the expected format of the kernel.

ToucheSir (changes requested):
Copied over my comment from the closed PR.

src/layers/conv.jl
vivekkumar7089: Thanks for the review.
I will make the necessary changes.

mcabbott: The paragraph "In other words ... a 2-tuple of integers." is about the case of images. The next paragraph is about N dimensions. The logic being that perhaps it's better to see a concrete example before talking about `N+2` etc.

It's possible that there are better ways to organise this. But I don't think that tacking a sentence about the N-dim case onto the 2-dim paragraph is a good idea.

ToucheSir: Ah, I thought the line break was spurious and a new paragraph was being started. I do think any description of what `filter` should look like ought to be before examples though. Perhaps that first paragraph can be extended and the description of `in` and `out` can be moved around a bit. IMO having a consistent structure for listing and describing each parameter is important. Though informative, the current docstring is a bit of a dog's breakfast in that regard.

mcabbott: Yea, there may be complete re-writes that are better. I'm not sure that trying to explain the parameters in complete generality should happen before an example -- just one, the one most likely to be familiar.

vivekkumar7089: Can insert a below paragraph  
```
Input data should be stored in LCN (Length, Channels, Batch), WHCN (Width, Height, Channels, Batch)
and WHDCN (Width, Height, Depth, Channels, Batch) order for 1D, 2D and 3D convolution, respectively
(notice channels and batch are 2nd last and last dimensions respectively).
```
after the 
```
Standard convolutional layer. `filter` is a tuple of integers specifying the size of the convolutional kernel;
`in` and `out` specify the number of input and output channels.
```
and before
```
Image data should be stored in WHCN order (width, height, channels, batch). In other words, a
100×100 RGB image would be a `100×100×3×1` array, and a batch of 50 would be a `100×100×3×50`
array. This has `N=2` spatial dimensions and needs a kernel size like `(5,5)`, a 2-tuple of integers.
```
In this way, we can explain it in a more general manner and also, after this giving an example will clarify it to the reader.
For 3D Conv, D will specify the 3rd spatial dimension for a 3D volume, and also it can identify the 3rd temporal dimension for videos.

mcabbott: Pasting it all together somewhere, maybe in a gist, is probably clearer than trying to imagine what goes where. IMO you need to see and read the whole thing. And the temptation is always to try to put everything first, which pushes everything else later... 

But perhaps what would be helpful it to explain exactly what wrong impressions or confusions etc. people might have, so you and we know what you are trying to fix. I listed some but not sure these are your concerns or not.

vivekkumar7089:
@ToucheSir @mcabbott , In the convolutional layer definition `Conv(filter, in => out, σ=identity; stride=1, pad=0, dilation=1, [bias, init])`, in place of `filter`, using `kernel` or `kernel_size` may be a good idea as here in [https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html](url). Because in actual here also we are passing the kernel size for convolution.

src/layers/conv.jl
```diff
@@ -42,8 +42,9 @@ and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
 a 2-tuple of integers.
 
-For `N` spatial dimensions, this layer expects as input an array
-with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of channels.
+For `N` dimensional data, this layer expects as input an array
+with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of input channels,
```
mcabbott, line 46: The reason not to say "`N` dimensional data" is that normally that would mean `ndims(data)`. That's what "spatial" was trying to say. 

Can it be said some other way if you don't like spatial? Like, "To take convolutions encoding N translations" or better "To take convolutions along `N` dimensions, "?

vivekkumar7089: Yeah, it will be a better way. Just specifying _spatial_ is not a good idea because there may also be some _temporal_ dimensions.

ToucheSir: Well, spatial vs temporal is purely a function of the data modality. Conv doesn't care waht you call those, and so we shouldn't either outside of examples.

mcabbott: It's not quite as dire as you guys make out -- the convolution is concerned with translations in the vector space R^N, which is still called that even if N=1 and it's encoding time in the physical world. But the potential for confusion is sufficient reason to find another way to say it, and the change looks fine here.

src/layers/conv.jl
```diff
@@ -42,8 +42,9 @@ and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
 a 2-tuple of integers.
 
-For `N` spatial dimensions, this layer expects as input an array
-with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of channels.
+To take convolutions along `N` dimensions, this layer expects as input an array
+with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of input channels,
+and the last dimension is `size(x,N+2) == batches`.
```
mcabbott, line 47:
```suggestion
with `ndims(x) == N+2`, where `size(x, N+1) == in` is the number of input channels,
and `size(x, ndims(x))` is (as always) the number of observations in a batch.
```

src/layers/conv.jl
```diff
@@ -42,8 +42,9 @@ and a batch of 50 would be a `100×100×3×50` array.
...
 Then:
 * `filter` should be a tuple of `N` integers.
```
mcabbott, line 49: >  in place of `filter`, using `kernel` 

I agree that `filter` seems an odd word for this, it already means something else in Julia, and "kernel" seems more standard. I think I tried to change it (back?) to this at some point, but don't remember what objections were raised, but settled for saying "kernel size" in the text a few times. Could say it here too.

ToucheSir: I'm partial to kernel size as well. Just kernel is still quite jargony.

src/layers/conv.jl
ToucheSir: Yes, but for better or worse that's likely going to fly over the head of most people using the library. User #2320 trying to build an audio classifier is probably not thinking about vector spaces :smile:

src/layers/conv.jl
mcabbott: I guess "window" is another possibility, used in `MaxPool(window::NTuple; pad=0, stride=window)`. IMO nice to avoid variable names so long the need underscores, if possible.

src/layers/conv.jl
mcabbott: Sure. I think elsewhere we also call these "feature dimensions" as opposed to channel, batch (in MaxPool docstring).

I can't put a comment there, but there is another use of "spatial" in the example of a 1D conv, below: 
https://github.com/FluxML/Flux.jl/pull/1646/files#diff-bb37ab740622543577e6e853376328054b738a6aa01bfc44de2d223abfb25d45L110

ToucheSir: Feature dimensions is a much better term! I think there's definitely a bias towards "spatial" because of how much CV dominates CNN usage, but as someone who works with time series I personally object to that framing ;)

mcabbott: :) So perhaps it should say "To take convolutions along `N` feature dimensions, ..."

src/layers/conv.jl
ToucheSir: Well, PyTorch and TF are no help because they use `kernel_size` (PyTorch also uses it for pooling layers, while TF uses `pool_size`). I think this is one of those places where the increased aversion to underscores in Julia makes our life harder.

src/layers/conv.jl
```diff
@@ -42,8 +42,9 @@ and a batch of 50 would be a `100×100×3×50` array.
 This has `N=2` spatial dimensions, and needs a kernel size like `(5,5)`,
 a 2-tuple of integers.
 
-For `N` spatial dimensions, this layer expects as input an array
-with `ndims(x) == N+2`, where `size(x,N+1) == in` is the number of channels.
+To take convolutions along `N` dimensions, this layer expects as input an array
```
mcabbott, line 45: I think this is what we liked:
```suggestion
To take convolutions along `N` feature dimensions, this layer expects as input an array
```

src/layers/conv.jl
mcabbott: Besides jargon, the "kernel" is really the function you convolve with (here encoded in the weights) not its support.

But I do still like "window", aren't all sorts of things a bit like this described this way, like input for a moving average?

mcabbott (approved):
This looks fine. It doesn't change the use of "filter" but that's not a regression.

ToucheSir (approved):
Sounds good, let's get this merged.

mcabbott:
Thanks @vivekkumar7089 !

PR #1649, opened 2021-07-02
Update conv.jl

colbec:
Relax strict equality; mark some constants as Float32 to match existing values

PR #1650, opened 2021-07-02
Use correct eltype and rtol in CrossCor tests

ToucheSir:
Ref. https://github.com/FluxML/Flux.jl/pull/1649.

### PR Checklist

- [x] Tests are added
- [ ] ~~Entry in NEWS.md~~
- [ ] ~~Documentation, if applicable~~
- [ ] ~~API changes require approval from a committer (different from the author, if applicable)~~

### DIFF
--- a/test/layers/conv.jl
+++ b/test/layers/conv.jl
@@ -118,7 +118,7 @@ end
   w = rand(Float32, 2,2,1,1)
   y = CrossCor(w, [0.0])
 
-  @test isapprox(sum(w .* x[1:2, 1:2, :, :]), y(x)[1, 1, 1, 1], rtol=1e-7)
+  @test sum(w .* x[1:2, 1:2, :, :]) ≈ y(x)[1, 1, 1, 1]  rtol=1e-7
 
   r = zeros(Float32, 28, 28, 1, 5)
   m = Chain(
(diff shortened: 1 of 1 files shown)


darsnack (approved):

darsnack:
Let's see CI and then merge.

test/layers/conv.jl
```diff
@@ -131,7 +131,7 @@ end
 
   @test size(m(r)) == (10, 5)
   @test y(x) != Conv(w, [0.0])(x)
-  @test CrossCor(w[end:-1:1, end:-1:1, :, :], [0.0])(x) == Conv(w, [0.0])(x)
+  @test isapprox(CrossCor(w[end:-1:1, end:-1:1, :, :], [0.0])(x), Conv(w, [0.0])(x), rtol=1e-7)
```
DhairyaLGandhi, line 134: Use the unicode symbol here for consistency. Note that you can still use the kwargs - see some tests that do that for ex in the normalisation layer tests

DhairyaLGandhi (commented):
Left a small comment, else lgtm

test/layers/conv.jl
ToucheSir: I used `isapprox` here because the other test in the suite used it. Happy to swap both over to the operator, looks like we need to do a cleanup pass in other files as well: https://github.com/FluxML/Flux.jl/blob/8a5b9774f0d6b254f60c88ca07a984997a016e75/test/layers/normalisation.jl#L73

DhairyaLGandhi: Pretty much. Thanks for looking into it!

ToucheSir: Done!

DhairyaLGandhi (approved):
Thanks

PR #1654, opened 2021-07-08
add Functors docs

### LINKED ISSUE #1653: Missing docs for @functor
The links to [Flux.@functor](https://fluxml.ai/Flux.jl/stable/models/advanced/models/@ref) appear to be broken, such as [here](https://fluxml.ai/Flux.jl/stable/models/advanced/#Custom-multiple-input-or-output-layer). I can't tell where the link is supposed to go, is it to [Functors.jl](https://github.com/FluxML/Functors.jl)?

DhairyaLGandhi:
Close #1653

### DIFF
--- a/docs/src/models/functors.md
+++ b/docs/src/models/functors.md
@@ -1,0 +1,10 @@
+## Functors.jl
+
+Flux makes use of the Functors.jl to represent many of the core functionalities it provides.
+
+```@docs
+Functors.isleaf
+Functors.children
+Functors.fcollect
+Functors.fmap
+```
--- a/docs/src/models/functors.md
+++ b/docs/src/models/functors.md
@@ -8,5 +8,6 @@ Functors.jl is a collection of tools designed to represent a [functor](https://e
 Functors.isleaf
 Functors.children
 Functors.fcollect
+Functors.functor
 Functors.fmap
 ```
--- a/docs/src/models/functors.md
+++ b/docs/src/models/functors.md
@@ -2,6 +2,8 @@
 
 Flux makes use of the Functors.jl to represent many of the core functionalities it provides.
 
+Functors.jl is a collection of tools designed to represent a [functor](https://en.wikipedia.org/wiki/Functor_(functional_programming)). Flux makes use of it to treat certain structs as functors. Notable examples include the layers that Flux defines.
+
 ```@docs
 Functors.isleaf
 Functors.children
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -1,5 +1,6 @@
 [deps]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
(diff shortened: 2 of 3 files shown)


docs/src/models/functors.md
```diff
@@ -0,0 +1,10 @@
+## Functors.jl
+
+Flux makes use of the Functors.jl to represent many of the core functionalities it provides.
```
CarloLucibello, line 3:
```suggestion
Flux makes use of the Functors.jl to perform recursive introspection into layers and models.
```

docs/src/models/functors.md
```diff
@@ -0,0 +1,10 @@
...
+
+```@docs
+Functors.isleaf
+Functors.children
+Functors.fcollect
+Functors.fmap
```
darsnack, line 9:
```suggestion
Functors.fmap
Functors.fmapstructure
```

docs/src/models/functors.md
darsnack, line 3:
```suggestion
Functors.jl allows for recursive traversal of deeply nested structures, making it possible to treat a complex struct like any other collection in Julia.

It is used extensively throughout [Flux.jl](https://github.com/FluxML/Flux.jl).
```

darsnack (changes requested):

docs/src/models/functors.md
DhairyaLGandhi: I am coming around to the notion that `fmapstructure` isn't a real API at all. If anything it can be put in the utils (with a different name) here and be removed from Functors.

docs/src/models/functors.md
DhairyaLGandhi: If we want to give details about Functors.jl I would rather say that it is a set of utility functions designed to help with treating certain safe structs as Functors. Notably ones include most of the layers that Flux defines.

Traversal/ nesting is strictly orthogonal.

darsnack: Okay then say that, but we should say what Functors.jl is instead of just that another package uses it.

docs/src/models/functors.md
darsnack: Then remove it in another PR, but if it's exported and has a docstring, then it should be in the docs.

docs/src/models/functors.md
ToucheSir: Alternatively, give Functors its own standalone docs site like Zygote and only talk about the Flux relevant parts here. After all, you can use both it and dependents like Optimisers without ever touching Flux.

darsnack: ....wow I totally thought this was a PR to Functors.jl this entire time

darsnack: (and I agree that Functors.jl should have its own docs site)

You can ignore my comments here.

DhairyaLGandhi: https://github.com/FluxML/Functors.jl/pull/17  does that too

darsnack (approved):

PR #1656, opened 2021-07-09
update Embedding layer

manikyabard:
Updates the Embedding layer to use `gather` for `AbstractVector` which earlier had an issue with repeated indices. Also adds special case for `outputsize`.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -518,7 +518,7 @@ end
 
 @functor Embedding
 
-Embedding((in, out)::Pair{<:Integer, <:Integer}; init = randn32) = Embedding(init(out, in))
+Embedding(dims::Pair{<:Integer, <:Integer}; init = randn32) = Embedding(init(last(dims), first(dims)))
   
 (m::Embedding)(x::Integer) = m.weight[:, x]
 (m::Embedding)(x::AbstractVector) = NNlib.gather(m.weight, x)
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -508,7 +508,7 @@ julia> m([6, 15, 15])  # applied to a batch
  4.01  22.01  22.01
  5.01  23.01  23.01
 
-julia> ans == m(Flux.OneHotMatrix([6, 15, 15], 26))
+julia> ans == m(Flux.onehotbatch("foo", 'a':'z'))
 true
 ```
 """
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -521,13 +521,10 @@ end
 Embedding(dims::Pair{<:Integer, <:Integer}; init = randn32) = Embedding(init(last(dims), first(dims)))
   
 (m::Embedding)(x::Integer) = m.weight[:, x]
-(m::Embedding)(x::AbstractVector) = NNlib.gather(m.weight, x)
+(m::Embedding)(x::AbstractVector{<:Integer}) = NNlib.gather(m.weight, x)
 (m::Embedding)(x::AbstractArray) = reshape(m(vec(x)), :, size(x)...)
-
-function (m::Embedding)(x::Union{OneHotVector{T,L}, OneHotMatrix{T,L}}) where {T,L}
-  size(m.weight, 2) == L || throw(DimensionMismatch("Matrix column must correspond with OneHot size: $(size(m.weight, 2)) != $L"))
-  return m(onecold(x))
-end
+(m::Embedding)(x::AbstractArray{Bool}) = reshape(m(reshape(x, size(x, 1), :)), :, size(x)[2:end]...)
+(m::Embedding)(x::AbstractVecOrMat{Bool}) = m.weight * x  # handles OneHotLikeVector, OneHotLikeMatrix
  
 function Base.show(io::IO, m::Embedding)
   print(io, "Embedding($(size(m.weight, 2)) => $(size(m.weight, 1)))")
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -483,7 +483,8 @@ function Base.show(io::IO, m::Parallel)
 end
 
 """
-    Embedding(in => out; init=randn)
+    Embedding(in => out; init=randn32)
+    Embedding(weight::AbstractMatrix)
 
 A lookup table that stores embeddings of dimension `out` 
 for a vocabulary of size `in`.
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -521,7 +521,7 @@ end
 Embedding(dims::Pair{<:Integer, <:Integer}; init = randn32) = Embedding(init(last(dims), first(dims)))
   
 (m::Embedding)(x::Integer) = m.weight[:, x]
-(m::Embedding)(x::AbstractVector{<:Integer}) = NNlib.gather(m.weight, x)
+(m::Embedding)(x::AbstractVector) = NNlib.gather(m.weight, x)
 (m::Embedding)(x::AbstractArray) = reshape(m(vec(x)), :, size(x)...)
 (m::Embedding)(x::AbstractArray{Bool}) = reshape(m(reshape(x, size(x, 1), :)), :, size(x)[2:end]...)
 (m::Embedding)(x::AbstractVecOrMat{Bool}) = m.weight * x  # handles OneHotLikeVector, OneHotLikeMatrix
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -493,25 +493,26 @@ The input to the layer can be either a vector of indexes
 or the corresponding [onehot encoding](@ref Flux.OneHotArray). 
 
 # Examples
+
 ```jldoctest
-julia> vocab_size, embed_size = 1000, 4;
+julia> m = Embedding(reshape(-6:45, 2, 26) .+ 0.01f0)
+Embedding(26 => 2)
 
-julia> model = Flux.Embedding(vocab_size => embed_size)
-Embedding(1000 => 4)  # 4_000 parameters
+julia> m(5)  # embedding vector for 5th element
+2-element Vector{Float32}:
+ 2.01
+ 3.01
 
-julia> vocab_idxs = [1, 722, 53, 220, 3];
+julia> m([6, 15, 15])  # applied to a batch
+2×3 Matrix{Float32}:
+ 4.01  22.01  22.01
+ 5.01  23.01  23.01
 
-julia> x = Flux.OneHotMatrix(vocab_idxs, vocab_size); summary(x)
-"1000×5 OneHotMatrix(::Vector{Int64}) with eltype Bool"
-
-julia> model(x) |> summary
-"4×5 Matrix{Float32}"
-
-julia> model(vocab_idxs) == model(x)
+julia> ans == m(Flux.OneHotMatrix([6, 15, 15], 26))
 true
 ```
 """
-struct Embedding{W}
+struct Embedding{W <: AbstractMatrix}
   weight::W
 end
 
@@ -529,5 +530,5 @@ function (m::Embedding)(x::Union{OneHotVector{T,L}, OneHotMatrix{T,L}}) where {T
 end
  
 function Base.show(io::IO, m::Embedding)
-  print(io, "Embedding(", size(m.weight, 2), " => ", size(m.weight, 1), ")")
+  print(io, "Embedding($(size(m.weight, 2)) => $(size(m.weight, 1)))")
 end
--- a/src/outputsize.jl
+++ b/src/outputsize.jl
@@ -169,5 +169,4 @@ for (fn, Dims) in ((:conv, DenseConvDims), (:depthwiseconv, DepthwiseConvDims))
   end
 end
 
-(m::Embedding)(x::AbstractVector{<:Nil}) = fill(nil, size(m.weight, 1), length(x))
-(m::Embedding)(x::AbstractArray{<:Nil}) = fill(nil, size(m.weight, 1), size(x)...)
+(m::Embedding)(x::AbstractVecOrMat{<:Nil}) = fill(nil, size(m.weight, 1), length(x))
--- a/src/outputsize.jl
+++ b/src/outputsize.jl
@@ -168,3 +168,6 @@ for (fn, Dims) in ((:conv, DenseConvDims), (:depthwiseconv, DepthwiseConvDims))
     end
   end
 end
+
+(m::Embedding)(x::AbstractVector{<:Nil}) = fill(nil, size(m.weight, 1), length(x))
+(m::Embedding)(x::AbstractArray{<:Nil}) = fill(nil, size(m.weight, 1), size(x)...)
--- a/src/outputsize.jl
+++ b/src/outputsize.jl
@@ -169,4 +169,6 @@ for (fn, Dims) in ((:conv, DenseConvDims), (:depthwiseconv, DepthwiseConvDims))
   end
 end
 
-NNlib.gather!(dst::AbstractArray, ::AbstractArray, ::AbstractArray{<:Nil}) = fill(nil, size(dst)...)
+function NNlib.gather(src::AbstractArray{Tsrc, Nsrc}, idx::AbstractArray{<:Nil}) where {Tsrc, Nsrc}
+  fill(nil, (size(src)[1:Nsrc-1]..., size(idx)...))
+end
--- a/src/outputsize.jl
+++ b/src/outputsize.jl
@@ -169,4 +169,4 @@ for (fn, Dims) in ((:conv, DenseConvDims), (:depthwiseconv, DepthwiseConvDims))
   end
 end
 
-(m::Embedding)(x::AbstractVecOrMat{<:Nil}) = fill(nil, size(m.weight, 1), length(x))
+NNlib.gather!(dst::AbstractArray, ::AbstractArray, ::AbstractArray{<:Nil}) = fill(nil, size(dst)...)
--- a/test/cuda/layers.jl
+++ b/test/cuda/layers.jl
@@ -127,13 +127,13 @@ gpu_gradtest("PixelShuffle 2d", pixelshuffle, rand(Float32, 3, 4, 18, 3), 3)
 gpu_gradtest("PixelShuffle 1d", pixelshuffle, rand(Float32, 3, 18, 3), 3)
 
 embedding = [Flux.Embedding]
-gpu_gradtest("Embedding", embedding, [1,3,5], 5, 2)
-gpu_gradtest("Embedding repeated indices", embedding, [1,3,5,3], 5, 2)
-gpu_gradtest("Embedding integer index", embedding, 1, 5, 2)
-gpu_gradtest("Embedding 2d index", embedding, [1 2; 3 4], 5, 2)
-gpu_gradtest("Embedding OneHotVec index", embedding, OneHotVector(1, 5), 5, 2)
-gpu_gradtest("Embedding OneHotMatrix index", embedding,  OneHotMatrix([1,2,3], 5), 5, 2)
-gpu_gradtest("Embedding OneHotMatrix repeated indices", embedding, OneHotMatrix([1,2,2], 5), 5, 2)
+gpu_gradtest("Embedding", embedding, [1,3,5], 5 => 2)
+gpu_gradtest("Embedding repeated indices", embedding, [1,3,5,3], 5 => 2)
+gpu_gradtest("Embedding integer index", embedding, 1, 5 => 2)
+gpu_gradtest("Embedding 2d index", embedding, [1 2; 3 4], 5 => 2)
+gpu_gradtest("Embedding OneHotVec index", embedding, OneHotVector(1, 5), 5 => 2)
+gpu_gradtest("Embedding OneHotMatrix index", embedding,  OneHotMatrix([1,2,3], 5), 5 => 2)
+gpu_gradtest("Embedding OneHotMatrix repeated indices", embedding, OneHotMatrix([1,2,2], 5), 5 => 2)
 
 @testset "function layers" begin
   x = rand(Float32, 3,3)
--- a/test/cuda/layers.jl
+++ b/test/cuda/layers.jl
@@ -128,12 +128,12 @@ gpu_gradtest("PixelShuffle 1d", pixelshuffle, rand(Float32, 3, 18, 3), 3)
 
 embedding = [Flux.Embedding]
 gpu_gradtest("Embedding", embedding, [1,3,5], 5 => 2)
-gpu_gradtest("Embedding repeated indices", embedding, rand(1:50, 10^6), 50 => 2)
+gpu_gradtest("Embedding repeated indices", embedding, rand(1:10, 10^3), 10 => 2)
 gpu_gradtest("Embedding integer index", embedding, 1, 5 => 2)
 gpu_gradtest("Embedding 2d index", embedding, [1 2; 3 4], 5 => 2)
 gpu_gradtest("Embedding OneHotVec index", embedding, OneHotVector(1, 5), 5 => 2)
 gpu_gradtest("Embedding OneHotMatrix index", embedding,  OneHotMatrix([1,2,3], 5), 5 => 2)
-gpu_gradtest("Embedding OneHotMatrix repeated indices", embedding, OneHotMatrix(rand(1:50, 10^6), 50), 50 => 2)
+gpu_gradtest("Embedding OneHotMatrix repeated indices", embedding, OneHotMatrix(rand(1:10, 10^3), 10), 10 => 2)
 
 @testset "function layers" begin
   x = rand(Float32, 3,3)
--- a/test/cuda/layers.jl
+++ b/test/cuda/layers.jl
@@ -128,12 +128,12 @@ gpu_gradtest("PixelShuffle 1d", pixelshuffle, rand(Float32, 3, 18, 3), 3)
 
 embedding = [Flux.Embedding]
 gpu_gradtest("Embedding", embedding, [1,3,5], 5 => 2)
-gpu_gradtest("Embedding repeated indices", embedding, [1,3,5,3], 5 => 2)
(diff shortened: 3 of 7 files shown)


src/outputsize.jl
```diff
@@ -168,3 +168,6 @@ for (fn, Dims) in ((:conv, DenseConvDims), (:depthwiseconv, DepthwiseConvDims))
     end
   end
 end
+
+(m::Embedding)(x::AbstractVector{<:Nil}) = fill(nil, size(m.weight, 1), length(x))
+(m::Embedding)(x::AbstractArray{<:Nil}) = fill(nil, size(m.weight, 1), last(size(x)))
```
DhairyaLGandhi, line 173: These methods should not be here.

DhairyaLGandhi (changes requested):

src/outputsize.jl
darsnack: This is exactly where they should be (see the other code in the file).

src/layers/basic.jl
```diff
@@ -418,7 +418,8 @@ function Embedding(in::Integer, out::Integer;
 end
 
 (m::Embedding)(x::Union{OneHotVector, OneHotMatrix}) = m.weight * x # equivalent to m.weight[:,onecold(x)]
```
darsnack, line 420: This will hit all the fast paths. You'll need to define the following at L48 of `onehot.jl`.
```julia
const OneHotLikeVector{T, L} = OneHotLike{T, L, 0, 1, T}
const OneHotLikeMatrix{T, L, I} = OneHotLike{T, L, 1, 2, I}
```

```suggestion
(m::Embedding)(x::Union{OneHotLikeVector, OneHotLikeMatrix}) = m.weight * x # equivalent to m.weight[:,onecold(x)]
```

src/outputsize.jl
DhairyaLGandhi: I mean, these methods should not be defined at all. If output size cannot find the expected size, then it should be fixed there.

darsnack (changes requested):
We'll need some tests, but this looks good.

DhairyaLGandhi:
Needs tests

src/layers/basic.jl
DhairyaLGandhi: Why not use `::OneHotLike` ? Is there reason to constrain it to the two cases?

src/outputsize.jl
darsnack: `outputsize`'s underlying mechanism was designed for numerical array operations which hits ~95% case even with custom layers. There will always be some layers that we might need to special case. A layer that uses indexing as the underlying operation is non-sensical with `outputsize`. How would you know the result of indexing an array with `Nil`?

DhairyaLGandhi: You only need the size, so maybe outputsize should know what to do with getindex.

src/layers/basic.jl
darsnack: I think the one hot fast path is for vectors and matrices only? So we want to reshape (code a couple lines below) before this.

src/outputsize.jl
darsnack: But the output of `getindex` depends on how many items you pull out. What is `x[I]` when `I isa Vector{Nil}`?

darsnack: If we have a good answer for what the behavior should be, then I absolutely agree the proper fix is defining `getindex` for `outputsize`.

CarloLucibello:
ref. #1516

src/outputsize.jl
DhairyaLGandhi: Well `Vector{Nil}` sounds like the wrong kind of thing to have here, but the size would be size of the indices plus handling any colons in the getindex call

darsnack: That's correct, but also just not how `outputsize` works.

lorenzoh: In any case, the first method is redundant I think

CarloLucibello:
I rebase the ref. #1516 on master since there were some conflicts. So I guess you should do `git rebase -i origin/cl/embed` here

CarloLucibello:
Or maybe it is just easier if I merge #1516 and then this targets master

manikyabard:
> Or maybe it is just easier if I merge #1516 and then this targets master

Yeah sure that works

src/outputsize.jl
darsnack: Yeah actually I think the first should be `AbstractVecOrMat` and the second gets deleted.

mcabbott: I don't know whether it's useful, but the ordinary function above accepts any array. They should surely match.

src/layers/basic.jl
```diff
@@ -371,3 +372,56 @@ function Base.show(io::IO, m::Parallel)
...
+  0.91139    0.670462    0.463217   0.670462    0.110932
+  0.247225  -0.0823874   0.698694  -0.0823874   0.945958
+ -0.393626  -0.590136   -0.545422  -0.590136    0.77743
+ -0.497621   0.87595    -0.870251   0.87595    -0.772696
+```
+
+julia> model(vocab_idxs) == model(x)
+true
+"""
+struct Embedding{W}
```
mcabbott, line 409: I think this only makes sense for matrices? If so I think it should be restricted -- to document this, to give earlier errors, etc. If not it should be tested on other things. 
```suggestion
struct Embedding{W <: AbstractMatrix}
```

src/layers/basic.jl
```diff
@@ -371,3 +372,56 @@ function Base.show(io::IO, m::Parallel)
...
+
+function Embedding(in::Integer, out::Integer;
+               init = (i...) -> randn(Float32, i...))
+  return Embedding(init(out, in))
+end
+
+(m::Embedding)(x::Union{OneHotVector, OneHotMatrix}) = m.weight * x # equivalent to m.weight[:,onecold(x)]
+(m::Embedding)(x::Int) = m.weight[:, x]
+(m::Embedding)(x::AbstractVector) = NNlib.gather(m.weight, x)
+(m::Embedding)(x::AbstractArray) = reshape(m(vec(x)), :, size(x)...)
```
mcabbott, line 423: Doing this would mean you get errors sooner, not sure though that you get more readable errors.
```suggestion
(m::Embedding)(x::Integer) = m.weight[:, x]
(m::Embedding)(x::AbstractVector{<:Integer}) = NNlib.gather(m.weight, x)
(m::Embedding)(x::AbstractArray{<:Integer}) = reshape(m(vec(x)), :, size(x)...)
```

src/layers/basic.jl
```diff
@@ -371,3 +372,56 @@ function Base.show(io::IO, m::Parallel)
...
+
+function Base.show(io::IO, m::Embedding)
+  print(io, "Embedding($(size(m.weight, 2)), $(size(m.weight, 1)))")
```
mcabbott, line 426: Since this is new code, perhaps we should consider standardising on `=>` like `Conv` for channel-like dimensions? It's always just a little confusing that they are backwards compared to the size of matrices. 

```suggestion
function Base.show(io::IO, m::Embedding)
  print(io, "Embedding($(size(m.weight, 2)) => $(size(m.weight, 1)))")
```
(Change obviously needs constructor to change too.)

src/layers/basic.jl
```diff
@@ -371,3 +372,56 @@ function Base.show(io::IO, m::Parallel)
...
+julia> model(x)
+4×5 Matrix{Float32}:
+  0.91139    0.670462    0.463217   0.670462    0.110932
+  0.247225  -0.0823874   0.698694  -0.0823874   0.945958
+ -0.393626  -0.590136   -0.545422  -0.590136    0.77743
+ -0.497621   0.87595    -0.870251   0.87595    -0.772696
+```
+
+julia> model(vocab_idxs) == model(x)
+true
```
mcabbott, line 407: Maybe an example without random numbers, so that it can be a doctest:
````suggestion
```jldoctest
julia> m = Embedding(reshape(-6:45, 2, 26) .+ 0.01f0)
Embedding(26 => 2)

julia> m(5)  # embedding vector for 5th element
2-element Vector{Float32}:
 2.01
 3.01

julia> m([6, 15, 15])  # applied to a batch
2×3 Matrix{Float32}:
 4.01  22.01  22.01
 5.01  23.01  23.01

julia> ans == m(Flux.OneHotMatrix([6, 15, 15], 26))
true
```
````

File: src/outputsize.jl (hunk 1 above)
Comment (mcabbott) [reply src/outputsize.jl:173]: Good to specialise on OneHot, but what happens if by some chance you collect it? Best case is that this gives you the same result, it's just an optimisation. (Indexing things often treat an array of Bools differently to other integers.) 

Second best is for it to be an error, which is what I get now, but is this guaranteed? (Perhaps it is without offset arrays?)
```
julia> m(Flux.OneHotMatrix([6, 15, 15], 26))
2×3 Matrix{Float32}:
 4.01  22.01  22.01
 5.01  23.01  23.01

julia> m(collect(Flux.OneHotMatrix([6, 15, 15], 26)))
ERROR: BoundsError: attempt to access 2×26 Matrix{Float32} at index [1:2, 78-element Vector{Bool}]
```

Comment (darsnack) [reply src/outputsize.jl:173]: > I don't know whether it's useful, but the ordinary function above accepts any array. They should surely match.

Yeah, but my intent was the >2D hits the reshape route in the embedding layer source. We only need to specify the `Nil` path for what finally gets called which is `AbstractVecOrMat`.

> Good to specialise on OneHot, but what happens if by some chance you collect it?

This looks related to your other suggestions. We should constraint on `AbstractVector{<:Integer}` and `AbstractVecOrMat{<:Bool}` separately? In the latter case, it should be a matmul (mathematically) or `m.weight[:, idx]`?

File: src/layers/basic.jl (hunk 3 above)
Comment (manikyabard) [reply src/layers/basic.jl:409]: Yeah we probably should just restrict it to matrices.

File: src/layers/basic.jl (hunk 6 above)
Comment (manikyabard) [reply src/layers/basic.jl:407]: Yeah this makes sense.

File: src/layers/basic.jl (hunk 5 above)
Comment (manikyabard) [reply src/layers/basic.jl:426]: Sure, we can do that.

**Comment by mcabbott** [comment]
Maybe labels other than integers starting at 1 deserve some thought. Flux's one-hot encoding lets you specify the set of labels, which are not stored, `onecold` returns integers starting at 1. 
```
julia> Flux.onehotbatch(collect("hello"), 'a':'z') 
26×5 Flux.OneHotArray{26,2,Vector{UInt32}}:
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 0  1  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 1  0  0  0  0
 0  0  0  0  0
 0  0  0  0  0
 ⋮           

julia> Flux.onecold(ans) # does not remember
5-element Vector{Int64}:
  8
  5
 12
 12
 15
```
Should this do something similar? Then `m = Embedding(0:9 => 5)` would have to store something, so that `m(0)` can give the first vector. Or is this unnecessary complication?

**Comment by darsnack** [comment]
Mapping to an indices or one-hot space is a standard data transformation for using a layer like `Embedding`. So I would say it's not a necessary feature for the layer. And we would always need to support 1:N which could get tricky if the "labels" are also an integer range.

If we did want to include it, I would store a "transform" function within the layer.

**Comment by mcabbott** [comment]
Ok, indeed this should probably be written something like `Chain(onehot('a':'z'), Embedding(26 => 5))` rather than duplicating this.

File: src/outputsize.jl (hunk 1 above)
Comment (manikyabard) [reply src/outputsize.jl:173]: Would something like this make more sense for `AbstractArray`?
```julia
(m::Embedding)(x::AbstractArray{<:Nil}) = fill(nil, size(m.weight, 1), size(x)...)
```

darsnack:
Should this be re-targeted for master?

CarloLucibello:
yes. Probably filing a new PR is easier

manikyabard:
I changed the base branch to master.

darsnack:
Looks like you need a rebase too

src/layers/basic.jl
```diff
@@ -428,3 +429,50 @@ function Base.show(io::IO, m::Parallel)
...
+"""
+struct Embedding{W <: AbstractMatrix}
+  weight::W
+end
+
+@functor Embedding
+
+Embedding(dims::Pair{<:Integer, <:Integer}; init = randn32) = Embedding(init(last(dims), first(dims)))
+  
+(m::Embedding)(x::Union{OneHotVector, OneHotMatrix}) = m.weight * x # equivalent to m.weight[:,onecold(x)]
```
mcabbott, line 471: This is where I think treating `collect(onehot(...))` the same as `onehot()` would be nice. They are both arrays of booleans, so something like this may work:
```suggestion
(m::Embedding)(x::AbstractArray{Bool}) = m.weight * x  # handles OneHotVector, OneHotMatrix
```
(Or it may require a bit more thought which is more specific.)

src/layers/basic.jl
```diff
@@ -428,3 +429,50 @@ function Base.show(io::IO, m::Parallel)
...
+(m::Embedding)(x::Integer) = m([x])
+(m::Embedding)(x::AbstractVector) = NNlib.gather(m.weight, x)
+(m::Embedding)(x::AbstractArray) = reshape(m(vec(x)), :, size(x)...)
```
mcabbott, line 474: I would also restrict these to integers
```suggestion
(m::Embedding)(x::AbstractVector{<:Integer}) = NNlib.gather(m.weight, x)
(m::Embedding)(x::AbstractArray{<:Integer}) = reshape(m(vec(x)), :, size(x)...)
```

test/layers/basic.jl
```diff
@@ -191,4 +191,29 @@ import Flux: activations
...
+    @test size(m.weight) == (embed_size, vocab_size)
+    
+    x = rand(1:vocab_size, 3)
+    y = m(x)
+    @test y isa Matrix{Float32}
+    @test y ≈ m.weight[:,x]
+    x2 = OneHotMatrix(x, vocab_size)
+    y2 = m(x2)
+    @test y2 isa Matrix{Float32}
+    @test y2 ≈ y
```
mcabbott, line 207: I think collecting the OneHot thing would ideally never change the answer. Maybe the test is like this:
```suggestion
    @test m(x2) isa Matrix{Float32}
    @test m(x2) ≈ y
    @test m(collect(x2)) ≈ y
```

src/layers/basic.jl
```diff
@@ -428,3 +429,50 @@ function Base.show(io::IO, m::Parallel)
...
+2-element Vector{Float32}:
+ 2.01
+ 3.01
+
+julia> m([6, 15, 15])  # applied to a batch
+2×3 Matrix{Float32}:
+ 4.01  22.01  22.01
+ 5.01  23.01  23.01
+
+julia> ans == m(Flux.OneHotMatrix([6, 15, 15], 26))
```
mcabbott, line 459: My fault, but I think the public function here is the lower-case onehotbatch. This could be
```suggestion
julia> ans == m(Flux.onehotbatch([6, 15, 15], 1:26))
```
or else, according to taste,
```suggestion
julia> ans == m(Flux.onehotbatch(['f', 'o', 'o'], 'a':'z'))
```

test/cuda/layers.jl
```diff
@@ -259,3 +259,39 @@ end
...
+
+  gs = gradient(() -> sum(tanh.(m(x))), params(m))
+  gs_g = gradient(() -> sum(tanh.(m_g(x_g))), params(m_g))
+  @test collect(gs_g[m_g.weight]) ≈ gs[m.weight]
+
+  @testset "repeated indexes" begin
+    vocab_size, embed_size = 5, 2
+    m = Flux.Embedding(vocab_size, embed_size)
+
+    x = [1, 3, 5, 3] # repeated indexes
```
mcabbott, line 286: Good to test, I'd be tempted to suggest much larger numbers here, like `x = rand(1:5, 10^6);`, to ensure there's enough work that it's very likely to be parallelised.

src/layers/basic.jl
darsnack: I guess it comes down to whether the `Vector{Bool}` case should be a matmul or index operation.

src/outputsize.jl
darsnack: Personally, I would keep the `Nil` overrides to a minimum and only do the vector case. Let the rest go through the normal routing for all arrays.

src/layers/basic.jl
darsnack: Maybe only for `AbstractVector` so that `Nil`-arrays can go through the reshape?

src/layers/basic.jl
mcabbott: Those should agree when it's actually one-hot (if not OneHot) up to a dropdims:
```
julia> [1 2 3; 4 5 6] * [false, true, false]
2-element Vector{Int64}:
 2
 5

julia> [1 2 3; 4 5 6][:, [false, true, false]]
2×1 Matrix{Int64}:
 2
 5
```
But not when it's not:
```
julia> [1 2 3; 4 5 6] * [false, true, true]
2-element Vector{Int64}:
  5
 11

julia> [1 2 3; 4 5 6][:, [false, true, true]]
2×2 Matrix{Int64}:
 2  3
 5  6
```
Should the latter case be an error? Maybe indexing would be faster although I'm not sure we care about optimising this. Logical indexing can't give you aliasing problems, if I'm thinking correctly.

src/layers/basic.jl
mcabbott: Oh I wasn't thinking about Nil, good point. Do you want Array{Nil} to behave like Array{Int} or like Array{Bool}? 

One other weird feature is that if you make a OneHotTensor, like a boolean 3-array, this will reshape it to a vector. Maybe that's too weird to worry about. Or maybe it should reshape `AbstractArray{Bool}` not with vec but with `m(reshape(size(x,1), :))`

src/layers/basic.jl
darsnack: I was mostly speaking from an optimization stand point. If we don't care about this case being fast (it is rare), then the suggested code works for me.

src/layers/basic.jl
darsnack: Yes, I think you are correct. Reshaping should be different when the first dimension encodes one-hot information. This definitely presents a problem for `Nil`, because the `Nil`-array is agnostic of whether the original input is one-hot or indices. In other words, the output size is input type dependent for this layer.

src/layers/basic.jl
manikyabard: Would it be better to change the current onehot paths from 
```julia
function (m::Embedding)(x::Union{OneHotLikeVector{T,L}, OneHotLikeMatrix{T,L,I}}) where {T,L,I}
    size(m.weight, 2) == L || throw(DimensionMismatch("Matrix column must correspond with OneHot size: $(size(m.weight, 2)) != $L"))
  return m(onecold(x))
end
``` 
back to a matmul?

darsnack: Yes definitely, the current implementation is just code duplication.

src/layers/basic.jl
manikyabard: In that case, is the only way for getting the correct output from `outputsize` just to ignore the first dimension of the one-hot input while passing in the `inputsizes` argument?

darsnack:
@manikyabard are you still up for rebasing this and moving forward? We ought to close up this loose end.

manikyabard:
> @manikyabard are you still up for rebasing this and moving forward? We ought to close up this loose end.

Yeah I can continue working on this, although I am not sure about the approach we should take. Maybe we can discuss this further in the next community call.

darsnack:
> Maybe we can discuss this further in the next community call.

Yeah, sounds good!

darsnack:
Summarizing what was discussed on the call:
- the issue is that the paths for `AbstractVector{<:Integer}` that do `getindex` calls don't make sense since `Vector{Nil}` is not a vector of indices
- Michael's PR to [make `Nil` a subtype of `Real`](https://github.com/FluxML/Flux.jl/pull/1830) can help avoid the cases where `Nil` should only be hitting the `AbstractArray{<:Real}` paths
- it still won't solve the issue when `outputsize` is used for a model utilizing the `AbstractVector{<:Integer}` path
    - this will need an `outputsize` override rule for `Embedding` + `AbstractVector{Nil}`
    - (this wasn't brought up during call but I just thought of it) a better solution would be to define `NNlib.gather` for `Nil` which will cover all indexing cases beyond just `Embedding`

manikyabard:
>   * (this wasn't brought up during call but I just thought of it) a better solution would be to define `NNlib.gather` for `Nil` which will cover all indexing cases beyond just `Embedding`

You mean something like this?
```Julia
NNlib.gather!(dst::AbstractArray, ::AbstractArray, ::AbstractArray{<:Nil}) = fill(nil, size(dst)...)
(m::Embedding)(x::AbstractVector{<:Nil}) = NNlib.gather(m.weight, x)
```

darsnack:
That looks right but you won't need to special case for `Embedding` anymore. It should go through the `NNlib.gather` rule automatically.

CarloLucibello:
this needs a rebase in master

darsnack:
Can you add a test for `outputsize` of `Embedding`?

CarloLucibello (changes requested):

src/layers/basic.jl
```diff
@@ -476,7 +476,7 @@ function Base.show(io::IO, m::Parallel)
 end
 
 """
-    Embedding(in, out; init=randn)
+    Embedding(in => out; init=randn)
```
mcabbott, line 479:
```suggestion
    Embedding(in => out; init=randn32)
    Embedding(weights::AbstractMatrix)
```

src/layers/basic.jl
```diff
@@ -487,47 +487,39 @@ or the corresponding [onehot encoding](@ref Flux.OneHotArray).
...
-julia> using Flux: Embedding
-
-julia> vocab_size, embed_size = 1000, 4;
-
-julia> model = Embedding(vocab_size, embed_size)
-Embedding(1000, 4)
-
-julia> vocab_idxs = [1, 722, 53, 220, 3]
+```jldoctest
+julia> m = Embedding(reshape(-6:45, 2, 26) .+ 0.01f0)
```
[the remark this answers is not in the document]
mcabbott: Indeed on the constructor. 

The virtue of this example is that it doesn't have random numbers, so it can be a doctest.

src/layers/basic.jl
```diff
@@ -487,47 +487,39 @@ or the corresponding [onehot encoding](@ref Flux.OneHotArray).
...
 """
-struct Embedding{W}
+struct Embedding{W <: AbstractMatrix}
   weight::W
 end
 
 @functor Embedding
 
-Embedding(in::Integer, out::Integer; init = randn32) = Embedding(init(out, in))
+Embedding(dims::Pair{<:Integer, <:Integer}; init = randn32) = Embedding(init(last(dims), first(dims)))
```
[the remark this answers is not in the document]
mcabbott: Ninja'd.

test/cuda/layers.jl
```diff
@@ -127,13 +127,13 @@ gpu_gradtest("PixelShuffle 2d", pixelshuffle, rand(Float32, 3, 4, 18, 3), 3)
...
 embedding = [Flux.Embedding]
-gpu_gradtest("Embedding", embedding, [1,3,5], 5, 2)
-gpu_gradtest("Embedding repeated indices", embedding, [1,3,5,3], 5, 2)
-gpu_gradtest("Embedding integer index", embedding, 1, 5, 2)
-gpu_gradtest("Embedding 2d index", embedding, [1 2; 3 4], 5, 2)
-gpu_gradtest("Embedding OneHotVec index", embedding, OneHotVector(1, 5), 5, 2)
-gpu_gradtest("Embedding OneHotMatrix index", embedding,  OneHotMatrix([1,2,3], 5), 5, 2)
-gpu_gradtest("Embedding OneHotMatrix repeated indices", embedding, OneHotMatrix([1,2,2], 5), 5, 2)
+gpu_gradtest("Embedding", embedding, [1,3,5], 5 => 2)
+gpu_gradtest("Embedding repeated indices", embedding, rand(1:50, 10^6), 50 => 2)
```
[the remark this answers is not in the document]
mcabbott: 10^6 seems quite big, but maybe there's a some value to bigger than 3, in case e.g. 3 is never parallelised. No idea if there's a particular cutoff.

src/outputsize.jl
```diff
@@ -168,3 +168,5 @@ for (fn, Dims) in ((:conv, DenseConvDims), (:depthwiseconv, DepthwiseConvDims))
     end
   end
 end
+
+NNlib.gather!(dst::AbstractArray, ::AbstractArray, ::AbstractArray{<:Nil}) = fill(nil, size(dst)...)
```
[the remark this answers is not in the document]
darsnack: Probably, we can modify the non-mutating variation: `NNlib.gather` which is what Flux layers will end up calling.

src/layers/basic.jl
```diff
@@ -487,47 +487,39 @@ or the corresponding [onehot encoding](@ref Flux.OneHotArray).
...
   
 
 (m::Embedding)(x::Integer) = m.weight[:, x]
 (m::Embedding)(x::AbstractVector) = NNlib.gather(m.weight, x)
 (m::Embedding)(x::AbstractArray) = reshape(m(vec(x)), :, size(x)...)
```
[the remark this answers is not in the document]
darsnack: This is what we had before (and we can still have it). The idea behind overriding `gather` instead is that anywhere we have indexing, we recommend using `gather` to be GPU-compatible. So, overriding `gather` will allow more custom layers with indexing to work with `outputsize`.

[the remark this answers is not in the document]
manikyabard: I have updated `gather` to work with Nil as of now. We can however not go this route, and just make this work only for Embedding layer.

test/cuda/layers.jl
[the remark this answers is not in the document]
manikyabard: I have updated it to be `10^3`. Would this also add a lot of time to CI?

CarloLucibello:

PR #1648, opened 2021-07-02
Use NNlibCUDA v0.1.5 to allow CUDA v3.3.1

DhairyaLGandhi:
Adds a lower bound to NNlibCUDA so it can pull in CUDA v3.3.1

Will release once CI passes.

PR #1657, opened 2021-07-10
remove Manifest

CarloLucibello:
let's be free of the Manifest


mcabbott (dismissed):
bors r+

mcabbott:
Failure might be new to this PR. It involves
```
[4] ctc_alpha(ŷ::CuArray{Float64, 2}, y::Vector{Int64})
  @ Flux.Losses /var/lib/buildkite-agent/builds/rtx4000-gpuci4-julia-csail-mit-edu/julialang/flux-dot-jl/src/losses/ctc-gpu.jl:211
```
at `/test/ctc-gpu.jl:26`. Looks odd to call this with a Vector & a CuArray, but seemingly worked before? Paging @maetshju 
```
@testset "ctc-gpu" begin
  x = rand(10, 50)
  y = rand(1:9, 30)
  x_cu = CuArray(x)
  g1 = gradient(ctc_loss, x_cu, y)[1]
```

CarloLucibello:
yeah I'm seeing the same failure in #1516

mcabbott:
Oh right. Slightly different stack trace but this line appears. CUDA v3.2.1 & NNlibCUDA v0.1.3, which is what the manifest there specifies.

CarloLucibello:
Even removing the ctc test I see those errors everywhere. cc @maleadt

DhairyaLGandhi:
Errors are due to be cuda and are tracked in #1648

mcabbott:
Ok, I'll update the title

CarloLucibello:
> Even removing the ctc test I see those errors everywhere. cc @maleadt

actually, that's wrong, there was a problem with my cuda setup. 

I fixed the `ctc_loss`, it shouldn't be doing scalar indexing now and tests should pass

test/outputsize.jl
```diff
@@ -10,6 +10,7 @@
   @test outputsize(m, (10,); padbatch=true) == (2, 1)
   @test outputsize(m, (10, 30)) == (2, 30)
 
+  @info "Don't mind the following error, it's for testing purpose."
```
mcabbott, line 13: 👍

CarloLucibello:
@mcabbott need an approval again

mcabbott (approved):
bors r+

PR #1658, opened 2021-07-10
Print channel dimensions of `Dense` like those of `Conv`

mcabbott:
This makes it print `Dense(5 => 2)` to match e.g. `Conv((3,3), 5 => 2)`, and accept this as input. Then we consistently print channel dimensions one way, and it's a little less confusing that this is `Dense(randn(2, 5))`. 

It does not make `Dense(5, 2)` into a depwarn, because that feels like more of an imposition, to make everyone change everything. It doesn't seem too confusing to just quietly accept both. 

Since the pretty-printing of `Chain` etc. will already break people's doctests, doing this in the same release is the least annoying time to do it.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -99,6 +99,7 @@ extraChain(::Tuple{}, x) = ()
 
 """
     Dense(in => out, σ=identity; bias=true, init=glorot_uniform)
+    Dense(in, out, [σ; keywords...])
     Dense(W::AbstractMatrix, [bias, σ])
 
 Create a traditional `Dense` layer, whose forward pass is given by:
@@ -117,7 +118,7 @@ The weight matrix and/or the bias vector (of length `out`) may also be provided 
 
 # Examples
 ```jldoctest
-julia> d = Dense(5, 2)
+julia> d = Dense(5 => 2)
 Dense(5 => 2)       # 12 parameters
 
 julia> d(rand(Float32, 5, 64)) |> size
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -301,7 +301,7 @@ function Base.show(io::IO, b::SkipConnection)
 end
 
 """
-    Bilinear(in1, in2, out, σ=identity; bias=true, init=glorot_uniform)
+    Bilinear((in1, in2) => out, σ=identity; bias=true, init=glorot_uniform)
     Bilinear(W::AbstractArray, [bias, σ])
 
 Creates a Bilinear layer, which operates on two inputs at the same time.
@@ -311,12 +311,14 @@ for all `i ∈ 1:out`:
     z[i] = σ(x' * W[i,:,:] * y + bias[i])
 
 If `x` and `y` are matrices, then each column of the output `z = B(x, y)` is of this form,
-with `B` a Bilinear layer.
+with `B` the Bilinear layer.
 
-If `y` is not given, it is taken to be equal to `x`, i.e. `B(x) == B(x, x)`
+If the second input `y` is not given, it is taken to be equal to `x`, i.e. `B(x) == B(x, x)`
 
 The two inputs may also be provided as a tuple, `B((x, y)) == B(x, y)`,
 which is accepted as the input to a `Chain`.
+
+If the two input sizes are the same, `in1 == in2`, then you may write `Bilinear(in => out, σ)`.
 
 The initialisation works as for [`Dense`](@ref) layer, with `W = init(out, in1, in2)`.
 By default the bias vector is `zeros(Float32, out)`, option `bias=false` will switch off
@@ -326,7 +328,8 @@ trainable bias. Either of these may be provided explicitly.
 ```jldoctest
 julia> x, y = randn(Float32, 5, 32), randn(Float32, 5, 32);
 
-julia> B = Flux.Bilinear(5, 5, 7);
+julia> B = Flux.Bilinear((5, 5) => 7)
+Bilinear(5 => 7)    # 182 parameters
 
 julia> B(x) |> size  # interactions based on one input
 (7, 32)
@@ -335,15 +338,15 @@ julia> B(x,y) == B((x,y))  # two inputs, may be given as a tuple
 true
 
 julia> sc = SkipConnection(
-                Chain(Dense(5, 20, tanh), Dense(20, 9, tanh)),
-                Flux.Bilinear(9, 5, 3, bias=false),
+                Chain(Dense(5 => 20, tanh), Dense(20, 9, tanh)),
+                Flux.Bilinear((9, 5) => 3, bias=false),
             );  # used as the recombinator, with skip as the second input
 
 julia> sc(x) |> size
 (3, 32)
 
 julia> Flux.Bilinear(rand(4,8,16), false, tanh)  # first dim of weight is the output
-Bilinear(8, 16, 4, tanh, bias=false)
+Bilinear((8, 16) => 4, tanh; bias=false)  # 512 parameters
 ```
 """
 struct Bilinear{F,A,B}
@@ -363,6 +366,9 @@ function Bilinear(in1::Integer, in2::Integer, out::Integer, σ = identity;
                   init = glorot_uniform, bias = true)
   Bilinear(init(out, in1, in2), bias, σ)
 end
+
+Bilinear(((in1, in2), out)::Pair{<:Tuple, <:Integer}, σ = identity; kw...) = Bilinear(in1, in2, out, σ; kw...)
+Bilinear((in12, out)::Pair{<:Integer, <:Integer}, σ = identity; kw...) = Bilinear(in12, in12, out, σ; kw...)
 
 function (a::Bilinear)(x::AbstractMatrix, y::AbstractMatrix)
   W, b, σ = a.weight, a.bias, a.σ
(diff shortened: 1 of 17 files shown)


DhairyaLGandhi:
Are there cases where such an API is an improvement? It seems odd to model a fully connected layer with a Conv layer.

ToucheSir:
I don't see how this is modeling a fully connected layer with a conv one? If anything, it's being more consistent about how we handle/display fixed-size input and output dimensions.

mcabbott:
Yes to consistency. I changed `Bilinear` too, where I think it also clarifies what the 3 integers mean (and lets you omit one without it being confusing -- like you can omit one input to mean `y = x`.) If we do this, then presumably `RNN`, `LSTM`, `GRU` should also change.

darsnack:
For an API, I think `In => Out` is much more clear that two numbers with a comma (we should support both constructors though).

Mathematically, I think it makes sense to think of the output dimension of `Dense` layers as analogous to convolution output channels. For example, this similarity came up during the orthogonal initialization (you could see my comment there for explanation).

CarloLucibello:
yes `Dense(2 => 5)` is clearer than `Dense(2, 5)`.
(e.g. years ago I was of the opinion that Dense(2, 5) should be changed to behave the other way around and be represented by a 2x5 matrix).  
The only problem is that the new constructors will require a little extra typing. Also `=>` add a bit of visual noise, but this is a personal opinion. 
It is also true that it doesn't take much to get used to the ordering of `Dense(2, 5)`. 
All-in-all I'm not fully convinced this PR is worth doing, but not opposed either.

src/layers/basic.jl
```diff
@@ -69,7 +69,7 @@ extraChain(::Tuple{}, x) = ()
 
 
 """
-    Dense(in, out, σ=identity; bias=true, init=glorot_uniform)
+    Dense(in => out, σ=identity; bias=true, init=glorot_uniform)
```
CarloLucibello, line 72: The intention here is to phase out of the doc the old constructor?
That would be a bit confusing if the old constructor is not deprecated

DhairyaLGandhi:
I don't see an explanation on where the current constructor is failing. And conflating layers is a slippery slope in API design, since you quickly end up with making it non-intuitive for most layers. There is also clear relationship between the two numbers, so it's not like the current situation makes it hard to parse what is happening. Are there popular examples of this kind of API?

darsnack:
Something doesn't need to be failing for us to make something better.

> And conflating layers is a slippery slope in API design, since you quickly end up with making it non-intuitive for most layers.

There isn't any conflation here. It always means the same thing: size in to size out.

> There is also clear relationship between the two numbers, so it's not like the current situation makes it hard to parse what is happening.

If you never read the Flux docs for `Dense` or you are new to ML, then what is the clear relationship between `(2, 5)`? One might easily expect it to be the size of `weight` but that would be wrong! Conversely, why is `Conv((3, 3), 3, 16)` not as intuitive as `Conv((3, 3), 3 => 16)`? And `Bilinear((3, 5) => 10)` is so much more intuitive than `Bilinear(3, 5, 10)`.

It's pretty easy to support both. I agree it's not a groundbreaking change, but a small convenience we can do in Julia that other frameworks can't.

src/layers/basic.jl
mcabbott: Haven't figured out the best approach to docstrings. It's probably confusing to leave something working without mentioning it at all. It's also confusing to explain why you support two forms.

Maybe it's less weird to just actually mark it depreciated. Even if there's no intent to delete it in one version's time. The depwarn message could say this. 

My concern was about imposing work on everyone who has tests running which include Flux. It's some effort to go and update things.

darsnack: Why not just keep both (in the code and docstring header)? In actual doc examples, we can prefer the new notation.

DhairyaLGandhi:
> Conversely, why is Conv((3, 3), 3, 16) not as intuitive as Conv((3, 3), 3 => 16)

But that is precisely my point. What makes sense for one layer, shouldn't be shoehorned into another. When one sees the `Dense` call it is already known that the user has indeed expected a `Dense` *layer*. That is where the relationship between the args is determined.

darsnack:
It's not being shoehorned...it still makes perfect sense. You've italicized "layer" but I don't understand the implications of that.

src/layers/basic.jl
mcabbott: I pushed a change doing that. Doesn't look too bad, maybe that's the right thing to do.

ToucheSir:
Put another way, I would be in full support of this PR even if Conv didn't use the => notation for channels.

CarloLucibello:
I think I will probably continue using `Dense(3, 2)`, since those 2 extra characters in `Dense(3 => 2)` require a little bit more finger gymnastics that probably I won't be willing to exert. Do you people see yourselves using this syntax in your code? 
Because if we are doing this only to make the construction less confusing for beginners, well, that confusion lasts only a minute and doesn't come back so better not give it too much weight

DhairyaLGandhi:
Yeah, the constructors are decent already and anyone who wants a layer output would also see the api in a number of ways anyway, which in itself is clear enough.

ToucheSir:
I'm less interested in the construction aspect than the display, which is the primary goal of this PR. Programmatic construction of layers may also become marginally easier, but that's a snall thing at best.

darsnack:
I'd probably use `Dense(3 => 2)` cause I'm not a fast typer anyways. I prefer the extra readability both in code and when displayed in the REPL.

I feel like these kinds of changes are always partially subjective. It's a marginal improvement, but it also doesn't seem to have any downsides. In the entire history of Flux, we've never wanted to use this syntax for something AFAIK.

DhairyaLGandhi:
Aren't Conv layers counter examples? I also worry that printing isn't a good enough motivation for adding an extra set of constructors. Also, `Dense(in, out)` also seems to support the "Dense layer taking `in` incoming features to `out` outgoing features" sentiment that the proposal is going after.

darsnack:
> Aren't Conv layers counter examples?

I would say they are inline examples (not counter). `Conv` is just the one layer that happened to already have a constructor of this form.

> Also, Dense(in, out) also seems to support the "Dense layer taking in incoming features to out outgoing features" sentiment that the proposal is going after.

Not really, `(in, out)` implies no directionality other than reading order. It could just as easily be `Dense(out, in)` (which is the shape of `weight` too so not an unreasonable conclusion).

mcabbott:
> It could just as easily be Dense(out, in)

Right. Once you've read it a million times you just know. But I do think it's kind-of confusing on first approach. If you type `Dense(randn(20, 30))` on day 1 at the REPL, and it prints `Dense(30 => 20)`, that seems a helpful hint. The numbers in the `=>` are, here and in Conv, and in Bilinear, the dimensions which are completely fixed, while image size & batch size can vary without changing the layer. 

I agree it's more annoying to type. I sincerely hope that typing out Dense layers' sizes isn't the bottleneck in anyone's day. Those who've done this a million times will continue to type the comma.

Yes it's subjective, communication always is. And yes to "marginal improvement", it's obviously not going to change the world. I made it now because, if anyone has doctests, they are going to have to update them anyway, so this is the minimally annoying time to change the printing.

darsnack:
> If you type Dense(randn(20, 30)) on day 1 at the REPL, and it prints Dense(30 => 20), that seems a helpful hint.

This is a really great point, as a plain tuple, `(in, out)` is flipped when specifying a weight matrix manually as size `(out, in)`. All the more reason we should be more explicit about ordering in the dims-based constructor.

ToucheSir:
Moreover, PyTorch is the only mainstream DL library that uses this `in, out` argument list. c.f. [TF](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dense), [CNTK](https://cntk.azurewebsites.net/pythondocs/layerref.html#dense), [MXNet](https://mxnet.apache.org/versions/1.8.0/api/python/docs/api/gluon/nn/index.html#mxnet.gluon.nn.Dense) and a couple [Jax](https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.Dense.html) [Frameworks](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.Linear).

mcabbott:
Thanks for the list, that would take me hours. Some of those look *really* annoying to type.

ToucheSir:
Well, all of them basically boil down to `Dense(out_features; ...)` because of shape inference. Thus side-stepping the need to distinguish between input and output dimensionalities. PyTorch has now followed suit with https://pytorch.org/docs/1.9.0/generated/torch.nn.LazyLinear.html?highlight=lazylinear#torch.nn.LazyLinear, but I think we can do better than both in Flux by using the Pair syntax and sprinkling in some metaprogramming magic for `outputsize`-driven shape inference.

CarloLucibello:
I'm neutral (or at most slightly pro) to this, but since there seems to be so much support I'm going to approve

CarloLucibello (approved):

src/layers/basic.jl
```diff
@@ -283,7 +287,7 @@ function Base.show(io::IO, b::SkipConnection)
 end
 
 """
-    Bilinear(in1, in2, out, σ=identity; bias=true, init=glorot_uniform)
+    Bilinear((in1, in2) => out, σ=identity; bias=true, init=glorot_uniform)
```
CarloLucibello, line 290: old constructor should be kept as well since we don't deprecate it

CarloLucibello:
I'll state the reasons for which I'm not 100% convinced (some of them discussed already)
- extra typing
- visual cluttering
- julia beginners could be unaware of the `Pair`s' syntax and of the Pair concept itself
- unless we deprecate the old constructor, we essentially create 2 different slang in the codebase

mcabbott:
Shall we do this, v0.13?

Latest version moves the old constructors explicitly to `deprecations.jl`, although it does not make them noisy, and we should probably leave them like for quite a long time, so that old things just work. But it seems too confusing to document both as official ways.

CarloLucibello:
I think I got increasingly more fond of this

darsnack (approved):
LGTM!

Issue #1519, opened 2021-02-25
Document OneHotArray


cossio:
It seems that the new `OneHotArray` (https://github.com/FluxML/Flux.jl/pull/1448) is not documented. There is nothing mentioned in https://fluxml.ai/Flux.jl/dev/data/onehot/ and it's not clear how one should use `OneHotArray`. Can we add some documentation? @darsnack

darsnack:
Yup the docs for the old `onehotbatch` etc. should be there, but I'll add more docs around the new type.

PR #1660, opened 2021-07-11
Printing & docstrings for `onehot` / `onehotbatch`

mcabbott:
Right now, the printing of OneHotArray lies about its type parameters. That's pretty confusing. The standard way to hide messy details the way `view` and `adjoint` do is via `Base.showarg`, so I did that. Then I also re-used the dots which LinearAlgebra's sparse matrix printing uses: 
```
julia> Flux.onehotbatch(collect("foo"), 'a':'z')  # before
26×3 Flux.OneHotArray{26,2,Vector{UInt32}}:
 0  0  0
 0  0  0
 0  0  0
 0  0  0
 0  0  0
 1  0  0
 0  0  0
 ...
 
julia> typeof(ans)
Flux.OneHotArray{UInt32, 26, 1, 2, Vector{UInt32}}

julia> Flux.onehotbatch(collect("foo"), 'a':'z')  # after
26×3 OneHotMatrix(::Vector{UInt32}) with eltype Bool:
 ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅
 1  ⋅  ⋅
 ⋅  ⋅  ⋅
```
I've also tried to tidy up things I thought were unclear in the docstrings. E.g. it looked `unk...` was indicating that you could specify multiple defaults, but in fact the splat is just an implementation trick. And following `Base.get` perhaps it's called `default`.

Should have no functional changes at all.

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -57,16 +57,15 @@ end
 
 # this is from /LinearAlgebra/src/diagonal.jl, official way to print the dots:
 function Base.replace_in_print_matrix(x::OneHotLike, i::Integer, j::Integer, s::AbstractString)
-    # CUDA.@allowscalar(x[i,j]) ? s : _isonehot(x) ? Base.replace_with_centered_mark(s) : s
     x[i,j] ? s : _isonehot(x) ? Base.replace_with_centered_mark(s) : s
 end
 function Base.replace_in_print_matrix(x::LinearAlgebra.AdjOrTrans{Bool, <:OneHotLike}, i::Integer, j::Integer, s::AbstractString)
-    CUDA.@allowscalar(x[i,j]) ? s : _isonehot(parent(x)) ? Base.replace_with_centered_mark(s) : s
+    x[i,j] ? s : _isonehot(parent(x)) ? Base.replace_with_centered_mark(s) : s
 end
 
-# Base.show(io::IO, x::OneHotLike) = show(io, convert(Array{Bool}, cpu(x))) # helps string(cu(y))
-
 Base.print_array(io::IO, X::OneHotLike{T, L, N, var"N+1", <:CuArray}) where {T, L, N, var"N+1"} = 
+  Base.print_array(io, cpu(X))
+Base.print_array(io::IO, X::LinearAlgebra.AdjOrTrans{Bool, <:OneHotLike{T, L, N, var"N+1", <:CuArray}}) where {T, L, N, var"N+1"} = 
   Base.print_array(io, cpu(X))
 
 _onehot_bool_type(x::OneHotLike{<:Any, <:Any, <:Any, N, <:Union{Integer, AbstractArray}}) where N = Array{Bool, N}
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -1,6 +1,12 @@
 import Adapt
 import .CUDA
 
+"""
+    OneHotArray{T,L,N,M,I} <: AbstractArray{Bool,M}
+
+These are constructed by [`onehot`](@ref) and [`onehotbatch`](@ref).
+Parameter `I` is the type of the underlying storage, and `T` its eltype.
+"""
 struct OneHotArray{T<:Integer, L, N, var"N+1", I<:Union{T, AbstractArray{T, N}}} <: AbstractArray{Bool, var"N+1"}
   indices::I
 end
@@ -15,7 +21,9 @@ _indices(x::Base.ReshapedArray{<: Any, <: Any, <: OneHotArray}) =
 const OneHotVector{T, L} = OneHotArray{T, L, 0, 1, T}
 const OneHotMatrix{T, L, I} = OneHotArray{T, L, 1, 2, I}
 
+@doc @doc(OneHotArray)
 OneHotVector(idx, L) = OneHotArray(idx, L)
+@doc @doc(OneHotArray)
 OneHotMatrix(indices, L) = OneHotArray(indices, L)
 
 # use this type so reshaped arrays hit fast paths
@@ -49,11 +57,17 @@ end
 
 # this is from /LinearAlgebra/src/diagonal.jl, official way to print the dots:
 function Base.replace_in_print_matrix(x::OneHotLike, i::Integer, j::Integer, s::AbstractString)
-    CUDA.@allowscalar(x[i,j]) ? s : _isonehot(x) ? Base.replace_with_centered_mark(s) : s
+    # CUDA.@allowscalar(x[i,j]) ? s : _isonehot(x) ? Base.replace_with_centered_mark(s) : s
+    x[i,j] ? s : _isonehot(x) ? Base.replace_with_centered_mark(s) : s
 end
 function Base.replace_in_print_matrix(x::LinearAlgebra.AdjOrTrans{Bool, <:OneHotLike}, i::Integer, j::Integer, s::AbstractString)
     CUDA.@allowscalar(x[i,j]) ? s : _isonehot(parent(x)) ? Base.replace_with_centered_mark(s) : s
 end
+
+# Base.show(io::IO, x::OneHotLike) = show(io, convert(Array{Bool}, cpu(x))) # helps string(cu(y))
+
+Base.print_array(io::IO, X::OneHotLike{T, L, N, var"N+1", <:CuArray}) where {T, L, N, var"N+1"} = 
+  Base.print_array(io, cpu(X))
 
 _onehot_bool_type(x::OneHotLike{<:Any, <:Any, <:Any, N, <:Union{Integer, AbstractArray}}) where N = Array{Bool, N}
 _onehot_bool_type(x::OneHotLike{<:Any, <:Any, <:Any, N, <:CuArray}) where N = CuArray{Bool, N}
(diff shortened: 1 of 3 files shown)


darsnack:
In general, I think this PR is great. I just want to note that the previous complex show code (as opposed to using the default `AbstractArray` show) existed for `OneHotArray`s on the GPU. We'll just want to make sure that works here.

darsnack:
There should be a test for GPU show already too

mcabbott:
Great, was just fiddling with this on a GPU'd machine, I think it now works:
```
julia> Flux.onehotbatch(collect("foo"), 'a':'z')' |> cu
3×26 adjoint(OneHotMatrix(::CuArray{UInt32, 1})) with eltype Bool:
 ⋅  ⋅  ⋅  ⋅  ⋅  1  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  1  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅
 ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  1  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅  ⋅

julia> reshape(ans, 6, :)
6×13 reshape(adjoint(OneHotMatrix(::CuArray{UInt32, 1})), 6, 13) with eltype Bool:
 0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  1  0  0  0  0  0
 0  0  0  0  0  0  0  1  0  0  0  0  0
 0  0  1  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0
```
Will check for this test, too.

src/onehot.jl
```diff
@@ -61,6 +47,27 @@ Base.getindex(x::OneHotArray{<:Any, L}, ::Colon, I...) where L = OneHotArray(x.i
...
+    return nothing
+end
+
+# this is from /LinearAlgebra/src/diagonal.jl, official way to print the dots:
+function Base.replace_in_print_matrix(x::OneHotLike, i::Integer, j::Integer, s::AbstractString)
+    x[i,j] ? s : _isonehot(x) ? Base.replace_with_centered_mark(s) : s
+end
+function Base.replace_in_print_matrix(x::LinearAlgebra.AdjOrTrans{Bool, <:OneHotLike}, i::Integer, j::Integer, s::AbstractString)
+    x[i,j] ? s : _isonehot(parent(x)) ? Base.replace_with_centered_mark(s) : s
+end
```
darsnack, line 64: My concern here is that the transpose of a one-hot array is no longer one-hot. Would this be confusing?

src/onehot.jl
```diff
@@ -95,72 +102,105 @@ Base.argmax(x::OneHotLike; dims = Colon()) =
     invoke(argmax, Tuple{AbstractArray}, x; dims = dims)
 
 """
-    onehot(l, labels[, unk])
+    onehot(x, labels, [default])
+
+Return a `OneHotVector` which is roughly a sparse representation of `x .== labels`.
```
darsnack, line 107:
```suggestion
Return a `OneHotVector` which compactly stores the result `x .== labels`.
```

src/onehot.jl
```diff
@@ -95,72 +102,105 @@ Base.argmax(x::OneHotLike; dims = Colon()) =
...
 
-Return a `OneHotVector` where only first occourence of `l` in `labels` is `1` and
-all other elements are `0`.
+Instead of storing say `Vector{Bool}`, it stores the index of the first occourence 
+of `x` in `labels`. If `x` is not found in labels, then it either returns `onehot(default, labels)`,
+or gives an error if no default is given.
```
darsnack, line 111:
```suggestion
Instead of storing the full vector `x .== labels`, it stores the index of the first occurrence 
of `x` in `labels`. If `x` is not found in labels, then it either returns `onehot(default, labels)`,
or gives an error if no default is given.
```

src/onehot.jl
```diff
@@ -95,72 +102,105 @@ Base.argmax(x::OneHotLike; dims = Colon()) =
...
 
-If `l` is not found in labels and  `unk` is present, the function returns
-`onehot(unk, labels)`; otherwise the function raises an error.
+See also [`onehotbatch`](@ref) to apply this to many `x`s, 
+and [`onecold`](@ref) for a `labels`-aware `argmax`.
```
darsnack, line 114:
```suggestion
and [`onecold`](@ref) to reverse this operation.
```

src/onehot.jl
```diff
@@ -95,72 +102,105 @@ Base.argmax(x::OneHotLike; dims = Colon()) =
...
+  i > 0 || return onehot(default, labels)
   OneHotVector{UInt32, length(labels)}(i)
 end
 
 """
-    onehotbatch(ls, labels[, unk...])
+    onehotbatch(xs, labels, [default])
+
+Returns a `OneHotMatrix` where `k`th column of the matrix is [`onehot(ls[k], labels)`](@ref onehot),
+for vector `xs`. This is a sparse matrix, which stores just `Vector{UInt32}` containing the indices.
```
darsnack, line 150:
```suggestion
Returns a `OneHotMatrix` where `k`th column of the matrix is [`onehot(xs[k], labels)`](@ref onehot).
This is a sparse matrix, which stores just a `Vector{UInt32}` containing the indices.
```

src/onehot.jl
```diff
@@ -95,72 +102,105 @@ Base.argmax(x::OneHotLike; dims = Colon()) =
...
+
+If one of the inputs in `xs` is not found in `labels`, that column is `onehot(default, labels)`
+if `default` is given, else an error.
 
-Return a `OneHotMatrix` where `k`th column of the matrix is `onehot(ls[k], labels)`.
+If `xs` is a matrix, then the result is an `AbstractArray{Bool, 3}` which is one-hot along
+the first dimension, i.e. `result[:, k...] == onehot(xs[k...], labels)`.
```
darsnack, line 156: This isn't really a common case, right? We should either re-phrase this for N-dimensional or just leave it out.

src/onehot.jl
```diff
@@ -95,72 +102,105 @@ Base.argmax(x::OneHotLike; dims = Colon()) =
...
 
-If one of the input labels `ls` is not found in `labels` and `unk` is given,
-return [`onehot(unk, labels)`](@ref) ; otherwise the function will raise an error.
+Matrix multiplication `M * onehotbatch(...)` is performed efficiently, by simply getting
+one element from every row of `M`. Some concatenation and reshape operations preserve onehot-ness.
+`OneHotArray`s can be moved to the GPU, to get for instance `OneHotMatrix(::CuArray{UInt32, 1})`.
```
darsnack, line 160: This seems like it shouldn't be in this docstring. Probably should go in the actual docs (and close #1519).

src/onehot.jl
```diff
@@ -95,72 +102,105 @@ Base.argmax(x::OneHotLike; dims = Colon()) =
...
+ 2  5  14  2  8  2  11  2  5  14  2
+ 3  6  15  3  9  3  12  3  6  15  3
 ```
 """
-onehotbatch(ls, labels, unk...) = batch([onehot(l, labels, unk...) for l in ls])
+onehotbatch(ls, labels, default...) = batch([onehot(l, labels, default...) for l in ls])
 
 """
-    onecold(y[, labels = 1:length(y)])
+    onecold(y, [labels])
```
darsnack, line 182:
```suggestion
    onecold(y, [labels = 1:size(y, 1)])
```

darsnack (commented):
Looking good, mostly just wording comments. We might be able to close #1519 with this too.

src/onehot.jl
mcabbott: It's a nod towards the N-dimensional case, I guess. They are probably all uncommon, so I think it makes sense to lead with the case of a OnehotMatrix, not with the generic one. I don't think it should be left out completely.

If you like it could say "If xs has more than one dimension, ..."?

src/onehot.jl
mcabbott: It's not exactly this, though. I wondered whether to write `x .== unique(vec(labels))` but that's also not exactly right... So I was hoping to give a helpful example before giving the details.

src/onehot.jl
mcabbott: Isn't it most useful not as the inverse, but applied to other matrices? The name is a cute inversion but that's not all.

src/onehot.jl
mcabbott: This is true. When I wrote this, I think I mis-read the `*` overloads as including `onehot' * dense`, but they don't actually define that.

src/onehot.jl
mcabbott: OK, perhaps you have a vision for more extensive docs. My tendency is always to put things in docstrings as they are easy to look up piecemeal & tend to get updated. For now I've removed this in favour of a comment on the example showing matmul.

src/onehot.jl
mcabbott: Done

src/onehot.jl
mcabbott: Now removed

mcabbott: I thought the other case used to catch this, but seems not to... anyway:
```
julia> reshape(hcat(αβγ...), 3,3,1)
3×3×1 reshape(OneHotMatrix(::Vector{UInt32}), 3, 3, 1) with eltype Bool:
[:, :, 1] =
 1  0  0
 0  1  0
 0  0  1

julia> reshape(hcat(αβγ...), 3,3,1) |> Flux._isonehot
true

julia> reshape(hcat(αβγ...), 3,3,1) isa Flux.OneHotLike
true
```

mcabbott: Oh I guess it's taking a slice, and those don't preserve it:
```
julia> reshape(hcat(αβγ...), 3,3,1)[:,:,1]
3×3 Matrix{Bool}:
 1  0  0
 0  1  0
 0  0  1

julia> @view reshape(hcat(αβγ...), 3,3,1)[:,:,1]
3×3 view(reshape(OneHotMatrix(::Vector{UInt32}), 3, 3, 1), :, :, 1) with eltype Bool:
 1  0  0
 0  1  0
 0  0  1

julia> Flux._isonehot(ans)
ERROR: MethodError: no method matching _isonehot(::SubArray
```

src/onehot.jl
mcabbott: My vote is to say this once, just below. (Partly because I'm never super-sure whether an `=` in the docstring signature means a keyword or a default.)

ToucheSir: Are we consistent about sequestering kwargs after a `;`? If so, I think there shouldn't be too much ambiguity.

mcabbott: Yea I don't mean actual ambiguity in the computer sense. I just mean in the plain-language sense, of quickly giving the right impression, and deciding which paragraph to put how much detail in to be minimally confusing.

src/onehot.jl
darsnack: "Return the `OneHotVector` representation of `x` given `labels`."

?

Then just explain in more detail below like you already are.

src/onehot.jl
darsnack: I've always used it as an inverse not as necessarily as an argmax. I just thought "`labels`-aware `argmax`" was confusing. It make sense if you intuitively understand the `onecold` function already but makes you double take otherwise.

src/onehot.jl
darsnack: Personally, I think the docstring headers should be very specific and not require reading the paragraph. I don't mind saying it once, though my vote is for in the header.

src/onehot.jl
mcabbott: My hope is to get across what it does, if you don't yet know what "one-hot" means.

src/onehot.jl
mcabbott: Neither is a complete description, I just think the concept that it's roughly the reverse is what the name already conveys, so any further description should be other facts. Could say nothing and just leave the link, I guess.

src/onehot.jl
mcabbott: Maybe it should still not have `[ ]`, since that's another way of saying that there exists a default. (Maybe that's part of what strikes me as keyword-ish.)
```suggestion
    onecold(y, labels = 1:size(y,1))
```

src/onehot.jl
mcabbott:
```suggestion
and [`onecold`](@ref) to reverse either of these, as well as to generalise `argmax`.
```

CarloLucibello (approved):

mcabbott:
Let's merge this and come back to any refinements in #1519. 

bors r+

PR #1661, opened 2021-07-11
Deprecate `Flux.zeros`

mcabbott:
Seems like a footgun to have a function with the same name & almost the same function as Base's. It seems to have been used inconsistently, with some functions defining their own closure to be less confusing. This gives it a new name, `zeros32`, and uses it everywhere. Ditto `ones32`.

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -18,3 +18,6 @@ function Base.getproperty(a::Dense, s::Symbol)
   end
   return getfield(a, s)
 end
+
+@deprecate zeros(arg...) zeros32(arg...) false  # don't export these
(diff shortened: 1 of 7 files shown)


darsnack (approved):

darsnack:
These would be breaking, right? Let's discuss how to handle 0.13 separately from bump patches. Historically, the idea of releasing a breaking version seems to block a lot of useful patches.

mcabbott:
If I've done what I meant to do, merging this should not alter any results, but will sometimes print warnings. Only in tests, I think. So I think that ought to be no problem to do right now, right? 

Deleting what's in deprecations.jl certainly can't be done before 0.13. But no great rush. Could even be left there until say 0.14.

Issue #1664, opened 2021-07-14
Allow user to change the channel axis for `BatchNorm` function and the likes


vboussange:
Hi all, 
Currently, the `BatchNorm` function selects by default the `N-1` dimension of the `N` dimensional array as the channel dimension. I think it would be useful to let the user decide on the dimension that corresponds to the channel dimension in his/her particular setting.

I have already a proposal, that I am happy to release as a pull request.

# Example
```julia
using Flux
channel_size = 3
channel_axis = 1
BN = BatchNorm(channel_size, dim = channel_axis)
x = randn(channel_size, 10, 10)
BN(x)
```

DhairyaLGandhi:
Thanks! We already support multiple dims in the norm layers for different data types. The question really is that in order to have a channel axis different in one set of layers compared to all the others in Flux. It might cause confusion and increase complexity of the package more generally.

ToucheSir:
Sounds like a reasonable (if somewhat uncommon) proposal to me if it is extended to GroupNorm and InstanceNorm as well. The main consideration I can think of would be how GPU-friendly this is.

vboussange:
Thanks for your answers.
@ToucheSir This is an option proposed in TensorFlow with [tf.keras.layers.BatchNormalization](https://www.tensorflow.org/api_docs/python/tf/keras/layers/BatchNormalization). It can be useful for scientific machine learning.
I am writing [a package](https://github.com/vboussange/HighDimPDE.jl) (paper in prep.) that relies on Flux.jl for solving high dimensional Partial Differential Equations. In this setting, I need to evaluate `BN` with an array `x` where `ndims(x)` is 2 or 3, but the "feature" axis is always the first one.
As regards to GPU performance, it works very well in my user case. 

I agree, this option should be also implemented for `GroupNorm` and `InstanceNorm`, which is what I did. Let me know if you want me to PR.

darsnack:
If you have code, a PR is always better to discuss a possible change.

This is not meant to be push back, but is using `permutedims` before the BN not an option? Is it just too cumbersome to write?

DhairyaLGandhi:
Well, something like this would need to be consistent through Conv*, Dense* etc too, not just *Norm for it to be used consistently in a model.

ToucheSir:
[...]

PR #1666, opened 2021-07-16
Allow user to change the channel axis for BatchNorm function and the likes

vboussange:
Proposition for issue #1664

This PR allows the user to customize the channel axis for normalisation functions (`BatchNorm`,` GroupNorm` and `InstanceNorm`).

## Example

```Julia
using Flux
channel_size = 3
channel_axis = 1
BN = BatchNorm(channel_size, dim = channel_axis)
x = randn(channel_size, 10, 10)
BN(x)
```

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -167,9 +167,8 @@ end
 # Compute the statistics on the slices specified by reduce_dims.
 # reduce_dims=[1,...,N-2,N] for BatchNorm
 # reduce_dims=[1,...,N-2] for InstanceNorm and GroupNorm
-function _norm_layer_forward(l, x::AbstractArray{T,N}; reduce_dims, affine_shape, dim=nothing) where {T, N}
-  # todo:change
-  isnothing(dim) ? dim = N-1 : nothing
+function _norm_layer_forward(l, x::AbstractArray{T,N}; reduce_dims, affine_shape) where {T, N}
+  isnothing(l.dim) ? dim = N-1 : dim = l.dim
   if !_isactive(l) && l.track_stats # testmode with tracked stats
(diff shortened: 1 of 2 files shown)


src/layers/normalise.jl
```diff
@@ -243,9 +245,11 @@ mutable struct BatchNorm{F,V,N,W}
   track_stats::Bool
   active::Union{Bool, Nothing}
   chs::Int # number of channels
+  dim::Union{Int, Nothing} # channel dimension
```
DhairyaLGandhi, line 248: This should never be `nothing`, but use `N - 1` as default.

ToucheSir: How do you know `N - 1` a priori though? The question is what is an appropriate sentinel value. Perhaps a negative offset from the end, since by default channels are at dim `end - 1`?

DhairyaLGandhi: Right so, it will need to be defined at runtime.

DhairyaLGandhi:
How well is cudnn able to handle this?

ToucheSir:
Looping back to answer the cuDNN question: it supports one other configuration for channel dim via `CUDNN_TENSOR_NHWC`. I think adding that to NNlibCUDA at https://github.com/FluxML/NNlibCUDA.jl/blob/96a334633ef3a3707c85fc1754c2c7eb8849db4e/src/cudnn/batchnorm.jl#L27 would be a good first step for getting this going again. Here are a couple example implementation PRs from [MXNet](https://github.com/apache/incubator-mxnet/pull/20615) and [PyTorch](https://github.com/pytorch/pytorch/pull/29361).

CarloLucibello:

PR #1670, opened 2021-07-20
fix deprecations of zeros

DhairyaLGandhi:
The deprecations were incorrect

cc @ChrisRackauckas

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -21,8 +21,9 @@ end
 
 function ones(dims...)
   Base.depwarn("Flux.ones(size...) is deprecated, please use Flux.ones32(size...) or Base.ones(Float32, size...)", :ones)
+  Base.ones(Float32, dims...)
 end
-ones(T::Type, dims...) = Base.ones(T, dims...) # no need for a message
+ones(T::Type, dims...) = Base.ones(T, dims...)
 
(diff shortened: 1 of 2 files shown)


DhairyaLGandhi:
Fixes the case like https://github.com/SciML/DiffEqFlux.jl/pull/592

src/deprecations.jl
```diff
@@ -26,6 +26,7 @@ ones(T::Type, dims...) = Base.ones(T, dims...) # no need for a message
 
 function zeros(dims...)
   Base.depwarn("Flux.zeros(size...) is deprecated, please use Flux.zeros32(size...) or Base.zeros(Float32, size...)", :ones)
```
CarloLucibello, line 28:
```suggestion
  Base.depwarn("Flux.zeros(size...) is deprecated, please use Flux.zeros32(size...) or Base.zeros(Float32, size...)", :zeros)
```
Not sure what that is

CarloLucibello:
should `ones` get a similar treatment?

DhairyaLGandhi:
I think its also a bit weird that there's an error message about forcing people to use a specific type in `zeros32`.

DhairyaLGandhi:
Merging since this is required by sciml

PR #1672, opened 2021-07-22
Add GPU activation tests for grouped conv

DhairyaLGandhi:
Good to test the groups in `Conv` with a few activations as well.


ToucheSir (approved):
Just noticed we're all green on CI now, nice :)

PR #1674, opened 2021-07-23
Generalise Parallel forwards pass

### LINKED ISSUE #1673: `Parallel` layer doesn't need to be tied to array input
https://github.com/FluxML/Flux.jl/blob/9931730129f514425d939683f72dd566a791472f/src/layers/basic.jl#L419

This line is needlessly restrictive, and breaks for a use case I'd like to make involving a custom type in my ChemistryFeaturization and AtomicGraphNets packages where the input to an `AGNConv` layer can be a `FeaturizedAtoms` type.

I'm not sure if the solution is just to change `AbstractArray` to `Any` or what...

DhairyaLGandhi:
Fix #1673

We should also do a sweep of other layers that may be more restrictive due to excessive typing. By default there should be no typing on the input, or even necessarily on the parameters.

The `(::Parallel)(::Tuple)` case is interesting since the current version assumes that all the elements need to spread out, but  layers may output multiple things as part of their forward pass, and ingest multiple arguments too. This is also the case in the Weave model for example. So we may want to treat tuples as invariant and treat them like a single entity. Currently, the API to pass tuples could be improved since passing tuples to layers would need an extra layer of nesting for example: `p(((x,),))`

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -415,8 +415,8 @@ Parallel(connection, layers...) = Parallel(connection, layers)
 
 @functor Parallel
 
-(m::Parallel)(x::AbstractArray) = mapreduce(f -> f(x), m.connection, m.layers)
-(m::Parallel)(xs::Vararg{<:AbstractArray}) = mapreduce((f, x) -> f(x), m.connection, m.layers, xs)
+(m::Parallel)(x) = mapreduce(f -> f(x), m.connection, m.layers)
+(m::Parallel)(xs...) = mapreduce((f, x) -> f(x), m.connection, m.layers, xs)
 (m::Parallel)(xs::Tuple) = m(xs...)
 
 Base.getindex(m::Parallel, i::Integer) = m.layers[i]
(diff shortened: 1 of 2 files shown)


darsnack (approved):

darsnack:
Let's tackle the tuple issue separately. Could someone open a separate issue describing the Weave model (or link to it)? I tried searching but couldn't find it.

DhairyaLGandhi:
Think of how any intermediate layer may return multiple outputs, and the next layer may be a `Parallel` with multiple branches which each expect multiple arguments (produced by the previous layer).

A simple pseudo API demo might look like
```julia
struct L1
  m
end

function (l::L1)(x)
  l.m(x), l.m(x)
end

struct L2
  m
end

function (l::L2)(x, y)
  l.m(x), l.m(y)
end

Chain(L1(Dense),
      Parallel(vcat, L2(Dense), L2(Dense)))
```

darsnack (dismissed):
Let's merge this fix now and deal with the more complex stuff in a separate PR

ToucheSir:
@darsnack is it fair to say this has your stamp of approval? If so, who wants to summon Bors?

darsnack (approved):
Yes this is good to go. If the author is a Flux committer, then only the author can trigger `bors`. (Or at least that's what the rules are supposed to be).

ToucheSir:
Got it, looks like I need to read through ColPrac again :sweat_smile:

Issue #1671, opened 2021-07-21
GRU implementation details


mkschleg:
There are multiple variations of the gated recurrent unit tied to different versions of the [arxiv](https://arxiv.org/abs/1406.1078) paper. The variation only changes where the reset vector (`r`) is applied (either before or after the matrix operation). Currently, we are implementing the version found in v1 of the paper, which follows from what [PyTorch does](https://pytorch.org/docs/stable/generated/torch.nn.GRU.html) . Tensorflow implements these different [variations](https://www.tensorflow.org/api_docs/python/tf/keras/layers/GRU), with the v3 version being the default. Another driver of this decision is CuDNN seems to only support GRU from v1 of the paper.

Would there be interest in a PR to add the v3 variation? 

If we were to add v3 (and maybe other variations), we should likely keep v1 as the default (so as not to break assumptions). But I think it is unclear in the literature exactly what people are using when using GRUs. This [paper](https://arxiv.org/pdf/1412.3555.pdf) from Yoshua Bengio's group seems to use v3 to empirically study GRUs. But everyone using Tensorflow or Pytorch are likely using v1 of the GRUs. Tensorflow says the default is v3, but it looks like they are setting the argument `reset_after=true`. This to me means they are actually using v1 (I haven't dug into their code to figure out what is doing yet because I am never a happy camper when looking at tensorflow...).

DhairyaLGandhi:
That sounds interesting! If you have an implementation of the different versions it would be great to have a PR.

ToucheSir:
This came up recently elsewhere, thanks for digging into it! Unsurprisingly, it seems like CuDNN is the bottleneck.

mkschleg:
I mentioned this in slack, so that might be what you are thinking about. I can work on a PR for this.

CarloLucibello:
Not strictly related, but we also miss bridging v1 to cudnn right now. From what I remember from @jeremiedb's posts, we would need to move away from the vector of vectors representation of a series to a matrix representation

ToucheSir:
Yeah, I don't see us trying to re-integrate with CuDNN unless the RNN interface itself undergoes some major changes. And even then, there may be a way to get comparable performance without bending over backwards to accommodate it like other frameworks have.

mkschleg:
[...]

PR #1675, opened 2021-07-23
Adding GRUv3 support.

mkschleg:
As per the starting discussion in #1671, we should provide support for variations on the GRU and LSTM cell. 

In this PR, I added support for the GRU found in v3 of the [original GRU paper](https://arxiv.org/abs/1406.1078). Current support in Flux is for v1 only. [Tensorflow](https://www.tensorflow.org/api_docs/python/tf/keras/layers/GRU) supports several variations, with this as one of the variations. 

While the feature is added and usable in this PR, this is only a first pass at a design and could use further iterations. Some questions I have:
- Should we have new types for each variation of these cells? (another possibility is through parametric options)
- Should we have a shared constructor similar to Tensorflow/Pytorch? (it might make sense to rename the current GRU to GRUv1 if we want to do this).

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -213,7 +213,8 @@ Base.show(io::IO, l::GRUCell) =
     GRU(in::Integer, out::Integer)
 
 [Gated Recurrent Unit](https://arxiv.org/abs/1406.1078) layer. Behaves like an
-RNN but generally exhibits a longer memory span over sequences.
+RNN but generally exhibits a longer memory span over sequences. This implements
+the cell proposed in v1 of the above paper.
 
 See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
@@ -268,7 +269,8 @@ Base.show(io::IO, l::GRUv3Cell) =
     GRUv3(in::Integer, out::Integer)
 
 [Gated Recurrent Unit](https://arxiv.org/abs/1406.1078) layer. Behaves like an
-RNN but generally exhibits a longer memory span over sequences.
+RNN but generally exhibits a longer memory span over sequences. This implements
+the version proposed in v3 of the above arxiv paper.
 
 See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -270,7 +270,7 @@ Base.show(io::IO, l::GRUv3Cell) =
 
 [Gated Recurrent Unit](https://arxiv.org/abs/1406.1078) layer. Behaves like an
 RNN but generally exhibits a longer memory span over sequences. This implements
-the version proposed in v3 of the above arxiv paper.
+the variant proposed in v3 of the referenced paper.
 
 See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -233,6 +233,61 @@ function Base.getproperty(m::GRUCell, sym::Symbol)
   end
 end
 
+
+# GRU v3
+
+struct GRUv3Cell{A,V,S}
+  Wi::A
+  Wh::A
+  b::V
+  Wh_h̃::A
+  state0::S
+end
+
+GRUv3Cell(in, out; init = glorot_uniform, initb = zeros32, init_state = zeros32) =
+    GRUv3Cell(init(out * 3, in), init(out * 2, out), initb(out * 3), 
+              init(out, out), init_state(out,1))
+
+function (m::GRUv3Cell{A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {A,V,T}
+  b, o = m.b, size(h, 1)
+  gx, gh = m.Wi*x, m.Wh*h
+  r = σ.(gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1))
+  z = σ.(gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2))
+  h̃ = tanh.(gate(gx, o, 3) .+ (m.Wh_h̃ * (r .* h)) .+ gate(b, o, 3))
+  h′ = (1 .- z) .* h̃ .+ z .* h
+  sz = size(x)
+  return h′, reshape(h′, :, sz[2:end]...)
+end
+
+@functor GRUv3Cell
+
+Base.show(io::IO, l::GRUv3Cell) =
+  print(io, "GRUv3Cell(", size(l.Wi, 2), ", ", size(l.Wi, 1)÷3, ")")
+
+"""
+    GRUv3(in::Integer, out::Integer)
+
+[Gated Recurrent Unit](https://arxiv.org/abs/1406.1078) layer. Behaves like an
+RNN but generally exhibits a longer memory span over sequences.
+
+See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
+for a good overview of the internals.
+"""
+GRUv3(a...; ka...) = Recur(GRUv3Cell(a...; ka...))
+Recur(m::GRUv3Cell) = Recur(m, m.state0)
+
+# TODO remove in v0.13
+function Base.getproperty(m::GRUv3Cell, sym::Symbol)
+  if sym === :h
+    Zygote.ignore() do
+      @warn "GRUv3Cell field :h has been deprecated. Use m::GRUv3Cell.state0 instead."
+    end
+    return getfield(m, :state0)
+  else
+    return getfield(m, sym)
+  end
+end
+
 @adjoint function Broadcast.broadcasted(f::Recur, args...)
   Zygote.∇map(__context__, f, args...)
 end
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -268,7 +268,7 @@ Base.show(io::IO, l::GRUv3Cell) =
 """
     GRUv3(in::Integer, out::Integer)
 
-[Gated Recurrent Unit](https://arxiv.org/abs/1406.1078) layer. Behaves like an
+[Gated Recurrent Unit](https://arxiv.org/abs/1406.1078v3) layer. Behaves like an
 RNN but generally exhibits a longer memory span over sequences. This implements
 the variant proposed in v3 of the referenced paper.
 
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -184,12 +184,12 @@ end
 # GRU
 
 function _gru_output(Wi, Wh, b, x, h)
-    o = size(h, 1)
-    gx, gh = Wi*x, Wh*h
-    r = σ.(gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1))
-    z = σ.(gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2))
+  o = size(h, 1)
+  gx, gh = Wi*x, Wh*h
+  r = σ.(gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1))
+  z = σ.(gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2))
 
-    return gx, gh, r, z
+  return gx, gh, r, z
 end
 
 struct GRUCell{A,V,S}
@@ -253,8 +253,8 @@ struct GRUv3Cell{A,V,S}
 end
 
 GRUv3Cell(in, out; init = glorot_uniform, initb = zeros32, init_state = zeros32) =
-    GRUv3Cell(init(out * 3, in), init(out * 2, out), initb(out * 3), 
-              init(out, out), init_state(out,1))
+  GRUv3Cell(init(out * 3, in), init(out * 2, out), initb(out * 3), 
+            init(out, out), init_state(out,1))
 
 function (m::GRUv3Cell{A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {A,V,T}
   b, o = m.b, size(h, 1)
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -183,7 +183,8 @@ end
 
 # GRU
 
-function _gru_output(Wi, Wh, x, h)
+function _gru_output(Wi, Wh, b, x, h)
+    o = size(h, 1)
     gx, gh = Wi*x, Wh*h
     r = σ.(gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1))
     z = σ.(gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2))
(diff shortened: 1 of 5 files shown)


mkschleg:
I added tests to `test/layers/recurrent.jl`. I'm not sure there are more tests I should add? Also added doc strings for the new method and updated the doc string for the original version to clarify it is v1 of the arxiv paper.

src/layers/recurrent.jl
```diff
@@ -233,6 +234,62 @@ function Base.getproperty(m::GRUCell, sym::Symbol)
...
+
+struct GRUv3Cell{A,V,S}
+  Wi::A
+  Wh::A
+  b::V
+  Wh_h̃::A
+  state0::S
+end
+
+GRUv3Cell(in, out; init = glorot_uniform, initb = zeros32, init_state = zeros32) =
```
DhairyaLGandhi, line 248: Needs an activation

src/layers/recurrent.jl
```diff
@@ -233,6 +234,62 @@ function Base.getproperty(m::GRUCell, sym::Symbol)
...
+    GRUv3(in::Integer, out::Integer)
+
+[Gated Recurrent Unit](https://arxiv.org/abs/1406.1078) layer. Behaves like an
+RNN but generally exhibits a longer memory span over sequences. This implements
+the version proposed in v3 of the above arxiv paper.
+
+See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
+for a good overview of the internals.
+"""
+GRUv3(a...; ka...) = Recur(GRUv3Cell(a...; ka...))
```
DhairyaLGandhi, line 278: How do the versions differ api wise? Does this need any extra terms?

DhairyaLGandhi (commented):
Thanks for the contribution! Looking forward to it!

I've left a couple of starter comments. Could you please also add CUDA tests?

src/layers/recurrent.jl
mkschleg: We shouldn't. The main difference is the `Wh` matrix needs to be split so we can apply the reset vector appropriately.  But this doesn't require any extra parameters for the constructor.

CarloLucibello:
there seems to be a lot of code duplication, probably adding a symbol parameter to the type, `GRU{..., Mode}`, and only dispatch the call method on that seems a better alternative

mkschleg:
Right. That is definitely the question I had. The conversation was derailed a bit in #1671, as we started discussing CuDNN support. 

If we were to do the parametric option version we would have to figure out how to do the operation in the line I tagged you in with the current struct layout (it is [this line](https://github.com/FluxML/Flux.jl/blob/2cf128816e73a704e31bc420cfaf57fdb017b0d1/src/layers/recurrent.jl#L257). I'm not sure how that would work tbh without extra unnecessary operations.

src/layers/recurrent.jl
```diff
@@ -233,6 +234,62 @@ function Base.getproperty(m::GRUCell, sym::Symbol)
...
+GRUv3Cell(in, out; init = glorot_uniform, initb = zeros32, init_state = zeros32) =
+    GRUv3Cell(init(out * 3, in), init(out * 2, out), initb(out * 3), 
+              init(out, out), init_state(out,1))
+
+function (m::GRUv3Cell{A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {A,V,T}
+  b, o = m.b, size(h, 1)
+  gx, gh = m.Wi*x, m.Wh*h
+  r = σ.(gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1))
+  z = σ.(gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2))
+  h̃ = tanh.(gate(gx, o, 3) .+ (m.Wh_h̃ * (r .* h)) .+ gate(b, o, 3))
```
mkschleg, line 257: @CarloLucibello This line.

mkschleg:
@DhairyaLGandhi cuda tests added.

DhairyaLGandhi:
It might make sense to have a function that is a no-op for the regular GRU layer (without changing the struct) and splits the array for v3.

mkschleg:
I remember having issues with views + zygote awhile ago. Is that still the case? (I've not tried since v0.10.x). If we can't use views the split function would create a larger memory footprint, no? I think @darsnack had some thoughts on this in terms of clarity of API.

DhairyaLGandhi:
Should be the same memory wise

src/layers/recurrent.jl
darsnack: AFAIK the activations in LSTM/GRU are very specifically chosen. That's why they are currently not options, and we should probably keep that consistent.

mkschleg:
I think I misunderstood what you suggested. This would be in the constructor, right? Then I agree memory would be the same.

We would have to add an extra parametric type to separate Wi and Wh (as they could be different depending on the mode). 

So the struct would be:

```julia
struct GRUCell{M, Ai, Ah, V, S}
  Wi::Ai
  Wh::Ah
  b::V
  state0::S
end
```

Where M is the mode. My instinct is to put this first to make dispatch very clear, but we could put it elsewhere.

darsnack:
Repeating what I mentioned from the other thread: `GRU(...; mode = :v3)`/`GRU{:v3}(...)` are just as verbose as `GRUv3(...)`. At first glance, the forward passes seem quite different for [the line @mkschleg highlighted](https://github.com/FluxML/Flux.jl/blob/2cf128816e73a704e31bc420cfaf57fdb017b0d1/src/layers/recurrent.jl#L257). Different forward pass == different types makes sense to me here. Having conditional forward passes based on a parameter is messier IMO, and I don't there is a function argument we can pass in that runs the two alternate passes (w/o this function being complex).

We can still reduce code duplication. As an example:
```julia
function _gru_output(Wi, Wh, x, h)
   gx, gh = Wi*x, Wh*h
   r = σ.(gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1))
   z = σ.(gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2))

   return r, z
end
```
Then use `_gru_output` in all the variants. IMO it makes sense to reduce duplication via the API when it results in a intuitive interpretation of what's going on under the hood. I don't think that's the case here.

src/layers/recurrent.jl
ToucheSir: Worth pointing out that TF and JAX-based libraries do allow you to customize the activation. I presume PyTorch doesn't because it lacks a non-CuDNN path for it's GPU RNN backend.

darsnack: Interesting, is it just the output activation or all of them?

mkschleg: All of them I believe: [tensorflow GRU](https://www.tensorflow.org/api_docs/python/tf/keras/layers/GRU)

ToucheSir: All of them, IIUC. There's a distinction between "activation" functions (by default `tanh`) and "gate" functions (by default `sigmoid`).

ToucheSir: For a non-Google implementation, here's [MXNet](https://mxnet.apache.org/versions/1.8.0/api/python/docs/api/gluon/rnn/index.html#mxnet.gluon.rnn.LSTMCell).

ToucheSir:
Exploring the other extreme for a second, I wonder if we could make `GRUv3Cell` the default GRU cell and set `Wh_h̃` to `Zeros` in the `GRUCell` constructor. In theory, this would reduce `.+ (m.Wh_h̃ * (r .* h))` to a no-op and allow us to use one codepath.

mkschleg:
Maybe I'm misunderstanding. The forward for the GRUv3Cell would then look like

```julia
  h̃ = tanh.(gate(gx, o, 3) .+ r .* gate(gh, o, 3) .+ (m.Wh_h̃ * (r .* h)) .+ gate(b, o, 3))
```

How would we turn off ```r .* gate(gh, o, 3)``` for the v1 version?

src/layers/recurrent.jl
ToucheSir: Prior art in Flux: https://github.com/FluxML/Flux.jl/pull/964

DhairyaLGandhi: Yeah, I agree that we should have the activations in Flux generally across all layers.

darsnack:
If I understand it correctly, `Wh` in v1 == `[Wh Wh_h̃]` in v3. So always adopting the `gate(Wh * h, o, 1) .+ Wh_h̃ * (r .* h))` means that `Wh` takes on a slightly different interpretation each time. In v1, `Wh` is plain `Wh` and `Wh_h̃ == Zeros()`. In v3, we would initialize `Wh` to be the first two slices and `Wh_h̃` to be the last.

ToucheSir:
Sorry, that was my mistake! Missed that the v1 also had a middle term and assumed it was another case of https://julialang.zulipchat.com/#narrow/stream/238249-machine-learning/topic/Elman.20RNN.20definition.

src/layers/recurrent.jl
```diff
@@ -233,6 +234,62 @@ function Base.getproperty(m::GRUCell, sym::Symbol)
...
+  b::V
+  Wh_h̃::A
+  state0::S
+end
+
+GRUv3Cell(in, out; init = glorot_uniform, initb = zeros32, init_state = zeros32) =
+    GRUv3Cell(init(out * 3, in), init(out * 2, out), initb(out * 3), 
+              init(out, out), init_state(out,1))
+
+function (m::GRUv3Cell{A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {A,V,T}
```
DhairyaLGandhi, line 252: Do we need to remove the types on the input and parameters?

darsnack: I believe these were introduced as part of #1521. We should tackle them separately for all recurrent cells.

mkschleg: Maybe we should have a central issue which details all the updates to recurrent cells the discussion in this PR and related issue has mentioned?

mkschleg: I'm definitely invested in the recurrent architectures for flux, so would like to help. But knowing all the outstanding issues is out of scope for what I can use my time for right now.

ToucheSir: Yeah, let's get this through and litigate general changes to the RNN interface in a separate issue.

mkschleg:
What needs to be done to push this PR through?

src/layers/recurrent.jl
```diff
@@ -233,6 +234,62 @@ function Base.getproperty(m::GRUCell, sym::Symbol)
...
+function Base.getproperty(m::GRUv3Cell, sym::Symbol)
+  if sym === :h
+    Zygote.ignore() do
+      @warn "GRUv3Cell field :h has been deprecated. Use m::GRUv3Cell.state0 instead."
+    end
+    return getfield(m, :state0)
+  else
+    return getfield(m, sym)
+  end
+end
```
darsnack, line 291: Don't need this since `GRUv3Cell.h` never existed previously.

src/layers/recurrent.jl
```diff
@@ -233,6 +234,62 @@ function Base.getproperty(m::GRUCell, sym::Symbol)
...
+
+GRUv3Cell(in, out; init = glorot_uniform, initb = zeros32, init_state = zeros32) =
+    GRUv3Cell(init(out * 3, in), init(out * 2, out), initb(out * 3), 
+              init(out, out), init_state(out,1))
+
+function (m::GRUv3Cell{A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {A,V,T}
+  b, o = m.b, size(h, 1)
+  gx, gh = m.Wi*x, m.Wh*h
+  r = σ.(gate(gx, o, 1) .+ gate(gh, o, 1) .+ gate(b, o, 1))
+  z = σ.(gate(gx, o, 2) .+ gate(gh, o, 2) .+ gate(b, o, 2))
```
darsnack, line 256: Separate this out into a helper for code re-use across `GRUCell` and `GRUv3Cell`.

src/layers/recurrent.jl
```diff
@@ -233,6 +234,62 @@ function Base.getproperty(m::GRUCell, sym::Symbol)
...
+
+Base.show(io::IO, l::GRUv3Cell) =
+  print(io, "GRUv3Cell(", size(l.Wi, 2), ", ", size(l.Wi, 1)÷3, ")")
+
+"""
+    GRUv3(in::Integer, out::Integer)
+
+[Gated Recurrent Unit](https://arxiv.org/abs/1406.1078) layer. Behaves like an
+RNN but generally exhibits a longer memory span over sequences. This implements
+the version proposed in v3 of the above arxiv paper.
```
darsnack, line 273:
```suggestion
the variant proposed in v3 of the referenced paper.
```

src/layers/recurrent.jl
```diff
@@ -233,6 +234,62 @@ function Base.getproperty(m::GRUCell, sym::Symbol)
...
+
+@functor GRUv3Cell
+
+Base.show(io::IO, l::GRUv3Cell) =
+  print(io, "GRUv3Cell(", size(l.Wi, 2), ", ", size(l.Wi, 1)÷3, ")")
+
+"""
+    GRUv3(in::Integer, out::Integer)
+
+[Gated Recurrent Unit](https://arxiv.org/abs/1406.1078) layer. Behaves like an
```
darsnack, line 271:
```suggestion
[Gated Recurrent Unit](https://arxiv.org/abs/1406.1078v3) layer. Behaves like an
```

src/layers/recurrent.jl
```diff
@@ -213,7 +213,8 @@ Base.show(io::IO, l::GRUCell) =
     GRU(in::Integer, out::Integer)
 
 [Gated Recurrent Unit](https://arxiv.org/abs/1406.1078) layer. Behaves like an
```
darsnack, line 215:
```suggestion
[Gated Recurrent Unit](https://arxiv.org/abs/1406.1078v1) layer. Behaves like an
```

src/layers/recurrent.jl
```diff
@@ -213,7 +213,8 @@ Base.show(io::IO, l::GRUCell) =
...
-RNN but generally exhibits a longer memory span over sequences.
+RNN but generally exhibits a longer memory span over sequences. This implements
+the cell proposed in v1 of the above paper.
```
darsnack, line 217:
```suggestion
the variant proposed in v1 of the referenced paper.
```

darsnack (changes requested):

src/layers/recurrent.jl
mkschleg: Done in latest commit.

src/layers/recurrent.jl
mkschleg: Done.

ToucheSir:
Running CI to see what it thinks. Otherwise I think the only other change would be to squash some of those intermediate update commits.

darsnack (changes requested):
Can you add an entry to NEWS.md?

src/layers/recurrent.jl
```diff
@@ -183,6 +183,15 @@ end
 
 # GRU
 
+function _gru_output(Wi, Wh, b, x, h)
```
DhairyaLGandhi, line 186: Needs a 2-space indent to match Flux style

mkschleg:
@darsnack Where should I put the announcement in NEWS.md?

darsnack:
As a new bullet under 0.12.7 (we're missing a 0.12.6 new entry for some reason).

mkschleg:
Also, how would I squash those commits? I've not had to do that before.

darsnack:
https://stackoverflow.com/questions/35703556/what-does-it-mean-to-squash-commits-in-git

It's just 3 extra commits than necessary, so I don't think you need to bother. In the future, you can go to the "Files" tab of the PR to add review suggestions to a single commit batch (instead of accepting each suggestion as a separate commit).

darsnack (approved):
Looks good, just gotta wait for CI to pass.

PR #1681, opened 2021-07-29
Support NamedTuples for Chain + Parallel

### LINKED ISSUE #1680: Support NamedTuples for Container Layers
This came up in https://github.com/FluxML/Metalhead.jl/pull/70 and surprised me, because I thought we supported this already! 

Other than `Chain`, are there any other layers we'd want to support this? The only other one that uses a tuple field is `Parallel`, but since it doesn't forward methods to `.layers` I'm not sure this is applicable.

Another thing using NamedTuples would open up is the possibility of defining aliases for `Chain`. e.g. 
```julia
const ResNetBlock{C1,C2,N} = Chain{NamedTuple{(:conv1, :conv2, :norm), Tuple{C1, C2, N}}}
```

mcabbott:
Closes #1680, WIP. Todo list includes:

- [x] add Parallel too
- [ ] ~~worry about whether any of this will upset Zygote, like https://github.com/FluxML/Zygote.jl/pull/909~~ or, kick that can down the road.
- [x] add tests

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -31,26 +31,31 @@ true
 struct Chain{T}
   layers::T
   Chain(xs...) = new{typeof(xs)}(xs)
-  Chain(; kw...) = new{typeof(values(kw))}(values(kw))
+  function Chain(; kw...)
+    :layers in Base.keys(kw) && throw(ArgumentError("a Chain cannot have a named layer called `layers`"))
+    isempty(kw) && return new{Tuple{}}(())
+    new{typeof(values(kw))}(values(kw))
+  end
 end
 
-Base.getproperty(c::Chain, s::Symbol) = getproperty(getfield(c, :layers), s)
+Base.parent(c::Chain) = getfield(c, :layers)
+Base.getproperty(c::Chain, s::Symbol) = s === :layers ? parent(c) : getproperty(parent(c), s)
 for fun in (:getindex, :length, :first, :last, :iterate, :lastindex, :propertynames, :keys)
-  @eval Base.$fun(c::Chain, args...) = Base.$fun(getfield(c, :layers), args...)
+  @eval Base.$fun(c::Chain, args...) = Base.$fun(parent(c), args...)
 end
 
-functor(::Type{<:Chain}, c) = getfield(c, :layers), ls -> Chain(ls...)
+functor(::Type{<:Chain}, c) = parent(c), ls -> Chain(ls...)
 
 applychain(::Tuple{}, x) = x
 applychain(fs, x) = applychain(tail(Tuple(fs)), first(fs)(x))
 
-(c::Chain)(x) = applychain(getfield(c, :layers), x)
+(c::Chain)(x) = applychain(parent(c), x)
 
-Base.getindex(c::Chain, i::AbstractArray) = Chain(getfield(c, :layers)[i]...)
+Base.getindex(c::Chain, i::AbstractArray) = Chain(parent(c)[i]...)
 
 function Base.show(io::IO, c::Chain)
   print(io, "Chain")
-  show(io, getfield(c, :layers))
+  show(io, parent(c))
 end
 
 # This is a temporary and naive implementation
@@ -63,11 +68,11 @@ end
 
 Calculate the forward results of each layers in Chain `c` with `input` as model input.
 """
-activations(c::Chain, input) = extraChain(getfield(c, :layers), input)
+activations(c::Chain, input) = extraChain(parent(c), input)
 
 function extraChain(fs::Tuple, x)
   res = first(fs)(x)
-  return (res, extraChain(Base.tail(Tuple(fs)), res)...)
+  return (res, extraChain(Base.tail(fs), res)...)
 end
 extraChain(::Tuple{}, x) = ()
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -38,25 +38,24 @@ struct Chain{T}
   end
 end
 
-Base.parent(c::Chain) = getfield(c, :layers)
-Base.propertynames(c::Chain) = (Base.keys(parent(c))..., :layers)
-Base.getproperty(c::Chain, s::Symbol) = s === :layers ? parent(c) : getproperty(parent(c), s)
+Base.propertynames(c::Chain) = (Base.keys(c)..., :layers)
+Base.getproperty(c::Chain, s::Symbol) = s === :layers ? getfield(c, :layers) : getproperty(getfield(c, :layers), s)
 
 @forward Chain.layers Base.getindex, Base.length, Base.first, Base.last,
   Base.iterate, Base.lastindex, Base.keys
 
-functor(::Type{<:Chain}, c) = parent(c), ls -> Chain(ls...)
+functor(::Type{<:Chain}, c) = getfield(c, :layers), ls -> Chain(ls...)
 
 applychain(::Tuple{}, x) = x
-applychain(fs, x) = applychain(tail(Tuple(fs)), first(fs)(x))
+applychain(fs, x) = applychain(tail(fs), first(fs)(x))
 
-(c::Chain)(x) = applychain(parent(c), x)
+(c::Chain)(x) = applychain(Tuple(getfield(c, :layers)), x)
 
-Base.getindex(c::Chain, i::AbstractArray) = Chain(parent(c)[i]...)
+Base.getindex(c::Chain, i::AbstractArray) = Chain(getfield(c, :layers)[i]...)
 
 function Base.show(io::IO, c::Chain)
   print(io, "Chain")
-  show(io, parent(c))
+  show(io, c.layers)  # allows NamedTuple, but prints a trailing comma sometimes
 end
 
 # This is a temporary and naive implementation
@@ -69,7 +68,7 @@ end
 
 Calculate the forward results of each layers in Chain `c` with `input` as model input.
 """
-activations(c::Chain, input) = extraChain(parent(c), input)
+activations(c::Chain, input) = extraChain(Tuple(c.layers), input)
 
 function extraChain(fs::Tuple, x)
   res = first(fs)(x)
@@ -442,11 +441,10 @@ Base.getindex(m::Parallel, i::Integer) = m.layers[i]
 Base.getindex(m::Parallel, s::Symbol) = m.layers[s]
 Base.getindex(m::Parallel, i::AbstractVector) = Parallel(m.connection, m.layers[i]...)
 
-Base.parent(m::Parallel) = getfield(m, :layers)
 Base.keys(m::Parallel) = Base.keys(getfield(m, :layers))
 Base.propertynames(m::Parallel) = (Base.keys(getfield(m, :layers))..., :connection, :layers)
 Base.getproperty(m::Parallel, s::Symbol) = s === :connection ? getfield(m, :connection) : 
-  s === :layers ? parent(m) : getproperty(parent(m), s)
+  s === :layers ? getfield(m, :layers) : getproperty(getfield(m, :layers), s)
 
 trainable(m::Parallel) = (m.connection, m.layers...)
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -429,10 +429,10 @@ end
 Parallel(connection, layers...) = Parallel(connection, layers)
 function Parallel(connection; kw...)
   layers = NamedTuple(kw)
-  if :layers in Base.keys(layers) || :layers in Base.keys(layers)
-    throw(ArgumentError("a Parallel layer cannot have a named sub-layer called `layers`"))
+  if :layers in Base.keys(layers) || :connection in Base.keys(layers)
+    throw(ArgumentError("a Parallel layer cannot have a named sub-layer called `connection` or `layers`"))
   elseif isempty(layers)
-    throw(ArgumentError("can't construct a Parallel layer with no paths"))
+    Parallel(connection, ())
   end
   Parallel(connection, layers)
 end
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -4,7 +4,7 @@
 
 Collects multiple layers / functions to be called in sequence
 on a given input. Supports indexing and slicing, `m[2]` or `m[1:end-1]`,
-and if names are given, `m.name == m[1]` etc.
+and if names are given, `m[:name] == m[1]` etc. (but not yet `m.name`).
 
 # Examples
 
@@ -24,7 +24,7 @@ true
 julia> m2 = Chain(enc = Chain(Flux.flatten, Dense(10, 5, tanh)), 
                   dec = Dense(5, 2));
 
-julia> m2(x) == (m2.dec ∘ m2.enc)(x)
+julia> m2(x) == (m2[:dec] ∘ m2[:enc])(x)
 true
 ```
 """
@@ -38,22 +38,19 @@ struct Chain{T}
   end
 end
 
-Base.propertynames(c::Chain) = (Base.keys(c)..., :layers)
-Base.getproperty(c::Chain, s::Symbol) = s === :layers ? getfield(c, :layers) : getproperty(getfield(c, :layers), s)
-
 @forward Chain.layers Base.getindex, Base.length, Base.first, Base.last,
   Base.iterate, Base.lastindex, Base.keys
 
-functor(::Type{<:Chain}, c) = getfield(c, :layers), ls -> Chain(ls...)
+functor(::Type{<:Chain}, c) = c.layers, ls -> Chain(ls...)
 
 applychain(::Tuple{}, x) = x
 applychain(fs::Tuple, x) = applychain(tail(fs), first(fs)(x))
 
-(c::Chain)(x) = applychain(Tuple(getfield(c, :layers)), x)
+(c::Chain)(x) = applychain(Tuple(c.layers), x)
 
-Base.getindex(c::Chain, i::AbstractArray) = Chain(getfield(c, :layers)[i]...)
+Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i]...)
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) = 
-  Chain(; NamedTuple{Base.keys(c)[i]}(Tuple(getfield(c, :layers))[i])...)
+  Chain(; NamedTuple{Base.keys(c)[i]}(Tuple(c.layers)[i])...)
 
 function Base.show(io::IO, c::Chain)
   print(io, "Chain(")
@@ -400,7 +397,7 @@ Called with one input `x`, this is equivalent to `reduce(connection, [l(x) for l
 If called with multiple inputs, they are `zip`ped with the layers, thus `Parallel(+, f, g)(x, y) = f(x) + g(y)`.
 
 Like [`Chain`](@ref), its sub-layers may be given names using the keyword constructor.
-These can be accessed by indexing or with a dot: `m[1] == m[:name] == m.name` is the first layer.
+These can be accessed by indexing or with a dot: `m[1] == m[:name]` is the first layer.
 
 # Examples
 
@@ -422,10 +419,10 @@ Parallel(
 julia> model2(rand(10), rand(5)) |> size
 (2,)
 
-julia> model2.α(rand(10)) |> size
+julia> model2[:α](rand(10)) |> size
 (2,)
 
-julia> model2.β == model2[2]
+julia> model2[:β] == model2[2]
 true
 ```
 """
@@ -455,9 +452,6 @@ Base.getindex(m::Parallel, i) = m.layers[i]
 Base.getindex(m::Parallel, i::AbstractVector) = Parallel(m.connection, m.layers[i]...)
 
 Base.keys(m::Parallel) = Base.keys(getfield(m, :layers))
-Base.propertynames(m::Parallel) = (Base.keys(getfield(m, :layers))..., :connection, :layers)
-Base.getproperty(m::Parallel, s::Symbol) = s === :connection ? getfield(m, :connection) : 
-  s === :layers ? getfield(m, :layers) : getproperty(getfield(m, :layers), s)
 
 trainable(m::Parallel) = (m.connection, m.layers...)
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -394,10 +394,13 @@ end
     Parallel(connection; name = layer, ...)
 
 Create a 'Parallel' layer that passes an input array to each path in
-`layers`, reducing the output with `connection`.
+`layers`, before reducing the output with `connection`.
 
 Called with one input `x`, this is equivalent to `reduce(connection, [l(x) for l in layers])`.
 If called with multiple inputs, they are `zip`ped with the layers, thus `Parallel(+, f, g)(x, y) = f(x) + g(y)`.
+
+Like [`Chain`](@ref), its sub-layers may be given names using the keyword constructor.
+These can be accessed by indexing or with a dot: `m[1] == m[:name] == m.name` is the first layer.
 
 # Examples
 
@@ -406,18 +409,24 @@ julia> model = Chain(Dense(3, 5),
                      Parallel(vcat, Dense(5, 4), Chain(Dense(5, 7), Dense(7, 4))),
                      Dense(8, 17));
 
-julia> size(model(rand(3)))
+julia> model(rand(3)) |> size
 (17,)
 
-julia> model = Parallel(+, Dense(10, 2), Dense(5, 2))
+julia> model2 = Parallel(+; α = Dense(10, 2, tanh), β = Dense(5, 2))
 Parallel(
   +,
-  Dense(10, 2),                         # 22 parameters
-  Dense(5, 2),                          # 12 parameters
+  α = Dense(10, 2, tanh),               # 22 parameters
+  β = Dense(5, 2),                      # 12 parameters
 )                   # Total: 4 arrays, 34 parameters, 392 bytes.
 
-julia> size(model(rand(10), rand(5)))
+julia> model2(rand(10), rand(5)) |> size
 (2,)
+
+julia> model2.α(rand(10)) |> size
+(2,)
+
+julia> model2.β == model2[2]
+true
 ```
 """
 struct Parallel{F, T}
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -4,7 +4,7 @@
 
 Collects multiple layers / functions to be called in sequence
 on a given input. Supports indexing and slicing, `m[2]` or `m[1:end-1]`,
-and if names are given, `m[:name] == m[1]` etc. (but not `m.name` at present).
+and if names are given, `m[:name] == m[1]` etc.
 
 # Examples
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -83,7 +83,7 @@ extraChain(::Tuple{}, x) = ()
     Dense(in, out, σ=identity; bias=true, init=glorot_uniform)
     Dense(W::AbstractMatrix, [bias, σ])
 
-Create a traditional fully-connected layer, whose forward pass is given by:
+Create a traditional `Dense` layer, whose forward pass is given by:
 
     y = σ.(W * x .+ bias)
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -1,11 +1,10 @@
 """
     Chain(layers...)
+    Chain(name = layer, ...)
 
-Chain multiple layers / functions together, so that they are called in sequence
-on a given input.
-
-`Chain` also supports indexing and slicing, e.g. `m[2]` or `m[1:end-1]`.
-`m[1:3](x)` will calculate the output of the first three layers.
+Collects multiple layers / functions to be called in sequence
+on a given input. Supports indexing and slicing, `m[2]` or `m[1:end-1]`,
+and if names are given, `m.name == m[1]` etc.
 
 # Examples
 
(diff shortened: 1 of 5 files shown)


src/layers/basic.jl
```diff
@@ -15,35 +14,43 @@ julia> m = Chain(x -> x^2, x -> x+1);
...
   layers::T
   Chain(xs...) = new{typeof(xs)}(xs)
+  Chain(; kw...) = new{typeof(values(kw))}(values(kw))
 end
 
-@forward Chain.layers Base.getindex, Base.length, Base.first, Base.last,
-  Base.iterate, Base.lastindex
+Base.getproperty(c::Chain, s::Symbol) = getproperty(getfield(c, :layers), s)
+for fun in (:getindex, :length, :first, :last, :iterate, :lastindex, :propertynames, :keys)
+  @eval Base.$fun(c::Chain, args...) = Base.$fun(getfield(c, :layers), args...)
```
mcabbott, line 39: `@forward` uses getproperty not getfield, so doesn't work here. 

But it's possible that `:layers` should be a reserved name, forbidden for naming your layers.

It's also possible that `getfield(c, :layers)` is going to behave badly with Zygote.

src/layers/basic.jl
```diff
@@ -15,35 +14,43 @@ julia> m = Chain(x -> x^2, x -> x+1);
...
+julia> x = rand(10, 32);
 
 julia> m(x) == m[2](m[1](x))
 true
+
+julia> m2 = Chain(enc = Chain(Flux.flatten, Dense(10, 5, tanh)), 
+                  dec = Dense(5, 2));
+
+julia> m2(x) == (m2.dec ∘ m2.enc)(x)
+true
```
mcabbott, line 28: This example looks like so:
```julia
julia> m2
Chain(
  enc = Chain(
    Flux.flatten,
    Dense(10, 5, tanh),                 # 55 parameters
  ),
  dec = Dense(5, 2),                    # 12 parameters
)                   # Total: 4 arrays, 67 parameters, 556 bytes.
```

src/layers/basic.jl
darsnack: Can we just update MacroTools to call `getfield` instead?

darsnack: I feel like `layers` could be changed to `_layers` and become forbidden. Or even just make `layers` forbidden if we are worried about changing the field name.

src/layers/basic.jl
mcabbott: First round of test failures were from this, maybe it's OK:
```julia
julia> Chain(Dense(2, 3),) |> println
Chain(Dense(2, 3),)
```

DhairyaLGandhi:
ref #1682

src/layers/basic.jl
mcabbott: Maybe least surprising to forbid it, there are no existing uses, but there may well be people writing `c.layers` in the wild. Done.

Agree MacroTools ought to be fixed, eventually.

mcabbott:
> ref #1682

Oh. The goal of this one is to be transparent. Just as the Tuple is never visible, you just provide arguments to chain and then `m[1]` retrieves them, so the NamedTuple should be invisible, and `m.name` should just work.

DhairyaLGandhi:
Seems like this PR requires taking on a significant amount of complication. And a `Chain` is not a `Tuple`. The parallel to draw might be `m[:key]`, as `m[1]`

A Chain is a struct, and the dot syntax denotes getting the fields (or properties) of the struct typically, but if it's desirable to have the dot syntax overloaded, one can define the following giving precedence to the field of the struct over that of the NamedTuple. Which is the expected outcome anyway.

```julia
julia> Base.getproperty(c::Chain, k::Symbol) = k == :layers ? Base.getfield(c, :layers) : Base.getproperty(Base.getfield(c, :layers), k)
```

src/layers/show.jl
```diff
@@ -13,16 +13,28 @@ for T in [
...
-    println(io, " "^indent, nameof(typeof(obj)), "(")
-    for c in children
-      _big_show(io, c, indent+2)
+    println(io, " "^indent, isnothing(name) ? "" : "$name = ", nameof(typeof(obj)), "(")
+    if obj isa Chain{<:NamedTuple} && children == getfield(obj, :layers)
+      # then we insert names -- can this be done more generically? 
+      for k in Base.keys(obj)
+        _big_show(io, obj[k], indent+2, k)
+      end
+    elseif obj isa Parallel{<:Any, <:NamedTuple}
```
mcabbott, line 27: A bit ugly, but allows (and parses) this:
```
Chain(
  I = Dense(3, 5),                      # 20 parameters
  II = Parallel(
    vcat,
    α = Dense(5, 4),                    # 24 parameters
    β = Chain(
      i = Dense(5, 7),                  # 42 parameters
      ii = Dense(7, 4),                 # 32 parameters
    ),
  ),
  III = Dense(8, 17),                   # 153 parameters
)                   # Total: 10 arrays, 271 parameters, 1.840 KiB.
```

darsnack: I think this looks great. It's definitely info dense, but that's a user choice (I assume the names here are not auto-inserted). I wouldn't name my layers so deep into the model.

mcabbott:
> one can define

Yes, but normally with `===`, like here:

https://github.com/FluxML/Flux.jl/pull/1681/files#diff-b6c250c343270a4d87fe8ae8dc99e882c5a1b7464e516e60b4b28e14856fb202R43

src/layers/basic.jl
```diff
@@ -15,35 +14,49 @@ julia> m = Chain(x -> x^2, x -> x+1);
...
   layers::T
   Chain(xs...) = new{typeof(xs)}(xs)
+  function Chain(; kw...)
+    :layers in Base.keys(kw) && throw(ArgumentError("a Chain cannot have a named layer called `layers`"))
+    isempty(kw) && return new{Tuple{}}(())
+    new{typeof(values(kw))}(values(kw))
+  end
 end
 
+Base.parent(c::Chain) = getfield(c, :layers)
```
darsnack, line 41: I suppose there is no harm in defining this, but it seems weird to me semantically. I've only ever seen this used for array types.

src/layers/basic.jl
```diff
@@ -15,35 +14,49 @@ julia> m = Chain(x -> x^2, x -> x+1);
...
 
-Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i]...)
+Base.getindex(c::Chain, i::AbstractArray) = Chain(parent(c)[i]...)
 
 function Base.show(io::IO, c::Chain)
-  print(io, "Chain(")
-  join(io, c.layers, ", ")
-  print(io, ")")
+  print(io, "Chain")
+  show(io, parent(c))
```
darsnack, line 59: This should fix the trailing comma. Is the current change here for another reason?

```suggestion
  join(io, parent(c), ", ")
```

src/layers/show.jl
mcabbott: Yes, this is just to check that the nested printing doesn't try to impose good taste on you. If you don't provide names, then the storage is a Tuple as before, and none are printed.

I have also wondered whether the printing should somehow number the un-named layers, so that you can see what `m[13]` is going to give you, but didn't think of a nice way.

src/layers/basic.jl
mcabbott: The change is a quick hack so that it shows you the names when there's NamedTuple. Maybe with more thought we can hack `join` & `zip` together nicely...

src/layers/basic.jl
mcabbott: Yes I'm not sure this is a great idea. Especially for Parallel. There may be no need to define such a function, just leave it `getfield` or dot-access.

src/layers/basic.jl
```diff
@@ -412,21 +422,38 @@ struct Parallel{F, T}
 end
 
 Parallel(connection, layers...) = Parallel(connection, layers)
+function Parallel(connection; kw...)
+  layers = NamedTuple(kw)
+  if :layers in Base.keys(layers) || :layers in Base.keys(layers)
+    throw(ArgumentError("a Parallel layer cannot have a named sub-layer called `layers`"))
+  elseif isempty(layers)
+    throw(ArgumentError("can't construct a Parallel layer with no paths"))
+  end
```
darsnack, line 431: What happens if we remove this case? In FastAI.jl, we have an automated builder that might create a `Parallel` with no branches.

mcabbott: Yes I meant to ask. You can construct one right now, but I think you cannot call it. (After 10s of trying.)

My vote is normally for an early error, rather than letting you make something which cannot make sense. But I could be persuaded. It should be an empty Tuple I presume, not an empty NamedTuple.

src/layers/show.jl
darsnack: Yeah in the original prototype in my head, I was going to force `NamedTuple`s and auto-generated names like `L$i`. But I prefer the current route. In general, I think if you aren't going to name a layer, then you only care about iterating the group, not indexing specific ones. So support for the `m[13]` case probably wouldn't be a greatly missed feature.

src/layers/basic.jl
darsnack: Something like
```julia
_show_layers(io, layers::Tuple) = join(io, layers, ", ")
_show_layers(io, layers::NamedTuple) = join(io, map((k, v) -> "$k = $v", pairs(layers)), ", ")
```

DhairyaLGandhi:
So, if we want the dot-access functionality, I can add that to #1681 over this, and this PR can work with the printing then.

src/layers/basic.jl
mcabbott: Now removed.

src/layers/basic.jl
mcabbott: Oh except `ERROR: map is not defined on dictionaries`. So I made it a comprehension.

src/layers/basic.jl
darsnack: Right you can't call it because of reducing an empty collection. But I was thinking we could specify `init` in the `mapreduce`s.

In our case, the input would be an empty array and the branches would be an empty tuple. The model here is a tabular model which has to main pre-processing branches — one for categorical data and another for continuous data. The output of these branches are concatenated and fed into a chain of FC layers. The categorical branch itself is another `Parallel`. When there are no categorical columns in the dataset, this branch should be a no-op (neatly handled as an empty array in our case). [Here is a link](https://github.com/FluxML/FastAI.jl/blob/eb2cfc85a7ec0e50bae86d9434540b5c26eb8401/src/models/tabularmodel.jl#L25-L33) to the code that would be iterating over nothing.

It wouldn't be the end of the world if it was forbidden though. Perhaps returning an empty array only makes sense in our use-case and would error in another type of model.

mcabbott: Ok, well for now I remove this error here, perhaps it's orthogonal to this PR. 

The other quirk I noticed on reading here is that `mapreduce` will accept mismatched lengths. Is this desired or a footgun?
```
julia> Parallel(hcat, x->x.+1, x->x.+2, x->x.+3)([0], [0])
1×2 Matrix{Int64}:
 1  2

julia> Parallel(hcat, x->x.+1, x->x.+2, x->x.+3)([0], [0], [0], [1000])
1×3 Matrix{Int64}:
 1  2  3
```

darsnack: Hmm that's the default behavior for `zip`, and the equivalence with `zip` is documented in the docstring for the layer. I guess it could be useful for the case where you create a bunch of branches but don't use all of them every pass. I could also see it being a footgun with reduction operators like `+`. Not sure the right call here.

mcabbott: Indeed, it does say `zip`, I missed that. Your call I guess, and also orthogonal to this PR, sorry.

DhairyaLGandhi:
Yeah, I think I'm uncomfortable taking on the dependency here especially since a simpler solution exists.

darsnack:
Maybe remove all the `getfield(c, :layers)` to minimize code changes? At this point, `c.layers` will route through `getproperty`. In which case, I think #1681 and #1682 are the same approach.

DhairyaLGandhi:
In that case, let #1682 handle the constructor, and this PR can handle printing.

mcabbott:
> Maybe remove all the getfield(c, :layers) to minimize

Maybe! What I think needs a close look before merging is that this doesn't trigger any bad behaviour from Zygote. There are issues around getproperty and getfield and literal_getproperty, I don't have the latest status in my head, but it needs a careful check that this doesn't produce some huge regression.

> let #1682 handle the constructor

Not sure this is such a huge change that it gains by being split across multiple PRs. What #1682 proposes to add which isn't added here is a constructor `Chain(::NamedTuple)`. That does run at present, but doesn't produce anything useful, since tuples aren't callable. There is no corresponding `Chain(::Tuple)` constructor, because the Tuple is really an implementation detail. Instead the default constructor takes the layers directly, and the named constructor here does the same.

DhairyaLGandhi:
Dot access doesn't seem terribly critical especially if the `Chain`s keys are already accessible through the indexing syntax and avoids gotchas with getproperty. And the code is simpler in #1682. It would also help to remove overloads that aren't consistent between the tuple and named tuple case.

mcabbott:
Yes, I think dot access is a nice-to-have, not an essential feature. It will be trivial to remove if it does turn out to cause problems we can't solve.

ToucheSir:
Can the `getproperty` changes be sequestered into their own PR? I'm not at all comfortable or convinced that they should exist, and `getindex` should work just fine in the meantime. The constructor and printing changes are quick wins, so getting them merged would let downstream start working while we litigate the rest.

mcabbott:
Sure, could do. That may indeed be quicker than getting up to speed on Zygote's sharp edges. Then this PR will be constructors, storage, printing, tests (easy, todo), and the later one can be dot access (and harder tests).

src/layers/basic.jl
```diff
@@ -15,36 +14,51 @@ julia> m = Chain(x -> x^2, x -> x+1);
...
+julia> m2(x) == (m2[:dec] ∘ m2[:enc])(x)
+true
 ```
 """
-struct Chain{T<:Tuple}
+struct Chain{T}
   layers::T
   Chain(xs...) = new{typeof(xs)}(xs)
+  function Chain(; kw...)
+    :layers in Base.keys(kw) && throw(ArgumentError("a Chain cannot have a named layer called `layers`"))
```
darsnack, line 35: Do we need this restriction now that we are punting on the dot-syntax?

mcabbott: It lets us safely punt, by preventing you from writing code which may later become ambiguous.

DhairyaLGandhi, line 35: This should definitely be removed here and in Parallel

DhairyaLGandhi (dismissed):
As requested, it would be better to factor out the changes to Chain

src/layers/basic.jl
mcabbott: Already mentioned above. Unless we think there is zero chance ever of adding `.field` access, this is low-cost bit of futre-proofing. Allowing .field access is one of the bits you explicitly approved of, if writing code for it counts as approval.

src/layers/basic.jl
```diff
@@ -412,21 +432,32 @@ struct Parallel{F, T}
...
+  end
+  Parallel(connection, layers)
+end
 
 @functor Parallel
 
-(m::Parallel)(x::AbstractArray) = mapreduce(f -> f(x), m.connection, m.layers)
-(m::Parallel)(xs::Vararg{<:AbstractArray}) = mapreduce((f, x) -> f(x), m.connection, m.layers, xs)
+(m::Parallel)(x::AbstractArray) = mapreduce(f -> f(x), m.connection, Tuple(m.layers))
+(m::Parallel)(xs::Vararg{<:AbstractArray}) = mapreduce((f, x) -> f(x), m.connection, Tuple(m.layers), xs)
```
darsnack, line 448: Since we're now explicitly allowing empty branches.

```suggestion
(m::Parallel)(x::AbstractArray) = mapreduce(f -> f(x), m.connection, Tuple(m.layers); init = [])
(m::Parallel)(xs::Vararg{<:AbstractArray}) = mapreduce((f, x) -> f(x), m.connection, Tuple(m.layers), xs; init = [])
```

darsnack (changes requested):

src/layers/basic.jl
mcabbott: Should this have a test?

darsnack: Probably should be `typeof(x)[]` but saving it for a separate PR is okay.

src/layers/basic.jl
```diff
@@ -1,11 +1,10 @@
...
+    Chain(name = layer, ...)
 
-Chain multiple layers / functions together, so that they are called in sequence
-on a given input.
-
-`Chain` also supports indexing and slicing, e.g. `m[2]` or `m[1:end-1]`.
-`m[1:3](x)` will calculate the output of the first three layers.
+Collects multiple layers / functions to be called in sequence
+on a given input. Supports indexing and slicing, `m[2]` or `m[1:end-1]`,
+and if names are given, `m[:name] == m[1]` etc. (but not yet `m.name`).
```
darsnack, line 7:
```suggestion
and if names are given, `m[:name] == m[1]` etc. (but not `m.name`).
```

src/layers/basic.jl
mcabbott: In fact, is it obvious this should be `mapreduce`? The docstring doesn't really specify whether the combinator is applied pairwise of vararg. Its examples use `+`, where these are equivalent, ditto things like `Parallel(vcat, ...)`. But the vararg version would be more efficient (although I cannot imagine this being a bottleneck).

src/layers/basic.jl
```diff
@@ -15,36 +14,51 @@ julia> m = Chain(x -> x^2, x -> x+1);
...
-julia> x = rand(10);
+julia> x = rand(10, 32);
 
 julia> m(x) == m[2](m[1](x))
 true
+
+julia> m2 = Chain(enc = Chain(Flux.flatten, Dense(10, 5, tanh)), 
+                  dec = Dense(5, 2));
+
+julia> m2(x) == (m2[:dec] ∘ m2[:enc])(x)
```
darsnack, line 27:
```suggestion
julia> m2(x) == m2[:dec](m2[:enc](x))
```

src/layers/basic.jl
```diff
@@ -378,32 +388,42 @@ end
...
 
 Create a 'Parallel' layer that passes an input array to each path in
-`layers`, reducing the output with `connection`.
+`layers`, before reducing the output with `connection`.
 
 Called with one input `x`, this is equivalent to `reduce(connection, [l(x) for l in layers])`.
 If called with multiple inputs, they are `zip`ped with the layers, thus `Parallel(+, f, g)(x, y) = f(x) + g(y)`.
 
+Like [`Chain`](@ref), its sub-layers may be given names using the keyword constructor.
+These can be accessed by indexing or with a dot: `m[1] == m[:name]` is the first layer.
```
darsnack, line 400:
```suggestion
These can be accessed by indexing: `m[1] == m[:name]` is the first layer.
```

darsnack (changes requested):
Some last minute tweaks

src/layers/basic.jl
DhairyaLGandhi: It may break #1673 again

src/layers/basic.jl
DhairyaLGandhi: If it added in the future, a future pr can handle the tradeoffs.

test/layers/basic.jl
```diff
@@ -19,12 +19,28 @@ import Flux: activations
...
+    @test_nowarn Chain(first = Dense(10, 5, σ), second = Dense(5, 2))(randn(10))
+    m = Chain(first = Dense(10, 5, σ), second = Dense(5, 2))
+    @test m[:first] == m[1]
+    @test m[1:2] == m
+
+    @test_throws ArgumentError Chain(layers = Dense(10, 10), two = identity) # reserved name
+
+    mt = Chain((Dense(2,2),)) # constructor accepts Any
+    @test_throws MethodError mt(rand(2)) # but Tuples aren't callable
+    mnt = Chain((; name = Dense(2,2),))
```
DhairyaLGandhi, line 32: I think this may be the contructor for #1682 and seems more explicit than kwargs which aren't necessarily part of the API

src/layers/basic.jl
mcabbott: This form is shown a few lines above, I thought perhaps it was nicer to show other ways to think about this, rather than simply repeat.

test/layers/basic.jl
mcabbott: This encodes the present behaviour in tests, not changed by this PR. If we wish to change it in a follow-up, such a follow-up can discuss the merits, and remove as needed.

src/layers/basic.jl
mcabbott: Oh I thought I got all of them, thanks.

src/layers/basic.jl
mcabbott:
```suggestion
and if names are given, `m[:name] == m[1]` etc. (but not `m.name` at present).
```

src/layers/basic.jl
mcabbott: I mean, technically, a future PR would then be a breaking change. This is a super-low-cost way to avoid that.

src/layers/basic.jl
mcabbott: I leave it alone for now? It's slightly odd to make an explicit branch for an empty cases which doesn't really work, but that case does mean that the PR overall doesn't change the behaviour here at all.

src/layers/basic.jl
darsnack: I feel like we shouldn't suggest a feature is coming. The wording makes it seem like it's definitively planned.

src/layers/basic.jl
darsnack: That's fair. I just thought the composition operator would be confusing for beginners.

src/layers/basic.jl
darsnack: I think we'd need to keep the explicit branch in anyways to not be breaking.

darsnack: At some point we should disallow it, #1673 is a good example where the right `init` would be hard to know.

mcabbott: Yes that's what I mean. It's just that making it explicit makes it look more deliberate, whereas before it silently worked and you could imagine it was deliberate or an oversight.

darsnack: Well it was previously an accident. And I think it's clear now that we want to disallow it. I believe fixing an accidental oversight counts as a bug fix in semver and not a breaking change?

src/layers/basic.jl
mcabbott: Maybe it's better to say nothing? It seems weird to me to explicitly disclaim the feature, as if you have a concrete reason not to want it. But perhaps it's also weird to dangle it as a coming attraction.

src/layers/basic.jl
```diff
@@ -1,11 +1,10 @@
...
+    Chain(name = layer, ...)
 
-Chain multiple layers / functions together, so that they are called in sequence
-on a given input.
-
-`Chain` also supports indexing and slicing, e.g. `m[2]` or `m[1:end-1]`.
-`m[1:3](x)` will calculate the output of the first three layers.
+Collects multiple layers / functions to be called in sequence
+on a given input. Supports indexing and slicing, `m[2]` or `m[1:end-1]`,
+and if names are given, `m[:name] == m[1]` etc. (but not `m.name` at present).
```
mcabbott, line 7:
```suggestion
and if names are given, `m[:name] == m[1]` etc.
```

src/layers/basic.jl
mcabbott: Yes, I think disallowing it would be fine. I made #1685 now to discuss.

src/layers/basic.jl
mcabbott: (Removed in another suggestion box, as you can't make another suggestion after committing one, TIL.)

darsnack (approved):
I think is good to go now that everything orthogonal will be handled by separate PRs.

darsnack (approved):

mcabbott:
Shall we see which of the overlapping sets of rules the robot follows?

bors r+

src/layers/basic.jl
```diff
@@ -15,36 +14,51 @@ julia> m = Chain(x -> x^2, x -> x+1);
...
 ```
 """
-struct Chain{T<:Tuple}
+struct Chain{T}
   layers::T
   Chain(xs...) = new{typeof(xs)}(xs)
+  function Chain(; kw...)
+    :layers in Base.keys(kw) && throw(ArgumentError("a Chain cannot have a named layer called `layers`"))
+    isempty(kw) && return new{Tuple{}}(())
+    new{typeof(values(kw))}(values(kw))
```
DhairyaLGandhi, line 37: Why not parameterize with the named tuple?

DhairyaLGandhi (commented):
Are we then to add #1682 on top to clean up the implementation?

src/layers/basic.jl
```diff
@@ -56,24 +70,20 @@ end
...
 extraChain(::Tuple{}, x) = ()
 
 
-
 """
     Dense(in, out, σ=identity; bias=true, init=glorot_uniform)
     Dense(W::AbstractMatrix, [bias, σ])
 
-Create a traditional `Dense` layer, whose forward pass is given by:
+Create a traditional fully-connected layer, whose forward pass is given by:
```
DhairyaLGandhi, line 86: I'd leave this be for now
```suggestion
Create a traditional `Dense` layer, whose forward pass is given by:
```

darsnack:
What's there to clean up? If you're talking about `Chain((; a = ..., b = ...))` vs `Chain(a = ..., b = ...)`, then I think the latter is more consistent with Flux's current behavior.

src/layers/basic.jl
mcabbott: Sure, this is orthogonal

src/layers/basic.jl
mcabbott: The type parameter `typeof(values(kw))` *is* the NamedTuple.

DhairyaLGandhi:
Well there are certain departures from Flux coding style in this PR (creating additional tuples, adding conditionals on keys etc), but #1682 extends the existing Chain implementation without needing the Tuples which adds some code noise which would be good to consolidate.

There's also kwarg only version in #1682. I think it's better to have the named tuple explicitly in there such that there's no ambiguity on the field-keys issue. Field access will always get you the field, and indexing will always fetch the key from the named tuple, making `:layers` a valid key in the named tuple. One must use indexing to access keys in a Chain always. Dot-Field access isn't ergonomic since it introduces the aforementioned ambiguity.

darsnack:
> Well there are certain departures from Flux coding style in this PR (creating additional tuples, adding conditionals on keys etc), but #1682 extends the existing Chain implementation without needing the Tuples which adds some code noise which would be good to consolidate.

The conditional on the keys has been discussed a couple times above. It's not necessary, but it is an easy way to allow us add dot-syntax in the future without releasing a breaking change. It's not pretty, but it is a very straightforward bit of code that will make downstream features easier. We can always remove it without a breaking change too.

Re: `Tuple`s, it's just a different way of making `applychain` work. If you want, this PR can easily be updated to use an additional dispatch path for `NamedTuple` like #1682. There's no need for a separate PR to add this.

> There's also kwarg only version in #1682. I think it's better to have the named tuple explicitly in there such that there's no ambiguity on the field-keys issue. Field access will always get you the field, and indexing will always fetch the key from the named tuple, making :layers a valid key in the named tuple. One must use indexing to access keys in a Chain always. Dot-Field access isn't ergonomic since it introduces the aforementioned ambiguity.

None of that is in this PR anymore. I really don't see how this PR and #1682 are materially different. There appear to  be only small differences like allowing `Chain(::NamedTuple)` as a constructor (could be added here but I don't see the benefit) and the `applychain` approach. I would say that feature-wise, this PR is strictly a superset of #1682.

mcabbott:
> creating additional tuples

You mean like `Chain(+, +)` creates an additional Tuple? That kind of "departures from Flux coding style"?

That "additional " tuple, an implementation detail hidden from the user, is *precisely* what this PR duplicates.

If you think this simple little mechanism needs multiple ways to invoke it, then you are of course welcome to argue for them, in follow-up issues or PRs or whatever. But they must stand on their own merits, and don't really block this one, which is the canonical simple obvious thing. Maybe there really has been a longstanding desire among some users to write `Chain((Dense(1,2,tanh), Dense(2,3))))` with extra brackets in it.

> Dot-Field access isn't ergonomic since it introduces the aforementioned ambiguity.

No, the ambiguity is *explicitly* avoided. Not one Chain in the wild has a layer named `:layers`, and none ever will, because an inner constructor forbids it. 

Dot field access is pretty ergonomic, since you don't have to type stupid brackets with your pinkies. That's why NamedTuples support it, in addition to `x[:y]`. Lots of structs pass dot field access on to constituents which aren't fields, that's literally the reason `getproperty` was introduced.

It got removed here solely because of concerns that this *may* be a (temporary) Zygote performance hurdle. It would probably have been much quicker to investigate that, than whatever it is we're all doing here a week later.

DhairyaLGandhi:
The additional tuples that I am referring to is `(c::Chain)(x) = applychain(Tuple(c.layers), x)` which can be handled with a dispatch. 

> there really has been a longstanding desire among some users to write [...] with extra brackets in it.
But #1682 also has the kwarg only constructor, so I'm not sure if this is fair.

> No, the ambiguity is explicitly avoided.
I think the argument was that there is a self consistent way to understand `:layers` if users insist. I guess if we are punting on `:layers` then 🤷

mcabbott:
> The additional tuples that I am referring to is (c::Chain)(x) = applychain(Tuple(c.layers), x) which can be handled with a dispatch.

Ok, I see Kyle was able to decode that. There are of course other ways to write it, but this seems like less characters than the others I tried. It emphasises that there is one path of functionality -- it's not dispatching to routines that do different things, just standardising.

darsnack (approved):

PR #1682, opened 2021-07-29
add NamedTuple chain constructor

DhairyaLGandhi:
Adds the `NamedTuple` constructor for `Chain`. its a much simpler design and respects all the usual expected behaviour.

```Julia
julia> c = Chain((a = Conv((3,3), 3 => 4),
                 b = Conv((3,3), 4 => 3)))
Chain(
  Conv((3, 3), 3 => 4),                 # 112 parameters
  Conv((3, 3), 4 => 3),                 # 111 parameters
)                   # Total: 4 arrays, 223 parameters, 1.574 KiB.

julia> r = rand(Float32, 28, 28, 3, 1);

julia> c(r) |> size
(24, 24, 3, 1)

julia> c[:a](r) |> size
(26, 26, 4, 1)

julia> c[:a](r) == c[1](r)
true
```

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -42,6 +42,14 @@ applychain(fs::Union{NamedTuple,Tuple}, x) = applychain(tail(fs), first(fs)(x))
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i]...)
 
+function Base.getproperty(c::Chain, k::Symbol)
+  if k === :layers
+    Base.getfield(c, :layers)
+  else
+    Base.getproperty(Base.getfield(c, :layers), k)
+  end
+end
+
 function Base.show(io::IO, c::Chain)
   print(io, "Chain(")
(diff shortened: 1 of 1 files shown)


ToucheSir:
Similar comments here as I left on #1681

DhairyaLGandhi:
As from my comments elsewhere - I don't believe that the dot-access is a good idea either. I agree that it should not be part of the API. Ditto for meddling with `propertynames` since dynamism in this regard is very small wins for potentially inconsistent behaviour.

DhairyaLGandhi:
Similarly, the concept of necessarily having a `kwargs` only constructor is a bit unnecessary for `Chain` since it's better to be explicit about what its doing. No one is going to mind if they can't add a `Tuple`, but can splat it I don't think. We have also trained users to add the layers independently.

DhairyaLGandhi:
Done

CarloLucibello:
closing as stale

PR #1683, opened 2021-07-29
use LearnBase dataset interface in the DataLoader

CarloLucibello:
Users can now define their own dataset types and as long as they define `LearnBase.getobs` and `LearnBase.nobs` the DataLoader will be able to handle them.

This PR also adopts dictionaries as one of the supported by default dataset types.


CarloLucibello:
where is github's CI? bah

src/data/dataloader.jl
```diff
@@ -102,20 +104,25 @@ function Base.length(d::DataLoader)
...
+    n = LearnBase.nobs(data[first(keys(data))])
     for i in keys(data)
-        ni = _nobs(data[i])
-        n == ni || throw(DimensionMismatch("All data inputs should have the same number of observations, i.e. size in the last dimension. " * 
-            "But data[$(repr(first(keys(data))))] ($(summary(data[1]))) has $n, while data[$(repr(i))] ($(summary(data[i]))) has $ni."))
+        ni = LearnBase.nobs(data[i])
+        n == ni || throw(DimensionMismatch("All data inputs should have the same number of observations, i.e. size in the last dimension. "))
     end
     return n
 end
```
darsnack, line 117: MLDataPattern.jl already provides these definitions (except maybe `AbstractDict`). Perhaps these belong in LearnBase.jl (or MLBase.jl in the future)?

src/data/dataloader.jl
```diff
@@ -102,20 +104,25 @@ function Base.length(d::DataLoader)
...
+                            Tuple{Vararg{AbstractArray}}, 
+                            NamedTuple{<:Any,<:Tuple{Vararg{AbstractArray}}}, 
+                            AbstractDict{<:Any,<:AbstractArray}}
 
-Base.eltype(::DataLoader{D}) where D = D
+Base.eltype(::DataLoader{D}) where {D <: WithSameEltype} = D
+
+LearnBase.getobs(data::AbstractArray, i) = data[ntuple(i -> Colon(), Val(ndims(data) - 1))..., i]
+LearnBase.getobs(data::Union{Tuple, NamedTuple}, i) = map(Base.Fix2(getobs, i), data)
+LearnBase.getobs(data::D, i) where {D<:AbstractDict} = D(k => getobs(v, i) for (k, v) in pairs(data))
```
darsnack, line 128: Same here

darsnack:
Let me try pushing through https://github.com/JuliaML/MLDataPattern.jl/pull/50 over the weekend and then we can also move some of these default definitions over to LearnBase.jl (or I could create MLBase.jl).

CarloLucibello:
It does feel like piracy indeed to have those definitions here. I'll add them to LearnBase, 
so that you should be able to remove most of the content here
https://github.com/JuliaML/MLDataPattern.jl/blob/9ba23b65f1eb00b75628448f05d7e154ede2961d/src/container.jl
while working at your PR

src/data/dataloader.jl
```diff
@@ -93,29 +95,11 @@ end
...
-        n == ni || throw(DimensionMismatch("All data inputs should have the same number of observations, i.e. size in the last dimension. " * 
-            "But data[$(repr(first(keys(data))))] ($(summary(data[1]))) has $n, while data[$(repr(i))] ($(summary(data[i]))) has $ni."))
-    end
-    return n
-end
-
-_getobs(data::AbstractArray, i) = data[ntuple(i -> Colon(), Val(ndims(data) - 1))..., i]
-_getobs(data::Union{Tuple, NamedTuple}, i) = map(Base.Fix2(_getobs, i), data)
-
-Base.eltype(::DataLoader{D}) where D = D
```
CarloLucibello, line 121: I had to remove this because it is actually wrong when batchsize=1. As a consequence though, when we collect the dataloader the get a Vector{Any}:
```julia
julia> using Flux.Data

julia> DataLoader(rand(4), batchsize=2) |> collect
2-element Vector{Any}:
 [0.5661901506001856, 0.28811712623170194]
 [0.9265719684408686, 0.05498394818719787]
```
I don't know how to fix this in a not overly complicated way. I think we will have to live with it.

CarloLucibello: Unless we can consider 
```julia
Base.eltype(d::DataLoader) = typeof(first(d))
```
an acceptable solution

darsnack: Probably this needs to be consistently done in LearnBase and MLDataPattern. Anything that implements `iterate` should specify `IteratorEltype` and consequently `eltype`.

darsnack: Meaning that if we implement the LearnBase interfaces, then we should use `MLDataPattern.eachbatch` and `MLDataPattern.shuffle`. And those containers/iterators should be implementing `eltype` stuff.

DhairyaLGandhi:
Seems like it is adding a dependency on LearnBase. Couldn't users who want to use LearnBase do so already with a loop?

src/data/dataloader.jl
CarloLucibello: ok, I'll have to take a look at that. In any case,  since the current assumptions on master are incorrect, we have to remove `eltype` for the time being, and we can merge this as it is while we wait for MLDataPattern update

darsnack:
LearnBase is a lightweight dependency. How would we be compatible with LearnBase interfaces without extending them for DataLoader?

CarloLucibello:
> Seems like it is adding a dependency on LearnBase. 

that's not a problem, LearnBase is an lean dependency

> Couldn't users who want to use LearnBase do so already with a loop?

the point here is to create a consistent dataset api for the whole ML ecosystem and have the DataLoader support it

CarloLucibello:
This shouldn't be merged until MLDataPattern is mase compatible with the new LearnBase version, otherwise FastAI.jl (that depends on MLDataPattern) won't be compatible with newest flux's releases. cc @darsnack

DhairyaLGandhi:
Since there is no hard limit on Fastai on which dataloaders to use, I don't see why we would want to make this a requirement here. Flux shouldn't assume such things so users can use what makes sense for their use case, or keep things simple and not force future work in Flux to be compatible with a particular design.

CarloLucibello:
I don't see your point. This PR is not imposing additional limitations. It is just giving the opportunity to people to define dataset types compatible with the DataLoader without having to hack into Flux's internals

DhairyaLGandhi:
Why would we need it to go via DataLoader anyway? The for loop is a simpler implementation for anyone wanting specific iterators.

CarloLucibello:
Closing as this is implemented in https://github.com/JuliaML/MLUtils.jl/pull/22

Issue #1678, opened 2021-07-26
Recurrent network interface updates/design


mkschleg:
While we were discussing #1675 and #1671 several improvements/updates to the recurrent network API came up. Instead of taking over #1675, @ToucheSir and myself thought it would be best to separate out the needed improvements into a separate issue so they can be worked on and discussed here. That way we can finish #1675, and move on with the other changes in lock-step.

- [ ] Adding CuDNN support back in, which means support for 3d arrays 
- [ ] Potentially add TimeDistributed from [TFs API](https://www.tensorflow.org/api_docs/python/tf/keras/layers/TimeDistributed)
- [ ] Adding parameters to change gate activation functions
- [ ] Bringing all cells up-to-date with #1521 

Any others I'm missing?

ToucheSir:
I added the ConvLSTM issue in the top post so we can track that here as well.

mkschleg:
I've started working on the Folded interface. It actually should be pretty easy to add without disrupting the current api, but I haven't thought through how CuDNN fits yet. I'll make a pr after we get #1675 merged, so we can iterate.

mkschleg:
I separated out CuDNNs from 3d array support. They both influence each other, but I think getting the FoldedRNN's api right needs a bit of iteration first.

mkschleg:
I added Bidirectional RNNs as per the conversation in #1686.

CarloLucibello:
Now that #1686 is merged I think we should disambiguate the interface (e.g. a 3d tensor input could be a single time step on batched grey images [width, height, batch_size] or multiple time steps on batched 1d inputs [num_features, batch_size, seq_length]).

I think that starting from next breaking release we should always assume that the input's last dimensions are batch and time,  and start introducing a deprecation path.

DhairyaLGandhi:
It's not something that needs deprecation yet.

CarloLucibello:
If we decide to go with this

> we should always assume that the input's last dimensions are batch and time, and start introducing a deprecation path.

at the very least we should immediately update the docs

ToucheSir:
:+1: for a docs update. I don't think we need to deprecate anything though, because it's not at all clear that `Recur` is the right struct to hang all this off of long term. What might be nice is if we took this opportunity to create the proposed experimental package/namespace and started prototyping there. My assumption is that most of the back and forth there will be about naming and API design, so being able to change both at a whim while still having something concrete for users to test would be nice.

dcecchini:
[...]

PR #1686, opened 2021-08-02
Adding support for folding RNNs over 3d arrays

mkschleg:
From #1678, adding a Recur like interface for a folded operation with support for 3-dimensional arrays. This is how many users expect RNNs to work if they are familiar with Pytorch and Tensorflow, and there seems to be some desire for support for this feature as per the discussion in #1671 and @jeremiedb .  This will also make a push to implementing support for the CuDNN versions of RNNs/GRUs/LSTMs more streamlined as this is the data layout that API expects. 

I did a barebones implementation to add support so we can start iterating on API.

There are several questions that I have lingering with this interface:
- ~Should we support different modes where we return all or only the last hidden state? Is there a better way to do the concat of the hidden states?~
- What kind of tests should we have? Just follow what we currently do for RNNs/LSTMs/GRUs?
- ~For the CPU version, does it make sense not to specialize on the different rnn types? We might be able to take more advantage of BLAS if we specialized on say `Folded{GRU}`.~
- ~Do we want to force the temporal dimension to be the 2nd?~
- ~Do we want this to be stateful? (i.e. allow the user to change what the starting hidden state is rather than state0).~

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -53,39 +53,6 @@ rnn.state = hidden(rnn.cell)
 reset!(m::Recur) = (m.state = m.cell.state0)
 reset!(m) = foreach(reset!, functor(m)[1])
 
-"""
-    FoldedRecur
-
-We fold over the second dimension as the time dimension, and return all the new hidden states concatenated on the time dimension.
-"""
-struct FoldedRecur{C}
-    cell::C
-end
-
-Flux.@functor FoldedRecur
-trainable(a::FoldedRecur) = (a.cell,)
-(m::FoldedRecur)(x) = m.cell(m.cell.state0, x)
-
-FoldedRNN(args...; kwargs...) = 
-    FoldedRecur(Flux.RNNCell(args...; kwargs...))
-
-function (m::Flux.RNNCell{F,A,V,<:AbstractMatrix{T}})(h, x::AbstractArray{<:Number, 3}) where {F,A,V,T}
-    σ, Wi, Wh, b = m.σ, m.Wi, m.Wh, m.b
-
-    # stride across the temporal dimension, i.e. 2
-    h_all = [h]
-    for t in 1:size(x, 2)
-        h = σ.(Wi*x[:, t, :] .+ Wh*h_all[end] .+ b)
-        push!(h_all, h)
-    end
-    sz = size(x)
-    h_ret = cat(reshape.(h_all[2:end], :, 1, sz[3:end]...)..., dims=2)
-
-    # h_ret = h_all[2:end]
-    return h_ret, h_ret
-end
-
-
 
 # TODO remove in v0.13
 function Base.getproperty(m::Recur, sym::Symbol)
@@ -100,6 +67,35 @@ function Base.getproperty(m::Recur, sym::Symbol)
 end
 
 flip(f, xs) = reverse(f.(reverse(xs)))
+
+
+"""
+    FoldedRecur
+
+We fold over the second dimension as the time dimension, and return all the new hidden states concatenated on the time dimension.
+"""
+struct FoldedRecur{C}
+  cell::C
+end
+
+Flux.@functor FoldedRecur
+trainable(a::FoldedRecur) = (a.cell,)
+
+function (m::FoldedRecur)(x::AbstractArray{<:Number, 3})
+  # stride across temporal dimension
+
+  h = m.cell.state0
+  h_all = Vector{typeof(h)}(undef, size(x, 2))
+  for t in 1:size(x, 2)
+      h, h_out = m.cell(h, x[:, t, :])
+      push!(h_all, h_out)
+  end
+  sz = size(x)
+  h_ret = cat(reshape.(h_all[2:end], :, 1, sz[3:end]...)..., dims=2)
+  return h_ret
+
+end
+
 
 # Vanilla RNN
 
@@ -136,6 +132,7 @@ The most basic recurrent layer; essentially acts as a `Dense` layer, but with th
 output fed back into the input each time step.
 """
 RNN(a...; ka...) = Recur(RNNCell(a...; ka...))
+FoldedRNN(args...; kwargs...) = FoldedRecur(Flux.RNNCell(args...; kwargs...))
 Recur(m::RNNCell) = Recur(m, m.state0)
 
 # TODO remove in v0.13
@@ -196,6 +193,7 @@ See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
 """
 LSTM(a...; ka...) = Recur(LSTMCell(a...; ka...))
+FoldedLSTM(args...; kwargs...) = FoldedRecur(Flux.LSTMCell(args...; kwargs...))
 Recur(m::LSTMCell) = Recur(m, m.state0)
 
 # TODO remove in v0.13
@@ -261,6 +259,7 @@ See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
 """
 GRU(a...; ka...) = Recur(GRUCell(a...; ka...))
+FoldedGRU(args...; kwargs...) = FoldedRecur(Flux.GRUCell(args...; kwargs...))
 Recur(m::GRUCell) = Recur(m, m.state0)
 
 # TODO remove in v0.13
@@ -315,6 +314,7 @@ See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
 """
 GRUv3(a...; ka...) = Recur(GRUv3Cell(a...; ka...))
+FoldedGRUv3(args...; kwargs...) = FoldedRecur(Flux.GRUv3Cell(args...; kwargs...))
 Recur(m::GRUv3Cell) = Recur(m, m.state0)
 
 
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -91,7 +91,7 @@ function (m::FoldedRecur)(x::AbstractArray{<:Number, 3})
       push!(h_all, h_out)
   end
   sz = size(x)
-  h_ret = cat(reshape.(h_all[2:end], :, 1, sz[3:end]...)..., dims=2)
+  h_ret = cat(reshape.(h_all, :, 1, sz[3:end]...)..., dims=2)
   return h_ret
 
 end
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -68,42 +68,10 @@ end
 
 flip(f, xs) = reverse(f.(reverse(xs)))
 
-
-"""
-    FoldedRecur
-
-We fold over the second dimension as the time dimension, and return all the new hidden states concatenated on the time dimension.
-
-TODO: Figure out how to use CuDNN for RNN, GRU, LSTM.
-
-"""
-struct FoldedRecur{C}
-    cell::C
-end
-
-Flux.@functor FoldedRecur
-trainable(a::FoldedRecur) = (a.cell,)
-
-# currently RNNs are only usable with 3-d arrays.
-function (m::FoldedRecur)(x::AbstractArray{<:Number, 3})
-
-  # stride across temporal dimension
-  h = m.cell.state0
-  h_all = if h isa Tuple
-      Vector{typeof(h[1])}(undef, size(x, 2))
-  else
-      Vector{typeof(h)}(undef, size(x, 2))
-  end
-
-  for t in axes(x, 2)
-      h, h_out = m.cell(h, x[:, t, :])
-      h_all[t] = h_out
-  end
-
+function (m::Recur)(x::AbstractArray{<:Number, 3})
+  h_ret = mapslices(m, x, dims=[1, 3])
   sz = size(x)
-  h_ret = cat(reshape.(h_all, :, 1, sz[3])..., dims=2)
   return h_ret
-
 end
 
 
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -82,7 +82,7 @@ end
 flip(f, xs) = reverse(f.(reverse(xs)))
 
 function (m::Recur)(x::AbstractArray{T, 3}) where T
-  h = [m(x[:, :, i]) for i in 1:size(x, 3)]
+  h = [m(view(x, :, :, i)) for i in 1:size(x, 3)]
   sze = size(h[1])
   reshape(reduce(hcat, h), sze[1], sze[2], length(h))
 end
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -68,8 +68,8 @@ end
 
 flip(f, xs) = reverse(f.(reverse(xs)))
 
-function (m::Recur)(x::AbstractArray{<:Number, 3})
-  h_ret = mapslices(m, x, dims=[1, 3])
+function (m::Recur)(x::AbstractArray{T, 3}) where T
+  h_ret = mapslices(m, x, dims = (1, 3))
   sz = size(x)
   return h_ret
 end
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -69,11 +69,8 @@ end
 flip(f, xs) = reverse(f.(reverse(xs)))
 
 function (m::Recur)(x::AbstractArray{T, 3}) where T
-  h_ret = mapslices(m, x, dims = (1, 3))
-  sz = size(x)
-  return h_ret
+    stack([m(x[:,:,i]) for i in 1:size(x, 3)], 3)
 end
-
 
 # Vanilla RNN
 
@@ -110,7 +107,6 @@ The most basic recurrent layer; essentially acts as a `Dense` layer, but with th
 output fed back into the input each time step.
 """
 RNN(a...; ka...) = Recur(RNNCell(a...; ka...))
-FoldedRNN(args...; kwargs...) = FoldedRecur(Flux.RNNCell(args...; kwargs...))
 Recur(m::RNNCell) = Recur(m, m.state0)
 
 # TODO remove in v0.13
@@ -171,7 +167,6 @@ See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
 """
 LSTM(a...; ka...) = Recur(LSTMCell(a...; ka...))
-FoldedLSTM(args...; kwargs...) = FoldedRecur(Flux.LSTMCell(args...; kwargs...))
 Recur(m::LSTMCell) = Recur(m, m.state0)
 
 # TODO remove in v0.13
@@ -237,7 +232,6 @@ See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
 """
 GRU(a...; ka...) = Recur(GRUCell(a...; ka...))
-FoldedGRU(args...; kwargs...) = FoldedRecur(Flux.GRUCell(args...; kwargs...))
 Recur(m::GRUCell) = Recur(m, m.state0)
 
 # TODO remove in v0.13
@@ -292,7 +286,6 @@ See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
 """
 GRUv3(a...; ka...) = Recur(GRUv3Cell(a...; ka...))
-FoldedGRUv3(args...; kwargs...) = FoldedRecur(Flux.GRUv3Cell(args...; kwargs...))
 Recur(m::GRUv3Cell) = Recur(m, m.state0)
 
 
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -53,6 +53,40 @@ rnn.state = hidden(rnn.cell)
 reset!(m::Recur) = (m.state = m.cell.state0)
 reset!(m) = foreach(reset!, functor(m)[1])
 
+"""
+    FoldedRecur
+
+We fold over the second dimension as the time dimension, and return all the new hidden states concatenated on the time dimension.
+"""
+struct FoldedRecur{C}
+    cell::C
+end
+
+Flux.@functor FoldedRecur
+trainable(a::FoldedRecur) = (a.cell,)
+(m::FoldedRecur)(x) = m.cell(m.cell.state0, x)
+
+FoldedRNN(args...; kwargs...) = 
+    FoldedRecur(Flux.RNNCell(args...; kwargs...))
+
+function (m::Flux.RNNCell{F,A,V,<:AbstractMatrix{T}})(h, x::AbstractArray{<:Number, 3}) where {F,A,V,T}
+    σ, Wi, Wh, b = m.σ, m.Wi, m.Wh, m.b
+
+    # stride across the temporal dimension, i.e. 2
+    h_all = [h]
+    for t in 1:size(x, 2)
+        h = σ.(Wi*x[:, t, :] .+ Wh*h_all[end] .+ b)
+        push!(h_all, h)
+    end
+    sz = size(x)
+    h_ret = cat(reshape.(h_all[2:end], :, 1, sz[3:end]...)..., dims=2)
+
+    # h_ret = h_all[2:end]
+    return h_ret, h_ret
+end
+
+
+
 # TODO remove in v0.13
 function Base.getproperty(m::Recur, sym::Symbol)
   if sym === :init
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -24,6 +24,19 @@ rnn.state   # 5
 rnn.(1:10)  # apply to a sequence
 rnn.state   # 60
 ```
+
+Folding over a 3d Array of dimensions `(features, batch, time)` is also supported:
+
+```julia
+accum(h, x) = (h .+ x, x)
+rnn = Flux.Recur(accum, zeros(Int, 1, 1))
+rnn([2])                    # 2
+rnn([3])                    # 3
+rnn.state                   # 5
+rnn(reshape(1:10, 1, 1, :)) # apply to a sequence of (features, batch, time)
+rnn.state                   # 60
+```
+
 """
 mutable struct Recur{T,S}
   cell::T
@@ -69,9 +82,9 @@ end
 flip(f, xs) = reverse(f.(reverse(xs)))
 
 function (m::Recur)(x::AbstractArray{T, 3}) where T
-    h = [m(x[:, :, i]) for i in 1:size(x, 3)]
-    sze = size(h[1])
-    reshape(reduce(hcat, h), sze[1], sze[2], length(h))
+  h = [m(x[:, :, i]) for i in 1:size(x, 3)]
+  sze = size(h[1])
+  reshape(reduce(hcat, h), sze[1], sze[2], length(h))
 end
 
 # Vanilla RNN
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -73,17 +73,21 @@ flip(f, xs) = reverse(f.(reverse(xs)))
     FoldedRecur
 
 We fold over the second dimension as the time dimension, and return all the new hidden states concatenated on the time dimension.
+
+TODO: Figure out how to use CuDNN for RNN, GRU, LSTM.
+
 """
 struct FoldedRecur{C}
-  cell::C
+    cell::C
 end
 
 Flux.@functor FoldedRecur
 trainable(a::FoldedRecur) = (a.cell,)
 
+# currently RNNs are only usable with 3-d arrays.
 function (m::FoldedRecur)(x::AbstractArray{<:Number, 3})
-  # stride across temporal dimension
 
+  # stride across temporal dimension
   h = m.cell.state0
   h_all = Vector{typeof(h)}(undef, size(x, 2))
   for t in 1:size(x, 2)
@@ -91,7 +95,7 @@ function (m::FoldedRecur)(x::AbstractArray{<:Number, 3})
       push!(h_all, h_out)
   end
   sz = size(x)
-  h_ret = cat(reshape.(h_all, :, 1, sz[3:end]...)..., dims=2)
+  h_ret = cat(reshape.(h_all, :, 1, sz[3])..., dims=2)
   return h_ret
 
 end
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -69,7 +69,9 @@ end
 flip(f, xs) = reverse(f.(reverse(xs)))
 
 function (m::Recur)(x::AbstractArray{T, 3}) where T
-    stack([m(x[:,:,i]) for i in 1:size(x, 3)], 3)
+    h = [m(x[:, :, i]) for i in 1:size(x, 3)]
+    sze = size(h[1])
+    reshape(reduce(hcat, h), sze[1], sze[2], length(h))
 end
 
 # Vanilla RNN
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -95,10 +95,11 @@ function (m::FoldedRecur)(x::AbstractArray{<:Number, 3})
       Vector{typeof(h)}(undef, size(x, 2))
   end
 
-  for t in 1:size(x, 2)
+  for t in axes(x, 2)
       h, h_out = m.cell(h, x[:, t, :])
       h_all[t] = h_out
   end
+
   sz = size(x)
   h_ret = cat(reshape.(h_all, :, 1, sz[3])..., dims=2)
   return h_ret
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -89,10 +89,15 @@ function (m::FoldedRecur)(x::AbstractArray{<:Number, 3})
 
   # stride across temporal dimension
   h = m.cell.state0
-  h_all = Vector{typeof(h)}(undef, size(x, 2))
+  h_all = if h isa Tuple
+      Vector{typeof(h[1])}(undef, size(x, 2))
+  else
+      Vector{typeof(h)}(undef, size(x, 2))
+  end
+
   for t in 1:size(x, 2)
       h, h_out = m.cell(h, x[:, t, :])
-      push!(h_all, h_out)
+      h_all[t] = h_out
   end
   sz = size(x)
   h_ret = cat(reshape.(h_all, :, 1, sz[3])..., dims=2)
--- a/test/cuda/curnn.jl
+++ b/test/cuda/curnn.jl
@@ -63,7 +63,5 @@ end
 
     cufy = (curnn(cufx); curnn(cufx))
     @test fy ≈ collect(cufy)
-
-      
   end
 end
--- a/test/cuda/curnn.jl
+++ b/test/cuda/curnn.jl
@@ -53,6 +53,17 @@ end
     y = (rnn(ohx); rnn(ohx))
 
     cuy = (curnn(cuohx); curnn(cuohx))
-    @test y ≈ collect(cuy)  
+    @test y ≈ collect(cuy)
+
+    Flux.reset!(rnn)
+    Flux.reset!(curnn)
+    fx = rand(Float32, 10, batch_size, 3)
+    cufx = gpu(fx)
+    fy = (rnn(fx); rnn(fx))
+
+    cufy = (curnn(cufx); curnn(cufx))
+    @test fy ≈ collect(cufy)
+
+      
   end
 end
--- a/test/layers/recurrent.jl
+++ b/test/layers/recurrent.jl
@@ -44,23 +44,28 @@ end
 
 @testset "BPTT-3D" begin
   seq = rand(Float32, (2, 1, 3))
-  for r ∈ [RNN,]
-    rnn = r(2, 3)
-    Flux.reset!(rnn)
-    grads_seq = gradient(Flux.params(rnn)) do
-        sum(rnn(seq)[:, :, 3])
-    end
-    Flux.reset!(rnn);
-    bptt = gradient(Wh -> sum(tanh.(rnn.cell.Wi * seq[:, :, 3] + Wh *
-                                  tanh.(rnn.cell.Wi * seq[:, :, 2] + Wh *
-                                        tanh.(rnn.cell.Wi * seq[:, :, 1] +
-                                            Wh * rnn.cell.state0
-                                        + rnn.cell.b)
-                                  + rnn.cell.b)
-                            + rnn.cell.b)),
-                    rnn.cell.Wh)
-    @test grads_seq[rnn.cell.Wh] ≈ bptt[1]
-end
+  rnn = RNN(2, 3)
+  Flux.reset!(rnn)
+  grads_seq = gradient(Flux.params(rnn)) do
+    sum(rnn(seq)[:, :, 3])
+  end
+  Flux.reset!(rnn);
+  bptt = gradient(rnn.cell.Wh) do Wh
+    # calculate state 1
+    s1 = tanh.(rnn.cell.Wi * seq[:, :, 1] +
+               Wh * rnn.cell.state0 +
+               rnn.cell.b)
+    #calculate state 2
+    s2 = tanh.(rnn.cell.Wi * seq[:, :, 2] +
+               Wh * s1 +
+               rnn.cell.b)
+    #calculate state 3
+    s3 = tanh.(rnn.cell.Wi * seq[:, :, 3] +
+               Wh * s2 +
+               rnn.cell.b)
+    sum(s3) # loss is sum of state 3
+  end
+  @test grads_seq[rnn.cell.Wh] ≈ bptt[1]
 end
 
 @testset "RNN-shapes" begin
--- a/test/layers/recurrent.jl
+++ b/test/layers/recurrent.jl
@@ -69,17 +69,17 @@ end
     m2 = R(3, 5)
     m3 = R(3, 5)
     x1 = rand(Float32, 3)
-    x2 = rand(Float32,3,1)
-    x3 = rand(Float32,3,1,2)
+    x2 = rand(Float32, 3, 1)
+    x3 = rand(Float32, 3, 1, 2)
     Flux.reset!(m1)
     Flux.reset!(m2)
     Flux.reset!(m3)
     @test size(m1(x1)) == (5,)
     @test size(m1(x1)) == (5,) # repeat in case of effect from change in state shape
-    @test size(m2(x2)) == (5,1)
-    @test size(m2(x2)) == (5,1)
-    @test size(m3(x3)) == (5,1,2)
-    @test size(m3(x3)) == (5,1,2)
+    @test size(m2(x2)) == (5, 1)
+    @test size(m2(x2)) == (5, 1)
+    @test size(m3(x3)) == (5, 1, 2)
+    @test size(m3(x3)) == (5, 1, 2)
   end
 end
 
--- a/test/layers/recurrent.jl
+++ b/test/layers/recurrent.jl
@@ -42,26 +42,52 @@ end
 end
 end
 
-@testset "RNN-shapes" begin
-    @testset for R in [RNN, GRU, LSTM, GRUv3]
-        m1 = R(3, 5)
-        m2 = R(3, 5)
-        x1 = rand(Float32, 3)
-        x2 = rand(Float32,3,1)
-        Flux.reset!(m1)
-        Flux.reset!(m2)
-        @test size(m1(x1)) == (5,)
-        @test size(m1(x1)) == (5,) # repeat in case of effect from change in state shape
-        @test size(m2(x2)) == (5,1)
-        @test size(m2(x2)) == (5,1)
+@testset "BPTT-3D" begin
+  seq = rand(Float32, (2, 1, 3))
+  for r ∈ [RNN,]
+    rnn = r(2, 3)
+    Flux.reset!(rnn)
+    grads_seq = gradient(Flux.params(rnn)) do
+        sum(rnn(seq)[:, :, 3])
     end
+    Flux.reset!(rnn);
+    bptt = gradient(Wh -> sum(tanh.(rnn.cell.Wi * seq[:, :, 3] + Wh *
+                                  tanh.(rnn.cell.Wi * seq[:, :, 2] + Wh *
+                                        tanh.(rnn.cell.Wi * seq[:, :, 1] +
+                                            Wh * rnn.cell.state0
+                                        + rnn.cell.b)
+                                  + rnn.cell.b)
+                            + rnn.cell.b)),
+                    rnn.cell.Wh)
+    @test grads_seq[rnn.cell.Wh] ≈ bptt[1]
+end
+end
+
+@testset "RNN-shapes" begin
+  @testset for R in [RNN, GRU, LSTM, GRUv3]
+    m1 = R(3, 5)
+    m2 = R(3, 5)
+    m3 = R(3, 5)
+    x1 = rand(Float32, 3)
+    x2 = rand(Float32,3,1)
+    x3 = rand(Float32,3,1,2)
+    Flux.reset!(m1)
+    Flux.reset!(m2)
+    Flux.reset!(m3)
+    @test size(m1(x1)) == (5,)
+    @test size(m1(x1)) == (5,) # repeat in case of effect from change in state shape
+    @test size(m2(x2)) == (5,1)
+    @test size(m2(x2)) == (5,1)
+    @test size(m3(x3)) == (5,1,2)
+    @test size(m3(x3)) == (5,1,2)
+  end
 end
 
 @testset "RNN-input-state-eltypes" begin
   @testset for R in [RNN, GRU, LSTM, GRUv3]
-      m = R(3, 5)
-      x = rand(Float64, 3, 1)
-      Flux.reset!(m)
-      @test_throws MethodError m(x)
+    m = R(3, 5)
+    x = rand(Float64, 3, 1)
+    Flux.reset!(m)
+    @test_throws MethodError m(x)
   end
 end


src/layers/recurrent.jl
```diff
@@ -67,6 +68,45 @@ end
...
+  end
+
+  for t in axes(x, 2)
+      h, h_out = m.cell(h, x[:, t, :])
+      h_all[t] = h_out
+  end
+
+  sz = size(x)
+  h_ret = cat(reshape.(h_all, :, 1, sz[3])..., dims=2)
+  return h_ret
```
ToucheSir, line 105: Ideally all of this could be accomplished with a call to `mapslices` on a `Recur`, but unfortunately the guarantee of that being left-right is implementation-only.

src/layers/recurrent.jl
```diff
@@ -67,6 +68,45 @@ end
...
+
+"""
+    FoldedRecur
+
+We fold over the second dimension as the time dimension, and return all the new hidden states concatenated on the time dimension.
+
+TODO: Figure out how to use CuDNN for RNN, GRU, LSTM.
+
+"""
+struct FoldedRecur{C}
```
ToucheSir, line 80: I feel like this _is_ `TimeDistributed`, as it always operates seq-seq and doesn't take a reduction operation like the fold functions do. Thoughts?

mkschleg: I agree, this is effectively `TimeDistributed`. The only difference is this needs to guarantee ordering, whereas `TimeDistributed` could be in any order. So, there might be some potential for parallelization in the future for `TimeDistributed` but not for `FoldedRecurs`. But we could definitely combine the two if we want.

src/layers/recurrent.jl
mkschleg: Right. I thought about that. We could implement something like `mapslices` in flux to use and guarantee order for the future, but I'm not sure that would be within scope of flux. I decided on this as a compromise. We could also always re-use recur here and have a specialized call on 3d arrays.

DhairyaLGandhi: Best to use Recur

mkschleg: Sounds good. I'll go back and remove `FoldedRecur` and just specialize over 3d arrays because both @ToucheSir and you think that would be better. I agree with this move after thinking about it.

mkschleg:
I've reimplemented using `mapslices` for now.  As @ToucheSir mentioned this is not guaranteed to be left-right in the future. So likely an alternative would be better. Left it for now as the implementation is now a single function of 3 lines. Looks quite nice.

src/layers/recurrent.jl
ToucheSir: After thinking about it a bit more, I wonder if perhaps `TimeDistributed` is more of a special case of `Parallel`? Specifically `Parallel(Flux.stack, submodel)`. Trying to think of what a good name for something that is order-dependent is difficult, however. The easiest would be `Map[LR]` and `Fold[LR]`, but I'm always leery of control flow like operations being introduced.

mkschleg: Hm. I'm not sure how `Parallel(Flux.stack, submodel)` would work. I can't get it to work at all from some testing. But my thought was that `TimeDistributed` would iterate a submodel over the "temporal" dimension. So if a layer doesn't support arbitrary dimensional data like Dense it would handle that. But maybe the better thing to do was make sure all the layers support that?

src/layers/recurrent.jl
```diff
@@ -67,6 +68,13 @@ end
 
 flip(f, xs) = reverse(f.(reverse(xs)))
 
+function (m::Recur)(x::AbstractArray{<:Number, 3})
```
DhairyaLGandhi, line 71:
```suggestion
function (m::Recur)(x::AbstractArray{T, 3}) where T
```

src/layers/recurrent.jl
```diff
@@ -67,6 +68,13 @@ end
...
+  h_ret = mapslices(m, x, dims=[1, 3])
```
DhairyaLGandhi, line 72:
```suggestion
  h_ret = mapslices(m, x, dims = (1, 3))
```
Temporaries

src/layers/recurrent.jl
```diff
@@ -162,6 +171,7 @@ See [this article](https://colah.github.io/posts/2015-08-Understanding-LSTMs/)
 for a good overview of the internals.
 """
 LSTM(a...; ka...) = Recur(LSTMCell(a...; ka...))
+FoldedLSTM(args...; kwargs...) = FoldedRecur(Flux.LSTMCell(args...; kwargs...))
```
DhairyaLGandhi, line 174: Seems nicer to not have this here api wise 

```Julia
Folded(LSTM(...))
```

Makes it clear that the folding representation is generic and any old recurrent cell can be thrown in automatically. We would need to have cudnn paths for these too and possibly make cudnn the default

src/layers/recurrent.jl
DhairyaLGandhi: You might need `(x...) -> cat(x..., dims = 3)`

DhairyaLGandhi: Also, `Dense` can work with extra dimensions already too.

src/layers/recurrent.jl
mkschleg: Oops! Forgot to remove that! That was a part of the old api.

src/layers/recurrent.jl
ToucheSir: `Parallel(Flux.stack, submodel)` is not going to work exactly as written right now, you'd have to use `eachslice` and a couple lambdas at the every least to get it to cooperate. My point is that the name "TimeDistributed" is really a misnomer, as it's actually just applying a network across the each slice of the first axis of the input array simultaneously (i.e. in parallel). That axis may represent time, but it doesn't necessarily have to.

ToucheSir: RE all layers supporting higher dimensions, that helps for some use cases but not all of them. For example applying a Conv -> RNNCell -> Dense chain on each timestep of a sequence (`[D(R(C(x1))), D(R(C(x2))), D(R(C(x3)))]`) is distinct from applying the each layer across the entire sequence in series (`D(R(C(X)))`). Of course we could decide to not support both, but as things stand we would need to accomodate the former pattern for backwards compatibility.

DhairyaLGandhi: I wouldn't want to drop support for either, but rather make leave it to users what they expect to do. As it stands, those are completely distinct models, and that seems reasonable.

jeremiedb:
Regarding the dimensions, my understanding is that the default for fused RNN operations (ex CUDNN) is to operate on input which is layout as `[features, batch_size, seq_length]`. See for example [Knet](https://github.com/denizyuret/Knet.jl/blob/6a2eccbd4c3476167205714da7e48f4132c2eb9b/src/ops20/rnn.jl#L17)(same with other frameworks). As such, I'd suggest to default to that behavior and assumes that time dimension is the 3rd.

src/layers/recurrent.jl
```diff
@@ -67,6 +68,13 @@ end
 
 flip(f, xs) = reverse(f.(reverse(xs)))
 
+function (m::Recur)(x::AbstractArray{<:Number, 3})
+  h_ret = mapslices(m, x, dims=[1, 3])
+  sz = size(x)
```
darsnack, line 73: Planning on doing some reshape with this? If not, can be deleted.

src/layers/recurrent.jl
darsnack: @ToucheSir your "special case of `Parallel`" argument makes me think this really should be expressed as something like `Folded` instead of specializing on the number of dimensions. Semantically, I prefer that to the current implementation.

src/layers/recurrent.jl
jeremiedb: Regarding the `mapslices` approach, had you tested on GPU? On a local test, it appears like CUDA.jl hasn't it implemented, resullting slow scalar operations. 
However, if going with: 
```julia
stack([m(x[:,:,i]) for i in 1:size(x, 3)], 3)
```

It did work fine on GPU, plus it does provide safety for the operation order.

ToucheSir: How AD-friendly is the comprehension? That and the splat in `stack` would be my biggest worries with this, but if they're non issues then it looks to be quite the elegant solution.

src/layers/recurrent.jl
jeremiedb: Regarding the ability to cover all use cases, I say potential upcoming conflict with the desire to integrate 3D seq with CUDNN's fused RNNs. The reason being that CUDNN's RNN directly operates on the entire sequence (so the `D(R(C(X)))` use case). Also, to have the `FoldedRecur` to mimic CUDNN's RNN, it will require some higher level utilities, notably the `bidirectional` and `seq_length` options which aren't relevant at the single cell level but only at the higher 3D level. So if we are to eventually have a 3D RNN compatible with CUDNN, shouldn't a distinct struct be defined so as to support those higher level functionalities?
In short, if keeping on the local implementation, then the current drection seems good, but if CUDNN support is in the plan, then I see some bumps ahead.

src/layers/recurrent.jl
jeremiedb: Not sure if I screw up something somewhere, but it's actually the mapslices that seems to hust the AD (with the infamous mutating arrays not supported!). Hope the following is self explanatory, essentially, `rnn1` is the comprehension approach while `rnn2` is the mapslices ones.  

```julia
using Flux
using Flux: stack, mse

m = RNN(64, 128)
d = Dense(128, 1)
x = rand(Float32, 64, 32, 16)
y = rand(Float32, 1, 32, 16)

function rnn1(x, m)
    stack([m(x[:,:,i]) for i in 1:size(x, 3)], 3)
end

function rnn2(x, m)
    mapslices(m, x, dims=[1, 2])
end

@time y1 = [m(x[:,:,i]) for i in 1:size(x, 3)];
@time y2 = mapslices(m, x, dims=[1, 2]);

@time rnn1(x, m);
@time rnn2(x, m);

loss1(x,y) = mse(rnn1(x, m) |> d, y)
loss1(x,y)
ps = Flux.params(m)
@time grads = gradient(() -> loss1(x, y), ps)

loss2(x,y) = mse(rnn2(x, m) |> d, y)
loss2(x,y)
ps = Flux.params(m)
@time grads = gradient(() -> loss2(x, y), ps)
```

DhairyaLGandhi: Not super ad friendly, since the splat is shown up as iteration, and also something cuda would not be happy with. Maybe a view based approach would work for both although incur some memory copies.

src/layers/recurrent.jl
DhairyaLGandhi: @mkschleg mind if commit the two review changes.

src/layers/recurrent.jl
jeremiedb: @DhairyaLGandhi I'm not observing any non AD friendly effect, at least not when compared to current 2D based  approach. 

```julia
########################
# 2-D approach
########################
x = [rand(Float32, 64,32) for i in 1:16]
y = rand(Float32, 1,32, 16)

m = Chain(RNN(64,128), Dense(128,1))
m(x[1])
loss_v1(x,y) = mse(stack(map(m, x), 3), y)

ps = Flux.params(m)
julia> @btime grads = gradient(() -> loss_v1($x, $y), $ps)
  7.646 ms (6801 allocations: 5.68 MiB)
```

The above is quite similar in speed but does have less allocations than:

```jullia
function rnn1B(x, m)
    stack([m(view(x, :,:,i)) for i in 1:size(x, 3)], 3)
end

loss1B(x,y) = mse(rnn1B(x, m) |> d, y)
@time loss1B(x,y);
ps = Flux.params(m)
julia> @btime grads = gradient(() -> loss1B($x, $y), $ps)
  7.753 ms (5213 allocations: 9.95 MiB)
```
 
(and notice that mapslices on previous example crashes on the gradient call). 

And has previously mentionned, it does  work fine on CUDA: 

```julia
using CUDA
x_g = x |> gpu
y_g = y |> gpu
m_g = m |> gpu
d_g = d |> gpu

loss1_g(x,y) = mse(rnn1(x, m_g) |> d_g, y)
@time loss1_g(x_g, y_g);
ps = Flux.params(m_g)
CUDA.@time grads = gradient(() -> loss1_g(x_g, y_g), ps)
```

src/layers/recurrent.jl
ToucheSir: Not necessarily. Wrapping a cell or Recur in this proposed Fold/Map/Scan (in XLA/Jax parlance) struct would be sufficient to support cudnn over 3D+ inputs. All other cases (vector of matrices, wrapping an arbitrary function or Chain) then gracefully degrade to simple iteration.

src/layers/recurrent.jl
mkschleg: Great! Thanks for all the advice and work on this @jeremiedb! The current `mapslices` was definitely just stand-in because I wasn't sure how else to go about it and haven't had time to iterate on my own. But I agree the stack operation might be the best option right now. I also have some concerns about the splat in stack, but I'm not sure of another way to concatenate the results together. I'll implement this and then go through the other suggested changes as well.

mkschleg:
I agree to moving the time index to the 3rd.

src/layers/recurrent.jl
mkschleg: I removed these.

src/layers/recurrent.jl
mkschleg: @darsnack I removed the superfluous size

src/layers/recurrent.jl
mkschleg: But wouldn't we need more information if you wanted to have a bidirectional RNN go through CUDNN? I think the goal of this direction is to eventually get to (mostly full) CUDNN support.

mkschleg: But maybe this is feature creep. We can address this in another PR and focus on getting basic 3d support here.

ToucheSir: That's why I posted `Map[LR]` and `Fold[LR]` as strawman names above. The question becomes one of API design: do you have a different struct for each direction (left, right, bidirectional) and each reduction type (last output, seq2seq), or do you change some of those into parameters (and then what values do you use to control those parameters)?

> I think the goal of this direction is to eventually get to (mostly full) CUDNN support

I'm not sure that's possible as long as we support `Vector{Array}` as an input format. As such, a more realistic goal is designing the interface so we can lean on dispatch as much as possible for identifying paths to send off to CuDNN or other backends.

For argument's sake, take the example of a seq2seq `BiDirectional{M}` layer which is parameterized by the model type. The dispatches could look something like this:
- `(::BiDirectional)(::AbstractVector{AbstractArray})` -> default implementation with iteration
- `(::BiDirectional{<:RNNCell})(::CuArray{N})`, where `N > 3` and the cell type is supported -> CuDNN
- `(::BiDirectional{<:RNNCell})(::CuArray{N})`, where `N > 3` and the cell type not supported -> default

mkschleg: Right. That is what I meant. I didn't mean "only CuDNN support". I get what you mean now! Yes this is a better/clearer interface than the heavy kwargs based ones from pytorch/tf. I agree with this and so for this PR we should just constrain to 3d support for `Recur`.

jeremiedb: Thanks @ToucheSir  for the clarification with the `BiDirectional` struct. Just for precision, I'd assume it would be `m::BiDirectional{<:Recur[T,S]}` rather than `::BiDirectional{<:RNNCell}` right? So that the function does essentially concat `m(x)` and `m(reverse(x))`?

I'm afraid this you still left some thorns with CUDNN integration as the on of its key feature for performance isn't only the ingestion of 3D arrays, but also the handling of a stack of layers.
So for exploiting CUDNN's RNN, it would need something like

```julia
m = Chain(
    Dense(64, 32),
    CUDNN_RNN(32, 128, num_layers=3), 
    Dense(128, 1))
``` 

instead of: 
```julia
m = Chain(
    Dense(64, 32),
    RNN(32, 128), 
    RNN(128, 128), 
    RNN(128, 128), 
    Dense(128, 1))
```

One might think of the option to have `Recur` adapted to handle a collection of cells: 
```julia
RNN(RNNCell(32, 128), 
    RNNCell(128, 128), 
    RNNCell(128, 128))
```

But I'd still see an issue for support distpach to CUDNN since AFAIK, it's not possible to assign a varying number of hidden neurons between the layers. Therefore, dispatch would also have to depend on the consistency of the number of neurons in each layer (Flux can handled varying number as the input to each layer is currently reprojected in Flux's cells, which isn't the case in CUDNN's RNN, except for the 1st layer). 

As a sidenote, if we add a `BiDirectional` layer, would there be an elegant way to handle the little akwardness that would result from the output size being 2 times the last cell size: 

```julia
m = Chain(
    Dense(64, 32),
    BiDirectional(RNN(32, 128)), 
    Dense(256, 1)) # 256 since each 128 direction is stacked
```

Given this context, I still feel like trying to fit in CUDNN's RNN within current design will likely result in weird contorsions. IMO, it shouldn't be considered as a universal layer as can be CUDNN's convolutions. I think it would be most efficient to not be bothered by its restrictive API and rather develop on the lean ground that Flux provide. And if needed, then I think a seperate dedicated struct respecting its convention should be developped, akin to the one in Knet. 

I know @AzamatB has dealt with recurrent stuff, hist perspective could be helpful.

ToucheSir: > Just for precision, I'd assume it would be `m::BiDirectional{<:Recur[T,S]}` rather than `::BiDirectional{<:RNNCell}` right?

It would probably be `m::BiDirectional{T} where T`, because T could be anything from an RNN cell to a `Chain` to some arbitrary function. That does bring up the question of how to construct one of these, but at the very least `BiDirectional(RNNCell(...), RNNCell(...))` would work. I see no point in getting `Recur` involved, because its ability to store state isn't required when the entire sequence is given.

As for supporting stacked layers, some combination of dispatch and runtime handoff seems necessary (again, no `Recur`s though). Dispatching on `Chain{<:NTuple{N, RNNCell} where N` ** would be one option, with a fallback to running each layer sequentially if the dimensions don't match. Another option is to have a dedicated layer for homogenous stacks which validates all of its children during construction.

**  + LSTMCell, GRUCell, etc.

jeremiedb: > no point in getting `Recur` involved, because its ability to store state isn't required when the entire sequence is given

I'm afraid I'm lost here as I still see `Recur` needed. For example in this current PR, for a vanilla left to right RNN, we iterate on a `Recur` so that state gets updated. How would the hidden and context outputs propagate if it's not through Recur's state (or an equivalent storage of cell's hidden state?) Mabe I'd need a minimal example to get the point.

darsnack: > Dispatching on (m::Chain{<:NTuple{N, RNNCell} where N)(x::CuArray) ** would be one option

Assuming the user packages the recurrent cells into a `Chain`. So perhaps a more formal API to push users in that direction would be better. `Recur` could behave more like `Chain` but with hidden state input/output semantics between layers.

ToucheSir: > or an equivalent storage of cell's hidden state?

Pretty much this. Using `Recur` means honouring its contract and setting the state appropriately at the end of the sequence. That might make sense for the default/fallback path, but constitutes unnecessary work for the optimized CuDNN one. Moreover, having the mutable struct field to stuff escaping allocations in and an additional function barrier to jump through may well stymie some optimizations. IMO we wouldn't have a mutable `Recur` at all in an ideal world, but until that day comes I'd like to rely on it as little as possible :)

DhairyaLGandhi: Right, doing the general thing first should be the priority. With the layer stacks, we can reduce multiple layers prior to calling the kernels at a later stage if needed.

src/layers/recurrent.jl
DhairyaLGandhi: It is easy to add copies and such with iteration and repeated operations with ad which theoretically should be fusable.

src/layers/recurrent.jl
darsnack: @ToucheSir so are you envisioning something like `BiDirectional(Cell(...))` as the basic block, then `BiDirectional(Chain(Cell(...), Cell(...), ...))` for the stack? Cause then the dispatch makes sense to me.

mkschleg:
Was able to play with this again.

We can get rid of the splat from `Flux.stack` by using `reduce` or `foldl` over an hcat. Unfortunately, foldl doesn't take advantage of reduce's optimizations for hcat and vcat. But maybe reduce here is fine? I wasn't sure if associativity would matter for hcat (i.e. if they break associativity in the future, will `reduce(hcat, x)` produce different orders of the arrays?)

You can see the results below for CPU and GPU. Overall the runtimes and memory consumption follow `reduce` < `stack` < `foldl`, so reduce is the best. 

The one optimization that could help even more is if there is a way to reduce directly into the final matrix instead of having to `cat`. I'm not sure if that kind of reduction operator exists, but would love to know so I can test it.

### setup

```julia
using Flux, BenchmarkTools

function recur_stack(m::Flux.Recur, x::AbstractArray{T, 3}) where T
    h = [m(x[:, :, i]) for i in 1:size(x, 3)]
    Flux.stack(h, 3)
end

function recur_foldl(m::Flux.Recur, x::AbstractArray{T, 3}) where T
    h = [m(x[:, :, i]) for i in 1:size(x, 3)]
    sze = size(h[1])
    reshape(foldl(hcat, h), sze[1], sze[2], length(h))
end

function recur_reduce(m::Flux.Recur, x::AbstractArray{T, 3}) where T
    h = [m(x[:, :, i]) for i in 1:size(x, 3)]
    sze = size(h[1])
    reshape(reduce(hcat, h), sze[1], sze[2], length(h))
end

function get_grad_stack(rnn, x)
    grads = gradient(Flux.params(rnn)) do 
        sum(recur_stack(rnn, x))
    end
end
# sim for foldl and stack

rnn = RNN(1024, 512)
rnn_gpu = RNN(1024, 512) |> gpu

x = cat([rand(Float32, 1024, 32) for t in 1:16]...; dims=3);
x_gpu = cat([rand(Float32, 1024, 32) for t in 1:16]...; dims=3) |> gpu;
```

### Forward pass CPU
```julia
@benchmark begin; Flux.reset!(rnn); recur_stack(rnn, x); end
BenchmarkTools.Trial: 698 samples with 1 evaluation.
 Range (min … max):  6.909 ms …   9.340 ms  ┊ GC (min … max): 0.00% … 24.63%
 Time  (median):     6.949 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   7.158 ms ± 662.046 μs  ┊ GC (mean ± σ):  2.82% ±  6.99%

  █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁ ▁ ▁
  7.17 ms         Histogram: frequency by time        6.94 ms <

 Memory estimate: 5.96 MiB, allocs estimate: 378.

@benchmark begin; Flux.reset!(rnn); recur_foldl(rnn, x); end
BenchmarkTools.Trial: 424 samples with 1 evaluation.
 Range (min … max):  11.058 ms … 17.287 ms  ┊ GC (min … max): 0.00% … 18.60%
 Time  (median):     11.127 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   11.789 ms ±  1.346 ms  ┊ GC (mean ± σ):  5.39% ±  8.89%

  █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄ ▆
  11.2 ms      Histogram: log(frequency) by time      14.3 ms <

 Memory estimate: 13.38 MiB, allocs estimate: 160.

@benchmark begin; Flux.reset!(rnn); recur_reduce(rnn, x); end
BenchmarkTools.Trial: 715 samples with 1 evaluation.
 Range (min … max):  6.745 ms …   9.305 ms  ┊ GC (min … max): 0.00% … 24.83%
 Time  (median):     6.784 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   6.989 ms ± 644.995 μs  ┊ GC (mean ± σ):  2.83% ±  6.98%

  █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁▁▁▁▁ ▁ ▁
  6.98 ms         Histogram: frequency by time        6.78 ms <

 Memory estimate: 5.94 MiB, allocs estimate: 132.

@benchmark begin; Flux.reset!(rnn); get_grad_1(rnn, x); end
BenchmarkTools.Trial: 107 samples with 1 evaluation.
 Range (min … max):  44.002 ms … 50.351 ms  ┊ GC (min … max):  7.44% … 18.70%
 Time  (median):     47.295 ms              ┊ GC (median):    13.60%
 Time  (mean ± σ):   46.977 ms ±  1.217 ms  ┊ GC (mean ± σ):  12.79% ±  2.37%

                                      █
  ▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁█▁▁▁▁▁▃▁▄▁▁▁▁▁▁▁▁▁▁▄▁▁▁ ▂
  46.8 ms         Histogram: frequency by time        47.5 ms <

 Memory estimate: 168.14 MiB, allocs estimate: 5231.

```
### Gradient CPU
```julia

julia> @benchmark begin; Flux.reset!(rnn); get_grad_stack(rnn, x); end                                                                                                                                                                   [51/1896]
BenchmarkTools.Trial: 107 samples with 1 evaluation.
 Range (min … max):  44.002 ms … 50.351 ms  ┊ GC (min … max):  7.44% … 18.70%
 Time  (median):     47.295 ms              ┊ GC (median):    13.60%
 Time  (mean ± σ):   46.977 ms ±  1.217 ms  ┊ GC (mean ± σ):  12.79% ±  2.37%

                                      █
  ▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁█▁▁▁▁▁▃▁▄▁▁▁▁▁▁▁▁▁▁▄▁▁▁ ▂
  46.8 ms         Histogram: frequency by time        47.5 ms <

 Memory estimate: 168.14 MiB, allocs estimate: 5231.

@benchmark begin; Flux.reset!(rnn); get_grad_foldl(rnn, x); end
BenchmarkTools.Trial: 141 samples with 1 evaluation.
 Range (min … max):  35.397 ms …  37.066 ms  ┊ GC (min … max): 13.22% … 12.84%
 Time  (median):     35.511 ms               ┊ GC (median):    13.43%
 Time  (mean ± σ):   35.544 ms ± 200.277 μs  ┊ GC (mean ± σ):  13.41% ±  0.08%

                                █
  ▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁ ▄
  35.4 ms       Histogram: log(frequency) by time      35.5 ms <

 Memory estimate: 175.61 MiB, allocs estimate: 6127.

julia> @benchmark begin; Flux.reset!(rnn); get_grad_reduce(rnn, x); end
BenchmarkTools.Trial: 143 samples with 1 evaluation.
 Range (min … max):  32.100 ms … 40.431 ms  ┊ GC (min … max):  7.43% … 11.60%
 Time  (median):     34.985 ms              ┊ GC (median):    13.44%
 Time  (mean ± σ):   35.152 ms ±  1.503 ms  ┊ GC (mean ± σ):  12.99% ±  1.94%

                                                              █
  ▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ █
  34.3 ms         Histogram: frequency by time        32.2 ms <

 Memory estimate: 168.10 MiB, allocs estimate: 4483.
```

## GPUs

```julia
@benchmark begin; Flux.reset!(rnn_gpu); recur_stack(rnn_gpu, x_gpu); end
BenchmarkTools.Trial: 5328 samples with 1 evaluation.
 Range (min … max):  793.573 μs … 23.339 ms  ┊ GC (min … max): 0.00% … 31.66%
 Time  (median):     811.077 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   934.020 μs ±  1.641 ms  ┊ GC (mean ± σ):  4.30% ±  2.35%

  █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁  ▁
  1.16 ms         Histogram: frequency by time          811 μs <

 Memory estimate: 158.67 KiB, allocs estimate: 3602.

@benchmark begin; Flux.reset!(rnn_gpu); recur_foldl(rnn_gpu, x_gpu); end
BenchmarkTools.Trial: 4919 samples with 1 evaluation.
 Range (min … max):  844.082 μs … 13.373 ms  ┊ GC (min … max): 0.00% … 35.85%
 Time  (median):     859.915 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.013 ms ±  1.351 ms  ┊ GC (mean ± σ):  5.88% ±  4.07%

  █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁
  1.15 ms         Histogram: frequency by time          866 μs <

 Memory estimate: 160.36 KiB, allocs estimate: 3821.

@benchmark begin; Flux.reset!(rnn_gpu); recur_reduce(rnn_gpu, x_gpu); end
BenchmarkTools.Trial: 5580 samples with 1 evaluation.
 Range (min … max):  762.537 μs … 22.104 ms  ┊ GC (min … max): 0.00% … 31.36%
 Time  (median):     777.677 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   892.435 μs ±  1.527 ms  ┊ GC (mean ± σ):  4.13% ±  2.32%

  █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁   ▁
  1.18 ms         Histogram: frequency by time          776 μs <

 Memory estimate: 141.77 KiB, allocs estimate: 3275.
```

## GPU Gradients

```julia
julia> @benchmark begin; Flux.reset!(rnn_gpu); get_grad_stack(rnn_gpu, x_gpu); end
BenchmarkTools.Trial: 1070 samples with 1 evaluation.
 Range (min … max):  3.710 ms … 9.461 ms  ┊ GC (min … max):  0.00% … 33.69%
 Time  (median):     3.782 ms             ┊ GC (median):     0.00%
 Time  (mean ± σ):   4.670 ms ± 1.989 ms  ┊ GC (mean ± σ):  10.64% ± 12.24%

  █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁▁▁▁▁ ▁
  4.16 ms        Histogram: frequency by time       3.86 ms <

 Memory estimate: 536.33 KiB, allocs estimate: 13396.

@benchmark begin; Flux.reset!(rnn_gpu); get_grad_foldl(rnn_gpu, x_gpu); end
BenchmarkTools.Trial: 974 samples with 1 evaluation.
 Range (min … max):  4.079 ms … 10.506 ms  ┊ GC (min … max):  0.00% … 32.17%
 Time  (median):     4.142 ms              ┊ GC (median):     0.00%
 Time  (mean ± σ):   5.129 ms ±  2.108 ms  ┊ GC (mean ± σ):  10.86% ± 12.39%

  █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁▁▁  ▁
  4.52 ms        Histogram: frequency by time        4.15 ms <

 Memory estimate: 584.59 KiB, allocs estimate: 15051.

@benchmark begin; Flux.reset!(rnn_gpu); get_grad_reduce(rnn_gpu, x_gpu); end
BenchmarkTools.Trial: 1095 samples with 1 evaluation.
 Range (min … max):  3.630 ms … 9.804 ms  ┊ GC (min … max):  0.00% … 33.95%
 Time  (median):     3.687 ms             ┊ GC (median):     0.00%
 Time  (mean ± σ):   4.564 ms ± 1.942 ms  ┊ GC (mean ± σ):  10.68% ± 12.29%

  █
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁▁▁▁▁▁ ▁
  4.05 ms        Histogram: frequency by time       3.72 ms <

 Memory estimate: 492.34 KiB, allocs estimate: 12522.
```

mkschleg:
I believe we can get away with using reduce, as the associativity doesn't matter. Only that it is commutative. i.e.

```julia
julia> x1, x2, x3 = rand(10), rand(10), rand(10);
julia> hcat(hcat(x1, x2), x3) .== hcat(x1, hcat(x2, x3))
true
```

ToucheSir:
> The one optimization that could help even more is if there is a way to reduce directly into the final matrix instead of having to `cat`.

`reduce([hv]cat)` should do this already, implementation in https://github.com/JuliaLang/julia/blob/master/base/abstractarray.jl#L1562. Order of evaluation shouldn't matter here since the RNN has already been run through the inputs in order via the comprehension above.

If we can assume that the network will output the same sized output at each timestep, then another possibility is constructing a [`Buffer`](https://fluxml.ai/Zygote.jl/latest/utils/#Zygote.Buffer) of shape `(outputsize..., timesteps)` and writing to it in a loop. I haven't tested the performance of this method, however.

mkschleg:
I just tested that, and it is a bit slower than the stack version (on a different computer, so I'm not going to put up results as they would be confusing). I think the reduce would be the best option with the current implementation.

mkschleg:
Given the above experimentation, i've changed the impl to be a `reduce` + a reshape both because it doesn't splat (so should only compile once if the sequence length changes) and it has better memory and computation time properties. 

Aside: I wonder if we should also change the Flux.Stack implementation to use reduce, but that is another issue.

mkschleg:
I'll start working on tests + docs now. I also want to make sure the gradients are consistent.

mkschleg:
I just added tests for BPTT and Size checking (on CPUs), and added a test for the gpu following the one hot encoding test. I also added documentation for the folding which follows the over what was there previously.

Finally. While digging through tests, I noticed that the RNN tests do broadcasting instead of the new for loop practice.

src/layers/recurrent.jl
mkschleg: I'm resolving, and added BiDirectional operations in #1678.

src/layers/recurrent.jl
```diff
@@ -67,6 +81,12 @@ end
 
 flip(f, xs) = reverse(f.(reverse(xs)))
 
+function (m::Recur)(x::AbstractArray{T, 3}) where T
+  h = [m(x[:, :, i]) for i in 1:size(x, 3)]
```
ToucheSir, line 85: Do we want to consider inputs of size `AxBx0`? I assume not because other layers would blow up even if we did, but putting a pin in this if anyone else wants to weigh in.

test/cuda/curnn.jl
```diff
@@ -53,6 +53,17 @@ end
...
+    Flux.reset!(rnn)
+    Flux.reset!(curnn)
+    fx = rand(Float32, 10, batch_size, 3)
+    cufx = gpu(fx)
+    fy = (rnn(fx); rnn(fx))
+
+    cufy = (curnn(cufx); curnn(cufx))
+    @test fy ≈ collect(cufy)
+
+
```
ToucheSir, line 67:
```suggestion
```

test/layers/recurrent.jl
```diff
@@ -42,26 +42,52 @@ end
...
+        sum(rnn(seq)[:, :, 3])
     end
+    Flux.reset!(rnn);
+    bptt = gradient(Wh -> sum(tanh.(rnn.cell.Wi * seq[:, :, 3] + Wh *
+                                  tanh.(rnn.cell.Wi * seq[:, :, 2] + Wh *
+                                        tanh.(rnn.cell.Wi * seq[:, :, 1] +
+                                            Wh * rnn.cell.state0
+                                        + rnn.cell.b)
+                                  + rnn.cell.b)
+                            + rnn.cell.b)),
```
ToucheSir, line 60: This can probably be in a `do` block with some intermediate variables :P

test/layers/recurrent.jl
```diff
@@ -42,26 +42,52 @@ end
...
+    x3 = rand(Float32,3,1,2)
+    Flux.reset!(m1)
+    Flux.reset!(m2)
+    Flux.reset!(m3)
+    @test size(m1(x1)) == (5,)
+    @test size(m1(x1)) == (5,) # repeat in case of effect from change in state shape
+    @test size(m2(x2)) == (5,1)
+    @test size(m2(x2)) == (5,1)
+    @test size(m3(x3)) == (5,1,2)
+    @test size(m3(x3)) == (5,1,2)
```
ToucheSir, line 82:
```suggestion
    x1 = rand(Float32, 3)
    x2 = rand(Float32, 3, 1)
    x3 = rand(Float32, 3, 1, 2)
    Flux.reset!(m1)
    Flux.reset!(m2)
    Flux.reset!(m3)
    @test size(m1(x1)) == (5,)
    @test size(m1(x1)) == (5,) # repeat in case of effect from change in state shape
    @test size(m2(x2)) == (5, 1)
    @test size(m2(x2)) == (5, 1)
    @test size(m3(x3)) == (5, 1, 2)
    @test size(m3(x3)) == (5, 1, 2)
```

ToucheSir (commented):
Thanks again for your work on this! I won't have an opportunity to do a full review until next week, but if someone else gets to it beforehand I think this is good to go.

test/layers/recurrent.jl
mkschleg: Oh, definitely. I just followed what already existed. The rnn tests could probably be refactored.

src/layers/recurrent.jl
mkschleg: Dense also doesn't handle this, and returns a `DivideError`. What would we want the behaviour to be for `AxBx0`?

ToucheSir: Should be fine to leave it as-is for now, then.

test/layers/recurrent.jl
ToucheSir: That they should! If it's not too much trouble, this would be a good start.

mkschleg: Done. I only did it for the new test, but can do for the old ones if that is something we want in this PR.

CarloLucibello:
looks good! add a news entry?

src/layers/recurrent.jl
```diff
@@ -67,6 +81,12 @@ end
...
+  sze = size(h[1])
+  reshape(reduce(hcat, h), sze[1], sze[2], length(h))
```
DhairyaLGandhi, line 87: I think this would create a bunch of intermediaries. Could we also check the benchmarks with larger batchsizes and models?

mkschleg: What would you want to compare against? The Zygote.buffer version? And what are we looking for? Memory issues?

ToucheSir: The only additional intermediary generated here is the vector `h` holding the outputs. From my own benchmarking with `Buffer`, this doesn't appear to add much if any overhead memory-wise, and is faster to compute + not dependent on Zygote OOTB.

ToucheSir: To clarify, the only way we can further reduce intermediate allocations is to make the RNNCell itself operate in-place. Since that's not possible at the moment, we eat an extra one for the output regardless. An extra 8 bytes for each additional timestep is insignificant in comparison, especially since `Buffer` usage appears to induce extra allocations of its own.

DhairyaLGandhi: Vs the stack implementation for longer time steps. And yes, I want to understand the memory impact in the GPU case specifically. Also, would it make sense to avoid the copy by sending a `view` of `x`?

ToucheSir: I forgot to mention it, but I was using `view` and it definitely helps! Here are the benchmarks from that:

Sequence length stress test: RNN(1, 1), batch size 1, sequence length 1024:
```
CPU:
f = recur_stack
  18.194 ms (17966 allocations: 15.17 MiB)
  68.377 ms (302842 allocations: 39.03 MiB)
f = recur_foldl
  620.649 μs (4100 allocations: 2.43 MiB)
  73.480 ms (358217 allocations: 78.00 MiB)
f = recur_reduce
  346.497 μs (3078 allocations: 300.42 KiB)
  47.053 ms (259184 allocations: 22.33 MiB)
f = recur_buffer
  341.854 μs (3075 allocations: 292.19 KiB)
  49.909 ms (279707 allocations: 20.87 MiB)

GPU:
f = recur_stack
  51.815 ms (161334 allocations: 22.73 MiB)
  213.182 ms (852717 allocations: 55.73 MiB)
f = recur_foldl
  38.556 ms (182209 allocations: 8.91 MiB)
  221.603 ms (964286 allocations: 60.72 MiB)
f = recur_reduce
  32.980 ms (142349 allocations: 7.51 MiB)
  190.225 ms (799857 allocations: 38.89 MiB)
f = recur_buffer
  36.472 ms (143886 allocations: 7.48 MiB)
  219.450 ms (836290 allocations: 38.26 MiB)
```

A more realistic scenario: RNN(32, 32), batch size 64, sequence length 128:
```
CPU:
f = recur_stack
  3.319 ms (2077 allocations: 4.23 MiB)
  41.513 ms (38493 allocations: 269.27 MiB)
f = recur_foldl
  7.595 ms (642 allocations: 67.54 MiB)
  48.827 ms (45359 allocations: 333.60 MiB)
f = recur_reduce
  2.766 ms (391 allocations: 4.04 MiB)
  37.872 ms (33394 allocations: 268.86 MiB)
f = recur_buffer
  2.866 ms (388 allocations: 4.04 MiB)
  41.226 ms (35456 allocations: 270.68 MiB)

GPU:
f = recur_stack
  3.959 ms (20388 allocations: 1.18 MiB)
  26.319 ms (116239 allocations: 5.66 MiB)
f = recur_foldl
  4.394 ms (23233 allocations: 1.12 MiB)
  28.567 ms (130096 allocations: 6.20 MiB)
f = recur_reduce
  3.701 ms (18189 allocations: 967.55 KiB)
  25.282 ms (109855 allocations: 5.22 MiB)
f = recur_buffer
  4.042 ms (18187 allocations: 960.36 KiB)
  27.730 ms (110834 allocations: 5.05 MiB)
```
`reduce(hcat)` is the fastest in every configuration, and at worst only uses marginally more memory than `Buffer`.

src/layers/recurrent.jl
DhairyaLGandhi, line 85:
```suggestion
  h = [m(@view(x[:, :, i])) for i in 1:size(x, 3)]
```
cc @ToucheSir

ToucheSir: Since:
```julia
julia> @macroexpand @view x[:, :, 1]
:(true && (view)(x, :, :, 1))
```
I'd favour:
```suggestion
  h = [m(view(x, :, :, i)) for i in 1:size(x, 3)]
```

src/layers/recurrent.jl
mkschleg: Thanks @ToucheSir! I got really busy so wasn't able to work on this. I think the reduce(hcat) + view is the best option right now.

DhairyaLGandhi:
I'd be curious to hear @mkschleg 's view on whether we should be more conservative on choosing what a specific dimension means in this way for future models.

Also ditto for @sdobber

DhairyaLGandhi (approved):

src/layers/recurrent.jl
```diff
@@ -67,6 +81,12 @@ end
 
 flip(f, xs) = reverse(f.(reverse(xs)))
 
+function (m::Recur)(x::AbstractArray{T, 3}) where T
```
DhairyaLGandhi, line 84: Should `3 => N` here such that we only consider the last dimension as time, leaving modellers to tune what different dimensions might mean for different tasks? It shouldn't be a very difficult charge but can help generalise our approach similar to how `conv` works.

ToucheSir: That dispatch would be ideal, but the problem is that it would override `(m::Recur)(x::AbstractArray{T, 1})` and `(m::Recur)(x::AbstractArray{T, 2})`, i.e. the single timestep cases. Even `(m::Recur)(x::AbstractArray{T, 3})` can be suspect if `Recur` is wrapping a model that expects 3+1D inputs (e.g. `Recur(Chain(Conv((3,), ...), GlobalMeanPool(), RNNCell(...))`). That scenario is less common, however, as I think users would be more likely to write `Chain(Conv((3,), ...), GlobalMeanPool(), RNN(...))`.

DhairyaLGandhi: `AbstractVector/Matrix` can take care of dispatches. The usage question is orthogonal, however, I'd imagine having the ability to write both the models is a more pressing concern.

ToucheSir: My hope was to avoid having 3 dispatches with the exact same implementation (even if it's in a helper function). As you mentioned, the longer-term solution is to add in more appropriate functionality for this use case (i.e. not trying to overload `Recur` even more).

DhairyaLGandhi: We need just one kernel and reshape the inputs as necessary

ToucheSir: We need at least the following:

1. `(m::Recur)(x)` for non-array inputs
2. `(m::Recur)(x::AbstractArray{T, N})` for this PR, which overrides 1. for sequences.
3. `(m::Recur)(x::AbstractArray{T, 1})` and `(m::Recur)(x::AbstractArray{T, 2})`, which use the same code as 1. but exist solely to undo the override in 2. for individual sequence elements + the current array-per-timestep pattern.

Now to be fair, this is only 2x2 lines of duplicated code at worst for case 3. However, it does show the perils of trying to guess what the user wants instead of creating more explicit and appropriate abstractions like we've discussed in #1678. For example, these dispatches will not work if someone wants to apply a Convolutional LSTM across individual timesteps.

Just to be clear, I do not want to hold up this PR. Far the contrary, this is a great stopgap that should allow us to start prototyping efficient CuDNN forwarding while working on #1678. Embracing that and not trying to expand the scope beyond said forwarding is advantageous because it means users won't hit unavoidable edge cases that only exist because `Recur` is trying to handle 2 conflicting usage patterns.

mkschleg: I think no matter what we do we will have to have a more complicated dispatch structure for recur to be able to support different assumptions about inputs (i.e. ConvLSTMs) as we support the folding version. But maybe we can do this by dispatching on `Recur{ConvLSTM}` and being clear in the documentation how this works. This will be a little bit unfortunate, but the amount code duplication is minimal and the default is pretty clear. I don't see how to design past this and have Recur be responsible for folding and per-step.

I think we can assume the final index is the temporal index (which seems reasonable given how arrays are stored in julia). I agree with @ToucheSir that this is a stop-gap and we can use it to iterate.

mkschleg:
> I'd be curious to hear @mkschleg 's view on whether we should be more conservative on choosing what a specific dimension means in this way for future models.
> 
> Also ditto for @sdobber

I think we need to be very conservative, but there might certain assumptions we might be able to make. Unfortunately, I think what we are doing here might break some assumptions in the [batch norm implementation](https://github.com/FluxML/Flux.jl/blob/7ada01cca14e1d87b946e258532cb152030ca671/src/layers/normalise.jl#L199). So we likely need to come up with strategies across Flux on what different dimensions may mean and how we can navigate the space. That means we probably need a design document of some kind.

mkschleg:
I think this might highlight some issues with `chain`. Sometimes I feel we want different behavior based on what layers are in a chain and what the input structure is. I'm not sure this is something we can solve, and we should just push people more towards building their own model structs in these instances.

Just as an example. In my work, I sometimes find myself searching through the chain to find certain layers so I can do a dynamic dispatch. I'm not sure there is a way around this given I need different input structure and different behavior in my updates based on what layers are in my chain.

ToucheSir:
There's always the option of allowing the user to specify which dimensions are important up-front. That of course would have to be the responsibility of a layer other than `Recur`, so I don't think we need to work it out for this PR.

mkschleg:
Sounds good. Are there any lingering questions w/ this PR? I think it would be beneficial to get this through so we can start working on CuDNN paths.

jeremiedb:
Just making a bump for having this PR merged. 
I think it would be preferable to have a new feature that might be imperfect (from the API perspective, I don't mean to compromise on correctness of the result) but that will be exposed and benefit from actual users' experience and feedback rather than trying to anticipate the perfect fit.

ToucheSir:
Sorry, I didn't realize this had already received an approval. Let's get it merged. bors r+

ToucheSir:
bors+

PR #1708, opened 2021-09-09
Add GPU Adaptor

DhairyaLGandhi:
Updates return types for CUDA/ CPU movement to hold structure information wherever possible.


DhairyaLGandhi:
The error is due to calling `Array(Zeros)` which seems unavoidable to some degree. We might need a different rule to handle `Zeros`.

DhairyaLGandhi:
@maleadt We currently have `typeof(gradient(x -> sum(gpu(cpu(x))), ::CuArray))` is `Tuple{Matrix}`. This is the main culprit behind the falling tests. I have added the adjoint to paper over it. We need to give Adapt some more information here. Ideally, I'd like to get rid of the adjoint.

test/cuda/cuda.jl
```diff
@@ -93,13 +93,13 @@ end
   @test gradient(x -> sum(abs, gpu(x)), a)[1] isa Matrix
   @test gradient(x -> sum(gpu(x)), a)[1] isa Matrix
   @test_skip gradient(x -> sum(gpu(x)), a')[1] isa Matrix  # sum(::Adjoint{T,CuArray}) makes a Fill
-  @test gradient(x -> sum(abs, cpu(x)), ca)[1] isa CuArray
-  @test gradient(x -> sum(cpu(x)), ca)[1] isa CuArray  # This involves FillArray, moved to GPU
-  @test gradient(x -> sum(cpu(x)), ca')[1] isa CuArray 
+  @test gradient(x -> sum(abs, cpu(x)), ca)[1] isa Matrix # CuArray
```
CarloLucibello, line 96: this test was previously correct and now it's wrong

test/cuda/cuda.jl
```diff
@@ -93,13 +93,13 @@ end
...
+  @test gradient(x -> sum(cpu(x)), ca)[1] isa Zygote.FillArrays.Fill  # This involves FillArray, which should be GPU compatible
+  @test gradient(x -> sum(cpu(x)), ca')[1] isa LinearAlgebra.Adjoint
```
CarloLucibello, line 98: this should test that the Adjoint wraps a CuArray

test/cuda/cuda.jl
```diff
@@ -93,13 +93,13 @@ end
...
 
   # Even more trivial: no movement
   @test gradient(x -> sum(abs, cpu(x)), a)[1] isa Matrix
-  @test gradient(x -> sum(abs, cpu(x)), a')[1] isa Matrix
+  @test gradient(x -> sum(abs, cpu(x)), a')[1] isa LinearAlgebra.Adjoint
```
CarloLucibello, line 102: this should test that the Adjoint wraps a Matrix

test/cuda/cuda.jl
```diff
@@ -108,6 +108,8 @@ end
   g0 = gradient(x -> sum(abs, (a * (a * x))), a)[1]
   @test g0 ≈ gradient(x -> sum(abs, cpu(ca * gpu(a * x))), a)[1]
   @test cu(g0) ≈ gradient(x -> sum(abs, gpu(a * cpu(ca * x))), ca)[1]
+  @test gradient(x -> sum(gpu(cpu(x))), a)[1] isa AbstractMatrix
```
CarloLucibello, line 111:
```suggestion
  @test gradient(x -> sum(gpu(cpu(x))), a)[1] isa Matrix
```

test/cuda/cuda.jl
```diff
@@ -117,21 +119,25 @@ end
...
   @test gradient(x -> cpu(gpu(x) * gpu(x))[1,2], Float32[1 2 3; 4 5 6; 7 8 9]) == ([2 6 8; 0 2 0; 0 3 0],)
-
 end
+
 @testset "gpu(x) and cpu(x) on structured arrays" begin
   # Check first that cpu() is a no-op on these, which adapt(Array, x) presently is not:
-  @test cpu(1:3) isa UnitRange
-  @test cpu(range(1,3,length=4)) isa AbstractRange
+  @test cpu(1:3) isa Vector
+  @test cpu(range(1, 3, length = 4)) isa Vector
```
CarloLucibello, line 127: this seems a regression. Can we preserve this old behavior for ranges instead of materializing them?

test/cuda/cuda.jl
```diff
@@ -117,21 +119,25 @@ end
...
+  # OneElement isn't GPU compatible
   g1 = Zygote.OneElement(1, (2,3), axes(ones(4,5)))
-  @test cpu(g1) isa Zygote.OneElement
-  g2 = Zygote.Fill(1f0,2)
-  @test cpu(g2) isa Zygote.Fill
+
+  @test cpu(g1) isa Matrix
+  g2 = Zygote.Fill(1f0, 2)
+  @test cpu(g2) isa Vector
```
CarloLucibello, line 133: these seems a regression

CarloLucibello:
I think this PR should keep most old tests in place to make sure we don't have regressions (apparently we do) 
and add new ones for the new features (preserve some wrappers?). 

It's not clear what this PR is fixing though, isn't Adapt supposed to take care of wrapped arrays?

test/cuda/cuda.jl
DhairyaLGandhi: This is the original behavior. This is materialized for working with CuArrays explicitly, making our job easier down the line. But we could see if preserving is feasible.

test/cuda/cuda.jl
DhairyaLGandhi: Agreed, but there was an impression to want to materialize more frequently. I can make it not materialise this case.

DhairyaLGandhi:
This is a more general design for data movement using an adaptor that is designed by CUDA.jl to work with movement and structured arrays between cpu and gpu memory.

CarloLucibello:
> using an adaptor that is designed by CUDA.jl

what do you mean? I don't see anything from CUDA.jl here

test/cuda/cuda.jl
DhairyaLGandhi: Should be fixed now

DhairyaLGandhi:
@maleadt suggested defining an adaptor for handling GPU movement cases as part of how Adapt is intended to be used. This way we have more control over how different array types are materialised or preserved.

CarloLucibello:
Can you add https://github.com/FluxML/Flux.jl/issues/1720#issuecomment-927128895 and related test?

src/functor.jl
```diff
@@ -61,7 +61,29 @@ function loadparams!(m, xs)
...
+adapt_storage(to::FluxCPUAdaptor, x::AbstractArray) = Array(x)
+adapt_storage(to::FluxCPUAdaptor, x::AbstractRange) = x
+adapt_storage(to::FluxCPUAdaptor, x::Zygote.FillArrays.AbstractFill) = x
+Zygote.@adjoint function Array(x::CUDA.CuArray)
+  Array(x), d -> (CUDA.cu(d),)
+end
+
+Zygote.@adjoint function Adapt.adapt_storage(to::FluxCPUAdaptor, x::CUDA.AbstractGPUArray)
+  adapt_storage(to, x), d -> (nothing, adapt_storage(FluxCUDAAdaptor(), d),)
+end
```
CarloLucibello, line 83: better use `rrule` for these

test/cuda/cuda.jl
```diff
@@ -117,21 +121,25 @@ end
...
   @test cpu(1:3) isa UnitRange
-  @test cpu(range(1,3,length=4)) isa AbstractRange
+  @test cpu(range(1, 3, length = 4)) isa AbstractRange
+  # OneElement isn't GPU compatible
   g1 = Zygote.OneElement(1, (2,3), axes(ones(4,5)))
-  @test cpu(g1) isa Zygote.OneElement
-  g2 = Zygote.Fill(1f0,2)
-  @test cpu(g2) isa Zygote.Fill
+
+  @test cpu(g1) isa Matrix
```
CarloLucibello, line 133: can we keep preserving Zygote.OneElement as well?

DhairyaLGandhi: no, it causes scalar indexing.

CarloLucibello: this is `cpu`, scalar indexing is not an issue.

I'm not sure how it should be handled on gpu, but @mcabbott put already some thought in it so better not disrupt his work without a valid reason

test/cuda/cuda.jl
```diff
@@ -117,21 +122,28 @@ end
...
+  g2 = Zygote.Fill(1f0, 2)
+  @test cpu(g2) isa Zygote.FillArrays.AbstractFill
   g3 = transpose(Float32[1 2; 3 4])
-  @test parent(cpu(g3)) isa Matrix
+
+  @test parent(cpu(g3)) isa Matrix{Float32}
+
+
+  @test cpu(sparse(rand(3,3))) isa AbstractSparseArray
+  @test gpu(sparse(rand(3,3))) isa AbstractSparseArray
```
CarloLucibello, line 143:
```suggestion
  x = sparse(rand(3,3))
  @test cpu(x) === x
  @test gpu(x) isa CUDA.CUSPARSE.CuSparseMatrix
  @test cpu(gpu(x)) isa SparseMatrixCSC
```

src/functor.jl
```diff
@@ -61,7 +62,31 @@ function loadparams!(m, xs)
...
+adapt_storage(to::FluxCUDAAdaptor, x) = CUDA.cu(x)
+adapt_storage(to::FluxCUDAAdaptor, x::Zygote.FillArrays.AbstractFill) = CUDA.cu(collect(x))
+
+# TODO: figure out the correct design for OneElement
+adapt_storage(to::FluxCUDAAdaptor, x::Zygote.OneElement) = CUDA.cu(collect(x))
+
+struct FluxCPUAdaptor end
+
+# define rules for handling structured arrays
+adapt_storage(to::FluxCPUAdaptor, x::AbstractArray) = Array(x)
```
CarloLucibello, line 75: this is making a copy, we don't want this, use `adapt(Array, x)`

src/functor.jl
```diff
@@ -61,7 +62,31 @@ function loadparams!(m, xs)
...
+adapt_storage(to::FluxCPUAdaptor, x::AbstractRange) = x
+adapt_storage(to::FluxCPUAdaptor, x::Zygote.FillArrays.AbstractFill) = x
+adapt_storage(to::FluxCPUAdaptor, x::AbstractSparseArray) = x
```
CarloLucibello, line 78:
```suggestion
adapt_storage(to::FluxCPUAdaptor, x::AbstractSparseArray) = adapt(SparseMatrixCSC, x)
```

src/functor.jl
CarloLucibello: Tests like 
```julia
x = rand(2)
@test cpu(x) === x
xg = gpu(x)
@test gpu(xg) === xg
```
should be added to make sure we don't regress on this

src/functor.jl
DhairyaLGandhi: This only works with Matrix types.

CarloLucibello: Seems to work also for vectors
```julia
julia> adapt(SparseMatrixCSC, sprand(2, 0.1))
2-element SparseVector{Float64, Int64} with 0 stored entries
```
The alternative is
```julia
adapt_storage(to::FluxCPUAdaptor, x::AbstractSparseMatrix) = adapt(SparseMatrixCSC, x)
adapt_storage(to::FluxCPUAdaptor, x::AbstractSparseVector) = adapt(SparseVector x)
```

test/cuda/cuda.jl
DhairyaLGandhi: We need the behaviour to be consistent across cpu and gpu. I think with the number of times it's come up recently and the special casing on master and in this pr, we will need to think about OneElement further.

src/functor.jl
DhairyaLGandhi: For the cpu case, there isn't any adapting necessary, and all the methods proposed here are already used via `CUDA.cu`

CarloLucibello: this is not true, we have to convert the `AbstractSparseArray`s that live in CUDA.CUSPARSE

test/cuda/cuda.jl
CarloLucibello: `cpu` should leave anything that lives on cpu unaltered

src/functor.jl
DhairyaLGandhi: I'm aware of that, but the methods you posted wouldn't achieve that, which is what got me confused. 

```Julia
julia> a = sparse(rand(Float32, 3,3))
3×3 SparseMatrixCSC{Float32, Int64} with 9 stored entries:
 0.33337   0.939036  0.874483
 0.455707  0.35652   0.511505
 0.642504  0.390601  0.925762

julia> ga = cu(a)
3×3 CUDA.CUSPARSE.CuSparseMatrixCSC{Float32} with 9 stored entries:
 0.33337   0.939036  0.874483
 0.455707  0.35652   0.511505
 0.642504  0.390601  0.925762

julia> adapt(SparseMatrixCSC, ga)
3×3 CUDA.CUSPARSE.CuSparseMatrixCSC{Float32} with 9 stored entries:
 0.33337   0.939036  0.874483
 0.455707  0.35652   0.511505
 0.642504  0.390601  0.925762
```

We will have to copy it to go from GPU to CPU.

test/cuda/cuda.jl
DhairyaLGandhi: There's no such guarantee.

src/functor.jl
DhairyaLGandhi: Done, I'll add the tests too.

src/functor.jl
DhairyaLGandhi: I've fixed this in the latest commit

test/cuda/cuda.jl
CarloLucibello: why do you mark as resolved? please use the tests I suggested

src/functor.jl
```diff
@@ -61,7 +62,31 @@ function loadparams!(m, xs)
...
+# TODO: figure out the correct design for OneElement
+adapt_storage(to::FluxCUDAAdaptor, x::Zygote.OneElement) = CUDA.cu(collect(x))
+
+struct FluxCPUAdaptor end
+
+# define rules for handling structured arrays
+adapt_storage(to::FluxCPUAdaptor, x::AbstractArray) = adapt(Array, x)
+adapt_storage(to::FluxCPUAdaptor, x::AbstractRange) = x
+adapt_storage(to::FluxCPUAdaptor, x::Zygote.FillArrays.AbstractFill) = x
+adapt_storage(to::FluxCPUAdaptor, x::T) where T <: CUDA.CUSPARSE.CUDA.CUSPARSE.AbstractCuSparseMatrix = adapt(Array, x)
```
CarloLucibello, line 78:
```suggestion
adapt_storage(to::FluxCPUAdaptor, x::T) where T <: CUDA.CUSPARSE.CUDA.CUSPARSE.AbstractCuSparseMatrix = SparseMatrixCSC(x)
```

test/cuda/cuda.jl
CarloLucibello: ...especially if you remove support for types we support already

src/functor.jl
DhairyaLGandhi: This should work as is. You should see.

```Julia
julia> cpu(gpu(sparse(rand(Float32, 3,3)))) 
3×3 SparseMatrixCSC{Float32, Int32} with 9 stored entries: 
 0.564721  …  0.253228
 0.86544      0.367709
 0.166146     0.153102
```

CarloLucibello: cool

test/cuda/cuda.jl
DhairyaLGandhi: We don't support OneElement, it is special cased either way.

test/cuda/cuda.jl
ToucheSir: Since tests pass, looks like this is resolved?

src/functor.jl
ToucheSir: I agree in general, but this need not hold up the PR. Really, both the `Adapt` method implementations and adjoints/rrules should live in Zygote itself. Something like this should be enough, and avoids having to re-define methods for our own adaptor:

```julia
function ChainRules.rrule(::typeof(Adapt.adapt_storage), a::Type{A}, x::AbstractArray{T})
    where {A <: Union{CuArray, CUDA.CuArrayAdaptor}, T}
  y = Adapt.adapt_storage(a, x)
  if x === y  # trivial case, e.g. gpu(x::CuArray)
    return y, dy -> (Zygote.ChainRules.NoTangent(), dy)
  else
    return y, dy -> (Zygote.ChainRules.NoTangent(), Adapt.adapt_storage(Array{T}, dy))
  end
end

function ChainRules.rrule(::typeof(Adapt.adapt_storage), a::Type{A}, x::CuArray{<:Any,<:Any,B})
    where {A <: Array, B}
    y = Adapt.adapt_storage(a, x)
    # FIXME T is not retained (needs upstream changes)
    return y, dy -> (Zygote.ChainRules.NoTangent(), Adapt.adapt_storage(CUDA.CuArrayAdaptor{B}, dy))
  end
end

function ChainRules.rrule(::typeof(Adapt.adapt_storage), a::Type{A}, x::CUDA.AbstractCuSparseArray{T})
    where {A <: Array, T}
    # FIXME memory location not retained (needs upstream changes)
    return y, dy -> (Zygote.ChainRules.NoTangent(), Adapt.adapt_storage(CuArray{T}, dy))
end
```
Good material for a follow-up PR!

test/cuda/cuda.jl
CarloLucibello: I was just suggesting a stricter test, this is fine though

src/functor.jl
CarloLucibello: yes this change should be non blocking

test/cuda/cuda.jl
CarloLucibello: We do support OneElement on master and you are removing support for no valid reason in what is supposed to be just a refactoring. It's not entirely clear to me the impact of the removed code, so I'd like to hear from @mcabbott

CarloLucibello:
aside from the fact that `cpu` shouldn't materialize Zygote.OneElement this LGTM

test/cuda/cuda.jl
DhairyaLGandhi: https://github.com/FluxML/Flux.jl/blob/d12ebfa124bc79d467cc4908181887b986cdff54/src/functor.jl#L152 can implicitly cause scalar indexing in hot loops without informing the user, and is special cased to do so. 
If you really want, we can add another dispatch for OneElement, but that wouldn't fix the underlying issue. If we were to remain generic and accept issues, we can fix them at the right place.

CarloLucibello: those scalar indexing are the best we can do right now I guess, converting to arrays is worse and the reason why OneElement was introduced (I didn't follow that story carefully though)

CarloLucibello (approved):

Issue #1597, opened 2021-05-12
Do not initialize CUDA during precompilation


maleadt:
This is wrong: https://github.com/FluxML/Flux.jl/blob/8a5b9774f0d6b254f60c88ca07a984997a016e75/src/losses/Losses.jl#L27

This should be guarded by a check `if !precompiling`: https://github.com/FluxML/Flux.jl/blob/8a5b9774f0d6b254f60c88ca07a984997a016e75/src/Flux.jl#L59-L61

Additionally, this shouldn't be done during `__init__`, but lazily when first calling the `gpu` function.

DhairyaLGandhi:
We already use the flag in `gpu` as well. Are you saying we skip this check here entirely and have `gpu` set the flag on first call and then use that consequently? Currently it's hidden behind a `Ref{Bool}` so we need something to say "hasn't been initialised yet", with `missing` or `nothing`

maleadt:
I think that would be better for users with a GPU, but who don't plan to use it (so that `using Flux` doesn't start downloading the large CUDA artifacts). I'll explore that in the multi GPU back-end PR.

darsnack:
In lieu of the long term fix, should we release a quick fix for the losses issue? Would using `use_cuda[]` instead of `CUDA.functional()` work?

DhairyaLGandhi:
It's the same either way. Fixing the init would be better.

IanButterworth:
It'd be nice to have this fixed. Downloading during precompilation is pretty awkward with the new Pkg auto precompilation, especially on a slow connection given the output during precomp is suppressed, so it's not clear that a download is happening

IanButterworth:
Friendly bump on this

Issue #1554, opened 2021-03-30
Very slow precompile (>50min) on julia 1.6.0 on Windows


yha:
```
    Updating registry at `C:\Users\sternlab\.julia\registries\General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
   Resolving package versions...
   Installed Requires ─────── v1.1.3
   Installed FillArrays ───── v0.11.7
   Installed ColorTypes ───── v0.10.12
   Installed StatsBase ────── v0.33.4
   Installed ChainRulesCore ─ v0.9.34
   Installed StaticArrays ─── v1.1.0
   Installed Zygote ───────── v0.6.7
   Installed Memoize ──────── v0.4.4
   Installed GPUCompiler ──── v0.10.0
   Installed CUDA ─────────── v2.6.2
    Updating `C:\Users\sternlab\.julia\environments\v1.6\Project.toml`
  [587475ba] + Flux v0.12.0
  [e88e6eb3] + Zygote v0.6.7
  [700de1a5] + ZygoteRules v0.2.1
    Updating `C:\Users\sternlab\.julia\environments\v1.6\Manifest.toml`
  [621f4979] + AbstractFFTs v1.0.1
  [1520ce14] + AbstractTrees v0.3.4
  [79e6a3ab] + Adapt v3.2.0
  [ab4f0b2a] + BFloat16s v0.1.0
  [fa961155] + CEnum v0.4.1
  [052768ef] + CUDA v2.6.2
  [082447d4] + ChainRules v0.7.56
  [d360d2e6] + ChainRulesCore v0.9.34
  [944b1d66] + CodecZlib v0.7.0
  [3da002f7] + ColorTypes v0.10.12
  [5ae59095] + Colors v0.12.6
  [bbf7d656] + CommonSubexpressions v0.3.0
  [34da2185] + Compat v3.25.0
  [9a962f9c] + DataAPI v1.6.0
  [864edb3b] + DataStructures v0.18.9
  [163ba53b] + DiffResults v1.0.3
  [b552c78f] + DiffRules v1.0.2
  [e2ba6199] + ExprTools v0.1.3
  [1a297f60] + FillArrays v0.11.7
  [53c48c17] + FixedPointNumbers v0.8.4
  [587475ba] + Flux v0.12.0
  [f6369f11] + ForwardDiff v0.10.17
  [d9f16b24] + Functors v0.2.1
  [0c68f7d7] + GPUArrays v6.2.0
[...]

PR #1711, opened 2021-09-15
Initialize CUDA lazily.

maleadt:
Closes https://github.com/FluxML/Flux.jl/issues/1597, fixes https://github.com/FluxML/Flux.jl/issues/1554.


DhairyaLGandhi (approved):

DhairyaLGandhi:
With #1708 `gpu` would be AD'd as well (ditto for the flux so far), so I was wondering if this would cause issue there.

PR #1716, opened 2021-09-24
Fix link in README

nilsmartel:
Points to deprecated package that used to be until v0.10.4, replaced by CUDA.jl in v0.11.0.

### PR Checklist

doesn't apply here, as far as I understant it.


darsnack (approved):
Looks great, thanks!

PR #1721, opened 2021-09-25
tag patch version

CarloLucibello:


DhairyaLGandhi:
We want to get #1708 in before releasing.

CarloLucibello:
I gave a review to #1708. It seems to have a few regressions and it's not even clear what it is fixing or where is the added value. So probably we don't want to have that as a release blocker (especially for a patch version). Bumping version again when it is ready is not a big deal in any case.

DhairyaLGandhi:
Master returns different types that are non breaking but reverting them would be breaking. I've added more context to #1078

In the meanwhile, if there's something specific needed from master, I'm happy to do a cherrypicked release.

CarloLucibello:
> Master returns different types that are non breaking but reverting them would be breaking.

What are these different types? If you refer to the changes due to ProjectTo, those are in Zygote and have already been tagged. 

If you refer to something #1704 (which is precisely the PR I need) I don't see what we would be reverting in a breaking way

mcabbott (approved):

PR #1726, opened 2021-09-30
add unbatch

CarloLucibello:
Add `unbatch` for convenience and consistency (we currently have `stack/unstack` but only `batch` with no `unbatch`). Can be specialized by downstream packages on their custom batched types. 

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -542,6 +542,8 @@ squeezebatch(x) = reshape(x, head(size(x)))
 
 Batch the arrays in `xs` into a single array.
 
+See also [`unbatch`](@ref)
+
 # Examples
 ```jldoctest
 julia> Flux.batch([[1,2,3],[4,5,6]])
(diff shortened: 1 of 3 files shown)


src/utils.jl
```diff
@@ -561,6 +563,27 @@ function batch(xs)
...
+```jldoctest
+julia> Flux.unbatch([1 3 5 7; 
+                     2 4 6 8])
+4-element Vector{Vector{Int64}}:
+ [1, 2]
+ [3, 4]
+ [5, 6]
+ [7, 8]
+"""
+unbatch(x::AbstractArray) = unstack(x, ndims(x))
```
darsnack, line 585: `batch` has this weird implementation where if the vector is not a vector of arrays, then it is a no-op. I suppose if we want perfect symmetry, we need a dispatch for this?

```suggestion
unbatch(x::AbstractVector) = x
unbatch(x::AbstractArray) = unstack(x, ndims(x))
```

Maybe we can also clean up the `batch` implementation too.

```julia
batch(x::AbstractVector{<:AbstractArray}) = stack(x, ndims(first(x)) + 1)
batch(x::AbstractVector) = x
```

darsnack:
Friendly bump, I think this is pretty much ready to go.

Issue #1730, opened 2021-10-01
additional arguments to loss function?


KenZhenLin:
How do I use `Flux.Optimise.train!` when my loss function has arguments in addition to `x` and `y`?

The code below works for loss function with x and y being the only arguments, that is, loss(x, y)

data = DataLoader((x, y), batchsize=2, shuffle=true, partial=true)
opt = ADAM(0.01, (0.9, 0.999))
for epoch = 1:100
    Flux.Optimise.train!(loss, params(W, b), data,  opt)
    println(loss(x, y))
end

What if my loss function is in the form `loss(x, y, z, W, b, c, d)` where `x, y, z` are the data for each batch, and `W, b` are the parameters I want to train, and `c, d` are the other fixed constant parameters for the loss function? How do I write/modify the above training code for this loss function?

Thanks!

DhairyaLGandhi:
There's many ways! First a little bit of context. The data you send in `train!` is in the form of `[itr1, itr2, ...]`. `itr` here is a collection of arguments to the loss function. Effectively, this will call `loss(itr...)`. So you need `itr` to be the form `(x, y, a, b, ...)`. If you're using a `DataLoader`, you want it to produce `itr` on iteration (hopefully that makes sense).

Another way, in the case you wish to supplement arguments that don't change, you can write your loss function as `(x...) -> loss(x..., arg1, arg2...)`.

See #1530 for how data loaders can be made to produce `itr` and DataLoaders.jl also.

KenZhenLin:
> Another way, in the case you wish to supplement arguments that don't change, you can write your loss function as `(x...) -> loss(x..., arg1, arg2...)`

Thanks a lot!

Also it seems that the parameters being updated have to be arrays. I observed that if ``W`` is an array, but ``b`` in ``params(W, b)`` is a real number, then ``params`` will ignore ``b``, and ``b`` will not be updated during the training. I had to make ``b`` a 1-d array to make things work. I noticed this by accident, so I wonder if this is the only restriction I need to pay attention to when doing the training.

darsnack:
[...]

PR #1731, opened 2021-10-02
Throw an error when someone calls params(::Params, ::Number, seen)

darsnack:
In response to #1730, this PR adds an error when someone tries to push a scalar onto `Params` (with a hint to wrap it as an array).

### PR Checklist

- [x] Tests are added
- [x] ~~Entry in NEWS.md~~
- [x] ~~API changes require approval from a committer (different from the author, if applicable)~~

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -39,7 +39,7 @@ Possible values include:
 trainmode!(m, mode = true) = mode isa Bool ? testmode!(m, !mode) : testmode!(m, mode)
 
 params!(p::Params, x::AbstractArray{<:Number}, seen = IdSet()) = push!(p, x)
-params!(::Params, ::Number, seen) = error("Tried to create params with a scalar (try wrapping it as a 1D array).")
+params!(::Params, ::Real, seen) = error("Tried to create params with a scalar (try wrapping it as a 1D array).")
 
 function params!(p::Params, x, seen = IdSet())
   x in seen && return
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -39,7 +39,9 @@ Possible values include:
 trainmode!(m, mode = true) = mode isa Bool ? testmode!(m, !mode) : testmode!(m, mode)
 
 params!(p::Params, x::AbstractArray{<:Number}, seen = IdSet()) = push!(p, x)
-params!(::Params, ::Real, seen) = @warn "Tried to create params with a scalar (try wrapping it as a 1D array)."
+params!(::Params, ::Real, seen) = Zygote.ignore() do
+  @warn "Tried to create params with a scalar (try wrapping it as a 1D array)."
+end
 
 function params!(p::Params, x, seen = IdSet())
   x in seen && return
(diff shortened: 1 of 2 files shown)


DhairyaLGandhi:
It can come as surprising behavior to have a number as something that can be mutated, but many higher order number types may be good candidates as paarms so we may want a less strict error here. More specifically, I am thinking that users only need dictate how to "update" their number type, and we can cast it to the correct type internally. We plan on doing that in Optimisers.jl (for working with images directly, say) so we would have the infrastructure to support this already.

src/functor.jl
```diff
@@ -39,6 +39,7 @@ Possible values include:
 trainmode!(m, mode = true) = mode isa Bool ? testmode!(m, !mode) : testmode!(m, mode)
 
 params!(p::Params, x::AbstractArray{<:Number}, seen = IdSet()) = push!(p, x)
+params!(::Params, ::Number, seen) = error("Tried to create params with a scalar (try wrapping it as a 1D array).")
```
DhairyaLGandhi, line 42:
```suggestion
params!(::Params, ::Real, seen) = error("Tried to create params with a scalar (try wrapping it as a 1D array).")
```

ToucheSir (commented):
I agree that we're not entirely sure what code in the wild might be doing and would thus suggest softening the error to a warning. That said, there's no way to use immutable number types with implicit params anyhow unless you're somehow maintaining an external mapping from path in model -> value (and praying the same value doesn't show up in 2 different locations and confusing `objectid`!) Either way, the overload needs an `ignore` to keep Zygote happy.

CarloLucibello:
I think this is too breaking. It's not uncommon to use `@functor MyType` and rely on the fact that scalar fields don't end up in the params.

darsnack:
All right, I've downgraded the message to a warning and restricted to `Real` instead of `Number`. Hopefully that allows more advanced number types while still being helpful for the common case error.

A warning should not be breaking and be soft enough to lift once Optimisers.jl stuff lands.

mcabbott:
This looks too strict to me. Storing some numbers which aren't parameters in a struct seems routine, for instance: 
```
julia> Flux.params(Conv((3,), 3 => 4))
┌ Warning: Tried to create params with a scalar (try wrapping it as a 1D array).
└ @ Flux REPL[59]:2
┌ Warning: Tried to create params with a scalar (try wrapping it as a 1D array).
└ @ Flux REPL[59]:2
┌ Warning: Tried to create params with a scalar (try wrapping it as a 1D array).
└ @ Flux REPL[59]:2
┌ Warning: Tried to create params with a scalar (try wrapping it as a 1D array).
└ @ Flux REPL[59]:2
Params([[-0.40309134 -0.20597205 0.08528712; -0.28718993 0.08592164 0.356715; 0.090567335 -0.2432247 0.47673503;;; -0.38878575 0.41411868 0.07338351; -0.2379244 -0.048532628 0.46593264; 0.23170501 -0.11007663 0.17844321;;; 0.11771456 -0.057944458 -0.37289232; -0.4776257 0.33483347 0.53414; 0.3548786 -0.24201536 0.3229901;;; 0.52154154 0.22303507 0.40377334; -0.4988539 -0.34985122 0.25671232; 0.38871452 0.1299994 0.0023408195], Float32[0.0, 0.0, 0.0, 0.0]])
```
You could restrict it to AbstractFloat, maybe that's less common... but the idea is to catch user mistakes like `params(rand(3), 4.5)` right? Could those be caught closer to source?

darsnack:
Yeah looks like we'll have to catch it earlier

CarloLucibello:

PR #1735, opened 2021-10-05
Don't export flatten

ChrisRackauckas:
```julia
julia> WARNING: both Flux and Iterators export "flatten"; uses of it in module DiffEqFlux must be qualified
```

Yeah no, that interferes with a very common Base module.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -10,7 +10,7 @@ using MacroTools: @forward
 using Zygote: Params, @adjoint, gradient, pullback, @nograd
 export gradient
 
-export Chain, Dense, Maxout, SkipConnection, Parallel, flatten,
+export Chain, Dense, Maxout, SkipConnection, Parallel,
        RNN, LSTM, GRU, GRUv3,
        SamePad, Conv, CrossCor, ConvTranspose, DepthwiseConv,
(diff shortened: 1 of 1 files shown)


ToucheSir:
Given `flatten` has been exported for [over 2 and a half years](https://github.com/FluxML/Flux.jl/commit/746e3310f18485c0c30c9975c71c88d53d00fe26) now, this would have to be a breaking change.

ChrisRackauckas:
Could anyone have ever actually used this though? If one package in the entire stack uses Iterators it's going to have an issue. I've never seen a package export something that shadows Base 😅

ToucheSir:
It's come up a few times in MWEs in the wild. I imagine this hasn't come up already because most folks use `Iterators` functions off the module. I don't think I've ever seen `using .Iterators` before :shrug:

ToucheSir:
Oh don't get me wrong, we shouldn't be exporting this. However, I don't think we should make this change until we're ready to cut a breaking release. It's a super trivial one, so as long as someone remembers we'll be good to go.

DhairyaLGandhi:
`flatten` is used. It clashing with a Base submodule is unfortunate.

CarloLucibello:
I don't see many breaking changes on the horizon, right? We could just unexport `flatten` and `params`, remove a few deprecations and tag v0.13

CarloLucibello:
Closing this in favor of #1751

PR #1742, opened 2021-10-11
Have destructure return only trainable params

### LINKED ISSUE #1727: Normalization Layers not interating well with destructure/restructure
MWE:
```julia
using Flux

model = Chain(Dense(2, 10, relu), BatchNorm(10))
p, re = Flux.destructure(model)

x = rand(Float32, 2, 1)

a, back = Flux.pullback(x, p) do _x, _p
    vec(re(_p)(_x))
end

back(a)
# ┌ Warning: Expected 70 params, got 30
# └ @ Flux ~/.julia/packages/Flux/ZnXxS/src/utils.jl:623
```

### LINKED ISSUE #1732: missing docstring for `Flux.params` and `trainable`
```julia
help?> Flux.params
  No documentation found.

  Flux.params is a Function.

  # 1 method for generic function "params":
  [1] params(m...) in Flux at /home/carlo/.julia/dev/Flux/src/functor.jl:51
```

### LINKED ISSUE #1733: inconsistency between params and destructure
In my understanding,  `destructure` is supposed to behave similarly to `params` and collect *trainable* params of a struct into a vector. Instead, it collects *all* arrays, as one can see in this example using BatchNorm:
```julia
julia> b = BatchNorm(2)
BatchNorm(2)        # 4 parameters, plus 4 non-trainable

julia> ps = params(b)

julia> p, re = Flux.destructure(b)
(Float32[0.0, 0.0, 1.0, 1.0, 0.0, 0.0, 1.0, 1.0], Flux.var"#60#62"{BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}}}(BatchNorm(2)))

julia> length(p)
8
```

Should we modify `destructure` to act like `params`, therefore recurse over `trainable(m)` instead of just applying `fmap` as it currently does?

Also, maybe we should add to both `params` and `destructure` a keyword argument `which` with possible values `:trainable`, `:buffer`, `:all`, and default value `:trainable`.

Related to #1727

CarloLucibello:
Have destructure return only trainable params
+
functorize RefValue (this part should go into Functors.jl)

Fix #1733, fix #1727, fix #1732

Adding also tests from #1614

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -1,7 +1,6 @@
 import Adapt: adapt, adapt_storage
 using  LinearAlgebra: Cholesky
 using Zygote: IdSet
-import Functors: Functors, @functor, functor, fmap, isleaf
 using SparseArrays: AbstractSparseArray
 
 trainable(m) = functor(m)[1]
@@ -37,6 +36,124 @@ Possible values include:
 - `:auto` or `nothing` for Flux to detect the mode automatically
 """
 trainmode!(m, mode = true) = mode isa Bool ? testmode!(m, !mode) : testmode!(m, mode)
+
+
+# Flattening models to weight vectors, and back
+
+function _restructure(m, xs)
+  i = 0
+  filter = (x, c) -> any(y -> c === y, trainable(x))
+  walk = filtered_walk(filter)
+  m̄ = fmap(m; walk) do x
+    x isa AbstractArray{<:Number} || return x
+    x = reshape(xs[i .+ (1:length(x))], size(x))
+    i += length(x)
+    return x
+  end
+  length(xs) == i || @warn "Expected $(i) params, got $(length(xs))"
+  return m̄
+end
+
+@adjoint function _restructure(m, xs)
+  m̄, numel = _restructure(m, xs), length(xs)
+  function _restructure_pullback(dm)
+    xs′ = destructure(dm)[1]
+    numel == length(xs′) || @warn "Expected $(numel) params, got $(length(xs′))"
+    return (nothing, xs′)
+  end
+  return m̄, _restructure_pullback
+end
+
+"""
+    destructure(m)
+Flatten a model's parameters into a single weight vector.
+    julia> m = Chain(Dense(10, 5, σ), Dense(5, 2), softmax)
+    Chain(Dense(10, 5, σ), Dense(5, 2), softmax)
+    julia> θ, re = destructure(m);
+    julia> θ
+    67-element Vector{Float32}:
+    -0.1407104
+    ...
+The second return value `re` allows you to reconstruct the original network after making
+modifications to the weight vector (for example, with a hypernetwork).
+    julia> re(θ .* 2)
+    Chain(Dense(10, 5, σ), Dense(5, 2), softmax)
+"""
+function destructure(m)
+  xs = Zygote.Buffer([])
+  collect_params!(xs, m)
+  return vcat(vec.(copy(xs))...), p -> _restructure(m, p)
+end
+
+function collect_params!(xs, m)
+  filter = (x, c) -> any(y -> c === y, trainable(x))
+  walk = filtered_walk(filter)
+  fmap(m; walk) do x
+    x isa AbstractArray{<:Number} && push!(xs, x)
+    return x
+  end
+end
+
+function filtered_walk(filter)
+  seen = IdSet()
+
+  function walk(f, x)
+    x in seen && return x
+    push!(seen, x)
+
+    children, reconstruct = functor(x)
+    mappedchildren = map(children) do c
+      filter(x, c) ? f(c) : c
+    end
+    reconstruct(mappedchildren)
+  end
+
+  return walk
+end
+
+
+"""
+  params(m...)
+
+Collect trainable parameters (a.k.a. numerical arrays)
+from the input model(s) `m` into a [`Zygote.Params`](@ref) object. 
+
+Only the parameters that can be reached by recursion 
+on the [`trainable`](@ref) children of
+the tree with root `m` are collected.
+
+# Usage
+
+```julia-repl 
+julia> m = Dense(ones(2, 3), zeros(2))
+Dense(3, 2)         # 8 parameters
+
+julia> ps = Flux.params(m)
+Params([[1.0 1.0 1.0; 1.0 1.0 1.0], [0.0, 0.0]])
+
+julia> x = ones(3)
+3-element Vector{Float64}:
+ 1.0
+ 1.0
+ 1.0
+
+julia> gs = gradient(() -> sum(2 .* m(x)), ps)
+Grads(...)
+
+julia> gs[m.weight]
+2×3 Matrix{Float64}:
+ 2.0  2.0  2.0
+ 2.0  2.0  2.0
+```
+"""
+function params end
+
+## TODO This causes some test regressions. Why?
+# function params(m...)
+#   ps = Params()
+#   collect_params!(ps, m)
+#   return ps
+# end
 
 params!(p::Params, x::AbstractArray{<:Number}, seen = IdSet()) = push!(p, x)
 
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -101,6 +101,9 @@ Chain(
 )         # Total: 6 trainable arrays, 77 parameters,
           # plus 2 non-trainable, 10 parameters, summarysize 836 bytes.
 ```
+
+Only numerical arrays are collected by `destructe`. Moreover, if the same array is nested multiple times in the same model (e.g. shared by some layers)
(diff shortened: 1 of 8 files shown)


CarloLucibello:
> I would go for a more specific fix for the sciml failures. We have seen the *(::Tuple) failures in a number of places. @ChrisRackauckas #1727 (comment) could you try with this definition of destructure. The failures related to CR would either need to be fixed there or we use a different adjoint.

This comment is not very clear, but maybe you are confusing things. The RefValue problem related to https://github.com/FluxML/Flux.jl/issues/1727#issuecomment-931749324 has already been fixed in Functors and Zygote. 
Here we fix the residual problem of having destructure return only trainable params (as the title suggests). 
The CR/Zygote changes impacting NeurlPDE and DiffEqFlux are not related to Flux or this PR, and undoubtedly will be fixed somewhere else.

test/functor.jl
```diff
@@ -0,0 +1,165 @@
...
+
+nobias(n) = Zeros()
+
+function testdense(m, bt)
+  @testset "Check layer $i" for (i, (l1, l2)) in enumerate(zip(m, dm(bt)))
+    @test l1.weight == l2.weight
+    @test l1.bias == l2.bias
+    @test typeof(l1.bias) === typeof(l2.bias)
+  end
+end
```
mcabbott, line 21: Could we take the opportunity to nuke weird tests like this which define functions which define functions which zip some numbers and magically maybe test all cases?

In favour of just a sampling of human-readable tests which look like things in the manual.

src/functor.jl
```diff
@@ -38,6 +37,124 @@ Possible values include:
...
+    numel == length(xs′) || @warn "Expected $(numel) params, got $(length(xs′))"
+    return (nothing, xs′)
+  end
+  return m̄, _restructure_pullback
+end
+
+"""
+    destructure(m)
+Flatten a model's parameters into a single weight vector.
+    julia> m = Chain(Dense(10, 5, σ), Dense(5, 2), softmax)
```
mcabbott, line 70: This wants to be a jldoctest, and should include something like BatchNorm, to illustrate which parameter count becomes the length of θ.

It should also say that `x isa AbstractArray{<:Number}` and unique `objectid(x)` are the criteria for inclusion.

src/functor.jl
```diff
@@ -38,6 +37,124 @@ Possible values include:
...
+    ...
+The second return value `re` allows you to reconstruct the original network after making
+modifications to the weight vector (for example, with a hypernetwork).
+    julia> re(θ .* 2)
+    Chain(Dense(10, 5, σ), Dense(5, 2), softmax)
+"""
+function destructure(m)
+  xs = Zygote.Buffer([])
+  collect_params!(xs, m)
+  return vcat(vec.(copy(xs))...), p -> _restructure(m, p)
```
mcabbott, line 85: Why does this copy? (And splat?)

And, how easy would it be to avoid Buffer somehow, to make this ready for not using Zygote?

src/functor.jl
```diff
@@ -38,6 +37,124 @@ Possible values include:
...
+function collect_params!(xs, m)
+  filter = (x, c) -> any(y -> c === y, trainable(x))
+  walk = filtered_walk(filter)
+  fmap(m; walk) do x
+    x isa AbstractArray{<:Number} && push!(xs, x)
+    return x
+  end
+end
+
+function filtered_walk(filter)
```
mcabbott, line 97: I think this should be `filtered_walk(cond::Function)`, with a type, and not re-using Base names.

src/functor.jl
```diff
@@ -38,6 +37,124 @@ Possible values include:
...
+  end
+
+  return walk
+end
+
+
+"""
+  params(m...)
+
+Collect trainable parameters (a.k.a. numerical arrays)
```
mcabbott, line 118: Not in fact a.k.a., since some arrays aren't trainable.

mcabbott (commented):
Slightly random comments, mostly on things not new to this PR:

test/functor.jl
CarloLucibello: I simplified and made more readable these tests, still not ideal but a bit better

src/functor.jl
mcabbott: A related question is: Is this the right test? Should these be independent?
```
julia> x = [1,2,3];

julia> objectid(x), objectid(x'), objectid(transpose(x))
(0x3aed6805416fa931, 0xab1cb79f2fe03e53, 0x731d5dafc3a51f2b)
```
You could for instance keep calling `parent` until it stops, and that the parameter. Or maybe such types should be unwrapped by Functors?

src/functor.jl
```diff
@@ -38,25 +37,126 @@ Possible values include:
...
+Chain(
+  Dense(10, 5, relu),                   # 55 parameters
+  BatchNorm(5),                         # 10 parameters, plus 10
+  Dense(5, 2),                          # 12 parameters
+  NNlib.softmax,
+)         # Total: 6 trainable arrays, 77 parameters,
+          # plus 2 non-trainable, 10 parameters, summarysize 836 bytes.
+```
+
+Only numerical arrays are collected by `destructe`. Moreover, if the same array is nested multiple times in the same model (e.g. shared by some layers)
```
ToucheSir, line 105:
```suggestion
Only numerical arrays are collected by `destructure`. Moreover, if the same array is nested multiple times in the same model (e.g. shared by some layers),
```

src/functor.jl
CarloLucibello: With 
```julia
function destructure(m)
  xs = AbstractArray[]
  collect_params!(xs, m)
  return vcat(vec.(xs)...), p -> _restructure(m, p)
end
```
Flux's tests still pass. I still have to test the interaction with DiffEqFlux, NeuralPDE, let's see

src/functor.jl
```diff
@@ -38,25 +37,126 @@ Possible values include:
...
+    return x
   end
+  length(xs) == i || @warn "Expected $(i) params, got $(length(xs))"
+  return m̄
 end
 
+@adjoint function _restructure(m, xs)
+  m̄, numel = _restructure(m, xs), length(xs)
+  function _restructure_pullback(dm)
+    xs′ = destructure(dm)[1]
```
mcabbott, line 60: This gradient can easily be wrong, it looks for duplicates *in the gradient* which can come from e.g. adding two parameters `x + y`. It is completely unaware of duplicates *in the original*. 

Demonstration:
https://github.com/FluxML/Flux.jl/issues/1826#issuecomment-1013783952

So at very least, we must (1) disable the removal of duplicates from destructure used here, and (2) throw an error if you try to use this adjoint when the original model had any gradients.

Or, failing that, we should remove it from v0.13 until someone can write a version which actually works.

src/functor.jl
CarloLucibello: As you know the situation with wrapped shared parameters is pretty complex, I'm not going to address those issues here

src/functor.jl
CarloLucibello: DiffEqFlux tests pass with both `Buffer()` and `AbstractArray[]`. 

@ChrisRackauckas you see any particular reason to keep Buffer?

ChrisRackauckas: I don't see a reason to use Buffer here.

ChrisRackauckas: I guess that's for some potential higher order AD issue?

ToucheSir: Or just first-order AD. AFAICT Flux's current test suite never tests the gradient of `destructure` (only restructuring) :see_no_evil:...

mcabbott: 🙈 is the Flux motto, really.

mcabbott: I think I wrote the correct de/restructure functions & gradients out last weekend. But not sure I have the stomach for trying to push them.

CarloLucibello:
@mcabbott should I close this? Maybe you have some better solution at this point

mcabbott:
I have written many things, e.g. https://github.com/FluxML/Functors.jl/pull/31, but they aren't going to be ready tomorrow. Maybe v0.13 should change the documented behaviour to trainable, without aspiring to fix all the bugs?

How difficult would it be to at least throw an error if there are repeated paramters?

mcabbott:
This is now the last issue on the v0.13 milestone. 

I think what we should do is simply delete `destructure` completely, and call the one from Optimisers. Apart from fixing bugs it ought to be a drop-in replacement for the one here -- that is, it also only keeps trainable parameters.

darsnack:
Agreed, we should try that and run the downstream tests. If everything passes, then there is no reason not to.

CarloLucibello:
agreed. I'll leave this PR open as a reminder for the milestone, but I think it is better to start clean in a new PR and cherry-pick from here some of the tests

PR #1746, opened 2021-10-13
Fix Saving and loading model output example

logankilpatrick:
Right now, the example here: https://fluxml.ai/Flux.jl/stable/saving/#Saving-and-Loading-Models is incorrect.

```Julia
julia> model = Chain(Dense(10,5,relu),Dense(5,2),softmax)
Chain(Dense(10, 5, NNlib.relu), Dense(5, 2), NNlib.softmax)
```

is not what happens and the loading model example at the bottom expects `Chain(Dense(10, 5, NNlib.relu), Dense(5, 2), NNlib.softmax)` so this fixes this inconsistency.

### DIFF
--- a/docs/src/saving.md
+++ b/docs/src/saving.md
@@ -10,7 +10,11 @@ Save a model:
 julia> using Flux
 
 julia> model = Chain(Dense(10, 5, NNlib.relu), Dense(5, 2), NNlib.softmax)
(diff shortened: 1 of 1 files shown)


logankilpatrick:
I guess I am on a very old version `Flux v0.8.3` but when I do `up Flux`, it does not change.

darsnack:
Okay let's merge this for now then.

bors r+

Issue #1745, opened 2021-10-13
Link to "train!" gives 404 page not found error on the website.


Chandu-4444:
I was looking through the docs [here](https://fluxml.ai/Flux.jl/stable/models/overview/). And when I press on `train!`, it redirected me to a 404 page.

logankilpatrick:
This will be closed by #1784 I believe

Chandu-4444:
> This will be closed by #1784 I believe

Yeah. I think the docs aren't updated yet to reflect the changes right? Cause I still get the same 404 error.

logankilpatrick:
You have to switch to the dev docs. It works for one of the mentions of `train` but the other does not yet work. They should both be fixed once that PR is merged (you will have to used the dev branch till a new release is tagged).

Chandu-4444:
Oh, ok thanks 👍

logankilpatrick:
We can leave this open till the PR is merged and verified it fixes the issue.

logankilpatrick:
Fixed!

PR #1748, opened 2021-10-16
Fix `train!` doc string 404

logankilpatrick:


logankilpatrick:
Closes #1745

logankilpatrick:
I think this is right since `train!` does not actually exist, it is https://fluxml.ai/Flux.jl/stable/training/training/#Flux.Optimise.train!

logankilpatrick:
From the [Documenter.jl docs](https://juliadocs.github.io/Documenter.jl/stable/man/syntax/#@ref-link):
> Note that depending on what the CurrentModule is set to, a docstring @ref may need to be prefixed by the module which defines it.

DhairyaLGandhi:
It should be exported from the Optimise submodule, and therefor in the Flux namespace? Either way, we can include this for the time being.

PR #1751, opened 2021-10-17
v0.13 deprecations

### LINKED ISSUE #629: Flux type piracy breaks REPL completions
```
> Base.show(<TAB>┌ Error: Error in the keymap
│   exception =
│    UndefVarError: T not defined
│    Stacktrace:
│     [1] show(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Type{Flux.Data.Tree{T}}) at C:\Users\Kristoffer\MachineLearning\dev\Flux\src\data\tree.jl:17
│     [2] show_datatype(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::DataType) at .\show.jl:526
│     [3] show(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::DataType) at .\show.jl:436
│     [4] print(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Type) at .\strings\io.jl:31
│     [5] string_with_env(::Base.ImmutableDict{Symbol,Any}, ::Type) at .\strings\io.jl:140
│     [6] argtype_decl(::Base.ImmutableDict{Symbol,Any}, ::Symbol, ::DataType, ::Int64, ::Int32,
::Bool) at .\methodshow.jl:42
│     [7] arg_decl_parts(::Method) at .\methodshow.jl:69
│     [8] #show#520(::DataType, ::Function, ::Base.GenericIOBuffer{Array{UInt8,1}}, ::Method) at
.\methodshow.jl:113
```

Type piracy is here:

https://github.com/FluxML/Flux.jl/blob/78876a14b3148e03a4238720f7b43091f4f4fb66/src/data/tree.jl#L16-L17

### LINKED ISSUE #1739: failing doc tests
The following doctests are currently failing due to Zygote now returning float where it used to return integers. 
I think this is intended and we should just fix the docs. Something that a bit annoys me though is that we have some `-0.0` instead of `0.0`. cc @mcabbott 
```
┌ Error: doctest failure in ~/work/Flux.jl/Flux.jl/docs/src/models/basics.md:25-30
│ 
│ ```jldoctest basics
│ julia> f(x, y) = sum((x .- y).^2);
│ 
│ julia> gradient(f, [2, 1], [2, 0])
│ ([0, 2], [0, -2])
│ ```
│ 
│ Subexpression:
│ 
│ gradient(f, [2, 1], [2, 0])
│ 
│ Evaluated output:
│ 
│ ([0.0, 2.0], [-0.0, -2.0])
│ 
│ Expected output:
│ 
│ ([0, 2], [0, -2])
│ 
│   diff =
│    Warning: Diff output requires color.
│    ([0, 2], [0, -2])([0.0, 2.0], [-0.0, -2.0])
└ @ Documenter.DocTests ~/.julia/packages/Documenter/bFHi4/src/DocTests.jl:373
┌ Error: doctest failure in ~/work/Flux.jl/Flux.jl/docs/src/models/basics.md:37-56
│ 
│ ```jldoctest basics
│ julia> x = [2, 1];
│ 
│ julia> y = [2, 0];
│ 
│ julia> gs = gradient(params(x, y)) do
│          f(x, y)
│        end
│ Grads(...)
│ 
│ julia> gs[x]
│ 2-element Vector{Int64}:
│  0
│  2
│ 
│ julia> gs[y]
│ 2-element Vector{Int64}:
│   0
│  -2
│ ```
│ 
│ Subexpression:
│ 
│ gs[x]
│ 
│ Evaluated output:
│ 
│ 2-element Vector{Float64}:
│  0.0
│  2.0
│ 
│ Expected output:
│ 
│ 2-element Vector{Int64}:
│  0
│  2
│ 
│   diff =
│    Warning: Diff output requires color.
│    2-element Vector{Int64}:
│     0
│     2Vector{Float64}:
│     0.0
│     2.0
└ @ Documenter.DocTests ~/.julia/packages/Documenter/bFHi4/src/DocTests.jl:373
┌ Error: doctest failure in ~/work/Flux.jl/Flux.jl/docs/src/models/basics.md:37-56
│ 
│ ```jldoctest basics
│ julia> x = [2, 1];
│ 
│ julia> y = [2, 0];
│ 
│ julia> gs = gradient(params(x, y)) do
│          f(x, y)
│        end
│ Grads(...)
│ 
│ julia> gs[x]
│ 2-element Vector{Int64}:
│  0
│  2
│ 
│ julia> gs[y]
│ 2-element Vector{Int64}:
│   0
│  -2
│ ```
│ 
│ Subexpression:
│ 
│ gs[y]
│ 
│ Evaluated output:
│ 
│ 2-element Vector{Float64}:
│  -0.0
│  -2.0
│ 
│ Expected output:
│ 
│ 2-element Vector{Int64}:
│   0
│  -2
│ 
│   diff =
│    Warning: Diff output requires color.
│    2-element Vector{Int64}:
│      0
│     -2Vector{Float64}:
│     -0.0
│     -2.0
```

### LINKED ISSUE #1779: Remove Juno.jl dependency
Given that Juno.jl is in maintenance-only mode and VSC is the recommended editor now, it should be considered to remove that dependency.

CarloLucibello:
- Removing datasets (have been deprecated for a year or so)
- unexporting `flatten` and `params`
- fixes a doctest issue #1739
- removed deprecated `getproperty` for recurrent layers.  
- deprecated `weight` keyword argument in Conv constructor

I guess the only controversial point is if we want to unexport `params`. That could cause some (maybe a lot) of breakage, although we have been using it mostly in a qualified way in docs and model-zoo examples. On one hand, I wish it was never exported (it's a generic and common name, and it clashes with the exported `Distributions.params`). On the other hand, I'm not sure it is worth causing breakage. 

Replaces #1735, Fix #1739, fix #629, fix #1779


mcabbott:
We could remove deprecations for Dense's getproperty, ref  https://github.com/FluxML/Zygote.jl/issues/1094#issuecomment-945972944 . And constructors too.

CarloLucibello:
> We could remove deprecations for Dense's getproperty, ref FluxML/Zygote.jl#1094 (comment) . And constructors too.

removed all v0.12 deprecations introduced more than 6 months ago

CarloLucibello:
@mcabbott there is also this https://github.com/FluxML/Flux.jl/blob/f8ebeadc8c206c3f56af0f654a4b729c3d0b7292/src/losses/utils.jl#L26
We have to decide if we want to make it an error already. It's been there 4 months, maybe we can wait the next release.

mcabbott:
Could go either way on that. As a middle path, can upgrade `@warn` to `@error` to make it louder?

Other things from a quick look today...

* `params` got a docstring 5 minutes ago #1786 but this should say `Flux.params`

* Juno https://github.com/FluxML/Flux.jl/issues/1779 -- if making a breaking release, could we just kill this, and add back other progress bars later, if someone is inspired?

* Parallel #1685 via #1698 or something lighter which just changes it to always call the combinator exactly once, nothing else.

* Channels with Pairs,  #1658. I share your discomfort with having two ways, but do think this would be clearer. Perhaps we should just bite the bullet and move the `,` method to depreciations & out of documentation, even if we never delete it.

* Some layer constructors have accumulated redundant methods. Letting `Conv((2,-99), 3=>4; weight=rand(2,3,4,5))` work without error doesn't sound like a good design, but removing it is breaking.

* Instead of #1272 we can now use `NNlib.fast_tanh`. Not certain whether making layers use this counts as a breaking change, changes should be tiny floating point effects.

ToucheSir:
No. 1 is already covered by #1643, just needs a quick bump to resolve VCS conflicts.

mkschleg:
I didn’t see this in the list, but I think we wanted to remove the [getproperty for recur](https://github.com/FluxML/Flux.jl/blob/0d3bd0fac3a33995a988a39da1f08686b116b886/src/layers/recurrent.jl#L70) as well for v0.13.

src/layers/basic.jl
```diff
@@ -131,23 +131,10 @@ struct Dense{F, M<:AbstractMatrix, B}
...
-  b = if bias === true && initb !== nothing
-    Base.depwarn("keyword initb is deprecated, please simply supply the bias vector, bias=initb(out)", :Dense)
-    initb(out)
-  else
-    bias
-  end
-
+  W = init(out, in)
+  b = bias
   return Dense(W, b, σ)
```
mcabbott, line 138:
```suggestion
  W = 
  b = bias
  return Dense(init(out, in), bias, σ)
```

CarloLucibello:
Implemented two of the @mcabbott's suggestions
> - Remove Juno.jl dependency #1779 -- if making a breaking release, could we just kill this, and add back other progress bars later, if someone is inspired?
> - Some layer constructors have accumulated redundant methods. Letting Conv((2,-99), 3=>4; weight=rand(2,3,4,5)) work without error doesn't sound like a good design, but removing it is breaking.

@mcabbott you want to add the other stuff to the v0.13 milestone?

PS I don't know why CI is being stupid complaining about the Manifest

src/layers/basic.jl
```diff
@@ -186,19 +173,9 @@ struct Diagonal{T}
...
-    Base.depwarn("keyword initβ is deprecated, please simply supply the desired vectors", :Diagonal)
-    initβ(sz...)
-  else
-    zeros32(sz...)
-  end
+function Diagonal(sz::Integer...)
+  α = ones32(sz...)
+  β = zeros32(sz...)
   Diagonal(α, β)
 end
```
mcabbott, line 180:
```suggestion
Diagonal(sz::Integer...) = Diagonal(ones32(sz...), zeros32(sz...))
```

src/layers/conv.jl
```diff
@@ -136,8 +134,13 @@ end
...
             init = glorot_uniform, stride = 1, pad = 0, dilation = 1, groups = 1,
-            weight = convfilter(k, ch; init, groups), bias = true) where N
-
+            weight = nothing, bias = true) where N
+    if weight !== nothing
+      # TODO remove in v0.14
+      Base.depwarn("The `weight` keyword arg is deprecated, use the Conv(weight, ...) constructor instead", :conv_weight)
+    else
+      weight = convfilter(k, ch; init, groups)
+    end
```
mcabbott, line 143: I would just delete these, as I think they were added recently enough and not advertised. Just reverting a mistake, not really removing a deliberate feature. But either way is fine if you don't mind the typing.

Same applies to ConvTranspose, DepthwiseConv, CrossCor

src/layers/conv.jl
```diff
@@ -61,8 +61,6 @@ Then:
 
 Keywords to control initialization of the layer:
 * `init` - Function used to generate initial weights. Defaults to `glorot_uniform`.
-* `weight` - Initial weights of the layer. Typically an array, and can be used to override
-  other configurations. By default, these are generated using [`convfilter`](@ref).
 * `bias` - Initial bias is zero by default, this can be disabled entirely by setting it to
   [`Flux.Zeros()`](@ref) or equivalently `false`, or another vector provided as
   `bias = randn(Float32, out)`.
```
mcabbott, line 66:
```suggestion
  `false`, or another vector explicitly as `bias = randn(Float32, out)`.
```

src/layers/conv.jl
```diff
@@ -61,8 +61,6 @@ Then:
```
mcabbott, line 61: Here "groups specifies groups" isn't very helpful.

mcabbott (commented):
For other changes:

I had a commit trying out `fast_tanh` on top of #1761, which we should merge. Some quite large forward speedups, modest with the gradient though, IIRC.

Parallel would also be good to sort out, haven't looked lately.

Edit: not on the list above, but 

* perhaps destructure & trainable #1742 should get a look?

* perhaps we should switch `Chain` to use `foldl` as in https://github.com/FluxML/Zygote.jl/issues/1126. There is a chance this may break 2nd derivatives; if we don't have any in the tests, then perhaps that's not a disaster -- it's a breaking change, and this will eventually be fixed. This is now #1809.

Also, ~~Juno is loaded here:~~ I pushed 2 commits here, hope you don't mind.

CarloLucibello:
ok I think this is good to go

src/layers/conv.jl
mcabbott: (but perhaps out of scope!)

mcabbott:
> * perhaps destructure & trainable #1742 should get a look?

One possibly evil idea is just to remove `destructure` since it's a breaking release.

Then we can make a PR adding it back, which must pass reasonably quality tests, i.e. It must have clear semantics, described in the docs; it must have tests, covering just about everything, and some attempt to invent adversarial examples. If you must have the old one, don't upgrade, until the new one is ready.

darsnack:
Isn't `destructure` critical for SciML stuff? Seems like a large "don't upgrade" umbrella. Instead of removing it, if we want to skip addressing it on this release, then let's just agree that a fast v0.14 is fine too (like @ToucheSir suggested on Slack).

mcabbott:
Well in my evil plan this desire to upgrade will inspire SciML types who need it to chip in to figuring out the solid PR adding this functionality -- to approximately ChainRules's standards. Instead of the thing being grandfathered in and intractable to work on without upsetting someone. But possibly this is too weird.

mcabbott (approved):
Since #1859 is merged, perhaps master is now sort-of 0.13-dev, and we should start merging things.

ToucheSir:
Another PR I'd like to submit for consideration is https://github.com/FluxML/Flux.jl/pull/1778. Making that non-breaking is going to be more of a pain than the alternative.

src/Flux.jl
```diff
@@ -4,19 +4,23 @@ module Flux
 
 using Base: tail
 using Statistics, Random, LinearAlgebra
+<<<<<<< HEAD
 using Zygote, MacroTools, ProgressLogging, Reexport
+=======
+using Zygote, MacroTools, Reexport
+>>>>>>> 3e86b468 (v0.13 deprecations)
```
mcabbott, line 11:
```suggestion
using Zygote, MacroTools, ProgressLogging, Reexport
```

PR #1752, opened 2021-10-17
Clarify that `params` updates

KronosTheLate:
I was confused when initally reading the overciew, because I though that `params` returned a tuple of vectors. I see now that the printing of `parameters` reveals the output to be something else, but I think it is relevant what a less experienced programmer may read it as. It later becomes apparent that it actually returns another data-structure, because it updates when parameters change. I have added a single sentence to clarify this point. Edits on the wording/phrasing are very welcome.

### DIFF
--- a/docs/src/models/overview.md
+++ b/docs/src/models/overview.md
@@ -117,7 +117,7 @@ julia> parameters = Flux.params(predict)
 Params([[-0.99009055], [0.0]])
 ```
 
-These are the parameters Flux will change, one step at a time, to improve predictions. Each time a parameter in your model changes, the `parameters` variable updates automatically. Each of the parameters comes from the `predict` model: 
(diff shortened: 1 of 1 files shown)


KronosTheLate:
What about `These are the parameters Flux will change, one step at a time, to improve predictions. The returned values are pointers, meaning that the variable `parameters` will always refer to the current parameters. Each of the parameters comes from the `predict` model: `?

On the verbose side, which I am not a fan on. Feel free to make suggestions/edits.

KronosTheLate:
Scrap my previous suggestion, I much prefer the following:

'parameters' will now always return the current model parameters. These are the parameters Flux will change, one step at a time, to improve predictions. Each of the parameters comes from the predict model:

Do you feel that it can be improved further, or should I edit the PR? (If you agree with the suggestion, you can also make the edit and merge in one motion if you'd like)

ToucheSir (approved):
Apologies for the delay, let's get this merged.

PR #1755, opened 2021-10-25
Remove Recurrent cell eltype restriction

### LINKED ISSUE #1565: Recurrent cell `eltype` restriction breaks `outputsize`
Since `outputsize` works by passing an array of `eltype(x) == Nil`, it will result in a `MethodError` when applied to recurrent cells.

rmsrosa:
Remove type restriction in the methods associated with recurrent cells `RNN`, `LSTM`, `GRU` and `GRUv3`.

Change conversion of `Nil` to other types, so `Nil()` gets converted to `NaN16, NaN32, NaN` for the corresponding float types and also complex types with float real and imaginary parts. Conversion errors for Integer and Rational types, in accordance to the errors when converting non integer floats.

This allows `outputsize` to work with recurrent cells as well.

Suitable tests were added.

This should fix #1565.

### PR Checklist

- [X ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -100,7 +100,7 @@ end
 RNNCell(in::Integer, out::Integer, σ=tanh; init=Flux.glorot_uniform, initb=zeros32, init_state=zeros32) = 
   RNNCell(σ, init(out, in), init(out, out), initb(out), init_state(out,1))
 
-function (m::RNNCell{F,A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {F,A,V,T}
+function (m::RNNCell)(h, x::Union{AbstractVecOrMat,OneHotArray})
   σ, Wi, Wh, b = m.σ, m.Wi, m.Wh, m.b
   h = σ.(Wi*x .+ Wh*h .+ b)
   sz = size(x)
@@ -154,7 +154,7 @@ function LSTMCell(in::Integer, out::Integer;
   return cell
 end
 
-function (m::LSTMCell{A,V,<:NTuple{2,AbstractMatrix{T}}})((h, c), x::Union{AbstractVecOrMat{T},OneHotArray}) where {A,V,T}
+function (m::LSTMCell)((h, c), x::Union{AbstractVecOrMat,OneHotArray})
   b, o = m.b, size(h, 1)
   g = m.Wi*x .+ m.Wh*h .+ b
   input = σ.(gate(g, o, 1))
@@ -222,7 +222,7 @@ end
 GRUCell(in, out; init = glorot_uniform, initb = zeros32, init_state = zeros32) =
   GRUCell(init(out * 3, in), init(out * 3, out), initb(out * 3), init_state(out,1))
 
-function (m::GRUCell{A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {A,V,T}
+function (m::GRUCell)(h, x::Union{AbstractVecOrMat,OneHotArray})
   b, o = m.b, size(h, 1)
   gx, gh, r, z = _gru_output(m.Wi, m.Wh, b, x, h)
   h̃ = tanh.(gate(gx, o, 3) .+ r .* gate(gh, o, 3) .+ gate(b, o, 3))
@@ -276,7 +276,7 @@ GRUv3Cell(in, out; init = glorot_uniform, initb = zeros32, init_state = zeros32)
   GRUv3Cell(init(out * 3, in), init(out * 2, out), initb(out * 3), 
             init(out, out), init_state(out,1))
 
-function (m::GRUv3Cell{A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {A,V,T}
+function (m::GRUv3Cell)(h, x::Union{AbstractVecOrMat,OneHotArray})
   b, o = m.b, size(h, 1)
   gx, gh, r, z = _gru_output(m.Wi, m.Wh, b, x, h)
   h̃ = tanh.(gate(gx, o, 3) .+ (m.Wh_h̃ * (r .* h)) .+ gate(b, o, 3))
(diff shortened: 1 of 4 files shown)


jeremiedb:
I'm afraid the proposed change results in hiding an error that arise from inconsistent eltype of input (X) and cell's initial state's (`state0`) when getting the gradients on GPU. For example, if feeding a RNN with Float64 while the model's `state0` is initialized with defaults Float32: 


```julia
using Flux
using Statistics: mean
using Random: seed!
using BenchmarkTools
using CUDA

#####################################
# RNN vanilla
#####################################
seed!(123)
feat = 8
h_size = 3
seq_len = 5
batch_size = 15

rnn = RNN(feat, h_size)

X = [rand(Float64, feat, batch_size) for i in 1:seq_len]
Y = rand(Float64, batch_size, seq_len) ./ 10

#### transfer to gpu ####
rnn_gpu = rnn |> gpu
X_gpu = CuArray{Float64}.(X)
Y_gpu = Float64.(Y)

θ = Flux.params(rnn)
θ_gpu = Flux.params(rnn_gpu)

function loss_cpu(x, y)
    # Flux.reset!(rnn)
    l = sum(Flux.stack(map(rnn, x), 2))
    return l
end

function loss_gpu(x, y)
    l = sum(Flux.stack(map(rnn_gpu, x), 2))
    return l
end

loss_cpu(X, Y)
loss_gpu(X_gpu, Y_gpu)

opt = ADAM(1e-3)
Flux.train!(loss_cpu, θ, [(X, Y)], opt)

opt_gpu = ADAM(1e-3)
Flux.train!(loss_gpu, θ_gpu, [(X_gpu, Y_gpu)], opt_gpu)

```

CPU's `train!` works fine, while GPU's results in:

```julia
ERROR: LoadError: this intrinsic must be compiled to be called
Stacktrace:
  [1] macro expansion
    @ C:\Users\jerem\.julia\packages\Zygote\rv6db\src\compiler\interface2.jl:0 [inlined]
  [2] _pullback(::Zygote.Context, ::Core.IntrinsicFunction, ::String, ::Type{Int64}, ::Type{Tuple{Ptr{Int64}}}, ::Ptr{Int64})
...
```
 
If using Float32 instead: `X_gpu = CuArray{Float32}.(X)`, there's no issue. 

Also, under the current forced parametrization on the RNNCell forward path definition function, the issue would be catched in advance through the missmatched types: 

```julia
ERROR: LoadError: MethodError: no method matching (::Flux.RNNCell{typeof(tanh), CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}, CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}})(::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, ::CuArray{Float64, 2, CUDA.Mem.DeviceBuffer})
Closest candidates are:
  (::Flux.RNNCell{F, A, V, var"#s526"} where var"#s526"<:AbstractMatrix{T})(::Any, ::Union{AbstractMatrix{T}, AbstractVector{T}, Flux.OneHotArray}) where {F, A, V, T} at 
c:\github\Flux.jl\src\layers\recurrent.jl:104
```

A way to properly handle Float32 params with Float64 input could be to specify the `init_state` type: 
```julia
rnn = RNN(feat, h_size, init_state=zeros)
```

which results in a valid model on CPU. However, moving to GPU, the ` |> gpu` approach converts all model parameters to Float32, which keeps the problem in place. 

I guess ideal scenario is for the GPU gradient calculation to handle where `state0` type differs from input/hidden state type, but I miss the underlying CUDA implications :\

DhairyaLGandhi:
Eltype promotion is how Julia handles multiple code paths and is a necessary component in (say) working with complex data, or duals etc. We also have tools to manage eltypes consistently (`f64`, `f32`), and type stability is a pretty common pattern to follow too. In `conv` we warn on mismatched eltypes but don't error, so this can be made consistent.

darsnack:
I think having at least a warning here is important. It's a more complex issue than the kind of promotion that occurs for convs, because of the type of the storage field.

I do think the better long term solution is an API. Per Julia's promotion system, the expected state type should be the promotion of the weights _and_ the input. Right now we "guess" the state type based on the weights alone.

jeremiedb:
Maybe there's a necessary compromise here since if the Recur's state isn't parametrized (in the struct definition), then the promotion will happen successfully on the state, even on GPU. However, Recur's output type then cannot be inferred, which has resulted in reported performance issue. 
But if the `intrinsic must be compiled` error on GPU can be resolved, it would result in the desired behavior right? Would be worth opening an issue in CUDA.jl?

As a side note, if one is aware of current gotchas and wants to initialize a model with Float32 weights and Float64 input and state, an issue is that that gpu conversion moves everything into Float32, thus losing the required Float64 for state:

```julia
rnn_cpu = RNN(feat, h_size, init=Flux.zeros32, initb=Flux.zeros32, init_state=zeros)
rnn_gpu = rnn_cpu |> gpu
julia> rnn_cpu.state
3×1 Matrix{Float64}:
 0.0
 0.0
 0.0

julia> rnn_gpu.state
3×1 CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}:
 0.0
 0.0
 0.0
```

Wouldn't it be desirable for `gpu` to preserve the types? If such was the case, then the handling of different types for weights vs input/state would be at least more convenient as the above would result in valid coded for GPU.

DhairyaLGandhi:
> Wouldn't it be desirable for gpu to preserve the types?

I'm not sure on that last bit, since GPUs can have very different performance properties depending on precision. Either way, if we know ahead of time that `gpu` produces `f32` we can much easier predict the eltypes we will get. Note that the error raised successfully prevents the user from mixing eltypes on the GPU which can have vastly larger differences in what performance we see. Is it common for users to need different eltypes for different parameters? Most cases I have seen don't seem to mind f32 everywhere, and don't actually intend to use the full precision of f64. Relaxing the eltypes makes it way easier to handle mixed precision cases without needing to hold on to too many specific kernels for different eltypes.

rmsrosa:
superseded by #2234

PR #1758, opened 2021-10-27
Add custom model example to docs.

Gregliest:
This PR adds a simple custom model example to the docs. As discussed in Slack, this example would have helped me understand how all the pieces fit together. But, happy to edit/move/delete as desired. 

cc @ToucheSir 

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/models/advanced.md
+++ b/docs/src/models/advanced.md
@@ -30,7 +30,7 @@ model = CustomModel(chain)
 model(rand(10))
 ```
 
-For an intro to Flux and automatic differentiation, see this [tutorial](https://fluxml.ai/tutorials/2020/09/15/deep-learning-flux.html)
+For an intro to Flux and automatic differentiation, see this [tutorial](https://fluxml.ai/tutorials/2020/09/15/deep-learning-flux.html).
 
 ## Customising Parameter Collection for a Model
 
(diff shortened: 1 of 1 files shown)


Gregliest:
Alright, I addressed all your comments. Looks like doctests are failing in basics.md, do you want me to have a look at that as well?

ToucheSir (dismissed):
Sorry about the delay. If you're still interested, I think fixing basics.md in a separate PR would be great (to my knowledge, it should be a simple matter of formatting). This is good to go though.

ToucheSir (approved):

logankilpatrick:
Can we bors this @ToucheSir ?

DhairyaLGandhi:
I'm wondering if this makes sense as a tutorial in the model zoo. Showing not just the structure based model, but also one that doesn't involve any structure and is very functional. We already have a similar example in the vae model and the cppn etc, but I wonder if an even more basic one is warranted.

ToucheSir:
@logankilpatrick Bors is failing to integrate because of failures on the `staging` branch and possibly `master`, see discussion in https://github.com/FluxML/Flux.jl/pull/1797.

Issue #737, opened 2019-04-15
saving ADAM optimizer is broken [@save] [BSON]


aldopareja:
this code should probably run:

```julia
using Flux
using BSON: @save
opt = ADAM(0.1)
@save "some_optim.bson" opt
```

devmotion:
I think the underlying problem is
```julia
using BSON: @save
dict = IdDict()
@save "test.bson" dict
```
and
```julia
using BSON: @save
dict = IdDict()
a = rand(5)
dict[a] = 5
@save "test.bson" dict
```

devmotion:
BTW, according to the documentation it should work: https://fluxml.ai/Flux.jl/stable/saving/

DilumAluthge:
The underlying cause of this issue is https://github.com/MikeInnes/BSON.jl/issues/43

bhvieira:
Anything on this? I just fell for this and lost some models I was checkpointing

DilumAluthge:
You could try using JLD2 instead of BSON.

We are still waiting for https://github.com/JuliaIO/BSON.jl/issues/43 to be fixed.

bhvieira:
But JLD and JLD2, as far as I can understand, can't save closures or functions in general.

bhvieira:
But I might try that, save the model with BSON and the optim with JLD

DilumAluthge:
I'm not sure about JLD. But JLD2 cannot save closures.

DilumAluthge:
> But I might try that, save the model with BSON and the optim with JLD

Yeah that might work.

bhvieira:
> I'm not sure about JLD. But JLD2 cannot save closures.

JLD can't either, trust me, just lost 2 days of work because of that

CarloLucibello:
This appears to be fixed by @dhairyagandhi96  in https://github.com/JuliaIO/BSON.jl/pull/70

NightMachinary:
Can the docs explain how saving IdDicts works? Don't they rely on the object ID of their keys? How can we know if these IDs are preserved when loading the parameters? (Does `BSON` do some magic to use the same IDs, or is this somehow guaranteed for any Julian serialization method?)

ToucheSir:
IIUC this was just about being able to save structs with `IdDicts` at all without errors, rather than being able to make use of them later. If you want to do the latter, give the following a try:
```julia
using Flux

# change opt type and IdDict field name for whatever you're using
function extract_opt_state(opt::ADAM, model)
[...]

PR #1762, opened 2021-11-06
Doc update (saving.md): removed outdated info; Typo fix.

NightMachinary:


ToucheSir (approved):

DhairyaLGandhi (changes requested):
The statement about IdDicts seems to be an issue with BSON rather than anything so I would retain the blurb about saving optimiser state. Would you mind holding on to the code snippet?

CarloLucibello:
I think the idea here is to not give false claims as "You can even store optimizer state alongside the model, to resume training exactly where you left off.", since we don't have the infrastructure to resume training. 
We need something like https://github.com/FluxML/Flux.jl/issues/737#issuecomment-962294707 but in the meantime we should avoid giving false information

docs/src/saving.md
```diff
@@ -118,10 +118,4 @@ revert to an older copy of the model if it starts to overfit.
...
 ```
 
-You can even store optimiser state alongside the model, to resume training
-exactly where you left off.
-
-```julia
-opt = ADAM()
-@save "model-$(now()).bson" model opt
-```
+Note that to resume a model's training, you might need to restore other stateful parts of your training loop. Possible examples are stateful optimizers (which usually utilize an `IdDict` to store their state, which is not automatically handled by `BSON`), and the randomness used to partition the original data into the training and validation sets.
```
CarloLucibello, line 121:
```suggestion
Note that to resume a model's training, you might need to restore other stateful parts of your training loop. Possible examples are stateful optimizers (which usually utilize an `IdDict` to store their state), and the randomness used to partition the original data into the training and validation sets.
```
I don't think BSON is the problem here, seems more on the flux side

NightMachinary: Do you think it's impossible to serialize `IdDict`s? I suggested the following strategy but @ToucheSir said `objectid` was not guaranteed to return unique IDs.

> -   Check if all the keys of the `IdDict` are present in the objects we are going to save to the BSON
> -   Put some ID tag on these objects (possibly their `objectid`)
> -   Store these IDs for keys of the `IdDict`
> -   When loading an `IdDict`, we can construct a new `IdDict` and populate it from the newly loaded objects

If `IdDict`s are fundamentally hard to serialize, perhaps Flux should switch to another dictionary type that is designed to be serializable?

---

Anyhow, the reason I added the `BSON` blurb was because `BSON` can handle `IdDict`s if a custom hack is employed. I.e., I wanted to emphasize that optimizers can be saved with some manual code, but they are not automagically "handled." Can you perhaps rephrase this better? Perhaps we can put a link to the workaround in the docs?

I don't think most new users will know what an `IdDict` is, or if `BSON` supports it or not. As a matter of fact, seeing the issue/PR history of `BSON`, I thought `BSON` did support `IdDict`s.

Another thing that comes to mind is `BSON`'s support for closures. We should mention this, as it is quite confusing, and not usually supported. (E.g., Python's FastAI disavows lambdas in its models exactly for this reason.) `BSON` itself is very sparsely documented; What are the limitations? E.g., when a closure is serialized, are its captured variables also serialized? Will this set captured globals in the global scope?
```julia
a = 0
f=()-> global a += 1
```

Perhaps our best approach is to add a limitations section to `BSON`'s readme, then link to it from the Flux docs.

ToucheSir: > If IdDicts are fundamentally hard to serialize, perhaps Flux should switch to another dictionary type that is designed to be serializable?

We are planning to switch away from the dict of params approach entirely, see https://github.com/FluxML/Flux.jl/pull/1481. This can't happen overnight, as a lot of downstream code relies on implicit params. In fact, it would be the biggest breaking change since Tracker was swapped out for Zygote ~2 years ago.

ToucheSir: Ref. https://github.com/FluxML/Flux.jl/issues/737#issuecomment-991942256, I was wrong about the inability to save `IdDict`s in optimizer state. I think it would be best to revert the changes in this section entirely in favour of adding a note after https://github.com/FluxML/Flux.jl/blob/master/docs/src/saving.md?plain=1#L127. That note could a) warn that models and optimizers must be saved together to have the latter work when restoring, and b) explain why (`BSON.@save filename xs...` can handle shared references/values in `xs`).

NightMachinary: @ToucheSir The other section of the PR is just a typo; I can close this PR, and you can submit a new one with the details about `BSON` and the typo fixed. BTW, adding this paragraph in that new PR is still helpful IMO:
```
Note that to resume a model's training, you might need to restore other stateful parts of your training loop. Possible examples are stateful optimizers (which usually utilize an `IdDict` to store their state), and the randomness used to partition the original data into the training and validation sets.
```

Or we can merge this PR, and you can re-add the section on BSON, with more details this time.

ToucheSir: The easiest way to go about things would be to restore the deleted code block below. That and this paragraph should be enough to merge.

NightMachinary: @ToucheSir  I pushed an update. Does it resolve all the issues?

docs/src/saving.md
```diff
@@ -118,10 +118,13 @@ revert to an older copy of the model if it starts to overfit.
 @save "model-$(now()).bson" model loss = testloss()
 ```
 
-You can even store optimiser state alongside the model, to resume training
-exactly where you left off.
+Note that to resume a model's training, you might need to restore other stateful parts of your training loop. Possible examples are stateful optimizers (which usually utilize an `IdDict` to store their state), and the randomness used to partition the original data into the training and validation sets.
+
+You can store the optimiser state alongside the model, to resume training
+exactly where you left off; BSON is smart enough to cache values and insert links when saving, but only if it knows everything to be saved up front. (See [here](https://github.com/JuliaIO/BSON.jl/blob/3b4a2cebda0afae11aab310f0a4d12b6a5234160/src/write.jl#L71).) So models and optimizers must be saved together to have the latter work when restoring.
```
ToucheSir, line 124:
```suggestion
exactly where you left off. BSON is smart enough to [cache values](https://github.com/JuliaIO/BSON.jl/blob/v0.3.4/src/write.jl#L71) and insert links when saving, but only only if it knows everything to be saved up front. Thus models and optimizers must be saved together to have the latter work after restoring.
```
```

ToucheSir (approved):
bors r+

PR #1763, opened 2021-11-08
Doc update (recurrence.md): fixed incorrect output dimensions, clarified batching.

NightMachinary:


docs/src/models/recurrence.md
```diff
@@ -175,9 +176,38 @@ x = [rand(Float32, 2, 4) for i = 1:3]
 y = [rand(Float32, 1, 4) for i = 1:3]
 ```
 
-That would mean that we have 4 sentences (or samples), each with 2 features (let's say a very small embedding!) and each with a length of 3 (3 words per sentence). Computing `m(batch[1])`, would still represent `x1 -> y1` in our diagram and returns the first word output, but now for each of the 4 independent sentences (second dimension of the input matrix).
+That would mean that we have 4 sentences (or samples), each with 2 features (let's say a very small embedding!) and each with a length of 3 (3 words per sentence). Computing `m(batch[1])`, would still represent `x1 -> y1` in our diagram and returns the first word output, but now for each of the 4 independent sentences (second dimension of the input matrix). We do not need to use `Flux.reset!(m)` here; Each sentence in the batch will output in its own "column", and the outputs of the different sentences won't mix.
```
ToucheSir, line 179:
```suggestion
That would mean that we have 4 sentences (or samples), each with 2 features (let's say a very small embedding!) and each with a length of 3 (3 words per sentence). Computing `m(batch[1])`, would still represent `x1 -> y1` in our diagram and returns the first word output, but now for each of the 4 independent sentences (second dimension of the input matrix). We do not need to use `Flux.reset!(m)` here; each sentence in the batch will output in its own "column", and the outputs of the different sentences won't mix. 
```

docs/src/models/recurrence.md
```diff
@@ -175,9 +176,38 @@ x = [rand(Float32, 2, 4) for i = 1:3]
 y = [rand(Float32, 1, 4) for i = 1:3]
 ```
 
-That would mean that we have 4 sentences (or samples), each with 2 features (let's say a very small embedding!) and each with a length of 3 (3 words per sentence). Computing `m(batch[1])`, would still represent `x1 -> y1` in our diagram and returns the first word output, but now for each of the 4 independent sentences (second dimension of the input matrix).
+That would mean that we have 4 sentences (or samples), each with 2 features (let's say a very small embedding!) and each with a length of 3 (3 words per sentence). Computing `m(batch[1])`, would still represent `x1 -> y1` in our diagram and returns the first word output, but now for each of the 4 independent sentences (second dimension of the input matrix). We do not need to use `Flux.reset!(m)` here; Each sentence in the batch will output in its own "column", and the outputs of the different sentences won't mix. 
+
+```julia
+#: This part is unchanged from our initial definiton; The batching comes for "free" from the way Julia does broadcasting and the rules of matrix multiplication.
```
ToucheSir, line 182: I would move these explanatory comments out to the markdown for consistency.

docs/src/models/recurrence.md
```diff
@@ -175,9 +176,38 @@ x = [rand(Float32, 2, 4) for i = 1:3]
...
+
+Both the hidden state and the output will now have a last dimension for batching; `h[:, n]` would be the hidden state of the nth sentence in the batch.
+
+```julia
+julia> size(h) == size(y) == (output_size, batch_size)
+true
+```
 
-In many situations, such as when dealing with a language model, each batch typically contains independent sentences, so we cannot handle the model as if each batch was the direct continuation of the previous one. To handle such situation, we need to reset the state of the model between each batch, which can be conveniently performed within the loss function:
+In many situations, such as when dealing with a language model, the sentences in each batch typically are independent (e.g., the last item of the first sentence of the first batch is independent from the first item of the first sentence of the second batch), so we cannot handle the model as if each batch was the direct continuation of the previous one. To handle such situations, we need to reset the state of the model between each batch, which can be conveniently performed within the loss function:
```
ToucheSir, line 210:
```suggestion
In many situations, such as when dealing with a language model, the sentences in each batch are independent (i.e. the last item of the first sentence of the first batch is independent from the first item of the first sentence of the second batch), so we cannot handle the model as if each batch was the direct continuation of the previous one. To handle such situations, we need to reset the state of the model between each batch, which can be conveniently performed within the loss function:
```

ToucheSir (changes requested):
Thanks! Just some small nits.

NightMachinary:
@ToucheSir I applied all the changes and squashed everything.

PS: I just found out about the convention around capitalization after semicolons. :) That's even weirder than the convention of putting punctuation inside quotes.

ToucheSir (approved):
Thanks, these are some nice improvements and I believe also fix an outstanding issue with the RNN docs :)

Issue #1696, opened 2021-08-22
Add better docs for the LSTM function


logankilpatrick:
Along the lines as this question: https://stackoverflow.com/questions/68885126/how-to-make-use-lstms-in-flux-jl it would be great to update the docs to specific what the two input variables actually relate to when creating an LSTM. Same thing goes for RNN and GRU. IMO, the docs for those should have enough substance to stand on their own.

logankilpatrick:
Thank you @mcognetta !

PR #1772, opened 2021-11-21
Expand RNN/LSTM/GRU docs

mcognetta:
This PR adds expanded documentation to the RNN/LSTM/GRU/GRUv3 docs, resolving #1696.

It addresses the `in` and `out` parameter meanings and adds a warning about a common gotcha (not calling reset when batch sizes are changed).

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -251,30 +251,7 @@ julia> l(rand(Float32, 3, 10)) |> size # batch size of 10
 (5, 10)
 ```
 
-The following is a demonstration of when failing to call `reset!` between batch size changes can cause erroneous outputs.
-
-```julia
-julia> l = LSTM(3, 5)
-Recur(
-  LSTMCell(3, 5),                       # 190 parameters
-)         # Total: 5 trainable arrays, 190 parameters,
-          # plus 2 non-trainable, 10 parameters, summarysize 1.062 KiB.
-
-julia> l.state .|> size 
-((5, 1), (5, 1))
-
-julia> l(rand(Float32, 3)) |> size
-(5,)
-
-julia> l(rand(Float32, 3, 10)) |> size # batch size of 10
-(5, 10)
-
-julia> l.state .|> size # state shape has changed
-((5, 10), (5, 10))
-
-julia> l(rand(Float32, 3)) |> size # erroneously outputs a length 5*10 = 50 vector.
-(50,)
-```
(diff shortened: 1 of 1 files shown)


ToucheSir (changes requested):
This is great! My only significant request would be to switch over the output array summaries to `|> size` like you have for `[recur].state`.  That would save some vertical space and keep consistency with other layers (e.g. https://github.com/FluxML/Flux.jl/blob/20f9535df4447784747f62f65c7ba5db569481e6/src/layers/conv.jl#L79-L92).

mcognetta:
Thanks for the comments, I condensed the examples as you suggested.

ToucheSir:
Doctests appear to be failing because the previous line comments of `# A demonstration of not using `reset!` when the batch size changes.` are interpreted as expected outputs. Can you shorten them a bit so that they can become end of line comments? Otherwise everything looks good to me.

mcognetta:
I split it into two code-blocks. The first one is a `jldoctest` and shows the expected usages, the second is just a `julia` block and shows the gotcha behavior regarding batch sizes. I figured it was not a good idea to make the second one a doctest block because that behavior is "bad" and not something we want to include in tests.

ToucheSir:
Apologies for yet more revisions, but WDYT about extracting the "bad" behaviour as a shared block under the `Recur` docstring? It seems to me that the only difference between the various RNN layers is the actual type name, so deduplicating that and inserting backrefs (with Documenter.jl warning style if it helps) in each layer's docstring would reduce the repetition.

mcognetta:
> Apologies for yet more revisions, but WDYT about extracting the "bad" behaviour as a shared block under the `Recur` docstring? It seems to me that the only difference between the various RNN layers is the actual type name, so deduplicating that and inserting backrefs (with Documenter.jl warning style if it helps) in each layer's docstring would reduce the repetition.

I changed all but the RNN docstring to have a reference to RNN in a warning about the batch size changing. The RNN docstring has the original example of the behavior.

I wasn't able to check that the docs rendered correctly though, do you mind taking a look?

ToucheSir:
Let me just check if docs previews can be made to work here. If not, I'll pull locally and confirm (if or you can, that would be swell).

ToucheSir (approved):
Ok, double-checked the syntax against https://github.com/FluxML/Flux.jl/blob/master/docs/src/models/recurrence.md?plain=1#L111-L123 and everything looks fine. Thanks for the PR!

mcognetta:
I was able to build them locally (https://github.com/JuliaDocs/Documenter.jl/issues/1413#issuecomment-687733990) and it looked good.

Thanks for the comments.

PR #1776, opened 2021-11-23
Use conjugates in optimizers to better learn on complex-valued inputs

staticfloat:
When weights are complex, the deltas to them will also be complex.  In
all optimizers that need a second-order estimate of gradient statistics,
we generally want to use the `x * conj(x)` pattern, rather than `x^2`.

We can see the effect this has on ADAM with the following test:

```julia
begin
    # This model will learn `W = I` and `bias = 0`
    complex_init(dims...) = Flux.glorot_uniform(dims...) .+ 1im .* Flux.glorot_uniform(dims...)
    model = Chain(
        Dense(4, 4, tanh; init=complex_init),
        Dense(4, 16, tanh; init=complex_init),
        Dense(16, 4, tanh; init=complex_init),
        Dense(4, 4, tanh; init=complex_init),
    )

    # Loss function; note we don't need the `abs()` if we update `Flux.Losses.mse()` as below
    function loss(x)
        return abs.(Flux.Losses.mse(model(x), x))
    end

    # Keep track of loss from epoch to epoch
    losses = Float64[]
    dataset = [(randn(ComplexF32, 4, 10),)]
    params = Flux.params(model)
    opt = Flux.Optimise.ADAM(0.001)
    for epoch_idx in 1:10000
        Flux.train!(loss, params, dataset, opt)
        epoch_loss = loss(dataset[1][1])
        push!(losses, epoch_loss)
        if epoch_idx % 100 == 0
            @info("epoch done", epoch_idx, epoch_loss)
        end
    end

    # Plot the loss
    fig = Figure()
    meta_ax = Axis(fig[1,1])
    lines!(meta_ax, log.(losses); label="Training loss")
    fig[1,2] = Legend(fig, meta_ax, "Learning Stats")
    fig
end
```

The training loss before the fix looks like this:

without_workaround

Whereas after both of these commits, it looks like this:

with_workaround

Note that while the absolute value of the loss is actually comparable in this simple example, the loss landscape is significantly more chaotic.  With a higher learning rate, the "fixed" version is able to learn much faster: 

download-1

Whereas the unfixed version simply diverges:

download-2

### DIFF
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -44,7 +44,8 @@ julia> Flux.mse(y_model, y_true)
 """
 function mse(ŷ, y; agg = mean)
   _check_sizes(ŷ, y)
-  agg((ŷ .- y) .^ 2)
+  error = ŷ .- y
+  real(agg(error .* conj(error)))
 end
 
 """
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -141,7 +141,7 @@ RMSProp(η = 0.001, ρ = 0.9) = RMSProp(η, ρ, IdDict())
 function apply!(o::RMSProp, x, Δ)
   η, ρ = o.eta, o.rho
   acc = get!(() -> zero(x), o.acc, x)::typeof(x)
-  @. acc = ρ * acc + (1 - ρ) * Δ^2
+  @. acc = ρ * acc + (1 - ρ) * Δ * conj(Δ)
   @. Δ *= η / (√acc + ϵ)
 end
 
@@ -179,7 +179,7 @@ function apply!(o::ADAM, x, Δ)
   end :: Tuple{typeof(x),typeof(x),Vector{Float64}}
 
   @. mt = β[1] * mt + (1 - β[1]) * Δ
-  @. vt = β[2] * vt + (1 - β[2]) * Δ^2
+  @. vt = β[2] * vt + (1 - β[2]) * Δ * conj(Δ)
   @. Δ =  mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + ϵ) * η
   βp .= βp .* β
 
@@ -221,7 +221,7 @@ function apply!(o::RADAM, x, Δ)
   end :: Tuple{typeof(x),typeof(x),Vector{Float64},Ref{Int}}
 
   @. mt = β[1] * mt + (1 - β[1]) * Δ
-  @. vt = β[2] * vt + (1 - β[2]) * Δ^2
+  @. vt = β[2] * vt + (1 - β[2]) * Δ * conj(Δ)
   ρ = ρ∞ - 2t[] * βp[2] / (1 - βp[2])
   if ρ > 4
     r = sqrt((ρ-4)*(ρ-2)*ρ∞/((ρ∞-4)*(ρ∞-2)*ρ))
@@ -311,7 +311,7 @@ function apply!(o::OADAM, x, Δ)
   end :: Tuple{typeof(x),typeof(x),typeof(x),Vector{Float64}}
 
   @. mt = β[1] * mt + (1 - β[1]) * Δ
-  @. vt = β[2] * vt + (1 - β[2]) * Δ^2
+  @. vt = β[2] * vt + (1 - β[2]) * Δ * conj(Δ)
   @. Δ = -Δ_
   @. Δ_ = η * mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + ϵ)
   @. Δ += 2Δ_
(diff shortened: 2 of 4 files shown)


staticfloat:
X-ref: https://github.com/pytorch/pytorch/issues/59998

ToucheSir:
There was a discussion about exactly this some weeks ago, but for the life of me I can't remember where or on what platform. @darsnack do you recall this?

darsnack (dismissed):
Looks good to me, thanks for this.

@ToucheSir Perhaps [this Zulip stream](https://julialang.zulipchat.com/#narrow/stream/238249-machine-learning/topic/Complex-valued.20optimizers). But I recall a longer discussion linking to the PyTorch issue above. I feel like it was `#autodiff` or `#math-optimization` on Slack, but those messages are long gone.

darsnack:
Is there a good test we could add here?

DhairyaLGandhi:
This is great, thanks @staticfloat! A good test might be to train with complex valued weights and inputs and checking for smoothness. We might need to do a similar pass over the other optimisers as well.

src/optimise/optimisers.jl
```diff
@@ -141,7 +141,7 @@ RMSProp(η = 0.001, ρ = 0.9) = RMSProp(η, ρ, IdDict())
 function apply!(o::RMSProp, x, Δ)
   η, ρ = o.eta, o.rho
   acc = get!(() -> zero(x), o.acc, x)::typeof(x)
-  @. acc = ρ * acc + (1 - ρ) * Δ^2
+  @. acc = ρ * acc + (1 - ρ) * Δ * conj(Δ)
```
baggepinnen, line 144: Isn't `Δ * conj(Δ)` the same as `abs2(Δ)`?

baggepinnen: no, they are not the same, `abs2` produces a real value while `Δ * conj(Δ)` produces a complex type with zero imaginary part.

staticfloat: Yeah, I didn't look deeply into these enough to figure out it returning a real value would cause type instability or whatnot, so I just left it in this format.

staticfloat:
Added some tests, and confirmed that this makes a big difference when training actual models.

ToucheSir (approved):

DhairyaLGandhi (approved):
lgtm. We need to be careful to preserve the behaviour for complex models moving on and review other losses for the same treatment as `mse`. Not a huge fan of the increased complexity in `mse` but we definitely need to be correct first.

bors r+

src/losses/functions.jl
```diff
@@ -44,7 +44,8 @@ julia> Flux.mse(y_model, y_true)
 """
 function mse(ŷ, y; agg = mean)
   _check_sizes(ŷ, y)
-  agg((ŷ .- y) .^ 2)
+  error = ŷ .- y
+  real(agg(error .* conj(error)))
```
staticfloat, line 48: @DhairyaLGandhi We can simplify this by using `agg(abs2.(ŷ .- y))` instead, as `abs2()` (as noted elsewhere) always returns a `Real`.  Would you prefer that?

DhairyaLGandhi: Yeah I think that would be good.

mcabbott:
While the test passes, something is wrong with the ADADelta  case here. I think it's passing because the last digits decline a bit, but overall the loss is very close to constant. 

Graph here:  https://github.com/FluxML/Optimisers.jl/pull/47  but also applies to Flux:
```
julia> opt_ctor = ADADelta;

julia> LOG = [];

julia> for idx in 1:10
                grads = Flux.gradient(loss, params)
                push!(LOG, loss()) 
                last_loss = loss()
                Flux.update!(opt, params, grads)
            end

julia> LOG
10-element Vector{Any}:
 2.0000000993025404
 2.0000000537095355
 2.000000116407697
 2.000000066966267
 2.000000134152019
 2.000000081155827
 2.000000152564042
 2.000000096253135
 2.000000171664382
 2.0000001122397846
```

test/optimise.jl
```diff
@@ -190,3 +190,40 @@ end
...
+        # Deterministic training data is the best training data
+        x = ones(1, 1) + 1im*ones(1, 1)
+
+        # Manually implement `mse()` to allow demonstration of brokenness
+        # on older Flux builds that don't have a fixed `mse()`
+        return sum(abs2.(w * x .- conj(x)))
+    end
+
+    params = Flux.Params([w])
+    opt = opt_ctor(1e-2)
```
mcabbott, line 218: This uses the same parameter for all. But ADADelta's first parameter wants to be close to 1, not 0.

src/optimise/optimisers.jl
cossio: Isn't `x * conj(x)` slower (albeit slightly) than `abs2(x)`?

PR #1778, opened 2021-11-23
add centered version of RMSProp

ludvigk:
Added centered version of RMSProp, which exists in other frameworks ([pytorch](https://pytorch.org/docs/stable/generated/torch.optim.RMSprop.html), [TF](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/RMSprop)).

Not sure if it makes more sense to have a separate `CenteredRMSProp` (or `CRMSProp`) struct for the centered version.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


src/optimise/optimisers.jl
```diff
@@ -110,39 +110,51 @@ function apply!(o::Nesterov, x, Δ)
...
 - Momentum (`ρ`): Controls the acceleration of gradient descent in the
                   prominent direction, in effect dampening oscillations.
+- Centered (`centered`): Whether to use the centered version of RMSProp.
 
 # Examples
 ```julia
 opt = RMSProp()
 
-opt = RMSProp(0.002, 0.95)
+opt = RMSProp(0.002, 0.95, True)
```
DhairyaLGandhi, line 134:
```suggestion
opt = RMSProp(0.002, 0.95, true)
```

src/optimise/optimisers.jl
```diff
@@ -110,39 +110,51 @@ function apply!(o::Nesterov, x, Δ)
 end
 
 """
-    RMSProp(η = 0.001, ρ = 0.9)
+    RMSProp(η = 0.001, ρ = 0.9, centered = False)
```
DhairyaLGandhi, line 113:
```suggestion
    RMSProp(η = 0.001, ρ = 0.9, centered = false)
```

DhairyaLGandhi (commented):
Thanks for looking into this! I've left a couple minor thoughts.

ludvigk:
> Thanks for looking into this! I've left a couple minor thoughts.

Thanks, Python bring bad habits.

src/optimise/optimisers.jl
```diff
@@ -110,39 +110,51 @@ function apply!(o::Nesterov, x, Δ)
...
 """
-    RMSProp(η = 0.001, ρ = 0.9)
+    RMSProp(η = 0.001, ρ = 0.9, centered = false)
 
 Optimizer using the
 [RMSProp](https://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf)
 algorithm. Often a good choice for recurrent networks. Parameters other than learning rate
 generally don't need tuning.
 
+The [centerd version](https://arxiv.org/pdf/1308.0850v5.pdf) of RMSProp maintains a moving
```
ToucheSir, line 120:
```suggestion
The [centered version](https://arxiv.org/pdf/1308.0850v5.pdf) of RMSProp maintains a moving
```

src/optimise/optimisers.jl
```diff
@@ -110,39 +110,51 @@ function apply!(o::Nesterov, x, Δ)
...
-opt = RMSProp(0.002, 0.95)
+opt = RMSProp(0.002, 0.95, true)
 ```
 """
   eta::Float64
   rho::Float64
-  acc::IdDict
+  centered::Bool
+  state::IdDict
```
ToucheSir, line 141: Do we consider this an API change? Might have to add a deprecation warning + getproperty overload forwarding `acc` to `state`.

ludvigk: I assume we'll have to overload setproperty! also then, to avoid issues where the optimiser state is saved to properly resume training? Still, if it's saved in a .bson file, there will be problems. The easiest way to avoid the API change is perhaps to just define the centered version as a different optimiser. I like the idea of just having one optimiser with the _centered_ flag though.

Another option is to create the new RMSProp version under a different name, and use a deprication warning for RMSProp.

DhairyaLGandhi: Might just leave `acc` as it is as the symbol. That would resolve the deprecation. Would there be a problem?

ludvigk: Does that help? The IdDict doesn't contain the same information anymore, so I don't see that helping.

ToucheSir: It does contain the same information when `centered = false` though, which is exactly what we want for backwards compatibility.

My concern was that both renaming and adding fields seem to be SemVer breaks. The former is easier to address, but the latter would not play well with code that uses reflection. Serialization isn't as big of a concern because optimizer state is basically useless once serialized (IdDict references will no longer match up).

src/optimise/optimisers.jl
```diff
@@ -110,39 +110,51 @@ function apply!(o::Nesterov, x, Δ)
...
+  acc, Δ_ave = get!(o.state, x) do
+    (zero(x), zero(x))
+  end :: Tuple{typeof(x),typeof(x)}
+
   @. acc = ρ * acc + (1 - ρ) * Δ^2
-  @. Δ *= η / (√acc + ϵ)
+  if o.centered
+    @. Δ_ave = ρ * Δ_ave + (1 - ρ) * Δ
+  end
+  @. Δ *= η / (√(acc - Δ_ave^2) + ϵ)
```
DhairyaLGandhi, line 157:
```suggestion
  @. Δ *= η / (√(acc - Δ_ave * conj(Δ_ave)) + ϵ)
```

DhairyaLGandhi (commented):
Could move `centered` to a type parameter like `RMSProp{true/false}` to avoid adding a field to the optimiser.

src/optimise/optimisers.jl
```diff
@@ -110,39 +110,51 @@ function apply!(o::Nesterov, x, Δ)
 end
 
 """
-    RMSProp(η = 0.001, ρ = 0.9)
+    RMSProp(η = 0.001, ρ = 0.9, centered = false)
```
mcabbott, line 113: On master there is `RMSProp(η::Real, ρ::Real, ϵ::Real)`. 

Probably this option should be controlled by a keyword, matching  https://github.com/FluxML/Optimisers.jl/pull/51

mcabbott:
This need not hold up v0.13, since it adds a feature without breaking anything.

cossio:
This looks almost ready to merge? What's missing?

ludvigk:
@cossio
> @ludvigk are you planning to follow up on this?
> 
> If not, I'd be willing to open a new PR with this.

I forgot about this. Feel free to open a new PR, or I can update this one this weekend.

ludvigk:
I added the suggestions from here to the pull request. Changed "state" back to "acc", and merged with the new epsilon argument for RMSprop. Not sure what's left?

mcabbott:
I think these two signatures are a bit too do-what-I-mean.
```
RMSProp(η::Real, ρ::Real, centered::Bool, ϵ::Real)
 RMSProp(η::Real, ρ::Real, ϵ::Real)
```
I would prefer it to be only a keyword argument, as 4 is a lot of positional arguments, especially if their order isn't fixed. 

Ideally the keyword should match https://github.com/FluxML/Optimisers.jl/pull/51 . Strong feelings about whether the constructor should take the verb "centre" vs the adjective "centred"?

CarloLucibello:

PR #1781, opened 2021-11-24
Fix AlphaDropout implementation and add tests

ToucheSir:
AFAICT, the original implementation never behaved as expected even pre-Zygote. This was likely not caught because the original PR didn't come with tests, so this PR should remedy that.

Behaviour and outputs are adapted from the PyTorch and TF implementations. Some points of note:
1. We have to special-case `p = 0` to avoid propagating NaNs when calcuating `A` and `B`.
2. Likewise for `p = 0`. TF just returns the input, but I think the PyTorch approach of returning all zeros (+/- depending on the input sign) is more in line with `Dropout`.
3. `ifelse` is used instead of something like https://github.com/keras-team/keras/blob/v2.7.0/keras/layers/noise.py#L200. I think it better reflects the conditional nature of the operation and it was also faster in local benchmarking.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -101,17 +101,20 @@ mutable struct AlphaDropout{F}
   end
 end
 
-function (a::AlphaDropout)(x)
+function (a::AlphaDropout)(x::AbstractArray{T}) where T
   _isactive(a) || return x
-  λ = eltype(x)(1.0507009873554804934193349852946)
-  α = eltype(x)(1.6732632423543772848170429916717)
-  α1 = eltype(x)(-λ*α)
-  noise = randn(eltype(x), size(x))
-  x = @. x*(noise > (1 - a.p)) + α1 * (noise < (1 - a.p))
-  A = sqrt(a.p + a.p * (1 - a.p) * α1^2)
-  B = -A * α1 * (1 - a.p)
-  x = @. A * x + B
-  return x
+  p = a.p
+  iszero(p) && return x
+  isone(p) && return sign.(x) .* T(0)
+
+  λ = T(1.0507009873554804934193349852946)
+  α = T(1.6732632423543772848170429916717)
+  α1 = T(-λ * α)
+  A = inv(sqrt((1 - p) * (1 + p * α1^2)))
+  B = -A * α1 * p
+
+  noise = rand!(similar(x))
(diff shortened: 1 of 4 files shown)


ToucheSir:
Unexpected bonus: this is now 100% GPU compatible and leaves us with one less broken test :)

CarloLucibello:
looks good!

CarloLucibello (dismissed):

darsnack:
Looks good, I think this deserves an entry in NEWS.md. Especially the GPU bit.

CarloLucibello (approved):

src/layers/normalise.jl
```diff
@@ -101,17 +101,18 @@ mutable struct AlphaDropout{F}
...
-  α1 = eltype(x)(-λ*α)
-  noise = randn(eltype(x), size(x))
-  x = @. x*(noise > (1 - a.p)) + α1 * (noise < (1 - a.p))
-  A = sqrt(a.p + a.p * (1 - a.p) * α1^2)
-  B = -A * α1 * (1 - a.p)
-  x = @. A * x + B
-  return x
+  p = a.p
+  iszero(p) && return x
+  isone(p) && return sign.(x) .* T(0)
```
DhairyaLGandhi, line 108: Is this correct? Are we intentionally creating `-0.0` values here.

src/layers/normalise.jl
```diff
@@ -101,17 +101,18 @@ mutable struct AlphaDropout{F}
...
+
+  α′ = T(-1.7580993408473766) # selu(-Inf) == -λα
```
DhairyaLGandhi, line 110: We can use `oftype` here

src/layers/normalise.jl
ToucheSir: Yes, see point 2 in https://github.com/FluxML/Flux.jl/pull/1781#issue-1062861991.

src/layers/normalise.jl
ToucheSir: `oftype` requires a value of type `T`, but the only way to get that in this function would be to extract an element of `x`would fail if `x` is empty and might also trigger the getindex adjoint unnecessarily). Since this function should be specializing on eltype anyhow, it makes sense to take advantage of that for casting.

PR #1783, opened 2021-11-25
Doc update (training.md): fix DataLoader example in Training section

eliascarv:
Change from:
```julia
X = rand(28, 28, 60000)
Y = rand(0:9, 60000)
data = DataLoader(X, Y, batchsize=128)
```
To:
```julia
X = rand(28, 28, 60000)
Y = rand(0:9, 60000)
data = DataLoader((X, Y), batchsize=128)
```


ToucheSir (approved):
Thanks!

PR #1785, opened 2021-11-26
Update train.jl to add a more detailed `train!` docstring

logankilpatrick:
Right now, the `train!` function docstring does not tell you what you need to do with respect to params. I tried to add that and moved a sentence from somewhere else in the docs about what the train function does to the top of the doc string.

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -91,8 +91,7 @@ respect to `params` through backpropagation and call the optimizer `opt`.
         
 If `d` is a tuple of arguments to `loss` call `loss(d...)`, else call `loss(d)`.
         
-To get the trainable parameters, generally you pass the model to the [`Flux.params`](@ref)
-function (the just the layers you want to train) like `train!(loss, params(model), ...)`.
+To pass trainable parameters, call [`Flux.params`](@ref) with your model or just the layers you want to train, like `train!(loss, params(model), ...)`.
 
 A callback is given with the keyword argument `cb`. For example, this will print
 "training" every 10 seconds (using [`Flux.throttle`](@ref)):
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -83,19 +83,21 @@ batchmemaybe(x::Tuple) = x
 """
     train!(loss, params, data, opt; cb)
         
-`train!` uses a loss function (`loss`) and training data (`data`) to improve the 
-parameters (`params`) of your model based on a pluggable optimiser (`opt`).
+`train!` uses a [Loss-Functions](@ref) (`loss`) and training [Datasets](@ref) (`data`) to improve the 
+[Model-parameters](@ref) (`params`) of your model based on a pluggable [Optimisers](@ref) (`opt`).
         
 For each datapoint `d` in `data`, compute the gradient of  `loss` with
 respect to `params` through backpropagation and call the optimizer `opt`.
 If `d` is a tuple of arguments to `loss` call `loss(d...)`, else call `loss(d)`.
         
-To pass trainable parameters, call [`Flux.params`](@ref) with your model or just the layers you want to train, like `train!(loss, params(model), ...)`.
+To pass trainable parameters, call [`Flux.params`](@ref) with your model or just the 
+layers you want to train, like `train!(loss, params(model), ...)` or `train!(loss, params(model[1:end-2), ...)` respectively.
 
-A callback is given with the keyword argument `cb`. For example, this will print
-"training" every 10 seconds (using [`Flux.throttle`](@ref)):
-    train!(loss, params, data, opt, cb = throttle(() -> println("training"), 10))
-
+[Callbacks]@ref) are given with the keyword argument `cb`. For example, this will print "training" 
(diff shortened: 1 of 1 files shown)


src/optimise/train.jl
```diff
@@ -82,11 +82,17 @@ batchmemaybe(x::Tuple) = x
...
+parameters (`params`) of your model based on a pluggable optimiser (`opt`).
+        
 For each datapoint `d` in `data`, compute the gradient of  `loss` with
 respect to `params` through backpropagation and call the optimizer `opt`.
-
+        
 If `d` is a tuple of arguments to `loss` call `loss(d...)`, else call `loss(d)`.
+        
+To get the trainable parameters, generally you pass the model to the [`Flux.params`](@ref)
+function (the just the layers you want to train) like `train!(loss, params(model), ...)`.
```
ToucheSir, line 95:
```suggestion
To pass trainable parameters call [`Flux.params`](@ref) with your model or just the layers you want to train, like `train!(loss, params(model), ...)`.
```
Once https://github.com/FluxML/Flux.jl/pull/1786 lands, I think it makes sense to lean on that doc here.

src/optimise/train.jl
```diff
@@ -82,11 +82,16 @@ batchmemaybe(x::Tuple) = x
...
     train!(loss, params, data, opt; cb)
-
+        
+`train!` uses a loss function (`loss`) and training data (`data`) to improve the 
+parameters (`params`) of your model based on a pluggable optimiser (`opt`).
+        
 For each datapoint `d` in `data`, compute the gradient of  `loss` with
 respect to `params` through backpropagation and call the optimizer `opt`.
-
+
```
ToucheSir, line 91:
```suggestion

```

src/optimise/train.jl
```diff
@@ -82,11 +82,16 @@ batchmemaybe(x::Tuple) = x
 
 """
     train!(loss, params, data, opt; cb)
-
+        
+`train!` uses a loss function (`loss`) and training data (`data`) to improve the 
+parameters (`params`) of your model based on a pluggable optimiser (`opt`).
+        
 For each datapoint `d` in `data`, compute the gradient of  `loss` with
```
ToucheSir, line 89:
```suggestion
For each datapoint `d` in the `data` iterator, compute the gradient of  `loss` with
```
I don't love this phrasing, but barring a larger docstring refactor it seems the easiest way to point out what interfaces `data` should support.

src/optimise/train.jl
```diff
@@ -82,11 +82,16 @@ batchmemaybe(x::Tuple) = x
 
 """
     train!(loss, params, data, opt; cb)
-
+        
+`train!` uses a loss function (`loss`) and training data (`data`) to improve the 
+parameters (`params`) of your model based on a pluggable optimiser (`opt`).
```
ToucheSir, line 87: Is there any way for "optimiser" to link to https://github.com/FluxML/Flux.jl/blob/master/docs/src/training/optimisers.md?

src/optimise/train.jl
logankilpatrick: Do we need to say "iterator", isn't it assumed that any time we do `for something in something else`, `something else` is an iterator / iterable?

src/optimise/train.jl
logankilpatrick: yes, then we should also link the loss function page as well

logankilpatrick: Should be simple per: https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Cross-Referencing, I can add both if desired

src/optimise/train.jl
ToucheSir: I'm not sure, but my inclination would be that "in" is sufficiently general verbiage that new users wouldn't pattern match it to iteration. We can always leave this out and revisit if people aren't getting it.

ToucheSir (approved):

src/optimise/train.jl
ToucheSir: That would be great :)

logankilpatrick:
Preview the PR here: https://fluxml.ai/Flux.jl/previews/PR1785/

ToucheSir:
First 2 links to loss functions and datasets aren't resolving :confused:

logankilpatrick:
Yeah I don't get why : ( I will open a issue for documenter.jl, I am sure I messed it up

logankilpatrick:
Same here @ToucheSir I am going to try and tackle all of these doc ref issues in a larger PR since I will need to add ID's and such

src/optimise/train.jl
```diff
@@ -82,16 +82,21 @@ batchmemaybe(x::Tuple) = x
 
 """
     train!(loss, params, data, opt; cb)
-
+        
+`train!` uses a [Loss Functions](@ref) (`loss`) and training [Datasets](@ref) (`data`) to improve the
```
ToucheSir, line 86:
```suggestion
`train!` uses a `loss` function and training `data` to improve the 
```
Let's remove the dangling links for now so readers aren't confused by them.

ToucheSir (approved):

PR #1786, opened 2021-11-26
Add docstring for `params`

logankilpatrick:
`params` is used all over the model zoo and tutorials in Flux. There should be a docstring or we should not publicly use it. Not sure if the proposed docstring gets the job done 100% but this seems like a step in the right direction.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -55,6 +55,8 @@ Given a model or specific layers from a model, create a `Params` object pointing
 
 This can be used with [`gradient`](@ref), or as input to the [`Flux.train!`](@ref Flux.train!) function.
 
+The behaviour of `params` on custom types can be customized using [`Functor.@functor`](@ref) or [`Flux.trainable`](@ref).
+
 # Examples
 ```jldoctest
 julia> params(Chain(Dense(ones(2,3))), softmax)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -54,7 +54,7 @@ end
 
 Given a model or specific layers from a model, create a `Params` object pointing to its trainable parameters.
 
-This can be used with [`gradient`](@ref), or as input to the [`Flux.train!`](@ref Flux.train!) function.
+This can be used with the `gradient` function, see [Taking Gradients](@ref), or as input to the [`Flux.train!`](@ref Flux.train!) function.
 
 The behaviour of `params` on custom types can be customized using [`Functor.@functor`](@ref) or [`Flux.trainable`](@ref).
 
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -51,8 +51,18 @@ end
 """
     params(model)
 
-Given a model or specific layers from a model, return the trainable parameters
-such that they can be used as input to the `train!` function.
+Given a model or specific layers from a model, create a `Params` object pointing to its trainable parameters.
+
+This can be used with [`gradient`](@ref), or as input to the [`Flux.train!`](@ref Flux.train!) function.
+
+# Examples
+```jldoctest
+julia> params(Chain(Dense(ones(2,3))), softmax)
+Params([[1.0 1.0 1.0; 1.0 1.0 1.0], [0.0, 0.0]])
+
+julia> params(BatchNorm(2, relu))
+Params([Float32[0.0, 0.0], Float32[1.0, 1.0]])
+```
 """
 function params(m...)
   ps = Params()
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -60,11 +60,23 @@ The behaviour of `params` on custom types can be customized using [`Functor.@fun
 
 # Examples
 ```jldoctest
-julia> params(Chain(Dense(ones(2,3))), softmax)
+julia> params(Chain(Dense(ones(2,3)), softmax))  # unpacks Flux models
 Params([[1.0 1.0 1.0; 1.0 1.0 1.0], [0.0, 0.0]])
 
-julia> params(BatchNorm(2, relu))
+julia> bn = BatchNorm(2, relu)
+BatchNorm(2, relu)  # 4 parameters, plus 4 non-trainable
+
+julia> params(bn)  # only the trainable parameters
 Params([Float32[0.0, 0.0], Float32[1.0, 1.0]])
+
+julia> params([1, 2, 3], [4.0])  # one or more arrays of numbers
+Params([[1, 2, 3], [4.0]])
+
+julia> params([[1, 2, 3], [4.0]])  # unpacks array of arrays
+Params([[1, 2, 3], [4.0]])
+
+julia> params(1, [2 2], (alpha=[3,3,3], beta=Ref(4), gamma=sin))  # ignores scalars, unpacks NamedTuples
+Params([[2 2], [3, 3, 3]])
 ```
 """
 function params(m...)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -48,6 +48,12 @@ function params!(p::Params, x, seen = IdSet())
   end
 end
 
+"""
+    params(model)
+
+Given a model or specific layers from a model, return the trainable parameters
+such that they can be used as input to the `train!` function.
+"""
 function params(m...)
   ps = Params()
   params!(ps, m)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -50,6 +50,7 @@ end
 
 """
     params(model)
+    params(layers...)
 
 Given a model or specific layers from a model, create a `Params` object pointing to its trainable parameters.
 
--- a/docs/src/training/training.md
+++ b/docs/src/training/training.md
@@ -70,6 +70,10 @@ Such an object contains a reference to the model's parameters, not a copy, such 
 
 Handling all the parameters on a layer by layer basis is explained in the [Layer Helpers](../models/basics.md) section. Also, for freezing model parameters, see the [Advanced Usage Guide](../models/advanced.md).
 
+```@docs
+Flux.params
+```
+
 ## Datasets
 
 The `data` argument of `train!` provides a collection of data to train with (usually a set of inputs `x` and target outputs `y`). For example, here's a dummy dataset with only one data point:


src/functor.jl
```diff
@@ -48,6 +48,12 @@ function params!(p::Params, x, seen = IdSet())
   end
 end
 
+"""
+    params(model)
+
+Given a model or specific layers from a model, return the trainable parameters
+such that they can be used as input to the `train!` function.
```
mcabbott, line 55: Good idea to add something. May I suggest to add some brief examples, and some cross-links? I am not 100% sure the links will work, but hope that CI will tell us.
````suggestion
Given a model or specific layers from a model, create a `Params` object pointing to its trainable parameters.

This can be used with [`gradient`](@ref), or as input to the [`Flux.train!`](@ref Flux.train!) function.

# Examples
```jldoctest
julia> params(Chain(Dense(ones(2,3))), softmax)
Params([[1.0 1.0 1.0; 1.0 1.0 1.0], [0.0, 0.0]])

julia> params(BatchNorm(2, relu))
```
````

Comment (CarloLucibello) [reply src/functor.jl:55]: I suggest also adding something like "The behaviour of `params` on custom types can be customized using [`Functor.@functor`](@ref) or [`Flux.trainable`](@ref)"

Comment (logankilpatrick) [reply src/functor.jl]: Added

File: src/functor.jl [hunk 2]
```diff
@@ -48,6 +48,24 @@ function params!(p::Params, x, seen = IdSet())
...
 end
 
+"""
+    params(model)
+
+Given a model or specific layers from a model, create a `Params` object pointing to its trainable parameters.
+
+This can be used with [`gradient`](@ref), or as input to the [`Flux.train!`](@ref Flux.train!) function.
+
+The behaviour of `params` on custom types can be customized using [`Functor.@functor`](@ref) or [`Flux.trainable`](@ref).
```
ToucheSir, line 58:
```suggestion
The behaviour of `params` on custom types can be customized using [`Functors.@functor`](@ref) or [`Flux.trainable`](@ref).
```
Unfortunately `Functors.@functor` won't resolve because it's a cross-package link. Not sure what we can do about that, perhaps a manual link to the latest version of the Functors docs?

src/functor.jl
```diff
@@ -48,6 +48,24 @@ function params!(p::Params, x, seen = IdSet())
   end
 end
 
+"""
+    params(model)
```
ToucheSir, line 52:
```suggestion
    params(model)
    params(layers...)
```
Having just the one vararg signature might be simpler, but given 90% of people would be passing in a single model struct my worry would be that it would be confusing.

src/functor.jl
```diff
@@ -48,6 +48,24 @@ function params!(p::Params, x, seen = IdSet())
...
+This can be used with [`gradient`](@ref), or as input to the [`Flux.train!`](@ref Flux.train!) function.
+
+The behaviour of `params` on custom types can be customized using [`Functor.@functor`](@ref) or [`Flux.trainable`](@ref).
+
+# Examples
+```jldoctest
+julia> params(Chain(Dense(ones(2,3))), softmax)
+Params([[1.0 1.0 1.0; 1.0 1.0 1.0], [0.0, 0.0]])
+
+julia> params(BatchNorm(2, relu))
```
ToucheSir, line 65: We should have an example and/or note that most built-in container types _will_ be unpacked if passed into `params`. This has bitten people frequently when they try to use it like `Zygote.Params` and pass in a couple of arrays (`Flux.params(array1)` will unpack _every_ element of `array1` instead of including the array itself...)

ToucheSir (changes requested):

src/functor.jl
mcabbott: According to CI there are many such failures, although they do not cause it to fail overall here.

ToucheSir: Yes, I saw a couple searching for other examples of this. Short of a solution in Documenter.jl, I guess our best option is a relative link?

src/functor.jl
mcabbott: Or the other way around, it's `Params(array)` which is surprising? Maybe something like:
```
julia> params([1, 2, 3])
Params([[1, 2, 3]])

julia> params(1, [2 2], (alpha=[3,3,3], beta=Ref(4), gamma=sin))
Params([[2 2], [3, 3, 3]])

julia> keys(ans.params.dict)
KeySet for a IdDict{Any, Nothing} with 2 entries. Keys:
  [2 2]
  [3, 3, 3]
```

src/functor.jl
mcabbott: Since the page this wants to link to is here https://fluxml.ai/Flux.jl/stable/models/functors/#Functors.jl I think it ought to be possible at least to have an automatic link to the heading of the page. Maybe that's what you mean? Like https://raw.githubusercontent.com/JuliaLang/julia/master/doc/src/manual/arrays.md it seems you can write e.g.
`## [Functors.jl](@id man-h2-functors)` and then refer to that.

src/functor.jl
ToucheSir: Not that I'm aware of?
```julia-repl
julia> Flux.Params([1, [2 2], [3,3,3], Ref(4), sin])
Params([1, [2 2], [3, 3, 3], Base.RefValue{Int64}(4), sin])
```
The docstring for `Params` is reasonably clear too (explicitly asks for an array).

mcabbott: What I mean is by surprise is this difference, you can't give `Params` just one naked array:
```
julia> x = [1,2,3]; gradient(() -> sum(x), Params(x))[x]
ERROR: KeyError: key [1, 2, 3] not found

julia> x = [1,2,3]; gradient(() -> sum(x), params(x))[x]
3-element Fill{Int64}, with entries equal to 1
```
If that's not what you meant, what did you mean people get bitten by? 

Giving `params` an array of arrays instead is OK:
```
julia> x = [1,2,3]; gradient(() -> sum(x), params([x, x]))[x]
3-element Fill{Int64}, with entries equal to 1
```

src/functor.jl
logankilpatrick: this makes sense to me

src/functor.jl
logankilpatrick: https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Cross-Referencing makes it seem like we can do `[Functors.jl](@ref)` in the doc string and it will just work

logankilpatrick:
Also, which file or section should we add the `params` docs too? I am not quite sure where it fits in best.

src/functor.jl
ToucheSir: For the life of me I can't seem to find it the GH issue or discourse discussions for it, but I think it was an issue with not expecting arrays of non-numbers to be unpacked. Whereas `Params` documents this unpacking, `params` doesn't and certainly doesn't document that it unpacks _recursively_ when given a nested structure.

ToucheSir:
I'm not sure either. Co-locating it with https://fluxml.ai/Flux.jl/stable/models/basics/#Building-Simple-Models is one option, but a little out of the way. WDYT about creating a new page under the "Training Models" section for parameter handling? It could just be the docstring for now, but that would give us an opportunity to expand later.

ToucheSir:
Actually, https://fluxml.ai/Flux.jl/stable/training/training/#Model-parameters doesn't look too bad.

logankilpatrick:
> Actually, https://fluxml.ai/Flux.jl/stable/training/training/#Model-parameters doesn't look too bad.

Agreed

logankilpatrick:
Preview the PR here: https://fluxml.ai/Flux.jl/previews/PR1786/training/training/#Flux.params

src/functor.jl
ToucheSir: Worth a try. Also not in scope for this PR, but `Functors.@functor` needs to be rendered correctly in the docs: https://fluxml.ai/Functors.jl/stable/api/

ToucheSir: Also `trainable` isn't documented either. Looks like we'll need a follow-up PR :)

logankilpatrick: Many PR's to come : )

darsnack:
Is it worth mentioning that a vector of arrays can also be passed to `params`? This allows very manual, fine-grained control over collecting parameters.

ToucheSir:
I think that would fit with my note in https://github.com/FluxML/Flux.jl/pull/1786#discussion_r757742057 that any vector of arrays passed to `params` will be unpacked, even if they are nested within another `functor`-able value.

logankilpatrick:
@ToucheSir Perhaps we can merge this and have one sweeping PR to try and get all the ref issues fixed in the docs? There are a lot that should be resolved.

DhairyaLGandhi:
Since `params([1,23])` would resolve correctly and the remaining issue is with `Params` I believe we can safely merge this. `Params` explicitly requires an array and that is documented too, but if one feels that an extra note that says numerical arrays need to be wrapped in a "container" array is required, that's fine too but that is an orthogonal change to this docstring.

src/functor.jl
```diff
@@ -48,6 +48,25 @@ function params!(p::Params, x, seen = IdSet())
...
+Params([Float32[0.0, 0.0], Float32[1.0, 1.0]])
```
mcabbott, line 67: Maybe a few more examples, to address things people said:
```suggestion
julia> params(Chain(Dense(ones(2,3))), softmax)  # unpacks Flux models
Params([[1.0 1.0 1.0; 1.0 1.0 1.0], [0.0, 0.0]])

julia> params(BatchNorm(2, relu))  # only the trainable parameters
Params([Float32[0.0, 0.0], Float32[1.0, 1.0]])

julia> params([1, 2, 3], [4.0])  # one or more arrays of numbers
Params([[1, 2, 3], [4.0]])

julia> params([[1, 2, 3], [4.0]])  # unpacks array of arrays
Params([[1, 2, 3], [4.0]])

julia> params(1, [2 2], (alpha=[3,3,3], beta=Ref(4), gamma=sin))  # ignores scalars, unpacks NamedTuples
Params([[2 2], [3, 3, 3]])
```

mcabbott (approved):
Looks like a good step forwards. Figuring out Functors links will involve touching many things, can be another PR.

mcabbott (approved):
bors r+

But it may not listen, Brian you may need to approve?

ToucheSir (approved):
Done.

PR #1784, opened 2021-11-26
Fix link to train in the docs

logankilpatrick:
Right now, in the dev docs, the first link to train on this page works and the 2nd one doesn't so I copied the link to hopefully make the second one work.

PR #1788, opened 2021-11-26
Create a PR comment with docs preview link

logankilpatrick:
It looks like per @ToucheSir 's comment here: #1784, there is already a documentation preview workflow setup. This simply adds a comment to the PR with a link like we do on the JuliaLang.org website repo.

### DIFF
--- a/.github/workflows/pr_comment.yml
+++ b/.github/workflows/pr_comment.yml
@@ -10,5 +10,5 @@ jobs:
         if: github.event_name == 'pull_request' && github.repository == github.event.pull_request.head.repo.full_name # if this is a pull request build AND the pull request is NOT made from a fork
         uses: thollander/actions-comment-pull-request@71efef56b184328c7ef1f213577c3a90edaa4aff
         with:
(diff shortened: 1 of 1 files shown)


ToucheSir:
The problem is, AIUI, that docs previews are only generated for PRs opened by members (with the right set of permissions). So we couldn't tack this onto any old PR.

logankilpatrick:
Yeah, it would be nice if it let us do that, the current version in this PR only runs when the PR is made from a branch, not from forks. So it will work as is.

.github/workflows/pr_comment.yml
```diff
@@ -0,0 +1,14 @@
+name: pr_comment
+on:
+  pull_request:
+    types: [opened]
+jobs:
+  pr_comment:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create PR comment
+        if: github.event_name == 'pull_request' && github.repository == github.event.pull_request.head.repo.full_name # if this is a pull request build AND the pull request is NOT made from a fork
```
logankilpatrick, line 10: @ToucheSir

ToucheSir: That's what I get for not scrolling to the end, thanks!

ToucheSir (approved):

PR #1789, opened 2021-11-27
Update basics.md to include gradient docs

logankilpatrick:
Right now, when you search `gradient` in the docs, the actual docstring is not being displayed in the docs anywhere. I am not Sure if this is the right place for it but it seems close.

### DIFF
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -5,7 +5,7 @@
 Flux's core feature is taking gradients of Julia code. The `gradient` function takes another Julia function `f` and a set of arguments, and returns the gradient with respect to each argument. (It's a good idea to try pasting these examples in the Julia terminal.)
 
 ```@docs
-Flux.gradient
+Flux.Zygote.gradient
 ```
 
 ```jldoctest basics
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -4,6 +4,10 @@
 
 Flux's core feature is taking gradients of Julia code. The `gradient` function takes another Julia function `f` and a set of arguments, and returns the gradient with respect to each argument. (It's a good idea to try pasting these examples in the Julia terminal.)
 
+```@docs
+Flux.gradient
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Ah, this comes from Zygote but is exported as a function here. It still makes sense in my mind to have the doc string in the docs since it is an exported and highly visible function.

logankilpatrick:
It seems like we could do this per: https://stackoverflow.com/questions/70137119/how-to-include-the-docstring-for-a-function-from-another-package-in-my-julia-doc/70137992?noredirect=1#comment123984276_70137992. Is including the some Zygote docs something we would want @DhairyaLGandhi @ToucheSir ?

ToucheSir:
I think it would be a little weird to have a docstring talking all about Zygote in the Flux docs. Is it not possible to do a cross-project docs link for now?

logankilpatrick:
It is, this would simply expose the zygote docs where we need them to be added (for specific functions).

ToucheSir:
Another question is whether we should be privileging `gradient`. IME a lot of users actually want `withgradient` so they can collect a loss value as well without using a closure going one level lower with `pullback`.

mcabbott:
Note that this doesn't seem to display correctly.

This is at quite a basic section of the docs. Maybe the whole Zygote.gradient docstring is too much detail to interrupt people with at this point. The next few paragraphs slowly unpack much the same things as it contains.

The paragraph just above says "Flux's core feature is taking gradients of Julia code" which feels not quite true. Maybe we should change that, to say something more like that Flux relies on automatic differentiation, and at present this is provided by a package called Zygote (link), whose principal function  `gradient` is re-exported by Flux.

logankilpatrick:
Agreed. Yes, this is broken right now but I will try to find time in the next week to update it, we should definitely include the docs for those functions coming out of zygote.

mcabbott:
There is now a Zygote.jl page of the docs

PR #1790, opened 2021-11-27
Added the wrapper `Bidirectional` for RNN layers

dcecchini:
### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### Description

The implementation was based on the implementation of `BiLSTM_CNN_CRF_Model` present on [TextModels.jl](https://github.com/JuliaText/TextModels.jl)  but has a few differences:

- It leaves the `Bidirectional` usage closer to `Keras` as it allows the creation of, for example, `Bidirectional(Flux.LSTM, in, out)` inside a `Chain` declarations.
- It explicitly adds the `dims=1` parameter on the `reverse` function, which was needed in my experiments but more generic implementation may be possible (I was not able to use `flip` probably for the same reason).

I also updated the `Base.show` and `_big_show` functions so that it is printed in the same style of `Flux` giving the number of trainable parameters.  The implementation added for the `_big_show` could be added to  `src/show.jl` later.

This is my first PR here, I read the information on `CONTRIBUTING.md`, but please let me know if it is not following the correct standards. I really enjoy `Julia` even though I am still learning. 

Hope I can contribute to this great project `Flux`.

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -1,4 +1,3 @@
-
 gate(h, n) = (1:h) .+ h*(n-1)
 gate(x::AbstractVector, h, n) = @view x[gate(h,n)]
 gate(x::AbstractMatrix, h, n) = x[gate(h,n),:]
@@ -434,10 +433,7 @@ end
 
 """
     Bidirectional{A,B} 
-
 A wrapper layer that allows bidirectional recurrent layers. 
-
-
 # Examples
 ```jldoctest
 julia> BLSTM = Bidirectional(LSTM, 3, 5)
@@ -450,11 +446,8 @@ Bidirectional(
   ),
 )         # Total: 10 trainable arrays, 380 parameters,
           # plus 4 non-trainable, 20 parameters, summarysize 2.141 KiB.
-
 julia> BLSTM(rand(Float32, 3)) |> size
-
 (10,)
-
 julia> model = Chain(Embedding(10000, 200), Bidirectional(LSTM, 200, 128), Dense(256, 5), softmax)
 Chain(
   Embedding(10000, 200),                # 2_000_000 parameters
@@ -481,25 +474,27 @@ end
 Bidirectional(forward, f_in::Integer, f_out::Integer, backward, b_in::Integer, b_out::Integer) = Bidirectional(forward(f_in, f_out), backward(b_in, b_out))
 
 # Constructor for forward and backward having the same size
-Bidirectional(forward, backward, in::Integer, out::Integer) = Bidirectional(forward, in, out, backward, in, out)
+Bidirectional(forward, backward, in::Integer, out::Integer) = Bidirectional(forward(in, out), backward(in, out))
 
 # Constructor to add the same cell as forward and backward with given input and output sizes
-Bidirectional(rnn, in::Integer, out::Integer) = Bidirectional(rnn, in, out, rnn, in, out)
+Bidirectional(rnn, in::Integer, out::Integer) = Bidirectional(rnn(in, out), rnn(in, out))
+
 
 # Concatenate the forward and reversed backward weights
 function (m::Bidirectional)(x::Union{AbstractVecOrMat{T},OneHotArray}) where {T}
   return vcat(m.forward(x), reverse(m.backward(reverse(x; dims=1)); dims=1))
 end
 
-Flux.@functor Bidirectional
+@functor Bidirectional
+Base.getproperty(m::Bidirectional, sym::Symbol) = getfield(m, sym)
 
 # Show adaptations
 function _big_show(io::IO, obj::Bidirectional, indent::Int=0, name=nothing)  
   println(io, " "^indent, isnothing(name) ? "" : "$name = ", nameof(typeof(obj)), "(")
   # then we insert names -- can this be done more generically? 
   for k in propertynames(obj)
-      Flux._big_show(io, getfield(obj, k), indent+2, k)
+      _big_show(io, getfield(obj, k), indent+2, k)
   end
 end
 
-Base.show(io::IO, m::Bidirectional) = _big_show(io, m)
+Base.show(io::IO, m::MIME"text/plain", x::Bidirectional) = _big_show(io, x)
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -439,9 +439,8 @@ A wrapper layer that allows bidirectional recurrent layers.
 
 
 # Examples
-```julia
-BLSTM = Bidirectional(LSTM, 3, 5)
-
+```jldoctest
+julia> BLSTM = Bidirectional(LSTM, 3, 5)
 Bidirectional(
   Recur(
     LSTMCell(3, 5),                     # 190 parameters
@@ -451,22 +450,12 @@ Bidirectional(
   ),
 )         # Total: 10 trainable arrays, 380 parameters,
           # plus 4 non-trainable, 20 parameters, summarysize 2.141 KiB.
-```
 
-```julia 
-BLSTM(rand(Float32, 3)) |> size
+julia> BLSTM(rand(Float32, 3)) |> size
 
 (10,)
-```
 
-```julia 
-model = Chain(
-  Embedding(10000, 200),
-  Bidirectional(LSTM, 200, 128),
-  Dense(256, 5),
-  softmax
-)
-
+julia> model = Chain(Embedding(10000, 200), Bidirectional(LSTM, 200, 128), Dense(256, 5), softmax)
 Chain(
   Embedding(10000, 200),                # 2_000_000 parameters
   Bidirectional(
@@ -497,7 +486,6 @@ Bidirectional(forward, backward, in::Integer, out::Integer) = Bidirectional(forw
 # Constructor to add the same cell as forward and backward with given input and output sizes
 Bidirectional(rnn, in::Integer, out::Integer) = Bidirectional(rnn, in, out, rnn, in, out)
 
-
 # Concatenate the forward and reversed backward weights
 function (m::Bidirectional)(x::Union{AbstractVecOrMat{T},OneHotArray}) where {T}
   return vcat(m.forward(x), reverse(m.backward(reverse(x; dims=1)); dims=1))
@@ -506,8 +494,7 @@ end
 Flux.@functor Bidirectional
 
 # Show adaptations
-function _big_show(io::IO, obj::Bidirectional, indent::Int=0, name=nothing)
-    
+function _big_show(io::IO, obj::Bidirectional, indent::Int=0, name=nothing)  
   println(io, " "^indent, isnothing(name) ? "" : "$name = ", nameof(typeof(obj)), "(")
   # then we insert names -- can this be done more generically? 
   for k in propertynames(obj)
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -430,3 +430,89 @@ Recur(m::GRUv3Cell) = Recur(m, m.state0)
 @adjoint function Broadcast.broadcasted(f::Recur, args...)
   Zygote.∇map(__context__, f, args...)
 end
+
+
+"""
+    Bidirectional{A,B} 
+
+A wrapper layer that allows bidirectional recurrent layers. 
+
+
+# Examples
+```julia
+BLSTM = Bidirectional(LSTM, 3, 5)
+
+Bidirectional(
+  Recur(
+    LSTMCell(3, 5),                     # 190 parameters
+  ),
+  Recur(
+    LSTMCell(3, 5),                     # 190 parameters
+  ),
+)         # Total: 10 trainable arrays, 380 parameters,
+          # plus 4 non-trainable, 20 parameters, summarysize 2.141 KiB.
+```
+
+```julia 
+BLSTM(rand(Float32, 3)) |> size
+
+(10,)
+```
+
+```julia 
+model = Chain(
+  Embedding(10000, 200),
+  Bidirectional(LSTM, 200, 128),
+  Dense(256, 5),
+  softmax
+)
+
+Chain(
+  Embedding(10000, 200),                # 2_000_000 parameters
+  Bidirectional(
+    Recur(
+      LSTMCell(200, 128),               # 168_704 parameters
+    ),
+    Recur(
+      LSTMCell(200, 128),               # 168_704 parameters
+    ),
+  ),
+  Dense(256, 5),                        # 1_285 parameters
+  NNlib.softmax,
+)         # Total: 13 trainable arrays, 2_338_693 parameters,
+          # plus 4 non-trainable, 512 parameters, summarysize 8.922 MiB.
+```
+"""
+struct Bidirectional{A,B} 
+  forward::A
+  backward::B
+end
+
+# Generic constructor for every case
+Bidirectional(forward, f_in::Integer, f_out::Integer, backward, b_in::Integer, b_out::Integer) = Bidirectional(forward(f_in, f_out), backward(b_in, b_out))
+
+# Constructor for forward and backward having the same size
+Bidirectional(forward, backward, in::Integer, out::Integer) = Bidirectional(forward, in, out, backward, in, out)
+
+# Constructor to add the same cell as forward and backward with given input and output sizes
+Bidirectional(rnn, in::Integer, out::Integer) = Bidirectional(rnn, in, out, rnn, in, out)
+
+
+# Concatenate the forward and reversed backward weights
+function (m::Bidirectional)(x::Union{AbstractVecOrMat{T},OneHotArray}) where {T}
+  return vcat(m.forward(x), reverse(m.backward(reverse(x; dims=1)); dims=1))
+end
+
+Flux.@functor Bidirectional
+
+# Show adaptations
+function _big_show(io::IO, obj::Bidirectional, indent::Int=0, name=nothing)
+    
+  println(io, " "^indent, isnothing(name) ? "" : "$name = ", nameof(typeof(obj)), "(")
+  # then we insert names -- can this be done more generically? 
+  for k in propertynames(obj)
+      Flux._big_show(io, getfield(obj, k), indent+2, k)
+  end
+end
+
+Base.show(io::IO, m::Bidirectional) = _big_show(io, m)
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -491,14 +491,8 @@ struct Bidirectional{A,B}
   backward::B
 end
 
-# Generic constructor for every case
-Bidirectional(forward, f_in::Integer, f_out::Integer, backward, b_in::Integer, b_out::Integer) = Bidirectional(forward(f_in, f_out), backward(b_in, b_out))
-
-# Constructor for forward and backward having the same size
-Bidirectional(forward, backward, in::Integer, out::Integer) = Bidirectional(forward(in, out), backward(in, out))
-
-# Constructor to add the same cell as forward and backward with given input and output sizes
-Bidirectional(rnn, in::Integer, out::Integer) = Bidirectional(rnn(in, out), rnn(in, out))
+# Constructor that creates a bidirectional with the same layer for forward and backward
+Bidirectional(rnn, a...; ka...) = Bidirectional(rnn(a...; ka...), rnn(a...; ka...))
 
 
 # Concatenate the forward and reversed backward weights
@@ -507,15 +501,3 @@ function (m::Bidirectional)(x::Union{AbstractVecOrMat{T},OneHotArray}) where {T}
 end
 
 @functor Bidirectional
-Base.getproperty(m::Bidirectional, sym::Symbol) = getfield(m, sym)
-
-# Show adaptations
-function _big_show(io::IO, obj::Bidirectional, indent::Int=0, name=nothing)  
-  println(io, " "^indent, isnothing(name) ? "" : "$name = ", nameof(typeof(obj)), "(")
-  # then we insert names -- can this be done more generically? 
-  for k in propertynames(obj)
-      _big_show(io, getfield(obj, k), indent+2, k)
-  end
-end
-
-Base.show(io::IO, m::MIME"text/plain", x::Bidirectional) = _big_show(io, x)
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -441,9 +441,14 @@ the forward layer weights are concatenated with the reversed order of the backwa
 It is intended to be used with recurrent layers such as `LSTM`, `GRU` or `RNN` to benefit from the sequential information that recurrent 
 layers have, but it will not raise an error if used with a different layer such as `Dense`, as long as the layer is compatible with the concatenation function `vcat`.
 
+For flexibility, it is possible to use the contructor `Bidirectional(rnn, in::Int, out::Int, a...; ka...)` by passing the input and output dimensions 
+together with the desired recurrent layers (one of `LSTM`, `GRU`, or `RNN`). Check the examples below for more details.
+
 # Examples
+
+1. Using the flexible constructor to create a bidirectional LSTM layer (BiLSTM):
 ```jldoctest
-julia> BLSTM = Bidirectional(LSTM, 3, 5)
+julia> BLSTM = Bidirectional(LSTM, 3, 5) 
 Bidirectional(
   Recur(
     LSTMCell(3, 5),                     # 190 parameters
@@ -453,9 +458,15 @@ Bidirectional(
   ),
 )         # Total: 10 trainable arrays, 380 parameters,
           # plus 4 non-trainable, 20 parameters, summarysize 2.141 KiB.
+```
+2. Checking the dimension after running the bidirectional layer on an input vector. Shows that the 
+dimension of the output is twice the dimension of the input:
+```jldoctest
 julia> Bidirectional(LSTM, 3, 5)(rand(Float32, 3)) |> size
 (10,)
-
+```
+3. It is possible to use the bidirectional layer inside the `Chain` container:
+```
 julia> model = Chain(Embedding(10000, 200), Bidirectional(LSTM, 200, 128), Dense(256, 5), softmax)
 Chain(
   Embedding(10000, 200),                # 2_000_000 parameters
@@ -472,6 +483,32 @@ Chain(
 )         # Total: 13 trainable arrays, 2_338_693 parameters,
           # plus 4 non-trainable, 512 parameters, summarysize 8.922 MiB.
 ```
+4. It is also possible to use the default constructor
+```jldoctest
+julia> BiLSTM = Bidirectional(GRU(3, 5), GRU(3, 5))
+Bidirectional(
+  Recur(
+    GRUCell(3, 5),                      # 140 parameters
+  ),
+  Recur(
+    GRUCell(3, 5),                      # 140 parameters
+  ),
+)         # Total: 8 trainable arrays, 280 parameters,
+          # plus 2 non-trainable, 10 parameters, summarysize 1.562 KiB.
+```
+5. And use other parameters available on the recurrent layers
+```jldoctest
+julia> BiLSTM = Bidirectional(RNN(3, 5, tanh; init=glorot_normal), LSTM(3, 5; initb=zeros32, init_state=zeros32))
+Bidirectional(
+  Recur(
+    RNNCell(3, 5, tanh),                # 50 parameters
+  ),
+  Recur(
+    LSTMCell(3, 5),                     # 190 parameters
+  ),
+)         # Total: 9 trainable arrays, 240 parameters,
+          # plus 3 non-trainable, 15 parameters, summarysize 1.500 KiB.
+```
 """
 struct Bidirectional{A,B} 
   forward::A
@@ -479,7 +516,8 @@ struct Bidirectional{A,B}
 end
 
 # Constructor that creates a bidirectional with the same layer for forward and backward
-Bidirectional(rnn, a...; ka...) = Bidirectional(rnn(a...; ka...), rnn(a...; ka...))
+# Needs to have `in` and `out` explicitly declared to avoid conflicts with the default construtor
+Bidirectional(rnn, in::Int, out::Int, a...; ka...) = Bidirectional(rnn(in, out, a...; ka...), rnn(in, out, a...; ka...))
 
 
 # Concatenate the forward and reversed backward weights
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -1,3 +1,4 @@
+
 gate(h, n) = (1:h) .+ h*(n-1)
 gate(x::AbstractVector, h, n) = @view x[gate(h,n)]
 gate(x::AbstractMatrix, h, n) = x[gate(h,n),:]
@@ -433,7 +434,13 @@ end
 
 """
     Bidirectional{A,B} 
-A wrapper layer that allows bidirectional recurrent layers. 
+
+A wrapper layer that allows the use of [bidirectional](https://ieeexplore.ieee.org/document/650093) layers. It contains two parts that are Flux layers: `forward` and `backward` where 
+the forward layer weights are concatenated with the reversed order of the backward layer weights.
+
+It is intended to be used with recurrent layers such as `LSTM`, `GRU` or `RNN` to benefit from the sequential information that recurrent 
+layers have, but it will not raise an error if used with a different layer such as `Dense`, as long as the layer is compatible with the concatenation function `vcat`.
+
 # Examples
 ```jldoctest
 julia> BLSTM = Bidirectional(LSTM, 3, 5)
@@ -446,8 +453,22 @@ Bidirectional(
   ),
 )         # Total: 10 trainable arrays, 380 parameters,
           # plus 4 non-trainable, 20 parameters, summarysize 2.141 KiB.
-julia> BLSTM(rand(Float32, 3)) |> size
+julia> Bidirectional(LSTM, 3, 5)(rand(Float32, 3)) |> size
 (10,)
+
+julia> Bidirectional(LSTM, 3, 5)(rand(Float32, 3))
+10-element Vector{Float32}:
+  0.009660141
+ -0.011628074
+ -0.017348368
+ -0.002131971
+  0.0152327195
+ -0.024785668
+  0.021315152
+ -0.015476399
+  0.01589005
+ -0.002576883
+
 julia> model = Chain(Embedding(10000, 200), Bidirectional(LSTM, 200, 128), Dense(256, 5), softmax)
 Chain(
   Embedding(10000, 200),                # 2_000_000 parameters
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -456,19 +456,6 @@ Bidirectional(
 julia> Bidirectional(LSTM, 3, 5)(rand(Float32, 3)) |> size
 (10,)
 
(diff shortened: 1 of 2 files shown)


src/layers/recurrent.jl
```diff
@@ -430,3 +429,72 @@ Recur(m::GRUv3Cell) = Recur(m, m.state0)
...
+# Show adaptations
+function _big_show(io::IO, obj::Bidirectional, indent::Int=0, name=nothing)  
+  println(io, " "^indent, isnothing(name) ? "" : "$name = ", nameof(typeof(obj)), "(")
+  # then we insert names -- can this be done more generically? 
+  for k in propertynames(obj)
+      _big_show(io, getfield(obj, k), indent+2, k)
+  end
+end
+
+Base.show(io::IO, m::MIME"text/plain", x::Bidirectional) = _big_show(io, x)
```
mcabbott, line 500: Without commenting on the actual functionality, I think it should not be necessary to touch much of the printing code. 

This should certainly define 2-arg show, which is what's called in many contexts. 
```suggestion
function Base.show(io::IO, b::Bidirectional)
  print(io, "Bidirectional(")
  show(io, b.forward)
  print(io, ", ")
  show(io, b.backward)
  print(io, ")")
end
```
That's enough to produce this:
```
julia> function Base.show(io::IO, b::Bidirectional)
         print(io, "Bidirectional(")
         show(io, b.forward)
         print(io, ", ")
         show(io, b.backward)
         print(io, ")")
       end;

julia> b = Bidirectional(LSTM, 200, 128)
Bidirectional(Recur(LSTMCell(200, 128)), Recur(LSTMCell(200, 128)))

julia> model = Chain(Embedding(10000, 200), Bidirectional(LSTM, 200, 128), Dense(256, 5), softmax)
Chain(
  Embedding(10000, 200),                # 2_000_000 parameters
  Bidirectional(
    Recur(
      LSTMCell(200, 128),               # 168_704 parameters
```
If in addition you think this is big enough that it ought to be expanded at top level, then it could be added to the list here:

https://github.com/FluxML/Flux.jl/blob/master/src/layers/show.jl#L3

src/layers/recurrent.jl
```diff
@@ -430,3 +429,72 @@ Recur(m::GRUv3Cell) = Recur(m, m.state0)
...
+Bidirectional(rnn, in::Integer, out::Integer) = Bidirectional(rnn(in, out), rnn(in, out))
+
+
+# Concatenate the forward and reversed backward weights
+function (m::Bidirectional)(x::Union{AbstractVecOrMat{T},OneHotArray}) where {T}
+  return vcat(m.forward(x), reverse(m.backward(reverse(x; dims=1)); dims=1))
+end
+
+@functor Bidirectional
+Base.getproperty(m::Bidirectional, sym::Symbol) = getfield(m, sym)
```
mcabbott, line 489: Also, I think this is the default. Did you run into some problem without it?

src/layers/recurrent.jl
```diff
@@ -430,3 +429,72 @@ Recur(m::GRUv3Cell) = Recur(m, m.state0)
...
+
+# Show adaptations
+function _big_show(io::IO, obj::Bidirectional, indent::Int=0, name=nothing)
```
DhairyaLGandhi, line 492: Why do we need this? Can we omit this method?

src/layers/recurrent.jl
```diff
@@ -430,3 +429,72 @@ Recur(m::GRUv3Cell) = Recur(m, m.state0)
...
+    ),
+    Recur(
+      LSTMCell(200, 128),               # 168_704 parameters
+    ),
+  ),
+  Dense(256, 5),                        # 1_285 parameters
+  NNlib.softmax,
+)         # Total: 13 trainable arrays, 2_338_693 parameters,
+          # plus 4 non-trainable, 512 parameters, summarysize 8.922 MiB.
+```
```
mcabbott, line 466: I also think explaining what this does would be good, and trying to think up very simple examples to demonstrate it. Perhaps
```
julia> Bidirectional(x -> @show(x), y -> 100*@show(y))([1 2; 3 4])
x = [1 2; 3 4]
y = [3 4; 1 2]
4×2 Matrix{Int64}:
   1    2
   3    4
 100  200
 300  400
```

src/layers/recurrent.jl
```diff
@@ -430,3 +429,72 @@ Recur(m::GRUv3Cell) = Recur(m, m.state0)
...
+end
+
+# Generic constructor for every case
+Bidirectional(forward, f_in::Integer, f_out::Integer, backward, b_in::Integer, b_out::Integer) = Bidirectional(forward(f_in, f_out), backward(b_in, b_out))
+
+# Constructor for forward and backward having the same size
+Bidirectional(forward, backward, in::Integer, out::Integer) = Bidirectional(forward(in, out), backward(in, out))
+
+# Constructor to add the same cell as forward and backward with given input and output sizes
+Bidirectional(rnn, in::Integer, out::Integer) = Bidirectional(rnn(in, out), rnn(in, out))
```
ToucheSir, line 480: I'm on the fence about whether these should exist. They disallow any passing of additional args (and more importantly, kwargs) to the cell or whatever other layer is being used. That, and
```julia
Bidirectional(RNN(f_in, f_out), RNN(b_in, b_out))
Bidirectional(RNN(f_in, f_out), RNN(f_in, f_out))
r = RNN(in, out)
Bidirectional(r, r)
```
Don't seem that much more verbose while being very explicit about which dimensions/weights are shared.

mcabbott: Note BTW that `Bidirectional(r, r)` shares parameters. Agree though that being explicit is probably better.

ToucheSir: Woops, disregard the last 2 lines then. I'd say this shows how having this particular set of convenience constructors can mess with one's intuition about what each one is doing.

src/layers/recurrent.jl
dcecchini: I was having some problems with the `show` method and after many changes, I thought it could be related but I was not certain. If it is the default then the problems I was having (error when using the `show`) were related something else.

src/layers/recurrent.jl
dcecchini: Do you mean the `getproperty`? I think if it is the default then we can commit it (as mentioned by @mcabbott above).

src/layers/recurrent.jl
dcecchini: In this example, I wanted to show how to create a `Bidirectional LSTM` layer by simply passing the `LSTM` layer as the first parameter and the input and output sizes. The `Bidirectional` wrapper could create Bidirectional `LSTM`, `GRU` or `RNN` layers in the same way. Depending on the choice or needs of the problem the person is solving (needs longer memory, faster computations, etc.), the person could choose between any of the recurrent layers in his model. 

The constructor is there so that it could be easily added as a layer within the `Chain` function (in a similar way as done in `Keras`).

src/layers/recurrent.jl
dcecchini: I can see your point and it makes total sense to allow additional arguments to be passed. I would say that at least the constructor
```
Bidirectional(rnn, in::Integer, out::Integer) = Bidirectional(rnn(in, out), rnn(in, out))
```
could be kept for the intended functionality to add bidirectional layers easily inside the `Chain` function. 

And probably it should be generalized to something like:

```
Bidirectional(rnn, a...; ka...) = Bidirectional(rnn(a...; ka...), rnn(a...; ka...))
```

src/layers/recurrent.jl
dcecchini: Actually, I think both cases in line 3 or line 53 of the `show.jl` could be used, I don't have a personal preference. I liked the way it is displayed in one line as you showed above, but sometimes the line could get too big. Do you have a preference?

src/layers/recurrent.jl
ToucheSir: What Dhairya and Michael are saying is that since it's the default, you don't have to write it out and the line can be removed :)

dcecchini: Sorry, I just saw a typo on my message, the intended word was `omit` it, not commit... sorry. I will commit the removal of it in the next changes ;)

src/layers/recurrent.jl
ToucheSir: I'd be ok with `Bidirectional(rnn, a...; ka...) = Bidirectional(rnn(a...; ka...), rnn(a...; ka...))` as long as it doesn't conflict with the default constructor. Not sure I understand the point about `Chain` though: why would having a convenience constructor make adding a bidi layer easier in that scenario in particular?

src/layers/recurrent.jl
mcabbott: My point is that, while this may be the intended use, nothing in the definition of this struct actually assumes anything about RNNs. It just calls whatever you provide, and then `vcat`. The docstring should explain this basic level of what's going to happen if you press this button, not just the level of under what conditions you ought to do so.

src/layers/recurrent.jl
mcabbott: Your call, I guess. Maybe the rule of thumb is that if it'll almost always fit on one line, then it probably shouldn't be expanded. It it's likely to get a whole `Chain` as one of its arguments (like Parallel) then probably it should be.

src/layers/recurrent.jl
dcecchini: It may be of personal preference, but it seems cleaner to me writing

```julia
model = Chain(Embedding(1000, 200), Bidirectional(LSTM, 200, 128), Dense(256, 5), softmax)
```

instead of 

```julia
model = Chain(Embedding(1000, 200), Bidirectional(LSTM(200, 128), LSTM(200, 128)), Dense(256, 5), softmax)
```

The thing is that in `keras` the forward and backward parts are by default the same kind of recurrent layer even though the option to use different is also there. So, the usual way to add them in the model are like:

```python
model = Sequential() \
    .add(Bidirectional(LSTM(10, return_sequences=True), input_shape=(5, 10))) \
    .add(Bidirectional(LSTM(10))) \
    .add(Dense(5)) \
    .add(Activation('softmax'))
``` 

I am not sure which is best, I tried to make it easy to decide which recurrent layer we want to use and the input and output dimensions that are needed. If it would be better, we could change the constructors to:

```julia
Bidirectional(rnn) = Bidirectional(rnn, rnn)
```

and use like:

```julia
model = Chain(Embedding(1000, 200), Bidirectional(LSTM(200, 128)), Dense(256, 5), softmax)
```

which would also accept `LSTM(a...; ka...)`. 

What do you think?

src/layers/recurrent.jl
dcecchini: Ok, I will treat it like `Parallel` and add the `:Bidirectional` to the `show.jl` file at line 3.

src/layers/recurrent.jl
dcecchini: I can see your point now. I am trying to improve the documentation with a better description:

> A wrapper layer that allows the use of [bidirectional](https://ieeexplore.ieee.org/document/650093) layers. It contains two parts that are Flux layers: `forward` and `backward` where the forward layer weights are concatenated with the reversed order of the backward layer weights.
>
> It is intended to be used with recurrent layers such as `LSTM`, `GRU` or `RNN` to benefit from the sequential information that recurrent layers have, but it will not raise an error if used with a different layer such as `Dense`, as long as the layer is compatible with the concatenation function `vcat`.

And also to add the example output when calling with a simple vector (the output 5 becomes 10 since it has forward and backward weights concatenated):

```julia
julia> Bidirectional(LSTM, 3, 5)(rand(Float32, 3))
10-element Vector{Float32}:
  0.009660141
 -0.011628074
 -0.017348368
 -0.002131971
  0.0152327195
 -0.024785668
  0.021315152
 -0.015476399
  0.01589005
 -0.002576883

```

src/layers/recurrent.jl
ToucheSir: Oh ok, I would say this is something that applies in general rather than just when in a `Chain` (recall that like PyTorch or JAX, Flux is "eager" and container layers like `Chain` are provided purely for convenience). Thanks for clarifying.

>  If it would be better, we could change the constructors to...

Note that

```julia
Bidirectional(rnn) = Bidirectional(rnn, rnn)
Bidirectional(LSTM(200, 128))
```

is not equivalent to

```julia
Bidirectional(rnn, in::Integer, out::Integer) = Bidirectional(rnn(in, out), rnn(in, out))
Bidirectional(LSTM, 200, 128)

```

because the RNN parameters (i.e. all the weights) are shared in the first case but completely independent in the second.

ToucheSir: Just to provide some context, the reason it may seem I'm being stingy about these constructors is that per SemVer, any removals/most signature tweaks in the public API are breaking changes. Thus it is far easier to add convenience functionality later if many people find the default too verbose than to inadvertently introduce an ambiguity/potentially confusing interface that we'd have to break compat to fix.

dcecchini: You are right, the weights should not b shared like that unless this is the intention. And yes, my perception is not related to `Chain` specifically, thanks for pointing that out. 

I would suggest, then, to keep the implementation as:

```julia
Bidirectional(rnn, a...; ka...) = Bidirectional(rnn(a...; ka...), rnn(a...; ka...))
```
that adds the possibility to use other parameters in the recurrent layer.

src/layers/recurrent.jl
```diff
@@ -430,3 +429,72 @@ Recur(m::GRUv3Cell) = Recur(m, m.state0)
...
+
+
+# Concatenate the forward and reversed backward weights
+function (m::Bidirectional)(x::Union{AbstractVecOrMat{T},OneHotArray}) where {T}
+  return vcat(m.forward(x), reverse(m.backward(reverse(x; dims=1)); dims=1))
```
ToucheSir, line 485:
```suggestion
  return vcat(m.forward(x), reverse(m.backward(reverse(x; dims=1)); dims=1))
```
Sorry, just found this. When applying Flux RNNs on dense sequences, the temporal dim is actually the last one. See https://github.com/FluxML/Flux.jl/blob/dd6b70ecf3116d15183ea9497983a490850a50bf/src/layers/recurrent.jl#L83 and https://github.com/FluxML/Flux.jl/blob/dd6b70ecf3116d15183ea9497983a490850a50bf/src/layers/recurrent.jl#L27.

ToucheSir:
@dcecchini do you have local tests that weren't pushed? I only see changes in `src/`.

src/layers/recurrent.jl
dcecchini: > Just to provide some context, the reason it may seem I'm being stingy about these constructors is that per SemVer, any removals/most signature tweaks in the public API are breaking changes. Thus it is far easier to add convenience functionality later if many people find the default too verbose than to inadvertently introduce an ambiguity/potentially confusing interface that we'd have to break compat to fix.

I see and understand. Maybe keep only the one above and discard the other two that I added that probably are too much.

dcecchini:
> @dcecchini do you have local (behavioural) tests that weren't pushed? I only see changes in `src/`.

I don't think so, I did all the changes on `src/` only and pushed all of them. I will make a few modifications as mentioned on the comments here and push them shortly.

src/layers/recurrent.jl
mcabbott: The collision concern is that `Bi(f, a...) = Bi(f(a...), f(a...))` also accepts two arguments, but does something quite different to `Bi(f, g)`. It could be `Bi(f, a::Int...)` if the goal is just to allow integer parameters.

(It could also be `Bidirectional(LSTM, in => out)` but maybe that's another story.)

src/layers/recurrent.jl
mcabbott: My complaint about showing this is that it still doesn't *show* what it actually does. This could equally well be `Parallel(vcat, ...)`.

Secondarily, printing random numbers is tricky for jldoctests.

src/layers/recurrent.jl
dcecchini: I tested to use the `Bidirectional(rnn, a...; ka...)` by calling `Bidirectional(LSTM, 3, 5)` and it worked but I didn't pay attention that it has the same number of parameters as the default constructor. I can see how confusing this can get.  It should be able to accept any parameter that the recurrent layers accept (like `in::Integer, out::Integer, σ=tanh; init=Flux.glorot_uniform, initb=zeros32, init_state=zeros32`)

src/layers/recurrent.jl
dcecchini: I'm quite confused now because I was expecting the variable `x` to be the result of an `OneHotMatrix` operation on a sentence and would be a matrix. In my experiments, I was using an array of one-hot encoded sentences (that were padded to fixed size) like this:

```julia
X_onehot = [OneHotMatrix(x, vocab_size) for x in X]
y_onehot = [OneHotMatrix(x, num_tags) for x in y]
```

where `X` is an array of padded sentences and y is the corresponding labels of each word in the sentence (I am experimenting with Named Entity Recognition models). So the input data would be a matrix `(batch, (time, features))` and not in the `(features, batch, time)` format. 

I am not sure how to proceed.

src/layers/recurrent.jl
dcecchini: The problem seems to arise only if we don't pass `in` and `out` together with the recurrent layer like `Bidirectional(LSTM, 3)`. That would create a struct with forward part equal to `LSTM` and the backward part equal to `3` that is a mistake. But I think we can assume that at least 3 parameters should be used because `in` and `out` at least are needed.

src/layers/recurrent.jl
dcecchini: My first try was to use the `Parallel` but I kept getting dimension mismatch errors. That's why I added the `dims=1` to solve that... I will remove the random numbers from the examples and think how to better show it.

src/layers/recurrent.jl
mcabbott: You could impose this, by writing `Bi(f, a::Int, b::Int, cs...) = Bi(f(a, b, cs...), f(a, b, cs...))`. Then there's a clear rule, but still a rule that needs more explaining than ideal. 

Are there other ways? `Bidirectional(() -> RNN(3,4,tanh))` is one pattern, a bit ugly.

dcecchini: Too ugly to try something like the below?

```julia
Bidirectional(rnn, a...; ka...) = length(a) >1 ? Bidirectional(rnn(a...; ka...), rnn(a...; ka...)) : Bidirectional(rnn, a)
```

ToucheSir: If the only goal is to avoid repeating the in and out dimensions, `Bidirectional(ntuple(_ -> RNN(3,4,tanh), 2)...)` would work. However, the bigger goal seems to be brevity.

dcecchini: I don't really understand the code you shared... is it creating a specific dimension `(3,4)` and a specific layer `RNN`? How can I use this to accept any of `LSTM`, `RNN`, or `GRU` and accept any specified input dimension and output dimension? 

Anyway, `Bi(f, a::Int, b::Int, cs...; ka...) = Bi(f(a, b, cs...; ka...), f(a, b, cs...; ka...))` would work fine and without conflict with the default constructor and would be ok for me.

src/layers/recurrent.jl
ToucheSir: https://fluxml.ai/Flux.jl/stable/models/recurrence/#Working-with-sequences has most of the nitty-gritty details about working with batched sequences for RNNs. In short, the only supported formats are `((features, batch), time)` and `(features, batch, time)`. Unlike Python frameworks, Flux puts the batch dim last for all layers because of column major layout. RNNs just add another time dim after that.

Since this definition of Bidirectional takes a contiguous array instead of a vector of arrays, `m.forward()` and `m.backward()` dispatch to `(m::Recur)(x::AbstractArray{T, 3})`. To support both, you'd need something like the following (note: untested!):
1. `(m::Bidirectional)(x::AbstractArray{T, 3}) where T` for `(features, batch, time)`
2. `(m::Bidirectional)(x::Vector{<:AbstractVecOrMat{T}}) where T` for `((features, batch), time)`

src/layers/recurrent.jl
ToucheSir: Sure, let's go with that. As long as the main docstring explains this signature clearly, I don't think people will be tripped up too much.

> I don't really understand the code you shared...

`ntuple(f, N)` calls `f` N and saves the results into a tuple of size N. `ntuple(_ -> RNN(3,4,tanh), 2) == (RNN(3,4,tanh), RNN(3,4,tanh))`. The `...` splats that into 2 arguments for the default `Bidirectional` constructor.

src/layers/recurrent.jl
mcabbott: Still needs 2-arg show, as above.

ToucheSir:
Testing wise, having something in https://github.com/FluxML/Flux.jl/blob/6168692ddb2680ebe02f461bf0c59c2b6cb0ff1e/test/layers/recurrent.jl that checks the logic (need not be complex) would be good. You can see how this is done for other composite layers such as [`SkipConnection`](https://github.com/FluxML/Flux.jl/blob/6168692ddb2680ebe02f461bf0c59c2b6cb0ff1e/test/layers/basic.jl#L131-L141) or [`Parallel`](https://github.com/FluxML/Flux.jl/blob/6168692ddb2680ebe02f461bf0c59c2b6cb0ff1e/test/layers/basic.jl#L189-L245).

src/layers/recurrent.jl
```diff
@@ -430,3 +430,61 @@ Recur(m::GRUv3Cell) = Recur(m, m.state0)
 @adjoint function Broadcast.broadcasted(f::Recur, args...)
   Zygote.∇map(__context__, f, args...)
 end
+
+
+"""
+    Bidirectional{A,B}
```
darsnack, line 436: Once we settle on the constructor, this line should be the various constructors not the type definition.

src/layers/recurrent.jl
dcecchini: Do you get any errors with this? I could instantiate the bidirectional layer and it shows as expected:

```julia
rnn = Flux.Bidirectional(Flux.LSTM, 3, 5)
Bidirectional(
  Recur(
    LSTMCell(3, 5),                     # 190 parameters
  ),
  Recur(
    LSTMCell(3, 5),                     # 190 parameters
  ),
)         # Total: 10 trainable arrays, 380 parameters,
          # plus 4 non-trainable, 20 parameters, summarysize 2.141 KiB.
```

mcabbott: Not errors, but e.g. `(rnn, rnn)` will use 2-arg show.

src/layers/recurrent.jl
DhairyaLGandhi: May be better to not have a `a...` variant since it doesn't seem to be very clean about constructing the layers. I think a single argument `Bidirectional(layer) = Bidirectional( layer, layer)` would be cleaner.

ToucheSir: The issue with that constructor is that (AIUI), the weights are generally not shared between both RNNs in the stack.

DhairyaLGandhi: Either way, it seems to me that avoiding confusion down the line is better.

dcecchini: I re-checked the `LSTMCell` and `GRUCell` and saw that there is no extra `a...` arguments, only the keyword arguments that can be used (apart from the `in` and `out`). The only cell that allows to change, for example, the activation function is the vanilla `RNNCell`:

```julia
struct RNNCell{F,A,V,S}
  σ::F
  Wi::A
  Wh::A
  b::V
  state0::S
end

# Can change the activation function σ
RNNCell(in::Integer, out::Integer, σ=tanh; init=Flux.glorot_uniform, initb=zeros32, init_state=zeros32) = 


struct LSTMCell{A,V,S}
  Wi::A
  Wh::A
  b::V
  state0::S
end

# Arguments are only in and out plus ka
function LSTMCell(in::Integer, out::Integer;
                  init = glorot_uniform,
                  initb = zeros32,
                  init_state = zeros32)
  cell = LSTMCell(init(out * 4, in), init(out * 4, out), initb(out * 4), (init_state(out,1), init_state(out,1)))
  cell.b[gate(out, 2)] .= 1
  return cell
end

struct GRUCell{A,V,S}
  Wi::A
  Wh::A
  b::V
  state0::S
end

# Same as LSTM
GRUCell(in, out; init = glorot_uniform, initb = zeros32, init_state = zeros32) =
  GRUCell(init(out * 3, in), init(out * 3, out), initb(out * 3), init_state(out,1))
```

So I would say we can have `Bidirectional(rnn, in::Int, out::Int; ka...)` only.

mcabbott: It should allow user-written layers too, though. And probably should be just `Bidirectional(rnn, in::Integer, args...; kw...)`, as that's enough to be unambiguous.

src/layers/recurrent.jl
mcabbott: But the rnn layers do not literally accept a vector of matrices. At the link this is iterated through by hand. Should method 2 of `Bidirectional(fwd, rev)(x)` handle that? 
```
julia> LSTM(3,5)(randn(Float32, 3,7)) |> size
(5, 7)

julia> LSTM(3,5)(randn(Float32, 3,7,11)) |> size
(5, 7, 11)

julia> LSTM(3,5)([randn(Float32, 3,7) for _ in 1:11]) |> size
ERROR: MethodError
```

src/layers/recurrent.jl
dcecchini: Adjusted [here](https://github.com/FluxML/Flux.jl/pull/1790/commits/18b406ad08a76a8ec1a172900309b68de1c19ce9)

src/layers/recurrent.jl
ToucheSir: > Should method 2 of `Bidirectional(fwd, rev)(x)` handle that?

Yes, exactly.

mcabbott: And applied to `randn(Float32, 3,7)`, should it fail, or still reverse in the 3rd dimension?

ToucheSir: The signatures specifically prohibit that, so it should fail. Did you have a specific edge case in mind that isn't caught by the above?

mcabbott: No, I just wonder about the mismatch between what LSTM accepts and what this accepts. Prohibiting things which don't make sense is good, you want to know soon.

ToucheSir: Right, the RNN layers can work with single timesteps but `Bidirectional` can't, as it needs to see the whole sequence up-front in order to reverse it. If anything the former should be brought in line with the latter, but that's a conversation for another issue (#1678 probably).

dcecchini: Thank you for the great conversation here and the suggestions. I just tested and the `Bidirectional`, as it is, is able to process  `randn(Float32, 3,7)`:

```julia
Bidirectional(LSTM, 3, 5)(randn(Float32, 3, 7)) |> size
(10, 7)
```

But about the format, I am still confused on how to preprocess the text data so that it would end in `(seq_length, (features, samples))` format. It seems counterintuitive to me. 

I usually follow: `read the data -> split sentences -> split words -> pad -> one-hot`. So my data would be an array with N sentences, where every sentence is described as a one-hot matrix of its words. In this way, I ended up with the `(samples, (features, seq_length))` format. How should I preprocess the text data so that I would end up with `(seq_length, (features, samples))`.

Also, by checking these formats I discovered that I should probably use `dims=2` in my formulation on the `reverse` function (not use the default to reverse in all dimensions, it should reverse only the time that in my case is the second dimension of the onehotmatrix).

ToucheSir:
To be clear, accepting that input is a bug, not a feature :P

Workflow-wise, the key is to put the batch dim in between the time and
feature ones. There are a few ways to do this, but one of the easiest would
be Flux.onehotbatch(transpose(reduce(hcat, batch_of_padded_sentences)),
vocab). See the Flux.onehotbatch
<https://fluxml.ai/Flux.jl/stable/data/onehot/#Flux.onehotbatch> docs for
more details. We're actively thinking about how to make working with RNNs
more ergonomic (the old interface is not terribly so), but in the meantime
getting your data into (samples, seq_length) or (samples, (seq_length)) first
and then doing the one-hot transform is the easiest way to go about things.

dcecchini:
> To be clear, accepting that input is a bug, not a feature :P Workflow-wise, the key is to put the batch dim in between the time and feature ones. There are a few ways to do this, but one of the easiest would be Flux.onehotbatch(transpose(reduce(hcat, batch_of_padded_sentences)), vocab). See the Flux.onehotbatch <https://fluxml.ai/Flux.jl/stable/data/onehot/#Flux.onehotbatch> docs for more details. We're actively thinking about how to make working with RNNs more ergonomic (the old interface is not terribly so), but in the meantime getting your data into (samples, seq_length) or (samples, (seq_length)) first and then doing the one-hot transform is the easiest way to go about things.

Yes! To add the batch between time and feature is not difficult, the problem is to have the batch as the last dimension! I was able to transform in `(time, (samples, features))` format (being an array of array of array and not an array of matrixes), but could not keep the samples at the end. I used: 

```julia
[[X_onehot[i][:, j] for i in 1:length(X_onehot)] for j in 1:size(X_onehot[1])[2]]
```

but after checking your example I was able to do this:

```julia
aux = reduce(hcat, X_pad); # size(aux) = (time, samples)
Xs = [Flux.onehotbatch(aux[i, :], 1:vocab_size) for i in 1:size(aux )[1]]; 
size(Xs) # (time, )
size(Xs[1]) # (features, samples)
```

So, at the end, I had to use the one-hot encoding in a different way: instead of using it in a sentence, use it in arrays representing first words, seconds words etc.

ToucheSir:
> `(time, (features, samples))` as input to the Embedding layer, but it was not working

You'd have to pass each timestep in separately to the embedding layer, because it takes an input of size `(samples,)` or `(features, samples)`. This holds true for the alternative format you posted as well, because `Embedding` doesn't know how to work with the outer vector otherwise.

> Also, I tried to pass it to LSTM cells and it didn't work as well.

Same story here. `LSTM` takes an input of shape `(features, samples)` (one timestep, you'll have to pass it each timestep separately if you're using `((features, samples), time)`, or `(features, samples, time)` (multiple timesteps).

Alternatively, you could use dense arrays all the way through. Given a set of inputs of shape `(samples, time)`, pass it to `Embedding` to get an output of `(features, samples, time)`. As you can see, this is already in the right shape to be handled by `LSTM` or any following RNN layer.



> ...it allows to send the whole sentence in a batch

Just to clarify, using the time dimension as the batch dim means that each token in the sentence will be processed in parallel...which is probably not what you want with an RNN :P

NeroBlackstone:
Any updates for this PR?
This is a very useful feature.
Are there any other examples of building bi-LSTM in Flux.jl 0.14?

NeroBlackstone:
Questions about this implementation of Bidirectional RNN:

My question is, since bidirectional RNN requires a full time step (forward hidden state calculation ➡️ and backward hidden state calculation⬅️), so, for a single time step (1/seq_length) input, we cannot calculate the hidden state of bidirectional RNN, right?

But if we initialize model in this PR:

``` julia
julia> model = Bidirectional(LSTM, 3, 5) 
Bidirectional(LSTMCell(3 => 5), LSTMCell(3 => 5))

julia> x = rand(Float32, 3)
3-element Vector{Float32}:
 0.40451288
 0.40546536
 0.7272219

julia> model(x)
10-element Vector{Float32}:
 -0.15061164
 -0.105684996
 -0.08399847
  0.113812014
 -0.03911737
  0.061232302
 -0.011352453
 -0.0018385417
 -0.06642
  0.006111855
```

It output single step result, is this work expected? 

I think the correct behavior is (for batch size 1):

``` julia
# input size(all_time_step) =  (sequence_len * features) 
julia> model(all_time_step)
# output size = (sequence_len * outsize*2)
```

https://github.com/FluxML/Flux.jl/blob/18b406ad08a76a8ec1a172900309b68de1c19ce9/src/layers/recurrent.jl#L523-L525

NeroBlackstone:
It seems that the original author has given up on implementing it, and I want to try to continue his work (this is related to my graduation,

src/layers/recurrent.jl
```diff
@@ -430,3 +430,106 @@ Recur(m::GRUv3Cell) = Recur(m, m.state0)
...
+# Constructor that creates a bidirectional with the same layer for forward and backward
+# Needs to have `in` explicitly declared to avoid conflicts with the default construtor
+Bidirectional(rnn, in::Integer, args...; ka...) = Bidirectional(rnn(in, args...; ka...), rnn(in, args...; ka...))
+
+# Concatenate the forward and reversed backward weights
+function (m::Bidirectional)(x::Union{AbstractVecOrMat{T},OneHotArray}) where {T}
+  return vcat(m.forward(x), reverse(m.backward(reverse(x; dims=1)); dims=1))
+end
+
+@functor Bidirectional
```
NeroBlackstone, line 527: `Flux.@layer` here?

NeroBlackstone:
> I think BRNN should take all the time step, so We can't use it like RNN that we have implemented, which means:
> 
> This usage is valid in RNN:
> 
> ```julia
> # 2 => feature, 3 => sequence_len, 4 => batch_size
> x = [[1 2 1 2; 3 4 3 4] for i = 1:3]
> rnn = RNN(2=>5)
> # out_dim = 3(sequence_len) * 5(hidden) * 4(batch_size)
> rnn_out = [rnn(xi) for xi in x]
> ```
> 
> But same usage is invalid in BRNN:
> 
> ```julia
> # 2 => feature, 3 => sequence_len, 4 => batch_size
> x = [[1 2 1 2; 3 4 3 4] for i = 1:3]
> brnn = Bidirectional(RNN,2,5)
> # out_dim = 3(sequence_len) * 10(hidden) * 4(batch_size)
> brnn_out = [brnn(xi) for xi in x]
> ```
> 
> Because in `(m::Bidirectional)(x::Union{AbstractVecOrMat{T},Flux.OneHotArray}) where {T}`:
> 
> ```julia
> function (m::Bidirectional)(x::Union{AbstractVecOrMat{T},Flux.OneHotArray}) where {T}
>     #size(x) = (2,4)(feature*batch_size)
>     return vcat(m.forward(x), reverse(m.backward(reverse(x; dims=1)); dims=1))
> end
> ```
> 
> the size of `x` is (feature*batch_size), we actually "reverse feature", not "reverse timestep"

NeroBlackstone:
So for BRNN input, I think correct dimensions is:

For *2d-matrix input* or *vector of vector input*, which means `seq_length * feature` with only one batch size.
``` julia
# seq_len = 3
# feature = 2

# vector of vector
x = [rand(Float32,2) for i = 1:3]
# or
x = rand(Float32,3,2)

# out put all hidden state, with double size.
brnn(x)
```

NeroBlackstone:
I think the correct implement is:

``` julia
function (m::Bidirectional)(x::Vector{Vector{T}}) where {T}
    vec_for = [m.forward(xi) for xi in x]
    vec_back = reverse([m.backward(xi) for xi in reverse(x)])
    return vcat.(vec_for,vec_back)
end

xb = [rand(Float32,2) for i = 1:3]
# size = 3(feature) * 10(double hidden_size)
brnn(xb)
# size = 3(feature) * 5(double hidden_size)
[rnn(xi) for xi in xb]
```

dcecchini:
Hi @NeroBlackstone , at that time I could not finish the implementation due to the difficulty to get something to work on the designed structure for RNN cells as you can see in the history above. I would be glad to work on this, but I have little time to spare.

NeroBlackstone:
> Hi @NeroBlackstone , at that time I could not finish the implementation due to the difficulty to get something to work on the designed structure for RNN cells as you can see in the history above. I would be glad to work on this, but I have little time to spare.  

Would you mind me open another PR to finish this job?

CarloLucibello:
closing as outdated

PR #1793, opened 2021-11-30
Allow NamedTuples in `batchmemaybe`

DhairyaLGandhi:
This allows named tuples to be treated exactly like tuples for `train!`. RFC for comments on the expected behaviour.

### DIFF
--- a/src/data/dataloader.jl
+++ b/src/data/dataloader.jl
@@ -13,7 +13,7 @@ struct DataLoader{D,R<:AbstractRNG}
 end
 
 """
-    Flux.DataLoader(data; batchsize=1, shuffle=false, partial=true, rng=GLOBAL_RNG)
+    Flux.DataLoader(data; batchsize = 1, shuffle = false, partial = true, rng = GLOBAL_RNG)
 
 An object that iterates over mini-batches of `data`, 
 each mini-batch containing `batchsize` observations
@@ -23,8 +23,8 @@ Takes as input a single data tensor, or a tuple (or a named tuple) of tensors.
 The last dimension in each tensor is the observation dimension, i.e. the one
 divided into mini-batches.
 
-If `shuffle=true`, it shuffles the observations each time iterations are re-started.
-If `partial=false` and the number of observations is not divisible by the batchsize, 
+If `shuffle = true`, it shuffles the observations each time iterations are re-started.
+If `partial = false` and the number of observations is not divisible by the batchsize, 
 then the last mini-batch is dropped.
 
 The original data is preserved in the `data` field of the DataLoader.
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -78,7 +78,7 @@ function stop()
 end
 
 batchmemaybe(x) = tuple(x)
-batchmemaybe(x::Tuple) = x
+batchmemaybe(x::Union{NamedTuple,Tuple}) = x
 
 """
     train!(loss, params, data, opt; cb)


ToucheSir (approved):
LGTM

CarloLucibello:
a test?

ToucheSir:
Did we not have one for `train!` already that covers this? I could've sworn I saw code using namedtuples at some point :/

ghlai9665:
Thank you @DhairyaLGandhi . Just tested on this piece of code and it works:

```
f(x) = 4x + 2
x_train, x_test = hcat(0:5...), hcat(6:10...)
y_train, y_test = f.(x_train), f.(x_test)
model = Dense(1, 1)
opt = Descent()

loss(x, y) = Flux.Losses.mse(model(x), y)

data = Flux.DataLoader((data=x_train, label=y_train), batchsize=4, shuffle=true);
params = Flux.params(model)

println("loss before training: ", loss(x_train, y_train))

for epoch in 1:10
    Flux.train!(loss, params, data, opt)
    @show loss(x_train,y_train)
end

println("loss after training: ", loss(x_train, y_train))

```

mcabbott:
Is ignoring the names of the namedtuple the expected behaviour? If it does this, you can't get them back. Whereas without it, you can choose how to handle them. 
```
julia> first(data)
(data = [1 5 3 2], label = [6 22 14 10])

julia> loss(first(data)...)  # what this PR does, discards names
76.624f0

julia> loss(; first(data)...)  # what you might expect? 
ERROR: MethodError: no method matching loss(; data=[5 2 3 4], label=[22 10 14 18])

julia> loss(nt::NamedTuple) = loss(nt...)
loss (generic function with 2 methods)

julia> for epoch in 1:10
           Flux.train!(loss, _params, data, opt)
           @show loss(x_train,y_train)
       end
loss(x_train, y_train) = 9.484035f0
loss(x_train, y_train) = 0.36190698f0
loss(x_train, y_train) = 0.16344242f0
loss(x_train, y_train) = 0.4328595f0
loss(x_train, y_train) = 0.5408588f0
loss(x_train, y_train) = 0.7149833f0
loss(x_train, y_train) = 1.3948113f0
loss(x_train, y_train) = 0.05702631f0
loss(x_train, y_train) = 0.07869858f0
loss(x_train, y_train) = 0.06192852f0

```

ToucheSir:
Come to think of it, either way is a breaking change no? All this makes me wonder if the original design behind `batchmemaybe` was too clever. Yes passing in a single value for data isn't the prettiest, but it avoids any consistency issues with splatting and function definitions can have destructured tuple parameters in Julia.

ToucheSir:
Which makes me wonder if we shouldn't change https://github.com/FluxML/Flux.jl/blob/master/src/optimise/train.jl#L110
to `loss(batchmemaybe(d))` as well. That would have the benefit of preserving NamedTuple keys as well. The downside would be that the loss function would have to handle a tuple, but `loss((x, y)) = ...` is only 2 more characters than `loss(x, y) = ...`

DhairyaLGandhi:
Enforcing a tuple argument to `loss` is too strict a requirement imo. If users need the keys, it's easy enough to send a struct instead to loss.

ToucheSir:
How is it any more strict of a requirement than forcing the loss function to be varargs? It's also one line to wrap the multi-arg loss in `Base.splat`. If anything, requiring a struct to get keys passed through seems more strict, especially when you consider that `DataLoader` [knows](https://github.com/FluxML/Flux.jl/blob/5c40161395d2b7ab48cd2c543f0a631b85f02993/src/data/dataloader.jl#L119) how to index/batch NamedTuples of arrays but not arbitrary structs.

DhairyaLGandhi:
By making it vararg, we don't make an assumption about the API `loss` has, and as such what the arguments to that function are. While we have existing patterns we follow, there isn't a strict requirement to use those at all. Similar to that, `data` isn't supposed to be a DataLoader, it can be any iterable. I can see the convenience in allowing namedtuples, but the question is whether to treat them like a tuple with keys, or an "atomic" type which cannot be deconstructed. If the argument is to splat NTs (effectively making them kwargs to `loss`) that is fine too, and `loss(; key1= ...)` can work in those cases.

ToucheSir:
If the concern is that single-arg loss functions shouldn't be forced to deal with tuples, I'd agree. Achieving that could be done by also dropping `batchmemaybe` and calling `loss(d)` directly.  

> By making it vararg, we don't make an assumption about the API `loss` has, and as such what the arguments to that function are

But we do: there's a strong assumption there that `loss` takes one arg for each item in the batch iff. the batch is a tuple. So Flux is taking the stance that batches behave differently when they are tuples. Which is not an invalid opinion, mind you, but it leads to awkwardness like this where one either has to give up some information (keys in namedtuple batches with splatting) or reimplement a chunk of what `DataLoader` handles automatically. 

Which takes me back to the first point here. If `train!` passed through `d` directly to `loss`, it would be making no assumptions about the signature of that function. Control over how to handle the batch is left completely to the user. Want to use varargs as before? Wrap with `splat`, or destructure in the signature. Need keys or want to do something with the tuple? Don't do anything fancy and the whole batch will pass through. I think it's worth pointing out that at least 90% of loss functions passed to `train!` are anonymous closures anyway (because they need to capture the model), so this would require minimal changes in user code.

CarloLucibello:
I think the whole point of having the `train!` function is convenience and conciseness for the most common use cases, therefore I wouldn't give up the convenient and commonly used 
```julia
loss(x, y) = Flux.mse(model(x), y)
train!(loss, ...)
```
for having the users splat the tuple themselves. That change would be hugely breaking and give no clear improvement. 
Regarding named tuples, since they are basically anonymous structs, I would keep handling them as we currently do and the same as any non-tuple object, to be handled by `loss(nt) = ...`. So basically between Brian's position "don't special-case anything" and Dhairya's "let's special-case both tuple and namedtuple (loosing the keys)" I prefer the middle ground where we special-case only tuples and leave things as they are.

ToucheSir:
The breaking change argument is certainly very strong, I was just ignoring it for the purpose of discussion since _every_ proposal other than keeping the status quo has been (necessarily) breaking thus far.

With that in mind, I don't see how a hypothetical
```
loss((x, y)) = Flux.mse(model(x), y)
```
Is meaningfully less convenient than `loss(x, y) = Flux.mse(model(x), y)`?

mcabbott:
I guess there are various degrees of breaking-ness. Removing the splat would affect essentially any beginner tutorial, random example: https://medium.com/coffee-in-a-klein-bottle/deep-learning-with-julia-e7f15ad5080b . And it would be difficult add a depreciation to keep things working.

The splat is part of the magic "do what I mean" of `train!`. I do agree that less such guessing would be better in general. Right now it's clearly documented to be only for tuples:
```
If d is a tuple of arguments to loss call loss(d...), else call loss(d).
```
To add more types I think you'd have to argue that this makes things more consistent with the one existing case, and easier to understand. It's pretty nice that NamedTuples work well with DataLoader to make iterators of iterators of iterators less confusing. But the fact that there are two splats possible, before or after the `;`, is an argument that this might be confusing.

CarloLucibello:
closing as stale

PR #1797, opened 2021-12-02
Mark destructure gradient test as broken

ToucheSir:
Temporary measure to unblock CI while we investigate the root cause. I think `@test_broken` is better than increasing the tolerance, as the latter makes it easier to forget about this one entirely.

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -416,7 +416,7 @@ end
       ∇m = gradient(m -> sum(m(x)), m)[1]
       p, re = destructure(m)
       ∇p = gradient(θ -> sum(re(θ)(x)), p)[1]
-      @test ∇p ≈ destructure(∇m)[1]
+      @test_broken ∇p ≈ destructure(∇m)[1]
     end
   end
 end


DhairyaLGandhi:
Happy to merge this for now. can we let go of the pr comment bot? It seems extraneous to most PRs and the links are always generated, so we aren't missing out.

ToucheSir:
@logankilpatrick is there a way to have the bot only trigger on markdown file, docstring changes and/or changes in `docs/`? That or have a manual trigger, though I think watching the docs folder should be conservative enough.

DhairyaLGandhi (approved):
bors r+

ToucheSir:
Do we have a flaky test here? https://buildkite.com/julialang/flux-dot-jl/builds/1893#efc52f50-bab7-4968-a07c-e223aa23f13f/402-581 failed on the bors run but not on the commit itself.

ToucheSir:
https://github.com/FluxML/Flux.jl/blob/master/test/cuda/layers.jl#L48 behaves differently on the PR branch vs `staging`, for  some reason?! :confused:

DhairyaLGandhi:
That test was added recently and overall seems to be reasonable. (Grouped)Conv layer tests weren't flaky earlier, although we need some cleanups both here and in nnlib.

DhairyaLGandhi:
Wondering if you could do a run with Julia 1.6 as a sanity check

ToucheSir:
I don't have a machine with 1.6 and a free GPU right now, but I can try over the weekend if someone doesn't get to this first.

DhairyaLGandhi:
I was thinking of upper boinding gha ci to 1.6 and doing a try with bors

ToucheSir:
Do we have GPU instances on GHA? AFACIT the non-CUDA grad tests have been rock stable.

DhairyaLGandhi:
Buildkite* sorry my bad.

CarloLucibello:
@ToucheSir it's ok to mark it as broken, but we should also add a test with a higher tolerance (if it passes) so that at least we are aware of further regressions

CarloLucibello:
let's merge this as quickly as possibly without bors if needed, having broken CI is pretty bad

DhairyaLGandhi:
The issue is not about a passing test with higher tolerance, since the code path should not be inducing any more error. I think it's a good idea to have a placeholder which can detect any other breakage.

DhairyaLGandhi:
bors try

DhairyaLGandhi (approved):
bors r+

ToucheSir:
@DhairyaLGandhi thanks! Do you know what (if anything) changed on CI?

PR #1798, opened 2021-12-02
Fix failing `params` doctests

ToucheSir:
```julia
julia> [[1, 2, 3], [4.0]]
2-element Vector{Vector{Float64}}:
 [1.0, 2.0, 3.0]
 [4.0]

julia> [[1, 2, 3], [4]]
2-element Vector{Vector{Int64}}:
 [1, 2, 3]
 [4]
```

### PR Checklist

- [x] Tests are added

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -69,11 +69,11 @@ BatchNorm(2, relu)  # 4 parameters, plus 4 non-trainable
 julia> params(bn)  # only the trainable parameters
 Params([Float32[0.0, 0.0], Float32[1.0, 1.0]])
 
-julia> params([1, 2, 3], [4.0])  # one or more arrays of numbers
-Params([[1, 2, 3], [4.0]])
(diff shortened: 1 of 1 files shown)


mcabbott:
Maybe only the 2nd was broken?

Could change all of these to `Flux.params` while you're there.

DhairyaLGandhi:
Best to keep changes minimal if possible.

ToucheSir:
I got a notification for a comment that doesn't appear for me on GH, but just to respond to the thread: I changed both to keep the intent of the original in showing `params(a, b) == params([a, b])`. `params([1.0, 2.0, 3.0], [4.0])` and `params([[1.0, 2.0, 3.0], [4.0]])` would've worked as well but been noisier to read.

DhairyaLGandhi (commented):
bors r+

DhairyaLGandhi (approved):
bors r+

PR #1799, opened 2021-12-03
Only add PR comment with docs build if the docs label is added 

logankilpatrick:
This should reduce the noise for non-docs PR's


DhairyaLGandhi (approved):
Thanks

CarloLucibello:
why a destructure/restructure test is failing now?

Issue #1634, opened 2021-06-26
Why does calling the gpu function not return an error when CUDA is unavailable


logankilpatrick:
Per the following code:
```Julia

julia> using CUDA

julia> CUDA.functional()
false

julia> m = Dense(10, 5)
Dense(10, 5)

julia> cpu(m)
Dense(10, 5)

julia> gpu(m)
Dense(10, 5)

julia> m
Dense(10, 5)

julia> CUDA.functional()
false
```
I would expect that when I call the GPU function on `m`, it would give an error given a GPU is not available to my understanding. Should there be a warn or print saying that it was not moved to the GPU? Otherwise, someone might think it was.

ToucheSir:
I know some code in the wild assumes `gpu` will act as a no-op when CUDA isn't present, but we could definitely `@warn`. Thoughts?

logankilpatrick:
Might be fair for more advanced users but personally I would assume things fail if I try to do that

ToucheSir:
Erroring would be a breaking change, so hence I think warning would get us more mileage sooner. It would also be good to have a `to_device` function that still does fall back to CPU if/when `gpu` no longer does so.

logankilpatrick:
Agreed!

DhairyaLGandhi:
I would rather not create fragmentation of the api. Its pretty convenient to sprinkle GPU calls without assuming the environment before hand and running on the most appropriate hardware. Its pretty common in our pitches that `gpu` is a no-op and warning too often seems like extra noise when it's totally safe to make that call. Its also a nice to have to make code portable and generic. 

`to_device` would have the potential to confuse regarding its application since it would have to do something close to the current setup anyway. `device` is only known to be a GPU in specialised circles, so I'd want to do something like `to_device(m, :cpu)` to be clearer. But now things are less easily readable. Added the fact that you'll need to make the same change `cpu -> gpu` everywhere anyway.

ToucheSir:
I don't really think we can get away from having a more explicit and complex interface eventually, especially for tasks like distributed training with multiple GPUs (where explicit device management is often necessary). `gpu` is also a bit of a poor name once you start factoring in other accelerators like TPUs. Ideally the function that offloads models to an accelerator when available should have a more accelerator-agnostic name,

DhairyaLGandhi:
[...]

PR #1800, opened 2021-12-04
Add more context on the behavior of the GPU function

logankilpatrick:
Following up the discussion here: https://github.com/FluxML/Flux.jl/issues/1634

### DIFF
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -49,7 +49,7 @@ m = fmap(cu, m)
 d(cu(rand(10)))
 ```
 
-As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing, but loading `CUDA` will cause it to move data to the GPU instead.
+As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing, meaning that you can safely call `gpu` on some data or model (as shown below), and the code will not error, regardless of whether the GPU is available or not. If the GPU loads successfully, `CUDA` will cause the data to move to the GPU from the CPU. As is shown below, this will change the type of something like a regualar array to a `CuArray`.
 
 ```julia
 julia> using Flux, CUDA
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -49,7 +49,7 @@ m = fmap(cu, m)
 d(cu(rand(10)))
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/gpu.md
```diff
@@ -49,7 +49,7 @@ m = fmap(cu, m)
 d(cu(rand(10)))
 ```
 
-As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing, but loading `CUDA` will cause it to move data to the GPU instead.
+As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing, meaning that you can safely call `gpu` on some data or model (as shown below), and the code will not error, regardless of whether the GPU is available or not. If the GPU loads successfully, `CUDA` will cause the data to move to the GPU from the CPU. As is shown below, this will change the type of something like a regualar array to a `CuArray`.
```
darsnack, line 52:
```suggestion
As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing. So, you can safely call `gpu` on some data or model (as shown below), and the code will not error, regardless of whether the GPU is available or not. If the GPU loads successfully, `CUDA` will cause the data to move to the GPU from the CPU. As is shown below, this will change the type of something like a regualar array to a `CuArray`.
```

docs/src/gpu.md
```diff
@@ -49,7 +49,7 @@ m = fmap(cu, m)
 d(cu(rand(10)))
 ```
 
-As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing, but loading `CUDA` will cause it to move data to the GPU instead.
+As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing. So, you can safely call `gpu` on some data or model (as shown below), and the code will not error, regardless of whether the GPU is available or not. If the GPU loads successfully, `CUDA` will cause the data to move to the GPU from the CPU. As is shown below, this will change the type of something like a regualar array to a `CuArray`.
```
ToucheSir, line 52:
```suggestion
As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing. So, you can safely call `gpu` on some data or model (as shown below), and the code will not error, regardless of whether the GPU is available or not. If the GPU library (CUDA.jl) loads successfully, `gpu` will move data   from the CPU to the GPU. As is shown below, this will change the type of something like a regular array to a `CuArray`.
```

ToucheSir (approved):

darsnack (approved):

darsnack:
Thanks!

bors r+

PR #1801, opened 2021-12-04
Add warning if the GPU function is called and CUDA is not available

logankilpatrick:
This should close https://github.com/FluxML/Flux.jl/issues/1634. Thanks @mcabbott for the help : )

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -184,7 +184,8 @@ function check_use_cuda()
       @warn "CUDA.jl found cuda, but did not find libcudnn. Some functionality will not be available."
     end
     if !(use_cuda[])
-      @warn "The GPU function is being called despite the GPU not being accessible. Defaulting back to the CPU." maxlog=1
+      @warn """The GPU function is being called but the GPU is not accessible. 
+               Defaulting back to the CPU. (No action is required if you want to run on the CPU).""" maxlog=1
     end
   end
 end
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -184,7 +184,7 @@ function check_use_cuda()
       @warn "CUDA.jl found cuda, but did not find libcudnn. Some functionality will not be available."
     end
     if !(use_cuda[])
-      @warn """The GPU function is being called but the GPU is not accessible. 
+      @info """The GPU function is being called but the GPU is not accessible. 
                Defaulting back to the CPU. (No action is required if you want to run on the CPU).""" maxlog=1
     end
   end
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -183,7 +183,7 @@ function check_use_cuda()
     if use_cuda[] && !CUDA.has_cudnn()
       @warn "CUDA.jl found cuda, but did not find libcudnn. Some functionality will not be available."
     end
-    if use_cuda[] == false
+    if !(use_cuda[])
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Sample output: 
```julia
julia> using Flux
[ Info: Precompiling Flux [587475ba-b771-5e3f-ad9e-33799f191a9c]

julia> m = Dense(10,5) |> gpu
┌ Warning: The GPU function is being called despite the GPU not being accessible. Defaulting back to the CPU.
└ @ Flux ~/Desktop/Flux.jl/src/functor.jl:187
Dense(10, 5)        # 55 parameters
```

DhairyaLGandhi (commented):
We typically encourage users to use the `gpu` function if they want to make their code portable both with and without GPUs. We keep the overhead of calling `gpu` low so that it doesn't hurt the performance but keeps code flexible. I'm worried this would discourage users from using `gpu` properly.

logankilpatrick:
Maybe @info would be better? I don't think it's a huge deal but it is rather unintuitive to me to call a function which doesn't actually do what I think it will do (move stuff to gpu).

src/functor.jl
```diff
@@ -183,6 +183,9 @@ function check_use_cuda()
     if use_cuda[] && !CUDA.has_cudnn()
       @warn "CUDA.jl found cuda, but did not find libcudnn. Some functionality will not be available."
     end
+    if use_cuda[] == false
```
DhairyaLGandhi, line 186:
```suggestion
    if !(use_cuda[])
```

Reads cleaner

logankilpatrick:
```Julia
julia> m = Dense(10,5) |> gpu
[ Info: The GPU function is being called despite the GPU not being accessible. Defaulting back to the CPU.
Dense(10, 5)        # 55 parameters
```
This seems a bit less aggressive?

src/functor.jl
```diff
@@ -183,6 +183,9 @@ function check_use_cuda()
     if use_cuda[] && !CUDA.has_cudnn()
       @warn "CUDA.jl found cuda, but did not find libcudnn. Some functionality will not be available."
     end
+    if !(use_cuda[])
+      @warn "The GPU function is being called despite the GPU not being accessible. Defaulting back to the CPU." maxlog=1
```
darsnack, line 187:
```suggestion
      @warn """The GPU function is being called but the GPU is not accessible. 
               Defaulting back to the CPU. (No action is required if you want to run on the CPU).""" maxlog=1
```

darsnack (commented):
I think it is better to inform the user when we silently do something unexpected, so I like this change. It only prints out once, so with the right language, I think should not discourage usage of `gpu`.

darsnack (approved):
LGTM!

ToucheSir:
bors try

ToucheSir:
(doing a bit of an experiment here to see if we can use `try` (which appears to work all the time) over `r+` (which has been consistently failing recently for some reason, along with master) until the whole Buildkite complaining on certain branches only thing is figured out).

ToucheSir (approved):
bors r+

ToucheSir:
Ok, everything but that one sometimes successful, sometimes broken ConvTranspose test works on integration. I don't think that should hold up a change as innocuous as this, so merging manually.

PR #1803, opened 2021-12-06
Fix typo in Buildkite config

ToucheSir:
Non-functional, but satisfies my inner pedant :)


ToucheSir:
Also took the opportunity to update the config name to reflect that we're no longer running on 1.6.

ToucheSir:
Superseded by #1805.

PR #1807, opened 2021-12-10
 Replace Juno with ProgressLogging

DhairyaLGandhi:
Supersedes #1643 due to a rebase bork


ToucheSir:
@pfitzseb because ProgressLogging landed https://github.com/JuliaLogging/ProgressLogging.jl/pull/35 and Juno did not, the former no longer allows many non-array types in `@progress`. Can [`_linindex`](https://github.com/JuliaLogging/ProgressLogging.jl/blob/master/src/ProgressLogging.jl#L510) be relaxed?

DhairyaLGandhi:
bump

ViralBShah:
Will be nice to remove the dependency on Juno and reduce the dependency graph of Flux.

PR #1808, opened 2021-12-12
[Do not merge] gathering MWE data for GroupedConvTranspose + selu iss…

ToucheSir:
…ues on Buildkite

For lack of being able to directly test on the workers or repro this locally, try to extract the random variables that cause the issue and use them locally.


ToucheSir:
bors try

DhairyaLGandhi:
Do we need this still?

ToucheSir:
Nope, I was looking for it but missed it skimming the PR list.

PR #1809, opened 2021-12-13
Replace unrolled `foldl` used to evaluate `Chain` with a better one

mcabbott:
~~This uses Base's `foldl` to feed data through a `Chain`, instead of a hand-written version. The reason to do so is that this greatly reduced time-to-first-gradient in https://github.com/FluxML/Zygote.jl/issues/1126 . It's possible that Flux ought to bound ChainRules with this. Or at least I should check whether Zygote's bound is high enough to guarantee it'll work. But first see if CI likes this.~~

Latest version does not call either `foldl` or `Base.afoldl`, but instead replaces one hand-written version with another, as this seems to perform better.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -44,7 +44,17 @@ end
 
 @functor Chain
 
-(c::Chain)(x) = foldl((y,f) -> f(y), (x, c.layers...))
+function (c::Chain)(x)
+  if order() < 2
+    foldl((y,f) -> f(y), (x, c.layers...))
+  else
+    # This hand-written foldl causes high latency
+    applychain(Tuple(c.layers), x)
+  end
+end
+
+applychain(::Tuple{}, x) = x
+applychain(fs::Tuple, x) = applychain(tail(fs), first(fs)(x))
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -44,7 +44,15 @@ end
 
 @functor Chain
 
-(c::Chain)(x) = foldl((y,f) -> f(y), (x, c.layers...))
+(c::Chain)(x) = applychain(c.layers, x)
+
+@generated function applychain(layers::Tuple{Vararg{<:Any,N}}, x) where {N}
+  symbols = vcat(:x, [gensym() for _ in 1:N])
+  calls = [:($(symbols[i+1]) = layers[$i]($(symbols[i]))) for i in 1:N]
+  Expr(:block, calls...)
+end
+
+applychain(layers::NamedTuple, x) = applychain(Tuple(layers), x)
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -44,17 +44,7 @@ end
 
 @functor Chain
 
-function (c::Chain)(x)
-  if order() < 2
-    foldl((y,f) -> f(y), (x, c.layers...))
-  else
-    # This hand-written foldl causes high latency
-    applychain(Tuple(c.layers), x)
-  end
-end
-
-applychain(::Tuple{}, x) = x
-applychain(fs::Tuple, x) = applychain(tail(fs), first(fs)(x))
+(c::Chain)(x) = foldl((y,f) -> f(y), (x, c.layers...))
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -27,8 +27,12 @@ julia> m2 = Chain(enc = Chain(Flux.flatten, Dense(10, 5, tanh)),
 julia> m2(x) == (m2[:dec] ∘ m2[:enc])(x)
 true
 ```
+
+For large models, there is a special type-unstable path which can reduce compilation
+times. This can be used by supplying a vector of layers `Chain([layer1, layer2, ...])`.
+This feature is somewhat experimental, beware!
 """
-struct Chain{T<:Union{Tuple, NamedTuple}}
+struct Chain{T<:Union{Tuple, NamedTuple, AbstractVector}}
   layers::T
 end
 
@@ -54,6 +58,13 @@ end
 
 applychain(layers::NamedTuple, x) = applychain(Tuple(layers), x)
 
+function applychain(layers::AbstractVector, x)  # type-unstable path, helps compile times
+  for f in layers
+    x = f(x)
+  end
+  x
+end
+
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
   Chain(NamedTuple{Base.keys(c)[i]}(Tuple(c.layers)[i]))
@@ -65,6 +76,7 @@ function Base.show(io::IO, c::Chain)
 end
 _show_layers(io, layers::Tuple) = join(io, layers, ", ")
 _show_layers(io, layers::NamedTuple) = join(io, ["$k = $v" for (k, v) in pairs(layers)], ", ")
+_show_layers(io, layers::AbstractVector) = (print(io, "["); join(io, layers, ", "); print(io, "]"))
 
 # This is a temporary and naive implementation
 # it might be replaced in the future for better performance
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -44,10 +44,7 @@ end
 
 @functor Chain
 
-applychain(::Tuple{}, x) = x
-applychain(fs::Tuple, x) = applychain(tail(fs), first(fs)(x))
-
-(c::Chain)(x) = applychain(Tuple(c.layers), x)
+(c::Chain)(x) = foldl((y,f) -> f(y), (x, c.layers...))
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
--- a/src/layers/show.jl
+++ b/src/layers/show.jl
@@ -102,7 +102,7 @@ function _big_finale(io::IO, m)
   end
 end
 
-_childarray_sum(f, x::AbstractArray) = f(x)
+_childarray_sum(f, x::AbstractArray{<:Number}) = f(x)
 _childarray_sum(f, x) = isleaf(x) ? 0 : sum(y -> _childarray_sum(f, y), Functors.children(x))
 
 # utility functions
--- a/src/layers/show.jl
+++ b/src/layers/show.jl
@@ -14,11 +14,12 @@ for T in [
 end
 
 function _big_show(io::IO, obj, indent::Int=0, name=nothing)
+  pre, post = obj isa Chain{<:AbstractVector} ? ("([", "])") : ("(", ") ")
   children = _show_children(obj)
   if all(_show_leaflike, children)
     _layer_show(io, obj, indent, name)
   else
-    println(io, " "^indent, isnothing(name) ? "" : "$name = ", nameof(typeof(obj)), "(")
+    println(io, " "^indent, isnothing(name) ? "" : "$name = ", nameof(typeof(obj)), pre)
     if obj isa Chain{<:NamedTuple} && children == getfield(obj, :layers)
       # then we insert names -- can this be done more generically? 
       for k in Base.keys(obj)
@@ -35,10 +36,10 @@ function _big_show(io::IO, obj, indent::Int=0, name=nothing)
       end
     end
     if indent == 0  # i.e. this is the outermost container
-      print(io, ")")
+      print(io, post)
       _big_finale(io, obj)
     else
-      println(io, " "^indent, "),")
+      println(io, " "^indent, post)
     end
   end
 end
@@ -90,12 +91,12 @@ function _big_finale(io::IO, m)
     noncnt = _childarray_sum(_->1, m) - length(ps)
     if noncnt > 0
       nonparam = underscorise(_childarray_sum(length, m) - sum(length, ps))
-      printstyled(io, " "^09, "# Total: ", length(ps), " trainable arrays, "; color=:light_black)
+      printstyled(io, " "^08, "# Total: ", length(ps), " trainable arrays, "; color=:light_black)
       println(io, pars, " parameters,")
       printstyled(io, " "^10, "# plus ", noncnt, " non-trainable, ", nonparam, " parameters, summarysize "; color=:light_black)
       print(io, bytes, ".")
     else
-      printstyled(io, " "^19, "# Total: ", length(ps), " arrays, "; color=:light_black)
+      printstyled(io, " "^18, "# Total: ", length(ps), " arrays, "; color=:light_black)
       print(io, pars, " parameters, ", bytes, ".")
     end
   end
--- a/src/layers/show.jl
+++ b/src/layers/show.jl
@@ -14,7 +14,7 @@ for T in [
 end
 
 function _big_show(io::IO, obj, indent::Int=0, name=nothing)
-  pre, post = obj isa Chain{<:AbstractVector} ? ("([", "])") : ("(", ") ")
+  pre, post = obj isa Chain{<:AbstractVector} ? ("([", "])") : ("(", ")")
   children = _show_children(obj)
   if all(_show_leaflike, children)
     _layer_show(io, obj, indent, name)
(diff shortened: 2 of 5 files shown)


src/layers/basic.jl
```diff
@@ -43,10 +43,7 @@ end
 
 functor(::Type{<:Chain}, c) = c.layers, ls -> Chain(ls...)
 
-applychain(::Tuple{}, x) = x
-applychain(fs::Tuple, x) = applychain(tail(fs), first(fs)(x))
```
ToucheSir, line 47: We should keep these lines, if just to not break compat.

ToucheSir:
Would there be any value in defining a `foldl` wrapper or `foldl`-like function that has `init` as a positional arg? That would allow for creating an equivalent (possibly simplified) version of https://github.com/JuliaDiff/ChainRules.jl/blob/a75193768775975fac5578c89d1e5f50d7f358c2/src/rulesets/Base/mapreduce.jl#L342-L377.

mcabbott:
The obvious candidate there is `Base.mapfoldr_impl(f, op, init, itr)`, which is 2 hops down from `foldl`. I've thought about it but haven't quite got there yet.

Agree the unstable path would be nice to have, too. This is just the minimum-effort PR to get started.

src/layers/basic.jl
mcabbott: I guess my take is that these are unexported implementation details, which you use at your own risk? And that this PR would only be part of v0.13, a breaking release.

CarloLucibello:
Should we have some second derivative test since you expressed concerns about it in  FluxML/Zygote.jl#1126 ?

ToucheSir:
That would be great, thanks for bringing it up.

src/layers/basic.jl
```diff
@@ -43,11 +43,18 @@ end
 
 functor(::Type{<:Chain}, c) = c.layers, ls -> Chain(ls...)
 
+function (c::Chain)(x)
+  if order() < 2
+    foldl((y,f) -> f(y), (x, c.layers...))
```
ToucheSir, line 48: Would it be too cheeky to do
```suggestion
    Base.afoldl(|>, x, c.layers...)
```
ref. https://github.com/JuliaLang/julia/blob/dba8f0344ab6f1a63ea4f7928610feb35f014d4f/base/operators.jl#L592.

mcabbott: This was slow when I tried it :( although I did not think of `|>`

mcabbott: Tidied up, here are times of some options on some very small networks:
```julia
julia> using Flux

julia> m1 = Chain(x->vcat(x,x), Dense(2,2,relu));

julia> m2 = Chain(Dense(20,30,relu), Dense(30,10), softmax);

# Tagged version

julia> @btime gradient(sum∘$m1, 3.14)
  16.584 μs (253 allocations: 9.52 KiB)
(0.33772213757038116,)

julia> @btime gradient(sum∘$m2, rand(20,20));
  42.792 μs (353 allocations: 68.00 KiB)

# foldl

julia> @eval Flux (c::Chain)(x) = foldl(|>, (x, c.layers...));

julia> Flux.Zygote.refresh()

julia> @btime gradient(sum∘$m1, 3.14)
  18.625 μs (277 allocations: 11.62 KiB)
(0.33772213757038116,)

julia> @btime gradient(sum∘$m2, rand(20,20));
  45.500 μs (385 allocations: 72.03 KiB)

# afoldl

julia> @eval Flux (c::Chain)(x) = Base.afoldl(|>, x, c.layers...);

julia> Flux.Zygote.refresh()

julia> @btime gradient(sum∘$m1, 3.14)
  84.167 μs (496 allocations: 19.03 KiB)
(0.33772213757038116,)

julia> @btime gradient(sum∘$m2, rand(20,20));
  117.917 μs (614 allocations: 77.62 KiB)

# Unstable for loop

julia> @eval Flux begin
       function (c::Chain)(x)
         y = identity(x)
         for f in c.layers
           y = f(y)
         end
         y
       end
       end

julia> Flux.Zygote.refresh()

julia> @btime gradient(sum∘$m1, 3.14)
  25.250 μs (357 allocations: 13.33 KiB)
(0.33772213757038116,)

julia> @btime gradient(sum∘$m2, rand(20,20));
  58.792 μs (512 allocations: 72.89 KiB)
  ```

Comment (ToucheSir) [reply src/layers/basic.jl:48]: Thanks. Looks like `afoldl` would need an `rrule` like `foldl` to be remotely competitive.

Comment (mcabbott) [reply src/layers/basic.jl:48]: I guess it's a sad reflection on Zygote that something so simple should benefit from a complicated hand-written rule. Diffractor:
```julia
julia> fs = (sin, cos, x -> vcat(x,x), sum, exp);

julia> @btime Zygote.gradient(x -> foldl(|>, (x, $fs...)), 3)
  6.900 μs (130 allocations: 3.88 KiB)
(2.017262629514019,)

julia> @btime Zygote.gradient(x -> Base.afoldl(|>, x, $fs...), 3)
  72.083 μs (306 allocations: 11.81 KiB)
(2.017262629514019,)

julia> @btime Diffractor.gradient(x -> foldl(|>, (x, $fs...)), 3)
  9.375 μs (205 allocations: 6.52 KiB)
(2.017262629514019,)

julia> @btime Diffractor.gradient(x -> Base.afoldl(|>, x, $fs...), 3)
  3.812 μs (90 allocations: 5.56 KiB)
(2.017262629514019,)
```

claforte: I haven't read the whole Flux documentation recently, but considering this isn't the first case of catastrophic perf degradation related to Zygote, I think it would be great if there were a simple, documented way to switch the AD Flux uses by default...

AriMKatz: @claforte Yea. Zygote unfortunately is just intrinsically slow for many use cases  because of its design. See [here](https://www.stochasticlifestyle.com/engineering-trade-offs-in-automatic-differentiation-from-tensorflow-and-pytorch-to-jax-and-julia/) for more details and how this is being solved.

mcabbott: IIRC the major hurdle for Flux is moving away from the implicit `Params` story, which Diffractor does not plan to support. Gradients themselves do run, although it's easy to find things which break.
```
julia> m = Chain(Dense(2,3,sigmoid),sum);

julia> Zygote.gradient(m -> m([0.1, 0.2]), m)
((layers = ((weight = [0.024973915178896296 0.04994783035779259; 0.024897234881931463 0.049794469763862925; 0.02492858706809792 0.04985717413619584], bias = Float32[0.24973916, 0.24897234, 0.24928588], σ = nothing), nothing),),)

julia> Diffractor.gradient(m -> m([0.1, 0.2]), m)
(Tangent{Chain{Tuple{Dense{typeof(σ), Matrix{Float32}, Vector{Float32}}, typeof(sum)}}}(layers = (Tangent{Dense{typeof(σ), Matrix{Float32}, Vector{Float32}}}(σ = ChainRulesCore.ZeroTangent(), bias = [0.24973915178896294, 0.24897234881931463, 0.2492858706809792], weight = [0.024973915178896296 0.04994783035779259; 0.024897234881931463 0.049794469763862925; 0.02492858706809792 0.04985717413619584]), ChainRulesCore.ZeroTangent()),),)
```
 I also don't know whether benchmarks like the one above are fair to Diffractor, or if major improvements can be expected soon. Tracker seems shockingly quick there. Maybe it was a mistake for Flux to move away from that.

ToucheSir: https://github.com/AStupidBear/TrackerFlux.jl hasn't been updated in a few months, but could potentially be revitalized along with Tracker itself (including a ChainRules retrofit if you want to be fancy).

Support for explicit gradients is mostly tied up with [Optimisers.jl](https://github.com/FluxML/Optimisers.jl) development. The other big prereq for Diffractor support would be porting normalization layer adjoints to `rrule`s (ideally by moving the underlying functions into NNlib).

mcabbott: Ok. From a look at its readme, TrackerFlux.jl stores the parameters & gradients together, as Flux used to. So the way you use it looks like the implicit parameter story. 

ReverseDiff seems to get more maintenance, and I think now uses ChainRules. Neither Tracker nor ReverseDiff seem happy with explicit non-array inputs. Although perhaps these could be overloaded to work? If so, it might be possible for Flux+Optimisers to be completely agnostic about AD, so long as `gradient(m -> loss(m,x,y), model)` returns a suitable nested structure.
```
julia> Tracker.gradient(x -> sum(x.a), (a=[1,2], b=[3,4]))
ERROR: MethodError: no method matching param(::NamedTuple{(:a, :b), Tuple{Vector{Int64}, Vector{Int64}}})

julia> ReverseDiff.gradient(x -> sum(x.a), (a=[1,2], b=[3,4]))
ERROR: MethodError: no method matching ReverseDiff.GradientConfig(::NamedTuple{(:a, :b), Tuple{Vector{Int64}, Vector{Int64}}})
```

AriMKatz: There's also https://github.com/dfdx/Yota.jl which seems like it could be potential pretty fast

ToucheSir: AD agnosticism is definitely a goal if the Optimisers.jl transition, yeah. Currently it doesn't handle Tangents, but we could steal some code from Zygote for that.

mcabbott: Oh right, indeed. This does understand structs, but fails on my `foldl` tests above, not sure why
```julia
julia> Yota.grad(x -> sum(x.a), (a=[1,2], b=[3,4]))
(3, (ChainRulesCore.ZeroTangent(), Tangent{NamedTuple{(:a, :b), Tuple{Vector{Int64}, Vector{Int64}}}}(a = [1, 1],)))

julia> Yota.grad(x -> foldl(|>, (x, fs2...)), 3.0)
ERROR: No deriative rule found for op %7 = foldl(|>, %5)::Float64, try defining it using 
	ChainRulesCore.rrule(::typeof(foldl), ::typeof(|>), ::Tuple{Float64, typeof(sin), typeof(cos), var"#13#14", typeof(sum), typeof(exp)}) = ...

julia> Yota.grad(x -> Base.afoldl(|>, x, fs2...), 3.0)
ERROR: MethodError: no method matching length(::typeof(sin))
```
Cc @dfdx in case he'd like to join this hallway discussion.

DhairyaLGandhi: The performance is really only due to zygote unrolling chains, which isn't per se necessary. It also seems `foldl` wouldn't work generically, esp when composing layers chains of chains etc.

DhairyaLGandhi (commented):
We should simplify this, because there are different rules now how composed chains would work.

src/layers/basic.jl
dfdx: @mcabbott The rule for `foldl` [exists][1] in ChainRules, so it must be a bug. I'm looking at it. Thanks for pinging me!

[1]: https://github.com/JuliaDiff/ChainRules.jl/blob/a75193768775975fac5578c89d1e5f50d7f358c2/src/rulesets/Base/mapreduce.jl#L331-L377

mcabbott: Latest version of this PR replaces `foldl` with `@generated` applychain. Applied to the tiny examples above https://github.com/FluxML/Flux.jl/pull/1809#discussion_r777762381 this is faster. And it has shorter compile times (shown elsewhere).
```julia
julia> using Flux

julia> m1 = Chain(x->vcat(x,x), Dense(2,2,relu));

julia> m2 = Chain(Dense(20,30,relu), Dense(30,10), softmax);

julia> @btime gradient(sum∘$m1, 3.14)
  min 14.917 μs, mean 16.347 μs (246 allocations, 9.42 KiB)
(0.310660183429718,)

julia> @btime gradient(sum∘$m2, rand(20,20));
  min 41.417 μs, mean 50.995 μs (353 allocations, 68.27 KiB)

julia> m1 = Chain([x->vcat(x,x), Dense(2,2,relu)]);

julia> m2 = Chain([Dense(20,30,relu), Dense(30,10), softmax]);

julia> @btime gradient(sum∘$m1, 3.14)
  min 34.333 μs, mean 35.980 μs (459 allocations, 15.06 KiB)
(0.0,)

julia> @btime gradient(sum∘$m2, rand(20,20));
  min 73.500 μs, mean 85.917 μs (681 allocations, 76.66 KiB)
```
The last times here show a chain with layers stored as a vector. This is not type-stable, but helps with compile times. On very small networks like this, it is significantly slower, so it probably cannot be the default.

mcabbott:
Latest version of this replaces `foldl` with an `@generated` function which explicitly unrolls. This is not only faster to compile, it is also faster to run. With the example from https://github.com/FluxML/Zygote.jl/issues/1126#issue-1065022370, the final gradient is 30% faster, and the first one is 2x faster:
```julia
julia> @time loss(lr_images);  # first Flux.Chain call, cold start, no Zygote
  4.490344 seconds (30.96 M allocations: 1.634 GiB, 7.28% gc time, 99.81% compilation time)  # Flux v0.12.8
  3.608894 seconds (34.26 M allocations: 1.801 GiB, 8.63% gc time, 99.84% compilation time)  # @generated
  3.699612 seconds (34.25 M allocations: 1.800 GiB, 8.80% gc time, 99.82% compilation time)

julia> @time loss(lr_images);
  0.005024 seconds (2.46 k allocations: 242.812 KiB)

julia> @btime loss($lr_images);
  min 7.202 ms, mean 7.648 ms (2443 allocations, 256.98 KiB)  # Flux v0.12.8
  min 5.533 ms, mean 5.696 ms (2461 allocations, 242.78 KiB)  # @generated
  min 5.639 ms, mean 5.881 ms (2368 allocations, 233.55 KiB)  # Vector chain

julia> loss_grad(x, ps) = gradient(() -> loss(x), ps);

julia> ps = Flux.params(model);

julia> @time loss_grad(lr_images, ps);  # first Zygote.gradient, cold start
 62.895168 seconds (67.77 M allocations: 3.523 GiB, 0.99% gc time, 99.92% compilation time)  # Flux v0.12.8
 29.646482 seconds (64.41 M allocations: 3.373 GiB, 2.02% gc time, 99.88% compilation time)  # @generated
 13.677340 seconds (63.15 M allocations: 3.308 GiB, 3.99% gc time, 99.74% compilation time)  # Vector chain

julia> @time loss_grad(lr_images, ps);  # second
  0.028015 seconds (12.95 k allocations: 2.275 MiB)

julia> @btime $loss_grad($lr_images, $ps);
  min 37.933 ms, mean 38.574 ms (13052 allocations, 2.93 MiB)  # Flux v0.12.8
  min 29.283 ms, mean 29.730 ms (12941 allocations, 2.28 MiB)  # @generated
  min 30.516 ms, mean 30.957 ms (13436 allocations, 1.62 MiB)  # Vector chain

julia> length(model.layers[1].layers.layers)  # count of resblock(channels) layers
15
```
Also included in this comparison is a Chain backed by a Vector. This is type-unstable, and not as fast for very small networks. But for larger ones, it is no slower, and improves compile time by another factor of 2. It's included in this PR for now, but obviously needs tests if it is to stay.

src/layers/basic.jl
mcabbott: Latest version again has such a function.

src/layers/basic.jl
```diff
@@ -59,6 +76,7 @@ function Base.show(io::IO, c::Chain)
 end
 _show_layers(io, layers::Tuple) = join(io, layers, ", ")
 _show_layers(io, layers::NamedTuple) = join(io, ["$k = $v" for (k, v) in pairs(layers)], ", ")
+_show_layers(io, layers::AbstractVector) = (print(io, "["); join(io, layers, ", "); print(io, "]"))
 
 # This is a temporary and naive implementation
 # it might be replaced in the future for better performance
```
mcabbott, line 82: BTW, in addition to a hand-written `foldl`, the function `Flux.activations` is just `accumulate(|>, m1.layers; init=x1)`. Since we don't support Julia < 1.5, we could just replace it.

DhairyaLGandhi:
For sciml we really cannot have regressions in higher order chains or regressions in smaller chains.

src/layers/basic.jl
ToucheSir: I thought `accumulate` would face the same issue as `foldl`, namely that the `rrule` doesn't consider `init`? This PR need not be concerned with `activations` either way, we can kick that can down the road until `rrules` are tweaked or someone complains about performance.

mcabbott: Indeed, you could likewise do `accumulate(|>, (x, m.layers...))`. But happy to leave it alone for now.

src/layers/basic.jl
dfdx: FYI: latest master of Yota now correctly handles `foldl`. I haven't tested it with the latest version of this PR though since I'm currently busy speeding up the underlying tracer faster, so I'm going to wait till the PR is done and catch up later.

ToucheSir:
@ChrisRackauckas would you be able to offer a thumbs up/down on whether this is something SciML could tolerate? Also @pxl-th I don't remember if you've tested this, but if not it may be interesting for your benchmark in https://github.com/FluxML/Zygote.jl/issues/1119.

ChrisRackauckas:
This is fine. We use `FastChain` for small chains anyways, and resort to Flux when things get big, so we should be fine. The small end gap will grow after https://github.com/SciML/DiffEqFlux.jl/pull/671, and I'm not sure what to do after that divergence but it'll be an interesting discussion to have.

src/layers/show.jl
```diff
@@ -35,10 +36,10 @@ function _big_show(io::IO, obj, indent::Int=0, name=nothing)
       end
     end
     if indent == 0  # i.e. this is the outermost container
-      print(io, ")")
+      print(io, post)
       _big_finale(io, obj)
     else
-      println(io, " "^indent, "),")
+      println(io, " "^indent, post)
```
ToucheSir, line 42: This appears to be the culprit for the doctest failure.

mcabbott: Oh I see, a missing comma, will fix.

mcabbott: Maybe #1862 should go first, and this after.

mcabbott:
Good to go?

ToucheSir (approved):

Issue #1818, opened 2021-12-28
Optimisers epsilon


cossio:
Why is epsilon defined as a global constant like this, https://github.com/FluxML/Flux.jl/blob/fe803a1a706eb7ecbbf46f7c7e6c20b01847146d/src/optimise/optimisers.jl#L6 ?

Given that it affects the behavior of the optimisers, I think it should be a parameter of each optimiser.

This would also allow users to modify its value, if they need to.

If people agree I can make a PR that makes epsilon a property of each optimiser.

DhairyaLGandhi:
we could make it so it's a parameter. Since one is likely to use one (maybe two?) Optimisers for most cases it's not likely to interfere much but makes it so our api expands for every optimiser.

Are there cases where we would need to set `eps` per optimiser? I can maybe understand if it causes unexpected promotion with different number types.

cossio:
I have seen some (rare) cases where people use a different value of epsilon and this helps with training stability.

Please see https://github.com/FluxML/Flux.jl/pull/1819.

darsnack:
It is a parameter in Optimisers.jl already. We could add it to current Flux.

Issue #1804, opened 2021-12-08
Buildkite CI failures with grad test of `ConvTranspose` + `selu`


ToucheSir:
One example: https://buildkite.com/julialang/flux-dot-jl/builds/1914#c62d9761-ab7f-415a-b995-51552eb2b1e5

I could not repro this locally on 2 separate GPU machines, master and 2 threads (same number Buildkite uses). Oddly, it's succeded twice (once on [trying](https://buildkite.com/julialang/flux-dot-jl/builds/1906) and once on [staging](https://buildkite.com/julialang/flux-dot-jl/builds/1912)), but fails with the same result every time, which makes me suspect some environmental discrepancy between workers.

ToucheSir:
Despite pulling the exact inputs and weights from failing buildkite runs, I was not able to replicate this locally. If someone has a znver2 machine they can test on, everything needed should  be in https://gist.github.com/ToucheSir/32fd6688d3932c9f498c78a42a0ea017. @DhairyaLGandhi and possibly @maleadt, is there any way to replicate the CI environment and run tests against that? I don't think it's necessary to have a GPU attached, because the inconsistencies are coming from the CPU calculation.

ToucheSir:
I could get the same summed result by exporting the pre-selu output and loading that locally, but still have not been able to generate a similar output and wasn't able to coax a container image into downloading the same dep versions. What's confusing is that the tests for ConvTranspose + identity activation _do_ pass on CI, yet the differing output for the selu version is happening before the activation function is applied. Either way, #1822 is now up for a stopgap until some brave soul decides to revisit this in the future.

ToucheSir:
Fixed by #1836, ref. https://github.com/FluxML/Flux.jl/pull/1836#issuecomment-1013745520.

PR #1819, opened 2021-12-28
make eps a parameter of optimisers

cossio:
Makes epsilon a parameter of each optimiser. Closes https://github.com/FluxML/Flux.jl/issues/1818.

### PR Checklist

- [X] Tests are added

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -133,16 +133,17 @@ opt = RMSProp(0.002, 0.95)
 mutable struct RMSProp <: AbstractOptimiser
   eta::Float64
   rho::Float64
+  epsilon::Float64
   acc::IdDict
 end
 
-RMSProp(η = 0.001, ρ = 0.9) = RMSProp(η, ρ, IdDict())
+RMSProp(η = 0.001, ρ = 0.9, ϵ = ϵ) = RMSProp(η, ρ, ϵ, IdDict())
 
 function apply!(o::RMSProp, x, Δ)
   η, ρ = o.eta, o.rho
   acc = get!(() -> zero(x), o.acc, x)::typeof(x)
   @. acc = ρ * acc + (1 - ρ) * Δ * conj(Δ)
-  @. Δ *= η / (√acc + ϵ)
+  @. Δ *= η / (√acc + o.epsilon)
 end
 
 """
@@ -166,10 +167,11 @@ opt = ADAM(0.001, (0.9, 0.8))
 mutable struct ADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
+  epsilon::Float64
   state::IdDict
 end
 
-ADAM(η = 0.001, β = (0.9, 0.999)) = ADAM(η, β, IdDict())
+ADAM(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = ADAM(η, β, ϵ, IdDict())
 
 function apply!(o::ADAM, x, Δ)
   η, β = o.eta, o.beta
@@ -180,7 +182,7 @@ function apply!(o::ADAM, x, Δ)
 
   @. mt = β[1] * mt + (1 - β[1]) * Δ
   @. vt = β[2] * vt + (1 - β[2]) * Δ * conj(Δ)
-  @. Δ =  mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + ϵ) * η
+  @. Δ =  mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + o.epsilon) * η
   βp .= βp .* β
 
   return Δ
@@ -207,10 +209,11 @@ opt = RADAM(0.001, (0.9, 0.8))
 mutable struct RADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
+  epsilon::Float64
   state::IdDict
 end
 
-RADAM(η = 0.001, β = (0.9, 0.999)) = RADAM(η, β, IdDict())
+RADAM(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = RADAM(η, β, ϵ, IdDict())
 
 function apply!(o::RADAM, x, Δ)
   η, β = o.eta, o.beta
@@ -225,7 +228,7 @@ function apply!(o::RADAM, x, Δ)
   ρ = ρ∞ - 2t[] * βp[2] / (1 - βp[2])
   if ρ > 4
     r = sqrt((ρ-4)*(ρ-2)*ρ∞/((ρ∞-4)*(ρ∞-2)*ρ))
-    @. Δ =  mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + ϵ) * η * r
+    @. Δ =  mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + o.epsilon) * η * r
   else
     @. Δ =  mt / (1 - βp[1]) * η
   end
@@ -256,10 +259,11 @@ opt = AdaMax(0.001, (0.9, 0.995))
 mutable struct AdaMax <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
+  epsilon::Float64
   state::IdDict
 end
 
-AdaMax(η = 0.001, β = (0.9, 0.999)) = AdaMax(η, β, IdDict())
+AdaMax(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = AdaMax(η, β, ϵ, IdDict())
 
 function apply!(o::AdaMax, x, Δ)
   η, β = o.eta, o.beta
@@ -270,7 +274,7 @@ function apply!(o::AdaMax, x, Δ)
 
   @. mt = β[1] * mt + (1 - β[1]) * Δ
   @. ut = max(β[2] * ut, abs(Δ))
-  @. Δ = (η/(1 - βp[1])) * mt/(ut + ϵ)
+  @. Δ = (η/(1 - βp[1])) * mt/(ut + o.epsilon)
   βp .= βp .* β
 
   return Δ
@@ -298,10 +302,11 @@ opt = OADAM(0.001, (0.9, 0.995))
 mutable struct OADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
+  epsilon::Float64
   state::IdDict
 end
 
-OADAM(η = 0.001, β = (0.5, 0.9)) = OADAM(η, β, IdDict())
+OADAM(η = 0.001, β = (0.5, 0.9), ϵ = ϵ) = OADAM(η, β, ϵ, IdDict())
 
 function apply!(o::OADAM, x, Δ)
   η, β = o.eta, o.beta
@@ -313,7 +318,7 @@ function apply!(o::OADAM, x, Δ)
   @. mt = β[1] * mt + (1 - β[1]) * Δ
   @. vt = β[2] * vt + (1 - β[2]) * Δ * conj(Δ)
   @. Δ = -Δ_
-  @. Δ_ = η * mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + ϵ)
+  @. Δ_ = η * mt / (1 - βp[1]) / (√(vt / (1 - βp[2])) + o.epsilon)
   @. Δ += 2Δ_
   βp .= βp .* β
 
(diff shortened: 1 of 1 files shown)


src/optimise/optimisers.jl
```diff
@@ -133,16 +133,17 @@ opt = RMSProp(0.002, 0.95)
 mutable struct RMSProp <: AbstractOptimiser
   eta::Float64
   rho::Float64
+  eps::Float64
   acc::IdDict
 end
 
-RMSProp(η = 0.001, ρ = 0.9) = RMSProp(η, ρ, IdDict())
+RMSProp(η = 0.001, ρ = 0.9, ϵ = ϵ) = RMSProp(η, ρ, ϵ, IdDict())
```
darsnack, line 140: This is what we use in Optimisers.jl.

```suggestion
RMSProp(η = 0.001, ρ = 0.9, ϵ = eps(typeof(η))) = RMSProp(η, ρ, ϵ, IdDict())
```

src/optimise/optimisers.jl
```diff
@@ -166,10 +167,11 @@ opt = ADAM(0.001, (0.9, 0.8))
 mutable struct ADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
+  eps::Float64
   state::IdDict
 end
 
-ADAM(η = 0.001, β = (0.9, 0.999)) = ADAM(η, β, IdDict())
+ADAM(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = ADAM(η, β, ϵ, IdDict())
```
darsnack, line 174:
```suggestion
ADAM(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = ADAM(η, β, ϵ, IdDict())
```

src/optimise/optimisers.jl
```diff
@@ -207,10 +209,11 @@ opt = RADAM(0.001, (0.9, 0.8))
 mutable struct RADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
+  eps::Float64
   state::IdDict
 end
 
-RADAM(η = 0.001, β = (0.9, 0.999)) = RADAM(η, β, IdDict())
+RADAM(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = RADAM(η, β, ϵ, IdDict())
```
darsnack, line 216:
```suggestion
RADAM(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = RADAM(η, β, ϵ, IdDict())
```

src/optimise/optimisers.jl
```diff
@@ -256,10 +259,11 @@ opt = AdaMax(0.001, (0.9, 0.995))
 mutable struct AdaMax <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
+  eps::Float64
   state::IdDict
 end
 
-AdaMax(η = 0.001, β = (0.9, 0.999)) = AdaMax(η, β, IdDict())
+AdaMax(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = AdaMax(η, β, ϵ, IdDict())
```
darsnack, line 266:
```suggestion
AdaMax(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = AdaMax(η, β, ϵ, IdDict())
```

src/optimise/optimisers.jl
```diff
@@ -298,10 +302,11 @@ opt = OADAM(0.001, (0.9, 0.995))
 mutable struct OADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
+  eps::Float64
   state::IdDict
 end
 
-OADAM(η = 0.001, β = (0.5, 0.9)) = OADAM(η, β, IdDict())
+OADAM(η = 0.001, β = (0.5, 0.9), ϵ = ϵ) = OADAM(η, β, ϵ, IdDict())
```
darsnack, line 309:
```suggestion
OADAM(η = 0.001, β = (0.5, 0.9), ϵ = eps(typeof(η))) = OADAM(η, β, ϵ, IdDict())
```

src/optimise/optimisers.jl
```diff
@@ -340,16 +345,17 @@ opt = ADAGrad(0.001)
 """
 mutable struct ADAGrad <: AbstractOptimiser
   eta::Float64
+  eps::Float64
   acc::IdDict
 end
 
-ADAGrad(η = 0.1) = ADAGrad(η, IdDict())
+ADAGrad(η = 0.1, ϵ = ϵ) = ADAGrad(η, ϵ, IdDict())
```
darsnack, line 352:
```suggestion
ADAGrad(η = 0.1, ϵ = eps(typeof(η))) = ADAGrad(η, ϵ, IdDict())
```

src/optimise/optimisers.jl
```diff
@@ -371,18 +377,19 @@ opt = ADADelta(0.89)
 """
 mutable struct ADADelta <: AbstractOptimiser
   rho::Float64
+  eps::Float64
   state::IdDict
 end
 
-ADADelta(ρ = 0.9) = ADADelta(ρ, IdDict())
+ADADelta(ρ = 0.9, ϵ = ϵ) = ADADelta(ρ, ϵ, IdDict())
```
darsnack, line 384:
```suggestion
ADADelta(ρ = 0.9, ϵ = eps(typeof(ρ))) = ADADelta(ρ, ϵ, IdDict())
```

src/optimise/optimisers.jl
```diff
@@ -409,22 +416,23 @@ opt = AMSGrad(0.001, (0.89, 0.995))
 mutable struct AMSGrad <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64, Float64}
+  eps::Float64
   state::IdDict
 end
 
-AMSGrad(η = 0.001, β = (0.9, 0.999)) = AMSGrad(η, β, IdDict())
+AMSGrad(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = AMSGrad(η, β, ϵ, IdDict())
```
darsnack, line 423:
```suggestion
AMSGrad(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = AMSGrad(η, β, ϵ, IdDict())
```

src/optimise/optimisers.jl
```diff
@@ -449,10 +457,11 @@ opt = NADAM(0.002, (0.89, 0.995))
 mutable struct NADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64, Float64}
+  eps::Float64
   state::IdDict
 end
 
-NADAM(η = 0.001, β = (0.9, 0.999)) = NADAM(η, β, IdDict())
+NADAM(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = NADAM(η, β, ϵ, IdDict())
```
darsnack, line 464:
```suggestion
NADAM(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = NADAM(η, β, ϵ, IdDict())
```

src/optimise/optimisers.jl
```diff
@@ -515,17 +524,18 @@ opt = AdaBelief(0.001, (0.9, 0.8))
 mutable struct AdaBelief
   eta::Float64
   beta::Tuple{Float64,Float64}
+  eps::Float64
   state::IdDict
 end
 
-AdaBelief(η = 0.001, β = (0.9, 0.999)) = AdaBelief(η, β, IdDict())
+AdaBelief(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = AdaBelief(η, β, ϵ, IdDict())
```
darsnack, line 531:
```suggestion
AdaBelief(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = AdaBelief(η, β, ϵ, IdDict())
```

darsnack (changes requested):
Only other comment is to call the field `epsilon` since that's the name we chose in Optimisers.jl (to avoid confusion with the function of the same name).

src/optimise/optimisers.jl
cossio: I didn't want to change the value of `eps` Flux is currently using, so as to not change behavior.
So I continue to use to the value of `1e-8` here.
But I did use the type of eta as you suggest.

cossio:
@darsnack  I didn't want to change the value of eps Flux is currently using, so as to not change behavior. So I prefer not to use `eps(typeof(eta))` as you suggest. I continue to use to the value of 1e-8 here.
But I did use the type of eta as you suggest.

Using `eps(typeof(eta))` is technically a breaking change so it could be adressed in a separate PR.

darsnack (approved):
Fair enough, I didn't realize the type was hard coded to `Float64`.

cossio:
Merge?

darsnack:
Failure seems related to #1808 (@ToucheSir)?

ToucheSir:
Yes, https://github.com/FluxML/Flux.jl/issues/1804 specifically. I still haven't been able to repro the issue locally (will try replicating the Buildkite env when I have time in a couple of days). Feel free to `@test_skip` that one if you need any PRs in sooner, otherwise help would be much appreciated on figuring out the CI failure.

cossio:
Merge this? buildkite failure looks unrelated to this PR, and all other tests pass.

darsnack:
@CarloLucibello @DhairyaLGandhi I think an org owner needs to merge.

DhairyaLGandhi:
This was a bit too wide a change for the api and can break a bunch of code...

darsnack:
> can break a bunch of code

How so? It adds an optional positional argument that has the same default value as before.

DhairyaLGandhi:
It adds a field to the optimisers, any code using the default "struct" constructors would break.

ToucheSir:
Sorry, I should've caught that after https://github.com/FluxML/Flux.jl/pull/1778. Shall we quarantine both off for 0.13?

DhairyaLGandhi:
#1778 is fine still because the disruption is limited to one type with a clear way of making a case for a non breaking type.

cossio:
Sorry I didn't realize that. To be clear, what you mean is that a call like:

```
RMSProp(η, ρ, IdDict())
```

no longer works, because it now needs to be replaced by:

```
RMSProp(η, ρ, ϵ, IdDict())
```

right?

A possible solution is to make another PR (overwriting this one), which makes ϵ an optional keyword argument. Would that work?

```
RMSProp(η, ρ, IdDict(); ϵ = ϵ)
```

DhairyaLGandhi:
Right

ToucheSir:
Well, I'm still not sure if adding struct fields constitutes a SemVer break, as (de)serialization is affected. If anyone is in the know about this, please do leave a comment.

darsnack:
@DhairyaLGandhi thanks for the explanation, sorry I missed that.

How frequently is the `IdDict` manually being passed in though? I'd rather have a breaking change for v0.13 than complicate the API by adding a keyword just to make it not breaking.

@ToucheSir adding fields is not breaking, just that the constructors conflict.

DhairyaLGandhi:
It's got less to do with frequency and more about necessity and API compactness. In cases where we need literal copies of gradients (for synchronisation for example) over different models it's helpful to store states and gradients as the same type for simplicity, and making users add in an epsilon fiels (especially when not needed in the optimisation step) is unnecessary.

darsnack:
I get that, but I'm saying if the frequency is low enough (i.e. a handful of users), then it's fair to ask that handful to add the new field into their code on a breaking release. Anyone manually passing in the `IdDict` is already passing in a bunch of parameters other than the LR. This seems reasonable to me in order to not have a global constant in the package affecting the behavior of many functions.

cossio:
> How frequently is the `IdDict` manually being passed in though? I'd rather have a breaking change for v0.13 than complicate the API by adding a keyword just to make it not breaking.

I don't see why making epsilon a keyword would complicate the API? It would be an optional keyword that most users need not even notice.

DhairyaLGandhi:
Passing in Dict explicitly is more than passing a bunch of arguments though, it is arguably more related to the use case of the optimisers than the value of an epsilon. It's already very rare to need to change epsilon.

darsnack:
I meant that if you are manually passing in the dict, then you are already writing `ADAM(lr, (0.9, 0.999), IdDict(...))`. No one changes the `(0.9, 0.999)` either, but these are extra arguments the user is copy-pasting in order to pass in the `IdDict`. So having to update the code to `ADAM(lr, (0.9, 0.999), 1e-8, IdDict(...))` on a breaking release seems okay to me.

DhairyaLGandhi:
But even a long list of arguments to the constructor maintains a rational order (going by what is likely to be used and requires changing for the majority of use cases), which epsilon in the position in the pr does not since it comes before the storage. It should be after.

darsnack:
Sure, a PR to fix the order is okay for me.

ToucheSir:
I would favour grouping the hparams together. Perhaps if the IdDict was the first parameter in the default constructor?

cossio:
I'd prefer
```
RMSProp(η, ρ, IdDict(); ϵ = ϵ)
```
over
```
RMSProp(η, ρ, IdDict(), ϵ)
```
In fact, having all hyper-params as keyword arguments wouldn't be bad I think,
```
RMSProp(IdDict(); η=η, ρ=ρ, ϵ=ϵ)
```
but that's another matter.

ToucheSir:
For `ADAM` at least, I'd argue `β` is not sufficiently more worthy of tuning than `ϵ` as to warrant being before the `IdDict` than after. Therein lies the challenge though: can we agree on which hyperparams are "privileged" and which are relegated to kwargs? If not, putting the `IdDict` first by convention prevents (most) any bikeshedding there.

cossio:
Any decision here?

DhairyaLGandhi:
I prefer the first proposal in https://github.com/FluxML/Flux.jl/pull/1819#issuecomment-1002764038. I'd imagine `ϵ` is common amongst many optimisers, and fitting it into different slots of optimiser API might get meddlesome, having it consistently as a kwarg makes sense. Besides, while experimenting with the hyper parameters of ADAM, I'd imagine modifying `β` would happen more frequently than `ϵ` (which strictly may not be a parameter at all).

cossio:
I can make a PR

PR #1820, opened 2021-12-29
Add `Split` as a convenience constructor

darsnack:
This adds a `Split` convenience constructor that just creates a `Parallel` layer. Can be useful for models learning parameters of a distribution (VAEs, etc.).

Closes #1289.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/test/layers/basic.jl
+++ b/test/layers/basic.jl
@@ -242,6 +242,16 @@ import Flux: activations
       @test gs[2].x ≈ gs_reg[2].x
       @test gs[3].x ≈ gs_reg[3].x
     end
+
+    @testset "Split" begin
+      x = rand(4, 5)
+      split = Split(Dense(4, 2), Dense(4, 1))
+      par = Parallel(tuple, split.layers...)
+
+      @test split(x) == par(x)
+      @test length(split(x)) == length(split.layers)
+      @test size.(split(x)) == ((2, 5), (1, 5))
+    end
   end
 
   @testset "Embedding" begin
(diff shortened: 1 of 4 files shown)


test/layers/basic.jl
```diff
@@ -242,6 +242,16 @@ import Flux: activations
...
     end
+
+    @testset "Split" begin
+      x = rand(4, 5)
+      split = Split(Dense(4, 2), Dense(4, 1))
+      par = Parallel(tuple, split.layers...)
+
+      @test split(x) == par(x)
+      @test length(split(x)) == length(split.layers)
+      @test size.(split(x)) == ((2, 5), (1, 5))
```
ToucheSir, line 253:
```suggestion
      split = Split(Dense(4, 3), Dense(4, 2), Dense(4, 1))
      par = Parallel(tuple, split.layers...)

      @test split(x) == par(x)
      @test length(split(x)) == length(split.layers)
      @test size.(split(x)) == ((3, 5), (2, 5), (1, 5))
```
IIUC the current implementation of `Parallel`, this should fail. If it does, making it pass will be a blocker for this PR.

DhairyaLGandhi: I thought we were punting on a split layer for the time being

darsnack:
I made this PR before @ToucheSir and @DhairyaLGandhi's comments on #1289. I agree with the comments, but the PR was simple enough that I just did it in case we want it.

My motivation for not outright rejecting the change was that `Parallel((x, y) -> (x..., y), layers...)` is a bit opaque for beginners.

test/layers/basic.jl
darsnack: Yeah I guess `(x, y) -> (x..., y)` would be necessary without #1698.

PR #1822, opened 2022-01-04
unbreak CI for now

ToucheSir:
After too long failing to repro https://github.com/FluxML/Flux.jl/issues/1804, getting CI back online for bors is the highest priority.

### DIFF
--- a/test/cuda/layers.jl
+++ b/test/cuda/layers.jl
@@ -46,7 +46,8 @@ function gpu_gradtest(name::String, layers::Vector, x_cpu = nothing, args...; te
           # test
           if test_cpu
             if VERSION >= v"1.7" && layer === GroupedConvTranspose && args[end] == selu
-              @test_broken y_gpu ≈ y_cpu rtol=1f-3 atol=1f-3
+              # FIXME revisit this after CUDA deps on CI are updated
+              @test y_gpu ≈ y_cpu rtol=2 atol=2
             else
               @test y_gpu ≈ y_cpu rtol=1f-3 atol=1f-3
             end
--- a/test/cuda/layers.jl
+++ b/test/cuda/layers.jl
@@ -46,7 +46,20 @@ function gpu_gradtest(name::String, layers::Vector, x_cpu = nothing, args...; te
           # test
           if test_cpu
             if VERSION >= v"1.7" && layer === GroupedConvTranspose && args[end] == selu
-              # FIXME revisit this after CUDA deps on CI are updated
+              # FIXME revisit this after CUDA deps on CI are updated. Here's a snapshot from 2021-12-25:
+              # CUDA toolkit 11.4, local installation
+              # NVIDIA driver 495.29.5, for CUDA 11.5
+              # CUDA driver 11.5
+              # Libraries:
+              # - CUBLAS: 11.6.1
+              # - CURAND: 10.2.5
+              # - CUFFT: 10.5.2
+              # - CUSOLVER: 11.2.0
+              # - CUSPARSE: 11.6.0
+              # - CUPTI: 14.0.0
+              # - NVML: 11.0.0+495.29.5
+              # - CUDNN: 8.20.4 (for CUDA 11.4.0)
+              # - CUTENSOR: missing
               @test y_gpu ≈ y_cpu rtol=2 atol=2
             else
               @test y_gpu ≈ y_cpu rtol=1f-3 atol=1f-3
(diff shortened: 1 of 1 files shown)


test/cuda/layers.jl
```diff
@@ -46,7 +46,8 @@ function gpu_gradtest(name::String, layers::Vector, x_cpu = nothing, args...; te
           # test
           if test_cpu
             if VERSION >= v"1.7" && layer === GroupedConvTranspose && args[end] == selu
-              @test_broken y_gpu ≈ y_cpu rtol=1f-3 atol=1f-3
+              # FIXME revisit this after CUDA deps on CI are updated
+              @test y_gpu ≈ y_cpu rtol=2 atol=2
```
DhairyaLGandhi, line 50: Doesn't really help here since the error bounds are pretty high and the broken test is already specific to ConvTranspose + selu. Can we specify a kind of failure we expect? Say we expect the test to fail but not error?

DhairyaLGandhi (changes requested):

test/cuda/layers.jl
ToucheSir: Nope, because the test doesn't always fail! It was a choice between this amd skipping the test entirely.

DhairyaLGandhi: I'd rather avoid having high error tolerances since that isn't very helpful in the real world, and and it's unlikely an error would be raised in this code path (although I'd rather retain the test). Something that doesn't error and gives inaccurate answers would be hard to debug! Can we compare against a standard (say tf/ pytorch) implementation?

ToucheSir: We could, but that doesn't address the issue of very high variance in results. The main problem is that we (or at least I) can't figure out _where_ that variance is coming from. It could be something deep within the bowels of cuDNN, and since the forward pass of `ConvTranspose` is literally 1 cuDNN call + broadcasted bias add + broadcasted activation, there'd be very little we could do about that.

All that said, I'm happy to change it to a `@test_skip` if you feel that's more appropriate.

DhairyaLGandhi: Is the issue that we might see Red ci spuriously if this test passes as it is on master? I don't think we've encountered that very frequently with the current setup right?

I'm fairly certain that the underlying issue would be in CUDA/ cudnn, and that would be pretty out of our hands at that point. To fix this we'd need Julia kernels which might not be the worst idea but seeing as the motivation is to fix one combination of conv and activation, it's fair to say that it would be low priority with little overall benefit.

ToucheSir: Not so spuriously, unfortunately. I think the failure rate is at least 25%, if not closer to 40%. The bigger problem is that this means all bors commands will be rolling of the dice, since those also run on Buildkite (and no merging will happen if a spurious test failure occurs).

DhairyaLGandhi: If we are to let this test be here with wide tolerance it would be good to know what CUDA deps are referred to in the comment and what we should do to be alerted when this combination works to a decent degree again.

ToucheSir: That's the thing, I don't know because I couldn't repro anything. It may well be that the CUDA deps are a red herring and the problem lies elsewhere (say with GPUCompiler + the new LLVM + compiler changes on Julia 1.7). I've added a commit with the last `CUDA.versioninfo()` output I got out of Buildkite.

ToucheSir: Thoughts? Concerns? Again, I'm happy to turn this into a `@test_skip` with the previous tolerance to get the PR merged.

darsnack: Is there a functional difference here between `@test_skip` and the adjusted tolerances? They are so wide, I would expect a 200% relative tolerance to be tantamount to skipping any testing. So, let's just change to `@test_skip` and merge.

ToucheSir: Done.

darsnack (approved):

ToucheSir:
What changed in the last 48 hours that we now have new failures despite no changes to Flux, Zygote or IRTools and no related changes to NNlib???

mcabbott:
Needs https://github.com/FluxML/NNlib.jl/pull/375 I think. Although Bors will probably still fail thanks to a red flag, in this bizarre system.

ToucheSir:
Seems to be superseded by #1836, no hacks needed.

DhairyaLGandhi:
Apologies for the delayed review request, I saw this only around when #1836 was opened

PR #1825, opened 2022-01-07
Update conv.jl

rkube:
Clarify meaning of `pad` parameter in `Conv`.

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -50,7 +50,7 @@ Then:
 * `filter` should be a tuple of `N` integers.
 * Keywords `stride` and `dilation` should each be either single integer,
   or a tuple with `N` integers.
-* Keyword `pad` specifies the number of elements added to the image borders. It can be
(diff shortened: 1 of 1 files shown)


ToucheSir (dismissed):
Thanks!

src/layers/conv.jl
```diff
@@ -50,7 +50,7 @@ Then:
 * `filter` should be a tuple of `N` integers.
 * Keywords `stride` and `dilation` should each be either single integer,
   or a tuple with `N` integers.
-* Keyword `pad` can be:
+* Keyword `pad` specifies the number of elements added to the image borders. It can be
```
DhairyaLGandhi, line 53:
```suggestion
* Keyword `pad` specifies the number of elements added to the borders of the data array. It can be
```

DhairyaLGandhi (dismissed):
Lgtm

DhairyaLGandhi (approved):
bors r+

thanks!

PR #1833, opened 2022-01-13
add step!

FelixBenning:
Add `step!` like suggested in #666 as a single step of `train!` to allow for more exotic optimisers to simply overload `step!` and be able to use the `train!` wrapper.

### PR Checklist

- [ ] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,7 +3,7 @@
 ## v0.12.9
 * Fixed incorrect output and added GPU compatibility for [AlphaDropout](https://github.com/FluxML/Flux.jl/pull/1781).
 * Add trilinear [Upsample layer](https://github.com/FluxML/Flux.jl/pull/1792).
-* Add `step!` as a single training step of `train!` to allow for more exotic
+* Add `optimstep!` as a single training step of `train!` to allow for more exotic
 optimisers (#666)
 
 ## v0.12.8
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 ## v0.12.9
 * Fixed incorrect output and added GPU compatibility for [AlphaDropout](https://github.com/FluxML/Flux.jl/pull/1781).
 * Add trilinear [Upsample layer](https://github.com/FluxML/Flux.jl/pull/1792).
+* Add `step!` as a single training step of `train!` to allow for more exotic
+optimisers (#666)
 
 ## v0.12.8
 * Optimized inference and gradient calculation of OneHotMatrix[pr](https://github.com/FluxML/Flux.jl/pull/1756)
--- a/docs/src/training/training.md
+++ b/docs/src/training/training.md
@@ -17,15 +17,27 @@ for d in datapoints
   # `d` should produce a collection of arguments
   # to the loss function
 
-  # Calculate the gradients of the parameters
-  # with respect to the loss function
-  grads = Flux.gradient(parameters) do
+  # Update the parameters based on the chosen
+  # optimiser (opt)
+  loss, grads = optimstep!(params, opt) do
     loss(d...)
   end
+end
+```
 
+`optimstep!` is the optimiser implementation and thus dispatches depending on
+the optimizer type. As an example, the default `optimstep!` for optimiser who
+use the gradient to update the parameters (e.g. gradient descent, momentum, ADAM, etc.) looks like this
+
+```julia
+function optimstep!(loss, params, opt)
+  # Calculate the gradients of the parameters
+  # with respect to the loss function
+  val, grads = Flux.withgradient(loss, parameters)
   # Update the parameters based on the chosen
   # optimiser (opt)
   Flux.Optimise.update!(opt, parameters, grads)
+  return val, grads
 end
 ```
 
--- a/src/optimise/Optimise.jl
+++ b/src/optimise/Optimise.jl
@@ -3,7 +3,7 @@ module Optimise
 using LinearAlgebra
 import ArrayInterface
 
-export train!, update!,
+export train!, step!, update!,
 	Descent, ADAM, Momentum, Nesterov, RMSProp,
 	ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW,RADAM, OADAM, AdaBelief,
 	InvDecay, ExpDecay, WeightDecay, stop, skip, Optimiser,
--- a/src/optimise/Optimise.jl
+++ b/src/optimise/Optimise.jl
@@ -3,7 +3,7 @@ module Optimise
 using LinearAlgebra
 import ArrayInterface
 
-export train!, step!, update!,
+export train!, optimstep!, update!,
 	Descent, ADAM, Momentum, Nesterov, RMSProp,
 	ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW,RADAM, OADAM, AdaBelief,
 	InvDecay, ExpDecay, WeightDecay, stop, skip, Optimiser,
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -1,5 +1,5 @@
 using Juno
-import Zygote: Params, gradient
+import Zygote: Params, withgradient
 
 """
     update!(x, x̄)
@@ -102,8 +102,9 @@ for d in data
 
 """
 function step!(loss, params, opt)
-  gs = gradient(loss, params)
+  val, gs = withgradient(loss, params)
   update!(opt, params, gs)
+  return val, gs
 end
 
 """
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -92,17 +92,16 @@ and calls `Flux.Optimise.update!` to adjust the parameters, but you can overload
 `step!` for specific types of `opt`. This can be useful if your optimization routine
 has does not follow the standard gradient descent procedure (e.g. gradient-free optimizers).
 
-While the loss function of `train!` still accepts data as input, the loss function
-of `step!` accepts no input. `train!` cycles through the data in a loop
-roughly like this
-
+Unlike `train!`, the loss function of `step!` accepts no input.
+Instead, `train!` cycles through the data in a loop and calls `step!`:
 ```julia
 for d in data
   step!(ps, opt) do
     loss(d)
   end
+end
 ```
-
+If you are writing [Custom Training loops](@ref), then you should follow this pattern.
 """
 function step!(loss, params, opt)
   val, gs = withgradient(loss, params)
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -81,6 +81,32 @@ batchmemaybe(x) = tuple(x)
 batchmemaybe(x::Tuple) = x
 
 """
+    step!(loss, params, opt)
+
+`step!` uses a `loss` function (with no inputs) to improve the [Model parameters](@ref) (`params`)
+based on a pluggable [Optimisers](@ref) (`opt`). It represents a single step in
+the training loop `train!`. While there is a default implementation for
+optimisers which are based on the `update!` function and only require gradient
+information, this `step!` has to be overloaded for more general optimisers.
+
+While the loss function of `train!` still accepts data as input, the loss function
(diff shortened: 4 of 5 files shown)


DhairyaLGandhi:
Similar to https://github.com/FluxML/Flux.jl/pull/1017 which should be complete

src/optimise/train.jl
```diff
@@ -80,6 +80,32 @@ end
...
+  step!(ps, opt) do
+    loss(d)
+  end
+```
+
+"""
+function step!(loss, params, opt)
+  gs = gradient(loss, params)
+  update!(opt, params, gs)
+end
```
darsnack, line 107: I would prefer this to be consistent with `train!`, and so users can use the default step in a custom loop without knowing `batchmemaybe`.

```suggestion
function step!(loss, params, data, opt)
  gs = gradient(params) do
    loss(batchmemaybe(data)...)
  end
  update!(opt, params, gs)
end
```

darsnack (changes requested):

src/optimise/train.jl
FelixBenning: To be honest I do not know what `batchmaybe` does and took the view of someone wanting to write a new optimiser (i.e. overloading `step!`). In that case I would not want to deal with batchmaybe, so I left it in `train!`. Assuming that end user would use `train!` anyway. I guess the first question is, what does batchmaybe do?

Besides that, I wonder if it is cleaner to leave data out of the step!, this would essentially split the concern of optimisation (i.e. implementing a `step!`) from handling data, i.e. defining a loss. I wonder why `train!` doesn't have batch or shuffling utilities though, as that would be the data handling part. Without that data handling functionality you might as well just use a for loop

darsnack: `batchmemaybe` just turns things that are not a tuple into one. It just another way of doing:
```julia
if data isa Tuple
  loss(data...)
else
  loss(data)
end
```

It's worth noting that the main use-case (i.e. 99%) for `step!` is not overloading it, but writing a custom `train!`:
```julia
function mytrain!(loss, params, data, opt)
  # do some stuff
  for d in data
    # do other stuff
    # okay now update my model like Flux does
    # I don't want to worry about the details
    step!(loss, params, data, opt)
    # do some more stuff
  end
end
```
Seems weird if `Flux.train!` and `mytrain!` can't accept the same loss cause `Flux.train!`/`step!` don't treat the loss in a consistent way.

darsnack: The note about why not use a for-loop vs. make `train!` more complex is heavily discussed. The maintainers are split between "`train!` is a mistake" and "we should add more to `train!`." I'd recommend let's just avoid diving into that hole for this PR and assume that `train!` is behaviorally unchanged.

FelixBenning: Do you think that could be a callback for the `train!` function? "do other stuff" in every step?

FelixBenning: okay so `batchmaybe` does not help partitioning the data. And that seems like it really should be a functionality of `train!`. Doing stuff like

```
partition(1:size(x_train, ndims(x_train)), batchsize)
```

to get a generator over the index ranges of your batches is annoying. And this is just the simplest minibatching version without shuffling. At least something should deal with this data handling, and this is where this `batchmaybe` function really belongs. It should not be a concern of the optimiser

darsnack: Okay, thinking about it more, it is reasonable that the person writing a custom train loop should know what `d in data` looks like and how to pass it to the loss function. I'm convinced this is OK, let's just solicit some opinions from other maintainers since this is introducing a new API.

@mcabbott @DhairyaLGandhi @CarloLucibello @ToucheSir any thoughts?

darsnack: One other suggestion is to use `Zygote.withgradient` instead of `gradient` and return both the loss and gradients from `step!`.

mcabbott: Bit of a drive-by but: Do we want to be adding new things which rely on zero-parameter functions and implicit Params? Aren't we trying to kill that entirely & move on?

darsnack: I thought about that, and right now, `train!` implicitly forces this. So, whenever we actually move to explicit, both `train!` and `step!` will be updated accordingly. Meaning this function would be `step!(loss, opt, args...)` and expect `loss(args...)`.

`train!` would be something like `loss(batchmemaybe(d)..., args...)` which is pretty hairy. That makes me think that the transition to explicit might be the right point to eliminate `train!` (fallout being that this PR would be irrelevant too).

darsnack: But also, it is pretty standard to write your loss function as `loss(x, y, m) = ...` so maybe not really that hairy other than the docstring.

FelixBenning: yeah, if the implicit params are gone, it would probably look like

```julia
function train!(...)
    for loss_sample in data_manager(loss, data)
        # data_manager generates from loss(params, x, y) a loss_sample(params) generator
        step!(loss, params, opt) # optimiser plugs params into loss
    end
end
```
So the same interface as now

FelixBenning: > One other suggestion is to use `Zygote.withgradient` instead of `gradient` and return both the loss and gradients from `step!`.

That is a good idea, what function should I use instead of `update!`then though?

darsnack: You would still use `update!`. `withgradient` returns two items: the loss and the gradients. The latter can be fed into `update!` as normal.

FelixBenning: so if we discard the loss anyway, why would you use withgradient?

darsnack: Sorry the suggestion is not to discard. We should return it along with the gradients at the end of `step!`. This will be useful for the custom loop case.

src/optimise/train.jl
```diff
@@ -102,8 +102,9 @@ for d in data
 
 """
 function step!(loss, params, opt)
-  gs = gradient(loss, params)
+  val, gs = withgradient(loss, params)
```
DhairyaLGandhi, line 105: If we want to be able to manage hooks, sensitivities and gradient clips better we should definitely use pullback here.  We would need this for things other than deep learning anyway.

FelixBenning: I don't know much about the differences between AD algos (what is pullback?). About the explicit params, no data is not needed. Because you generate loss function realizations with data, you can just pass one such realization to step!. As I suggested in a previous post something like

```julia
function train!(...)
    for loss_sample in data_manager(loss, data)
        # data_manager generates from loss(params, x, y) a loss_sample(params) generator
        step!(loss_sample, params, opt) # optimiser plugs params into loss_sample
    end
end

function data_manager(loss, data)
     for x,y in data
         yield param -> loss(param, x,y)
     end
end
```
I am not sure if such a yield statement exists like in python, but the idea is to create a generator of loss realizations

FelixBenning: An alternative to `simplePass` could be

```julia

function bootstrapping(loss, data, samples)
    return imap(1:samples) do 
         x,y = rand(data)
         return param -> loss(param, x,y)
    end
end
```

darsnack: Either way, `train!` already forces implicit params. I think it is better to deal with such issues separately in a different PR.

src/optimise/train.jl
```diff
@@ -80,6 +80,33 @@ end
...
 batchmemaybe(x::Tuple) = x
 
+"""
+    step!(loss, params, opt)
+
+`step!` uses a `loss` function (with no inputs) to improve the [Model parameters](@ref) (`params`)
+based on a pluggable [Optimisers](@ref) (`opt`). It represents a single step in
+the training loop `train!`. While there is a default implementation for
+optimisers which are based on the `update!` function and only require gradient
+information, this `step!` has to be overloaded for more general optimisers.
```
darsnack, line 90:
```suggestion
`step!` uses a `loss` function (with no inputs) to improve the [Model parameters](@ref) (`params`)
based on a pluggable [Optimisers](@ref) (`opt`). It represents a single step in
the training loop `train!`.

The default implementation for `step!` is takes the gradient of `loss`
and calls `Flux.Optimise.update!` to adjust the parameters, but you can overload
`step!` for specific types of `opt`. This can be useful if your optimization routine
has does not follow the standard gradient descent procedure (e.g. gradient-free optimizers).
```

src/optimise/train.jl
```diff
@@ -80,6 +80,33 @@ end
...
+of `step!` accepts no input. `train!` cycles through the data in a loop
+roughly like this
+
+```julia
+for d in data
+  step!(ps, opt) do
+    loss(d)
+  end
+```
+
```
darsnack, line 102:
````suggestion
Unlike `train!`, the loss function of `step!` accepts no input.
Instead, `train!` cycles through the data in a loop and calls `step!`:
```julia
for d in data
  step!(ps, opt) do
    loss(d)
  end
end
```
If you are writing [Custom Training loops](@ref), then you should follow this pattern.
````

**Review by darsnack — CHANGES_REQUESTED** [review]
A couple of docstring changes. The implementation looks good to me.

Can you also update the docs for "Custom Training loops" to use this new function? And add the docstring into the docs in that location?

One other suggestion is to call this `trainstep!`, since `step!` is very generic, and we might want to hold onto the name for something else.

File: src/optimise/train.jl [hunk 5]
```diff
@@ -81,29 +81,29 @@ batchmemaybe(x) = tuple(x)
 batchmemaybe(x::Tuple) = x
 
 """
-    step!(loss, params, opt)
+    optimstep!(loss, params, opt)
```
FelixBenning, line 84: I suggest `optimstep!` instead of `trainstep!` to indicate that this is the optimiser interface and keep the ML jargon to a minimum

docs/src/training/training.md
```diff
@@ -17,15 +17,27 @@ for d in datapoints
   # `d` should produce a collection of arguments
   # to the loss function
 
-  # Calculate the gradients of the parameters
-  # with respect to the loss function
-  grads = Flux.gradient(parameters) do
+  # Update the parameters based on the chosen
```
FelixBenning, line 20: This is right at the beginning instead of in the `Custom Training Loop Section`. It seems to me like the custom training loop section might either be redundant or demonstrate how to have a custom gradient calculation now.

NEWS.md
```diff
@@ -3,6 +3,8 @@
 ## v0.12.9
 * Fixed incorrect output and added GPU compatibility for [AlphaDropout](https://github.com/FluxML/Flux.jl/pull/1781).
 * Add trilinear [Upsample layer](https://github.com/FluxML/Flux.jl/pull/1792).
+* Add `step!` as a single training step of `train!` to allow for more exotic
```
ToucheSir, line 6:
```suggestion
* Add `optimstep!` as a single training step of `train!` to allow for more exotic
```
To match recent changes below.

src/optimise/Optimise.jl
```diff
@@ -3,7 +3,7 @@ module Optimise
 using LinearAlgebra
 import ArrayInterface
 
-export train!, update!,
+export train!, step!, update!,
```
ToucheSir, line 6:
```suggestion
export train!, optimstep!, update!,
```
This needs to be updated too.

src/optimise/train.jl
mcabbott: One vote for something evoking `train!` to stress that they are closely related. 

If the longer-term plan is to use Optimisers.jl, this may not fit with `train!` at all -- some recent discussion here: https://github.com/FluxML/Flux.jl/pull/1902#discussion_r821239364 . In which case there will be an implicit-style `train!` & `Params` story, and an explicit-style `gradient` and `Optimisers.update!`. With such a divide, this function wants to be clearly on the `train!` & `Params` side. 

Maybe it should just be 3-arg `train!`? Without a data iterator, there is no iteration, that's all:
```julia
train!(loss, data, ::Params, ::AbstractOptimiser)  # calls loss(d...) for d in data
train!(loss, ::Params, ::AbstractOptimiser)        # calls loss() since there is no data
```

CarloLucibello:
we have Flux.trainstep!

PR #1836, opened 2022-01-15
Try using latest cu(DNN) binaries

ToucheSir:
Possible alternative to #1822.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -17,6 +17,8 @@ steps:
     agents:
       queue: "juliagpu"
       cuda: "*"
+    env:
+      JULIA_CUDA_USE_BINARYBUILDER: "true"
     timeout_in_minutes: 60
 
   # - label: "GPU nightly"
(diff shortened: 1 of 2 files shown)


ToucheSir:
bors try

DhairyaLGandhi:
Was there some background to this? Is there sometji g abou the cudnn we were using causing this?

DhairyaLGandhi (approved):
bors r+

ToucheSir:
Testing whether this note from https://docs.nvidia.com/deeplearning/cudnn/release-notes/rel_8.html#rel-830 applies:
> In some cases, cudnnConvolutionBackwardFilter() generates numerically imprecise results when used with algo set to CUDNN_CONVOLUTION_BWD_FILTER_ALGO_1. This issue is most frequently encountered with three dimensional spatial tensors. Users of the backend API may explicitly avoid backend engine 2032 or consider the numerical notes of engines and reject any marked as offering reduced precision reduction (CUDNN_NUMERICAL_NOTE_REDUCED_PRECISION_REDUCTION).

PR #1832, opened 2022-01-12
Use NNlib.conv_bias_act for Conv

darsnack:
This closes #1302 (supersedes) and fixes #558.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md
- [ ] Documentation, if applicable
- [ ] API changes require approval from a committer (different from the author, if applicable)

PR #1837, opened 2022-01-16
Use faster activation functions

mcabbott:
This substitutes `tanh_fast` where possible, since `tanh` often dominates the forward pass of small networks. Builds on #1761, but in the months that sat waiting, I have mislaid the benchmarks I ran. Best case 2x better forwards, worst case no improvement, modulo noise.

Intersects with https://github.com/FluxML/Flux.jl/pull/1832, which would also do this to conv layers, but not to RNNs.

Closes #1272; this version is significantly faster, and this PR applies it to more cases.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -154,7 +154,7 @@ end
 @functor Dense
 
 function (a::Dense)(x::AbstractVecOrMat)
-  W, b= a.weight, a.bias
+  W, b = a.weight, a.bias
   σ = NNlib.fast_act(a.σ, x)  # replaces tanh => tanh_fast, etc
   return σ.(W*x .+ b)
 end
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -154,7 +154,8 @@ end
 @functor Dense
 
 function (a::Dense)(x::AbstractVecOrMat)
-  W, b, σ = a.weight, a.bias, a.σ
+  W, b= a.weight, a.bias
+  σ = NNlib.fast_act(a.σ, x)  # replaces tanh => tanh_fast, etc
   return σ.(W*x .+ b)
 end
 
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -161,7 +161,8 @@ end
 @functor Conv
 
 function (c::Conv)(x::AbstractArray)
-  σ, b = c.σ, reshape(c.bias, ntuple(_ -> 1, length(c.stride))..., :, 1)
+  b = reshape(c.bias, ntuple(_ -> 1, length(c.stride))..., :, 1)
+  σ = NNlib.fast_act(c.σ, x)
   cdims = DenseConvDims(x, c.weight; stride = c.stride, padding = c.pad, dilation = c.dilation, groups = c.groups)
   σ.(conv(x, c.weight, cdims) .+ b)
 end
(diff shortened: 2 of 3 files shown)


src/layers/basic.jl
```diff
@@ -154,7 +154,8 @@ end
 @functor Dense
 
 function (a::Dense)(x::AbstractVecOrMat)
-  W, b, σ = a.weight, a.bias, a.σ
+  W, b= a.weight, a.bias
```
CarloLucibello, line 157:
```suggestion
  W, b = a.weight, a.bias
```

CarloLucibello:
It's a pity to pollute the code a bit but I guess the performance increase is worth it. 

I wonder what could be a more aesthetically pleasing alternative. Pointing `NNlib.sigmod` to `sigmoid_fast` and defining a `NNlib.tanh` as `tanh_fast`?

mcabbott:
I wouldn't claim this as an aesthetic improvement. But it is a performance one.

I'm against less explicit names like `NNlib.tanh`, I think it's confusing to read packages which don't work how they look like they work, because some Base name means something different. `Flux.rand` used to be some such trick, and we ripped it out, for explicit `rand32`.

CarloLucibello:
Ok. Should we wait for a final tag of the v0.12 before merging (#1838)? I wouldn't say this PR is breaking though

mcabbott:
Not so breaking as to demand a release, but if there's one nearby perhaps that's the safe time. So after last 0.12.x tag, before v0.13, as you say.

ToucheSir:
Buildkite is failing with weird errors again :grimacing:

ToucheSir:
bors try

src/layers/basic.jl
ToucheSir, line 157:
```suggestion
  W, b = a.weight, a.bias
```

src/layers/conv.jl
```diff
@@ -161,7 +161,8 @@ end
 @functor Conv
 
 function (c::Conv)(x::AbstractArray)
-  σ, b = c.σ, reshape(c.bias, ntuple(_ -> 1, length(c.stride))..., :, 1)
+  b = reshape(c.bias, ntuple(_ -> 1, length(c.stride))..., :, 1)
```
ToucheSir, line 164: This line could be pulled into a helper function. I also noticed that `Conv` uses `ntuple` while the others use `map`.

mcabbott: Could be. Maybe revisit with #1832?

ToucheSir: How about an internal only function for now? Or barring that, updating this line to match.
```suggestion
  b = reshape(c.bias, map(_->1, c.stride)..., :, 1)
```

src/layers/recurrent.jl
```diff
@@ -224,8 +225,8 @@ function (m::LSTMCell{A,V,<:NTuple{2,AbstractMatrix{T}}})((h, c), x::Union{Abstr
   b, o = m.b, size(h, 1)
   g = m.Wi*x .+ m.Wh*h .+ b
   input, forget, cell, output = multigate(g, o, Val(4))
-  c′ = @. σ(forget) * c + σ(input) * tanh(cell)
-  h′ = @. σ(output) * tanh(c′)
+  c′ = @. sigmoid_fast(forget) * c + sigmoid_fast(input) * tanh_fast(cell)
+  h′ = @. sigmoid_fast(output) * tanh_fast(c′)
```
ToucheSir, line 229: Ought these to get the `fast_act` treatment too? I'm ok with revisiting too, RNN cell inflexibility is a bit of a long standing issue.

ToucheSir (approved):

src/layers/recurrent.jl
mcabbott: Oh right. If we decide to disable `fast_tanh` on CuArrays, that will be ignored here. But perhaps revisit if & when... it's a bit more clutter to squeeze that in.

ToucheSir: It may be a blessing in disguise, as currently plain `tanh.(...)` will hit the likely obsolete https://github.com/FluxML/NNlibCUDA.jl/blob/master/src/cudnn/activations.jl.

PR #1840, opened 2022-01-17
move eps to the end

cossio:
https://github.com/FluxML/Flux.jl/pull/1819 turned out to be a breaking change, see comment https://github.com/FluxML/Flux.jl/pull/1819#issuecomment-1002707317.

This PR overwrites that one, doing a similar change, but in a way that should not be breaking.

@DhairyaLGandhi @CarloLucibello @ToucheSir

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -133,11 +133,11 @@ opt = RMSProp(0.002, 0.95)
 mutable struct RMSProp <: AbstractOptimiser
   eta::Float64
   rho::Float64
-  epsilon::Float64
   acc::IdDict
+  epsilon::Float64
 end
 
-RMSProp(η = 0.001, ρ = 0.9, ϵ = ϵ) = RMSProp(η, ρ, ϵ, IdDict())
+RMSProp(η = 0.001, ρ = 0.9; ϵ = ϵ) = RMSProp(η, ρ, IdDict(), ϵ)
 
 function apply!(o::RMSProp, x, Δ)
   η, ρ = o.eta, o.rho
@@ -167,11 +167,11 @@ opt = ADAM(0.001, (0.9, 0.8))
 mutable struct ADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
-  epsilon::Float64
   state::IdDict
+  epsilon::Float64
 end
 
-ADAM(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = ADAM(η, β, ϵ, IdDict())
+ADAM(η = 0.001, β = (0.9, 0.999); ϵ = ϵ) = ADAM(η, β, IdDict(), ϵ)
 
 function apply!(o::ADAM, x, Δ)
   η, β = o.eta, o.beta
@@ -209,11 +209,11 @@ opt = RADAM(0.001, (0.9, 0.8))
 mutable struct RADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
-  epsilon::Float64
   state::IdDict
+  epsilon::Float64
 end
 
-RADAM(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = RADAM(η, β, ϵ, IdDict())
+RADAM(η = 0.001, β = (0.9, 0.999); ϵ = ϵ) = RADAM(η, β, IdDict(), ϵ)
 
 function apply!(o::RADAM, x, Δ)
   η, β = o.eta, o.beta
@@ -259,11 +259,11 @@ opt = AdaMax(0.001, (0.9, 0.995))
 mutable struct AdaMax <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
-  epsilon::Float64
   state::IdDict
+  epsilon::Float64
 end
 
-AdaMax(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = AdaMax(η, β, ϵ, IdDict())
+AdaMax(η = 0.001, β = (0.9, 0.999); ϵ = ϵ) = AdaMax(η, β, IdDict(), ϵ)
 
 function apply!(o::AdaMax, x, Δ)
   η, β = o.eta, o.beta
@@ -302,11 +302,11 @@ opt = OADAM(0.001, (0.9, 0.995))
 mutable struct OADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
-  epsilon::Float64
   state::IdDict
+  epsilon::Float64
 end
 
-OADAM(η = 0.001, β = (0.5, 0.9), ϵ = ϵ) = OADAM(η, β, ϵ, IdDict())
+OADAM(η = 0.001, β = (0.5, 0.9); ϵ = ϵ) = OADAM(η, β, IdDict(), ϵ)
 
 function apply!(o::OADAM, x, Δ)
   η, β = o.eta, o.beta
@@ -345,11 +345,11 @@ opt = ADAGrad(0.001)
 """
 mutable struct ADAGrad <: AbstractOptimiser
   eta::Float64
-  epsilon::Float64
   acc::IdDict
+  epsilon::Float64
 end
 
-ADAGrad(η = 0.1, ϵ = ϵ) = ADAGrad(η, ϵ, IdDict())
+ADAGrad(η = 0.1; ϵ = ϵ) = ADAGrad(η, IdDict(), ϵ)
 
 function apply!(o::ADAGrad, x, Δ)
   η = o.eta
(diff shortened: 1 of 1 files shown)


CarloLucibello:
One minor thing: do we have other cases of unicode keyword args or a policy about them?

That said, in a future PR I would deprecate all constructors and introduce keywords only ones.

cossio:
> That said, in a future PR I would deprecate all constructors and introduce keywords only ones.

I also prefer this. But that's breaking, so maybe leave that for a future release.

src/optimise/optimisers.jl
```diff
@@ -133,11 +133,11 @@ opt = RMSProp(0.002, 0.95)
 mutable struct RMSProp <: AbstractOptimiser
   eta::Float64
   rho::Float64
-  epsilon::Float64
   acc::IdDict
+  epsilon::Float64
 end
 
-RMSProp(η = 0.001, ρ = 0.9, ϵ = ϵ) = RMSProp(η, ρ, ϵ, IdDict())
+RMSProp(η = 0.001, ρ = 0.9; ϵ = ϵ) = RMSProp(η, ρ, IdDict(), ϵ)
```
mcabbott, line 140: Apparently we have a global constant with a one-letter unicode name:
```
julia> Flux.Optimise.ϵ
1.0e-8
```
If touching this, can this change name to EPSILON or something?

It seems like a bug magnet, in that if e.g. you type the wrong unicode epsilon in your function, you will get the global variable.

cossio: I agree. So you want me to rename `Flux.Optimise.ϵ` in this PR?

mcabbott: I guess I'm also not sure why have it at all. Why not hard-code 1e-8 alongside 0.001? That is less confusing if you look at the source. What's gained by moving it elsewhere?

ToucheSir: If we consider it an internal constant, then it could be removed completely. Otherwise I'd vote to put in on the chopping block for v0.13.

src/optimise/optimisers.jl
```diff
@@ -167,11 +167,11 @@ opt = ADAM(0.001, (0.9, 0.8))
 mutable struct ADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
-  epsilon::Float64
   state::IdDict
+  epsilon::Float64
 end
 
-ADAM(η = 0.001, β = (0.9, 0.999), ϵ = ϵ) = ADAM(η, β, ϵ, IdDict())
+ADAM(η = 0.001, β = (0.9, 0.999); ϵ = ϵ) = ADAM(η, β, IdDict(), ϵ)
```
mcabbott, line 174: I have another comment which perhaps I make a different thread for. This uses epsilon (well, varepsilon) as a keyword argument, and thus something you may have to type to read. The previous state used greek only for positional arguments. Is this a good idea?

ToucheSir: I think that was Carlo's question, and I can see good arguments either way.

mcabbott: Oh I see, I didn't read enough of the thread.

One vote for keeping the logical new order of the struct, parameters then IdDict. 

Arguably these constructors should all have types, `ADAM(η::Real = 0.001, β::Tuple = (0.9, 0.999), ϵ::Real = ϵ)`. And since an IdDict isn't a real number, you can keep a method which accepts the old order of arguments.

cossio:
> One minor thing: do we have other cases of unicode keyword args or a policy about them?

I could rename the kwarg `ϵ -> eps`

src/optimise/optimisers.jl
cossio: > One vote for keeping the logical new order of the struct, parameters then IdDict.

That would be breaking. I'd say it could be done in a different PR. And if it's breaking go all the way and make all parameters kwargs.

mcabbott: Can you say what this breaks?

src/optimise/optimisers.jl
cossio: I'd say this issue can be dealt with in a separate PR.

src/optimise/optimisers.jl
cossio: Because of https://github.com/FluxML/Flux.jl/pull/1819#issuecomment-1002707317 ? Maybe I misunderstood you.

mcabbott: But this is solved by multiple disptach. The constructor can have several methods, which detect where the IdDict is located. One of which will catch the old usage.

cossio: Can you be explicit about the arg order you are suggesting?

mcabbott:
```
ADAM(η::Real = 0.001, β::Tuple = (0.9, 0.999), ϵ::Real = 1e-8, dict::IdDict=IdDict()) = ADAM(η, β, ε, dict)  # 0,1,2,3,4 arg functions, with new order

ADAM(η::Real, β::Tuple, dict::IdDict) = ADAM(η, β, 1e-8, dict)  # one more method, matches previous usage
```
And perhaps the 2nd is in the deprecations file, possibly define with a macro / printing a depwarn by hand.

src/optimise/optimisers.jl
DhairyaLGandhi: Maybe I'm not understanding something clearly but you can't refer to the constant from outside of this file, and have to qualify it even then. And the name of the (kw)arg will be how it is referred to in the constructor as a local variable. Either way it is fine to remove considering its local to this file.

DhairyaLGandhi:
Kwarg only constructors sounds non-julian. Are we using the global constant anywhere else?

src/optimise/optimisers.jl
DhairyaLGandhi: I don't think we need to deprecate it considering its a convenient constructor to have anyway. Requiring an `eps` is really niche enough that we can save users from needing to specify it if they need to specify a container for their parameters.

CarloLucibello:
> I could rename the kwarg ϵ -> eps

let's do this

src/optimise/optimisers.jl
CarloLucibello: I suggest we go keywords only for all arguments, except maybe the learning rate, and don't spend too much time discussing ordering (which is why the reason why I suggest we go keyword only, besides better script clarity and not having to type in the defaults of the args in first positions)

darsnack (approved):

mcabbott (commented):
Should the docstrings mention this, somewhere? Maybe briefly, after they have discussed the more familiar parameters?
```
    ADAM(η = 0.001, β::Tuple = (0.9, 0.999), ε = $EPS)
[ADAM](https://arxiv.org/abs/1412.6980) optimiser.
# Parameters
```

src/optimise/optimisers.jl
```diff
@@ -136,8 +136,8 @@ mutable struct RMSProp <: AbstractOptimiser
   epsilon::Float64
   acc::IdDict
 end
-
-RMSProp(η = 0.001, ρ = 0.9, ϵ = ϵ) = RMSProp(η, ρ, ϵ, IdDict())
+RMSProp(η::Real = 0.001, ρ::Real = 0.9, ϵ::Real = EPS) = RMSProp(η, ρ, ϵ, IdDict())
+RMSProp(η::Real, ρ::Real, acc::IdDict) = RMSProp(η, ρ, EPS, acc)
```
DhairyaLGandhi, line 140: Why

src/optimise/optimisers.jl
```diff
@@ -136,8 +136,8 @@ mutable struct RMSProp <: AbstractOptimiser
   epsilon::Float64
   acc::IdDict
 end
-
-RMSProp(η = 0.001, ρ = 0.9, ϵ = ϵ) = RMSProp(η, ρ, ϵ, IdDict())
+RMSProp(η::Real = 0.001, ρ::Real = 0.9, ϵ::Real = EPS) = RMSProp(η, ρ, ϵ, IdDict())
```
DhairyaLGandhi, line 139: `Real -> Number` would be better. 

Is there an issue with having `eps` be a kwarg consistently across all the optimisers?

cossio: Why `Number`? It's a `Float64` within RMSProp anyway.

I'd say either all args are changed to kwargs, or none of them are. Why eps is singled out as kwarg and the others not?

cossio:
> Should the docstrings mention this, somewhere? Maybe briefly, after they have discussed the more familiar parameters?

Not sure, the docstrings don't mention `IdDict` either.

mcabbott:
The argument both to mention epsilon, and not to make it a special keyword argument, is that we should follow the source:
Screenshot 2022-01-23 at 11 43 46
The IdDict, however, is just an internal detail of how Flux presently keeps track of parameters.

cossio:
Ok I mention eps in the argument list.

cossio:
Can we tag a release with this?

PR #1845, opened 2022-01-22
add token secret for codecov

ToucheSir:
This should resolve the (non-terminating) upload errors on CI and actually let us collect coverage info.


mcabbott (approved):
Not so secret :) but I guess you know what you are doing, it's got to write somewhere...

ToucheSir:
I was pointed to the docs at https://github.com/JuliaGPU/buildkite#using-secrets, and presumably those folks know what they're doing :smile:

PR #1849, opened 2022-01-25
Add RNG support for Dropout/AlphaDropout

### LINKED ISSUE #1617: Allow specification of RNG in Dropout

darsnack:
Adds support for custom RNGs to `Dropout` and `dropout`. For dealing with the GPU, the RNG field is mapped to the corresponding default RNG for CUDA.jl when it is `Random.default_rng()`. All other RNGs will throw an error (as these are not supported by CUDA.jl). Still needs more tests.

Fixes #1617.

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -96,6 +96,9 @@ end
 struct FluxCUDAAdaptor end
 adapt_storage(to::FluxCUDAAdaptor, x) = CUDA.cu(x)
 adapt_storage(to::FluxCUDAAdaptor, x::Zygote.FillArrays.AbstractFill) = CUDA.cu(collect(x))
+adapt_storage(to::FluxCUDAAdaptor, x::Random.GLOBAL_RNG) = CUDA.CURAND.default_rng()
+adapt_storage(to::FluxCUDAAdaptor, x::AbstractRNG) =
+  error("Cannot map RNG of type $(typeof(x)) to GPU. GPU execution only supports Random.default_rng().")
 
 # TODO: figure out the correct design for OneElement
 adapt_storage(to::FluxCUDAAdaptor, x::Zygote.OneElement) = CUDA.cu(collect(x))
@@ -109,6 +112,8 @@ adapt_storage(to::FluxCPUAdaptor, x::Zygote.FillArrays.AbstractFill) = x
 adapt_storage(to::FluxCPUAdaptor, x::T) where T <: CUDA.CUSPARSE.CUDA.CUSPARSE.AbstractCuSparseMatrix = adapt(Array, x)
 adapt_storage(to::FluxCPUAdaptor, x::Zygote.OneElement) = x
 adapt_storage(to::FluxCPUAdaptor, x::AbstractSparseArray) = x
+adapt_storage(to::FluxCPUAdaptor, x::CUDA.CURAND.RNG) = Random.default_rng()
+adapt_storage(to::FluxCPUAdaptor, x::AbstractRNG) = x
 
 Zygote.@adjoint function Array(x::CUDA.CuArray)
   Array(x), d -> (CUDA.cu(d),)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -96,7 +96,11 @@ end
 struct FluxCUDAAdaptor end
 adapt_storage(to::FluxCUDAAdaptor, x) = CUDA.cu(x)
 adapt_storage(to::FluxCUDAAdaptor, x::Zygote.FillArrays.AbstractFill) = CUDA.cu(collect(x))
-adapt_storage(to::FluxCUDAAdaptor, x::Random.TaskLocalRNG) = CUDA.default_rng()
+if VERSION >= v"1.7"
+  adapt_storage(to::FluxCUDAAdaptor, x::Random.TaskLocalRNG) = CUDA.default_rng()
+else
+  adapt_storage(to::FluxCUDAAdaptor, x::Random._GLOBAL_RNG) = CUDA.default_rng()
+end
 adapt_storage(to::FluxCUDAAdaptor, x::CUDA.RNG) = x
 adapt_storage(to::FluxCUDAAdaptor, x::AbstractRNG) =
   error("Cannot map RNG of type $(typeof(x)) to GPU. GPU execution only supports Random.default_rng().")
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -96,7 +96,7 @@ end
 struct FluxCUDAAdaptor end
 adapt_storage(to::FluxCUDAAdaptor, x) = CUDA.cu(x)
 adapt_storage(to::FluxCUDAAdaptor, x::Zygote.FillArrays.AbstractFill) = CUDA.cu(collect(x))
-adapt_storage(to::FluxCUDAAdaptor, x::Random._GLOBAL_RNG) = CUDA.default_rng()
+adapt_storage(to::FluxCUDAAdaptor, x::Random.TaskLocalRNG) = CUDA.default_rng()
 adapt_storage(to::FluxCUDAAdaptor, x::CUDA.RNG) = x
 adapt_storage(to::FluxCUDAAdaptor, x::AbstractRNG) =
   error("Cannot map RNG of type $(typeof(x)) to GPU. GPU execution only supports Random.default_rng().")
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -97,6 +97,7 @@ struct FluxCUDAAdaptor end
 adapt_storage(to::FluxCUDAAdaptor, x) = CUDA.cu(x)
 adapt_storage(to::FluxCUDAAdaptor, x::Zygote.FillArrays.AbstractFill) = CUDA.cu(collect(x))
 adapt_storage(to::FluxCUDAAdaptor, x::Random._GLOBAL_RNG) = CUDA.default_rng()
+adapt_storage(to::FluxCUDAAdaptor, x::CUDA.RNG) = x
 adapt_storage(to::FluxCUDAAdaptor, x::AbstractRNG) =
   error("Cannot map RNG of type $(typeof(x)) to GPU. GPU execution only supports Random.default_rng().")
 
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -96,7 +96,7 @@ end
 struct FluxCUDAAdaptor end
 adapt_storage(to::FluxCUDAAdaptor, x) = CUDA.cu(x)
 adapt_storage(to::FluxCUDAAdaptor, x::Zygote.FillArrays.AbstractFill) = CUDA.cu(collect(x))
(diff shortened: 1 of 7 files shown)


src/functor.jl
```diff
@@ -96,6 +96,9 @@ end
 struct FluxCUDAAdaptor end
 adapt_storage(to::FluxCUDAAdaptor, x) = CUDA.cu(x)
 adapt_storage(to::FluxCUDAAdaptor, x::Zygote.FillArrays.AbstractFill) = CUDA.cu(collect(x))
+adapt_storage(to::FluxCUDAAdaptor, x::Random.GLOBAL_RNG) = CUDA.CURAND.default_rng()
```
ToucheSir, line 99:
```suggestion
adapt_storage(to::FluxCUDAAdaptor, x::Random._GLOBAL_RNG) = CUDA.default_rng()
```

src/functor.jl
```diff
@@ -109,6 +112,8 @@ adapt_storage(to::FluxCPUAdaptor, x::Zygote.FillArrays.AbstractFill) = x
 adapt_storage(to::FluxCPUAdaptor, x::T) where T <: CUDA.CUSPARSE.CUDA.CUSPARSE.AbstractCuSparseMatrix = adapt(Array, x)
 adapt_storage(to::FluxCPUAdaptor, x::Zygote.OneElement) = x
 adapt_storage(to::FluxCPUAdaptor, x::AbstractSparseArray) = x
+adapt_storage(to::FluxCPUAdaptor, x::CUDA.CURAND.RNG) = Random.default_rng()
```
ToucheSir, line 115:
```suggestion
adapt_storage(to::FluxCPUAdaptor, x::CUDA.RNG) = Random.default_rng()
```
ref. https://github.com/JuliaGPU/CUDA.jl/blob/master/src/random.jl

ToucheSir (changes requested):
See if these help with the tests. One outstanding question is how task local RNGs factor into this, thoughts from the floor welcome.

darsnack:
I was under the impression that `default_rng` was task local.

ToucheSir:
Do you mean `CUDA.default_rng`? I believe it's used for both the global RNG and task local ones.

darsnack:
`Random.default_rng()` will expand to `default_rng(Threads.threadid())`. Not sure about CUDA RNGs. Since using a CUDA RNG involves calling `gpu` and that must come after `CUDA.device!(...)`, it's possible it works depending on `CUDA.default_rng`.

Probably the types for dispatch like `Random._GLOBAL_RNG` and `CUDA.RNG` are not the correct things to use.

ToucheSir:
The big headache is that neither of Base's default RNGs (global and task local) use an algorithm available in CUDA and vice versa. Random123.jl does have a Philox implementation, but expecting users to depend on it just to make their models GPU compatible is a stretch. The implication is that converting RNGs between devices completely destroys the internal state, whereas with parameters it keeps as much semantically meaningful internal state as possible.

@Ablaom if I understand https://github.com/FluxML/MLJFlux.jl/pull/166 correctly, you only need the ability to specify a consistent RNG for the whole model and not individual layers, right? One way to get around the challenges of storing the RNG in the layer would be to use Cassette-style contextual dispatch to hook `default_rng`. Then whenever `Dropout` calls `default_rng()`, it'll receive the user-specified RNG instead of the global or task local one. A refinement of this idea would be to define a higher-level function `Flux/NNlib.rng_from_array(::AbstractArray)` which `Dropout` calls. This can be overloaded for `CuArrays` and would also be safer to use contextual dispatch on.

darsnack:
Just for clarity, are you suggesting:
```julia
function dropout_mask(x, p; dims=:)
  y = rand!(rng_from_array(x), similar(x, _dropout_shape(x, dims)))
  return y
end
```

ablaom:
> you only need the ability to specify a consistent RNG for the whole model and not individual layers, right?

Yes.

darsnack:
Currently, CUDA.jl only exposes a single RNG. I think it is reasonable that `Random.default_rng` and `CUDA.default_rng` can be swapped back and forth, since these are the default cases AKA the cases where users aren't specifying an RNG at all. Currently, this PR conservatively throws an error for any other type of RNG, since they only exist on the CPU side of things.

Let's say the low level contextual approach is like
```julia
Flux.rng_from_array(x) = MersenneTwister(123)
m = Dropout(0.1)
m(rand(Float32, 10)) # works, uses MersenneTwister
m(CUDA.rand(10)) # errors due to Flux.rng_from_array(::CuArray) inside Flux.jl
```
Then you can have block like
```julia
m = Dropout(0.1)
withrng(MersenneTwister(123)) do
  m(rand(Float32, 10))
end
```

But I think this becomes trickier for packages that want the RNG on construction instead of the forward pass. Of course, if you have a wrapper around the model, then the wrapper forward pass could use `withrng`, but I don't know if that will always be the case.

ToucheSir:
Kyle beat me to it, but thoughts below.

> Just for clarity, are you suggesting (or some equivalent):
> ...

I was thinking a couple layers higher. The addition of `dropout(rng, x)` in this PR make a lot of sense, so the change would be from this:
```julia
dropout(x, p; kwargs...) = dropout(Random.default_rng(), x, p; kwargs...)
dropout(x::CuArray, p; kwargs...) = dropout(CUDA.CURAND.default_rng(), x, p; kwargs...)
```
to this:
```julia
rng_from_array(_) = Random.default_rng()
rng_from_array(::CuArray) = CUDA.default_rng()

dropout(x, p; kwargs...) = dropout(rng_from_array(x), x, p; kwargs...)
```
Likewise for `(d::Dropout)(x)`. Then the override would look like:
```julia
using Cassette

Cassette.@context RNGOverrideCtx
Cassette.overdub(ctx::RNGOverrideCtx, ::typeof(Flux.rng_from_array), _) = rng.metadata

with_rng_override(f, rng) = Cassette.overdub(RNGOverrideCtx(metadata=rng), f)

...
model, x, rng = ...
with_rng_override(rng) do
  model(x)
end
```

ToucheSir:
> I think it is reasonable that `Random.default_rng` and `CUDA.default_rng` can be swapped back and forth, since these are the default cases AKA the cases where users aren't specifying an RNG at all.

To clarify, are you referring to any instance of the types those functions return or to the returned instances themselves? The latter would be ideal, but it's not clear to me whether the return value of `default_rng` is consistent (that is, `===`) over time. I also don't understand enough about the task system to know whether moving a model constructed with the global RNG to a task would affect performance down the line, as IIRC the global RNG requires locking to be thread-safe.

darsnack:
That implementation makes sense to me, and it would be safer. One thing we could do to the current implementation is
```julia
dropout_mask(rng::CUDA.RNG, x::CuArray, p; kwargs...) = _dropout_mask(rng, x, p; kwargs...)
dropout_mask(rng, x::CuArray, p; kwargs...) = ArgumentError("dropout_mask only supports CUDA.RNG when x isa CuArray.")
dropout_mask(rng, x, p; kwargs...) = _dropout_mask(rng, x, p; kwargs...)
function _dropout_mask(rng, x, p; dims=:)
  y = rand!(rng, similar(x, _dropout_shape(x, dims)))
  return y
end
```
I think that will raise both approaches to the same level of safety without considering task-related issues.

> To clarify, are you referring to any instance of the types those functions return or to the returned instances themselves?

The latter, I think though I am similarly unsure about the task system and the interaction with the RNGs.

darsnack:
Okay, now only any `TaskLocalRNG` and `CUDA.RNG` are considered "swappable" though the state might be destroyed.

Trying to move `Random.GLOBAL_RNG` will error right now, but we could warn instead. Presumably, you wouldn't be using this with tasks, so you'd be happy to accept `CUDA.RNG`.

test/layers/normalisation.jl
```diff
@@ -4,85 +4,103 @@ using Zygote: pullback
...
+
+    testmode!(m, false)
+    y = evalwgrad(m, x)
+    @test mean(y) ≈ 0 atol=0.1
+    @test var(y) ≈ 1 atol=0.1
+
+    # Known good value ranges
+    # Values taken from https://github.com/pytorch/pytorch/blob/v1.10.0/test/cpp/api/modules.cpp#L1337-L1338
+    x = ones(100)
+    @test 40 < sum(evalwgrad(m, x)) < 130
```
ToucheSir, line 95: These bounds were copied from the PyTorch test suite. IIUC PyTorch's default CPU number generator is a mersenne twister, but maybe the seed is less than ideal?

darsnack:
Seems like `AlphaDropout` is extremely sensitive to the RNG. Do we still want to include it in this PR?

test/layers/normalisation.jl
```diff
@@ -4,85 +4,108 @@ using Zygote: pullback
...
+    @test var(y) ≈ 1 atol=0.2
+
+    # Known good value ranges
+    # Values taken from https://github.com/pytorch/pytorch/blob/v1.10.0/test/cpp/api/modules.cpp#L1337-L1338
+    x = ones(100)
+    if isempty(rng_kwargs)
+      @test 40 < sum(evalwgrad(m, x)) < 130
+    else
+      # FIXME: this breaks spuriously for MersenneTwister
+      @test_skip 40 < sum(evalwgrad(m, x)) < 130
```
ToucheSir, line 99: This seems reasonable to me, opened https://github.com/FluxML/Flux.jl/issues/1851 for a proper test. Anything left assuming tests pass?

darsnack:
Screen Shot 2022-01-27 at 11 15 06 AM

😱

darsnack:
It looks like we'll need a `rng_from_array` anyways to avoid making the codebase littered with `if VERSION ...`

ToucheSir:
😱indeed. I thought the current code would be generic enough to handle 1.6, but it's a good enough opportunity as any to introduce the cleaner API.

ToucheSir (approved):
LGTM

src/layers/normalise.jl
```diff
@@ -28,49 +31,63 @@ automatically managed using the [`Dropout`](@ref) layer instead of the
...
+dropout_mask(rng, x, p; kwargs...) = _dropout_mask(rng, x, p; kwargs...)
+function _dropout_mask(rng, x, p; dims=:)
+  y = rand!(rng, similar(x, _dropout_shape(x, dims)))
   y .= _dropout_kernel.(y, p, 1 - p)
   return y
 end
 
 """
-    Dropout(p; dims=:)
+    Dropout(p; dims=:, rng = default_rng())
```
darsnack, line 58:
```suggestion
    Dropout(p; dims=:, rng = rng_from_array())
```

src/layers/normalise.jl
```diff
@@ -83,7 +100,7 @@ function Base.show(io::IO, d::Dropout)
 end
 
 """
-    AlphaDropout(p)
+    AlphaDropout(p; rng = default_rng())
```
darsnack, line 103:
```suggestion
    AlphaDropout(p; rng = rng_from_array())
```

PR #1855, opened 2022-02-02
Block parameters for RNNs (for cudnn paths)

mkschleg:
This changes the RNN implementation such that it stores all its parameters in a block. We include helper views to access specific parts of the weight matrix more easily. This is from some rumblings I've been doing to get make cudnn paths possible for Flux RNNs. The idea would be to port all these changes to the other recurrent cells, but that should happen after we are happy with the move.

 Some points to consider:
- Gradients are only calculated for W and state0
- The views have to be recreated on the forward pass so gradients can be calculated, 
   - we could instead do something else for this, but I thought this was easier at first.
- In cudnn, they also block weights for multiple RNN layers in a row. Likely we may want to do this as well.
- You can see some of the fuller things necessary for a forward pass for cudnn [here](https://github.com/mkschleg/FluxRNNs.jl/blob/c05f5b5b251e57441e5ec25342ebb000db8d07d2/src/FluxRNNs.jl#L154).

I also cleaned up some of the testing code for recurrent cells. Because:
1. they were incompatible with this change
2. they are great tests, but scary at first look.

Also a test currently fails (the eltype test), which I thought we were wanting to move away from anyway. But maybe I'm misremembering.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/.github/workflows/pr_comment.yml
+++ b/.github/workflows/pr_comment.yml
@@ -8,7 +8,7 @@ jobs:
     steps:
       - name: Create PR comment
         if: github.event_name == 'pull_request' && github.repository == github.event.pull_request.head.repo.full_name && github.event.label.name == 'documentation' # if this is a pull request build AND the pull request is NOT made from a fork
-        uses: thollander/actions-comment-pull-request@dadb7667129e23f12ca3925c90dc5cd7121ab57e
+        uses: thollander/actions-comment-pull-request@d61db783da9abefc3437960d0cce08552c7c004f
         with:
           message: 'Once the build has completed, you can preview any updated documentation at this URL: https://fluxml.ai/Flux.jl/previews/PR${{ github.event.number }}/ in ~20 minutes'
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


mkschleg:
Errors are in utils tests. Maybe I should comment those out to just have recurrent tests?

mkschleg:
Something else to think about. When using CUDNN the weight matrix is not organized like it is in the cpu implementation. I  have yet to think of a good solution for this. The best I can think of is not ensuring how W is organized in docs/user facing details. We can have a forward facing api (through views like I've done here) and make guarantees on that.

mkschleg:
Sorry. Dumping more info/thoughts here. There is a lot of metadata required to pass to cudnn. I'm wondering if we should have a flux side metadata struct to deal with these particularities, especially if we decide to make multi-layer RNNs apart of the same interface. But maybe these are taken care of in the parametric types.

ToucheSir:
I haven't had time yet to go through this PR, but for:

> I'm wondering if we should have a flux side metadata struct to deal with these particularities

If the metadata is CUDA-specific, one place to put it would be NNlibCUDA. It would also be cool to have a multi-layer RNN interface, but given the complexity of that I wouldn't worry about solving it in this PR :)

src/layers/recurrent.jl
```diff
@@ -105,31 +105,57 @@ end
...
+          out)
+
+function (m::RNNCell)(h, x)
+  # σ, Wi, Wh, b = m.σ, m.Wi, m.Wh, m.b
+  σ, W = m.σ, m.W
+  in, out = size(x, 1), size(h, 1)
+  
+  Wi = view(W, :, 1:in)
+  Wh = view(W, :, (in+1):(out+in))
+  b = view(W, :, size(W, 2))
```
ToucheSir, line 140: Unfortunately, `view` is pretty inefficient at present because AD will allocate a full-sized zeros buffer for each call. You may want to explore something like https://github.com/FluxML/Flux.jl/blob/v0.12.9/src/layers/recurrent.jl#L6-L18 here.

mkschleg: Oh nice! I hadn't seen that!

mkschleg:
Updated to use efficient view adjoint based on multigate. I also changed some things to better represent Flux's current state for recurrent layers. I was behind on some things.

src/layers/recurrent.jl
```diff
@@ -105,31 +105,69 @@ end
...
-struct RNNCell{F,A,V,S}
+struct RNNCell{F, VA, VB, A, S}
   σ::F
-  Wi::A
-  Wh::A
-  b::V
+  Wi::VA
+  Wh::VA
+  b::VB
+  W::A
```
ToucheSir, line 113: Speculative, but instead of storing `W` as a field, I wonder if this could just call `parent([Wi|Wh|b])`. Possibly too janky?

test/layers/recurrent.jl
```diff
@@ -5,18 +5,20 @@
...
-    bptt = gradient(Wh -> sum(tanh.(rnn.cell.Wi * seq[3] + Wh *
-                                  tanh.(rnn.cell.Wi * seq[2] + Wh *
-                                        tanh.(rnn.cell.Wi * seq[1] +
-                                            Wh * rnn.cell.state0
-                                        + rnn.cell.b)
-                                  + rnn.cell.b)
-                            + rnn.cell.b)),
-                    rnn.cell.Wh)
-    @test grads_seq[rnn.cell.Wh] ≈ bptt[1]
+      bptt = gradient(rnn.cell.W) do W
```
ToucheSir, line 11: Some formatting and indentation funniness to work out after the implementation is locked in. These new test cases look much cleaner though!

src/layers/recurrent.jl
mkschleg: That feels a bit _too_ janky for my tastes. I agree the struct is a bit cluttered. We could override getproperty for these, but that might also be not great. I think having these is preferable to not. 

One thing we maybe should do with these changes is rename `W` to `weight` so it is consistent with other parts of Flux.

mkschleg:
I just realized this new version doesn't have an ability to not have a bias unit. I'm not sure how to solve that w/o a flag of some kind.

ToucheSir:
The bias thing is an implementation pain for other frameworks as well. I recall either some documentation or a discussion thread about how much Keras had to bend over backwards to make it work.

mkschleg:
I dealt with the bias issue by dispatching on constructors and the `rnn_weights` function. So now we support no bias with `initb=Flux.Zeros`. I also added a test case for this based on BPTT-2D just to make sure. 

I haven't tested on GPU. But I expect it should work.

CarloLucibello:
Are we sure that using `cudnn` is worth all this trouble? Do we have benchmarks?

If we go with this, we should make sure that using views interact nicely with `|> gpu` movement (i.e. a view of an array becomes the view of the gpu-moved array, from some rapid check this is the case) and gradients (which seems to be handled and tested already in this PR)

Do we need to store the views in the struct at all? Is that for retrocompatibility? Can we overload getproperty instead?

darsnack:
It seems that a lot of the complexity here comes from taking the storage cuDNN wants (packed array block) and de-sugaring it into what we like (separate blocks). Would it not be easier to reverse the logic?

We store separate fields for each semantic block. The constructor by default will make these fields views of a monolithic block. The individual views are what we take gradients w.r.t. and writes to them should handle the accumulation and opt state correctly? We can have a level of indirection like `_cellforward(x, h, Wi, Wh, ...)`.

For cuDNN and functoring, we define a `packcell` function that accepts the individual blocks and returns a monolithic array. When all the blocks are views, we check that they share the same parent then return `parent`. Otherwise, we naively concatenate them (though this path should never be taken in practice). The cuDNN path can then call this `packcell` function to get the monolithic array it needs, and its gradient rule can handle distributing the gradient to the views. It would do this by having a separate dispatch for `_cellforward`?

PR #1856, opened 2022-02-03
Fix type-stability for normalization layers

pxl-th:
This PR fixes type-stability for normalization layers.
This also makes ResNet model type-stable (as well as others).

While there is [PR](https://github.com/FluxML/Flux.jl/pull/1509) that reworks normalization layers, it is not clear what is the status of it.
@ToucheSir has also suggested that functional part should be moved into NNlib.jl, so I'm not sure if this PR should be accepted in the first place...
But at least we can use it to look at what improvements type-stability can bring in this case...

~~Also I feel it is not the cleanest solution, as it essentially computes the output type `o::_basetype(typeof(x)){O, N}`.~~

### Benchmark:

```julia
using Test
using Flux
using BenchmarkTools

function main()
    x = rand(Float32, 28, 28, 64, 4)
    bn = BatchNorm(64)
    θ = Flux.params(bn)

    @info "forward"
    @time bn(x)
    @btime $bn($x)

    @info "grad"
    @time gradient(θ) do
        sum(bn(x))
    end
    @btime gradient($θ) do
        sum($bn($x))
    end

    @inferred bn(x) # master fails here
end
```

Before:

```julia
[ Info: forward
  0.254515 seconds (856.55 k allocations: 43.045 MiB, 98.02% compilation time)
  911.059 μs (29 allocations: 785.33 KiB)
[ Info: grad
  4.050493 seconds (12.31 M allocations: 635.845 MiB, 3.65% gc time, 99.81% compilation time)
  2.838 ms (804 allocations: 13.05 MiB)
```

After:

```julia
[ Info: forward
  0.000917 seconds (15 allocations: 1.532 MiB)
  455.609 μs (15 allocations: 1.53 MiB)
[ Info: grad
  2.245501 seconds (7.16 M allocations: 372.776 MiB, 3.75% gc time, 99.17% compilation time)
  2.337 ms (694 allocations: 12.28 MiB)
```

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -188,13 +188,6 @@ function Base.show(io::IO, l::LayerNorm)
   print(io, ")")
 end
 
-_maybe_promote_type(::Type{T1}, ::Type{T2}) where {T1, T2} = promote_type(T1, T2)
-_maybe_promote_type(::Type{Nothing}, ::Type{T2}) where T2 = T2
-_maybe_promote_type(::Type{T1}, ::Type{Nothing}) where T1 = T1
-
-_maybe_eltype(::Type{T}) where T <: AbstractArray = eltype(T)
-_maybe_eltype(::Type{Nothing}) = Nothing
-
 # For InstanceNorm, GroupNorm, and BatchNorm.
 # Compute the statistics on the slices specified by reduce_dims.
 # reduce_dims=[1,...,N-2,N] for BatchNorm
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -146,13 +146,13 @@ testmode!(m::AlphaDropout, mode=true) =
     LayerNorm(sz, λ=identity; affine=true, ϵ=1fe-5)
 
 A [normalisation layer](https://arxiv.org/abs/1607.06450) designed to be
-used with recurrent hidden states. 
-The argument `sz` should be an integer or a tuple of integers. 
-In the forward pass, the layer normalises the mean and standard 
+used with recurrent hidden states.
+The argument `sz` should be an integer or a tuple of integers.
+In the forward pass, the layer normalises the mean and standard
 deviation of the input, the applied the elementwise activation `λ`.
 The input is normalised along the first `length(sz)` dimensions
 for tuple `sz`, along the first dimension for integer `sz`.
-The input  is expected to have first dimensions' size equal to `sz`. 
+The input  is expected to have first dimensions' size equal to `sz`.
 
 If `affine=true` also applies a learnable shift and rescaling
 as in the [`Diagonal`](@ref) layer.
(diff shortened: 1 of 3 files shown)


src/layers/normalise.jl
```diff
@@ -188,39 +188,78 @@ function Base.show(io::IO, l::LayerNorm)
...
-  if hasaffine(l)
-    γ = reshape(l.γ, affine_shape)
-    β = reshape(l.β, affine_shape)
-    return l.λ.(γ .* (x .- μ) ./ sqrt.(σ² .+ l.ϵ) .+ β)
-  else
-    return l.λ.((x .- μ) ./ sqrt.(σ² .+ l.ϵ))
-  end
+
+  O = _promote_to_output(l, x)
+  o::_basetype(typeof(x)){O, N} = ((x .- μ) ./ sqrt.(σ² .+ l.ϵ))
```
ToucheSir, line 238: I wonder if this could be done via an inference barrier (i.e. helper function) instead of having to rely on an assertion. Thoughts?

src/layers/normalise.jl
```diff
@@ -188,39 +188,78 @@ function Base.show(io::IO, l::LayerNorm)
...
+  mtm = bn.momentum
+  res_mtm = one(V) - mtm
+  m = prod(size(x, i) for i in reduce_dims)
+
+  μnew = vec(N ∈ reduce_dims ? μ : mean(μ, dims=N))
+  σ²new = vec(N ∈ reduce_dims ? σ² : mean(σ², dims=N))
+
+  bn.μ = res_mtm .* bn.μ .+ mtm .* μnew
+  bn.σ² = res_mtm .* bn.σ² .+ mtm .* (m / (m - one(V))) .* σ²new
+  nothing
```
CarloLucibello, line 259:
```suggestion
  return nothing
```

src/layers/normalise.jl
pxl-th: Thanks, @ToucheSir! That worked!

ToucheSir (approved):
How do the times look now? More or less the same?

pxl-th:
> How do the times look now? More or less the same?

Yes, the times are the same.

```julia
[ Info: forward
  0.000813 seconds (15 allocations: 1.532 MiB)
  461.305 μs (15 allocations: 1.53 MiB)
[ Info: grad
  1.970701 seconds (7.04 M allocations: 366.512 MiB, 4.10% gc time, 99.81% compilation time)
  2.209 ms (634 allocations: 12.28 MiB)
```

ToucheSir:
Magnificent, thanks!

P.S.
> @ToucheSir has also suggested that functional part should be moved into NNlib.jl, so I'm not sure if this PR should be accepted in the first place...

For anyone reading this comment, here's a "good first issue" ;)

PR #1859, opened 2022-02-04
Use ProgressLogging instead of Juno

### LINKED ISSUE #1779: Remove Juno.jl dependency

darsnack:
This replaces Juno with ProgressLogging. Closes #1807 and fixes #1779.

This overcomes the `_linindex` issue by manually updating the progress using `@logprogress` and calling `length` instead. In the case of infinite iterators, the code will still run but we don't have a clear way to measure progress (so the bar will jump from 0 to 100).

One possible issue is infinite iterators that improperly implement the Base iterator interface. Right now this PR takes the "implement the interface correctly" stance, but we could also try/catch on `length`.

### PR Checklist

- [x] ~~Tests are added~~
- [x] ~~Entry in NEWS.md~~


CarloLucibello:
Compat bound missing

CarloLucibello (approved):

src/optimise/train.jl
```diff
@@ -120,6 +121,7 @@ function train!(loss, ps, data, opt; cb = () -> ())
         rethrow(ex)
       end
     end
+    @logprogress i / n
```
ToucheSir, line 124: How does `@logprogress` handle `Inf` here? Is it possible to make the progress bar indeterminate if n == 0?

darsnack: The progress will just show up as 0% until the end of the loop then jump to 100%. I think `@logprogress` accepts NaN or Inf as part of its API to indicate a "no-op."

darsnack: There is no API for indeterminate, just lots of discussion on the ProgressLogging issues. It's unclear to me that we'd be able to take advantage of the proposed APIs too, because we don't know the criteria used to measure convergence.

ToucheSir: I was thinking something along the lines of what tqdm does (just report `it/s` or `s/it`), but good to know.

ToucheSir (approved):

src/optimise/train.jl
darsnack: Yeah maybe we can use something like this eventually: https://github.com/JuliaLogging/ProgressLogging.jl/issues/23. Right now I think it requires a specific frontend for ProgressLogging to work how we want it. It doesn't work with the VSCode frontend.

pfitzseb: > There is no API for indeterminate

https://github.com/JuliaLogging/ProgressLogging.jl/blob/0e7933005233722d6214b0debe3316c82b4d14a7/src/ProgressLogging.jl#L79

darsnack: Thanks, I thought `nothing` has the same behavior as `NaN` (at least in VS Code).

PR #1861, opened 2022-02-05
Speed up `onehotbatch`

### LINKED ISSUE #1844: onehotbatch performance
After a [discourse thread](https://discourse.julialang.org/t/flux-1d-convolutions-on-genomic-data/74874), I was recommended to create a github issue about the performance of `onehotbatch`. Let's use the following MWE:

```julia
using Flux
using BenchmarkTools

const bases_dna = ['A', 'C', 'G', 'T']

function ohe_custom(sequence)
    return collect(sequence) .== permutedims(bases_dna)
end

function ohe_flux(sequence)
    return Flux.onehotbatch(collect(sequence), bases_dna)
end

sequence = "CCGAGGGCTATGGTTTGGAAGTTAGAACCCTGGGGCTTCTCGCGGA"
```

So right now the dimensions of the two functions are different (transposed), but that can easily be changed with a `permutedims` applied to any one of them, otherwise they return the same onehot encoded matrix. So far, so good. However, when benchmarking them, we find the following:

```julia
@btime ohe_custom(sequence);
# output: 550.514 ns (5 allocations: 464 bytes)
```
and
```julia
@btime ohe_flux(sequence);
# output: 69.274 μs (374 allocations: 17.30 KiB)
```

As we can see, `ohe_flux` is more than 100 times slower than `ohe_custom` and with 70 times more allocations. 

Another minor detail is the size of the different outputs:
```julia
Base.summarysize(sequence)
# output: 54
```
```julia
Base.summarysize(ohe_custom(sequence))
# output: 96
```
```julia
Base.summarysize(ohe_flux(sequence))
# output: 232
```

mcabbott:
Closes #1844. Before:
```julia
julia> @btime ohe_custom($sequence);  # from #1844
  min 908.784 ns, mean 962.810 ns (5 allocations, 464 bytes)

julia> @btime ohe_flux($sequence);
  min 67.083 μs, mean 69.835 μs (374 allocations, 17.30 KiB)

julia> seq = String(rand('a':'z', 10));

julia> @btime Flux.onehotbatch($seq, 'a':'z');
  min 14.750 μs, mean 15.071 μs (84 allocations, 3.84 KiB)
```
After:
```julia
julia> @btime Flux.onehotbatch($sequence, $bases_dna);  # vector of labels
  min 857.246 ns, mean 920.303 ns (4 allocations, 528 bytes)

julia> @btime Flux.onehotbatch($sequence, $(Tuple(bases_dna)));  # faster if you make the tuple
  min 586.572 ns, mean 639.020 ns (3 allocations, 496 bytes)

julia> seq = String(rand('a':'z', 10));

julia> @btime Flux.onehotbatch($seq, 'a':'z');  # 26 classes, quite a long tuple
  min 1.883 μs, mean 1.946 μs (5 allocations, 480 bytes)

julia> @btime Flux.onehotbatch($seq, $(Tuple('a':'z')));
  min 717.391 ns, mean 774.093 ns (3 allocations, 208 bytes)
```
There is probably more that could be done, e.g. to make `batch` type-stable & save a copy, but this is a start.

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -163,8 +163,8 @@ If `xs` has more dimensions, `M = ndims(xs) > 1`, then the result is an
 `AbstractArray{Bool, M+1}` which is one-hot along the first dimension, 
 i.e. `result[:, k...] == onehot(xs[k...], labels)`.
 
-Notice that `xs::String` is fine, it does not need to be collected.
-For less than 32 classes, `labels::Tuple` will give best performance.
+Note that `xs` can be any iterable, such as a string. And that using a tuple
+for `labels` will often speed up construction, certainly for less than 32 classes.
 
 # Examples
 ```jldoctest
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -131,16 +131,23 @@ julia> hcat(αβγ...)  # preserves sparsity
 ```
 """
 function onehot(x, labels)
-  i = something(findfirst(isequal(x), labels), 0)
-  i > 0 || error("Value $x is not in labels")
+  i = _findval(x, labels)
+  isnothing(i) && error("Value $x is not in labels")
   OneHotVector{UInt32, length(labels)}(i)
 end
 
 function onehot(x, labels, default)
-  i = something(findfirst(isequal(x), labels), 0)
-  i > 0 || return onehot(default, labels)
+  i = _findval(x, labels)
+  isnothing(i) && return onehot(default, labels)
   OneHotVector{UInt32, length(labels)}(i)
 end
+
+_findval(val, labels) = findfirst(isequal(val), labels)
+# Fast unrolled method for tuples:
+function _findval(val, labels::Tuple, i::Integer=1)
+  ifelse(isequal(val, first(labels)), i, _findval(val, Base.tail(labels), i+1))
+end
+_findval(val, labels::Tuple{}, i::Integer) = nothing
 
 """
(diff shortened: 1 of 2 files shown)


src/onehot.jl
```diff
@@ -156,9 +163,12 @@ If `xs` has more dimensions, `M = ndims(xs) > 1`, then the result is an
 `AbstractArray{Bool, M+1}` which is one-hot along the first dimension, 
 i.e. `result[:, k...] == onehot(xs[k...], labels)`.
 
+Notice that `xs::String` is fine, it does not need to be collected.
+For less than 32 classes, `labels::Tuple` will give best performance.
```
CarloLucibello, line 167:
```suggestion

For less than 32 classes, `labels::Tuple` will give the best construction performance.
```

src/onehot.jl
```diff
@@ -156,9 +163,12 @@ If `xs` has more dimensions, `M = ndims(xs) > 1`, then the result is an
 `AbstractArray{Bool, M+1}` which is one-hot along the first dimension, 
 i.e. `result[:, k...] == onehot(xs[k...], labels)`.
 
+Notice that `xs::String` is fine, it does not need to be collected.
```
CarloLucibello, line 166:
```suggestion
Notice that `xs` can be any iterable, e.g. a `String`, it does not need to be collected into an array beforehand.
```

CarloLucibello (approved):

PR #1862, opened 2022-02-05
Simplify `trainable`, `functor` and `Parallel`

### LINKED ISSUE #1685: `Parallel` edge-cases

### LINKED ISSUE #1857: `Chain` forgets names under `fmap`
From https://github.com/FluxML/Optimisers.jl/pull/49#issuecomment-1028165105, this isn't great:

```julia
julia> fmap(identity, Chain(x=sin, y=cos))
Chain(sin, cos)
```
The culprit is this `functor` method, which could be extended to deal with this. 

https://github.com/FluxML/Flux.jl/blob/fe803a1a706eb7ecbbf46f7c7e6c20b01847146d/src/layers/basic.jl#L44

Or we could delete it, and let Chain be more like Parallel:

```julia
julia> Flux.functor(Chain(x=sin, y=cos))
((x = sin, y = cos), Flux.var"#154#155"())

julia> Flux.functor(Parallel(vcat, (x=sin, y=cos)))
((connection = vcat, layers = (x = sin, y = cos)), Flux.var"#178#179"())
```

Then we would want delete the inner constructor for Chain, so that it accepts a tuple, `Chain((sin, cos)) == Chain(sin, cos)`. That would line up with https://github.com/FluxML/Flux.jl/pull/1809 allowing `Chain([sin, cos])` as a type-unstable variant.

mcabbott:
This does a few things to do with Functors / Optimisers:
* Changes all uses of `trainable` to return a NamedTuple, which is what Optimisers now wants.
* Removes the use of `trainable` on Parallel, since all fields are trainable, the layers are just one wrapper deeper.
* Removes the inner constructor from `Chain`, so that it can simply be `@functor Chain`. Fixes #1857 . 
* Likewise removes the inner constructor from `Maxout`.
* Adjusts the `show` code not to use `trainable`. This is most of why all these changes are in one PR.

The downside of no longer hiding the Tuple inside Chain from Functors.jl is that `fcollect` and hence `Flux.modules` will see it. ~~This is why the tests fail. I'm not too sure what this is for, and whether this matters.~~ Tests updated to allow this.

The PR also changes Parallel to call its `connection` exactly once, always. And to allow, with N layers, either 1 input or exactly N inputs. It used to zip, but (IMO) allowing N-1 inputs just seems like a bug magnet. These can be separated if anyone feels strongly. Fixes #1685, closes #1698.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -409,8 +409,8 @@ end
 Create a `Parallel` layer that passes an input array to each path in
 `layers`, before reducing the output with `connection`.
 
-Called with one input `x`, this is equivalent to `reduce(connection, [l(x) for l in layers])`.
-If called with multiple inputs, they are `zip`ped with the layers, thus `Parallel(+, f, g)(x, y) = f(x) + g(y)`.
+Called with one input `x`, this is equivalent to `connection([l(x) for l in layers]...)`.
+If called with multiple inputs, one is passed to each layer, thus `Parallel(+, f, g)(x, y) = f(x) + g(y)`.
 
 Like [`Chain`](@ref), its sub-layers may be given names using the keyword constructor.
 These can be accessed by indexing: `m[1] == m[:name]` is the first layer.
@@ -445,7 +445,7 @@ julia> model2[:β] == model2[2]
 true
 ```
 """
-struct Parallel{F, T}
+struct Parallel{F, T<:Union{Tuple, NamedTuple}}
   connection::F
   layers::T
 end
@@ -455,24 +455,30 @@ function Parallel(connection; kw...)
   layers = NamedTuple(kw)
   if :layers in Base.keys(layers) || :connection in Base.keys(layers)
     throw(ArgumentError("a Parallel layer cannot have a named sub-layer called `connection` or `layers`"))
-  elseif isempty(layers)
-    Parallel(connection, ())
   end
+  isempty(layers) && return Parallel(connection, ())
   Parallel(connection, layers)
 end
 
 @functor Parallel
 
-(m::Parallel)(x) = mapreduce(f -> f(x), m.connection, Tuple(m.layers))
-(m::Parallel)(xs...) = mapreduce((f, x) -> f(x), m.connection, Tuple(m.layers), xs)
+(m::Parallel)(x) = m.connection(map(f -> f(x), Tuple(m.layers))...)
 (m::Parallel)(xs::Tuple) = m(xs...)
+function (m::Parallel)(xs...)
+  nl = length(m.layers)
+  nx = length(xs)
+  if nl != nx
+    throw(ArgumentError("Parallel with $nl sub-layers can take one input or $nl inputs, but got $nx inputs"))
+  end
+  m.connection(map(|>, xs, Tuple(m.layers))...)
+end
 
 Base.getindex(m::Parallel, i) = m.layers[i]
-Base.getindex(m::Parallel, i::AbstractVector) = Parallel(m.connection, m.layers[i]...)
+Base.getindex(m::Parallel, i::AbstractVector) = Parallel(m.connection, m.layers[i])
+Base.getindex(m::Parallel{<:Any, <:NamedTuple}, i::AbstractVector) =
+  Parallel(m.connection, NamedTuple{Base.keys(m)[i]}(Tuple(m.layers)[i]))
 
 Base.keys(m::Parallel) = Base.keys(getfield(m, :layers))
-
-trainable(m::Parallel) = (m.connection, m.layers...)
 
 function Base.show(io::IO, m::Parallel)
   print(io, "Parallel(", m.connection, ", ")
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -28,29 +28,30 @@ julia> m2(x) == (m2[:dec] ∘ m2[:enc])(x)
 true
 ```
 """
-struct Chain{T}
+struct Chain{T<:Union{Tuple, NamedTuple}}
   layers::T
-  Chain(xs...) = new{typeof(xs)}(xs)
-  function Chain(; kw...)
-    :layers in Base.keys(kw) && throw(ArgumentError("a Chain cannot have a named layer called `layers`"))
-    isempty(kw) && return new{Tuple{}}(())
-    new{typeof(values(kw))}(values(kw))
-  end
+end
+
+Chain(xs...) = Chain(xs)
+function Chain(; kw...)
+  :layers in Base.keys(kw) && throw(ArgumentError("a Chain cannot have a named layer called `layers`"))
+  isempty(kw) && return Chain(())
+  Chain(values(kw))
 end
 
 @forward Chain.layers Base.getindex, Base.length, Base.first, Base.last,
   Base.iterate, Base.lastindex, Base.keys
 
-functor(::Type{<:Chain}, c) = c.layers, ls -> Chain(ls...)
+@functor Chain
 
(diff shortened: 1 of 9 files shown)


darsnack:
`modules` provides a flattened iterator for 1-level above the leaves. So it gives you the ordered list of nodes seen during the traversal. It's use case is regularizers like
```julia
sum(regularize(l) for l in modules(model))
```
where `regularize` can dispatch by layer type.

So I think for a breaking change, we can put a note in the docs/NEWS and update the tests, but this will not affect the expected use case.

mcabbott:
Ok. Possibly it should filter out all types owned by Base? With `Parallel` you get a tuple of layers, before this PR, which you don't get with `Chain`
```
julia> Flux.modules(Chain(Parallel(+, Dense(2,2,sin), Dense(2,2,cos)), Dense(2,2,tan)))
6-element Vector{Any}:
 Chain(Parallel(+, Dense(2, 2, sin), Dense(2, 2, cos)), Dense(2, 2, tan))  # 18 parameters
 Parallel(+, Dense(2, 2, sin), Dense(2, 2, cos))  # 12 parameters
 (Dense(2, 2, sin), Dense(2, 2, cos))
 Dense(2, 2, sin)    # 6 parameters
 Dense(2, 2, cos)    # 6 parameters
 Dense(2, 2, tan)    # 6 parameters
```

test/runtests.jl
```diff
@@ -8,55 +8,58 @@ using CUDA
 
 Random.seed!(0)
 
-@testset "Utils" begin
-  include("utils.jl")
-end
+@testset verbose=true "Flux.jl" begin
```
mcabbott, line 11: This PR also adds an overall testset, so that all tests are run even if one fails near the start.

darsnack:
> Or perhaps that's just more confusing and staying closer to the literal structure is better.

I think so. And ultimately, only the last three elements above will matter for the expected use case.

src/layers/normalise.jl
```diff
@@ -82,8 +82,7 @@ function Dropout(p; dims=:, rng = rng_from_array())
 end
 
 @functor Dropout
```
ToucheSir, line 84:
```suggestion
@functor Dropout (rng,)
```
While we're at it?

src/layers/basic.jl
```diff
@@ -460,25 +455,31 @@ function Parallel(connection; kw...)
...
+    throw(ArgumentError("Parallel with $nl sub-layers can take one input or $nl inputs, but got $nx inputs"))
+  end
+  m.connection(map(|>, xs, Tuple(m.layers))...)
+end
 
 Base.getindex(m::Parallel, i) = m.layers[i]
-Base.getindex(m::Parallel, i::AbstractVector) = Parallel(m.connection, m.layers[i]...)
+Base.getindex(m::Parallel, i::AbstractVector) = Parallel(m.connection, m.layers[i])
+Base.getindex(m::Parallel{<:Any, <:NamedTuple}, i::AbstractVector) =
+  Parallel(m.connection, NamedTuple{Base.keys(m)[i]}(Tuple(m.layers)[i]))
```
ToucheSir, line 479:
```suggestion
  Parallel(m.connection, NamedTuple{Base.keys(m)[i]}(m.layers))
```
Does the same thing, I believe. Alas neither is type stable.

src/layers/basic.jl
```diff
@@ -28,29 +28,30 @@ julia> m2(x) == (m2[:dec] ∘ m2[:enc])(x)
...
 applychain(fs::Tuple, x) = applychain(tail(fs), first(fs)(x))
 
 (c::Chain)(x) = applychain(Tuple(c.layers), x)
 
-Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i]...)
-Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) = 
-  Chain(; NamedTuple{Base.keys(c)[i]}(Tuple(c.layers)[i])...)
+Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
+Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
+  Chain(NamedTuple{Base.keys(c)[i]}(Tuple(c.layers)[i]))
```
ToucheSir, line 54:
```suggestion
  Chain(NamedTuple{Base.keys(c)[i]}(c.layers))
```
Same as `Parallel`.

src/layers/basic.jl
mcabbott: TIL:
```
julia> NamedTuple{(:a, :c)}((a=1, b=2, c=3))
(a = 1, c = 3)
```

src/layers/normalise.jl
mcabbott: I see, to protect `p` from training, but still move RNG to the GPU.

Are there any downsides to doing it this way, rather than `trainable(d:: Dropout) = (;)`?

ToucheSir: They're complementary changes. I don't think it's super necessary to selectively functor for Dropout since it doesn't have any non-trainable array params, but at some point it would be good to clean house.

mcabbott:
Good to go?

ToucheSir (approved):

PR #1863, opened 2022-02-05
Replace `@adjoint` with `rrule`

mcabbott:
To allow use without Zygote, we should move to defining rules via ChainRules. 

Most of these are mechanical, but perhaps deserve a quick look to see if there are tests. Comments on particular ones below.


src/cuda/cudnn.jl
```diff
@@ -11,10 +11,11 @@ function (BN::Flux.BatchNorm)(x::Union{CuArray{T,2},CuArray{T,4},CuArray{T,5}},
                   training=Flux._isactive(BN)))
 end
 
-@adjoint function batchnorm(g, b, x, running_mean, running_var, momentum; kw...)
+function ChainRulesCore.rrule(::typeof(batchnorm), g, b, x, running_mean, running_var, momentum; kw...)
```
mcabbott, line 14: Not sure why this is CUDA-only. Can this whole file be deleted, and things moved out / to NNlib / NNlibCUDA?

src/layers/normalise.jl
```diff
@@ -38,12 +38,6 @@ function dropout(rng, x, p; dims=:, active::Bool=true)
 end
 dropout(x, p; kwargs...) = dropout(rng_from_array(x), x, p; kwargs...)
 
-@adjoint function dropout(rng, x, p; dims=:, active::Bool=true)
-  active || return x, Δ -> (Δ, nothing)
-  y = dropout_mask(rng, x, p, dims=dims)
-  return x .* y, Δ -> (nothing, Δ .* y, nothing)
-end
```
mcabbott, line 45: Marking the mask `@non_differentiable` should make it safe to delete this

src/layers/recurrent.jl
```diff
@@ -434,7 +435,7 @@ julia> g(rand(Float32, 3, 10)) |> size # batch size of 10
 GRUv3(a...; ka...) = Recur(GRUv3Cell(a...; ka...))
 Recur(m::GRUv3Cell) = Recur(m, m.state0)
 
-
+# TODO move to ChainRulesCore? 
 @adjoint function Broadcast.broadcasted(f::Recur, args...)
   Zygote.∇map(__context__, f, args...)
 end
```
mcabbott, line 441: I think the point of this is that the gradient for `map` reverses iteration order. That's a little dodgy, since `map` makes no such promise (and IIRC it only happens for some argument types, vectors but not 1-column matrices?). Should we just make broadcasting an RNN within a gradient an error?

src/utils.jl
```diff
@@ -789,6 +789,7 @@ L2 (generic function with 1 method)
 modules(m) = [x for x in Functors.fcollect(m) if !isleaflike(x)]
 
 @nograd modules
+ChainRulesCore.@non_differentiable modules(::Any)  # is this correct?
```
mcabbott, line 792: If the intention of modules is that something roughly like `loss + sum(norm, modules(m))` should work, then doesn't this need to pass gradients through?

ToucheSir:
This ought to make debugging easier as well. A potential next step would be moving more bits out to NNlib(CUDA).

src/losses/utils.jl
```diff
@@ -23,6 +23,9 @@ end
   res, Δ -> (nothing, Zygote.unbroadcast(x, xlogy.(Δ, y)), Zygote.unbroadcast(y, Δ .* x ./ y))
 end
 
+ChainRulesCore.@scalar_rule xlogy(x, y) (log(y), x/y)  # is this good enough?
+ChainRulesCore.@scalar_rule xlogx(x) (log(y) + true)
```
mcabbott, line 27: Can't literally translate `broadcasted(::typeof(xlogy)` rule to a Zygote-free world, as `unbroadcast` (which sums as necessary for mismatched shapes) belongs to Zygote.

I hope that Diffractor's broadcasting will work via `@scalar_rule`. But the rule as written is slightly different, as it doesn't treat `Δ==0` as a strong zero, when `y==0`. Does that matter?

ToucheSir: Are these needed if https://github.com/JuliaStats/LogExpFunctions.jl/blob/c8a4c28ffe7b6e4f8d5253e01cef091bb8d2f42c/src/chainrules.jl#L1-L2 is are already loaded through a transitive dep?

src/utils.jl
ToucheSir: Good catch. I have a sinking feeling this might be one of those things that works with implicit gradients but not with explicit ones.

src/layers/normalise.jl
ToucheSir: Happy to see it go :)

src/losses/utils.jl
mcabbott: Flux could switch to those. It has branches not ifelse, and different NaN behaviour, not sure if that matters:

https://github.com/JuliaStats/LogExpFunctions.jl/blob/584442d9bd4c4abadfb5daed86cefa5fabfff645/src/basicfuns.jl#L17-L30

And 5 dependencies.

mcabbott: But for now perhaps it's evidence that the scalar rules are ok?

src/utils.jl
mcabbott: Likewise. Xref https://github.com/FluxML/Functors.jl/issues/35 I guess -- is  `fmapreduce(x -> norm(x.weight), +, m; exclude = x -> x isa Dense)` where we want to end up?

src/layers/recurrent.jl
mcabbott: Moved to here, which I think should give the same results, but also warn on the forward pass:

https://github.com/FluxML/Flux.jl/pull/1863/files#diff-5b453f8f7fb34afbebfc6f688a8209aa0532c8b1c3e95393f97afcbc37a473e7R41-R46

src/utils.jl
ToucheSir: that would be one way of doing things. The big question with any approach is how to prevent AD from balking on the cache mutation + lookup.

src/losses/utils.jl
ToucheSir: Are you looking to do some testing soon with this and Diffractor/not Zygote? Otherwise I think it would be cleaner to have a separate PR that removes all of the code above in favour of https://github.com/FluxML/Zygote.jl/blob/master/src/lib/logexpfunctions.jl and the `@scalar_rule`s in LogExpFunctions.

mcabbott: I can remove these rules for now if you prefer. The functions ought to be differentiable without special rules, mostly. The PR just wants to translate as many things as possible over for now.

ToucheSir: https://github.com/FluxML/Flux.jl/pull/1866

src/layers/normalise.jl
```diff
@@ -1,6 +1,6 @@
 istraining() = false
 
-@adjoint istraining() = true, _ -> nothing
+ChainRulesCore.rrule(::typeof(istraining)) = true, _ -> (NoTangent(),)
```
ToucheSir, line 3: I'm surprised there isn't an equivalent for this in ChainRules already.

mcabbott: Somewhere I was writing a function like `CRC.order().back > 0`... would be good to have.

ToucheSir:
bors try

ToucheSir:
If you wouldn't mind rebasing, we can get this merged assuming that fixes the cuda tests.

src/cuda/cuda.jl
```diff
@@ -3,8 +3,9 @@ module CUDAint
 using ..CUDA
 
 import ..Flux: Flux
-import Zygote
-using Zygote: @adjoint
+# import Zygote
+# using Zygote: @adjoint
```
ToucheSir, line 7:
```suggestion
```

ToucheSir (approved):

PR #1866, opened 2022-02-05
Use LogExpFunctions for losses

ToucheSir:
This package is already in the dep tree through multiple paths (Zygote -> ForwardDiff, StatsBase, etc.), so we might as well make use of it.

### PR Checklist

- [x] Tests are ~~added~~ removed ;)
- [ ] Entry in NEWS.md
~~- [ ] Documentation, if applicable~~


test/losses.jl
```diff
@@ -17,22 +16,6 @@ const ALL_LOSSES = [Flux.Losses.mse, Flux.Losses.mae, Flux.Losses.msle,
...
-  @inferred xlogx(0)
-
-  @test iszero(xlogy(0, 1))
-  @test isnan(xlogy(NaN, 1))
-  @test isnan(xlogy(1, NaN))
-  @test isnan(xlogy(NaN, NaN))
-  @test xlogy(2, 3) ≈ 2.0 * log(3.0)
-  @inferred xlogy(2, 3)
-  @inferred xlogy(0, 1)
-end
```
mcabbott, line 34: Do these tests pass?

src/losses/utils.jl
```diff
@@ -1,31 +1,6 @@
...
-
-"""
-    xlogy(x, y)
-
-Return `x * log(y)` for `y > 0`, and zero when `x == 0`.
-"""
-function xlogy(x, y)
-  result = x * log(y)
-  ifelse(iszero(x), zero(result), result)
-end
```
mcabbott, line 19: One question is whether there are any performance differences, and whether we care. IIRC the replacements have `if else` instead of `ifelse`, but perhaps the compiler sorts it out?

test/losses.jl
ToucheSir: I tested locally before removing and they do. https://github.com/JuliaStats/LogExpFunctions.jl/blob/master/test/basicfuns.jl also looks like a strict superset of the Flux tests.

src/losses/utils.jl
ToucheSir: I found https://github.com/JuliaStats/LogExpFunctions.jl/pull/26. GPU is the big question mark, but if https://github.com/FluxML/Flux.jl/pull/1791 is any indication there may not be a difference there either.

mcabbott: I gret similar numbers.

mcabbott:
I'm not strongly opposed but I wonder what we gain here, the functions are so simple.

Surely one could write a very fast `xlogx(::Float32)` via rational functions etc, since it doesn't need the hard part of log. But LogExpFunctions doesn't do that. And I doubt it's ever a bottleneck here.

Flux currently depends indirectly on this package (and its dependencies), but would ultimately like not to depend on Zygote, ForwardDiff, et. al. I lean towards a few lines of duplication being better than a tight web of every package depending on every other one.

darsnack:
I think the main benefit of external packages over code duplication is maintainability and testing. If it is a relatively lightweight dep and we use it for only a few functions, then I see no harm in having it. It'd be a different story if we take on a dep that provides many functions in Flux, making them intricately woven.

mcabbott:
Right, I just feel that maintenance and testing of lines
```
  result = x * log(y)
  ifelse(iszero(x), zero(result), result)
```
here might be less hassle than maintenance of lines in Project.toml, breaking changes downstream, etc. If we want it never to change, then never changing it seems simpler than hoping some other package won't. I agree this is subjective though.

Besides such tradeoffs, the `broadcasted` gradients here aren't duplicated there. Ideally the future implementation of broadcasting will work off `@scalar_rule` but not yet.

ToucheSir:
I suppose I hold the opposite perspective. There is a not significant amount of, for lack of a better term, derelict code kicking around in FluxML packages. Think a lot of the utility functions here (some which are not tested :see_no_evil:) and certain adjoints in Zygote.

For this particular case, I don't think the extra dep is a problem for a few reasons. Firstly, if LogExpFunctions breaks then Flux is going to feel it either way, since it's on a critical path of some direct dependencies. Secondly, `xlogx` and `xlogy` is a small surface area to break, but have enough edge cases that I think the expanded test coverage in LogExpFunctions is a good thing. Lastly, there is a better chance of silently breaking changes being caught downstream because more packages rely on this functionality than Flux. We can all think of examples where routines here were doing the wrong thing but not erroring, so having that canary would allow us to focus effort on more core parts of the library.

About broadcasting, I had a look back through the blame and the adjoint in question was added [2 years ago](https://github.com/FluxML/Flux.jl/commit/f5a8900ffb50b7b5fede3d3c085740c41ab595ff). Nowadays, every AD we care about has a fast path for broadcasting `LogExpFunctions.xlogy`, whether that's through DiffRules or ChainRules (Diffractor). Perhaps if it needed the rrule to store some intermediate results for the pullback, but since it doesn't I don't see why the dedicated broadcast rule would be necessary.

ToucheSir:
@mcabbott another temperature check on this now that #1863 is ready for review?

darsnack:
I'm still in favor overall

CarloLucibello:

PR #1867, opened 2022-02-06
Updated Dropout for more input types.

ShoofLLC:
Replacing the call to rand! in the _dropout_mask function to account
for complex (float) data types.

See discussion: https://github.com/FluxML/Flux.jl/pull/1572

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


src/layers/normalise.jl
```diff
@@ -49,7 +49,8 @@ dropout_mask(rng, x::CuArray, p; kwargs...) =
   throw(ArgumentError("x isa CuArray, but rng isa $(typeof(rng)). dropout_mask only support CUDA.RNG for CuArrays."))
 dropout_mask(rng, x, p; kwargs...) = _dropout_mask(rng, x, p; kwargs...)
 function _dropout_mask(rng, x, p; dims=:)
-  y = rand!(rng, similar(x, _dropout_shape(x, dims)))
+  fptype = float(real(eltype(x)))
+  y = rand!(rng, similar(x, fptype, _dropout_shape(x, dims)))
```
darsnack, line 53: Just so we remember what this is for.

```suggestion
  realfptype = float(real(eltype(x)))
  y = rand!(rng, similar(x, realfptype, _dropout_shape(x, dims)))
```

darsnack (approved):

ShoofLLC:
The return type of dropout if **NOT** "active" for Int types would be Int.
But when "active" it would be Float.

That doesn't seem like a good thing, so I think that's a no on this one too.

darsnack (changes requested):
Can you add an entry in NEWS.md?

ShoofLLC:
> Can you add an entry in NEWS.md?

This still needs a new review, I've found something while testing with int types.

src/layers/normalise.jl
```diff
@@ -34,7 +34,7 @@ The [`Dropout`](@ref) layer is what you should use in most scenarios.
 function dropout(rng, x, p; dims=:, active::Bool=true)
   active || return x
   y = dropout_mask(rng, x, p, dims=dims)
-  return x .* y
+  return x .* iszero.(y)
```
ToucheSir, line 37:
```suggestion
  return x .* y
```
This change is incorrect, since `y` is not a binary mask (it contains scaling factors to compensate for the dropped elements).

darsnack:
> The return type of dropout if NOT "active" for Int types would be Int.
But when "active" it would be Float.

When `x isa Int`? Wouldn't the `iszero` take care of this?

ToucheSir:
> The return type of dropout if **NOT** "active" for Int types would be Int.
> But when "active" it would be Float.

This behaviour is already present on master. I don't think it needs to be addressed in this PR. If anything, this PR improves compatibility with int inputs, as before `dropout_mask` was calling `rand!(::AbstractArray{Int})` and silently doing the wrong thing...

darsnack:
The fix should make the inactive path return a chosen float type. It doesn't make sense to return an integer with the scaling.

ToucheSir:
The tricky part is that the inactive path doesn't require any scaling. Or put another way, it's semantically equivalent to `p=0`, which leads to a scaling factor `1/(1-p) === 1`. I suppose we could specifically convert or error on integer arrays?

darsnack:
True, but I think the return type should be stable across active/inactive regardless of the actual computation. `1 / (1 - 0)` is a float, so I think always converting is reasonable here.

ShoofLLC:
> True, but I think the return type should be stable across active/inactive regardless of the actual computation. `1 / (1 - 0)` is a float, so I think always converting is reasonable here.



> The tricky part is that the inactive path doesn't require any scaling. Or put another way, it's semantically equivalent to `p=0`, which leads to a scaling factor `1/(1-p) === 1`. I suppose we could specifically convert or error on integer arrays?

Since scaling is not optional, that kinda makes sense, it would at least let the user know that something is wrong.

darsnack:
It's not exactly an error, right? Mathematically the correct output exists, but it cannot be an integer except for very limited values of `p`. Perhaps we can warn though cause converting the type might cause unexpected downstream errors.

darsnack (approved):
Either way, I agree with @ToucheSir that we should merge this and then deal with the type stability issue in a separate PR.

darsnack:
Shall we merge this?

PR #1870, opened 2022-02-13
Mark dropout_mask as non-differentiable

ToucheSir:
Should allow for nested AD of Dropout.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md
~~- [ ] Documentation, if applicable~~


mcabbott (approved):

src/layers/normalise.jl
```diff
@@ -55,6 +55,9 @@ function _dropout_mask(rng, x, p; dims=:)
   return y
 end
 
+# TODO move this to NNlib
+Zygote.ChainRulesCore.@non_differentiable dropout_mask(rng, x, p)
```
ToucheSir, line 59:
```suggestion
Zygote.ChainRulesCore.@non_differentiable dropout_mask(::Any, ::Any, ::Any)
```
I noticed #1863 uses this signature. Is there a functional difference between the two?

test/layers/normalisation.jl
```diff
@@ -458,3 +458,8 @@ end
     @test BN(x) ≈ GN(x)
   end
 end
+
+@testset "second derivatives" begin
+  m1 = Dropout(0.5)
+  @test Zygote.hessian_reverse(sum∘m1, [1.0,2.0,3.0]) == ones(3, 3)
```
ToucheSir, line 464: Previously broken conv tests at https://github.com/FluxML/Flux.jl/blob/master/test/layers/basic.jl#L318 are now passing. Should they be marked un-broken as part of this PR?

ToucheSir (commented):
Need to fix a test and add news, a couple comments while I do:

test/layers/normalisation.jl
mcabbott: This is https://github.com/FluxML/NNlib.jl/pull/387 no? I would mark them fixed, and raise the bound on NNlib.

ToucheSir: Yes indeed. I've added a separate commit for this.

Issue #1700, opened 2021-08-25
Make it possible to run subset of tests


mkschleg:
Right now when running tests you have to run the entire suite, which takes a few minutes. Might be convenient for development if we adopt the use of `test_args` to allow for running specific tests (see https://github.com/JuliaLang/Pkg.jl/pull/1226).

CarloLucibello:
Done — the test suite uses ParallelTestRunner and forwards `ARGS`, so a single file can be run directly:

```
julia --project=test/ test/runtests.jl layers/conv
julia --project=test/ test/runtests.jl --list
```

Closing.

PR #1871, opened 2022-02-13
Recurrent benchmarks

mkschleg:
Given the conversation in #1855, I think we should have benchmarks in the main repo we can use to iterate on for rnn performance. This touches benchmark_utils to add some of the `Flux.Recur` specific machinery we need. I only added tests for `RNN` using both the vector and 3d interfaces, but we can add the other cells once we are happy w/ how these operate.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/perf/bench_utils.jl
+++ b/perf/bench_utils.jl
@@ -28,7 +28,7 @@ function run_benchmark(model, x; cuda=true)
     end
 
     ps = Flux.params(model)
-    y, back = if model isa Flux.Recur
+    y, back = if model isa Flux.Recur && eltype(x) isa AbstractVector
         pullback(() -> sum(sum([model(x_t) for x_t in x])), ps)
     else
         pullback(() -> sum(model(x)), ps)
--- a/perf/bench_utils.jl
+++ b/perf/bench_utils.jl
@@ -28,7 +28,7 @@ function run_benchmark(model, x; cuda=true)
     end
 
     ps = Flux.params(model)
-    y, back = if model isa Flux.Recur && eltype(x) isa AbstractVector
+    y, back = if model isa Flux.Recur && eltype(x) isa AbstractArray
         pullback(() -> sum(sum([model(x_t) for x_t in x])), ps)
     else
         pullback(() -> sum(model(x)), ps)
--- a/perf/bench_utils.jl
+++ b/perf/bench_utils.jl
@@ -1,12 +1,24 @@
 using BenchmarkTools
 using Flux
 using CUDA
-using Zygote: pullback
+using Zygote: pullback, ignore
 
 
 fw(m, x) = m(x)
 bw(back) = back(1f0)
 fwbw(m, ps, x) = gradient(() -> sum(m(x)), ps)
+
+# Need to specialize for flux.recur.
+fw(m::Flux.Recur, X::Vector{<:AbstractArray}) = begin
+    ignore() do
+      Flux.reset!(m)
+    end
+    [m(x) for x in X]
+end
+fwbw(m::Flux.Recur, ps, X::Vector{<:AbstractArray}) = gradient(ps) do
+    y = fw(m, X)
+    sum(sum(y))
+end
   
 function run_benchmark(model, x; cuda=true)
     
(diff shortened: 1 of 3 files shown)


ToucheSir:
One thing I was experimenting with locally was a more structured benchmarking system like https://github.com/JuliaGPU/CUDA.jl/blob/master/perf/runbenchmarks.jl. It would be nice to have a way to visualize results from these smaller-scale performance tests. https://speed.fluxml.ai/ currently has just E2E examples.

mkschleg:
Having good tools for testing performance and for testing is up there with things I wish we had in Flux (#1700). With the performance measurements I think it would be good to also capture how much memory we use (to more easily catch things like #1872).

mkschleg:
@CarloLucibello and @ToucheSir I'm cool with waiting to add RNN tests until we figure out a better framework if that is preferable! Otherwise, what do we want done in this PR to get this merged?

perf/bench_utils.jl
```diff
@@ -1,12 +1,24 @@
...
+fw(m::Flux.Recur, X::Vector{<:AbstractArray}) = begin
+    ignore() do
+      Flux.reset!(m)
+    end
+    [m(x) for x in X]
+end
+fwbw(m::Flux.Recur, ps, X::Vector{<:AbstractArray}) = gradient(ps) do
+    y = fw(m, X)
+    sum(sum(y))
+end
```
ToucheSir, line 21: When benchmarking for #1761, I used a wrapper struct for this instead of special-casing `Recur`. See https://gist.github.com/ToucheSir/2fecbe99e5b304fed11e25e42c535cc3#file-rnn-jl.

mkschleg: I followed almost exactly what you did in the gist, with a few modifications.

mkschleg:
I made some modifications. Primarily from @ToucheSir's suggestions. I also modified the bench_utils slightly to make overloading functionality easier. Primarily by adding a pullback function `pb(m, ps, x)` and changing some of the other generic versions to use `fw` instead of a generic call.

perf/bench_utils.jl
```diff
@@ -1,13 +1,14 @@
...
-using Zygote: pullback
+using Zygote: pullback, ignore
 
 
 fw(m, x) = m(x)
 bw(back) = back(1f0)
-fwbw(m, ps, x) = gradient(() -> sum(m(x)), ps)
-  
+fwbw(m, ps, x) = gradient(() -> sum(fw(m, x)), ps)
+pb(m, ps, x) = pullback(()->sum(fw(m, x)), ps)
```
ToucheSir, line 10:
```suggestion
pb(m, ps, x) = pullback(() -> sum(fw(m, x)), ps)
```

perf/recurrent.jl
```diff
@@ -0,0 +1,62 @@
...
+
+fwbw(r::RNNWrapper, ps, X::Vector{<:AbstractArray}) = gradient(ps) do
+  y = fw(r, X)
+  sum(sum(y))
+end
+
+pb(r::RNNWrapper, ps, X::Vector{<:AbstractArray}) = pullback(ps) do
+  y = fw(r, X)
+  sum(sum(y))
+end
```
ToucheSir, line 27: I believe most of this could be handled by defining `(r::RNNWrapper)(X::Vector{<:AbstractArray})` and `(r::RNNWrapper)(X)`, no?

mkschleg: Right. I guess I focused on using the fw function instead, is because otherwise I'm not sure why that exists. I think using this it as I have is a cleaner way to separate benchmark specific code and flux code. I don't really care either way and can change pretty easily if we want.

ToucheSir: My reading was that `fw`, `bw` and `fwbw` exist mostly to save a few keystrokes. The code in `perf/` could use some love and none of it should be considered sacred, so feel free to dice stuff up if it makes more sense to you.

The purpose of the suggestion above was mostly to avoid writing more than one forward pass. On second reading, I realize that's not possible without adding an overload for `(::Recur)(::Vector{<:AbstractArray})`. So let's go with your current code, and ruminate separately over whether that overload should be part of the Flux API :)

Issue #1282, opened 2020-07-15
Data loading & preprocessing pipeline feature


ageron:
The `DataLoader` is nice, but if I understand correctly it requires the dataset to fit in memory. For large datasets that don't fit in memory, it would be nice to have an easy way to load & preprocess the data efficiently, similar to TensorFlow's tf.data API. Maybe something like this exists already?

If not, perhaps one option would be to provide custom [transducers](https://github.com/JuliaFolds/Transducers.jl) to make it possible to write things like:

```julia
data = (csv_file_paths |> Shuffle(length(csv_file_paths)) |> Interleave(CSV.File; threads=4)
             |> Map(preprocess_sample) |> Shuffle(100_000) |> Batch(32) |> Prefetch(1))
```

This would load records from multiple files (in random file order), pick 4 randomly, interleave their records, preprocess every record, shuffle records using a 100,000 element buffer, and batch the records with batch size 32, and prefetch 1 batch (so the CPU can prepare the next batch while the GPU is working on the previous batch). Then the `data` could be used for training.

pxl-th:
Right now it is possible to use `DataLoader` with datasets that do not fit into memory, but feels somewhat hacky... or not.
Let's say we have dataset of images, which we want to load on demand.
Then we can define a custom struct subtyping `AbstractArray`.

```julia
struct Dataset{T, N} <: AbstractArray{T, N}
    frame_template::FormatExpr
    targets::AbstractArray
end

Dataset{T}(frame_template, targets) where {T} =
    Dataset{T, 1}(frame_template, targets)
```

Then defining `getindex` function that describes how we load one item

```julia
function Base.getindex(d::Dataset{T}, i::Int) where {T}
    path = format(d.frame_template, i - 1)
    image = path |> FileIO.load |> Images.channelview .|> T
    image, d.targets[[i]]
end
```

and how we load mini-batch

```julia
function Base.getindex(d::Dataset{T}, ids::Array) where {T}
    x, y = d[ids[1]]
    xs_last_dim = ntuple(i -> Colon(), ndims(x))
    ys_last_dim = ntuple(i -> Colon(), ndims(y))

    xs = Array{T}(undef, size(x)..., length(ids))
    ys = Array{T}(undef, size(y)..., length(ids))

    xs[xs_last_dim..., 1] .= x
    ys[ys_last_dim..., 1] .= y

    for (i, id) in enumerate(ids[2:end])
        x, y = d[id]
[...]

Issue #1159, opened 2020-05-04
DataLoader type instability


cossio:
Here is an example:

```
using Flux, MLDatasets
train_x, train_y = MNIST.traindata()
train_loader = Flux.Data.DataLoader(train_x_reduced, batchsize=16,
                                    shuffle=true, partial=false)
eltype(train_loader) # Any
```

Since the `eltype` is `Any`, iterating over the `DataLoader` leads to non-inferrable code.

```julia
function do_something(data)
    s = 0.0
    for d in data
        s += mean(d)
    end
    return s
end
```

`@code_warntype do_something(train_loader)` prints:

![image](https://user-images.githubusercontent.com/10637482/80954186-30038200-8dfd-11ea-9eb4-e3a8320fe24d.png)

and `@inferred do_something(train_loader)` fails.

cossio:
Fixed by https://github.com/FluxML/Flux.jl/pull/1152

cossio:
@CarloLucibello Reopen?

CarloLucibello:
Closing here: `DataLoader` now lives in MLUtils.jl, and its `eltype` is concrete rather than `Any`. Any remaining inference issue is best filed against MLUtils.jl.

PR #1874, opened 2022-02-14
use MLUtils

### LINKED ISSUE #1227: Cannot do double iteration of DataLoader
The current `DataLoader` cannot be used to iterate on two data-points at the same time (or it doesn't work as expected). Suppose I write:

```julia
data = DataLoader((randn(5,5,100), rand(Bool, 100)); shuffle=true, batchsize=16);
for d1 in data, d2 in data
...
end
```

The problem is that `data.indices` stores the shuffled indices for *both* `d1` and `d2`. We should have independent shuffled indices for each iteration. That is, the shuffled indices should be stored in the state of iterator, instead of in the iterator itself, I think.

CarloLucibello:
This is replacing some utility functions and the `DataLoader` with their equivalent versions in [MLUtils.jl](https://github.com/JuliaML/MLUtils.jl). 

## On the DataLoader inferrability
After this, the `DataLoader` will accept generic datasets exposing the `MLUtils.numobs` and `MLUtils.getobs` interface (something that we should probably document somewhere).  This has been discussed in #1282. 

Unfortunately we lose inferrability, e. g. `@inferred first(loader)` will fail. This is because the previous implementation was just wrong for generic datasets. It was
```julia
Base.eltype(loader::DataLoader{D}) where D = D
```
where `D` is the underlying data type. So this PR will reopen #1159. This is an example of the current problematic behavior on master:
```julia
julia> using Flux.Data: Dataloader

julia> d = DataLoader(1:5);

julia> [x for x in d]  # iteration is fine
5-element Vector{Vector{Int64}}:
 [1]
 [2]
 [3]
 [4]
 [5]

julia> collect(d)
ERROR: MethodError: Cannot `convert` an object of type Vector{Int64} to an object of type UnitRange{Int64}
Closest candidates are:
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray at ~/julia/julia-1.7.1/share/julia/stdlib/v1.7/LinearAlgebra/src/factorization.jl:58
  convert(::Type{T}, ::AbstractRange) where T<:AbstractRange at ~/julia/julia-1.7.1/share/julia/base/range.jl:215
  convert(::Type{T}, ::T) where T<:AbstractArray at ~/julia/julia-1.7.1/share/julia/base/abstractarray.jl:16
  ...
Stacktrace:
 [1] setindex!(A::Vector{UnitRange{Int64}}, x::Vector{Int64}, i1::Int64)
   @ Base ./array.jl:903
 [2] copyto!(dest::Vector{UnitRange{Int64}}, src::DataLoader{UnitRange{Int64}, Random._GLOBAL_RNG})
   @ Base ./abstractarray.jl:893
 [3] _collect(cont::UnitRange{Int64}, itr::DataLoader{UnitRange{Int64}, Random._GLOBAL_RNG}, #unused#::Base.HasEltype, isz::Base.HasLength)
   @ Base ./array.jl:655
 [4] collect(itr::DataLoader{UnitRange{Int64}, Random._GLOBAL_RNG})
   @ Base ./array.jl:649
 [5] top-level scope
   @ REPL[9]:1
 [6] top-level scope
   @ ~/.julia/packages/CUDA/bki2w/src/initialization.jl:52
```
## TODO:
- [x] Add news
- [x] check if it is closing #1227. It does!

Close #1227


docs/src/utilities.md
```diff
@@ -7,15 +7,17 @@ callback functions.
 
 ## Working with Data
 
+Utilities for data processing are re-exported from [MLUtils.jl](https://github.com/JuliaML/MLUtils.jl). Below is a non-exhaustive list.
```
mcabbott, line 10: Provided by, maybe?

src/onehot.jl
```diff
@@ -87,7 +87,7 @@ Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 2}} =
 Base.hcat(x::T, xs::T...) where {L, T <: OneHotLike{<:Any, L, <:Any, 1}} =
   OneHotMatrix(reduce(vcat, _indices.(xs); init = _indices(x)), L)
 
-batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(_indices.(xs), L)
+MLUtils.batch(xs::AbstractArray{<:OneHotVector{<:Any, L}}) where L = OneHotArray(_indices.(xs), L)
```
mcabbott, line 90: Is this right?

CarloLucibello: In general, it's not clear to me what `batch(xs::AbstractArray{T, N})` is supposed to mean. Add `N` dimensions at the end (assuming this makes sense for `T`) and stack the elements along them? If that's the case this definition for OneHot seems consistent. I'm not sure if that's your perplexity.  

In any case we should document better `batch`'s behavior, or decide we only support `batch(xs::AbstractVector)`.

mcabbott: Oh sorry, I thought this `batch` was a function which belonged only to onehot stuff, and I saw many tests failing for it. 

More generally why is there "batch" and "stack"? Might this move to MLUtils be the right time to see how many of these functions can be killed? And to give the rest clearer descriptions than "batch: Batch the arrays"...

CarloLucibello: In GNN.jl I defined `batch(gs::Vector{GNNGraph) -> GNNGraph` to "batch" multiple graphs into a single one with a block-diagonal structure as it is usually done when using graph neural networks. I guess I could have used `stack` instead.

Maybe it is ok to have the distinction:`stack` is for arrays; `batch` is for generic datasets, meant to be extensible, and when used for arrays it falls back to `stack` with a specific dims argument (N+1). This is the way `batch` is currently implemented in MLUtils.jl.

mcabbott: That sounds OK, but maybe you should own the verb? Such a block-diagonal-within-some-struct use doesn't seem close to:
```
help?> Flux.batch
  batch(xs)

  Batch the arrays in xs into a single array.

  See also unbatch

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  julia> Flux.batch([[1,2,3],[4,5,6]])
  3×2 Matrix{Int64}:
   1  4
   2  5
   3  6
```
This example is exactly what `Base.stack` will (I hope) do.

CarloLucibello: With respect to `stack`, `batch` also has special handling of tuples, named tuples, and dicts. We could say that `batch` is a very opinionated version of `stack`.

CarloLucibello: > but maybe you should own the verb? 

I'd like to use a common interface. For example, AlphaZeros.jl calls `batch` to group observed states and feed the to prediction models. In some situations, states can be represented by graphs and the model is a graph neural network...

mcabbott: But why both then? And if you have to read the source to figure out what the different utils do, and they're each one line, maybe they shouldn't exist? Can we delete `stack` and actually document `batch`? `unstack` appears to be `eachslice` except more confusing... maybe it's a fossil from before Julia 1.0? 

> Equivalent to SparseArrays.blockdiag

This still sounds to me like it should be a distinct function. What do they share, except the english word "batch"?

CarloLucibello: > Can we delete stack and actually document batch? unstack appears to be eachslice except more confusing... maybe it's a fossil from before Julia 1.0?

`stack` is familiar to numpy/pytorch users. Yes, we can replace it by adding a `dims` keyword to the `batch` specialization on arrays. I don't see a particular advantage in doing that though. 

> This makes it sounds like it should be a distinct function. What do they share, except the english word "batch"? I guess I don't really know what this is doing, is there some sense in which this is the same operation, e.g. (tograph ∘ batch)(x) == (batch ∘ tograph)(x)?


For deep learning on graphs, the node dimension is very similar to the batch dimension. You work with graphs with features on nodes, so let's say represented by a  `N x N` adjacency matrix and  `D x N` feature matrix. When you batch graphs together you obtain a graph with `Ntot x Ntot` adj mat and `D x Ntot` feature vectors. So a GNNGraph contains both feature arrays and some representation of the topology, and the meaning of batching is an extension of the meaning we commonly apply to arrays.

ToucheSir: You'd want methods on the same function for collating data in a dataloader, otherwise some data types will just be completely incompatible. It may be that we want to make that knob tweakable at some point on the data loading side, but having `batch` as a default makes a lot of sense. In contrast, `stack` IMO is just a workaround for not having an equivalent function in Base that users from other frameworks expect to have.

mcabbott: Ok. Can this `batch` be documented carefully though, what it's expected to do, what types the basic one handles, how extensions are expected to interact with DataLoader? 

Depending on how likely you think  https://github.com/JuliaLang/julia/pull/43334 is to land, it might be worth making methods line up here.

CarloLucibello:
Merge?

CarloLucibello:
mergity merge?

darsnack:
Other question is what happens when MLUtils -> MLDataUtils?

darsnack (approved):
No more concerns on my end

mcabbott:
There are many warnings like this in the tests:
```
┌ Warning: `unsqueeze(x::AbstractArray, dims::Int)` is deprecated, use `unsqueeze(x; dims = dims)` instead.
│   caller = macro expansion at Test.jl:463 [inlined]
└ @ Core ~/.julia/dev/julia/usr/share/julia/stdlib/v1.8/Test/src/Test.jl:463
┌ Warning: `stack(x, dims)` is deprecated, use `stack(x; dims = dims)` instead.
│   caller = macro expansion at utils.jl:249 [inlined]
└ @ Core ~/.julia/dev/Flux/test/utils.jl:249
```
No rush but it would be nice to update all of them.

cossio:
Can we have a release with the nicer DataLoader?

darsnack:
These are the blockers: https://github.com/FluxML/Flux.jl/milestone/3

PR #1875, opened 2022-02-15
Add a structural `loadparams!`

darsnack:
This replaces `loadparams!` with `loadmodel!` which uses `fmap` to structurally walk the model and copy parameters over. Right now it mutates destination model, so fields like the activation are not copied.

I opted to have a more verbose implementation than the one-liner `fmap(loadto!, m, mbar)`. It allows us to have more informative error messages for the standard layers. Custom layers will fallback to the error thrown by Functors.jl.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md


ToucheSir:
Because no API addition is complete without bikeshedding, might I link to the discussion in https://github.com/beacon-biosignals/LegolasFlux.jl/pull/4#issuecomment-875858899 :stuck_out_tongue_winking_eye:.

(`loadmodel!` is a fine name, just FYI if you haven't seen it already).

src/loading.jl
```diff
@@ -0,0 +1,47 @@
...
+loadto!(x::AbstractArray, x̄::AbstractArray) = copyto!(x, x̄)
+for T in [:Dense, :Bilinear, :Conv, :ConvTranspose, :DepthwiseConv, :CrossCor]
+  @eval begin
+    function loadto!(m::$T, m̄::$T)
+      if (size(m.weight) != size(m̄.weight)) || (size(m.bias) != size(m̄.bias))
+        throw(DimensionMismatch("Tried to load $m̄ into $m but the parameter sizes do not match."))
+      else
+        return fmap(loadto!, m, m̄)
+      end
+    end
```
ToucheSir, line 18: I wonder if part of this could be pulled into a helper function. Maybe
```suggestion
    function loadto_helper!(m, m̄, ps)
      for p in ps
        if size(getproperty(m, p)) != size(getproperty(m̄, p))
          throw(DimensionMismatch("Tried to load $m̄ into $m but the parameter sizes do not match."))
        end
      end
      return fmap(loadto!, m, m̄)
    end
```
And then
```suggestion
  loadto!(m::$T, m̄::$T) = loadto_helper!(m, m̄, (:weight, :bias)) 
```

mcabbott:
Can you write briefly what this does & doesn't do? E.g. it copies non-trainable parameter arrays, unlike restructure, but not integer sizes/strides, not activation functions, how about dropout rate? At least as a docstring.  

What's the argument for the name change?

What happens with immutable arrays? Maybe it should work like `update!` in that it mutates if it can but returns the actual truth. 

What this custom traverse doesn't allow is recovering all the weights from something like `fmapstructural(identity, m)`, is this a good idea? Should the source be allowed to completely omit e.g. an activation? 

It seems you could still have nice errors, even for custom layers, by just keeping track of the one-layer-below name/summary. Not `fmap` I guess, but could still be one function. That might end up more compact & easier to test well.
```
function load!(dst::AbstractArray, src::AbstractArray; str=summary(dst))
    size(dst) == size(src) || throw(DimensionMismatch("wrong size within $str"))
    copyto!!(dst, src)
end

function load!(dst, src; str="")
    din, re = functor(dst)
    sin, _ = functor(typeof(dst), src)  # could try-catch this for pretty error on missing field?
    str = string(dst)
    re(map((d,s) -> load!(d,s; str), din, sin))
end

copyto!!(dst::DenseArray, src) = copyto!(dst, src)
copyto!!(dst::AbstractArray, src) = typeof(dst)(src)  # or something, to allow immutable
```

Do we need the `xs::AbstractVector` method in addition to Params? 

That's a lot of questions, sorry!

darsnack:
Yes, this PR still needs docstrings, examples, and doc updates before it's ready.

> (loadmodel! is a fine name, just FYI if you haven't seen it already).

I had not seen that! "State" might be better than "model" here. I'm open to changing it or keeping the old name.

> What's the argument for the name change?

Flux uses "parameters" to mean "trainable parameters," while this function loads both the trainable parameters and the non-trainable state. I wanted the name to make clear that we are considering the model's structure when loading. My main motivation was so that a user noticed the change, but I'm happy to keep the old name.

> Can you write briefly what this does & doesn't do?

> What happens with immutable arrays?

Right now they would just fail (though I could give a better error message). It would be easy enough to make this behave like `Optimisers.update!`, and my original implementation did just that as well as included *every* functor'd field. We can also allow immutable arrays but not reconstruct every field. We can also have something like `trainable` which allows someone to specify which fields should be "loadable."

But I am waffling on what's the right behavior. Before calling `loadmodel!(m, mbar)`, a user would have constructed `m`. That means setting the stride, padding, activation functions, dropout rate, etc. When they call `loadmodel!`, I expect that what they want is the parameters and state that came from training `mbar` but not *everything*. So, I walked back my initial implementation.

As it stands, Flux doesn't work with immutable parameters. The gradients can be immutable, but not the parameters. I figure we can make this function behave like `Optimisers.update!` when we have optimizers that support immutable parameters.

> What this custom traverse doesn't allow is recovering all the weights from something like fmapstructural(identity, m), is this a good idea? Should the source be allowed to completely omit e.g. an activation?

Believe this is asking a similar thing to what I addressed above, but I wasn't sure. Could you rephrase if I misunderstood?

> It seems you could still have nice errors, even for custom layers, by just keeping track of the one-layer-below name/summary.

Good point, I'll refactor.

> Do we need the xs::AbstractVector method in addition to Params?

I'm not sure if it is used somewhere, but the old `loadparams!` would accept it.

ToucheSir:
The two valid models requirement doesn't seem necessary for this PR. Assuming structural similarity, I don't see any obvious barriers to using the nested namedtuple as the second arg instead of an actual model struct. Medium-long term something like you mention would be ideal, but I see this PR as a good MVP so that users can actually start getting rid of implicit params for serialization. Better to be a little conservative while we figure out the edge cases of a more general approach.

darsnack:
Okay, I see what you mean. That would be a larger step than this PR, but something I am willing to do. This would basically be two pieces:
- a "saving" step that produces the `NamedTuple` that has everything that can be saved
- a "loading" step that accepts that `NamedTuple`

The "loading" step can already be done by this PR as Brian mentioned. In the `somefun!(m, m0)` case, I don't want to use the `re` that produced `m0`. The reason I created this PR is that we don't currently have a good way of saying "`m` is the ground truth structure, copy the parameters into `m` while making sure the structure matches." So, using a `re` would circumvent that check.

ericphanson:
re-documentation, it would be great if the docs made sure to specify the interface for custom layers to participate (e.g. when to define `_loadleaf` etc), not just how to use it with built-in layers.

ToucheSir:
Personally, I'm not a fan of the `out, re = fn(args...)` API unless it's absolutely necessary (as in `destructure`), as you have to keep `re` around and it may close over a bunch of data that could otherwise be GCed. Having an API that effectively does `Fix1(restructure, model)` where `restructure(model, params) = ...` would be fine, but it should not be the default.

mcabbott:
> somefun!(m, mtrained): effectively call somefun!(m)[2](somefun!(mtrained)[1])

This seems redundant, I would hope that `somefun!(m)[2](somefun!(mtrained)[1]) == somefun!(m)[2](mtrained)`, since (by definition) `somefun!(mtrained)[1]` is a nested struct with all the same fields as `mtrained`.

darsnack:
I agree that between `de/restructure` and `loadmodel!`, we are talking about the same underlying core with slightly different use cases. Sharing code paths would be good.

The fact that `loadmodel!` works without needing a closure suggest to me that `somefun!` should be something that walks 1 structure and produces the simplified form, or walks 2 structures copying from the second to the first. Then mapping the simplified form to a flat vector can be a separate piece where closing over might be required.

ToucheSir:
Let me be more precise: `destructure` returns both an output (flattened params) and additional bookkeeping information (`re`). The latter is required in order to restructure, so even if the API was `remake!(m, v)` one would have to keep it around. Hence it makes a lot of sense to have `re` be callable (as FluxML/Optimisers.jl#54 did), as it needs to be carried around anyways.

In contrast, `somefun!` should only have to return the serialization-friendly model state and not any additional bookkeeping info. Therefore, there's no need to construct a reconstruction function and carry that around. `Restructure(m)(v)` should be the same as `restructure(m, v)`, so there's no need for the former outside of syntactic convenience (for which you can use `Fix1` or define a curried version).

darsnack:
I guess I am in agreement at the highest level of this discussion, but I'm confused about what's actually being proposed. So, can we make `somefun!` more concrete? What's an actual possible name for this function, and what does it do?

darsnack:
> One reason I asked about the ::AbstractVector method above was wondering whether loadparams!(m, rand(100)) with a vector of numbers should just automatically do that. Do we want it to? If so that may change our opinions on how the rest should work.

Ultimately, I want to deprecate that path entirely, but I plan on doing it once implicit params are gone.

mcabbott:
Maybe I should think some more and write the options somewhere. But not today. Sorry about derailing the PR!

tl;dr is that I'd vote not to introduce a new `loadmodel!` function until we've thought a bit more about what we want. Perhaps using something like `fmap(copyto!, m, m2)` for now is safest, as this won't break.

ToucheSir:
> I think you are assuming that most uses of `destructure` wants the `re` not just the vector, while most uses of `somefun!(m)` would want only the simplified struct, not the `re`.
> 
> But this "most" doesn't seem absolute. You may want flat parameters to e.g. use them for some regularisation within the loss. You may want Base structs to save them. On the other hand, you may also (as in the present PR) want only the `re` part, `somefun!(m, mtrained)`, and you may similarly call `_, re = destructure(m)` so that you can put `re(x,p)` inside some loss.

All this makes sense for `destructure`, but I'm not sure how it applies to saving and loading model state here? In fact,

> you may also (as in the present PR) want only the `re` part, `somefun!(m, mtrained)`

To me is a great argument for having the two separate load/save functions, because otherwise you're incurring unnecessary work to generate both of `ps, re` just to throw away one.

More generally, the two return value API is honestly kinda _weird_ for users, especially those coming from Python. My understanding has always been that it was a necessary evil in order to get acceptable performance for `destructure`, but were that not the case then 

> `remake!(m, v)` could equally well make it later (modulo benchmarking).

would've been a less confusing option.

Moreover, while `re` from `destructure` is somewhat easier to thread deep into a training loop by including it in a loss function, a `re` from this theoretical `somefun` would not be. e.g. instead of being able to call `(save|load)model!` directly for checkpoints, the loop would have to hold onto a `re` and update it every so often. That runs the risk of closing over possibly stale model state and makes everything from serialization to distributed training more difficult.

But my biggest concern (which I apologize for not thinking of earlier) is more basic. In PyTorch, I can do `model.nested[1].path.load_state_dict(sd["nested.1.path"])` with impunity. With `loadparams!`, this would be a direct translation to `loadparams!(model.nested[1].path, sd.nested[1].path)`. With a simple implementation of `m0, re = somefun(...)`, this is not possible because `re` only works at the top level. One could make some fancy overloads for `getindex`, `getproperty` etc. on the type of `re` to enable such nested loading, but to me that seems like adding unnecessary internal complexity just to maintain an already unfamiliar user-facing API. This kind of subset loading is not a niche case either. I've seen/used it in real world code for transfer learning, self-supervised training and other forms of model surgery.

mcabbott:
> that it was a necessary evil in order to get acceptable performance for destructure, but were that not the case then

No, I don't think this is true. The old implementation literally closes over `m` and calls a reconstructor function, no special work is done up front. 

The separated version was what I proposed in https://github.com/FluxML/Functors.jl/pull/31,  but I got the impression everyone preferred the `destructure` form, hence https://github.com/FluxML/Optimisers.jl/pull/54 .

ToucheSir:
I had a look back at the original issues, and I think I got some wires crossed reading https://github.com/FluxML/Flux.jl/pull/986#issuecomment-589619486. For posterity, https://github.com/FluxML/Flux.jl/pull/799 seems to have the actual design discussion.

That said, the point about performance is true now that `Optimisers.destructure` exists and saves offsets for faster restructuring. FWIW I thought `fvec` was fine, but it made a lot more sense to have the functionality in Optimisers instead of Functors.

darsnack:
> It could be helpful to know what features are missing from a basic implementation like that.

The main missing feature is the kind of structural error checking that you get from this design. Collecting into a flat vector could always silently be wrong, and when it does catch an error, the best you can say is "some parameters are missing." Here, we can be more helpful since we know exactly what structure is being loaded.

It also allows for the convenient syntax of `loadmodel!(m1[1][3], m2[4])` like Brian mentioned.

----

Okay, so we've had more discussion on this PR than I expected (which is good)! And I think at a high level the concept of "model -> simple structure" is shared between this and `de/restructure`. There's some disagreement around when/where `re` is necessary, and whether including in loading makes sense at all. I think we need to concretely say what's the final function(s) and how the behave to be able to move on this discussion.

For now, we can let this PR stew if needed. I will say that this function is needed to make Metalhead.jl work with pre-trained models. So, I propose the following path forward:
- if we come to an agreement, we merge and move on
- if we can't agree, we use the same old `loadparams!` name and defer this discussion to a later PR (this way we are simply extending what's already there and not introducing anything new)

darsnack:
I've also significantly simplified the implementation so that most custom types will participate in the thorough error checking for free.

This PR considers any type for which `all(Functors.isleaf, Functors.children(x))` is true as "block" for loading. Any type that wants to be treated the same but fails that check can override `isloadleaf`. A type that wants custom error checks/messages can override `loadto!(m::T, mbar)` for their `T`. Neither of these overrides will be necessary for the vast majority of types. All that's required is `@functor` as always.

I will add documentation soon too which should hopefully help narrow the discussion as well.

darsnack:
What are our final thoughts here?

I see several transformations we want to do:
1. `T1`: Nested structure -> simple nested structure
2. `T2`: (Simple) nested structure -> flat structure
3. `T3`: Flat structure -> (simple) nested structure
4. `T4`: Simple nested structure -> nested structure

`T2` must produce a "reconstructor" and `T3` must consume it. `T1` and `T4` don't need a reconstructor, they just need some sense of alignment in the structures that it walks.

There are all kinds of ways to write the implementations for these transforms to share code. Ultimately, I think the code that they really share is the concept of walking a tree from Functors.jl. So my vote here is not to write `restructure` and co. to "just work" for `loadparams!` but to move as much shared code into `fmap`.

Given #1882, I would suggest we move forward with this PR and place either `loadparams!` (name change) or `loadparams!(m, ::Params)` on the deprecation path. We can make note in the error message that these forms don't work for `bias=false`. I can rebase this PR once #1882 is merged and fix the edge cases.

ToucheSir:
I'm not sure if `fmap` is the best place to hang all this common functionality, but agreed that we can worry about abstracting bits out after this interface has landed.

darsnack:
Okay I think this is ready. The broken tests after #1882 are now passing, and I have marked them as such. The only broken CI should be the `libblas` related issues on master.

@ericphanson you shouldn't need to override any of this for a custom layer, but I have included a section in the docs anyways (feel free to provide some feedback).

ericphanson:
Perhaps the tests should be more extensive and include batchnorm layers and such? I’ve been using this model in LegolasFlux’s tests: https://github.com/beacon-biosignals/LegolasFlux.jl/blob/80569ab63a8248a8a063c76e0bbf701f4ada9bd4/examples/digits.jl#L33 - you could copy it here.

src/loading.jl
```diff
@@ -0,0 +1,111 @@
...
+
+Copy `x̄` to `x` or throw `err` when their sizes are mismatched.
+By default, use `copyto!` when `x` and `x̄` are arrays.
+Otherwise, just return `x`.
+"""
+loadleaf!(x, x̄, err) = x
+function loadleaf!(x::AbstractArray, x̄, err)
+  x .= x̄
+  return x
+end
```
mcabbott, line 24: I think this `x .= x̄` method exists specifically for when the src model has bias=false, and the destination has an array.

In the reverse situation of the dst having false & src having an array, at present the array is simply dropped. That seems dangerous. Shouldn't it check that this is all zero first?

I still wonder a bit whether this case is worth catering to at all. Why handle this exact one kind of models differing, but make all others errors? I guess it is special-cased in that the "same" layers permit you to have or not have this parameter, but still.

src/loading.jl
```diff
@@ -0,0 +1,111 @@
...
+  err = DimensionMismatch("Tried to load $m̄ into $m but the parameter sizes do not match.")
+  foreach((l, l̄) -> loadleaf!(l, l̄, err), ls, l̄s)
+
+  return m
+end
+
+"""
+    loadmodel!(m, m̄)
+
+Copy all the parameters (trainable and non-trainable) from `m̄` to `m`.
```
mcabbott, line 55: One notation comment: Elsewhere `m̄` is usually the gradient with respect to `m`. I think it's a bit confusing to also use it for just another version. Could this be called something else? 

`loadmodel!(dst, src)` is more verbose but super-clear. Or if unicode, maybe there's something more suggestive?

ToucheSir: `dst` and `src` has the benefit of consistency and familiarity. The fact this is working with models is already covered by the method name :)

src/loading.jl
```diff
@@ -0,0 +1,111 @@
...
+function loadto!(m::T, m̄::S) where {T, S}
+  (nameof(T) == nameof(S)) || throw(ArgumentError("Tried to load $m̄ into $m."))
+  
+  ls, _ = functor(m)
+  l̄s, _ = functor(m̄)
+  (keys(ls) == keys(l̄s)) ||
+    throw(ArgumentError("Tried to load $m̄ into $m but the structures do not match."))
+
+  err = DimensionMismatch("Tried to load $m̄ into $m but the parameter sizes do not match.")
+  foreach((l, l̄) -> loadleaf!(l, l̄, err), ls, l̄s)
```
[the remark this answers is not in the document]
darsnack: The reason was to use `fmap`. In general, I wish we didn't have to write out the recursion manually. Walking the models is a hotbed for corner case bugs, so I try to use Functors built-ins as much as possible to catch and fix them in one place.

Either way, the non-`fmap` implementation is much simpler and automated. So definitely worth a custom recursion.

src/loading.jl
```diff
@@ -0,0 +1,111 @@
...
+    isloadleaf(x)
+
+Return `true` whenever `x` should be treated as a "leaf node"
+for the purposes of loading parameters.
+By default, `isloadleaf` returns `true` if [`Functors.isleaf`](@ref)
+is `true` for all [`Functors.children(x)`](@ref `Functors.children`).
+
+You can override this function for a specific type if needed.
+"""
+isloadleaf(x) = all(Functors.isleaf, Functors.children(x))
```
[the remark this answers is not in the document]
darsnack: The reason was getting `fmap` to stop in the right place for almost every layer, but also allow overriding when the guess is wrong. But the new implementation doesn't have this issue.

test/utils.jl
```diff
@@ -340,18 +340,42 @@ end
...
+    @test m1[2].σ == relu
+    loadmodel!(m5, m6)
+    @test m5[1].weight == m6[1].weight
+    @test m5[2][1].weight == m6[2][1].weight
+    @test m5[2][1].bias == Zeros()
+
+    @test_throws ArgumentError loadmodel!(m1, m3)
+    @test_throws DimensionMismatch loadmodel!(m1, m4)
+    @test_throws ArgumentError loadmodel!(m1, m5)
+  end
```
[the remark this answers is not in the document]
darsnack: Added a bunch more tests. Getting the variations of `Chain` to work required dropping support for `loadmodel!(m, ::AbstractVector)`. I could always add a layer of indirection to keep support, but I don't think this variation is used anywhere as you mentioned. So, I'll just drop it for now.

src/loading.jl
darsnack: These paths take care of the `bias=false` cases in both directions, but that is not all they are for. The nice part of `bias` not being `Zero` is that the "correct" thing isn't a special rule but just one case of a larger policy.

That policy is to only copy/write to a destination that is a mutable array. The entire function here is strictly mutating. I'll leave reconstructing/non-mutable children for a separate PR. I think it needs more discussion on how we decide which non-mutable children should be copied or not. Perhaps this PR should get more strict and only copy on `AbstractArray{<:Number}`.

darsnack:
Bump?

src/loading.jl
```diff
@@ -0,0 +1,86 @@
...
+Copy `src` to `dst` or throw `err` when their sizes are mismatched.
+By default, use `copyto!` when `dst` and `src` are arrays.
+When only `dst` is an array, set every element to `src`.
+Otherwise, just return `dst`.
+"""
+loadleaf!(dst, src, err) = dst
+function loadleaf!(dst::AbstractArray, src, err)
+  dst .= src
+  return dst
+end
```
mcabbott, line 13: > These paths take care of the bias=false cases in both directions

This still seems asymmetrical to me. If src has bias=false it zeros dst, but if dst has bias=false, then it will silently throw away nonzero values. Is this clearly the right thing to do? Are there use cases which need it? 

I wonder if it wouldn't be simpler to demand that all arrays must match (although allowing non-array content like differing activation functions).

But if we do allow `dst .= false`, then I think `dst === false` should check `iszero(src)`, else error.

Also, I am not a fan of documenting `loadleaf`. I think the description of the one API function `loadmodel!` should be clear enough to know what it will do, without knowing any internals of Functors, or this other function it happens to call.

src/loading.jl
```diff
@@ -0,0 +1,86 @@
...
+julia> dst[2].bias == src[2].bias
+true
+```
+
+See [`Flux.loadleaf!`](@ref) for more details on the copy behavior.
+
+!!! warning
+    This function allows `src` to be a `Params` for backwards-compatibility.
+    You should avoid using `loadmodel!` this way, because it skips most of the structural
+    checking used when `src` is also a nested structure. Silent errors may occur.
```
mcabbott, line 65: Could this fallback just keep the old name, and make the new name only structural?

test/utils.jl
```diff
@@ -340,18 +340,136 @@ end
...
+    loadmodel!(chain2[end - 1], chain1[end - 1])
+    loadmodel!(chain2[3], chain1[3])
+    @test chain2[end - 1].weight == chain1[end - 1].weight
+    @test chain2[3].μ == chain1[3].μ
+    @test chain2[2].bias != chain1[2].bias
+
+    # test shared weights
+    m1 = Chain(Dense(10 => 5), Dense(5 => 2))
+    m2 = Chain(Dense(transpose(m1[2].weight)), Dense(permutedims(m1[1].weight)))
+    m3 = Chain(Dense(m1[1].weight), Dense(m1[2].weight))
```
mcabbott, line 463: Can you say what you mean by shared here? I don't think the fmap cache would see any duplicates here.

src/loading.jl
```diff
@@ -0,0 +1,86 @@
...
+julia> src = Chain(Dense(5 => 2), Dense(2 => 1));
+
+julia> all(isone, dst[1].weight)
+true
+
+julia> dst = loadmodel!(dst, src)
+Chain(
+  Dense(5 => 2),                        # 12 parameters
+  Dense(2 => 1),                        # 3 parameters
+)                   # Total: 4 arrays, 15 parameters, 316 bytes.
```
mcabbott, line 48: Since this doesn't reconstruct, I think what's displayed below should be the mutated destination, not the returned value:
```suggestion
julia> loadmodel!(dst, src);
```
And perhaps no need to print it.

src/loading.jl
```diff
@@ -0,0 +1,86 @@
...
+
+julia> all(isone, dst[1].weight)
+false
```
mcabbott, line 51: Maybe less cryptic? Or just omit?
```suggestion
julia> dst[1].weight ≈ ones(2, 5)
false
```

src/loading.jl
```diff
@@ -0,0 +1,86 @@
...
+"""
+    loadmodel!(dst, src)
+
+Copy all the parameters (trainable and non-trainable) from `src` to `dst`.
+
+`loadmodel!` recursively walks the [`Functors.children`](@ref) of `dst` and `src`
+calling `loadleaf!` on any pair of children where [`Functors.isleaf`](@ref) is true.
+It throws an error whenever:
+- `dst` and `src` do not share the same fields (at any level)
+- the sizes of leaf nodes are mismatched between `dst` and `src`
```
mcabbott, line 28: I wonder if "parameter arrays" is clearer, in case you think Dropout has a non-trainable scalar parameter:
```suggestion
Copy the contents of all parameter arrays (trainable and non-trainable) from `src` to `dst`.

Recursively walks `dst` and `src` together, using Functors.jl, throwing an error if these 
do not share the same field names (at any level). Parameter arrays must be of matching size. 
Non-array elements (such as activation functions) need not match.
An all-zero bias array can be copied to or from absent bias, encoded `bias = false`.
```
I don't think "the sizes of leaf nodes are mismatched" is quite accurate since `size(false)` need not match at present. If we do allow that, maybe it should be explained.

I'd give less weight to the details of Functors.

test/utils.jl
darsnack: Okay I have a better idea of what you meant by shared weights now. Something like the model below, but the tests I wrote just document the current behavior. What do we want? If someone loads an untied set of weights into a set of tied weights, should we warn or error?

> fmap cache

Note that there is no longer `fmap` here, so it's caching behaviors are unused. This kind of diversion is how we recurse things is why I prefer to use `fmap` over manually using `functor`. But there is no way to hold "one level up" in the error message with `fmap` (at least not that I could figure out).

```suggestion
    # test shared weights
    encoder = Chain(Dense(10 => 5), Dense(5 => 2))
    decoder = Chain(Dense(transpose(m1[2].weight)), Dense(permutedims(m1[1].weight)))
    m1 = Chain(encoder, decoder)
    m2 = Chain(Chain(Dense(10 => 5), Dense(5 => 2)), Chain(Dense(2 => 5), Dense(5 => 10)))
    loadmodel!(m1, m2)
    # what behavior do we want
    @test encoder[2].weight === parent(decoder[1].weight)
    @test encoder[2].weight == permutedims(m2[2][1].weight)
    @test encoder[1].weight == m2[1].weight
    @test encoder[1].weight != permutedims(decoder[2].weight)
```

src/loading.jl
darsnack: Sure but I eventually want to deprecate that path altogether.

mcabbott: Agreed. But what I mean is, perhaps cleaner to leave it part of the deprecated function, rather than ever have a method of the new function which does this. Nobody should update code twice.

test/utils.jl
darsnack: Erroring seems like the correct thing to me too.

mcabbott: My quick attempt to write this, starting from `fmap` and adapting...
```julia
function modelcopy!(x, y; cache = IdDict())
  if haskey(cache, x)
    # If we've seen (and thus updated) x already, check that new y is the same, or nothing:
    leaf_tiecheck(x, y)
    return cache[x]
  end
  cache[x] = if Functors.isleaf(x)
    leafcopy!(x, y)
  else
    # _default_walk has mismarch errors from functor(T, y), not as pretty as could be
    Functors._default_walk((x, y) -> modelcopy!(x, y; cache), x, y)
  end
end

# Main case, error on immutable array x (at present):
leafcopy!(x::AbstractArray, y::AbstractArray) = x .= y  # demands equal size, could make a prettier error
leaf_tiecheck(x::AbstractArray, y::AbstractArray) = x == y || error("tied weights in `dst` must be identical in `src`, or else `nothing`")

# Bool represents absent bias, equivalent to zero bias:
leafcopy!(x::AbstractArray, y::Bool) = x .= y
leafcopy!(x::Bool, y::AbstractArray) = iszero(y) ? x : error("can't copy nonzero bias in `src` to absent bias in `dst`")
leaf_tiecheck(x::AbstractArray, y::Bool) = iszero(x) || error("tied weights in `dst` cannot have some zero/absent bias, others nonzero")
leaf_tiecheck(x::Bool, y::AbstractArray) = iszero(x) || error("tied weights in `dst` cannot have some zero/absent bias, others nonzero")

# Fallback, for e.g. activation functions, but also missing nodes:
leafcopy!(x, y) = x
leaf_tiecheck(x, y) = nothing


# This fmapstructure produces a tree of tuples, with all ties removed, ideal for saving?
sh = [1,2]; m = (x=sh, y=[3,4], z=sh, b=false, a=sin);
mflat = fmapstructure(identity, m; prune=nothing)
# That is accepted as src by modelcopy!
modelcopy!(m, (x = [1, 20], y = [300, 4000], z = nothing, b=[0,0], a=nothing))
modelcopy!(m, (x=[10,2], y=[3,4], z=[10,2], b=false, a=nothing))  # also ok
modelcopy!(m, (x=nothing, y=[30,40], z=nothing, b=nothing, a=nothing))  # doesn't change m.x
# Mismatch of a weight which should be shared is an error:
modelcopy!(m, (x=[1,2], y=[3,4], z=[100,200], b=nothing, a=nothing))

# Could simplify further:
function simpletree(m)
    fmapstructure(m; prune=nothing) do x
        # We know isleaf(x), but further keep only values modelcopy! will accept:
        x isa AbstractArray && return x
        x === false && return x
        nothing
    end
end
simpletree(m)  # has a = nothing
```

darsnack:
Okay I think this is ready for another review. All the comments should have been addressed, and I added support for tied parameters.

src/loading.jl
```diff
@@ -0,0 +1,96 @@
...
+"""
+function loadmodel!(dst, src; cache = Base.IdSet())
+  ldsts, _ = functor(dst)
+  lsrcs, _ = functor(src)
+  (keys(ldsts) == keys(lsrcs)) ||
+    throw(ArgumentError("Tried to load $src into $dst but the structures do not match."))
+
+  err = DimensionMismatch("Tried to load $src into $dst but the parameter sizes do not match.")
+  foreach(ldsts, lsrcs) do ldst, lsrc
+    if _parent(ldst) in cache # we already loaded this parameter before
```
mcabbott, line 79: Can you explain what `_parent` is trying to do?

src/loading.jl
```diff
@@ -0,0 +1,96 @@
...
+
+_parent(x) = x
+_parent(x::AbstractArray) = parent(x)
+
+_tie_check(dst::AbstractArray, src::AbstractArray) = dst == src
+_tie_check(dst, src) = true
+
+_bool_tie_check(dst::Bool, src::AbstractArray) = iszero(dst)
+_bool_tie_check(dst::AbstractArray, src::Bool) = iszero(dst) && iszero(src)
+_bool_tie_check(dst, src) = true
```
mcabbott, line 25: Can't these just be methods of one function?

src/loading.jl
darsnack: It handles cases like `transpose` which are shown in the tests. It's just `Base.parent` but with a no-op path for non-arrays.

mcabbott: I guess what I mean is: Why differ from Functors here? Having two different rule for which parameters are tied seems likely to result in lots of weird edge cases.

I agree that Functors ought to regard a matrix and its transpose as being tied. But it got stuck on the precise rule for this. If Functors 0.3 has this, then Flux can change to match exactly.

src/loading.jl
darsnack: Done.

src/loading.jl
darsnack: Makes sense, done.

src/loading.jl
```diff
@@ -0,0 +1,90 @@
...
+
+"""
+    loadmodel!(dst, src)
+
+Copy all the parameters (trainable and non-trainable) from `src` to `dst`.
+
+`loadmodel!` recursively walks the [`Functors.children`](@ref) of `dst` and `src`
+calling `copyto!` on any pair of children where [`Functors.isleaf`](@ref) is true.
+It also handles "absent" parameters such as `bias == false`.
+It throws an error whenever:
```
mcabbott, line 35:
```suggestion
Copy all the parameters (trainable and non-trainable) from `src` into `dst`.

Recursively walks `dst` and `src` together using [`Functors.children`](@ref),
and calling `copyto!` on parameter arrays.
Non-array elements (such as activation functions) need not match.
An all-zero bias array can be copied to or from absent bias, encoded `bias = false`.

Throws an error when:
```
It's inaccurate to say it calls `copyto!` on all leaf nodes. Stressing "parameter arrays" so that e.g. dropout `p` is excluded.

I think we can say exactly how it "handles" false, and should avoid "such as" when there's exactly one such case.

src/loading.jl
mcabbott: Or print it if you want to stress that it does return `dst`, but don't assign to `dst`, since it always works by mutation.

mcabbott (approved):
LGTM, just docstring comments

src/loading.jl
```diff
@@ -0,0 +1,88 @@
...
+_bool_tie_check(dst, src) = true
+
+"""
+    loadmodel!(dst, src)
+
+Copy all the parameters (trainable and non-trainable) from `src` into `dst`.
+
+Recursively walks `dst` and `src` together using [`Functors.children`](@ref),
+and calling `copyto!` on parameter arrays.
+Non-array elements (such as activation functions) need not match.
```
ericphanson, line 34: It's not clear to me what this means-- I think it would be more clear if it's after "Throws an error when", since at this point in the docstring I'm thinking about copying from `src` to `dst`, so I don't know what "needs not match" means (sometimes it'll copy? idk). But what I think it means is it ignores these and doesn't copy or throw errors.

src/loading.jl
```diff
@@ -0,0 +1,88 @@
...
+An all-zero bias array can be copied to or from absent bias, encoded `bias = false`.
```
ericphanson, line 35: > An all-zero bias array can be copied to or from absent bias, encoded `bias = false`.

This sentence is unclear to me

darsnack:
@ericphanson let me know if the new version clears things up

src/loading.jl
```diff
@@ -0,0 +1,91 @@
...
+"""
+    loadmodel!(dst, src)
+
+Copy all the parameters (trainable and non-trainable) from `src` into `dst`.
+
+Recursively walks `dst` and `src` together using [`Functors.children`](@ref),
+and calling `copyto!` on parameter arrays.
+Non-array elements (such as activation functions) are not copied
+and do not need to match between `dst` and `src`.
+Inactive parameters, encoded by `false` in place on an array,
```
ericphanson, line 36: What about this? (Is this the right scenario, or can `src` have `false` too?)
```suggestion
If `dst` has the boolean `false` where `src` has an all-zero array, no error will be raised (and no values will be copied). This facillitates using `false` for inactive parameters.
```

src/loading.jl
```diff
@@ -0,0 +1,91 @@
...
+can be copied to and from all-zero arrays.
+Attempting to copy a non-zero array to/from an inactive parameter will throw an error.
```
ericphanson, line 38: Maybe this is more clear?
```suggestion
Inactive parameters can be encoded by using the boolean value `false` instead of an array.
If `src` or `dst` has `false` where the other model has an all-zero array, no error will be raised (and no values copied). However, attempting to copy a non-zero array to/from an inactive parameter will throw an error.
```

darsnack: Okay I used this with a few modifications, because your comment made me realize that the behavior is not 1-1 like the docstring implies.

src/loading.jl
```diff
@@ -0,0 +1,92 @@
...
+Copy all the parameters (trainable and non-trainable) from `src` into `dst`.
+
+Recursively walks `dst` and `src` together using [`Functors.children`](@ref),
+and calling `copyto!` on parameter arrays.
+Non-array elements (such as activation functions) are not copied
+and do not need to match between `dst` and `src`.
+Inactive parameters can be encoded by using the boolean value `false` instead of an array.
+If `dst == false` and `src` is an all-zero array, no error will be raised (and no values copied);
+however, attempting to copy a non-zero array to an inactive parameter will throw an error.
+Likewise, copying `src == false` to any `dst` array is valid, but copying `src == true` will error.
```
ericphanson, line 39: tiny nitpick: `src` isn't `== false`, but rather one of its values:
```suggestion
Likewise, copying a `src` value of `false` to any `dst` array is valid, but copying a `src` value of `true` will error.
```

src/loading.jl
mcabbott: Can we shorten this somehow? It seems a very tiny edge case about layers with `bias=false` vs. models with an actual bias array. Most likely this will never happen in real life. Yet somehow it gets an essay describing all possible paths... 

How about just "Zero bias and `bias=false` are considered equivalent."?

ericphanson: Could put it under extended help? See point 11 of https://docs.julialang.org/en/v1/manual/documentation/.

IMO magic should at least be documented...

src/loading.jl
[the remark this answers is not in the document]
darsnack: Strictly speaking, the rule applies to anything not just bias. But bias should be the only occurrence of this rule in practice.

@ericphanson how do you like this shortened version?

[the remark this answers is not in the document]
ericphanson: I don't really like it since like you say, it sounds like this only applies to `bias` (and to vectors), and it doesn't give the full semantics. Re-https://github.com/FluxML/Flux.jl/pull/1875#discussion_r841768510, I think if we want a short docstring, then we should just put more of the details under extended help, so it only shows up in the online docs or if you do `?? loadparams!` in the REPL.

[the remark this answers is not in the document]
mcabbott: Yes. But `bias=false` is the one official API for making variants of layers, which aims to handle. Others, like `affine=false`, are not -- the models must simply match.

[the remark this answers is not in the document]
mcabbott: Alternatively, the need to write an essay describing this suggests that it might be unnecessary complication. Is there an actual use case? If you have saved a model, and now you reconstruct and load the weights, then the models agree -- in this and all other details. 

It seems a pretty weird edge case to have saved a model whose training never touched the bias, and also to have edited the original model before re-loading, and now want loading to just work.

src/loading.jl
```diff
@@ -0,0 +1,92 @@
+loadleaf!(dst, src, err) = dst
```
mcabbott, line 1: I also wonder if there should be more errors here:
```suggestion
loadleaf!(dst, src, err) = dst
loadleaf!(dst::AbstractArray, src, err) = error()
loadleaf!(dst, src::AbstractArray, err) = error()
```

src/loading.jl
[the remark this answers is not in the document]
mcabbott: BTW I think the reason allowing other mismatches is, I think, that the other half of this "model serialization tool" is something like 
```julia
function simpletree(m)
    fmapstructure(m; prune=nothing) do x
        # We know isleaf(x), but further keep only values modelcopy! will accept:
        x isa AbstractArray && return x
        x === false && return x  # if we keep that...
        nothing
    end
end
```
which should produce a nested set of NamedTuples, with only the details this thing will load --- no layer types, no activation functions, and tied arrays appear only once. If `nothing` is the magic value for this, then we probably want a method to ignore it on loading:
```
loadleaf!(dst, src::Nothing, err) = dst
```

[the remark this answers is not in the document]
darsnack: I think saving a model trained w/o bias and loading into a model w/ bias that you intend to fine-tune is a pretty reasonable/common use case. This is the pre-trained model flow, not the save my own model and load my own model flow.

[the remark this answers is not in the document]
ToucheSir: I've mentioned it before, but PyTorch has a pretty compelling model for dealing with these mismatches: [`load_state_dict`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.load_state_dict) errors by default, but also has a non-strict mode where it returns a symmetric diff of the source and destination model trees. All this to say that the behaviour here need not be set in stone, and that we should strive to be at least as good about telling the user about _how_/_why_ loading failed when it does.

darsnack:
Okay I went with the extended help suggestion, but if special casing `false` is the only thing holding up this PR, then let's save it for another day and merge the rest today.

PR #1877, opened 2022-02-17
Truncated normal initialisation for weights

theabhirath:
This PR adds a truncated normalisation method for the initialisation of weights in layers. This has been used in recent papers including [ConvNeXt](https://arxiv.org/abs/2201.03545) (work is currently ongoing to implement a Julia version of the same at https://github.com/FluxML/Metalhead.jl/pull/119) and is also incorporated in [PyTorch master](https://github.com/pytorch/pytorch/blob/c22b8a42e6038ed2f6a161114cf3d8faac3f6e9a/torch/nn/init.py#L22). Unfortunately there's an additional dep in the form of SpecialFunctions.jl, but thankfully it seems to be quite lightweight

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


src/utils.jl
```diff
@@ -301,6 +302,45 @@ end
...
+  if (μ < a - 2σ) || (μ > b + 2σ)
+    @warn "mean is more than 2 std from [a, b] in truncated_normal. The distribution of values 
+           may be incorrect."
+  end
+  l = norm_cdf((a - μ) / σ)
+  u = norm_cdf((b - μ) / σ)
+  x = rand(rng, dims...) * 2(u - l) .+ (2l - 1)
+  x = erfinv.(x)
+  x = clamp.(x * σ/√2 .+ μ, a, b)
+  return convert.(Float32, x)
```
mcabbott, line 338: For efficiency these broadcasts should be combined. 

Why not work in Float32 the whole way, did this cause problems? 

Also, something is wrong here:
```
julia> std(truncated_normal(10^6; a=-5, b=5))
0.49986008f0

julia> extrema(truncated_normal(10^6; a=-2, b=4))
(-0.99999887f0, 1.9829093f0)
```

src/utils.jl
```diff
@@ -301,6 +302,45 @@ end
 sparse_init(dims...; kwargs...) = sparse_init(Random.GLOBAL_RNG, dims...; kwargs...)
 sparse_init(rng::AbstractRNG; init_kwargs...) = (dims...; kwargs...) -> sparse_init(rng, dims...; init_kwargs..., kwargs...)
 
+"""
+    truncated_normal([rng=GLOBAL_RNG], dims...; μ = 0, σ = 1, a = -2., b = 2.)
```
mcabbott, line 306: Do we want `μ, σ, a, b` as keywords? Elsewhere I think the goal is never to require anyone to type greek letters. And I think `a,b` are much too obscure, `left, right` or `lo, hi` seem more obvious.

```suggestion
    truncated_normal([rng=GLOBAL_RNG], dims...; mean = 0, std = 1, lo = -2, hi = 2)
    truncated_normal([rng]; kw...) = (dims...) -> truncated_normal(rng, dims...; kw...)
```
I think this should mention the curried form somehow. But perhaps writing it out like this is too verbose?

src/utils.jl
```diff
@@ -301,6 +302,45 @@ end
...
+`σ`, with values outside `[a, b]` redrawn until they are within the bounds. The method used for 
+generating the random values works best when `a ≤ mean ≤ b`.
+
+# Examples
+```jldoctest; setup = :(using Random; Random.seed!(0))
+julia> Flux.truncated_normal(3, 2)
+3×2 Matrix{Float32}:
+ -0.113785  -0.627307
+ -0.676033   0.198423
+  0.509005  -0.554339
```
mcabbott, line 319: How much does anyone learn from this? Maybe something like:
```
julia> truncated_normal(3, 4) |> summary
"3×4 Matrix{Float32}"

julia> round.(extrema(truncated_normal(10^6)); digits=3)
(-1.0f0, 1.0f0)

julia> using Plots

julia> histogram(truncated_normal(10^6; σ=2, a=-4, b=4); bins=-4:0.1:4, alpha=0.5)

julia> histogram!(randn(10^6); bins=-4:0.1:4, alpha=0.5)
```
Or even nicer, something inline with UnicodePlots.

test/utils.jl
```diff
@@ -146,6 +146,15 @@ end
     end
   end
 
+  @testset "truncated_normal" begin
+    for sz in [(100,), (100, 100), (2, 3, 32, 64), (2, 3, 4, 32, 64)]
+      v = truncated_normal(sz...)
```
mcabbott, line 151: I don't think that testing the same thing for many sizes helps much. 

The failure modes which tests should catch include:
* Are the extrema of the result close to the numbers provided, both default and not?
* Does the distribution have the right statistics?
* If I call it with `mean = 0.5`, does the result become Float64? Sim. other parameters.
* Does the curried form work, pass arguments through? Does the form which takes RNG work?

theabhirath:
> This doesn't seem to be doing the resampling strategy like the docstring suggests. Instead, it is sampling from a uniform distribution and applying the inverse CDF for the truncated normal.
> 
Yep, started working on one approach and ended up with another because I found it to be more common 🤦🏽‍♂️ I'll fix the docstring

> In theory, the `clamp` should only be necessary for floating point errors. Can we remove the `clamp` and see how often it fails (if at all)?

I was wondering about this too. Mathematically it's well within the bounds unless possibly some absolute sorcery is unleashed, but the PyTorch master version had it and I was worried I would miss out on some edge cases

src/utils.jl
```diff
@@ -301,6 +302,45 @@ end
...
+  end
+  l = norm_cdf((a - μ) / σ)
+  u = norm_cdf((b - μ) / σ)
+  x = rand(rng, dims...) * 2(u - l) .+ (2l - 1)
+  x = erfinv.(x)
+  x = clamp.(x * σ/√2 .+ μ, a, b)
+  return convert.(Float32, x)
+end
+
+truncated_normal(dims::Integer...; kwargs...) = truncated_normal(Random.GLOBAL_RNG, dims...; kwargs...)
```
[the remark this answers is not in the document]
theabhirath: I just followed the style of the rest of the init layers 😅

mcabbott:
It seems OK to talk about resampling as an explanation of what the truncated normal distribution *is*. I guess it should't imply that this is how the implementation works, but what algorithm `randn` uses is also not something we need to know to call it.

test/utils.jl
theabhirath: These are great suggestions, thanks! I'll get to writing these, I was breaking my head over what I could possibly write 😅 For the mean = 0.5 - yep, it indeed does result in a Float64 result. This was why I had a blanket convert earlier, but instead maybe I could convert the individual parameters instead to Float32 and then start the calculations?

src/utils.jl
theabhirath: For Float32 I was throwing the kitchen sink because I wanted to account for individual params. I've mentioned that below 😅. For the results, I don't understand what exactly seems wrong?

src/utils.jl
theabhirath: I'll fix the Greek letters 👍🏽 The curried form hasn't been mentioned for the others, so I just followed precedent - I can change that if it seems more appropriate

src/utils.jl
mcabbott: If I truncate far from the centre, then this should become a normal distribution with the given standard deviation, right?

And, the limits I provide should be the extrema of the result, whereas instead it seems to use half their values.

theabhirath: Ooh yeah that change seems to be missing, there's a multiplication sign gone wrong. You're right. I'll push that

test/utils.jl
theabhirath: The Float64 conversion in the end definitely helps accuracy slightly but I'll take the tradeoff because the arrays will most likely be large

mcabbott: Since `rand(Float32, ...)` is sure to return an Array, maybe all operations should write back into this same array. Then there will be just one allocation. And since `setindex!` automatically converts, there should be no need to sanitise the inputs, as they cannot cause type promotion. That sounds like the fast & tidy way.

theabhirath:
Okay the API and the docstring should be fine now. I'll come up with more specific cases for tests and then push this through for another round of reviews😅

test/utils.jl
theabhirath: Experiments since showed that there's too much error if we breach the [-2, 2] region for `lo` and `hi`. I switched back to Float64 ops with an explicit convert at the end but in case the consensus is that not many people will want to tweak the lo and hi too much then I can switch it back. The error is significant thought, that much is for sure - even at [-5, 5] it's around 0.1

darsnack:
There are some review comments from my previous review related to the `Float32` issue that don't seem resolved? I didn't see a decision to use `Float64`, so I have unmarked those comments as resolved. Do you mind taking a look again?

All the constants and `rand` should be `Float32` when they are created instead of converting at the end (like discussed above). This does keep things open to someone setting the keyword arguments to `Float64` and getting a `Float64` result. But this is a standard application of Julia's type promotion system, so I personally don't think we should forbid it. I see that other initialization functions are inconsistent about this behavior. Perhaps we should settle in this PR what kind of behavior we want.

Even if the decision is to always return a `Float32`, it is cheaper to convert the scalar output of `norm_cdf` to `Float32` than to convert the array at the end.

darsnack:
> I was wondering about this too. Mathematically it's well within the bounds unless possibly some absolute sorcery is unleashed, but the PyTorch master version had it and I was worried I would miss out on some edge cases

My worry is that we silently sample the wrong distribution since the support of the samples will appear correct. So if we keep it in, I'd like to know that it's only for numerical reasons first (by seeing the resulting histogram without it).

src/utils.jl
darsnack: UnicodePlots has a `histogram` function. Using that would be nice here.

test/utils.jl
mcabbott: What is 0.1, I'm not sure I follow?

But even if you compute in Float64, it can probably still be one fused loop writing back into the same array. 
```
xs = rand(Float32, size...)
broadcast!(xs, xs) do x
  y = Float64(x) + ... # do stuff
end
```

theabhirath: > What is 0.1, I'm not sure I follow?

The error 😬

> But even if you compute in Float64, it can probably still be one fused loop writing back into the same array.

Sure, I'll make this change

theabhirath:
> There are some review comments from my previous review related to the Float32 issue that don't seem resolved?

Oh, sorry about that. As I noted [here](https://github.com/FluxML/Flux.jl/pull/1877#discussion_r809310827), I did switch to Float32 initially but the error was quite noticeable and so I decided to switch back to calculating the result as a Float64 and converting it in the end. I'll write back to the same array and see if that works out

theabhirath:
Writing back to the array seems to be working fine - the error is under control too for now, so I think it should be okay. I've also changed the RNGs for the init layers to reflect [this](https://github.com/FluxML/Flux.jl/pull/1877#discussion_r809305659)

src/utils.jl
```diff
@@ -188,9 +188,50 @@ function kaiming_normal(rng::AbstractRNG, dims...; gain = √2f0)
...
+  if (mean < lo - 2 * std) || (mean > hi + 2 * std)
+    @warn "Mean is more than 2 std from [a, b] in truncated_normal. The distribution of values may be incorrect." maxlog=1
+  end
+  l = norm_cdf((lo - mean) / std)
+  u = norm_cdf((hi - mean) / std)
+  xs = rand(rng, Float32, dims...)
+  broadcast!(xs, xs) do x
+    x = x * 2(u - l) + (2l - 1)
+    x = erfinv(x)
+    x = clamp.(x * std * √2f0 + mean, lo, hi)
```
mcabbott, line 227: Since x is Float64 here, may as well use`√2`. And no `.` needed:
```suggestion
    x = clamp(x * std * √2 + mean, lo, hi)
```

I don't see NaN, but do see Inf, without the clamp:
```julia
julia> count(isnan, truncated_normal(10^8; lo=-100, hi=100, std=10, mean=7))
0

julia> count(isinf, truncated_normal(10^8; lo=-100, hi=100, std=10, mean=7))
6

julia> extrema(truncated_normal(10^8; lo=-100, hi=100, std=10, mean=7))
(-Inf32, 59.94704f0)
```

src/utils.jl
```diff
@@ -188,9 +188,50 @@ function kaiming_normal(rng::AbstractRNG, dims...; gain = √2f0)
...
+
+# References
+[1] Burkardt, John. "The Truncated Normal Distribution" 
+[PDF](https://people.sc.fsu.edu/~jburkardt/presentations/truncated_normal.pdf). 
+Department of Scientific Computing website.
+"""
+function truncated_normal(rng::AbstractRNG, dims...; mean = 0, std = 1, lo = -2, hi = 2)
+  norm_cdf(x) = 0.5 * (1 + erf(x/√2))
+  if (mean < lo - 2 * std) || (mean > hi + 2 * std)
+    @warn "Mean is more than 2 std from [a, b] in truncated_normal. The distribution of values may be incorrect." maxlog=1
```
mcabbott, line 219:
```suggestion
    @warn "Mean is more than 2 std outside the limits in truncated_normal, so the distribution of values may be inaccurate." maxlog=1
```

src/utils.jl
```diff
@@ -451,7 +493,7 @@ The second return value `re` allows you to reconstruct the original network afte
 modifications to the weight vector (for example, with a hypernetwork).
 
     julia> re(θ .* 2)
-    Chain(Dense(10, 5, σ), Dense(5, 2), softmax)
+    Chain(Dense(10, 5, std), Dense(5, 2), softmax)
```
mcabbott, line 496: Here & two others:
```suggestion
    Chain(Dense(10, 5, σ), Dense(5, 2), softmax)
```

src/utils.jl
```diff
@@ -188,9 +188,50 @@ function kaiming_normal(rng::AbstractRNG, dims...; gain = √2f0)
...
+kaiming_normal(dims...; kwargs...) = kaiming_normal(rng_from_array(), dims...; kwargs...)
 kaiming_normal(rng::AbstractRNG; init_kwargs...) = (dims...; kwargs...) -> kaiming_normal(rng, dims...; init_kwargs..., kwargs...)
 
+"""
+    truncated_normal([rng=GLOBAL_RNG], dims...; mean = 0, std = 1, lo = -2., hi = 2.)
+  
+Return an `Array` of size `dims` where each element is drawn from a truncated normal distribution.
+The values are generated by using a truncated uniform distribution and then using the inverse CDF 
+for the normal distribution. The method used for generating the random values works best when 
+`lo ≤ mean ≤ hi`.
```
mcabbott, line 200: Maybe this should be two paragraphs, one on what you get, the second on the algorithm used. And  perhaps the first ought to say that mean/std belongs to the un-truncated gaussian, they are not moments of what's returned.

Also IMO we should never write `2.`, it's either important enough to be `2.0` or `2f0`, or it's just two.

```suggestion
    truncated_normal([rng=GLOBAL_RNG], dims...; mean = 0, std = 1, lo = -2, hi = 2)
  
Return an `Array{Float32}` of size `dims` where each element is drawn from a truncated normal distribution.
The numbers are distributed like `filter(x -> lo<=x<=hi, mean .+ std .* rand(10^6))`.

The values are generated using uniform `rand()` and then the inverse CDF. 
This method works best when `lo ≤ mean ≤ hi`.
```

theabhirath:
> the error is under control too for now

Welp. CI shows there's an error because 3.98 is calculated instead of 4 (tolerance for now is 0.01). Should I increase the tolerance? And also, doctests keep failing - I'm not sure why, but the values it gets seem to be different from mine. I'm on Julia 1.8 on a Mac if that matters

mcabbott:
I don't think random numbers are stable between Julia versions. They may only be tested on one version, but better, most of the doctests are written never to print them, always show size of `summary` etc instead.

src/utils.jl
theabhirath:
```julia
julia> histogram(reduce(vcat, truncated_normal(100, 100)))
                ┌                                        ┐ 
   [-2.0, -1.5) ┤███████▍ 448                              
   [-1.5, -1.0) ┤████████████████▎ 982                     
   [-1.0, -0.5) ┤█████████████████████████▍ 1538           
   [-0.5,  0.0) ┤████████████████████████████████▌ 1974    
   [ 0.0,  0.5) ┤██████████████████████████████████  2066  
   [ 0.5,  1.0) ┤█████████████████████████▋ 1556           
   [ 1.0,  1.5) ┤███████████████▋ 952                      
   [ 1.5,  2.0) ┤███████▉ 484                              
                └                                        ┘ 
                                 Frequency            
```
This looks pretty nice to me

theabhirath:
> I don't think random numbers are stable between Julia versions. They may only be tested on one version, but better, most of the doctests are written never to print them, always show size of `summary` etc instead.

Should I fix this for the other init layers as well? And change them to have plots too?

src/utils.jl
```diff
@@ -39,7 +39,7 @@ epseltype(x) = eps(float(eltype(x)))
 Create an instance of the RNG most appropriate for `x`.
 The current defaults are:
 - `x isa AbstractArray`
-  - Julia version is < 1.7: `Random.GLOBAL_RNG`
+  - Julia version is < 1.7: `rng_from_array()`
```
darsnack, line 42: This change shouldn't be here (the old docstring was correct).

src/utils.jl
```diff
@@ -188,9 +188,51 @@ function kaiming_normal(rng::AbstractRNG, dims...; gain = √2f0)
...
 
-kaiming_normal(dims...; kwargs...) = kaiming_normal(Random.GLOBAL_RNG, dims...; kwargs...)
+kaiming_normal(dims...; kwargs...) = kaiming_normal(rng_from_array(), dims...; kwargs...)
 kaiming_normal(rng::AbstractRNG; init_kwargs...) = (dims...; kwargs...) -> kaiming_normal(rng, dims...; init_kwargs..., kwargs...)
 
+"""
+    truncated_normal([rng=GLOBAL_RNG], dims...; mean = 0, std = 1, lo = -2, hi = 2)
+  
+Return an `Array{Float32}` of size `dims` where each element is drawn from a truncated normal distribution.
+The numbers are distributed like `filter(x -> lo<=x<=hi, mean .+ std .* rand(10^6))`.
```
darsnack, line 198: Shouldn't this say `mean .+ std .* randn(dims)`?

src/utils.jl
```diff
@@ -188,9 +188,51 @@ function kaiming_normal(rng::AbstractRNG, dims...; gain = √2f0)
...
+
+The values are generated using uniform `rand()` and then the inverse CDF.
```
darsnack, line 200:
```suggestion
The values are generated by sampling a Uniform(0, 1) (`rand()`) and then
applying the inverse CDF of the truncated normal distribution
(see the references for more info).
```

src/utils.jl
theabhirath: Works. Given that this is something that needs a little more hashing out (as noted [here](https://github.com/FluxML/Flux.jl/pull/1877#issuecomment-1045596625)), I think I'll try and open up a separate PR addressing these concerns for all the init strategies

src/utils.jl
darsnack: Or is this trying to say the distribution after sampling a million times? Shouldn't it still be normal?

theabhirath: The number of elements _is_ the number of times it's sampled, right?

darsnack: Yeah that's the number of samples from the truncated normal. I was thinking maybe this was trying to approximately describe the resampling procedure somehow (so > `dims` samples from the underlying normal). But I typed that out before seeing your comment that it's just an example, so you can ignore it.

src/utils.jl
```diff
@@ -188,9 +188,56 @@ function kaiming_normal(rng::AbstractRNG, dims...; gain = √2f0)
...
+Return an `Array{Float32}` of size `dims` where each element is drawn from a truncated normal distribution.
+The numbers are distributed like `filter(x -> lo<=x<=hi, mean .+ std .* randn(dims...))`.
+
+The values are generated by sampling a Uniform(0, 1) (`rand()`) and then
+applying the inverse CDF of the truncated normal distribution
+(see the references for more info).
+This method works best when `lo ≤ mean ≤ hi`.
+
+# Examples
+```jldoctest; setup = :(using Statistics)
```
darsnack, line 206:
```suggestion
```jldoctest
julia> using Statistics

```

NEWS.md
```diff
@@ -9,6 +9,7 @@ been removed in favour of MLDatasets.jl.
 * `Dropout` gained improved compatibility with Int and Complex arrays and is now twice-differentiable.
 * Many utily functions and the `DataLoader` are [now provided by MLUtils.jl](https://github.com/FluxML/Flux.jl/pull/1874).
 * The DataLoader is now compatible with generic dataset types implementing `MLUtils.numobs` and `MLUtils.getobs`.
+* Added truncated normal initialisation of weights.
```
darsnack, line 12:
```suggestion
* Added [truncated normal initialisation](https://github.com/FluxML/Flux.jl/pull/1877) of weights.
```

darsnack (approved):

Issue #1765, opened 2021-11-12
Figure out consistent code style


ToucheSir:
I've seen a few of files now in FluxML repos where the diffs ping-pong back and forth between different styles. Most of the discrepancy seems to be around function and type signature whitespace. Trailing whitespace too, my editor config is set to trim them and a surprising number of core files seem to have them.

ChainRulesCore recently [added JuliaFormatter](https://github.com/JuliaDiff/ChainRulesCore.jl/pull/482) to their CI and it seems to be working well, but even just a couple of sentences in CONTRIBUTING.md would go a long way.

DhairyaLGandhi:
We can add a line to contributing.md, but we follow two spaces and no trailing whitespace. Additionally we use linux file endings. We use whitespace around types and around default arguments too (`kwarg = value`).

ToucheSir:
Well that's just the thing, the current codebase is anything but self-consistent! Just going through the top level of `src/`, https://github.com/FluxML/Flux.jl/blob/master/src/deprecations.jl has no spaces while https://github.com/FluxML/Flux.jl/blob/master/src/functor.jl does. A big advantage of auto-formatting is that it's hard to argue with a machine :smile:, but regardless of whether we add one there should be some sort of rough consensus on this style question.

DhairyaLGandhi:
Yeah deprecations.jl is not following the style and should be fixed.

Chandu-4444:
I'd like to help with this issue. Can you tell me what can I do?

ToucheSir:
Well, the first step is to get most contributors to agree on a convention :P. You could try doing something like the ChainRulesCore PR linked above and see what people say.

darsnack:
I like four spaces because it matches docstrings/markdown, but I'm happy with Dhairya's suggestion.

theabhirath:
Now that there is a [consistent code style for Metalhead](https://github.com/FluxML/Metalhead.jl/pull/168), could this be extended to the Flux repo as well?

ToucheSir:
That would be great. Something like https://github.com/JuliaDiff/ChainRulesCore.jl/blob/main/.github/workflows/format.yml would be nice to have too. We can trial this on a smaller FluxML repo if Flux turns out to hit too many formatting edge cases.

theabhirath:
I was about to set that up for Metalhead but decided to hold off until https://github.com/domluna/JuliaFormatter.jl/issues/604 is resolved

[...]

PR #1881, opened 2022-02-19
Extending `Diagonal`

theabhirath:
This PR tries to extend some of the functionality for the built-in layer `Diagonal` following the discussions in https://github.com/FluxML/Metalhead.jl/pull/119. This makes very minimal changes to Diagonal to allow it to function without a bias and also take in a weight initialisation strategy for the weight matrix. This has the advantage that:

1. It allows for more flexibility to implement custom layers like LayerScale; and
2. It leaves the current working unchanged (what worked before still will without any modifications)

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -185,7 +185,8 @@ if `bias` is true, and
 
 otherwise. The learnable arrays are initialised `α = ones(Float32, size)` and
 `β = zeros(Float32, size)`. If `init` is specified, the function given to it is 
-called and used to initialise α.
+called and used to initialise α. The weight matrix and/or the bias vector 
+(with the same size as x) may also be provided explicitly.
 
 Used by [`LayerNorm`](@ref).
 """
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -110,7 +110,7 @@ as an `in × N` matrix, or any array with `size(x,1) == in`.
 The out `y` will be a vector  of length `out`, or a batch with
 `size(y) == (out, size(x)[2:end]...)`
 
-Keyword `bias = false` will switch off trainable bias for the layer.
+Keyword `bias=false` will switch off trainable bias for the layer.
 The initialisation of the weight matrix is `W = init(out, in)`, calling the function
 given to keyword `init`, with default [`glorot_uniform`](@doc Flux.glorot_uniform).
 The weight matrix and/or the bias vector (of length `out`) may also be provided explicitly.
@@ -127,7 +127,7 @@ julia> d(rand(Float32, 5, 1, 1, 64)) |> size  # treated as three batch dimension
 (2, 1, 1, 64)
 
 julia> d1 = Dense(ones(2, 5), false, tanh)  # using provided weight matrix
-Dense(5 => 2, tanh; bias = false)  # 10 parameters
+Dense(5 => 2, tanh; bias=false)  # 10 parameters
 
 julia> d1(ones(5))
 2-element Vector{Float64}:
@@ -167,7 +167,7 @@ end
 function Base.show(io::IO, l::Dense)
   print(io, "Dense(", size(l.weight, 2), " => ", size(l.weight, 1))
   l.σ == identity || print(io, ", ", l.σ)
-  l.bias == false && print(io, "; bias = false")
+  l.bias == false && print(io, "; bias=false")
   print(io, ")")
 end
 
@@ -182,11 +182,13 @@ Create an element-wise linear layer, which performs
 if `bias` is true, and
 
     y = x .* scale
+    
+otherwise.
 
-otherwise. The learnable arrays are initialised `scale = ones(Float32, size)` and
-`bias = zeros(Float32, size)`. If `init` is specified, the function given to it is 
-called and used to initialise scale. The weight matrix and/or the bias vector 
-(with the same size as x) may also be provided explicitly.
+The learnable parameters are initialised `scale = Flux.ones32(size)` and 
+`bias = Flux.zeros32(size)`. Alternatively, specify `init` to customize how `scale` 
+is initialised. The weight matrix and/or the bias vector (with the same size as `x`) 
+may also be provided explicitly.
 
 Used by [`LayerNorm`](@ref).
 """
@@ -207,7 +209,7 @@ Diagonal(sz::Integer...; bias = true, init = ones32) = Diagonal(init(sz...), bia
 
 function Base.show(io::IO, l::Diagonal)
   print(io, "Diagonal(", join(size(l.scale), ", "))
-  l.bias == false && print(io, "; bias = false")
+  l.bias == false && print(io, "; bias=false")
   print(io, ")")
 end
 
@@ -330,7 +332,7 @@ which is accepted as the input to a `Chain`.
 If the two input sizes are the same, `in1 == in2`, then you may write `Bilinear(in => out, σ)`.
 
 The initialisation works as for [`Dense`](@ref) layer, with `W = init(out, in1, in2)`.
-By default the bias vector is `zeros(Float32, out)`, option `bias = false` will switch off
+By default the bias vector is `zeros(Float32, out)`, option `bias=false` will switch off
 trainable bias. Either of these may be provided explicitly.
 
 # Examples
@@ -348,14 +350,14 @@ true
 
 julia> sc = SkipConnection(
                 Chain(Dense(5 => 20, tanh), Dense(20 => 9, tanh)),
-                Flux.Bilinear((9, 5) => 3, bias = false),
+                Flux.Bilinear((9, 5) => 3, bias=false),
             );  # used as the recombinator, with skip as the second input
 
 julia> sc(x) |> size
 (3, 32)
 
 julia> Flux.Bilinear(rand(4,8,16), false, tanh)  # first dim of weight is the output
-Bilinear((8, 16) => 4, tanh; bias = false)  # 512 parameters
+Bilinear((8, 16) => 4, tanh; bias=false)  # 512 parameters
 ```
 """
 struct Bilinear{F,A,B}
@@ -406,7 +408,7 @@ function Base.show(io::IO, l::Bilinear)
     print(io, "Bilinear((", size(l.weight, 2), ", ", size(l.weight, 3), ") => ", size(l.weight, 1))
   end
   l.σ == identity || print(io, ", ", l.σ)
-  l.bias === false && print(io, "; bias = false")
+  l.bias === false && print(io, "; bias=false")
   print(io, ")")
 end
 
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -172,23 +172,18 @@ function Base.show(io::IO, l::Dense)
 end
 
 """
-    Diagonal(scale, bias)
     Diagonal(size::Integer...; bias = true, init = ones32)
+    Diagonal(scale::AbstractArray, [bias])
 
 Create an element-wise linear layer, which performs
 
     y = scale .* x .+ bias
 
-if `bias` is true, and
-
-    y = x .* scale
-    
-otherwise.
-
-The learnable parameters are initialised `scale = Flux.ones32(size)` and 
-`bias = Flux.zeros32(size)`. Alternatively, specify `init` to customize how `scale` 
-is initialised. The weight matrix and/or the bias vector (with the same size as `x`) 
-may also be provided explicitly.
+with no activation function.
+ 
+The learnable parameters are initialised `scale = init(size...)` and 
+`bias = zeros32(size...)`, with `init = ones32` by default. You may specify the function `init`, 
+turn off trainable bias with `bias = false`, or provide the array(s) explicitly.
 
 Used by [`LayerNorm`](@ref).
 """
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -172,7 +172,7 @@ function Base.show(io::IO, l::Dense)
 end
 
 """
-    Diagonal(size::Integer...; bias = true, init = ones32)
+    Diagonal(size::Integer...; bias=true, init=ones32)
     Diagonal(scale::AbstractArray, [bias])
 
 Create an element-wise linear layer, which performs
@@ -181,9 +181,9 @@ Create an element-wise linear layer, which performs
 
 with no activation function.
  
-The learnable parameters are initialised `scale = init(size...)` and 
-`bias = zeros32(size...)`, with `init = ones32` by default. You may specify the function `init`, 
-turn off trainable bias with `bias = false`, or provide the array(s) explicitly.
+The learnable scale & bias are initialised `init(size...)` and `zeros32(size...)`,
+with `init=ones32` by default. You may specify the function `init`, 
+turn off trainable bias with `bias=false`, or provide the array(s) explicitly.
 
 Used by [`LayerNorm`](@ref).
 """
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -110,7 +110,7 @@ as an `in × N` matrix, or any array with `size(x,1) == in`.
 The out `y` will be a vector  of length `out`, or a batch with
 `size(y) == (out, size(x)[2:end]...)`
 
-Keyword `bias=false` will switch off trainable bias for the layer.
+Keyword `bias = false` will switch off trainable bias for the layer.
 The initialisation of the weight matrix is `W = init(out, in)`, calling the function
 given to keyword `init`, with default [`glorot_uniform`](@doc Flux.glorot_uniform).
 The weight matrix and/or the bias vector (of length `out`) may also be provided explicitly.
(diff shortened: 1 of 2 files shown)


theabhirath:
PS: a weird error occurs when I write the forward pass as `(a::Diagonal)(x) = a.α .* x .+ a.β` (the implementation in the PR has two lines for this reason):

```julia
ERROR: MethodError: no method matching length(::Flux.Zeros)
Closest candidates are:
  length(::Union{Base.KeySet, Base.ValueIterator}) at ~/julia/usr/share/julia/base/abstractdict.jl:58
  length(::Union{LinearAlgebra.Adjoint{T, <:Union{StaticArrays.StaticVector{<:Any, T}, StaticArrays.StaticMatrix{<:Any, <:Any, T}}}, LinearAlgebra.Diagonal{T, <:StaticArrays.StaticVector{<:Any, T}}, LinearAlgebra.Hermitian{T, <:StaticArrays.StaticMatrix{<:Any, <:Any, T}}, LinearAlgebra.LowerTriangular{T, <:StaticArrays.StaticMatrix{<:Any, <:Any, T}}, LinearAlgebra.Symmetric{T, <:StaticArrays.StaticMatrix{<:Any, <:Any, T}}, LinearAlgebra.Transpose{T, <:Union{StaticArrays.StaticVector{<:Any, T}, StaticArrays.StaticMatrix{<:Any, <:Any, T}}}, LinearAlgebra.UnitLowerTriangular{T, <:StaticArrays.StaticMatrix{<:Any, <:Any, T}}, LinearAlgebra.UnitUpperTriangular{T, <:StaticArrays.StaticMatrix{<:Any, <:Any, T}}, LinearAlgebra.UpperTriangular{T, <:StaticArrays.StaticMatrix{<:Any, <:Any, T}}, StaticArrays.StaticVector{<:Any, T}, StaticArrays.StaticMatrix{<:Any, <:Any, T}, StaticArrays.StaticArray{<:Tuple, T}} where T) at ~/.julia/packages/StaticArrays/h2TVC/src/abstractarray.jl:1
  length(::Union{LinearAlgebra.Adjoint{T, S}, LinearAlgebra.Transpose{T, S}} where {T, S}) at ~/julia/usr/share/julia/stdlib/v1.8/LinearAlgebra/src/adjtrans.jl:172
  ...
Stacktrace:
 [1] _similar_shape(itr::Flux.Zeros, #unused#::Base.HasLength)
   @ Base ./array.jl:660
 [2] _collect(cont::UnitRange{Int64}, itr::Flux.Zeros, #unused#::Base.HasEltype, isz::Base.HasLength)
   @ Base ./array.jl:715
 [3] collect(itr::Flux.Zeros)
   @ Base ./array.jl:709
 [4] broadcastable(x::Flux.Zeros)
   @ Base.Broadcast ./broadcast.jl:704
 [5] broadcasted
   @ ./broadcast.jl:1302 [inlined]
 [6] (::Flux.Diagonal{Vector{Float32}, Flux.Zeros})(x::Vector{Float64})
   @ Flux ~/Code/Flux.jl/src/layers/basic.jl:207
 [7] top-level scope
   @ REPL[2]:1
```

darsnack:
Probably need to define
```julia
Base.broadcastable(::Zeros) = Ref(Zeros())
```

theabhirath:
> Probably need to define this in `zeros.jl`.
> 
> ```julia
> Base.broadcastable(::Zeros) = Ref(Zeros())
> ```

This didn't help, the error message changed to:

```julia
ERROR: MethodError: no method matching +(::Float64, ::Flux.Zeros)
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at ~/julia/usr/share/julia/base/operators.jl:591
  +(::T, ::T) where T<:Union{Float16, Float32, Float64} at ~/julia/usr/share/julia/base/float.jl:377
  +(::Union{Float16, Float32, Float64}, ::BigFloat) at ~/julia/usr/share/julia/base/mpfr.jl:407
  ...
Stacktrace:
 [1] _broadcast_getindex_evalf
   @ ./broadcast.jl:670 [inlined]
 [2] _broadcast_getindex
   @ ./broadcast.jl:643 [inlined]
 [3] getindex
   @ ./broadcast.jl:597 [inlined]
 [4] copy
   @ ./broadcast.jl:899 [inlined]
 [5] materialize
   @ ./broadcast.jl:860 [inlined]
 [6] (::Flux.Diagonal{Array{Float32, 3}, Flux.Zeros})(x::Array{Float64, 3})
   @ Flux ~/Code/Flux.jl/src/layers/basic.jl:207
 [7] top-level scope
   @ REPL[2]:1
```

There's also broadcasting rules defined already, and astonishingly an _extremely_ similar definition works for the Dense layer 🤔

src/layers/basic.jl
```diff
@@ -174,27 +174,40 @@ end
...
+(with the same size as x) may also be provided explicitly.
 
 Used by [`LayerNorm`](@ref).
 """
-struct Diagonal{T}
-  α::T
-  β::T
+struct Diagonal{A, B}
+  α::A
+  β::B
```
mcabbott, line 196: Can I suggest that we change the field names while we're at it? If all bias fields are called `bias`, that will occasionally simplify life:
```suggestion
  scale::A
  bias::B
```

src/layers/basic.jl
```diff
@@ -174,27 +174,40 @@ end
...
 @functor Diagonal
 
-(a::Diagonal)(x) = a.α .* x .+ a.β
+function (a::Diagonal)(x)
+  x = a.α .* x
+  x = x .+ a.β
+end
 
 function Base.show(io::IO, l::Diagonal)
   print(io, "Diagonal(", join(size(l.α), ", "), ")")
```
mcabbott, line 213: Also, copying Dense:
```suggestion
function Base.show(io::IO, l::Diagonal)
  print(io, "Diagonal(", join(size(l.α), ", "))
  l.bias == Zeros() && print(io, "; bias=false")
  print(io, ")")
```

src/layers/basic.jl
```diff
@@ -174,27 +174,40 @@ end
...
-Diagonal(sz::Integer...) = Diagonal(ones32(sz...), zeros32(sz...))
+Diagonal(sz::Integer...; bias = true, init = ones32) = Diagonal(init(sz...), bias)
 
 @functor Diagonal
 
-(a::Diagonal)(x) = a.α .* x .+ a.β
+function (a::Diagonal)(x)
+  x = a.α .* x
+  x = x .+ a.β
+end
```
mcabbott, line 210: What's the purpose of this?

I think it ought to stay one fused broadcast. And that it should not re-use `x` to mean 3 different things.

theabhirath: https://github.com/FluxML/Flux.jl/pull/1881#issuecomment-1046010390 😅 It's a hack for now until I figure out what's going on

darsnack:
> There's also broadcasting rules defined already, and astonishingly an extremely similar definition works for the Dense layer

Yes, that's cause `Dense` only has one level of broadcasting. `Zeros` probably never worked for any complex nested broadcasting. This basically leaves us with making `Zeros` a full blown `AbstractArray` with dimensions or defining scalar rules as well.

Alternatively, we can consider replacing `Zeros` with FillArrays. What do you think @mcabbott?

src/layers/basic.jl
mcabbott: Oh, sorry, I missed that.

mcabbott:
We can also just delete Zeros completely.

darsnack:
Are you up for reviving your efforts in that direction? I'm all for it.

src/layers/basic.jl
mcabbott: I see, the error is from making the scale a scalar:
```
julia> [1,2,3] .+ Flux.Zeros()
3-element Vector{Int64}:
 1
 2
 3

julia> 2 .+ Flux.Zeros()
ERROR: MethodError: no method matching length(::Flux.Zeros)
```

darsnack: I think it's also when it gets nested:
```julia
julia> [1, 2, 3] .* [5, 6, 7] .+ Flux.Zeros()
ERROR: MethodError: no method matching length(::Flux.Zeros)

julia> [1, 2, 3] .* [5 6; 8 9; 10 11] .+ Flux.Zeros()
ERROR: MethodError: no method matching length(::Flux.Zeros)
```

src/layers/basic.jl
```diff
@@ -174,27 +174,40 @@ end
...
+`β = zeros(Float32, size)`. If `init` is specified, the function given to it is 
+called and used to initialise α. The weight matrix and/or the bias vector 
+(with the same size as x) may also be provided explicitly.
 
 Used by [`LayerNorm`](@ref).
 """
-struct Diagonal{T}
-  α::T
-  β::T
+struct Diagonal{A, B}
```
mcabbott, line 194: In fact, one more suggestion:
```suggestion
struct Diagonal{A<:AbstractArray, B}
```
I think we should either forbid, or explicitly allow,document, and test, the use of a number for the scale. 

Right now you can create such a thing, but there are no tests. And is not handled by `create_bias`:
```
julia> Flux.Diagonal(1.2, 3.4)
Diagonal()

julia> Flux.create_bias(1.2, true, 2,2)
ERROR: MethodError: no method matching create_bias(::Float64, ::Bool, ::Int64, ::Int64)
```

src/layers/basic.jl
darsnack: The issue seems to be the `Zeros` is written with rules to behave like an `AbstractArray`, but it is actually a scalar by [broadcasting's definition](https://docs.julialang.org/en/v1.7/manual/interfaces/#man-interfaces-broadcasting).

mcabbott: I see. I guess Dense does something subtly different:
```
julia> identity.([5, 6, 7] .+ Flux.Zeros())
3-element Vector{Int64}:
 5
 6
 7
```
The error is because the `.*` makes a Broadcasted not an array, I think, so it misses the methods defined for the weird Zero thing. One more reason to kill it.

darsnack: Exactly, there are probably many cases where `Zeros` plain won't behave like it is supposed to.

src/layers/basic.jl
mcabbott: In fact the printing lies, too. What `Diagonal()` creates is a zero-dimensional array. 
```
julia> Flux.Diagonal() |> dump
Flux.Diagonal{Array{Float32, 0}}
  α: Array{Float32}(()) fill(1.0f0)
  β: Array{Float32}(()) fill(0.0f0)
```
Which should be fine, it's one scalar parameter. Although probably nothing is tested with this.

theabhirath:
I'm a little uncertain - I'm happy to incorporate the changes but given that `create_bias` and `Zeros` seem to be a point of contention, I think I'll let this stew for a bit until there's a clear direction regarding that?

src/layers/basic.jl
mcabbott: Ok, I made #1882 to kill this thing.

mcabbott: Done, Zeros is now gone (on master).

src/layers/basic.jl
```diff
@@ -110,7 +110,7 @@ as an `in × N` matrix, or any array with `size(x,1) == in`.
 The out `y` will be a vector  of length `out`, or a batch with
 `size(y) == (out, size(x)[2:end]...)`
 
-Keyword `bias=false` will switch off trainable bias for the layer.
+Keyword `bias = false` will switch off trainable bias for the layer.
```
darsnack, line 113: Annoyingly, these should be reverted b/c the Flux style.

src/layers/basic.jl
```diff
@@ -167,36 +167,48 @@ end
...
-The learnable arrays are initialised `α = ones(Float32, size)` and
-`β = zeros(Float32, size)`.
+if `bias` is true, and
+
+    y = x .* scale
+
+otherwise. The learnable arrays are initialised `scale = ones(Float32, size)` and
+`bias = zeros(Float32, size)`. If `init` is specified, the function given to it is 
+called and used to initialise scale. The weight matrix and/or the bias vector 
+(with the same size as x) may also be provided explicitly.
```
darsnack, line 189:
```suggestion
otherwise.

The learnable parameters are initialised `scale = Flux.ones32(size)` and `bias = Flux.zeros32(size)`.
Alternatively, specify `init` to customize how `scale` is initialised.
The weight matrix and/or the bias vector (with the same size as `x`) may also be provided explicitly.
```

test/layers/basic.jl
```diff
@@ -91,16 +91,17 @@ import Flux: activations
     @test length(Flux.Diagonal(10)(randn(10))) == 10
     @test length(Flux.Diagonal(10)(1)) == 10
     @test length(Flux.Diagonal(10)(randn(1))) == 10
+    @test length(Flux.Diagonal(10; bias = false)(randn(10))) == 10
     @test_throws DimensionMismatch Flux.Diagonal(10)(randn(2))
 
     @test Flux.Diagonal(2)([1 2]) == [1 2; 1 2]
     @test Flux.Diagonal(2)([1,2]) == [1,2]
-    @test Flux.Diagonal(2)([1 2; 3 4]) == [1 2; 3 4]
+    @test Flux.Diagonal(2; bias = false)([1 2; 3 4]) == [1 2; 3 4]
```
darsnack, line 99: Why does these tests need `bias=false`?

darsnack (changes requested):

test/layers/basic.jl
theabhirath: They don't really, I kept those just to check that `bias=false` doesn't trip anything

src/layers/basic.jl
```diff
@@ -172,31 +172,45 @@ function Base.show(io::IO, l::Dense)
...
+if `bias` is true, and
+
+    y = x .* scale
+    
+otherwise.
+
+The learnable parameters are initialised `scale = Flux.ones32(size)` and 
+`bias = Flux.zeros32(size)`. Alternatively, specify `init` to customize how `scale` 
+is initialised. The weight matrix and/or the bias vector (with the same size as `x`) 
+may also be provided explicitly.
```
mcabbott, line 191: `y = scale .* x .+ bias` is true in both cases, but it seems a little confusing to say that this is true if `bias=true`, since the `true` isn't what goes into that formula. Perhaps we can be quite brief since Dense works the same way:

```suggestion
with no activation function.
 
The learnable parameters are initialised `scale = init(size...)` and 
`bias = zeros32(size...)`, with `init = ones32` by default. You may specify the function `init`, 
turn off trainable bias with `bias = false`, or provide the array(s) explicitly.
```

src/layers/basic.jl
```diff
@@ -172,31 +172,45 @@ function Base.show(io::IO, l::Dense)
 end
 
 """
-    Diagonal(α, β)
-    Diagonal(size::Integer...)
+    Diagonal(scale, bias)
+    Diagonal(size::Integer...; bias = true, init = ones32)
```
mcabbott, line 176: To match Dense etc. I'd put the method which takes integers first:
```suggestion
    Diagonal(size::Integer...; bias = true, init = ones32)
    Diagonal(scale::AbstractArray, [bias])
```

src/layers/basic.jl
```diff
@@ -172,31 +172,40 @@ function Base.show(io::IO, l::Dense)
 end
 
 """
-    Diagonal(α, β)
-    Diagonal(size::Integer...)
+    Diagonal(size::Integer...; bias = true, init = ones32)
```
mcabbott, line 175:
```suggestion
    Diagonal(size::Integer...; bias=true, init=ones32)
```

src/layers/basic.jl
```diff
@@ -172,31 +172,40 @@ function Base.show(io::IO, l::Dense)
...
-    y = α .* x .+ β
+    y = scale .* x .+ bias
 
-The learnable arrays are initialised `α = ones(Float32, size)` and
-`β = zeros(Float32, size)`.
+with no activation function.
+ 
+The learnable parameters are initialised `scale = init(size...)` and 
+`bias = zeros32(size...)`, with `init = ones32` by default. You may specify the function `init`, 
+turn off trainable bias with `bias = false`, or provide the array(s) explicitly.
```
mcabbott, line 186:
```suggestion
The learnable scale & bias are initialised `init(size...)` and `zeros32(size...)`,
with `init=ones32` by default. You may specify the function `init`, 
turn off trainable bias with `bias=false`, or provide the array(s) explicitly.
```

mcabbott (approved):
This looks fine to me.

There's a style question about whether we want spaces around keywords. Mostly Flux don't have such in docstrings. I am not sure whether Kyle's comment https://github.com/FluxML/Flux.jl/pull/1881#r820132004 is suggesting to remove them? Which is what my suggestions below would do.

ToucheSir:
Per https://github.com/FluxML/Flux.jl/issues/1765 I don't think that style question is settled, and outside of installing a autoformatter to enforce it don't personally have the energy to try to litigate it.

mcabbott:
Yes, I don't mean to litigate it. This may not have been what Kyle meant above. Approved by me with or without.

darsnack:
Yeah no spaces is what I meant, but I would be very happy with a decision to litigating this stuff! Approved by me then.

mcabbott:
Ok, done. Can you approve though, as your red card still blocks my button? Or just merge, in fact.

darsnack (approved):
Whoops my bad

darsnack:
Also, I meant above to say I would be happy _not_ to litigate 😅

PR #1882, opened 2022-02-19
rm Flux.Zeros

mcabbott:
This is the same change as #1407, but easier to replace than to rebase.

This year's motivation for killing Flux.Zeros is that its broadcasting rules are broken for what's needed in #1881. Last year, and the one before... well, there's a long history. The extent of this thing was reduced in #1379.

Closes #1407. Fixes #1392. Fixes  #1402. Does not re-break #1332.


src/deprecations.jl
```diff
@@ -17,3 +17,12 @@ zeros32(::Type, dims...) = throw(ArgumentError("Flux.zeros32 is always Float32,
 
 # v0.13 deprecations
 @deprecate frequencies(xs) group_counts(xs)
+
+struct Zeros # was used both Dense(10, 2, initb = Zeros) and Dense(rand(2,10), Zeros()), long ago
+   function Zeros()
+     Base.depwarn("Flux.Zeros is no more, has ceased to be, is bereft of life, is an ex-boondoggle... please simply use bias=false instead", :Zeros)
```
darsnack, line 23: Cute, I definitely laughed, but a shorter warning would be better?

test/utils.jl
```diff
@@ -407,16 +351,16 @@ end
...
-    @testset "$b1 to $b2" for (b1, b2, be) in (
-      (Flux.zeros32, Flux.ones32, Flux.ones32),   # Load ones as bias to a model with zeros as bias -> model gets ones as bias
-      (Flux.ones32, nobias, Flux.zeros32), # Load Zeros as bias to a model with ones as bias-> model gets zeros as bias
-      (nobias, Flux.ones32, nobias),     # Load ones as bias to a model with Zeros as bias-> model bias does not change
-    )
-      m1 = dm(b1)
-      m2 = dm(b2)
-      loadparams!(m1, b1 == nobias ? weights(m2) : pararray(m2))
-      testdense(m1, be)
-    end
```
mcabbott, line 419: These are the tests where I'm not quite sure what's being tested or what the intent is. Does anyone know?

darsnack: I think the intent is if the model already has the `bias` as `false` (`Zeros`), then loading a numeric bias into it should be ignored. The opposite direction is not true, loading a `false` (`Zeros`) into a numeric bias should zero-out the entries of the numeric bias without disabling it.

darsnack: These are some special cases that could easily be handled in #1875.

src/deprecations.jl
ToucheSir: Yes, leaving it as a comment would be fine though. Ideally with a mention of pining for the fjords.

mcabbott: I hope this will seldom be called, since the API has been to use `bias=false` for a long time. But I think there should be some depwarn, and (clearly) I don't think really needs to be dead-minimal and completely straight-faced.

test/utils.jl
mcabbott: Ok, will fiddle a bit. But possibly test these facts in a less awful way.

mcabbott: 8acad3e adds some tests. 

> loading a false (Zeros) into a numeric bias should zero-out the entries of the numeric bias without disabling it.

This does not happen on tagged flux, nor with this PR. I've marked it broken.

>  if the model already has the bias as false (Zeros), then loading a numeric bias into it should be ignored. 

This is the current behaviour, and is unchanged here. But there is some chance it ought to be an error, at least when the source vector isn't zero.

As you say, these can be adjusted as desired with structural `loadparams!` alla #1875.

src/utils.jl
```diff
@@ -441,17 +441,17 @@ rand32(dims...) = Base.rand(Float32, dims...)
...
-    create_bias(weights, bias, length)
+    create_bias(weights, bias, size...)
 
 Return a bias parameter for a layer, based on the value given
 to the constructor's keyword `bias=bias`.
 
 * `bias == true` creates a zero vector, of the same type as weights.
-* `bias == false` returns `Zeros()`, a special struct which exists only to encode the absence of bias.
+* `bias == false` returns `false` now, which is understood by AD to be non-differentiable.
 * `bias::AbstractArray` uses the array provided, provided it has the correct size and eltype. If the type is wrong, it will be converted.
```
mcabbott, line 451: This line is a lie, it does not at present fix eltypes. I think I'd prefer it to insist they match, is there a use for differing types which isn't just a mistake?

mcabbott: For now I edited the words to match the code.

test/utils.jl
DrChainsaw: Trying to make some amends for awful code :)

The tests are indeed testing some edge cases which one might want to error, .e.g loading params from a model without bias into one which has. The behaviour of `loadparams!` was a hot topic in the previous iteration of no bias (although I feel it might have been a proxy for some other more ideological issue) and I just tried to test what the code kind of tried to do.

I think I was a bit worried that testing with just a dense layer as the model would mask alot of errors just because the bias happened to be the last parameter in the list of parameters (and therefore easily ignored by `zip`). Using a `Chain` with multiple dense makes the tests a bit more verbose but I fully admit that I went overboard with helper functions (and "succinct" naming) here.

Furthermore, the test in https://github.com/FluxML/Flux.jl/commit/8acad3e074a40522c30594a54e80d8b3e08cb5ea which is marked as broken fails only because Params does not collect the bias, so I guess one can just replace it with e.g. `[m0.weights, m0.bias]`.

I suppose this is all pretty moot if #1875 will anways change the behaviour of `loadparams!` (hopefully to make it more fail-fast).

mcabbott: Hah, well thanks for having a thick skin! Somehow I thought these tests were from the beginning of time... they certainly matched the surrounding code style perfectly. 

I've changed the test to use a Chain. Now all the mismatched cases are errors, before and after the PR. 

Because while `Params` claims to be an IdSet of parameters, in fact `loadparams!` relies on the order in vector of parameters. I have no idea whether that's a bug or a feature, the whole thing is hopelessly under-specified. Nevertheless, `Flux.params` is clearly the public interface here, and using that, mismatched models in `loadparams!` gave an error before this PR, and still give an error after it.

You can make it do other things by constructing this internal vector of parameters yourself. You can put numbers in there:
```
julia> false in Params([[1 2; 3 4], false])
true
```
which was what https://github.com/FluxML/Flux.jl/pull/1731 wanted to make an error, since they don't fit with the IdSet picture at all. No doubt you can construct vectors which make `loadparams!` do almost anything, or fail at anything. Are we obliged to preserve any of that?

DrChainsaw: > they certainly matched the surrounding code style perfectly.

Nice save 🙂 

> You can make it do other things by constructing this internal vector of parameters yourself.

Yeah, this is probably the key point w.r.t this issue. While one can put `false` in there, it has no more relevance to "no bias" than putting `13.44` and whether or not this is something that should work in `loadparams!` does not belong here (don't have a strong opinion fwiw).

mcabbott: Shall we do this? Maybe it's time.

src/utils.jl
```diff
@@ -441,17 +441,18 @@ rand32(dims...) = Base.rand(Float32, dims...)
...
+    create_bias(weights, bias, size...)
 
 Return a bias parameter for a layer, based on the value given
 to the constructor's keyword `bias=bias`.
 
-* `bias == true` creates a zero vector, of the same type as weights.
-* `bias == false` returns `Zeros()`, a special struct which exists only to encode the absence of bias.
-* `bias::AbstractArray` uses the array provided, provided it has the correct size and eltype. If the type is wrong, it will be converted.
+* `bias == true` creates a trainable array of the given size, of the same type as `weights`, initialised to zero.
+* `bias == false` returns `false` now, which is understood by AD to be non-differentiable.
```
CarloLucibello, line 450:
```suggestion
* `bias == false` returns `false`, which is understood by AD to be non-differentiable.
```

test/utils.jl
```diff
@@ -442,6 +375,26 @@ end
...
+  @test m1[1].bias == 7:9
+  @test sum(m1[1].weight) == 21
+
+  # load from a model without bias -- should ideally recognise the `false` but `Params` doesn't store it
+  @test_broken Flux.loadparams!(m1, Flux.params(m0))
+  @test_broken iszero(m1[1].bias)
+  @test sum(m1[1].weight) == 6  # written before error
+
+  # load into a model without bias -- should it ignore the parameter which has no home, or error?
+  @test_broken Flux.loadparams!(m0, Flux.params(m2))
```
CarloLucibello, line 393: I think being strict and erroring is fine. Also not worth thinking too much about this now as we phase out `Params`.

CarloLucibello (approved):

src/layers/basic.jl
```diff
@@ -301,7 +301,7 @@ end
     Bilinear((in1, in2) => out, σ=identity; bias=true, init=glorot_uniform)
     Bilinear(W::AbstractArray, [bias, σ])
 
-Creates a bilinear layer, which operates on two inputs at the same time.
+Creates a fully connected layer which operates on two inputs.
```
darsnack, line 304: I thought the old version was clearer

darsnack (approved):
Think it's better to deal with the `loadparams!` separately in  #1875.

src/layers/basic.jl
mcabbott: What I would like to do is to remove all docstrings which explain "X" by saying "This makes an X", because that seems circular.

Maybe this can be better without being self-referential?

darsnack: Ah, I see that makes sense. Maybe leave in the "at the same time." part?

mcabbott: Maybe like so, or perhaps too wordy?
```suggestion
Creates a fully connected layer which operates on two inputs at the same time,
with a learned weight for every triple of two input elements and an output element `x[i], y[j], z[k]`.

```

darsnack: Yeah might be too confusing. What about
```suggestion
Creates a fully connected layer which operates on two inputs at the same time,
with a separate learned weight for each input-output pair.
```

mcabbott: Maybe just:
```suggestion
Creates a layer which is fully connected between two inputs and the output, and otherwise like [`Dense`](@ref).
```
"fully connected" as an adjective, since as a noun "fully connected layer" it sort-of means Dense. And leave further description to the formula two lines down.

ToucheSir: I think the PyTorch description is pretty short and sweet: https://pytorch.org/docs/stable/generated/torch.nn.Bilinear.html. The current docstring takes 3-4 sentences to say basically the same thing.

mcabbott: That hides a lot though, as the formula is for a scalar output. But they have a stronger tradition of mysterious auto-broadcasting over extra dimensions (you are to infer that the extra dimension of A goes to the output, while those of x & y are zipped together) which their "shape" section unpacks a little. 

And I'm not sure whether saying "bilinear transformation" breaks the circular definition rule. Theirs doesn't include activation, but it does include bias, and thus isn't really linear in its inputs.

mcabbott: Anyway maybe this is good enough? Sorry about the tangent, didn't want to re-write the whole thing, just regard circularity as a bug to be fixed when noticed.

darsnack: Last suggestion is good for me

mcabbott: Done! Thanks all.

ToucheSir: I think the shape section is great and we ought to add it for all Flux layers instead of the ad-hoc current state of things, but that's a discussion for another issue. The main point is that we tend to spend a lot of prose on descriptions that could be captured with less, more structured doc text.

mcabbott: Yes that's probably true. And bad prose is worse than mediocre structured info.

One simple way to do this is to add `size` examples. If you pick distinct primes for the sizes then they can fix the result, without adding another layer of "what does this notation mean?" to things, and without the "maybe that's a typo?" worry as it's tested.

In the Pytorch one, what's the distinction between ``H_in1`` and `in1_feature` exactly, are those two things or the same thing? And they resort to prose for "All but the last dimension of the inputs should be the same." which I think is saying that the * and the N mean the same thing, there is no broadcasting. (But can the N be omitted entirely? Maybe not.)

PR #1884, opened 2022-02-20
Contrastive Loss function added

arcAman07:
Contrastive loss function added and exported in the Losses folder.
If it is perfect and without any errors, will write the corresponding test and update the docs of this loss function.
Wanted to keep this PR small so that loss function implemented is fine.

### DIFF
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -527,7 +527,7 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
     agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=dims))
 end
 
-function contrastive_loss(ŷ, y, margin = epseltype(ŷ), agg = mean)
+function contrastive_loss(ŷ, y; agg = mean, margin = epseltype(ŷ))
     square_pred = ŷ .^ 2
     max_val = max(sum(margin .- ŷ), 0)
     if max_val > 0
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -561,7 +561,6 @@ true
 See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
 
 """
-
 function contrastive_loss(ŷ, y; agg = mean, margin = epseltype(ŷ))
     square_pred = ŷ .^ 2
     max_val = max(sum(margin .- ŷ), 0)
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -563,7 +563,7 @@ See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
 """
 function contrastive_loss(ŷ, y; agg = mean, margin = epseltype(ŷ))
     square_pred = ŷ .^ 2
-    max_val = max(sum(margin .- ŷ), 0)
+    max_val = max(sum(ŷᵢ -> (margin - ŷᵢ), ŷ), 0)
     if max_val > 0
         margin_square = (margin .- ŷ) .^ 2
     else
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -537,7 +537,8 @@ function contrastive_loss(ŷ, y; agg = mean, margin = epseltype(ŷ))
     end
     return agg(y .* square_pred + (1 .- y) .* margin_square)
 end
-"""
+
+
 
 ```@meta
 DocTestFilters = nothing
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -527,9 +527,9 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
     agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=dims))
 end
 
-function contrastive_loss(ŷ, y, margin = epseltype(ŷ),agg=mean)
+function contrastive_loss(ŷ, y, margin = epseltype(ŷ), agg = mean)
     square_pred = ŷ .^ 2
-    max_val = max(sum(margin.-ŷ),0)
+    max_val = max(sum(margin .- ŷ), 0)
     if max_val > 0
         margin_square = (margin .- ŷ) .^ 2
     else
@@ -537,7 +537,6 @@ function contrastive_loss(ŷ, y, margin = epseltype(ŷ),agg=mean)
     end
     return agg(y .* square_pred + (1 .- y) .* margin_square)
 end
-
 
 ```@meta
 DocTestFilters = nothing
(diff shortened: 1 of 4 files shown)


test/losses.jl
```diff
@@ -6,15 +6,15 @@ using Flux.Losses: xlogx, xlogy
...
-                    Flux.Losses.binary_focal_loss, Flux.Losses.focal_loss]
+  Flux.Losses.crossentropy, Flux.Losses.logitcrossentropy,
+  Flux.Losses.binarycrossentropy, Flux.Losses.logitbinarycrossentropy,
+  Flux.Losses.kldivergence,
+  Flux.Losses.huber_loss,
+  Flux.Losses.tversky_loss,
+  Flux.Losses.dice_coeff_loss,
+  Flux.Losses.poisson_loss,
+  Flux.Losses.hinge_loss, Flux.Losses.squared_hinge_loss,
+  Flux.Losses.binary_focal_loss, Flux.Losses.focal_loss,Flux.Losses.contrastive_loss]
```
CarloLucibello, line 17: these should be indented back

src/losses/functions.jl
```diff
@@ -527,6 +527,18 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
     agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=dims))
 end
 
+function contrastive_loss(ŷ, y; agg = mean, margin = epseltype(ŷ))
```
CarloLucibello, line 530: docstring needed

arcAman07: I did add some dummy documentation for the time being just the loss function syntax.

arcAman07:
Having some troubles with the PR

src/losses/functions.jl
```diff
@@ -521,12 +521,60 @@ true
...
+julia> Flux.focal_loss(ŷ, y) ≈ 1.1277571935622628
+true
+```
+
+See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
+
+"""
+
+function contrastive_loss(ŷ, y; agg = mean, margin = epseltype(ŷ))
+    square_pred = ŷ .^ 2
```
mcabbott, line 565: You don't use `square_pred`, so this can be fused into the last line.

src/losses/functions.jl
```diff
@@ -521,12 +521,60 @@ true
...
+    max_val = max(sum(margin .- ŷ), 0)
```
mcabbott, line 566: More efficiently:
```suggestion
    max_val = max(sum(ŷᵢ -> (margin - ŷᵢ), ŷ), 0)
```

src/losses/functions.jl
```diff
@@ -19,8 +19,8 @@ julia> Flux.mae(y_model, 1:3)
 ```
 """
 function mae(ŷ, y; agg = mean)
-  _check_sizes(ŷ, y)
-  agg(abs.(ŷ .- y))
+    _check_sizes(ŷ, y)
+    agg(abs.(ŷ .- y))
 end
```
mcabbott, line 24: Your editor has changed many other lines. Can you tell it not to? 

The PR should only change what's being changed. And Flux uses 2 spaces, not 4 (although there are mistakes).

src/losses/functions.jl
```diff
@@ -19,8 +19,8 @@ julia> Flux.mae(y_model, 1:3)
 ```
 """
 function mae(ŷ, y; agg = mean)
-  _check_sizes(ŷ, y)
-  agg(abs.(ŷ .- y))
+    _check_sizes(ŷ, y)
+    agg(abs.(ŷ .- y))
```
CarloLucibello, line 23: indentation is flux is 2 whitespaces. Please don't change code all around the place. A PR should introduce a focused set of changes.

src/losses/functions.jl
```diff
@@ -521,12 +521,59 @@ true
...
+"""
+function contrastive_loss(ŷ, y; agg = mean, margin = epseltype(ŷ))
+    square_pred = ŷ .^ 2
+    max_val = max(sum(margin .- ŷ), 0)
+    if max_val > 0
+        margin_square = (margin .- ŷ) .^ 2
+    else
+        margin_square = 0
+    end
+    return agg(y .* square_pred + (1 .- y) .* margin_square)
```
mcabbott, line 572: I think this `if` should probably just be two final results, and not construct this unstable intermediate.

src/losses/functions.jl
arcAman07: Really sorry for this, I was just understanding the codebase and this might have happened during it. Wouldn't happen again.

src/losses/functions.jl
CarloLucibello: is it more efficient? Are we that good at higher-order functions' derivatives now?

src/losses/functions.jl
arcAman07: Yeah I just noticed it now. I don't really know how my editor is messing up the spaces in the codebase, even fixing it has become an issue. ( used VsCode for it )

src/losses/functions.jl
mcabbott: Good point, in the gradient this might have to take a more generic path, while the broadcast `-` has a hand-written one. In fact re-writing it with `sum( ŷ .- margin)` would probably be even better, IIRC that'll save a copy.

mcabbott: (With a simple enough function, `sum(f, x)` should be very efficient after https://github.com/JuliaDiff/ChainRules.jl/pull/529. But not with a lamda.)

src/losses/functions.jl
mcabbott: (Carlo was 30s quicker)

src/losses/functions.jl
arcAman07: Yeah can be directly added to the last line

src/losses/functions.jl
arcAman07: Would add a ternary operator

src/losses/functions.jl
mcabbott: Oh and I missed a reply. Somehow you need to turn off auto-formatting. And then what I'd do is copy the change out to another file, reset your branch back to master, then paste the change back as a new commit, then `push --force` the new version to github. 

Ah, in fact all the changes seem to be on one commit, https://github.com/FluxML/Flux.jl/pull/1884/commits/03d8fbd4a08735b0028c078be57adaf4db525dc2 . So perhaps easier is to just revert that one.

arcAman07: Just turned the autoformatter off, gave so many troubles right now. Made some changes to the loss function for it to be more efficent. ( Added your change along with the ternary operator, also removed the extra variable square_pred I had added in the first line )

src/losses/functions.jl
mcabbott: No, I don't mean to save lines. I mean that this introduces a type instability, and I think it'd be clearer to have the final answer be `agg(y .* ŷ .^ 2)` on one branch.

PR #1885, opened 2022-02-20
Addition of Siamese contrastive loss

arcAman07:
Loss function added and exported.
Looked into formatting and checked it.
Tests added.
Docs added. ( Incomplete currently will edit if PR merged )


arcAman07:
I think it's ready to be merged. Will write more tests and complete the docs when its merged.

src/losses/functions.jl
```diff
@@ -527,6 +527,46 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+true
+```
+                                    
+See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
+                                    
+"""
+function contrastive_loss(ŷ, y,margin = epseltype(ŷ); agg=mean)
+    max_val = max(sum(ŷᵢ -> (margin - ŷᵢ), ŷ), 0)
+    max_val > 0 ? margin_square = (margin .- ŷ) .^ 2 : margin_square = 0
+    return agg(y .* (ŷ .^ 2) + (1 .- y) .* margin_square)
```
mcabbott, line 567: Continuing #1884, I think the best way is probably something like this:
```suggestion
    max_val = max(-sum(ŷ .- margin), 0)
    if max_val > 0
        return agg(@. y * ŷ^2 + (1 - y) * (margin - ŷ)^2)
    else
        return agg(@. y * ŷ^2)
    end
end    
```

arcAman07: will commit the above changes and send the PR again then

arcAman07:
The last committed change was not working, so I reverted back to my previous code. It's working properly, but for extra performance it needs to be checked.

src/losses/functions.jl
mcabbott: The mistake is just that it has `end` twice, sorry.

src/losses/functions.jl
```diff
@@ -527,6 +527,46 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+julia> Flux.focal_loss(ŷ, y) ≈ 1.1277571935622628
+true
+```
+                                    
+See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
+                                    
+"""
+function contrastive_loss(ŷ, y, margin = epseltype(ŷ);agg = mean)
+    max_val = max(sum(ŷᵢ -> (margin - ŷᵢ), ŷ), 0)
+    max_val > 0 ? margin_square = (margin .- ŷ) .^ 2 : margin_square = 0
```
ToucheSir, line 566:
```suggestion
    margin_square = max_val > 0 ? abs2.(margin .- ŷ) : 0
```
Note: I'm not sure if the `abs2` is correct.

mcabbott: If `ŷ` comes from `softmax`, won't it always be real?

src/losses/functions.jl
mcabbott: (Now fixed above)

src/losses/functions.jl
ToucheSir: Does it always come from softmax? Does this function not work with binary or multilabel targets?

mcabbott: I was just going by the docstring: "The input, 'ŷ', is expected to be normalized (i.e. [`softmax`](@ref) output)." 

`abs2` surely won't hurt though. But IMO we should write it https://github.com/FluxML/Flux.jl/pull/1885/files#r810673814 first, no need to allocate a separate array `margin_square` and introduce a type-instability.

src/losses/functions.jl
mcabbott: Version with `abs2`. Not sure which is better really.
```suggestion
    max_val = max(-sum(ŷ .- margin), 0)
    if max_val > 0
        return agg(@. y * abs2(ŷ) + (1 - y) * abs2(margin - ŷ))
    else
        return agg(@. y * abs2(ŷ))
    end
```

src/losses/functions.jl
arcAman07: Don't go into the docstring for the moment ( had used the same focal loss documentation just to see the added loss function was ready to be merged). Will revamp the docstring along with more tests if the code is fine

src/losses/functions.jl
arcAman07: I was having the same confusion,  which one to go for hence I did not change anything. If I can get a 👍 for any one of the above which is better performant,  will commit the changes

src/losses/functions.jl
arcAman07: Will get those changes done

src/losses/functions.jl
mcabbott: I guess the question is whether `ŷ` can even be complex here. Maybe someone will try, and this should err on the side of caution?

arcAman07: Yep , continuous usage of this function on a variety of models( for training purposes) can help us know more about it.

arcAman07:
Reviewed changes added. Just the docs ( in the dosctring ) and more tests can be added.

arcAman07:
Anything else to be added to it or modifed ?, will do the needful

darsnack (changes requested):
The PyTorch maintainers went with "Siamese contrastive loss" since "contrastive loss" was too broad. Just searching for info on the implementation issues raised in the suggestions, I'm inclined to agree here. "Contrastive loss" is too poorly defined to indicate a specific function.

arcAman07:
Also the function name changed from "contrastive_loss" to "siamese_contrastive_loss" to make it specific as I agree contrastive loss functions is too broad a topic which has had multiple implementations. 
All the reviewed changes, have been made 👍

src/losses/functions.jl
```diff
@@ -527,6 +527,47 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+ 0.0900306  0.0900306  0.0900306  0.0900306  0.0900306
+ 0.244728   0.244728   0.244728   0.244728   0.244728
+ 0.665241   0.665241   0.665241   0.665241   0.665241
+                                    
+julia> Flux.focal_loss(ŷ, y) ≈ 1.1277571935622628
+true
+```
+                                    
+See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
+
```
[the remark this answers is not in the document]
arcAman07: Cool, I completed the reviewed changes. Will work on the documentation for this loss function and will update the docstring 👍. Any other changes that needs to be worked on?

test/losses.jl
```diff
@@ -210,3 +210,13 @@ end
...
 end
+
+@testset "contrastive_loss" begin
+    y1 = [1 0
+          0 0
+          0 1]
+    ŷ1 = [0.4 0.2
+          0.5 0.5
+          0.1 0.3]
+    @test Flux.contrastive_loss(ŷ1, y1,1) ≈ 0.3666666666666667
```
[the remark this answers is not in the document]
darsnack: Running a pair of inputs through the loss function and saving the output for use in a test will only test that the function doesn't error. It does not test whether the function is correct.

You need to write your tests without using the function that you implemented to produce the output. Either it should be calculated by hand mathematically, or it can be copied from a trusted reference elsewhere (like PyTorch or Tensorflow tests).

Same as the docstring, you don't need to implement everything in a single commit, but you will need to implement thorough testing before the PR is ready to be merged.

src/losses/functions.jl
```diff
@@ -527,6 +527,47 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+```
+                                    
+See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
+                                    
+"""
+function siamese_contrastive_loss(ŷ, y, margin = epseltype(ŷ); agg = mean)
+    _check_sizes(ŷ, y)
+    max_val = max(sum(margin .- ŷ), 0)
+    margin_square = max_val > 0 ? abs2.(margin .- ŷ) : 0
+    return agg(y .* margin_square + (1 .- y) .* (ŷ .^ 2))
```
mcabbott, line 568: I still don't think we should define `margin_square` like this. It materialises a broadcast that you don't have to, and introduces a type instability which you don't need.

Also, it should either use `abs2` everywhere, or ^2 everywhere.

src/losses/functions.jl
```diff
@@ -527,6 +527,47 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+```
+                                    
+See also: [`Losses.binary_focal_loss`](@ref) for binary (not one-hot) labels
+                                    
+"""
+function contrastive_loss(ŷ, y, margin = epseltype(ŷ);agg = mean)
+    _check_sizes(ŷ, y)
+    max_val = max(sum(ŷᵢ -> (margin - ŷᵢ), ŷ), 0)
+    margin_square = max_val > 0 ? abs2.(margin .- ŷ) : 0
+    return agg(y .* (ŷ .^ 2) + (1 .- y) .* margin_square)
```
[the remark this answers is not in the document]
darsnack: This issue still needs to be considered by other maintainers cause I'm not sure my suggestion is the best choice.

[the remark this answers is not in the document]
arcAman07: For me I have seen the reference of this loss function in pytorch as suggested above, I have created these siamese network models using variety of datasets ( from preloaded datasets like MNIST, Fashion MNIST to real life datasets) with tensorflow using this custom loss function. Hence I went for this implementation of the siamese_contrastive_loss function which always gave me amazing results. If it needs to be changed, I am up for it. It was just my personal preference

src/losses/functions.jl
darsnack: `ŷ` is supposed to be a Euclidean distance between input samples by definition. It will not be complex.

arcAman07: Yep 👍, just returns the euclidean distance which is a test for similarity for this loss function.

CarloLucibello:
Docs have to come in this PR, no need for future PRs and we are in no hurry. Let's also think a bit about https://github.com/FluxML/Flux.jl/pull/1885#discussion_r812025466.

arcAman07:
> Docs have to come in this PR, no need for future PRs and we are in no hurry. Let's also think a bit about [#1885 (comment)](https://github.com/FluxML/Flux.jl/pull/1885#discussion_r812025466).

Cool, working on the docs but would need some help as I am not too used to it, will try though and add the commit.

arcAman07:
Yep implementation of it differs between pytorch and tensorflow, but the core outcome is the same. Indeed will need to ponder over that issue so that we know which to go for.

src/losses/functions.jl
```diff
@@ -527,6 +527,48 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
     agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=dims))
 end
 
+"""
+    siamese_contrastive_loss(ŷ, y, margin = epseltype(ŷ); agg = mean)
```
CarloLucibello, line 531: default margin is supposed to be 1 or what?

arcAman07: Yeah default margin value is 1, but I haven't implemented it till now. Will add default value to margin as 1

src/losses/functions.jl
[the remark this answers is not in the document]
ToucheSir: I think the PyTorch PR is actually consistent with the other two, just that the terms are in the opposite order for some reason.

arcAman07:
Docs also fully added, along with default margin value is set to 1 ( as also mentioned in the papers ).
Hopefully have added all the necessary things for the implementation of this loss function in this PR.
If certain changes to be made, ready to inculcate it 👍

arcAman07:
Not sure what is the error , can't find out but one of the tests isnt being passed. Can somebody please help me with it, rest all good.

src/losses/functions.jl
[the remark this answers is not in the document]
arcAman07: I am also now kinda confused here, I was just using this:
image
with tensorflow, but after seeing the pytorch implementation, am a bit confused.

src/losses/functions.jl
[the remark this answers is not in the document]
ToucheSir: Woops, you are indeed correct and I'm not sure how the wires got crossed on my end. Given the "TF" version is in the (not terribly well vetted) addons repo and not core, I have more reason to doubt its implementation. This should definitely be changed.

src/losses/functions.jl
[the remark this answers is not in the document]
ToucheSir: Woops, you are indeed correct and I'm not sure how the wires got crossed on my end. Given the "TF" version is in the (not terribly well vetted) addons repo and not core, I have more reason to doubt its implementation. This should definitely be changed.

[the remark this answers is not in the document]
arcAman07: Yeah probably it needs to be changed then, we have implemented the loss function as it is implemented in pytorch and the paper. So we are all good about it

src/losses/functions.jl
```diff
@@ -527,6 +527,43 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+                                    
+Return the [siamese_contrastive_loss](http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf)
+which can be used in training Siamese Networks.
```
darsnack, line 534:
```suggestion
Return the [contrastive loss](http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf)
which can be useful for training Siamese Networks.
```

src/losses/functions.jl
```diff
@@ -527,6 +527,43 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+                                 
+margin: Integer, defines the baseline for distance for which pairs should be classified as dissimilar. - (default is 1).
```
darsnack, line 536:
```suggestion
Specify `margin` to set the baseline for distance at which pairs are dissimilar.
```

src/losses/functions.jl
```diff
@@ -527,6 +527,43 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+julia> Flux.siamese_contrastive_loss(ŷ1, y1,1) ≈ 0.2333333333333333
+true
+```                                    
+"""
+function siamese_contrastive_loss(ŷ, y, margin = 1; agg = mean)
+    _check_sizes(ŷ, y)
+    max_val = max(sum(margin .- ŷ), 0)
+    margin_square = max_val > 0 ? abs2.(margin .- ŷ) : 0
+    return agg(y .* margin_square + (1 .- y) .* abs2.(ŷ))
+end
```
darsnack, line 565: `abs2` can be changed back to `^2` since `ŷ` is a Euclidean distance by definition (so it is real). And @mcabbott's suggestion from [here](https://github.com/FluxML/Flux.jl/pull/1885#discussion_r810673814) still needs to be incorporated to minimize type instabilities.

darsnack (changes requested):

src/losses/functions.jl
CarloLucibello: the docstring for the margin default needs to be changed. 
As a general comment,  I think this PR has been a little bit rushed, changes are pushed too quickly. Take your time, we are in no hurry! 30 commits have been pushed for a simple PR! My advice is to do things calmy and locally, check everything carefully, and only when you feel things are ready push the commit.

src/losses/functions.jl
```diff
@@ -527,6 +527,43 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+ 0  0
+ 0  1
+                                    
+julia> ŷ1 = [0.4 0.2
+             0.5 0.5
+             0.1 0.3]
+3×2 Matrix{Float64}:
+ 0.4  0.2
+ 0.5  0.5
+ 0.1  0.3
```
CarloLucibello, line 554:
```suggestion
julia> y = [1 0
                 0 0
                 0 1];
           
julia> ŷ = [0.4 0.2
                0.5 0.5
                0.1 0.3];
```

src/losses/functions.jl
```diff
@@ -527,6 +527,43 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+                                    
+julia> Flux.siamese_contrastive_loss(ŷ1, y1,1) ≈ 0.2333333333333333
```
CarloLucibello, line 556:
```suggestion
julia> Flux.siamese_contrastive_loss(ŷ, y,1) ≈ 0.2333333333333333
```

CarloLucibello: ops, I accidentally committed this change

src/losses/functions.jl
arcAman07: Cool, will just do the changes now 👍

src/losses/functions.jl
CarloLucibello: You should also compare the docstrings to this: as you should notice, the mathematical definition of the operation is missing

arcAman07 (commented):
Done with proposed changes, abs2 removed , went with the suggestion @mcabbott suggested

arcAman07 (commented):
Changes inculcated in the future commits.

arcAman07 (commented):
Changes inculcated 👍

arcAman07 (commented):
Done 👍

arcAman07 (commented):
Done 👍

arcAman07 (commented):
Done 👍

arcAman07 (commented):
Done 👍

src/losses/functions.jl
```diff
@@ -527,6 +527,38 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+```jldoctest
+julia> y = [1 0
+                 0 0
+                 0 1];
+           
+julia> ŷ = [0.4 0.2
+                0.5 0.5
+                0.1 0.3];
+                                    
+julia> Flux.siamese_contrastive_loss(ŷ, y,1) ≈ 0.2333333333333333
```
mcabbott, line 548: Is there an example you can think up from which we learn something? If all we see is that some random numbers give a random number, then it's probably better to omit it.

Will `y` almost always be Bool? Then probably write it as onehot to remind people. Will y^ always add to 1? Then call `softmax` to remind people. Does this penalise something more heavily than other loss functions? Then compare it to its nearest relative. Will people forget which y has the hat? Things like `...(softmax(y_model), y_label)` will remind them. What does `margin` do, can you invent a small example which shows two different values somehow?

No rush, but good examples can be very helpful.

src/losses/functions.jl
mcabbott: Also:

* Should `margin` be a positional argument or a keyword?
* Should it be `margin::Real`, or are there other uses? If there are, they probably need tests. 
* The other losses take a `dims` keyword, probably this one should too.

src/losses/functions.jl
arcAman07: Will probably need to think of them and add it as tests. Mind was stuck and adamant at getting the core functionality bang on. If possible we can add further better tests as further PRs( just am not able to thing of them at the moment which can show the wide applications and range of this loss function).
Was thinking of adding a Fashion MNIST siamese network tutorial using the preloaded dataset , along with the loss function on the core website tutorial . Thought that will be a pretty good idea.

src/losses/functions.jl
arcAman07: It actually just ranges in single digit number to give good results and for experimentation, it's default is mostly used everywhere which is 1
So it's  one of the extra arguments which is not that important to write separate tests for it in my opinion. But we can discuss it further and if need be we can do that too. Just didn't feel dims keyword needed to be added, but again am open for suggestion if it needs to be added

darsnack:
It's okay if you can't think of more examples or tests right at this moment. We are no hurry here; good PRs can take weeks or months to complete. Please take your time, and don't feel the need to immediately have an answer for every review comment. If you don't respond for a week, no one will be upset.

src/losses/functions.jl
mcabbott: I guess there is in fact no `dims` here. 

I do think `margin` should be a keyword. No other loss functions take 3 positional arguments. 

I think you are saying it is always a real number, so please encode that in the type. 

There should always be a test. If the argument is important enough to include at all, then it's important enough to test. Then you will notice when you make a typo while messing with it next year sometime, rather than breaking everyone else's code.

src/losses/functions.jl
mcabbott: No rush, take your time. But a good docstring describing what this is, and ideally a good compact example showing it, are what need to be in the first PR before it's merged. That and tests. 

Bigger things like a "Fashion MNIST siamese network tutorial" are great too, but will be too big for the docstring, and should not be in this PR. It's fine to work on e.g. a model zoo PR with such things, which needs this PR, at the same time if you like. Sometimes that helps figure out the best way to do things.

arcAman07:
Currently getting 1 test failed at "no spurious promotions" loss functions. I didn't change that but getting an error there, will need some help. => the main problem currently
Updates ( For the rest of the features ):  Margin added as a keyword with type of Float64 and a default value of 1 ( which is a standard is siamese contrastive loss functions and people don't change that very often )
Added More Robust Testing : Made the testing much more robust and relatable ( relatable functions like softmax used, along wiht variety of examples where one can see the change in the loss function when margin is incremented or decremented by a certain value).
Also changed the docstring made it much more simple and sweet.

mcabbott:
The promotion is because `margin = 1.0` causes an array `eltype(y) == Float32` to give an answer which is Float64.

I think you want `margin::Real` so that `margin=1` is accepted, as this won't promote floats. 

If you want to go further, you can (in the function body) have something like `m32 = convert(float(eltype(y^)), margin)` and use that in the calculation. Then even if someone provides `margin = 1.5` not `margin = 1.5f0`, it still won't cause promotion. But this may be more clutter than it's worth. Making sure the default doesn't promote is the more important step.

arcAman07:
> The promotion is because `margin = 1.0` causes an array `eltype(y) == Float32` to give an answer which is Float64.
> 
> I think you want `margin::Real` so that `margin=1` is accepted, as this won't promote floats.
> 
> If you want to go further, you can (in the function body) have something like `m32 = convert(float(eltype(y^)), margin)` and use that in the calculation. Then even if someone provides `margin = 1.5` not `margin = 1.5f0`, it still won't cause promotion. But this may be more clutter than it's worth. Making sure the default doesn't promote is the more important step.

Thanks a lot , I forgot to notice that. Looks like everything is sorted now. Any more changes or feedbacks will be appreciated

arcAman07:
> It's okay if you can't think of more examples or tests right at this moment. We are no hurry here; good PRs can take weeks or months to complete. Please take your time, and don't feel the need to immediately have an answer for every review comment. If you don't respond for a week, no one will be upset.

Sure actually whatever issue I take I try to fix it asap rather than procrastinate.👍

mcabbott:
It's not about procrastination. The rush creates a lot of noise (130 notifications and counting), and I think many comments are getting lost. It wastes time if people have to repeat themselves. Please read above about about indent, docstrings, signatures, tests.

Ideally you should make one commit with your best attempt. Then discuss different things in review, each in a thread. Think about them, fix them locally. Test locally. Re-read that you didn't miss some comments. Explain why you think your way is better than some suggestions if you don't like them, in the threads. Then push your changes. Of course it never quite works that way, and that's fine... but this isn't really a chat system, considered replies make it easier for others to see what's going on. 

If you are editing things on the web interface, it's probably worth figuring out how to do this locally on your computer instead, where it's easier to try things out. Without making a commit for every save.

arcAman07:
> It's not about procrastination. The rush creates a lot of noise (130 notifications and counting), and I think many comments are getting lost. It wastes time if people have to repeat themselves. Please read above about about indent, docstrings, signatures, tests.
> 
> Ideally you should make one commit with your best attempt. Then discuss different things in review, each in a thread. Think about them, fix them locally. Test locally. Re-read that you didn't miss some comments. Explain why you think your way is better than some suggestions if you don't like them, in the threads. Then push your changes. Of course it never quite works that way, and that's fine... but this isn't really a chat system, considered replies make it easier for others to see what's going on.
> 
> If you are editing things on the web interface, it's probably worth figuring out how to do this locally on your computer instead, where it's easier to try things out. Without making a commit for every save.

First of all I would really like to apologise to all the reviewers, I made it a cumbersome process with just way too many commits.
Being really new to the julia language ( and the community ) I was just way too excited , hence I was testing so many things that it led to various commits, which I promise won't happen in the future. I really made this cluttered more like a chat rather than a focused review of the PR. As I keep experimenting, try out various libraries and go deeper into the language I will make sure I read up about the various things above like "docstrings" , "signatures" and tests. I worked with VS code initially for the first few commits in the PR, but as soon as the indentation error ( due to configurations of my VS code ) which really messed up the entire codebase, I was too scared of using it and hence used the jupyter notebook for testing and committed using the web interface. Have changed the settings thus far of my editor to make sure it never happens again, so that I can create a much more focused PR in the future. These all mistakes were from a sincere place and I hope to not repeat them in the future.
Again Apologies for kinda ruining the PR.

arcAman07 (commented):
Changes implemented

arcAman07 (commented):
Changes implemented

arcAman07 (commented):
Changes implemented

src/losses/functions.jl
```diff
@@ -527,6 +527,46 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+ 0  0
+ 0  1									
+           
+julia> ŷ = [0.4 0.2
+            0.5 0.5
+            0.1 0.3]
+3×2 Matrix{Float64}:
+ 0.4  0.2
+ 0.5  0.5
+ 0.1  0.3
```
darsnack, line 554: Since `y` and `ŷ` are "distances" between samples, maybe a more informative example would start with two samples (`x`) and show the output of this function when we compare samples that are same vs. different.

darsnack (changes requested):
I left a few test and docstring suggestions. Don't worry, the PR review process is not obvious, and we want to help you learn the process. On this round, try to follow @mcabbott's suggestions and take things slow, think about the comments, make sure you haven't missed any comments, and only ask for a new review when you are confident that the code is the best you think it can be.

arcAman07:
bors try

arcAman07:
Tests updated according to the reviews (
1) Test Case added when similar input and output is generated
2) Test Case added when margin = 0
3) Contrastive loss function updated to throw a "DomainError" when margin is less than 0. Suitable test case added for that when user enters a margin value less than zero)
Doc strings updated to show the difference in the output when input and ouputs are similar vs inputs and outputs are different.

GPU integration with julia v1.6 is failing in buildkite, how to fix that?
All reviewed changes applied.

CarloLucibello:
Why did you close it?

arcAman07:
I tried it using VS code while I was committing I saw the codebase was fine with no indentation but after the PR was updated it changed the codebase by again changing the indentation from 2 to 4 , it also changed the codebase of tests of loss functions. Hence I had no other choice other than deleting the forked repository. So then I just fixed the auto formatter, forked the repo again and made the changes to send a updated PR inculcating the reviewed changes.

CarloLucibello:
Ok good, you did the right thing, just could have added a cross-reference to the new PR, something like 

"Closing in favor of #1892 "

PR #1886, opened 2022-02-20
Make RNN layers accept `in => out`

mcabbott:
This does #1658 for all the RNN layers.

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -187,7 +187,7 @@ struct LSTMCell{A,V,S}
   state0::S
 end
 
-function LSTMCell((in, out)::Pair{<:Integer, <:Integer};
+function LSTMCell((in, out)::Pair;
                   init = glorot_uniform,
                   initb = zeros32,
                   init_state = zeros32)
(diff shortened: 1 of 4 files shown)


src/layers/recurrent.jl
```diff
@@ -187,7 +187,7 @@ struct LSTMCell{A,V,S}
   state0::S
 end
 
-function LSTMCell(in::Integer, out::Integer;
+function LSTMCell((in, out)::Pair{<:Integer, <:Integer};
```
ToucheSir, line 190: Why is the additional eltype constraint required here in particular?

mcabbott: This signature just happens to already be written out on different lines.

ToucheSir: Sorry, not sure I follow. Is the plan to update `RNNCell` and `GRU(v3)Cell` to match?

mcabbott: No. I typed this one this way. Then I thought it doesn't really matter, and isn't worth the clutter of changing the others to be multi-line definitions. 

Now fixed, they all agree. 

(Previously, `GRUCell(in, out;` took Any, Any, btw.)

ToucheSir (approved):
LGTM!

PR #1888, opened 2022-02-22
Speeding up onehotbatch by creating OneHotArray directly

TLipede:
- Speeding up `onehotbatch` by constructing array directly

Hey there.
I noticed that the `onehotbatch` function calls the `onehot` function repeatedly, and then merges the resulting `OneHotVector`s.
This leads to the `OneHotVector` constructor being called many times, especially as data grows large. This PR changes the behaviour slightly, constructing a `OneHotArray` directly.

EDIT - Testing against master:
EDIT2 - Updated after broadcasts replaced with maps
EDIT3 - Added example benchmark cases from #1861 / #1844 
EDIT4 - Update after replacing maps with comprehension

Before:
```julia
julia> small_data = rand(MersenneTwister(0), 'a':'e', 20);
julia> small_labels = 'a':'e'
julia> big_data = rand(MersenneTwister(0), '.':'z', 10^6);
julia> big_labels = '.':'z'

julia> @btime onehotbatch($small_data, $small_labels);
  694.440 ns (5 allocations: 416 bytes)

julia> @btime onehotbatch($big_data, $big_labels);
  1.474 s (8000007 allocations: 358.58 MiB

julia> @btime onehotbatch($small_data, $big_labels);
  24.796 μs (165 allocations: 7.55 KiB)

julia> @btime onehotbatch($small_data, labels) setup=(labels=Tuple(small_labels));
  415.618 ns (3 allocations: 304 bytes)

julia> @btime onehotbatch(data, $big_labels, 'c') setup=(data=rand(MersenneTwister(0), 1:50, 10^6));
  1.657 s (8000007 allocations: 358.58 MiB)

# Example cases from #1861 / #1844
julia> const bases_dna = ['A', 'C', 'G', 'T'];
julia> dna_sequence = "CCGAGGGCTATGGTTTGGAAGTTAGAACCCTGGGGCTTCTCGCGGA";
julia> short_string = String(rand('a':'z', 10));
julia> string_labels = 'a':'z';

julia> @btime onehotbatch($dna_sequence, $bases_dna);
  793.137 ns (4 allocations: 528 bytes)

julia> @btime onehotbatch($short_string, $string_labels);
  1.512 μs (5 allocations: 480 bytes)
```

After:
```julia
julia> small_data = rand(MersenneTwister(0), 'a':'e', 20);
julia> small_labels = 'a':'e'
julia> big_data = rand(MersenneTwister(0), '.':'z', 10^6);
julia> big_labels = '.':'z'

julia> @btime onehotbatch($small_data, $small_labels);
  646.539 ns (4 allocations: 272 bytes)

julia> @btime onehotbatch($big_data, $big_labels);
  9.802 ms (3 allocations: 3.81 MiB)

julia> @btime onehotbatch($small_data, $big_labels);
  522.220 ns (2 allocations: 160 bytes)

julia> @btime onehotbatch($small_data, labels) setup=(labels=Tuple(small_labels));
  371.029 ns (2 allocations: 160 bytes)

julia> @btime onehotbatch(data, $big_labels, 'c') setup=(data=rand(MersenneTwister(0), 1:50, 10^6));
  266.723 ms (3 allocations: 3.81 MiB)

# Example cases from #1861 / #1844
julia> const bases_dna = ['A', 'C', 'G', 'T'];
julia> dna_sequence = "CCGAGGGCTATGGTTTGGAAGTTAGAACCCTGGGGCTTCTCGCGGA";
julia> short_string = String(rand('a':'z', 10));
julia> string_labels = 'a':'z';

julia> @btime onehotbatch($dna_sequence, $bases_dna);
  753.800 ns (3 allocations: 288 bytes)

julia> @btime onehotbatch($short_string, $string_labels);
  1.456 μs (4 allocations: 384 bytes)
```

For very small data/labels, the older implementation is a bit faster, but then the newer one tends to do better.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -190,7 +190,7 @@ function _onehotbatch(data, labels)
   n_labels = length(labels)
   indices = _findval.(data, Ref(labels))
   if nothing in indices
-    unexpected_values = unique(data[indices .== -1])
+    unexpected_values = unique(data[indices .== nothing])
     error("Values $unexpected_values are not in labels")
   end
   OneHotArray(indices, n_labels)
@@ -200,7 +200,7 @@ function _onehotbatch(data, labels, default)
   n_labels = length(labels)
   indices = _findval.(data, Ref(labels))
   if nothing in indices
-    indices = replace!(indices, -1 => _findval(default, labels))
+    indices = replace!(indices, nothing => _findval(default, labels))
   end
   return OneHotArray(indices, n_labels)
 end
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -188,7 +188,7 @@ onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ?
 # NB function barrier:
 function _onehotbatch(data, labels)
   n_labels = length(labels)
-  indices = _findval.(data, Ref(labels))
+  indices = map(x -> _findval(x, labels), data)
   if nothing in indices
     unexpected_values = unique(data[indices .== nothing])
     error("Values $unexpected_values are not in labels")
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -198,11 +198,15 @@ end
 
 function _onehotbatch(data, labels, default)
   n_labels = length(labels)
-  indices = _findval.(data, Ref(labels))
+  indices = map(x -> _findval(x, labels), data)
   if nothing in indices
-    indices = replace(indices, nothing => _findval(default, labels))
+    default_index = _findval(default, labels)
+    isnothing(default_index) && error("Default value $default_index is not in labels")
+    replaced_indices = replace(indices, nothing => default_index)
+    return OneHotArray(replaced_indices, n_labels)
+  else
+    return OneHotArray(indices, n_labels)
   end
-  return OneHotArray(indices, n_labels)
 end
 
 """
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -190,7 +190,7 @@ function _onehotbatch(data, labels)
   n_labels = length(labels)
   indices = _findval.(data, Ref(labels))
   if nothing in indices
-    unexpected_values = unique(data[indices .== nothinghl])
+    unexpected_values = unique(data[indices .== nothing])
     error("Values $unexpected_values are not in labels")
   end
   return OneHotArray(indices, n_labels)
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -190,7 +190,7 @@ function _onehotbatch(data, labels)
   n_labels = length(labels)
   indices = _findval.(data, Ref(labels))
   if nothing in indices
-    unexpected_values = unique(data[indices .== nothing])
+    unexpected_values = unique(data[indices .== nothinghl])
     error("Values $unexpected_values are not in labels")
   end
   OneHotArray(indices, n_labels)
@@ -200,7 +200,7 @@ function _onehotbatch(data, labels, default)
   n_labels = length(labels)
   indices = _findval.(data, Ref(labels))
   if nothing in indices
-    indices = replace!(indices, nothing => _findval(default, labels))
+    indices = replace(indices, nothing => _findval(default, labels))
   end
   return OneHotArray(indices, n_labels)
 end
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -193,7 +193,7 @@ function _onehotbatch(data, labels)
     unexpected_values = unique(data[indices .== nothinghl])
     error("Values $unexpected_values are not in labels")
   end
-  OneHotArray(indices, n_labels)
+  return OneHotArray(indices, n_labels)
 end
 
 function _onehotbatch(data, labels, default)
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -188,8 +188,9 @@ onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ?
 function _onehotbatch(data, labels)
   indices = UInt32[something(_findval(i, labels), 0) for i in data]
   if 0 in indices
-    unexpected_values = unique(data[indices .== 0])
-    error("Values $unexpected_values are not in labels")
+    for x in data
+      isnothing(_findval(x, labels)) && error("Value $x not found in labels")
+    end
   end
   return OneHotArray(indices, length(labels))
 end
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -183,9 +183,27 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
  3  6  15  3  9  3  12  3  6  15  3
 ```
 """
-onehotbatch(ls, labels, default...) = _onehotbatch(ls, length(labels) < 32 ? Tuple(labels) : labels, default...)
-# NB function barier:
-_onehotbatch(ls, labels, default...) = batch([onehot(l, labels, default...) for l in ls])
+onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
+
+# NB function barrier:
+function _onehotbatch(data, labels)
+  n_labels = length(labels)
+  indices = _findval.(data, Ref(labels))
+  if nothing in indices
+    unexpected_values = unique(data[indices .== -1])
+    error("Values $unexpected_values are not in labels")
+  end
+  OneHotArray(indices, n_labels)
+end
+
+function _onehotbatch(data, labels, default)
+  n_labels = length(labels)
+  indices = _findval.(data, Ref(labels))
+  if nothing in indices
+    indices = replace!(indices, -1 => _findval(default, labels))
+  end
+  return OneHotArray(indices, n_labels)
+end
 
 """
     onecold(y::AbstractArray, labels = 1:size(y,1))
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -184,6 +184,7 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
 ```
 """
 onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
+onehotbatch(data::AbstractString, labels, default...) = onehotbatch([_ for _ in data], labels, default...)
 
 # NB function barrier:
 function _onehotbatch(data, labels)
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -187,7 +187,7 @@ onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ?
 
 # NB function barrier:
 function _onehotbatch(data, labels)
-  indices = [_findval(i, labels) for i in data]
+  indices = UInt32[_findval(i, labels) for i in data]
   if nothing in indices
     unexpected_values = unique(data[indices .== nothing])
     error("Values $unexpected_values are not in labels")
@@ -198,7 +198,7 @@ end
 function _onehotbatch(data, labels, default)
   default_index = _findval(default, labels)
   isnothing(default_index) && error("Default value $default is not in labels")
-  indices = [isnothing(_findval(i, labels)) ? default_index : _findval(i, labels) for i in data]
+  indices = UInt32[isnothing(_findval(i, labels)) ? default_index : _findval(i, labels) for i in data]
   return OneHotArray(indices, length(labels))
 end
 
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -202,7 +202,7 @@ function _onehotbatch(data, labels, default)
   indices = map(x -> _findval(x, labels), data)
   if nothing in indices
     default_index = _findval(default, labels)
-    isnothing(default_index) && error("Default value $default_index is not in labels")
+    isnothing(default_index) && error("Default value $default is not in labels")
     replaced_indices = replace(indices, nothing => default_index)
     return OneHotArray(replaced_indices, n_labels)
   else
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -184,30 +184,22 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
 ```
 """
 onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
-onehotbatch(data::AbstractString, labels, default...) = onehotbatch([_ for _ in data], labels, default...)
 
 # NB function barrier:
 function _onehotbatch(data, labels)
-  n_labels = length(labels)
-  indices = map(x -> _findval(x, labels), data)
+  indices = [_findval(i, labels) for i in data]
   if nothing in indices
     unexpected_values = unique(data[indices .== nothing])
     error("Values $unexpected_values are not in labels")
   end
-  return OneHotArray(indices, n_labels)
+  return OneHotArray(indices, length(labels))
 end
 
 function _onehotbatch(data, labels, default)
-  n_labels = length(labels)
-  indices = map(x -> _findval(x, labels), data)
-  if nothing in indices
-    default_index = _findval(default, labels)
-    isnothing(default_index) && error("Default value $default is not in labels")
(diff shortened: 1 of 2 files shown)


src/onehot.jl
```diff
@@ -183,9 +183,27 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
  3  6  15  3  9  3  12  3  6  15  3
 ```
 """
-onehotbatch(ls, labels, default...) = _onehotbatch(ls, length(labels) < 32 ? Tuple(labels) : labels, default...)
-# NB function barier:
-_onehotbatch(ls, labels, default...) = batch([onehot(l, labels, default...) for l in ls])
+onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
```
darsnack, line 186:
```suggestion
onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
onehotbatch(data::AbstractString, labels, default...) = onehotbatch(collect(data), labels, default...)
```

src/onehot.jl
```diff
@@ -183,9 +183,27 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
...
+
+# NB function barrier:
+function _onehotbatch(data, labels)
+  n_labels = length(labels)
+  indices = _findval.(data, Ref(labels))
+  if nothing in indices
+    unexpected_values = unique(data[indices .== nothinghl])
+    error("Values $unexpected_values are not in labels")
+  end
+  OneHotArray(indices, n_labels)
```
darsnack, line 196:
```suggestion
  return OneHotArray(indices, n_labels)
```

src/onehot.jl
```diff
@@ -183,9 +183,27 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
...
-# NB function barier:
-_onehotbatch(ls, labels, default...) = batch([onehot(l, labels, default...) for l in ls])
+onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
+
+# NB function barrier:
+function _onehotbatch(data, labels)
+  n_labels = length(labels)
+  indices = _findval.(data, Ref(labels))
+  if nothing in indices
+    unexpected_values = unique(data[indices .== nothinghl])
```
darsnack, line 193:
```suggestion
    unexpected_values = unique(data[indices .== nothing])
```

src/onehot.jl
```diff
@@ -183,9 +183,27 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
...
+  end
+  OneHotArray(indices, n_labels)
+end
+
+function _onehotbatch(data, labels, default)
+  n_labels = length(labels)
+  indices = _findval.(data, Ref(labels))
+  if nothing in indices
+    indices = replace(indices, nothing => _findval(default, labels))
+  end
```
darsnack, line 204:
```suggestion
  (nothing in indices) && replace!(indices, nothing => _findval(default, labels))
```

darsnack (changes requested):
Thanks! These changes might help.

src/onehot.jl
mcabbott: Are you sure this is better? It will probably leave the array `indices` a `Union{Nothing, X}`, rather than make a new one of uniform type.

With the current code, I'd suggest avoiding re-using the name `indices` for a new array of a different type, something like this:
```
indices2 = nothing in indices ? replace(indices, nothing => _findval(default, labels)) : indices
```
I believe that will evaluate `_findval(default, labels)` once, but it might still be clearer to pull it out so that there is no doubt.

src/onehot.jl
```diff
@@ -183,9 +183,27 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
...
 """
-onehotbatch(ls, labels, default...) = _onehotbatch(ls, length(labels) < 32 ? Tuple(labels) : labels, default...)
-# NB function barier:
-_onehotbatch(ls, labels, default...) = batch([onehot(l, labels, default...) for l in ls])
+onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
+
+# NB function barrier:
+function _onehotbatch(data, labels)
+  n_labels = length(labels)
+  indices = _findval.(data, Ref(labels))
```
mcabbott, line 191: I think using `map` is better than broadcasting here. 
```suggestion
  indices = map(x -> _findval(x, labels), data)
```
If `data` is a non-array iterable (like a generator) then this will save collecting it first.

src/onehot.jl
mcabbott: I think there should be no need to collect the string, if you `map` the `_findval` over it, not broadcast.

mcabbott (commented):
This looks promising. Most recent update is https://github.com/FluxML/Flux.jl/pull/1861, can you test this on those examples too?

src/onehot.jl
mcabbott: Even better might be to never create such an array. Maybe something like this would be ideal?
```
def_ind = _findval(default, labels) # and check that's not nothing?
indices = map(x -> something(_findval(x, labels), def_ind, data)
```
But benchmark to check.

TLipede:
> This looks promising. Most recent update is #1861, can you test this on those examples too?

Heya, sorry the description wasn't very clear. This is being tested against master. I'll edit the description.

src/onehot.jl
TLipede: Thanks, good shout.

mcabbott:
Comparing against master is good. But what I meant is, that PR (and the issue that inspired it) have some different examples, with strings not arrays. Finding problems to benchmark that you didn't make up for the purpose occasionally reveals blind spots.

src/onehot.jl
darsnack: I tried that but `map(f, ::AbstractString)` requires that `f` returns an `AbstractChar`.

mcabbott: Oh right, sorry. Then maybe it should be a comprehension?

darsnack: Yeah, that also makes dealing with `nothing` easier.

src/onehot.jl
TLipede: @mcabbott the solution with `something` is really cool! Will have to remember that one. It turns out to be a little slower, so I've just cleaned it up a bit.
@darsnack `replace!` won't work if all of the values in `data` aren't in `labels`, since then `indices` is `Vector{Nothing}`. I played around with having `_findval` return -1 instead, but it didn't seem to speed things up.

mcabbott: Oh that's a good catch. Do we have tests which hit the all-default path?

src/onehot.jl
TLipede: This appears to still require the collect.

src/onehot.jl
TLipede: I don't think so, will put something in.

src/onehot.jl
TLipede: Thanks both, I've gone with the comprehension.

TLipede:
I see what you mean now @mcabbott, added those examples now. There's a test and an error now if `default` isn't provided, but the error will only get triggered if unexpected values are encountered. Are we okay with that behaviour, or should the function always error?

src/onehot.jl
```diff
@@ -183,9 +183,32 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
...
-onehotbatch(ls, labels, default...) = _onehotbatch(ls, length(labels) < 32 ? Tuple(labels) : labels, default...)
-# NB function barier:
-_onehotbatch(ls, labels, default...) = batch([onehot(l, labels, default...) for l in ls])
+onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
+onehotbatch(data::AbstractString, labels, default...) = onehotbatch([_ for _ in data], labels, default...)
+
+# NB function barrier:
+function _onehotbatch(data, labels)
+  n_labels = length(labels)
+  indices = map(x -> _findval(x, labels), data)
```
darsnack, line 192: I think the comprehension suggestion is to use one here instead of `map` (not to add a different method for `AbstractString`). This way you can also deal with `nothing` directly in the comprehension.

darsnack (changes requested):

src/onehot.jl
TLipede: Ah, fair enough. I follow now 👍🏿.

src/onehot.jl
```diff
@@ -183,9 +183,24 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
...
 """
-onehotbatch(ls, labels, default...) = _onehotbatch(ls, length(labels) < 32 ? Tuple(labels) : labels, default...)
-# NB function barier:
-_onehotbatch(ls, labels, default...) = batch([onehot(l, labels, default...) for l in ls])
+onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
+
+# NB function barrier:
+function _onehotbatch(data, labels)
+  indices = UInt32[_findval(i, labels) for i in data]
+  if nothing in indices
```
mcabbott, line 191: Why doesn't this give errors? 
```
julia> UInt32[i==2 ? nothing : i for i in 1:4]
ERROR: MethodError: Cannot `convert` an object of type Nothing to an object of type UInt32
```

src/onehot.jl
```diff
@@ -183,9 +183,24 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
...
+    unexpected_values = unique(data[indices .== nothing])
+    error("Values $unexpected_values are not in labels")
+  end
+  return OneHotArray(indices, length(labels))
+end
+
+function _onehotbatch(data, labels, default)
+  default_index = _findval(default, labels)
+  isnothing(default_index) && error("Default value $default is not in labels")
+  indices = UInt32[isnothing(_findval(i, labels)) ? default_index : _findval(i, labels) for i in data]
```
mcabbott, line 201: Does running `_findval(i, labels)` twice pay for itself somehow? I would have thought that `something(_findval(i, labels), default_index)` would be faster.

src/onehot.jl
```diff
@@ -183,9 +183,24 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
  3  6  15  3  9  3  12  3  6  15  3
 ```
 """
-onehotbatch(ls, labels, default...) = _onehotbatch(ls, length(labels) < 32 ? Tuple(labels) : labels, default...)
-# NB function barier:
-_onehotbatch(ls, labels, default...) = batch([onehot(l, labels, default...) for l in ls])
+onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
+
+# NB function barrier:
```
mcabbott, line 188:
```suggestion
```
I'd drop this comment. It was sandwiched between two functions which you might be tempted to combine, but now seems marooned...

src/onehot.jl
TLipede: yeah you're right, forgot about something again

src/onehot.jl
TLipede: fixed now 👍🏿

test/onehot.jl
```diff
@@ -18,10 +18,13 @@ using Test
   @test onehotbatch("abc", 'a':'c') == Bool[1 0 0; 0 1 0; 0 0 1]
   @test onehotbatch("zbc", ('a', 'b', 'c'), 'a') == Bool[1 0 0; 0 1 0; 0 0 1]
 
+  @test onehotbatch([10, 20], [30, 40, 50], 30) == Bool[1 1; 0 0; 0 0]
+
   @test_throws Exception onehotbatch([:a, :d], [:a, :b, :c])
   @test_throws Exception onehotbatch([:a, :d], (:a, :b, :c))
   @test_throws Exception onehotbatch([:a, :d], [:a, :b, :c], :e)
   @test_throws Exception onehotbatch([:a, :d], (:a, :b, :c), :e)
+  @test_throws Exception onehotbatch([:a, :e], (:a, :b, :c), :d)
```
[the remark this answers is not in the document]
TLipede: This was supposed to test that we get an error for the default value not being present, even if no unexpected values get encountered. Should be fixed now.

tobi-lipede-oodle:
I'm not clear what happened with the buildkite test (it says MLUtils is not registered). Did I do something to cause that?

TLipede:
I'm not clear what happened with the buildkite test (it says MLUtils is not registered). Did I do something to cause that?

mcabbott:
Looks like it's failing on master, too: https://buildkite.com/julialang/flux-dot-jl/builds/2414

src/onehot.jl
```diff
@@ -183,9 +183,23 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
...
-onehotbatch(ls, labels, default...) = _onehotbatch(ls, length(labels) < 32 ? Tuple(labels) : labels, default...)
-# NB function barier:
-_onehotbatch(ls, labels, default...) = batch([onehot(l, labels, default...) for l in ls])
+onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
+
+function _onehotbatch(data, labels)
+  indices = UInt32[something(_findval(i, labels), 0) for i in data]
+  if 0 in indices
+    unexpected_values = unique(data[indices .== 0])
+    error("Values $unexpected_values are not in labels")
```
mcabbott, line 192: What this won't handle is non-indexable data:
```
julia> _onehotbatch((x for x in [10, 20, 10, 20, 30]), [10, 20])
ERROR: MethodError: no method matching getindex(::Base.Generator{Vector{Int64}, typeof(identity)}, ::BitVector)

julia> _onehotbatch("abcd", "abc")
ERROR: ArgumentError: logical indexing not supported for strings
```
Not quite as informative an error, but one compact solution is:
```suggestion
    for x in data
      isnothing(_findval(x, labels)) && error("Value $x not found in labels")
    end
```

TLipede: @mcabbott would this be reasonable? Or overcomplicate things, do you think?

```julia
function _unexpected_values(data, indices)
  unique_values = Set{eltype(data)}()
  it = 1
  for (i, (idx, val)) in enumerate(zip(indices, data))
    it = i
    idx != 0 && continue
    push!(unique_values, val)
    length(unique_values) == 10 && break
  end
  string_values = "(" * join(unique_values, ", ")
  closing_brac = it == length(data) ? ")" : "...)"
  return string_values * closing_brac
end

function _onehotbatch(data, labels)
  indices = UInt32[something(_findval(i, labels), 0) for i in data]
  if 0 in indices
    unexpected_values = _unexpected_values(data, indices)
    error("Values $unexpected_values are not in labels")
  end
  return OneHotArray(indices, length(labels))
end
```

```
julia> onehotbatch((x for x in [10, 20, 10, 20, 30]), [10, 20])
ERROR: Values (30) are not in labels

julia> onehotbatch("abcd", "ab")
ERROR: Values (c, d) are not in labels

julia> onehotbatch(1:10^8, "abcd")
ERROR: Values (5, 4, 6, 7, 2, 10, 9, 8, 3, 1...) are not in labels
```

mcabbott: I don't think runtime is a concern at all, but my vote is that this is probably more complexity than is justified. 

You get this error when you mess up and use labels "dog, cat, ..." with data [1,3,2, ...], right? There's some value to printing one wrong point, to lead you to the mistake, but probably not much value in collecting all of them.

But you could equally well mess up the data or the labels. Maybe better than 
```
ERROR: Values (5, 4, 6, 7, 2, 10, 9, 8, 3, 1...) are not in labels
```
would be something like this?
```
ERROR: Value 5 not found in labels "abcd"
```

TLipede: That's fair enough, maybe I'll use your initial suggestion then (just in case labels is a really large array/string).

mcabbott (approved):

darsnack (approved):

TLipede:
Cheers both!

PR #1892, opened 2022-02-28
Addition of Siamese Contrastive Loss function ( Updated )

arcAman07:
Addition of Siamese Contrastive Loss function. This is an Updated Pull request with all the reviewed changes added from the previous PR. Fixed the code formatting error which happened due to VS code so hence had to send a new PR ( previous forked repo was deleted ). Hopefully all necessary changes are added and this works well.

### PR Checklist

- Tests are added ( 
1)Test Case added when similar input and output is generated
2 )Test Case added when margin = 0
3) Contrastive loss function updated to throw a "DomainError" when margin is less than 0. Suitable test case added for that when user enters a margin value less than zero))

- Doc strings updated to show the difference in the output when input and ouputs are similar vs inputs and outputs are different.

- Updated in the docs/ also the loss function updated and exported.

### DIFF
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -528,7 +528,7 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
 end
 
 """
-    siamese_contrastive_loss(ŷ, y, margin = 1; agg = mean)
+    siamese_contrastive_loss(ŷ, y; margin = 1, agg = mean)
                                     
 Return the [contrastive loss](http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf)
 which can be useful for training Siamese Networks.
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -578,12 +578,7 @@ true
 function siamese_contrastive_loss(ŷ, y; agg = mean, margin::Real = 1)
     _check_sizes(ŷ, y)
     margin < 0 && throw(DomainError(margin, "Margin must be non-negative"))
-    max_val = max(sum(margin .- ŷ), 0)
-    if max_val > 0
-        return agg(@. y * (margin - ŷ)^2 + (1 - y) * ŷ^2)
-    else
-        return agg(@. y * (margin - ŷ)^2)
-    end
+    return agg(@. (1 - y) * ŷ^2 + y * max(0, margin - ŷ)^2)
 end
 
 ```@meta
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -531,9 +531,9 @@ end
     siamese_contrastive_loss(ŷ, y; margin = 1, agg = mean)
                                     
 Return the [contrastive loss](http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf)
-which can be useful for training Siamese Networks.
+which can be useful for training Siamese Networks. It is given by
                                     
-Formula: [agg(@. (1 - y) * ŷ^2 + y * max(0, margin - ŷ)^2)]                                    
+    agg(@. (1 - y) * ŷ^2 + y * max(0, margin - ŷ)^2)                           
                                  
 Specify `margin` to set the baseline for distance at which pairs are dissimilar.
                                     
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -577,15 +577,12 @@ true
 """
 function siamese_contrastive_loss(ŷ, y; agg = mean, margin::Real = 1)
     _check_sizes(ŷ, y)
-    if margin >= 0
-        max_val = max(-sum(ŷ .- margin), 0)
-        if max_val > 0
-            return agg(@. y * (margin - ŷ)^2 + (1 - y) * ŷ^2)
-        else
-            return agg(@. y * (margin - ŷ)^2)
-        end
+    margin < 0 && throw(DomainError(margin, "Margin must be non-negative"))
+    max_val = max(sum(margin .- ŷ), 0)
+    if max_val > 0
+        return agg(@. y * (margin - ŷ)^2 + (1 - y) * ŷ^2)
     else
-        throw(DomainError(margin, "Margin must be non-negative"))
+        return agg(@. y * (margin - ŷ)^2)
     end
 end
 
--- a/src/losses/functions.jl
+++ b/src/losses/functions.jl
@@ -537,45 +537,6 @@ which can be useful for training Siamese Networks. It is given by
                                  
 Specify `margin` to set the baseline for distance at which pairs are dissimilar.
                                     
-# Example
-```jldoctest
-julia> y = [1 0
-            0 0
-            0 1]
-3×2 Matrix{Int64}:
- 1  0
- 0  0
- 0  1									
-           
-julia> ŷ = [1 0
-            0 0
-            0 1]
-3×2 Matrix{Int64}:
- 1  0
- 0  0
- 0  1									
-                                    
-julia> Flux.siamese_contrastive_loss(ŷ, y) ≈ 0.0
-true
-julia> y1 = [1 0
-             0 0
-             0 1]
-3×2 Matrix{Int64}:
- 1  0
- 0  0
- 0  1									
-           
-julia> ŷ1 = [0.4 0.2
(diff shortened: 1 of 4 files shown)


src/losses/functions.jl
```diff
@@ -527,6 +527,68 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+    if margin >= 0
+        max_val = max(-sum(ŷ .- margin), 0)
+        if max_val > 0
+            return agg(@. y * (margin - ŷ)^2 + (1 - y) * ŷ^2)
+        else
+            return agg(@. y * (margin - ŷ)^2)
+        end
+    else
+        throw(DomainError(margin, "Margin must be non-negative"))
+    end
```
CarloLucibello, line 589:
```suggestion
    margin < 0 && throw(DomainError(margin, "Margin must be non-negative"))
    max_val = max(sum(margin .- ŷ), 0)
    if max_val > 0
        return agg(@. y * (margin - ŷ)^2 + (1 - y) * ŷ^2)
    else
        return agg(@. y * (margin - ŷ)^2)
    end
```

src/losses/functions.jl
```diff
@@ -527,6 +527,68 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+julia> ŷ1 = [0.4 0.2
+             0.5 0.5
+             0.1 0.3]
+3×2 Matrix{Float64}:
+ 0.4  0.2
+ 0.5  0.5
+ 0.1  0.3									
+                                    
+julia> Flux.siamese_contrastive_loss(ŷ1, y1) ≈ 0.2333333333333333
+true
```
CarloLucibello, line 575: these doc examples are not really informative, maybe we can remove them entirely. 
What would be useful is to have the formula in the docstring, as with any other loss function. 
Unfortunately, this formula is a bit more complex than the others. Maybe it can be written as
```julia
cond = sum(margin .- ŷ) > 0
agg(@. y * (margin - ŷ)^2 + cond * (1 - y) * ŷ^2)
```

CarloLucibello (approved):

src/losses/functions.jl
arcAman07: Will add the formula even though it might look a bit complex to the users.
Was told in the previous review to show contrastive loss for similar vs different inputs and outputs.( hence above example was added ).
Am fine with removing it and just adding the formula along with the information and use case of this loss function.

CarloLucibello:
This function assumes that distance is precomputed and given as ŷ. I think we should also have the version where inputs are given :
```julia
siamese_contrastive_loss(x1, x2, y) = siamese_contrastive_loss(sum(abs2, x1 - x2; dims=1), y)
```

src/losses/functions.jl
```diff
@@ -527,6 +527,65 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
...
+"""
+function siamese_contrastive_loss(ŷ, y; agg = mean, margin::Real = 1)
+    _check_sizes(ŷ, y)
+    margin < 0 && throw(DomainError(margin, "Margin must be non-negative"))
+    max_val = max(sum(margin .- ŷ), 0)
+    if max_val > 0
+        return agg(@. y * (margin - ŷ)^2 + (1 - y) * ŷ^2)
+    else
+        return agg(@. y * (margin - ŷ)^2)
+    end
```
CarloLucibello, line 586: This is much better:
```suggestion
    return agg(@. (1 - y) * ŷ^2 + y * max(0, margin - ŷ)^2)
```

CarloLucibello: And actually, the previous formulation was wrong, you don't have to compute the max over a sum

test/losses.jl
```diff
@@ -210,3 +210,29 @@ end
...
+@testset "siamese_contrastive_loss" begin
+  y = [1 0
+       0 0
+       0 1]
+  ŷ = [0.4 0.2
+       0.5 0.5
+       0.1 0.3]
+  y1 = [1 0 0 0 1
+        0 1 0 1 0
+        0 0 1 0 0]
```
CarloLucibello, line 223: Is there any reason why these are matrices instead of vectors (one distance and one label for each example in a batch)?

src/losses/functions.jl
CarloLucibello: Ok. Just notice that your previous formula was wrong, the correct one is 
```julia
agg(@. (1 - y) * ŷ^2 + y * max(0, margin - ŷ)^2)
```

test/losses.jl
arcAman07: Plan of using matrices was to maybe provide a bit more complex testing example for the contrastive loss function even though it is not realistic in deep learning models.
I can replace these matrices test example with vectors ( as it is generally for each input in a batch ) if it needs to be or I can even add seperate vector examples along with these matrices examples.

src/losses/functions.jl
arcAman07: Yeah noted, it was an error on my part, will update the siamese_contrastive_loss function with the formula given above ( with the suggested change you had given ) and will check the corresponding tests again.

arcAman07:
> This function assumes that distance is precomputed and given as ŷ. I think we should also have the version where inputs are given :
> 
> ```julia
> euclidean_distance(x1::AbstractMatrix, x2::AbstractMatrix) = sqrt.(sum(abs2, x1 - x2; dims=1)) |> vec
> 
> siamese_contrastive_loss(x1::AbstractMatrix, x2::AbstractMatrix, y::AbstractVector) =
>     siamese_contrastive_loss(euclidean_distance(x1, x2), y)
> 
> function siamese_contrastive_loss(ŷ::AbstractVector, y::AbstractVector)
>     ...
> end
> ```

Had plans of adding this where inputs be the normal ones rather than precomputed euclidean distance vectors. Then thought there are packages where user can easily compute these euclidean distance and then use this contrastive loss function.  Also wasn't sure to add this in the initial PR as it could be a bit complex and sort of heavy. Wanted to check whether the initial function is well received then thought of maybe adding this feature too. Am fine with adding this in this PR or maybe in a future PR.

arcAman07:
1) Loss function corrected
2) Added vector test cases => to show the application of this loss function.
3) Let me know whether even the docstring example needs to be revamped with the vector test case example or that should be added to the docstring also.
4) Would want to the function when the inputs can be the normal features rather than euclidean distances after calculation ( most probably in another PR to avoid this PR being too heavy and complex, it will be a bit easier for me ). Am fine with adding that to even this too.

src/losses/functions.jl
CarloLucibello: let's remove all the examples entirely, they don't say much.

src/losses/functions.jl
```diff
@@ -527,6 +527,62 @@ function focal_loss(ŷ, y; dims=1, agg=mean, γ=2, ϵ=epseltype(ŷ))
     agg(sum(@. -y * (1 - ŷ)^γ * log(ŷ); dims=dims))
 end
 
+"""
+    siamese_contrastive_loss(ŷ, y; margin = 1, agg = mean)
+                                    
+Return the [contrastive loss](http://yann.lecun.com/exdb/publis/pdf/hadsell-chopra-lecun-06.pdf)
+which can be useful for training Siamese Networks.
+                                    
+Formula: [agg(@. (1 - y) * ŷ^2 + y * max(0, margin - ŷ)^2)]
```
CarloLucibello, line 536:
```suggestion
which can be useful for training Siamese Networks. It is given by
                                    
    agg(@. (1 - y) * ŷ^2 + y * max(0, margin - ŷ)^2)                           
```

src/losses/functions.jl
arcAman07: Cool will do that, does those examples need to be replaced with anything currently. Or after some time we will add much more suitable examples to the docs?

arcAman07: Examples removed from the docs.

CarloLucibello (approved):

CarloLucibello:
thank you for the contribution!

arcAman07:
> thank you for the contribution!

Learnt a lot while contributing, hoping to contribute more in the future ( especially with the implementation of the precomputed euclidean distance vectors as inputs for the loss function )

PR #1896, opened 2022-03-02
Updation to the Siamese Contrastive Loss function

arcAman07:
Addition of another generic method attached to the siamese contrastive Loss function. In the previous PR, the user would have to give the euclidean distance vectors to the loss function to calculate it's loss ( euclidean distance would have to be calculated first ). After discussions during the review of the previous PR, it was agreed that it would be easier for the user just to give the matrix inputs and let the loss function itself calculate the euclidean distance between the inputs and then also the corresponding loss. 
1) Euclidean distance function added ( exported as well for testing )
2) Another generic method siamese_contrastive_loss function added to calculate the loss
3) Variety of test cases added to understand how this method works, dependency on the margin value,etc

As this was a generic method to the same function, was having trouble on how to write the docs for it and how should we exactly structure it so that user can properly understand and use it. Would need some help for that, while rest of the functionalities are added.

### PR Checklist

- [ ✓] Tests are added
- [ ] Entry in NEWS.md


arcAman07:
Cam anybody please help me out regarding the error which is coming and how to solve it. It's happening in no spurious promotions and it's probably related to float datatype.

Issue #1860, opened 2022-02-04
Is the two-arg update!(x, d) ever called?


cossio:
This method:

https://github.com/FluxML/Flux.jl/blob/d0a5b776308a33085f876f09a40781b0528d716c/src/optimise/train.jl#L9

is never called. It's also not tested.
Only the three-arg method is used.

Do we need this two-arg method of this function here?

PR #1902, opened 2022-03-08
RFC: Restrict `train!` to `AbstractOptimiser`

mcabbott:
Flux's optimisers all (except 1 mistake?) have a supertype. So I think we can restrict `train!` and friends to demand this. 

The reason to do so is that I think it makes it possible to add a `train!` method which accepts the ones from Optimisers.jl (which have no supertype), later. And have them coexist within one version of Flux.

I guess this counts as breaking, but are there many other optimisers without the supertype in the wild? Will it break anything else?

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -1,5 +1,4 @@
 name: Downstream
-
 on:
   push:
     branches: [master]
@@ -10,6 +9,8 @@ jobs:
   test:
     name: ${{ matrix.package.repo }}/${{ matrix.package.group }}
     runs-on: ${{ matrix.os }}
+    env:
+      GROUP: ${{ matrix.package.group }}
     strategy:
       fail-fast: false
       matrix:
@@ -27,7 +28,7 @@ jobs:
           - {user: SciML, repo: DiffEqFlux.jl, group: Layers}
           - {user: SciML, repo: NeuralPDE.jl, group: NNPDE}
           - {user: SciML, repo: OperatorLearning.jl, group: All}
-    if: contains(github.event.pull_request.labels.*.name, 'run downstream test')
+    # if: contains(github.event.pull_request.labels.*.name, 'run downstream test')
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -28,7 +28,7 @@ jobs:
           - {user: SciML, repo: DiffEqFlux.jl, group: Layers}
           - {user: SciML, repo: NeuralPDE.jl, group: NNPDE}
           - {user: SciML, repo: OperatorLearning.jl, group: All}
-    # if: contains(github.event.pull_request.labels.*.name, 'run downstream test')
+    if: contains(github.event.pull_request.labels.*.name, 'run downstream test')
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -553,7 +553,7 @@ mutable struct Optimiser <: AbstractOptimiser
   os::Vector{Any}
 end
 
-Optimiser(o...) = Optimiser(Any[o...])
+Optimiser(opts::AbstractOptimiser...) = Optimiser(Any[opts...])
 
 @forward Optimiser.os Base.getindex, Base.first, Base.last, Base.lastindex, Base.push!, Base.setindex!
 @forward Optimiser.os Base.iterate
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -521,7 +521,7 @@ opt = AdaBelief()
 opt = AdaBelief(0.001, (0.9, 0.8))
 ```
 """
-mutable struct AdaBelief
+mutable struct AdaBelief <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
   epsilon::Float64
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -93,8 +93,7 @@ The callback can call [`Flux.stop`](@ref) to interrupt the training loop.
 
 Multiple optimisers and callbacks can be passed to `opt` and `cb` as arrays.
 """
-function train!(loss, ps, data, opt::AbstractOptimiser; cb = () -> ())
-  ps = Params(ps)
+function train!(loss, ps::Params, data, opt::AbstractOptimiser; cb = () -> ())
   cb = runall(cb)
   n = (Base.IteratorSize(typeof(data)) == Base.HasLength()) ? length(data) : 0
   @withprogress for (i, d) in enumerate(data)
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -1,14 +1,6 @@
 using ProgressLogging: @progress, @withprogress, @logprogress
 import Zygote: Params, gradient
 
-"""
-    update!(x, x̄)
-
-Update the array `x` according to `x .-= x̄`.
-"""
-function update!(x::AbstractArray, x̄)
-  x .-= x̄
-end
 
 """
     update!(opt, p, g)
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -73,25 +73,42 @@ batchmemaybe(x) = tuple(x)
 batchmemaybe(x::Tuple) = x
 
 """
-    train!(loss, params, data, opt; cb)
+    train!(loss, pars::Params, data, opt::AbstractOptimiser; [cb])
         
-`train!` uses a `loss` function and training `data` to improve the 
-[Model parameters](@ref) (`params`) based on a pluggable [Optimisers](@ref) (`opt`).
-        
-For each datapoint `d` in `data`, compute the gradient of  `loss` with
-respect to `params` through backpropagation and call the optimizer `opt`.
-If `d` is a tuple of arguments to `loss` call `loss(d...)`, else call `loss(d)`.
-        
-To pass trainable parameters, call [`Flux.params`](@ref) with your model or just the 
-layers you want to train, like `train!(loss, params(model), ...)` or `train!(loss, params(model[1:end-2), ...)` respectively.
+Uses a `loss` function and training `data` to improve the 
+model's parameters according to a particular optimisation rule `opt`.
 
-[Callbacks](@ref) are given with the keyword argument `cb`. For example, this will print "training" 
-every 10 seconds (using [`Flux.throttle`](@ref)):
-`train!(loss, params, data, opt, cb = throttle(() -> println("training"), 10))`
-        
+For each `d in data`, first the gradient of the `loss` is computed like this:
+```
+    gradient(() -> loss(d...), pars)  # if d isa Tuple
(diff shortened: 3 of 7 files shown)


src/optimise/train.jl
```diff
@@ -101,7 +93,7 @@ The callback can call [`Flux.stop`](@ref) to interrupt the training loop.
 
 Multiple optimisers and callbacks can be passed to `opt` and `cb` as arrays.
 """
-function train!(loss, ps, data, opt; cb = () -> ())
+function train!(loss, ps, data, opt::AbstractOptimiser; cb = () -> ())
```
mcabbott, line 96: Further restricting to `ps::Params` gives errors:
```
Training Loop: Error During Test at /Users/me/.julia/dev/Flux/test/optimise.jl:48
  MethodError: no method matching train!(::var"#57#63", ::Tuple{}, ::Base.Iterators.Take{Base.Iterators.Repeated{Tuple{}}}, ::Descent)
  Closest candidates are:
    train!(::Any, ::Params, ::Any, ::Flux.Optimise.AbstractOptimiser; cb) at ~/.julia/dev/Flux/src/optimise/train.jl:105
  Stacktrace:
...
DataLoader: Error During Test at /Users/me/.julia/dev/Flux/test/data.jl:3
  MethodError: no method matching train!(::var"#loss#93", ::Vector{Vector{Float64}}, ::IterTools.NCycle{DataLoader{Matrix{Float64}, Random._GLOBAL_RNG}}, ::Descent)
  Closest candidates are:
    train!(::Any, ::Params, ::Any, ::Flux.Optimise.AbstractOptimiser; cb) at ~/.julia/dev/Flux/src/optimise/train.jl:105
  Stacktrace:
```

src/optimise/train.jl
```diff
@@ -1,14 +1,6 @@
 using ProgressLogging: @progress, @withprogress, @logprogress
 import Zygote: Params, gradient
 
-"""
-    update!(x, x̄)
-
-Update the array `x` according to `x .-= x̄`.
-"""
-function update!(x::AbstractArray, x̄)
-  x .-= x̄
```
mcabbott, line 10: As noted by https://github.com/FluxML/Flux.jl/issues/1860, this function seems never to be used. 

Maybe someone else uses it, and it deserves a depwarn?

src/optimise/optimisers.jl
```diff
@@ -521,7 +521,7 @@ opt = AdaBelief()
 opt = AdaBelief(0.001, (0.9, 0.8))
 ```
 """
-mutable struct AdaBelief
+mutable struct AdaBelief <: AbstractOptimiser
```
mcabbott, line 524: I presume this was an oversight, all others have supertype.

src/optimise/train.jl
ToucheSir: Looking at the blame it's hard to say how intentional the use of `()` in https://github.com/FluxML/Flux.jl/blob/master/test/optimise.jl#L51-L74 was, but I'd be fine with updating the tests to match.

.github/workflows/Downstream.yml
```diff
@@ -27,7 +28,7 @@ jobs:
           - {user: SciML, repo: DiffEqFlux.jl, group: Layers}
           - {user: SciML, repo: NeuralPDE.jl, group: NNPDE}
           - {user: SciML, repo: OperatorLearning.jl, group: All}
-    if: contains(github.event.pull_request.labels.*.name, 'run downstream test')
+    # if: contains(github.event.pull_request.labels.*.name, 'run downstream test')
```
ToucheSir, line 31: If I'm reading https://github.com/FluxML/Flux.jl/actions/runs/1948923390 correctly, the downstream tests only take 2-3 minutes each? If so, we could probably get away with either removing this or inverting the condition to skip them.

src/optimise/train.jl
mcabbott: Ok. Thinking aloud a bit, the idea is that all of these can then co-exist:

* `train!(loss, ps::Params, data, opt::AbstractOptimiser)` as before
* `train!(loss, ps::Params, data, opt)` would use implicit parameters + Optimisers.jl optimisers. I haven't written it but I think it wouldn't be so hard to store their states in an IdDict. That would let us delete Flux's optimisers completely, and many use cases would not need to change anything. Still calls  `gradient(() -> loss(x, y), ps)`.
* `train!(loss, model, data, opt)` would call `gradient(m -> loss(m, x, y), model)` and be fully structural. Only works with new optimisers of course.

Maybe that's too many steps and clearer to just do the first and last. But the more restrictive we are in the initial 0.13.0, the more room we have to decide.

.github/workflows/Downstream.yml
mcabbott: I see, you mean that the purpose of this line was to restrict when these are run? 

I just blindly copied what NNlib does, as there (and on this PR) they show up nicely in the checks. I'm not certain whether they run less often before this PR. But as you say they don't take hours.

ToucheSir: They didn't run at all without the label I just added (in this case, for demonstration purposes only).

darsnack: No they are only taking a short time because of version incompatibility (the downstream testing doesn't run on breaking releases)

mcabbott: I set the version to 0.12.99 here, temporarily, so that they would run.

src/optimise/train.jl
mcabbott: The Optimisers.jl method could look like this:
```
function train!(loss, model, data, opt; cb = [])
  opt isa AbstractOptimiser && error("old-style AbstractOptimiser can only be used with Params")
  model isa Params && error("implicit Params can only be used with old-style AbstractOptimiser")
  cb = runall(cb)
  state = Optimisers.setup(opt, model)
  @withprogress for (i, d) in enumerate(data)
    try
      grad = gradient(model) do m
        loss(m, batchmemaybe(d)...)
      end
      state, model = update!(state, model, grad...)
      cb()
    catch ex
      # ...
    end
    @logprogress i / length_or_zero(data)
  end
  model
end
```
The downside of this is that `state` is thrown away, so `@epochs 100 train!(...)` will reset each time. Whereas the present method mutates `opt` and stores it there.

darsnack: Hmm then we'd lose momentum info, etc. over epochs. What about the interface being `train!(loss, model, data, opt, state = Optimisers.setup(opt, model))`?

mcabbott: Could do. But 5 positional arguments is a lot. 

You never need `opt` again once the setup is done, but you do need to pass keep it around. 

Maybe it's less confusing not to replace `train!`, and instead have something like:
```julia
s = setup(m, Descent())
for d in data
    s, m = step!(loss, s, m, d...)   # calls gradient(m -> loss(m, d...), m)
end
```

mcabbott: Another idea:
```julia
opt = fluxoptwrap(Optimisers.Momentum(), model)  # makes mutable <: AbstractOptimiser

train!(loss, model, data, opt)          # calls gradient(m -> loss(m, d...), model)

# perhaps also:
train!(loss, params(model), data, opt)  # calls gradient(() -> loss(d...), ps)
```

src/optimise/optimisers.jl
```diff
@@ -553,7 +553,7 @@ mutable struct Optimiser <: AbstractOptimiser
   os::Vector{Any}
 end
 
-Optimiser(o...) = Optimiser(Any[o...])
+Optimiser(opts::AbstractOptimiser...) = Optimiser(Any[opts...])
```
mcabbott, line 556: This chain thing calls `apply!(opt, x, Δ)` on each element, which only accepts the various AbstractOptimiser types. So it should be safe to restrict --- and doing so may avoid some mistakes with Optimisers.OptimiserChain.

src/optimise/train.jl
```diff
@@ -81,28 +73,44 @@ batchmemaybe(x) = tuple(x)
...
+[Callbacks](@ref) are given with the keyword argument `cb`.
+For example, this will print "training" every 10 seconds (using [`Flux.throttle`](@ref)):
+```
+    train!(loss, params, data, opt, cb = throttle(() -> println("training"), 10))
+```
+    
 The callback can call [`Flux.stop`](@ref) to interrupt the training loop.
 
-Multiple optimisers and callbacks can be passed to `opt` and `cb` as arrays.
+Multiple callbacks can be passed to `cb` as array.
```
mcabbott, line 111: There seem to be no tests of this "Multiple optimisers ...`opt` ... as array". Presumably you should really use `Optimiser(o...)` for this?

ToucheSir: Indeed, the [backing code](https://github.com/FluxML/Flux.jl/commit/97244e0a68fa8cbae17f8065160126897a674009) has long since disappeared.

src/optimise/optimisers.jl
ToucheSir: Is the splat still necessary then?

src/optimise/train.jl
ToucheSir: It is [documented](https://fluxml.ai/Flux.jl/stable/training/optimisers/#Flux.Optimise.update!), so a depwarn unless we're considering this a breaking change for 0.13.

mcabbott: I didn't see that, thanks.

This whole PR is certainly breaking. Do you think it's worth bothering with a deprecation for this?

src/optimise/optimisers.jl
mcabbott: Didn't think about it. The type specifies Vector{Any}. It forwards `Base.push!, Base.setindex!`. I could try removing these to store a tuple, see if anything breaks?

ToucheSir: In that case I don't think it's worth the hassle. With any luck we'll be removing it soon.

src/optimise/train.jl
ToucheSir: What would be an example of an immutable rule? I'm trying to work out why the wrapper would be necessary.

mcabbott: The rules from Optimisers.jl are all immutable.

ToucheSir: The rules are, but the state they're manipulating shouldn't be? Unless you were considering not returning the optimizer state from `train!` and keeping it in the wrapper instead.

mcabbott: Yes exactly. The `fluxoptwrap` idea is to let you use `train!` mutating both model and optimiser state, with the Optimisers.jl stuff hidden away. Changing existing code to use this would be a much smaller change than using Optimisers directly.

I'm not certain that's a good idea, and it isn't in this PR. But it's one path we could take.

ToucheSir: Let's not worry about the Optimisers.jl integration for now then.

src/optimise/train.jl
ToucheSir: I think it would be a reasonably small change and consistent with the other 0.13 deprecations we've added.

.github/workflows/Downstream.yml
mcabbott: Reverted

mcabbott:
Maybe this is ready to go?

src/optimise/train.jl
ToucheSir: If you wouldn't mind adding this, everything else LGTM.

mcabbott: Sure, done.

ToucheSir (approved):

PR #1908, opened 2022-03-20
Mark initialisations nograd, restrict signatures

mcabbott:
Inspired by this:
https://discourse.julialang.org/t/mutating-arrays-not-supported-toy-problem/78057
this PR adds `@non_differentiable` to all initialisation functions.

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -81,6 +81,7 @@ end
       init = (args...) -> sparse_init(args...; sparsity=0.5)
     else
       @test size(init(3)) == (3,)
+      @test size(init(3, 4, 5)) == (3, 4, 5)
     end
     @test size(init(3, 4)) == (3, 4)
     @test_skip size(init((3, 4, 5))) == (3, 4, 5)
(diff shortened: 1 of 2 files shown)


test/utils.jl
```diff
@@ -69,6 +69,34 @@ end
...
+      truncated_normal,
+      identity_init,
+    ]
+    if init == sparse_init
+      init = (args...) -> sparse_init(args...; sparsity=0.5)
+    else
+      @test size(init(3)) == (3,)
+    end
+    @test size(init(3, 4)) == (3, 4)
+    @test_skip size(init((3, 4, 5))) == (3, 4, 5)
```
mcabbott, line 86: Not all of them accept a size tuple like this. Should they?

Or perhaps they should all only accept `init(size::Integer...)`?

ToucheSir: +1 for the latter.

ToucheSir:
Is there a use case for having init functions succeed in a gradient context? Could we throw a more informative error in the rrule instead?

mcabbott:
I can't think of a use. But if they are variants of `rand` / `randn`, then nograd seems sensible:
```
julia> gradient(x -> sum(rand(3) .* x), 4)
(2.438304105970437,)

julia> gradient(x -> sum(rand(3) .* x), 4)
(1.9420922421421238,)
```

test/utils.jl
mcabbott: Done in [9d1f5d5](https://github.com/FluxML/Flux.jl/pull/1908/commits/9d1f5d5ed976f299b43b719894fed3b563509bc9) 

I guess that makes this breaking, hence v0.13

CarloLucibello (approved):

PR #1910, opened 2022-03-20
Add `MLUtils` to doc dependencies and fix missing `DataLoader` docstring

### LINKED ISSUE #1909: Missing docstring for `Flux.Data.Dataloader`
The `dev` documentation shows a missing docstring error here - https://fluxml.ai/Flux.jl/dev/data/dataloader/

image

`DataLoader` was replaced by the `DatLoader` from `MLUtils` here - https://github.com/FluxML/Flux.jl/pull/1874

Should the documentation say something like - `Flux.jl` now uses `DataLoader` from `MLUtils`.  For more documentation on `DataLoader`, refer [here](https://juliaml.github.io/MLUtils.jl/dev/api/#MLUtils.DataLoader). Or should I manually add the documentation in the .`md` file?

Saransh-cpp:
Fixes #1909 

Fixed the missing docstring by adding a new doc dependency (`MLUtils`)!

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,7 +1,7 @@
-using Documenter, Flux, NNlib, Functors
+using Documenter, Flux, NNlib, Functors, MLUtils
 
 DocMeta.setdocmeta!(Flux, :DocTestSetup, :(using Flux); recursive = true)
-makedocs(modules = [Flux, NNlib, Functors],
+makedocs(modules = [Flux, NNlib, Functors, MLUtils],
          doctest = VERSION == v"1.5",
          sitename = "Flux",
          pages = ["Home" => "index.md",
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -2,6 +2,7 @@
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 NNlib = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
+MLUtils = "f1d291b0-491e-4a28-83b9-f70985020b54"
 
 [compat]
 Documenter = "0.26"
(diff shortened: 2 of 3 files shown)


docs/make.jl
```diff
@@ -1,7 +1,7 @@
-using Documenter, Flux, NNlib, Functors
+using Documenter, Flux, NNlib, Functors, MLUtils
 
 DocMeta.setdocmeta!(Flux, :DocTestSetup, :(using Flux); recursive = true)
-makedocs(modules = [Flux, NNlib, Functors],
+makedocs(modules = [Flux, NNlib, Functors, MLUtils],
          doctest = VERSION == v"1.5",
          sitename = "Flux",
          pages = ["Home" => "index.md",
```
mcabbott, line 7: Possibly the page `"DataLoader" => "data/dataloader.md"` should be replaced with something more like that for "NNlib" & "Functors"? Covering also the functions now in utilities.md, since they now live in MLUtils.

Saransh-cpp: Yes, that would make it much more organized, but where should I add the new `"MLUtils"` page? Given that it will now contain some utility functions, as well as, `"DataLoader"`.

mcabbott: Maybe the same place, under "Handling Data", still makes sense? One-hot encoding will (I think) soon be a separate package too.

Saransh-cpp: Added a new page for `MLUtils`!

docs/src/models/nnlib.md
```diff
@@ -1,4 +1,4 @@
-# NNlib
+# NNlib.jl
```
mcabbott, line 1: Notice this, on master, for Functors but not for NNlib:

Screenshot 2022-03-21 at 11 21 12

So it seems to show another heading if the H1 doesn't match the page title? 

Maybe these page titles & headings deserve a re-think. Something like "Activations etc. from NNlib.jl", "Nested structure using Functors.jl" might be better than just the package names (as these don't help much if you don't already know them) but won't fit in the sidebar.

mcabbott (approved):

docs/src/models/nnlib.md
mcabbott: Ah that must be it. Sure, making them all H1 sounds like a good idea.

PR #1911, opened 2022-03-20
Improvements for LayerNorm

theabhirath:
Inspired by [this](https://discourse.julialang.org/t/flux-layernorm-slower-than-pytorch/78084) discussion, this PR achieves some speedup in LayerNorm by a. passing the mean to `std` as an argument (there was a note to make this change after it had been implemented but it never really happened) and b. tweaking some mild implementation details (it removes a broadcasted `identity` that was proving to be a bottleneck for large array sizes, for example).

Before this PR (Julia 1.8.0-beta2.3, Flux master):
```julia
julia> l = LayerNorm(768);

julia> x = rand(Float32, 768, 128, 2);

julia> @benchmark $(l)($x)
 Range (min … max):  200.292 μs …  10.650 ms  ┊ GC (min … max):  0.00% … 96.39%
 Time  (median):     311.625 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   412.579 μs ± 625.478 μs  ┊ GC (mean ± σ):  23.83% ± 14.70%

  ▃█▄▁                                                          ▁
  ████▇▄▄▄▁▄▁▃▁▃▁▁▁▁▁▁▁▁▃▁▄▁▃▁▄▃▁▁▄▁▄▅▃▄▅▄▆▆▄▅▅▅▆▆▅▄▃▅▅▅▄▅▄▅▄▄▅ █
  200 μs        Histogram: log(frequency) by time       4.42 ms <

 Memory estimate: 2.25 MiB, allocs estimate: 24.
```

After this PR:
```julia
julia> @benchmark $(l)($x)
 Range (min … max):  167.875 μs …   5.326 ms  ┊ GC (min … max):  0.00% … 95.32%
 Time  (median):     243.792 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   282.340 μs ± 341.526 μs  ┊ GC (mean ± σ):  14.86% ± 11.55%

    █                                                            
  ▃▆█▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▁▁▂▂▂▁▂▂▂▂▂▂▂▂▁▁▁▂▂▂▂ ▂
  168 μs           Histogram: frequency by time         2.39 ms <

 Memory estimate: 1.50 MiB, allocs estimate: 16.
```

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -178,7 +178,7 @@ end
 
 function Base.show(io::IO, l::LayerNorm)
   print(io, "LayerNorm($(l.size)")
-  a.λ === identity || print(io, ", $(l.λ)")
+  l.λ === identity || print(io, ", ", l.λ)
   hasaffine(l) || print(io, ", affine=false")
   print(io, ")")
 end
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -164,7 +164,7 @@ struct LayerNorm{F,D,T,N}
   affine::Bool
 end
 
-function LayerNorm(sz, λ=nothing; affine=true, ϵ=1f-5)
+function LayerNorm(sz, λ=identity; affine=true, ϵ=1f-5)
   sz = sz isa Integer ? (sz,) : sz
   diag = affine ? Diagonal(sz...) : identity
   return LayerNorm(λ, diag, ϵ, sz, affine)
@@ -174,12 +174,12 @@ end
 
 function (a::LayerNorm)(x)
   x = a.diag(normalise(x, dims=1:length(a.size), ϵ=a.ϵ))
-  return isnothing(a.λ) ? x : a.λ.(x)
+  return a.λ === identity ? x : a.λ.(x)
 end
 
 function Base.show(io::IO, l::LayerNorm)
   print(io, "LayerNorm($(l.size)")
-  isnothing(l.λ) || print(io, ", $(l.λ)")
+  a.λ === identity || print(io, ", $(l.λ)")
   hasaffine(l) || print(io, ", affine=false")
   print(io, ")")
 end
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -139,7 +139,7 @@ testmode!(m::AlphaDropout, mode=true) =
   (m.active = (isnothing(mode) || mode == :auto) ? nothing : !mode; m)
 
 """
-    LayerNorm(sz, λ=identity; affine=true, ϵ=1fe-5)
+    LayerNorm(sz, λ=nothing; affine=true, ϵ=1fe-5)
 
 A [normalisation layer](https://arxiv.org/abs/1607.06450) designed to be
 used with recurrent hidden states.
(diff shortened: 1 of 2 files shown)


src/layers/normalise.jl
```diff
@@ -164,22 +164,22 @@ struct LayerNorm{F,D,T,N}
...
 end
 
 @functor LayerNorm
 
 function (a::LayerNorm)(x)
-  x = normalise(x, dims=1:length(a.size), ϵ=a.ϵ)
-  a.diag === nothing ? a.λ.(x) : a.λ.(a.diag(x))
+  x = a.diag(normalise(x, dims=1:length(a.size), ϵ=a.ϵ))
+  return isnothing(a.λ) ? x : a.λ.(x)
 end
```
mcabbott, line 178: I like storing `diag = identity`, that helps clarity. But don't like storing `λ = nothing`, it could equally well test `λ === identity` instead of `isnothing`.

Note that in a gradient, Zygote will remove the broadcast identity.

A bigger change, but it's possible that instead of storing the activation function separately, it should become part of Diagonal?

theabhirath: > Note that in a gradient, Zygote will remove the broadcast identity.

Yep this was more to reduce the time taken by the forward pass

> A bigger change, but it's possible that instead of storing the activation function separately, it should become part of Diagonal?

That will mean that we can't have an activation function without an affine transform, won't it?

mcabbott: It could store something like `diag = Base.Fix1(broadcast, λ)` for the case affine=false.

src/layers/normalise.jl
```diff
@@ -139,7 +139,7 @@ testmode!(m::AlphaDropout, mode=true) =
   (m.active = (isnothing(mode) || mode == :auto) ? nothing : !mode; m)
 
 """
-    LayerNorm(sz, λ=identity; affine=true, ϵ=1fe-5)
+    LayerNorm(sz, λ=nothing; affine=true, ϵ=1fe-5)
```
mcabbott, line 142: Revert to match code:
```suggestion
    LayerNorm(sz, λ=identity; affine=true, ϵ=1fe-5)
```

src/layers/normalise.jl
```diff
@@ -166,20 +166,20 @@ end
 
 function LayerNorm(sz, λ=identity; affine=true, ϵ=1f-5)
   sz = sz isa Integer ? (sz,) : sz
```
mcabbott, line 168: Tidying up?
```suggestion
function LayerNorm(sz, λ=identity; affine::Bool=true, ϵ::Real=1f-5)
```
Also, do we really have a public keyword called `ϵ`? I thought they were supposed to be ascii...

I guess "`sz` should be an integer or a tuple of integers." is a bit ugly too, can't it just be integers?

src/layers/normalise.jl
```diff
@@ -166,20 +166,20 @@ end
...
-  diag = affine ? Diagonal(sz...) : nothing
+  diag = affine ? Diagonal(sz...) : identity
   return LayerNorm(λ, diag, ϵ, sz, affine)
```
mcabbott, line 170:
```suggestion
  return LayerNorm(λ, diag, ϵ, Tuple(sz), affine)
```

src/layers/normalise.jl
```diff
@@ -164,22 +164,21 @@ struct LayerNorm{F,D,T,N}
...
-  x = normalise(x, dims=1:length(a.size), ϵ=a.ϵ)
-  a.diag === nothing ? a.λ.(x) : a.λ.(a.diag(x))
+  x = a.diag(normalise(x, dims=1:length(a.size), ϵ=a.ϵ))
+  return a.λ === identity ? x : a.λ.(x)
 end
 
 function Base.show(io::IO, l::LayerNorm)
   print(io, "LayerNorm($(l.size)")
-  l.λ == identity || print(io, ", $(l.λ)")
+  a.λ === identity || print(io, ", $(l.λ)")
```
mcabbott, line 181:
```suggestion
  l.λ === identity || print(io, ", $(l.λ)")
```

mcabbott (approved):
Looks good, it's a step forwards.

Remaining questions are (1) should Diagonal take an activation function? (2) can we kill the ε (no, ϵ) keyword in favour of ascii? For all normalisation layers, (3) Can they have more useful docstring examples? It seems the printing may not be tested right now. 

These need not be this PR. Perhaps (2) is worth considering for v0.13?

mcabbott (approved):

PR #1912, opened 2022-03-20
Improve docs for initialisation

mcabbott:
This tidies up the docs for initialisation functions:
* It removes the incomplete "see also" lists, but completes the list in the docs
* It removes some examples which printed random numbers, in favour of something interpretable
* It documents the curried forms
* It documents `zeros32` and `randn32` etc.

It also adds a few methods which seemed to be missing, like `randn32(rng, size...)`, and `glorot_normal(; gain=2)`.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -164,7 +164,7 @@ julia> round.(extrema(Flux.kaiming_uniform(100, 100)), digits=3)
 
 [1] He, Kaiming, et al. "Delving deep into rectifiers: Surpassing human-level performance on imagenet classification." _Proceedings of the IEEE international conference on computer vision_. 2015.
 """
-function kaiming_uniform(rng::AbstractRNG, dims::Integer...; gain = √2)
+function kaiming_uniform(rng::AbstractRNG, dims::Integer...; gain::Real = √2)
   bound = Float32(√3 * gain / sqrt(first(nfan(dims...)))) # fan_in
   return (rand(rng, Float32, dims...) .- 0.5f0) .* 2bound
 end
@@ -201,7 +201,7 @@ julia> round(std(Flux.kaiming_normal(1000, 1000)), digits=3)
 
 [1] He, Kaiming, et al. "Delving deep into rectifiers: Surpassing human-level performance on imagenet classification." _Proceedings of the IEEE international conference on computer vision_. 2015.
 """
-function kaiming_normal(rng::AbstractRNG, dims::Integer...; gain = √2f0)
+function kaiming_normal(rng::AbstractRNG, dims::Integer...; gain::Real = √2f0)
   std = Float32(gain / sqrt(first(nfan(dims...)))) # fan_in
   return randn(rng, Float32, dims...) .* std
 end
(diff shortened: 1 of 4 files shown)


src/utils.jl
```diff
@@ -223,11 +226,6 @@ julia> round.(extrema(Flux.truncated_normal(10^6)); digits=3)
 julia> round(std(Flux.truncated_normal(10^6; lo = -100, hi = 100)))
 1.0f0
 ```
-
-# References
-[1] Burkardt, John. "The Truncated Normal Distribution" 
-[PDF](https://people.sc.fsu.edu/~jburkardt/presentations/truncated_normal.pdf).
```
mcabbott, line 229: I removed this reference as (1) truncating a normal seems to standard to really need one, and (2) it describes how to sample by drawing and rejecting, which isn't what's implemented here.

src/utils.jl
```diff
@@ -53,98 +52,104 @@ else
...
+
+julia> std(Flux.glorot_normal(100, 100))
+0.10008307f0
+```
 
 # References
 
 [1] Glorot, Xavier, and Yoshua Bengio. "Understanding the difficulty of training deep feedforward neural networks." _Proceedings of the thirteenth international conference on artificial intelligence and statistics_. 2010.
 """
-glorot_normal(rng::AbstractRNG, dims::Integer...) = randn(rng, Float32, dims...) .* sqrt(2.0f0 / sum(nfan(dims...)))
```
mcabbott, line 118: While `kaiming_normal` takes a `gain` keyword, as do several others, this did not. I assume there's no strong reason not to do so.

src/utils.jl
```diff
@@ -315,34 +307,36 @@ function orthogonal(rng::AbstractRNG, d1::Integer, ds::Integer...; kwargs...)
...
 
 ChainRulesCore.@non_differentiable orthogonal(::Any...)
 
 """
-    sparse_init([rng=GLOBAL_RNG], dims...; sparsity, std = 0.01)
+    sparse_init([rng=GLOBAL_RNG], rows, cols; sparsity, std = 0.01) -> Array
+    sparse_init([rng]; kw...) -> Function
 
-Return an `Array` of size `dims` where each column contains a fixed fraction of
+Return a `Matrix{Float32}` of size `rows, cols` where each column contains a fixed fraction of
```
mcabbott, line 318: `sparse_init` is the only one which doesn't make arbitrary-size arrays, previously undocumented.

src/utils.jl
```diff
@@ -361,71 +355,64 @@ function sparse_init(rng::AbstractRNG, dims::Integer...; sparsity, std = 0.01)
...
 
 sparse_init(dims::Integer...; kwargs...) = sparse_init(rng_from_array(), dims...; kwargs...)
-sparse_init(rng::AbstractRNG; init_kwargs...) = (dims...; kwargs...) -> sparse_init(rng, dims...; init_kwargs..., kwargs...)
+sparse_init(rng::AbstractRNG=rng_from_array(); init_kwargs...) = (dims...; kwargs...) -> sparse_init(rng, dims...; init_kwargs..., kwargs...)
 
 ChainRulesCore.@non_differentiable sparse_init(::Any...)
 
 """
-    identity_init([rng=GLOBAL_RNG], dims...; gain=1, shift=0)
+    identity_init(size...; gain=1, shift=0) -> Array
```
mcabbott, line 363: Does in fact accept `rng`,  but discards it. My take is that it's there only for testing convenience, seems a bit confusing to note it.

ToucheSir: How about putting the comment added below into the docstring? One non-testing related use case for this would be easier programmatic construction of models.

src/utils.jl
darsnack: I think it also describes how to do it using the inverse CDF which is what is done here. It is a useful implementation reference, but I agree that truncated normal is self-explanatory.

darsnack:
Thoughts on moving the initialization functions to their own file(s)? `utils.jl` is unwieldy to sift through right now.

src/utils.jl
mcabbott: Oh right, sorry, it does mention both. 

In fact https://en.wikipedia.org/wiki/Truncated_normal_distribution mentions these and some other ways. Maybe that's evidence that this isn't obscure.

src/utils.jl
mcabbott: Done.

mcabbott:
Maybe I vote to kick the can down the road on moving things. I agree it's not optimal, but don't mind long files too much, and do like git blame sometimes.

mcabbott:
Good to go?

darsnack:
Only comment is that we should double check that nothing promote to `Float64` when the keywords are `Float64` (the current docstrings imply this behavior). From my pass, at least `orthogonal_init` seems like it would promote for a `Float64` gain.

mcabbott:
Indeed, that should be tested, and `orthogonal` used to promote. Done.

darsnack (approved):

PR #1232, opened 2020-06-16
add Flux.skip()

Moelf:
per #821

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md
- [x] Documentation, if applicable
- [ ] Final review from `@MikeInnes` or `@dhairyagandhi96` (for API changes).

PR #1913, opened 2022-03-21
Examples for `skip` and `stop`

mcabbott:
`skip` claims that if used in the callback passed to `train!`, the `update!` step will not happen. This isn't true, it's called too late for that. Its tests don't try to test that, instead they test that if used in the loss function, it doesn't evaluate the rest of the loss.

So right now this PR adds an example showing what it does do. Although perhaps better to simply remove it? Github's search can't find a single use of it in the wild. Originally added in #1232.

Also adds an example for `stop`, which does work as described.

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -43,14 +43,43 @@ struct SkipException <: Exception end
 """
     skip()
 
-Call `Flux.skip()` in a callback to indicate when a callback condition is met.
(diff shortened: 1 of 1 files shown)


ToucheSir:
I have definitely seen it on code posted to Discourse, so removal might be too disruptive. The "proper" design would be to allow users to specify where in the training loop each callback runs, but it's a slippery slope from there to something like FluxTraining.jl.

PR #1916, opened 2022-03-22
Add a ton of doctests + fix outdated documentation in `.md` files

Saransh-cpp:
First of all, thank you for such excellent documentation! I have been learning `Flux.jl` through its documentation, and it has been amazing so far! I noticed, however, that a lot of examples were outdated; in the sense that their outputs did not match those running on Flux's master branch.

I've tried fixing such examples, and I have also added a lot of doctests to make sure that this does not happen again! I've skipped adding doctests to a few examples as that would require some discussions with the maintainers, but I'll be happy to add them in this PR itself! :)

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -6,7 +6,7 @@ NVIDIA GPU support should work out of the box on systems with CUDA and CUDNN ins
 
 By default, Flux will run the checks on your system to see if it can support GPU functionality. You can check if Flux identified a valid GPU setup by typing the following:
 
-```julia
+```jldoctest gpu; setup = :(using Random; Random.seed!(0))
 julia> using CUDA
 
 julia> CUDA.functional()
@@ -51,39 +51,62 @@ d(cu(rand(10)))
 
 As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing. So, you can safely call `gpu` on some data or model (as shown below), and the code will not error, regardless of whether the GPU is available or not. If the GPU library (CUDA.jl) loads successfully, `gpu` will move data from the CPU to the GPU. As is shown below, this will change the type of something like a regular array to a `CuArray`.
 
-```julia
+```jldoctest gpu
 julia> using Flux, CUDA
 
 julia> m = Dense(10, 5) |> gpu
-Dense(10 => 5)
+Dense(10 => 5)      # 55 parameters
 
 julia> x = rand(10) |> gpu
-10-element CuArray{Float32,1}:
- 0.800225
- ⋮
- 0.511655
+10-element CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}:
+ 0.066846445
+ 0.15663664
+ 0.60529673
+ 0.13574456
+ 0.8381178
+ 0.914712
+ 0.30007496
+ 0.7228498
+ 0.11965257
+ 0.76706964
 
 julia> m(x)
-5-element CuArray{Float32,1}:
- -0.30535
- ⋮
- -0.618002
+5-element CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}:
+ -0.99992573
+  0.08930303
+  0.4309393
+ -0.5205649
+ -0.547261
 ```
 
 The analogue `cpu` is also available for moving models and data back off of the GPU.
 
-```julia
+```jldoctest gpu
 julia> x = rand(10) |> gpu
-10-element CuArray{Float32,1}:
- 0.235164
- ⋮
- 0.192538
+10-element CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}:
+ 0.8019236
+ 0.03534455
+ 0.48466054
+ 0.8991991
+ 0.9516907
+ 0.8011185
+ 0.12432273
+ 0.114268765
+ 0.07955447
+ 0.7766742
 
 julia> x |> cpu
-10-element Array{Float32,1}:
- 0.235164
- ⋮
- 0.192538
+10-element Vector{Float32}:
+ 0.8019236
+ 0.03534455
+ 0.48466054
+ 0.8991991
+ 0.9516907
+ 0.8011185
+ 0.12432273
+ 0.114268765
+ 0.07955447
+ 0.7766742
 ```
 
 ## Disabling CUDA or choosing which GPUs are visible to Flux
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -6,7 +6,7 @@ NVIDIA GPU support should work out of the box on systems with CUDA and CUDNN ins
 
 By default, Flux will run the checks on your system to see if it can support GPU functionality. You can check if Flux identified a valid GPU setup by typing the following:
 
-```jldoctest gpu; setup = :(using Random; Random.seed!(0))
+```julia
 julia> using CUDA
 
 julia> CUDA.functional()
@@ -51,7 +51,7 @@ d(cu(rand(10)))
 
 As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing. So, you can safely call `gpu` on some data or model (as shown below), and the code will not error, regardless of whether the GPU is available or not. If the GPU library (CUDA.jl) loads successfully, `gpu` will move data from the CPU to the GPU. As is shown below, this will change the type of something like a regular array to a `CuArray`.
 
-```jldoctest gpu
+```julia
 julia> using Flux, CUDA
 
 julia> m = Dense(10, 5) |> gpu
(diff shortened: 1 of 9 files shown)


docs/src/models/overview.md
```diff
@@ -28,7 +28,7 @@ This example will build a model to approximate the `actual` function.
 
 Use the `actual` function to build sets of data for training and verification:
 
-```julia
+```jldoctest overview
```
CarloLucibello, line 31: curiosity: what is this overview option?

Saransh-cpp: This preserves variables, function definitions, etc. in all the blocks labeled "overview". 

"Every doctest block is evaluated inside its own module. This means that definitions (types, variables, functions etc.) from a block can not be used in the next block"

More information on this here - https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Preserving-Definitions-Between-Blocks

docs/src/gpu.md
```diff
@@ -55,35 +55,58 @@ As a convenience, Flux provides the `gpu` function to convert models and data to
...
+ 0.914712
+ 0.30007496
+ 0.7228498
+ 0.11965257
+ 0.76706964
 
 julia> m(x)
-5-element CuArray{Float32,1}:
- -0.30535
- ⋮
```
ToucheSir, line 69: I'm not sure if these ellipses were added automatically before, but they should probably be there. Printing out 10 random numbers just takes up vertical space.

docs/src/models/regularisation.md
```diff
@@ -6,66 +6,79 @@ add the result to the overall loss.
...
 When working with layers, Flux provides the `params` function to grab all
 parameters at once. We can easily penalise everything with `sum`:
 
-```julia
+```jldoctest regularisation
 julia> Flux.params(m)
-2-element Array{Any,1}:
- param([0.355408 0.533092; … 0.430459 0.171498])
- param([0.0, 0.0, 0.0, 0.0, 0.0])
+Params([Float32[0.34704182 -0.48532376 … -0.06914271 -0.38398427; 0.5201164 -0.033709668 … -0.36169025 -0.5552353; … ; 0.46534058 0.17114447 … -0.4809643 0.04993277; -0.47049698 -0.6206029 … -0.3092334 -0.47857067], Float32[0.0, 0.0, 0.0, 0.0, 0.0]])
```
ToucheSir, line 36: There's no reason to make this a doctest. If anything, it'd cause more headaches if anything around RNGs changes.

docs/src/models/regularisation.md
```diff
@@ -6,66 +6,79 @@ add the result to the overall loss.
...
+julia> sqnorm(x) = sum(abs2, x)
+sqnorm (generic function with 1 method)
 
-loss(x, y) = logitcrossentropy(m(x), y) + sum(sqnorm, Flux.params(m))
+julia> loss(x, y) = logitcrossentropy(m(x), y) + sum(sqnorm, Flux.params(m))
+loss (generic function with 1 method)
 
-loss(rand(28^2), rand(10))
+julia> loss(rand(28^2), rand(10))
+300.76693683244997
```
ToucheSir, line 62: Same here.

docs/src/models/regularisation.md
```diff
@@ -6,66 +6,79 @@ add the result to the overall loss.
...
 julia> activations(c, rand(10))
-3-element Array{Any,1}:
- Float32[0.84682214, 0.6704139, 0.42177814, 0.257832, 0.36255655]
- Float32[0.1501253, 0.073269576]                                 
- Float32[0.5192045, 0.48079553]                                  
+([0.3274892431795043, 0.5360197770386552, 0.3447464835514667, 0.5273025865532305, 0.7513168089280781], [-0.3533774181890544, -0.010937055274926138], [0.4152168057978045, 0.5847831942021956])
 
 julia> sum(sqnorm, ans)
-2.0710278f0
+1.9953131077618562
```
ToucheSir, line 81: And here. Really anywhere with output that isn't semantically meaningful (or stable) shouldn't be tested.

docs/src/saving.md
```diff
@@ -90,16 +92,23 @@ The new `model` we created will now be identical to the one we saved parameters
...
 
-evalcb = throttle(30) do
-  # Show loss
-  @save "model-checkpoint.bson" model
-end
+julia> evalcb = throttle(30) do
+         # Show loss
+         @save "model-checkpoint.bson" model
+       end
+(::Flux.var"#throttled#109"{Flux.var"#throttled#105#110"{Bool, Bool, var"#1#2", Int64}}) (generic function with 1 method)
```
ToucheSir, line 111: Last comment about doctest outputs for now. Here the problem is that the gensym numbers will not always be consistent. Also there's no point testing the return type, the function is literally defined above ;)

docs/src/gpu.md
mcabbott: Possibly the example should be shortened so that displaying the honest output doesn't need manual shortening? (Or it might be done automatically by making the window really small.)

I presume that these CuArray ones can't be doctests anyway.

docs/src/models/regularisation.md
mcabbott: I used to say this. But now I wonder if it isn't better just to make as many things as possible doctests, and lock them to run only on 1.6, with a fixed seed (as is I think done here). When we want to raise the docs to 1.9 or whatever, it's pretty easy to automatically update them all, once. But in the meantime it helps ensure that things don't go stale.

mcabbott: For example, `activations(c, rand(10))` is shown as returning an array, but has in fact changed to return a tuple. Not deadly, but it's nice if all such things are correct.

docs/src/saving.md
mcabbott: Yes, no way around `var"#throttled#109"` being random. But you can add a semicolon. 

In fact, quite a few lines should probably have one. While I do like the pretty printing, it's a bit much to have it on every little example, when it's literally printing what was just typed. Likewise things like `loss (generic function with 1 method)` should often be hidden, I think.

docs/src/models/regularisation.md
ToucheSir: I think for scenarios like that, `summary` would be a better tool because it keeps all but the actual contents of the value (which we wouldn't care about).

docs/src/gpu.md
Saransh-cpp: Yes, I realized this after creating the PR. GitHub Actions don't provide GPU. I will add the ellipses again.

Saransh-cpp:
Hi @ToucheSir, @mcabbott!

This is on my radar for this week. Could you please let me know if I should keep the random outputs? Or should I suppress the outputs generated by RNGs using `;`?

docs/src/models/regularisation.md
ToucheSir: @Saransh-cpp see comment above and https://github.com/FluxML/Flux.jl/pull/1916#discussion_r832768986 to answer your question.

docs/src/models/overview.md
```diff
@@ -129,22 +129,22 @@ The first parameter is the weight and the second is the bias. Flux will adjust p
...
 And check the loss:
 
-```
+```jldoctest overview
 julia> loss(x_train, y_train)
-267.8037f0
+116.38745f0
 ```
 
 It went down. Why?
```
mcabbott, line 143: Printing the loss seems like part of the point of this discussion.

If you do that, then either it should not be tested (as before) or it must be tested with a fixed seed and version (as in this PR). 

And once one jldoctest depends on the seed, maybe there's little cost to letting the others do so. When we change the version, it's pretty easy to auto-update all of them at once. 

While I'm normally in favour of writing tests which don't care about the seed, it would require quite a lot of changes to make this page so. And there's something to be said for printing out the real objects just as you would see them.

So my vote is that this PR is a good idea.

ToucheSir: If we're really going down this path, every page that needs it should be swapped over to StableRNGs.

mcabbott: But at the cost of moving it further from being a "here's the simplest way to use this" tutorial, introducing lots of explicit RNG handling. Unless there's a hack to make the global RNG magically the stable one?

It doesn't seem like a great sin to have the docs only succeed on one version. I agree it's not ideal. But having them not tested at all seems worse, as they can be silently wrong.

Saransh-cpp: I think we can filter the doctests using `regex`. There are some simple examples documented here - https://juliadocs.github.io/Documenter.jl/stable/man/doctests/#Filtering-Doctests. The tests would then effectively ignore the randomly generated values and will only test the deterministic quantities. Should I try this?

Saransh-cpp:
Filtering doctests using `regex` worked! The tests are not failing on not setting the random seed.

docs/src/models/overview.md
mcabbott: This seems like a good solution, I think.

Saransh-cpp: Yes, I've added `filter` in every doctest having randomly generated numbers. It seems to work!

CarloLucibello (approved):

CarloLucibello:
Thanks for all this work @Saransh-cpp

PR #1925, opened 2022-04-02
Allow activation function for Diagonal

theabhirath:
Following up on #1911 , this PR allows for an activation function argument to be passed to `Flux.Diagonal` and reroutes the `LayerNorm` activation function to `Flux.Diagonal` now. This reduces memory and time taken for `LayerNorm` with an activation function.

Before this PR (Julia master, Flux master):

```julia
julia> l = LayerNorm(768, relu);

julia> x = rand(Float32, 768, 128, 2);

julia> @benchmark $(l)($x)
 Range (min … max):  193.291 μs …   4.904 ms  ┊ GC (min … max):  0.00% … 91.62%
 Time  (median):     310.625 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   385.946 μs ± 415.163 μs  ┊ GC (mean ± σ):  19.75% ± 16.07%

  ▂▃█▇▂                                                         ▁
  █████▆▅▅▄▃▁▁▁▁▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▃▄▄▆▆▆▇▅▆▇▆▇▇▆▆▆▅▅▆▅▅▅▆▅▅▁▅ █
  193 μs        Histogram: log(frequency) by time       2.75 ms <

 Memory estimate: 2.25 MiB, allocs estimate: 18.
```

After this PR:
```julia
julia> @benchmark $(l)($x)
 Range (min … max):  168.333 μs …   1.870 ms  ┊ GC (min … max):  0.00% … 84.42%
 Time  (median):     245.667 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   281.375 μs ± 209.594 μs  ┊ GC (mean ± σ):  13.78% ± 14.90%

  ▄▃▄█▇▃                                                        ▁
  ███████▅▅▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▄▅▅▆▆▇▇██▇▇▇▇ █
  168 μs        Histogram: log(frequency) by time       1.43 ms <

 Memory estimate: 1.50 MiB, allocs estimate: 16.
```

PS: there are some formatting changes that I just made because they caught my eye, but if they're jarring I can always remove them

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -138,13 +138,13 @@ julia> Flux.params(d1)  # no trainable bias
 Params([[1.0 1.0 … 1.0 1.0; 1.0 1.0 … 1.0 1.0]])
 ```
 """
-struct Dense{F, M<:AbstractMatrix, B}
+struct Dense{M<:AbstractMatrix, B, F}
   weight::M
   bias::B
   σ::F
   function Dense(W::M, bias = true, σ::F = identity) where {M<:AbstractMatrix, F}
     b = create_bias(W, bias, size(W,1))
-    new{F,M,typeof(b)}(W, b, σ)
+    new{M, typeof(b), F}(W, b, σ)
   end
 end
 
@@ -158,7 +158,7 @@ end
 function (a::Dense)(x::AbstractVecOrMat)
   W, b = a.weight, a.bias
   σ = NNlib.fast_act(a.σ, x)  # replaces tanh => tanh_fast, etc
-  return σ.(W*x .+ b)
+  return σ.(W * x .+ b)
 end
 
 (a::Dense)(x::AbstractArray) = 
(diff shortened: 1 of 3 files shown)


theabhirath:
There was no type restriction on the input for `Flux.Diagonal` earlier - there's a test to that effect checking if it works with just an integer as an input (which it did earlier). The fast activation path now is incompatible with that though, so is this something I should manually work around or is an `AbstractArray` restriction actually better for the layer given the general usecase?

src/layers/basic.jl
```diff
@@ -138,13 +138,13 @@ julia> Flux.params(d1)  # no trainable bias
 Params([[1.0 1.0 … 1.0 1.0; 1.0 1.0 … 1.0 1.0]])
 ```
 """
-struct Dense{F, M<:AbstractMatrix, B}
+struct Dense{M<:AbstractMatrix, B, F}
```
darsnack, line 141: I think this would be breaking to reorder

darsnack (changes requested):

src/layers/basic.jl
```diff
@@ -172,35 +171,37 @@ function Base.show(io::IO, l::Dense)
...
 
-Diagonal(sz::Integer...; bias = true, init = ones32) = Diagonal(init(sz...), bias)
+Diagonal(sz::Integer...; σ = identity, bias = true, init = ones32) = Diagonal(init(sz...), bias, σ)
 
 @functor Diagonal
 
-(a::Diagonal)(x) = a.scale .* x .+ a.bias
+function (a::Diagonal)(x::AbstractArray)
+  σ = NNlib.fast_act(a.σ, x)  # replaces tanh => tanh_fast, etc
+  return σ == typeof(identity) ? a.scale .* x .+ a.bias : σ.(a.scale .* x .+ a.bias)
```
ToucheSir, line 203:
```suggestion
  return σ === typeof(identity) ? a.scale .* x .+ a.bias : σ.(a.scale .* x .+ a.bias)
```
Needs the extra `=` to trigger compiler smarts, I believe. That and Zygote might make a mess of the conditional logic, so can you test backwards pass performance as well?

theabhirath: Checked the backwards pass and while there's no improvement, there are no regressions

CarloLucibello:
I think restricting to abstract array is fine and this is a good moment to introduce a breaking change

CarloLucibello:
> I think restricting to abstract array is fine and this is a good moment to introduce a breaking change

adding this PR to the milestone in case we decide to go with this

ToucheSir:
Let's do it, I highly doubt anyone was passing non-array inputs anyhow.

CarloLucibello (approved):

PR #1926, opened 2022-04-03
Upgrade warnings for v0.13

mcabbott:
Options for v0.12 deprecations are:
* do nothing, leave things working
* make them louder, so the message is printed always not just in tests
* turn them into errors, now that there has been time to update.

Thoughts on which to do, where?

### DIFF
--- a/src/losses/utils.jl
+++ b/src/losses/utils.jl
@@ -26,12 +26,11 @@ end
 ChainRulesCore.@scalar_rule xlogy(x, y) (log(y), x/y)  # should help Diffractor's broadcasting
 ChainRulesCore.@scalar_rule xlogx(x) (log(y) + true)
 
-# This can be made an error in Flux v0.13, for now just a warning
 function _check_sizes(ŷ::AbstractArray, y::AbstractArray)
   for d in 1:max(ndims(ŷ), ndims(y)) 
-    if size(ŷ,d) != size(y,d)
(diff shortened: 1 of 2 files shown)


src/losses/utils.jl
```diff
@@ -26,12 +26,11 @@ end
...
   for d in 1:max(ndims(ŷ), ndims(y)) 
-    if size(ŷ,d) != size(y,d)
-      @warn "Size mismatch in loss function! In future this will be an error. In Flux <= 0.12 broadcasting accepts this, but may not give sensible results" summary(ŷ) summary(y) maxlog=3 _id=hash(size(y))
-    end
+   size(ŷ,d) == size(y,d) || throw(DimensionMismatch(
+      "loss function expects size(ŷ) = $(size(ŷ)) to match size(y) = $(size(y))"
+    ))
   end
 end
 _check_sizes(ŷ, y) = nothing  # pass-through, for constant label e.g. y = 1
```
mcabbott, line 36: I still wonder if this should go further, BTW, and forbid comparing things which aren't arrays. Things like this:

https://stackoverflow.com/questions/71540612/no-decrease-in-loss-after-lots-traning

darsnack:
I would just make them as loud as possible without an error. I feel that the more time you give to update, the less people can complain when they are finally forced to do it. The current set of warnings is just a simple naming issue, there is not reason to force anyone's hand. Something that we decide is plain wrong is candidate for error ASAP.

mcabbott:
Happy with the levels here? `Flux.ones` should now print a warning every time. And loss functions with mismatched sizes will error.

ToucheSir (approved):

PR #1927, opened 2022-04-04
Rename `Diagonal` to `Scale`

mcabbott:
This avoids the awkward collision with LinearAlgebra.Dense. 

It also changes it to be written `Scale(2,3,tanh)` without making a tuple of the sizes, and similarly allows `LayerNorm(2,3,tanh)`.

Builds on #1925

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -211,17 +211,17 @@ julia> Flux.params(b)
 Params([[1 2 3 4]])
 ```
 """
-struct Scale{F<:Function, A<:AbstractArray, B}
+struct Scale{F, A<:AbstractArray, B}
   scale::A
   bias::B
   σ::F
-  function Scale(scale::A, bias = true, σ::F = identity) where {A<:AbstractArray, F<:Function}
+  function Scale(scale::A, bias::B = true, σ::F = identity) where {A<:AbstractArray, B<:Union{Bool, AbstractArray}, F}
     b = create_bias(scale, bias, size(scale)...)
     new{F, A, typeof(b)}(scale, b, σ)
   end
 end
 
-Scale(size::Integer...; bias = true, init = ones32, _act = identity) = Scale(init(size...), bias, _act)
+Scale(s1::Integer, s23::Integer...; bias = true, init = ones32, _act = identity) = Scale(init(s1, s23...), bias, _act)
 Scale(size_act...; bias = true, init = ones32) = Scale(size_act[1:end-1]...; bias, init, _act = size_act[end])
 
 @functor Scale
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -189,7 +189,7 @@ Used by [`LayerNorm`](@ref) with `affine=true`.
 # Examples
 ```jldoctest
 julia> a = Flux.Scale(2)
-Scale(2)      # 4 parameters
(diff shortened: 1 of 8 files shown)


src/layers/basic.jl
```diff
@@ -171,40 +171,69 @@ function Base.show(io::IO, l::Dense)
...
+Scale(size_act...; bias = true, init = ones32) = Scale(size_act[1:end-1]...; bias, init, _act = size_act[end])
 
-@functor Diagonal
+@functor Scale
 
-function (a::Diagonal)(x::AbstractArray)
+function (a::Scale)(x::AbstractArray)
   σ = NNlib.fast_act(a.σ, x)  # replaces tanh => tanh_fast, etc
-  return σ === typeof(identity) ? a.scale .* x .+ a.bias : σ.(a.scale .* x .+ a.bias)
+  σ.(a.scale .* x .+ a.bias)
```
mcabbott, line 231: I removed this ` ? : `  because I think the fused `identity.()` should be free. Zygote knows to ignore it too.

src/layers/normalise.jl
```diff
@@ -164,17 +164,19 @@ struct LayerNorm{F,D,T,N}
   affine::Bool
 end
 
-function LayerNorm(sz, λ=identity; affine::Bool=true, ϵ::Real=1f-5)
-  diag = affine ? Diagonal(sz...; σ = λ) : Base.Fix1(broadcast, λ)
-  return LayerNorm(λ, diag, ϵ, Tuple(sz), affine)
+function LayerNorm(size::Tuple{Vararg{Int}}, λ=identity; affine::Bool=true, ϵ::Real=1f-5)
+  diag = affine ? Scale(size..., λ) : λ!=identity ? Base.Fix1(broadcast, λ) : identity
+  return LayerNorm(λ, diag, ϵ, size, affine)
```
mcabbott, line 169: Here I added one more branch: `identity.(x)` has nothing to fuse with here, so it will cost a copy.

src/layers/basic.jl
```diff
@@ -171,40 +171,69 @@ function Base.show(io::IO, l::Dense)
...
+
+julia> Flux.params(b)
+Params([[1 2 3 4]])
+```
 """
-struct Diagonal{A<:AbstractArray, B, F}
+struct Scale{F<:Function, A<:AbstractArray, B}
   scale::A
   bias::B
   σ::F
```
johnnychen94, line 217: Why we limit `F <: Function` here? I don't see a clear reason to exclude functors as activation functions.

mcabbott: No very strong reason, it makes dumb things like `Flux.Scale(1,2,3.)` give an error on construction not later. 

Are there any examples of non-Function activation functions which might be useful?

johnnychen94: > Are there any examples of non-Function activation functions which might be useful?

Activation layers with trainable parameters, e.g., [PReLU](https://arxiv.org/abs/1502.01852).

mcabbott: Ok, but that has a coefficient per channel, so it needs to see the whole array right? Unless it hacks broadcasting in some awful way.

mcabbott: Anyway, now removed, and thus matches the other layers in Flux.

CarloLucibello (approved):

PR #1928, opened 2022-04-05
Don't import Losses in Flux namespace

CarloLucibello:


ToucheSir:
Does anything outside of the tests need updating?

CarloLucibello:
I changed my mind on this, `Flux.mse` and others are used in many docs, doing this now would be too breaking. Maybe also there is really no need for it (I just think it makes more clear to the user where definitions live)

mcabbott:
One downside of the module, btw, is that they don't autocomplete: `Flux.ms <tab>` shows nothing.

PR #1930, opened 2022-04-06
cleanup Flux.Losses documentation

CarloLucibello:
afterthought of #1928

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,6 +1,11 @@
 using Documenter, Flux, NNlib, Functors, MLUtils
 
 DocMeta.setdocmeta!(Flux, :DocTestSetup, :(using Flux); recursive = true)
+DocMeta.setdocmeta!(Flux.Losses, :DocTestSetup, :(using Flux.Losses); recursive = true)
+
+# In the Losses module, doctests which differ in the printed Float32 values won't fail
+DocMeta.setdocmeta!(Flux.Losses, :DocTestFilters, :(r"[0-9\.]+f0"); recursive = true)
+
 makedocs(modules = [Flux, NNlib, Functors, MLUtils],
          doctest = false,
          sitename = "Flux",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,10 +1,6 @@
 using Documenter, Flux, NNlib, Functors, MLUtils
 
 DocMeta.setdocmeta!(Flux, :DocTestSetup, :(using Flux); recursive = true)
-DocMeta.setdocmeta!(Flux.Losses, :DocTestSetup, :(using Flux.Losses); recursive = true)
-
-# In the Losses module, doctests which differ in the printed Float32 values won't fail
-DocMeta.setdocmeta!(Flux.Losses, :DocTestFilters, :(r"[0-9\.]+f0"); recursive = true)
 
 makedocs(modules = [Flux, NNlib, Functors, MLUtils],
          doctest = false,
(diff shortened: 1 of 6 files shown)


docs/make.jl
```diff
@@ -1,6 +1,11 @@
 using Documenter, Flux, NNlib, Functors, MLUtils
 
 DocMeta.setdocmeta!(Flux, :DocTestSetup, :(using Flux); recursive = true)
+DocMeta.setdocmeta!(Flux.Losses, :DocTestSetup, :(using Flux.Losses); recursive = true)
```
CarloLucibello, line 4: it seems this line is not doing its job and docs CI is failing because of this

src/losses/functions.jl
```diff
@@ -32,13 +28,14 @@ Return the loss corresponding to mean square error:
...
+# Examples
+
 ```jldoctest
 julia> y_model = [1.1, 1.9, 3.1];
 
 julia> y_true = 1:3;
 
-julia> Flux.mse(y_model, y_true)
+julia> mse(y_model, y_true)
 0.010000000000000018
```
mcabbott, line 39: Why change these? 

Elsewhere in Flux, all (ideally!) non-exported functions are qualified in doctests. It seems important to know you have to, seems bad to write examples which won't work at the repl.

docs/src/models/losses.md
```diff
@@ -3,43 +3,36 @@
...
+In Flux's convention, the order of the arguments is the following:
 
 ```julia
 loss(ŷ, y)
 ```
 
+They are commonly passed as arrays of size `num_target_features x num_examples_in_batch`. 
+
 Most loss functions in Flux have an optional argument `agg`, denoting the type of aggregation performed over the
 batch:
```
mcabbott, line 23: This is not quite true, BTW. Sometimes `agg` is the only reduction applied, to all dimensions, not just the batch:
```
julia> Flux.crossentropy([1 2; 3 4], [5 6; 7 8]; agg=identity)
1×2 Matrix{Float64}:
 -7.69029  -15.2492

julia> Flux.mse([1 2; 3 4], [5 6; 7 8]; agg=identity)
2×2 Matrix{Int64}:
 16  16
 16  16
```
IIRC this was done to to follow what Pytorch means by mean/sum (scaled by the length, not just the batch size) and for efficiency. But it perhaps means we shouldn't promise too much here, it's kind-of sort-of the batch reduction, but what `loss(ŷ, y, agg=x->sum(x, dims=2))` will do, best you check the source.

docs/make.jl
darsnack: I believe these lines are useless since `makedocs` have the tests disabled. In the GH CI.yml, there is a separate version of these lines, and that's the one that matters.

src/losses/functions.jl
CarloLucibello: added `using Flux.Losses: mse` statements

docs/make.jl
CarloLucibello: turns out that there is an additional problem, `DocMeta.setdocmeta!` doesn't accept the `DocTestFilters` that I also needed, so I had to revert this change entirely

ToucheSir:
Thoughts on rebasing this since we've had some QoL changes to the docs build recently? cc @Saransh-cpp

Saransh-cpp:
Yes, rebasing should work! Additionally, I think these lines - https://github.com/FluxML/Flux.jl/blob/8bc0c35932c4a871ac73b42e39146cd9bbb1d446/src/losses/functions.jl#L1-L4
are redundant, and can be removed.

Rebase should also get the DocBot to comment here. Currently, the updated docs can be viewed through the [`documenter/deploy`](https://fluxml.github.io/Flux.jl/previews/PR1930/) action.

CarloLucibello:

Issue #1929, opened 2022-04-05
Chain ignores Base.show function of custom layer


fpartl:
Hello, 

I have created custom BiLSTM layer like this:
```julia
using Flux

mutable struct BiLSTM{L<:Flux.Recur{T} where T<:Flux.LSTMCell}
    fw::L
    bw::L
end

function BiLSTM(in::Int, out::Int)
    if out % 2 != 0
        throw(ArgumentError("Output size must be divisible by 2."))
    end

    return BiLSTM(
        LSTM(in, out ÷ 2), 
        LSTM(in, out ÷ 2)
    )
end

Flux.@functor BiLSTM
Flux.trainable(b::BiLSTM) = (b.fw, b.bw)

Base.show(io::IO, ::MIME"text/plain", b::BiLSTM) = 
    print(io, "BiLSTM($(size(b.fw.cell.Wi, 2)), $(size(b.fw.cell.Wi, 1)÷2))")

# ...
```

When I instantiate `BiLSTM` cell using `BiLSTM` function the `Base.show` function is used as expected. 
```julia
julia> BiLSTM(12, 16)
BiLSTM(12, 16)
```
But if I create a `Chain` of BiLSTM cells my custom `Base.show` function is ignored. 
```julia
julia> Chain(BiLSTM(12, 16), BiLSTM(16, 16), BiLSTM(16, 16))
Chain(
  BiLSTM(
    Recur(
      LSTMCell(12, 8),                  # 688 parameters
    ),
    Recur(
      LSTMCell(12, 8),                  # 688 parameters
    ),
  ),
  BiLSTM(
    Recur(
      LSTMCell(16, 8),                  # 816 parameters
    ),
    Recur(
      LSTMCell(16, 8),                  # 816 parameters
    ),
  ),
  BiLSTM(
    Recur(
      LSTMCell(16, 8),                  # 816 parameters
    ),
    Recur(
      LSTMCell(16, 8),                  # 816 parameters
    ),
  ),
)         # Total: 30 trainable arrays, 4_640 parameters,
          # plus 12 non-trainable, 96 parameters, summarysize 20.141 KiB.
```

Is there any way to make this less verbose? Something like this would be nice. 
```julia
julia> Chain(BiLSTM(12, 16), BiLSTM(16, 16), BiLSTM(16, 16))
Chain(
  BiLSTM(12, 16),                      # 1376 parameters
  BiLSTM(16, 16),                      # 1632 parameters
[...]

PR #1932, opened 2022-04-07
Add a macro to opt-in to fancy printing, and to everything else

### LINKED ISSUE #2044: Macro to display model struct the way Flux does
Semi-related to #2043 😄. Custom model structs use
```julia
@functor MyModel
```
to define `Functors.children` etc. It would be nice if there was also
```julia
@define_layer_show MyModel
```
to generate a method for `Base.show`, like
```julia
using Flux
function Base.show(io::IO, m::MIME"text/plain", x::MyModel)
    if get(io, :typeinfo, nothing) === nothing  # e.g. top level in REPL
      Flux._big_show(io, x)
    elseif !get(io, :compact, false)  # e.g. printed inside a Vector, but not a Matrix
      Flux._layer_show(io, x)
    else
      show(io, x)
    end
  end
end
```

Alternatives:
* I guess this could also be built into `@functor` but I suspect it might be better to keep it separate/ opt-in so you can define your own show method if you want. Also, adding it to `@functor` would also presumably be breaking since folks may have custom show methods already and that would generate method overwrite warnings and possibly break precompilation.
* This could be done with an abstact type, but I think it's nice if folks are free to have their model partipicate in their own type hierarchy. But an abstract type seems OK too.

mcabbott:
### Originally, `@big_show`

This adds a macro which lets you tell Flux to treat some container type much the same way as it treats `Chain`, starting the recursive printing walk over all children.

Prompted by #1929. Closes #2044

Metalhead also has `show` methods for top-level layers which enable the recursive printing, which could be replaced by this macro:

https://github.com/FluxML/Metalhead.jl/blob/aba6fb832093d88dc2d2b4d5b1d2d63a0f21eb9c/src/Metalhead.jl#L53-L57

https://github.com/FluxML/Metalhead.jl/blob/c8f0a88e4d24274c53b62c2b740822aa6a781709/src/utilities.jl#L49-L52

Searching for other uses on github, I see only two:

https://github.com/avik-pal/ExplicitFluxLayers.jl/blob/8e1cff447afda225bc12144ff27ae1370e8fc3da/src/show_layers.jl

https://github.com/maxfreu/SegmentationModels.jl/blob/7bfdbaa438910baf49543f03f2931de765dbd761/src/unet.jl#L99-L112

### Later, `@layer`

Now aims to also replace `@functor`, in addition to handling `show`.

Discussed at https://github.com/FluxML/Flux.jl/pull/2028 -- a layer supertype is another way to make opting into pretty printing easier. (We wouldn't do both, so this closes https://github.com/FluxML/Flux.jl/pull/2028 .)

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/docs/src/models/basics.md
+++ b/docs/src/models/basics.md
@@ -257,8 +257,8 @@ m(5) # => 26
 
 There is still one problem with this `Affine` layer, that Flux does not know to look inside it. This means that [`Flux.train!`](@ref) won't see its parameters, nor will [`gpu`](@ref) be able to move them to your GPU. These features are enabled by the [`@functor`](@ref Functors.@functor) macro:
 
-```
-Flux.@functor Affine
+```julia
+Flux.@layer Affine
 ```
 
 Finally, most Flux layers make bias optional, and allow you to supply the function used for generating random weights. We can easily add these refinements to the `Affine` layer as follows, using the helper function [`create_bias`](@ref Flux.create_bias):
@@ -271,4 +271,9 @@ function Affine((in, out)::Pair; bias=true, init=Flux.randn32)
 end
 
 Affine(3 => 1, bias=false, init=ones) |> gpu
+```
+
+```@docs
+Flux.@layer
+Flux.create_bias
 ```
--- a/src/layers/macro.jl
+++ b/src/layers/macro.jl
@@ -131,18 +131,20 @@ function _default_functor(::Type{T}, x) where {T}
      F = fieldnames(T)
      args = map(sy -> :(getfield(x, $(QuoteNode(sy)))), F)
      C = Base.typename(T).wrapper  # constructor
-     recon = VERSION > v"1.9-" ? :(Splat($C)) : :(Base.splat($C))
+     # recon = VERSION > v"1.9-" ? :(Splat($C)) : :(Base.splat($C))
+     recon = :(Base.splat($C))
      :((NamedTuple{$F}(($(args...),)), $recon))
    else
      # Getting this parameterless type takes about 2μs, every time:
-     spl = VERSION > v"1.9-" ? Splat : Base.splat
+     # spl = VERSION > v"1.9-" ? Splat : Base.splat
+     spl = Base.splat
      namedtuple(x), spl(Base.typename(T).wrapper)
    end
 end
 
 function _custom_functor(::Type{T}, x, ::Val{which}) where {T,which}
    if false
-     # TODO write the @generated version
+     # TODO write the @generated version. Or decide we don't care, or should forbid this?
    else
      remake(nt) = Base.typename(T).wrapper(map(f -> f in which ? getfield(nt, f) : getfield(x, f), fieldnames(T))...)
      NamedTuple{which}(map(s -> getfield(x, s), which)), remake
--- a/src/layers/macro.jl
+++ b/src/layers/macro.jl
@@ -1,0 +1,173 @@
+
+"""
+    @layer Dense
+    @layer :expand Chain
+    @layer BatchNorm trainable=(β,γ)
+    @layer Struct children=(α,β) trainable=(β,)
+
+This macro replaces most uses of `@functor`. Its basic purpose is the same:
+When you define a new layer, this tells Flux to explore inside it
+to see the parameters it trains, and also to move them to the GPU, change precision, etc.
+Like `@functor`, this assumes your struct has the default constructor, to enable re-building.
+
+Some keywords allow you to limit this exploration, instead of visiting all `fieldnames(T)`.
+Note that it is never necessary to tell Flux to ignore non-array objects such as functions or sizes.
+* If some fields look like parameters but should not be trained,
+  then `trainable` lets you specify which fields to include, while the rest are ignored.
+* You can likewise add restrictions to Functors's `children` (although this is seldom a good idea),
+  equivalent to `@functor Struct (α,β)`. Any `trainable` limitation must then be a subset of `children`.
+
+The macro also handles overloads of `show` for pretty printing.
+* By default, it adds methods to 3-arg `Base.show` to treat your layer much like `Dense` or `Conv`.
+* If your layer is a container, more like `Chain` or `Parallel`, then `:expand` makes `show` unfold its contents.
+* To disable all `show` overloads, there is an `:ignore` option too.
+
+(You probably still want to define 2-arg `show(io::IO, x::Layer)`, the macro does not touch this.)
+
+Note that re-running the macro with different options may not overwrite all methods, you will need to restart.
+
+# Example
+```jldoctest
+julia> struct Trio; a; b; c end
+
+julia> tri = Trio(Dense([1.1 2.2], [0.0], tanh), Dense(hcat(3.3), false), Dropout(0.4))
+Trio(Dense(2 => 1, tanh), Dense(1 => 1; bias=false), Dropout(0.4))
+
+julia> Flux.destructure(tri)  # parameters are not yet visible to Flux
+(Bool[], Restructure(Trio, ..., 0))
+
+julia> Flux.@layer :expand Trio
+
+julia> Flux.destructure(tri)  # now gpu, params, train!, etc will see inside too
+([1.1, 2.2, 0.0, 3.3], Restructure(Trio, ..., 4))
+
+julia> tri  # and layer is printed like Chain
+Trio(
+  Dense(2 => 1, tanh),                  # 3 parameters
+  Dense(1 => 1; bias=false),            # 1 parameters
+  Dropout(0.4),
+)                   # Total: 3 arrays, 4 parameters, 224 bytes.
+```
+
+"""
+macro layer(exs...)
+  out = quote end
+
+  # These functions are defined in show.jl, and each return an expression overloading Base.show
+  type, rest... = if exs[1] == QuoteNode(:expand)
+    push!(out.args, _macro_big_show(esc(exs[2])))  
+    exs[2:end]
+  elseif exs[1] == QuoteNode(:ignore)
+    exs[2:end]
+  elseif exs[1] isa QuoteNode
+    error("`@layer` accepts only two options before the layer type, `:expand` and `:ignore` (to control `show`)")
+  else
+    push!(out.args, _macro_layer_show(esc(exs[1])))
+    exs
+  end
+
+  # This function exists only for depwarns when you use @functor directly
+  push!(out.args, :(Flux._check_new_macro(::$(esc(type))) = nothing))
+  
+  i = findfirst(ex -> Meta.isexpr(ex, :(=)) && ex.args[1] == :children, rest)
+  if isnothing(i)  # then default like @functor Layer
+    push!(out.args, _macro_functor(esc(type)))
+  else
+    push!(out.args, _macro_functor(esc(type), rest[i].args[2]))
+  end
+  for j in 1:length(rest)
+    j == i && continue
+    ex = rest[j]
+    Meta.isexpr(ex, :(=)) || error("The macro `@layer` expects here `keyword = (fields...,)`, got $ex")
+
+    name = if ex.args[1] == :trainable
+      :(Optimisers.trainable)
+    elseif ex.args[1] == :functor
+      error("Can't use `functor=(...)` as a keyword to `@layer`. Use `childen=(...)` to define a method for `functor`.")
+    else
+      error("`@layer` cannot define a method for `$(ex.args[1])` at the moment, sorry.")
+      # @warn "Trying to define a method for `$(ex.args[1])` in your scope... this is experimental" maxlog=1
+      # esc(ex.args[1])
+    end
+    push!(out.args, _macro_trainable(esc(type), name, ex.args[2]))
+  end
+
+  out
+end
+
+# Temporary depwarn function, called within `params`, is also called by `show`.
+
+function _check_new_macro(x::T) where T
+  Functors.isleaf(x) && return
+  Base.depwarn("This type should probably now use `Flux.@layer` instead of `@functor`: $T", Symbol("@functor"))
+end
+_check_new_macro(::Tuple) = nothing  # defined by Functors.jl, not by users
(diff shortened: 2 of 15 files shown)


src/layers/show.jl
```diff
@@ -88,7 +88,7 @@ end
 _show_leaflike(x) = isleaf(x)  # mostly follow Functors, except for:
 _show_leaflike(::Tuple{Vararg{<:Number}}) = true         # e.g. stride of Conv
 _show_leaflike(::Tuple{Vararg{<:AbstractArray}}) = true  # e.g. parameters of LSTMcell
-_show_leaflike(::Diagonal) = true                        # appears inside LayerNorm
+_show_leaflike(::Scale) = true                        # appears inside LayerNorm
```
mcabbott, line 91: This change should have been in #1927 but I missed it. On master:
```
julia> Chain(LayerNorm(2))
Chain(
  LayerNorm(
    identity,
    Scale(2),                           # 4 parameters
    1.0f-5,
    (2,),
    true,
  ),
) 
```
vs. with this fix
```
julia> Chain(LayerNorm(2))
Chain(
  LayerNorm(2),                         # 4 parameters
) 
```

ToucheSir:
Looks reasonable, so if you'd let me indulge in some light bikeshedding how about `@show_expanded` over `@big_show`?

mcabbott:
Not at all set on the name. 

Another question is whether there should be a matching way to disable recursion. Overloading `_show_leaflike(::Diagonal)` to make LayerNorm not expand is slightly weird. It's a hack to re-use the fact that layers all of whose children area leaflike aren't expanded, and has side-effects like this:
```
julia> Chain(Flux.Scale(2), Flux.Scale(2))
Chain(Scale(2), Scale(2))  # 8 parameters

julia> Chain(Flux.Scale(2), Flux.Scale(2), Dense(2=>2))
Chain(
  Scale(2),                             # 4 parameters
  Scale(2),                             # 4 parameters
  Dense(2 => 2),                        # 6 parameters
)                   # Total: 6 arrays, 14 parameters, 440 bytes.
```
If this needs a public face, then perhaps it should be more like `@show_noexpand LayerNorm` i.e. a property of the container, not the contents.

CarloLucibello:
`show_leaflike(x::MyType) = true` allowing to customize on the specific object seems more flexible and it feels ok to me as an interface, I would prefer that over a macro. 

We could also merge this PR now since it seems ready and expose that interface later

mcabbott:
[6370374](https://github.com/FluxML/Flux.jl/pull/1932/commits/6370374714605bd69558d484bc120cd6fa53d129) upgrades this to a replacement for `@functor`. The idea is to have just one magic macro do several things for us. Discussed at  https://github.com/FluxML/Flux.jl/pull/2028 as an alternative to the supertype idea there. 

Still WIP, but roughly works, I think.

src/layers/macro.jl
```diff
@@ -0,0 +1,177 @@
...
+This macro replaces most uses of `@functor` in Flux 0.14. Its basic purpose is the same:
+When you define a new layer, this tells Flux to explore inside it
+to see the parameters it trains, and also to move them to the GPU, change precision, etc.
+
+Some "keywords" allow control of the recursion:
+* If some fields look like parameters but should not be trained,
+  then `Optimisers.trainable` lets you specify fields to include, and ignore the rest.
+* We can likewise add restructions to `Functors.functor`, but not yet written.
+* In fact you can provide an arbitrary keyword with this syntax, and it will
+  overload this function alla `trainable`... that might be a terrible idea.
```
ToucheSir, line 17: As a start, we could limit the allowable keyword list.

src/layers/macro.jl
```diff
@@ -0,0 +1,177 @@
+
+"""
+    @layer Dense
+    @layer :expand Chain
+    @layer BatchNorm trainable=(β,γ)
+    @layer Struct functor=(α,β) trainable=(β,)
```
ToucheSir, line 6: While we're at it, perhaps `functor` could be aliased to something less obscure? `params` might be a name conflict but is the most familiar (perhaps `parameters`? Too cheeky?). `buffers` is close to PyTorch but still kind of obscure.

darsnack: Why not `leaves` or `children`?

src/layers/macro.jl
mcabbott: Yes. Nice to pick a notation that can grow if desired though.

src/layers/macro.jl
mcabbott: Could do. I guess I wanted to match the function name, and if this one is obscure, that's OK, you should almost never use it. `children` does match *a* function name.

ToucheSir: `children` also has the benefit of having the same signature of trainable: `() -> Union{Tuple,NamedTuple,Array,...}` instead of  `() -> Tuple{Union{Tuple,NamedTuple,Array,...}, Function}` for `functor`.

mcabbott: Latest version uses `children` as the keyword.

src/layers/macro.jl
```diff
@@ -0,0 +1,171 @@
...
+
+  out
+end
+
+# Temporary depwarn function, called within `params`, is also called by `show`.
+
+function _check_new_macro(x::T) where T
+  Functors.isleaf(x) && return
+  Base.depwarn("This type should probably now use `Flux.@layer` instead of `@functor`: $T", Symbol("@functor"))
+end
```
mcabbott, line 101: I toned this warning down from `@warn`. I believe this makes the PR non-breaking, could be in a 0.13 release. Possibly 0.14 should always warn?

src/layers/macro.jl
```diff
@@ -0,0 +1,171 @@
...
+    j == i && continue
+    ex = rest[j]
+    Meta.isexpr(ex, :(=)) || error("The macro `@layer` expects here `keyword = (fields...,)`, got $ex")
+
+    name = if ex.args[1] == :trainable
+      :(Optimisers.trainable)
+    elseif ex.args[1] == :functor
+      error("Can't use `functor=(...)` as a keyword to `@layer`. Use `childen=(...)` to define a method for `functor`.")
+    else
+      @warn "Trying to define a method for `$(ex.args[1])` in your scope... this is experimental" maxlog=1
```
mcabbott, line 87: Tests check that `@layer Trio trainable=(a,b) test=(c,)` works, although noisily. Not entirely sure it should stay in.

src/layers/macro.jl
```diff
@@ -0,0 +1,171 @@
...
+   end
+end
+
+function _custom_functor(::Type{T}, x, ::Val{which}) where {T,which}
+   if false
+     # TODO write the @generated version
+   else
+     remake(nt) = Base.typename(T).wrapper(map(f -> f in which ? getfield(nt, f) : getfield(x, f), fieldnames(T))...)
+     NamedTuple{which}(map(s -> getfield(x, s), which)), remake
+   end
```
mcabbott, line 147: The path when you specify `children` isn't as fast. Maybe this doesn't matter for now, since no layer within Flux does so.

src/layers/normalise.jl
```diff
@@ -352,8 +352,7 @@ function BatchNorm(chs::Int, λ=identity;
             nothing, chs)
 end
 
-@functor BatchNorm
-trainable(bn::BatchNorm) = hasaffine(bn) ? (β = bn.β, γ = bn.γ) : (;)
+@layer BatchNorm trainable=(β,γ)
```
mcabbott, line 355: Normalisation layers all have permanent `trainable`, but the fields are `nothing` when not in use, should be safe?

src/layers/recurrent.jl
```diff
@@ -207,7 +206,7 @@ function (m::RNNCell{F,A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T}
   return h, reshape_cell_output(h, x)
 end
 
-@functor RNNCell
+@layer RNNCell  # trainable=(Wi, Wh, b)
```
mcabbott, line 209: Why is `state0` included in RNNCell's trainable fields? I didn't try but perhaps it should not be?

src/layers/show.jl
```diff
@@ -1,22 +1,30 @@
+@nospecialize  # just for this file, for startup time
```
mcabbott, line 1: Sometimes show is visibly slow on first loading. This doesn't really help, a test with `@time show(stdout, MIME"text/plain"(), model);` goes from 2.28s to 2.15s.

docs/src/models/basics.md
```diff
@@ -216,7 +216,7 @@ m(5) # => 26
 Flux provides a set of helpers for custom layers, which you can enable by calling
 
 ```julia
-Flux.@functor Affine
+Flux.@layer Affine
```
mcabbott, line 219: "set of helpers for custom layers" is more accurate after than before. Where else would docs need changing?

src/layers/macro.jl
mcabbott: A related question. Currently many people use `Flux.@functor`. We could define such a macro as a pass-through to `@layer`, which will print a depwarn once on the macro call. Should we?

src/layers/macro.jl
ToucheSir: Can we error here too for now? I feel like letting the macro out for a while in the wild before loosening the restriction would be prudent.

docs/src/models/basics.md
ToucheSir: https://github.com/FluxML/Flux.jl/blob/master/docs/src/models/advanced.md is the big one, off the top of my head.

src/layers/recurrent.jl
ToucheSir: There was a whole debate about this some years back and IIRC the resulting impasse favoured the status quo. Let me see if I can find that issue.

src/layers/normalise.jl
ToucheSir: Yes. Really, the affine params should be wrapped up in a `Scale` sublayer like LayerNorm has, but that's a PR for another day.

src/layers/macro.jl
ToucheSir: This may be why Functors.jl uses `@eval`. The generated version seems fine though.

src/layers/recurrent.jl
mcabbott: I see, thanks. Certainly not this PR's can of worms.

mcabbott:
```suggestion
@layer RNNCell  # state0 is trainable, see issue 807 about this.
```

src/layers/macro.jl
mcabbott: Done.

ToucheSir:
Doctest failure looks real, but I'm not sure why it's only showing up now.

docs/src/models/basics.md
mcabbott: I had a go at editing that. It's quite a crazy scheme we have going, where `trainable`, `functor`, `adapt` define 3 different recursions. Do we really need to ways to customise `functor`?

mcabbott:
Doc failure seems to be that this doesn't work on 1.6, where the first argument is a vector:
```
julia> Dense([3.3;;], false)
Dense(1 => 1; bias=false)  # 1 parameters
```

mcabbott:
Failures look unrelated:

* GeometricFlux.jl I see only `create_bias` failures, https://github.com/FluxML/Flux.jl/issues/2049
* FastAI.jl has failures like "Expression: isempty(stderr_content) [1105] Evaluated: isempty("┌ Warning: `ignore(f)` is deprecated, use `ChainRulesCore.ignore_derivatives(f)` instead.\n│   caller =" 
* AtomicGraphNets.jl has failures like "MethodError: no method matching needs_concrete_A(::Nothing)  Closest candidates are: needs_concrete_A(::LinearSolve.AbstractFactorization) at ~/.julia/packages/LinearSolve/AoYJI/src/LinearSolve.jl:34" 
* OperatorLearning.jl has `create_bias` only

ToucheSir:
So just the doctest fixup for 1.6?

src/layers/macro.jl
mcabbott: For sure it can be done with a `@generated` block. I was just lazy to actually write and check and benchmark it.

darsnack: Can this be done with `@generated`? It seems you can't define closures in generated functions, and we need to close over `x`.

mcabbott: I guess that's true. But perhaps it can instead return a struct like `Remake{(:some, :fields)}(x, constructor)` which gets some fields from `x` and some from what it's called on. But I didn't try.

mcabbott: I'm also horrified by the number of ways to customise recursive behaviour we have, and I wonder if this one should just be forbidden.

darsnack: Clever, let me try that.

darsnack: Wait isn't returning a `Remake` struct the same as returning a closure (in terms of performance not whether it can be in a generated function)? I thought the benefit here was that you can manually write out the iteration accessing the fields.

mcabbott: Yes I would guess so. If necc. its call could itself be a generated thing, but quite likely some ntuple / map would just compile down. 

Not precisely sure what list of fields this thing ought to be passed, the ones you get or the ones you don't or some mix such that `merge` sorts it out?

mcabbott:
Rebased in 2024. We should just do this, right?

The biggest questions is: Should this support children at all? Or can we simply forbid that... do we really need 4 ways to customise things, alla #2385? Maybe if you want to hack things with Functors.jl directly you are welcome, but Flux.jl can pretend this doesn't exist.

darsnack:
I agree. Given that you almost never want to set `children`, having to support and explain `children` vs `trainable` is just confusing for users. Anyone wanting to restrict `children` probably needs to take a minute to understand `@functor` first.

mcabbott:
OK, `children` is gone. 

This should be ready to merge, I think. As always there may be further cleaning up of the docs possible.

PR #1942, opened 2022-04-12
Fix type instabilities in apply!(optimizer, ...)

ancapdev:
Minor change to fix type instabilities in optimisers.

Ran into this when trying to reduce allocations during training (GC is my biggest obstacle to scalable multi-threaded fitting of small models at the moment). In particular the `RADAM` `apply!` function was yielding 24 allocations in my test primarily due to instability on `t`, which was being asserted to an abstract type.

### DIFF
--- a/src/optimise/optimisers.jl
+++ b/src/optimise/optimisers.jl
@@ -168,7 +168,7 @@ mutable struct ADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
   epsilon::Float64
-  state::IdDict
+  state::IdDict{Any, Any}
 end
 ADAM(η::Real = 0.001, β::Tuple = (0.9, 0.999), ϵ::Real = EPS) = ADAM(η, β, ϵ, IdDict())
 ADAM(η::Real, β::Tuple, state::IdDict) = ADAM(η, β, EPS, state)
@@ -210,7 +210,7 @@ mutable struct RADAM <: AbstractOptimiser
   eta::Float64
   beta::Tuple{Float64,Float64}
   epsilon::Float64
-  state::IdDict
+  state::IdDict{Any, Any}
 end
 RADAM(η::Real = 0.001, β::Tuple = (0.9, 0.999), ϵ::Real = EPS) = RADAM(η, β, ϵ, IdDict())
 RADAM(η::Real, β::Tuple, state::IdDict) = RADAM(η, β, EPS, state)
@@ -221,7 +221,7 @@ function apply!(o::RADAM, x, Δ)
 
   mt, vt, βp, t = get!(o.state, x) do
       (zero(x), zero(x), Float64[β[1], β[2]], Ref(1))
-  end :: Tuple{typeof(x),typeof(x),Vector{Float64},Ref{Int}}
+  end :: Tuple{typeof(x),typeof(x),Vector{Float64},Base.RefValue{Int}}
 
   @. mt = β[1] * mt + (1 - β[1]) * Δ
   @. vt = β[2] * vt + (1 - β[2]) * Δ * conj(Δ)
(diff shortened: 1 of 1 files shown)


DhairyaLGandhi:
by making it infer as `Any` might cause instability for the full optimisation pipeline. Could you test a before and after against a basic inferred training loop? Perhaps grabbing the one from the optimization tests would be an option.

DhairyaLGandhi:
Yes actually. In the case for a multi node training scheme, using Params and Grads is pretty clunky (due to the IdDict) so I had to pass regular dicts instead. The final version doesn't have this limitation though, but there are reasons to not use the default IdDict.

ancapdev:
I see. In the case of optimisers they don't accept anything except `IdDict` for their state as it is, so is there anything that needs to change with my changes here?

ToucheSir:
The RefValue change looks good to me, thanks! Can you provide a couple of benchmarks for the IdDict changes?

ancapdev:
Impact is constant regardless of input size, so in practice mostly applicable to small parameters. A simple benchmark:

```Julia
for N in [8, 16, 32]
    W = rand(Float32, N, N)
    Δ = rand(Float32, N, N) ./ 1000
    opt = RADAM()
    @btime Flux.Optimise.apply!($opt, $W, $Δ)
end
```

### Baseline
```
  1.347 μs (23 allocations: 560 bytes)
  2.782 μs (23 allocations: 560 bytes)
  8.546 μs (23 allocations: 560 bytes)
```

### RefValue
```
  630.219 ns (1 allocation: 32 bytes)
  2.068 μs (1 allocation: 32 bytes)
  7.768 μs (1 allocation: 32 bytes) 
```

### IdDict{Any, Any}
```
  584.508 ns (0 allocations: 0 bytes)
  2.032 μs (0 allocations: 0 bytes)
  7.722 μs (0 allocations: 0 bytes)
```

Most of the benefit comes from fixing the `Ref` type instability, much more marginal from fixing `state`, although it's hard to say from micro-benchmarks like this. You can generate the most amazing codegen in the world, but when you run across a large number of threads with the current Julia GC you very quickly become GC bottlenecked. I'm just trying to eliminate all unnecessary allocations in my fitting loop.

ToucheSir:
Right, so as I mentioned the RefValue change is a definite shoe-in. For the state IdDict, there's the question of whether there would be any benefit of passing one with a narrower type at optimizer construction. Likewise, some types may be able to constrain the upper type bound of their state dict further. e.g. `InvDecay`'s state is a single Int and the state dict type upper bound could be `IdDict{Any,Int}` instead.

ancapdev:
I think a narrower type for the dict value is in general not going to help if you still need to type assert the value on extraction (if it depends on `x`). `InvDecay` is a good example to the counter though, and I can see that one isn't even type asserted at the moment. Shall I narrow the `state` type there?

ToucheSir:
> I think a narrower type for the dict value is in general not going to help if you still need to type assert the value on extraction (if it depends on `x`)

Assuming the type of all parameters and the input `x` are consistent (say `Matrix{Float32}`), then you can provide a more narrowly-typed IdDict upon optimizer construction and the type assertion should compile away as a no-op. I don't anticipate this would happen super often, but @CarloLucibello and @darsnack might have some thoughts from their real-world usage.

One more practical concern is whether or not constraining the IdDict type messes with deserializing optimizer state from prior Flux versions. I suspect the answer is probably no, but if you wouldn't mind testing that would be great.

ancapdev:
> One more practical concern is whether or not constraining the IdDict type messes with deserializing optimizer state from prior Flux versions. I suspect the answer is probably no, but if you wouldn't mind testing that would be great.

I tested the change to `RADAM` and `InvDecay` with native serialization and `BSON.jl`, both work fine.

Good point if you know the exact type of all parameters. The type assert is very cheap though.

ToucheSir (approved):
Hopefully most of this will be obselete soon since Optimisers.jl (rightfully) does not try to put every parameter + state into a homogeneously-typed dict, but until then an improvement is an improvement :)

PR #1945, opened 2022-04-13
Add Base.firstindex(c::Chain) = 1

KronosTheLate:
I see no reason not to have this, and I ran into missing it, as I have made a habit of indexing with `begin` as opposed to 1.

I am not sure - should this be added to NEWS.md? It seems too minor for it...
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -68,7 +68,6 @@ end
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
   Chain(NamedTuple{Base.keys(c)[i]}(Tuple(c.layers)[i]))
-Base.firstindex(c::Chain) = 1
 function Base.show(io::IO, c::Chain)
   print(io, "Chain(")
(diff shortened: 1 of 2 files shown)


src/layers/basic.jl
```diff
@@ -68,7 +68,7 @@ end
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
   Chain(NamedTuple{Base.keys(c)[i]}(Tuple(c.layers)[i]))
-
+Base.firstindex(c::Chain) = 1
```
ToucheSir, line 71:
```suggestion
```
This line shouldn't be necessary with the forwarded definition above.

KronosTheLate: Cool! I have never seen forwarding before, but I trust you got it right.

ToucheSir:
Do you mind adding a test or two for begin/end while you're at it? I don't believe we currently check the latter either.

ToucheSir (approved):
This is a nice little QoL change, thanks!

Issue #1947, opened 2022-04-19
Type instability in Recur for 3 dimensional arrays


MarcoVela:
While working with LSTM I noticed that Recur is not type stable for 3 dimensional arrays.

If we execute:
```jl
using Flux
model = LSTM(1=>1)
x = rand(Float32, 1, 1, 1)
@code_warntype model(x)
```
Result:
```jl
MethodInstance for (::Flux.Recur{Flux.LSTMCell{Matrix{Float32}, Vector{Float32}, Tuple{Matrix{Float32}, Matrix{Float32}}}, Tuple{Matrix{Float32}, Matrix{Float32}}})(::Array{Float32, 3})
  from (m::Flux.Recur)(x::AbstractArray{T, 3}) where T in Flux at ~\.julia\packages\Flux\18YZE\src\layers\recurrent.jl:88
Static Parameters
  T = Float32
Arguments
  m::Flux.Recur{Flux.LSTMCell{Matrix{Float32}, Vector{Float32}, Tuple{Matrix{Float32}, Matrix{Float32}}}, Tuple{Matrix{Float32}, Matrix{Float32}}}
  x::Array{Float32, 3}
Locals
  #269::Flux.var"#269#270"{Flux.Recur{Flux.LSTMCell{Matrix{Float32}, Vector{Float32}, Tuple{Matrix{Float32}, Matrix{Float32}}}, Tuple{Matrix{Float32}, Matrix{Float32}}}}
  sze::Any
  h::Any
Body::Any
1 ─ %1  = %new(Flux.var"#269#270"{Flux.Recur{Flux.LSTMCell{Matrix{Float32}, Vector{Float32}, Tuple{Matrix{Float32}, Matrix{Float32}}}, Tuple{Matrix{Float32}, Matrix{Float32}}}}, m)::Flux.var"#269#270"{Flux.Recur{Flux.LSTMCell{Matrix{Float32}, Vector{Float32}, Tuple{Matrix{Float32}, Matrix{Float32}}}, Tuple{Matrix{Float32}, Matrix{Float32}}}}
│   %2  = invoke Base.var"#eachslice##kw"()($(QuoteNode((dims = 3,)))::NamedTuple{(:dims,), Tuple{Int64}}, Flux.eachslice::typeof(eachslice), x::Array{Float32, 3})::Base.Generator{Base.OneTo{Int64}}
[...]

PR #1948, opened 2022-04-20
Recur type stability for 3d arrays

MarcoVela:
Fixes https://github.com/FluxML/Flux.jl/issues/1947

Defines the function eachlastdim (we can change the name for something more descriptive) that returns a generator over the last dimension of an Array, it also defines its corresponding rrule using ChainRule.jl [∇eachslice](https://github.com/JuliaDiff/ChainRules.jl/blob/f13e0a45d10bb13f48d6208e9c9d5b4a52b96732/src/rulesets/Base/indexing.jl#L77)

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -18,6 +18,17 @@ function ChainRulesCore.rrule(::typeof(multigate), x::AbstractArray, h, c)
   return multigate(x, h, c), multigate_pullback
 end
 
+# Type stable and AD-friendly helper for iterating over the last dimension of an array
+function eachlastdim(A::AbstractArray{T,N}) where {T,N}
+  inds_before = ntuple(Returns(:), N-1)
+  return (view(A, inds_before..., i) for i in axes(A, N))
+end
+
+function ChainRulesCore.rrule(::typeof(eachlastdim), x::AbstractArray{T,N}) where {T,N}
+  lastdims(dy) = (NoTangent(), Zygote.ChainRules.∇eachslice(unthunk(dy), x, Val(N)))
+  collect(eachlastdim(x)), lastdims
+end
+
 reshape_cell_output(h, x) = reshape(h, :, size(x)[2:end]...)
 
 # Stateful recurrence
@@ -86,7 +97,7 @@ reset!(m) = foreach(reset!, functor(m)[1])
 flip(f, xs) = reverse([f(x) for x in reverse(xs)])
 
 function (m::Recur)(x::AbstractArray{T, 3}) where T
-  h = [m(x_t) for x_t in eachslice(x, dims=3)]
+  h = [m(x_t) for x_t in eachlastdim(x)]
   sze = size(h[1])
   reshape(reduce(hcat, h), sze[1], sze[2], length(h))
 end
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -20,7 +20,7 @@ end
 
 # Type stable and AD-friendly helper for iterating over the last dimension of an array
 function eachlastdim(A::AbstractArray{T,N}) where {T,N}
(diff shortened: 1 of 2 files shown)


src/layers/recurrent.jl
```diff
@@ -18,6 +18,17 @@ function ChainRulesCore.rrule(::typeof(multigate), x::AbstractArray, h, c)
...
 end
 
+# Type stable and AD-friendly helper for iterating over the last dimension of an array
+function eachlastdim(A::AbstractArray{T,N}) where {T,N}
+  inds_before = ntuple(_ -> :, N-1)
+  return (view(A, inds_before..., i) for i in axes(A, N))
+end
+
+function ChainRulesCore.rrule(::typeof(eachlastdim), x::AbstractArray{T,N}) where {T,N}
+  lastdims(dy) = (NoTangent(), Zygote.ChainRules.∇eachslice(unthunk(dy), x, Val(N)))
```
ToucheSir, line 28: One thing we've been trying with Flux is to avoid relying on anything Zygote-specific. In this case, the options to do so would be:
1. Take a direct dep on ChainRules. That seems suboptimal since doing so is really the responsibility of the AD, but it would be the fewest LOC.
2. Copy over https://github.com/JuliaDiff/ChainRules.jl/blob/f13e0a45d10bb13f48d6208e9c9d5b4a52b96732/src/rulesets/Base/indexing.jl#L77. This would be suboptimal because we're duplicating code, but it might also be an opportunity to simplify the rrule since we don't have to worry about `dims != ndims(x)`.

I'm leaning towards option 2. Would that work for you?

MarcoVela: Don't like the idea of code duplication, but until a fix to eachslice comes out I think option 2 works for us.

test/layers/recurrent.jl
```diff
@@ -106,3 +106,17 @@ end
   @test res == sum(x[1:2, :]) + 2sum(x[5:6, :])
   @test dx == [ones(2, 5); zeros(2, 5); fill(2, 2, 5)]
 end
+
+@testset "eachlastdim" begin
```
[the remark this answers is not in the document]
MarcoVela: sure, should I add a test for the RNNs too?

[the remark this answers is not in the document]
MarcoVela: It´s my first time using `@inferred` for testing, do these look okay to you?

src/layers/recurrent.jl
```diff
@@ -18,6 +18,37 @@ function ChainRulesCore.rrule(::typeof(multigate), x::AbstractArray, h, c)
...
+      else
+          copyto!(slice, dys[i])
+      end
+  end
+  return ProjectTo(x)(dx)
+end
+
+function ChainRulesCore.rrule(::typeof(eachlastdim), x::AbstractArray{T,N}) where {T,N}
+  lastdims(dy) = (NoTangent(), ∇eachlastdim(unthunk(dy), x))
+  collect(eachlastdim(x)), lastdims
```
CarloLucibello, line 49: why this  collect here?

MarcoVela: That's how it was implemented originally in [ChainRules](https://github.com/JuliaDiff/ChainRules.jl/blob/f13e0a45d10bb13f48d6208e9c9d5b4a52b96732/src/rulesets/Base/indexing.jl#L66), I think it's because the pullback excepts a vector in order to index into it.

ToucheSir (approved):
LGTM, thanks @MarcoVela!

MarcoVela:
> LGTM, thanks @MarcoVela!

Thanks a lot, I'm using Flux for my thesis and it makes me happy to be able to contribute to it

PR #1951, opened 2022-04-24
Resolve two warnings in the test suite

mcognetta:
This PR resolves two warnings in the test suite by updating outdated or discouraged code.

The errors are:

```julia
┌ Warning: Broadcasting is not safe to use with RNNs, as it does not guarantee an iteration order.
│ Re-writing this as a comprehension would be better.
│   caller = ip:0x0
└ @ Core :-1
┌ Warning: Flux.Diagonal is now Flux.Scale, and also allows an activation function.
│   caller = macro expansion at outputsize.jl:17 [inlined]
└ @ Core ~/.julia/packages/Flux/18YZE/test/outputsize.jl:17
```

### DIFF
--- a/test/layers/recurrent.jl
+++ b/test/layers/recurrent.jl
@@ -5,7 +5,7 @@
     rnn = r(2 => 3)
     Flux.reset!(rnn)
     grads_seq = gradient(Flux.params(rnn)) do
-        sum(rnn.(seq)[3])
+        sum([rnn(s) for s in seq][3])
     end
     Flux.reset!(rnn);
     bptt = gradient(Wh -> sum(tanh.(rnn.cell.Wi * seq[3] + Wh *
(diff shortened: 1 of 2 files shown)


test/layers/recurrent.jl
```diff
@@ -5,7 +5,7 @@
     rnn = r(2 => 3)
     Flux.reset!(rnn)
     grads_seq = gradient(Flux.params(rnn)) do
-        sum(rnn.(seq)[3])
+        sum([rnn(s) for s in seq][3])
```
DhairyaLGandhi, line 8: We definitely need benchmarks because this can cause include extra frames in AD since this would now be parsed as a generator/ loop. Maybe the warning can be relaxed? We do need to make sure we are getting the right results though. Broadcasting can have undefined execution order in some cases (this case should be safe though)

CarloLucibello: this is just  a test

CarloLucibello (approved):

PR #1953, opened 2022-04-24
Update documentation on Split layer

JLDC:
Minimal update to the documentation. Removes `tuple(...)` from the output of the [`Split` layer](https://fluxml.ai/Flux.jl/stable/models/advanced/#Multiple-outputs:-a-custom-Split-layer). The output is already a tuple in the desired form.

### DIFF
--- a/docs/src/models/advanced.md
+++ b/docs/src/models/advanced.md
@@ -196,7 +196,7 @@ Split(paths...) = Split(paths)
 
 Flux.@functor Split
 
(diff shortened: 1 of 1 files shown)


CarloLucibello (approved):

darsnack:
It could be a vector depending on how the constructor was called. I would restrict the type `T` to be a subtype of `Tuple`. A follow up PR would be appreciated if you don't mind!

PR #1954, opened 2022-04-25
Training mod

XWHtorrentx:
Add a progress bar with display of training loss in train!(...). For training less than 1 batch there is no display of progress bar and training loss.


darsnack:
Thanks for this contribution!

If I understand this correctly, this specifically uses ProgressMeter.jl as the front end for progress logging? Our current approach uses ProgressLogging.jl with integrates with VS Code and offers a choice of front ends. One advantage to ProgressMeter.jl is the ability to print the speed for infinite length sequences. We would need to discuss whether this is worth the change.

PR #1955, opened 2022-04-27
[docs] suggest using ADAM with LR=1 when combined with ExpDecay

ericphanson:
This only tweaks the example in the `ExpDecay` documentation.

The total LR is the product of the ADAM's learning rate with the current ExpDecay LR, which is kind of non-intuitive. By setting the ADAM's LR to 1, all the LR happens in the ExpDecay.


cossio:
Actually I'd prefer that `ExpDecay` be the one with lr = 1 ...

ToucheSir (approved):
Hopefully we'll be able to offer https://github.com/darsnack/ParameterSchedulers.jl or a similar library as an officially blessed + documented alternative before then :)

PR #1959, opened 2022-05-06
onehotbatch with CuArray

CarloLucibello:
Not a great solution but this used to error before (when scalar indexing is deactivated) due to the following operation:
```julia
function _onehotbatch(data, labels)
      indices = UInt32[something(_findval(i, labels), 0) for i in data]
      ...
```
so I guess going to cpu and back to gpu is strictly better.

### DIFF
--- a/src/onehot.jl
+++ b/src/onehot.jl
@@ -185,6 +185,8 @@ julia> reshape(1:15, 3, 5) * oh  # this matrix multiplication is done efficientl
 """
 onehotbatch(data, labels, default...) = _onehotbatch(data, length(labels) < 32 ? Tuple(labels) : labels, default...)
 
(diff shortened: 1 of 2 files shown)


ToucheSir (approved):
I can't think of a great way of doing this either. If anyone does, we can test it out in OneHotArrays.jl first.

MilkshakeForReal:
What about this?
```julia
 onehotbatch(data::CuArray,labels)= cu(labels).==reshape(data, 1,size(data)...)
```
soure: https://discourse.julialang.org/t/all-the-ways-to-do-one-hot-encoding/64807/9

ToucheSir:
That defeats the purpose of having a one-hot array type in the first place. If you can find a way to preserve the outer wrapper and use less memory than a dense matrix, we would be very interested.

MilkshakeForReal:
I will leave it to the smart ones

PR #1965, opened 2022-05-16
Explain `stride≠1` case for SamePad

KronosTheLate:
This PR adds a line to explain what `SamePad()` does when stride is not equal to 1.


ToucheSir (approved):
LGTM, thanks!

PR #1968, opened 2022-05-21
fast sigmoid

### LINKED ISSUE #1967: sigmoid_fast in GRU?
Hi!
Is there any particular reason that `sigmoid_fast` is not used in the helper function `_gru_output` in the GRU layer in reccurent.jl? I see it used in the LSTM layer.
```
function _gru_output(gxs, ghs, bs)
  r = @. σ(gxs[1] + ghs[1] + bs[1])
  z = @. σ(gxs[2] + ghs[2] + bs[2])
  return r, z
end
```
In my (not so simple) use case i got a 5% speed up in the forward step using `sigmoid_fast` on cpu compared to the above version.

oysteinsolheim:
Fixed https://github.com/FluxML/Flux.jl/issues/1967

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -281,8 +281,8 @@ Recur(m::LSTMCell) = Recur(m, m.state0)
 # GRU
 
 function _gru_output(gxs, ghs, bs)
-  r = @. σ(gxs[1] + ghs[1] + bs[1])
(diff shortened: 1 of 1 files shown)


ToucheSir (approved):
Thanks!

PR #1971, opened 2022-05-25
`PairwiseFusion` layer

theabhirath:
This PR implements a `PairwiseFusion` layer similar to the one in [Lux](https://github.com/avik-pal/Lux.jl/blob/0ca5a265b7ef8c6d3da2b2dfacfefc56a39f5163/src/layers/basic.jl#L335-L388). This might be especially useful for implementing models like [Res2Net](https://github.com/FluxML/Metalhead.jl/pull/123), since the forward pass for the bottleneck layer is quite similar to this.

### TODO:

- [x] Add pretty-printing similar to `Parallel`

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -67,7 +67,7 @@ end
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
-  Chain(NamedTuple{Base.keys(c)[i]}(Tuple(c.layers)[i]))
+  Chain(NamedTuple{keys(c)[i]}(Tuple(c.layers)[i]))
 function Base.show(io::IO, c::Chain)
   print(io, "Chain(")
   _show_layers(io, c.layers)
@@ -566,35 +566,37 @@ end
 
 `PairwiseFusion` returns a tuple of length `N` with the output of each fusion ((`y1`, `y2`, ..., `yN`) in the example above).
 """
-struct PairwiseFusion{F, T<:Union{Tuple, NamedTuple}}
+struct PairwiseFusion{F, T <: NamedTuple}
   connection::F
   layers::T
 end
 
-PairwiseFusion(connection, layers...) = PairwiseFusion(connection, layers)
-function PairwiseFusion(connection; kw...)
-  layers = NamedTuple(kw)
-  if :layers in keys(layers) || :connection in keys(layers)
-    throw(ArgumentError("a Parallel layer cannot have a named sub-layer called `connection` or `layers`"))
-  end
-  isempty(layers) && return Parallel(connection, ())
-  return PairwiseFusion(connection, layers)
+function PairwiseFusion(connection, layers...)
+  names = ntuple(i -> Symbol("layer_$i"), length(layers))
+  return PairwiseFusion(connection, NamedTuple{names}(layers))
 end
 
 function (m::PairwiseFusion)(x::T) where {T}
-  nlayers = length(m.layers)
-  getinput(i) = T <: Tuple ? x[i] : x
-  if T <: Tuple
-    nx = length(x)
-    if nx != nlayers
-      throw(ArgumentError("PairwiseFusion with $nlayers layers takes $nlayers inputs, but got $nx inputs"))
-    end
+  lx = length(x)
+  N = length(m.layers)
+  if T <: Tuple && lx != N
+    throw(ArgumentError("PairwiseFusion with $N sub-layers can take one input or $N inputs, but got $lx inputs"))
   end
-  outputs = [m.layers[1](getinput(1))]
-  for i in 2:nlayers
-    push!(outputs, m.layers[i](m.connection(getinput(i), outputs[i - 1])))
-  end
-  return outputs
+  applypairwisefusion(m.layers, m.connection, x)
+end
+
+@generated function applypairwisefusion(layers::NamedTuple{names}, connection, x::T) where {names, T}
+  N = length(names)
+  y_symbols = [gensym() for _ in 1:(N + 1)]
+  getinput(i) = T <: Tuple ? :(x[$i]) : :x
+  calls = [:($(y_symbols[N + 1]) = $(getinput(1)))]
+  append!(calls,
+            [:($(y_symbols[i]) = layers[$i]($(y_symbols[N + 1]));
+               $(y_symbols[N + 1]) = connection($(y_symbols[i]), $(getinput(i + 1)))) 
+             for i in 1:N - 1])
+  push!(calls, :($(y_symbols[N]) = layers[$N]($(y_symbols[N + 1]))))
+  push!(calls, :(return $(y_symbols[N])))
+  return Expr(:block, calls...)
 end
 
 @functor PairwiseFusion
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -595,18 +595,23 @@ end
                $(y_symbols[N + 1]) = connection($(y_symbols[i]), $(getinput(i + 1)))) 
              for i in 1:N - 1])
   push!(calls, :($(y_symbols[N]) = layers[$N]($(y_symbols[N + 1]))))
-  push!(calls, :(return $(y_symbols[N])))
+  push!(calls, :(return tuple($(Tuple(y_symbols[1:N])...))))
   return Expr(:block, calls...)
 end
 
 @functor PairwiseFusion
 
 Base.getindex(m::PairwiseFusion, i) = m.layers[i]
-Base.getindex(m::PairwiseFusion, i::AbstractVector) = PairwiseFusion(m.connection, m.layers[i])
 Base.getindex(m::PairwiseFusion{<:Any, <:NamedTuple}, i::AbstractVector) =
   PairwiseFusion(m.connection, NamedTuple{keys(m)[i]}(Tuple(m.layers)[i]))
 
 Base.keys(m::PairwiseFusion) = keys(getfield(m, :layers))
+
+function Base.show(io::IO, m::PairwiseFusion)
+  print(io, "PairwiseFusion(", m.connection, ", ")
+  _show_layers(io, m.layers)
+  print(io, ")")
+end
 
 """
     Embedding(in => out; init=randn)
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -582,11 +582,15 @@ function PairwiseFusion(connection; kw...)
 end
 
 function (m::PairwiseFusion)(x::T) where {T}
-  getinput(i) = T <: Union{Tuple, Vector} ? x[i] : x
-  nx = length(x)
   nlayers = length(m.layers)
-  if nx != nlayers
-    throw(ArgumentError("PairwiseFusion with $nlayers layers takes $nlayers inputs, but got $nx inputs"))
+  if T <: Union{Tuple, Vector}
+    getinput(i) = x[i]
+    nx = length(x)
+    if nx != nlayers
+      throw(ArgumentError("PairwiseFusion with $nlayers layers takes $nlayers inputs, but got $nx inputs"))
+    end
+  else
+    getinput(i) = x
   end
   outputs = [m.layers[1](getinput(1))]
   for i in 2:nlayers
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -544,7 +544,7 @@ x1 --> layer1 --> y1
 
 This layer behaves differently based on input type:
 
-1. Input `x` is a tuple/vector of length `N`. Then `layers` must be a tuple of length `N`. The computation is as follows:
+1. Input `x` is a tuple of length `N`. Then `layers` must be a tuple of length `N`. The computation is as follows:
 
 ```julia
 y = x[1]
@@ -583,14 +583,12 @@ end
 
 function (m::PairwiseFusion)(x::T) where {T}
   nlayers = length(m.layers)
-  if T <: Union{Tuple, Vector}
-    getinput(i) = x[i]
+  getinput(i) = T <: Tuple ? x[i] : x
+  if T <: Tuple
     nx = length(x)
     if nx != nlayers
       throw(ArgumentError("PairwiseFusion with $nlayers layers takes $nlayers inputs, but got $nx inputs"))
     end
-  else
-    getinput(i) = x
   end
   outputs = [m.layers[1](getinput(1))]
   for i in 2:nlayers
@@ -644,7 +642,7 @@ end
 @functor Embedding
 
 Embedding((in, out)::Pair{<:Integer, <:Integer}; init = randn32) = Embedding(init(out, in))
-  
+
 (m::Embedding)(x::Integer) = m.weight[:, x]
 (m::Embedding)(x::AbstractVector) = NNlib.gather(m.weight, x)
 (m::Embedding)(x::AbstractArray) = reshape(m(vec(x)), :, size(x)...)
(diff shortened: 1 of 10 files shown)


src/layers/basic.jl
```diff
@@ -510,16 +510,102 @@ end
...
+    if nx != nlayers
+      throw(ArgumentError("PairwiseFusion with $nlayers layers takes $nlayers inputs, but got $nx inputs"))
+    end
+  end
+  outputs = [m.layers[1](getinput(1))]
+  for i in 2:nlayers
+    push!(outputs, m.layers[i](m.connection(getinput(i), outputs[i - 1])))
+  end
+  return outputs
+end
```
ToucheSir, line 598: I fear this will be very AD unfriendly. Lux uses a [generated function](https://github.com/avik-pal/Lux.jl/blob/0ca5a265b7ef8c6d3da2b2dfacfefc56a39f5163/src/layers/basic.jl#L403-L420)

src/layers/basic.jl
```diff
@@ -510,16 +510,102 @@ end
...
+struct PairwiseFusion{F, T<:Union{Tuple, NamedTuple}}
+  connection::F
+  layers::T
+end
+
+PairwiseFusion(connection, layers...) = PairwiseFusion(connection, layers)
+function PairwiseFusion(connection; kw...)
+  layers = NamedTuple(kw)
+  if :layers in keys(layers) || :connection in keys(layers)
+    throw(ArgumentError("a Parallel layer cannot have a named sub-layer called `connection` or `layers`"))
```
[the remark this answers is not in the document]
theabhirath: Ah, the perils of copy-pasting 🤦🏽

src/layers/basic.jl
theabhirath: Hmm, this does seem to be throwing up an error when I try a gradient with a very simple example:

```julia

julia> x = (rand(1, 10), rand(30, 10));

julia> model = PairwiseFusion(+,  Dense(1, 30),  Dense(30, 10));

julia> @benchmark Zygote.gradient(p -> sum($model(p)), $x)
ERROR: DimensionMismatch: dimensions must match: a has dims (Base.OneTo(30), Base.OneTo(10)), b has dims (Base.OneTo(10), Base.OneTo(10)), mismatch at 1
Stacktrace:
  [1] promote_shape
    @ ./indices.jl:178 [inlined]
  [2] promote_shape(a::Matrix{Float64}, b::Matrix{Float64})
    @ Base ./indices.jl:169
  [3] +(A::Matrix{Float64}, Bs::Matrix{Float64})
    @ Base ./arraymath.jl:14
  [4] adjoint
    @ ~/.julia/packages/Zygote/DkIUK/src/lib/array.jl:769 [inlined]
  [5] _pullback
    @ ~/.julia/packages/ZygoteRules/AIbCs/src/adjoint.jl:65 [inlined]
  [6] _pullback
    @ ~/Code/Flux.jl/src/layers/basic.jl:602 [inlined]
  [7] _pullback(ctx::Zygote.Context, f::PairwiseFusion{typeof(+), Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}, args::Tuple{Matrix{Float64}, Matrix{Float64}})
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/interface2.jl:0
  [8] _pullback
    @ ./REPL[8]:1 [inlined]
  [9] _pullback(ctx::Zygote.Context, f::var"#7#8"{PairwiseFusion{typeof(+), Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}}, args::Tuple{Matrix{Float64}, Matrix{Float64}})
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/interface2.jl:0
 [10] _pullback(f::Function, args::Tuple{Matrix{Float64}, Matrix{Float64}})
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/interface.jl:34
 [11] pullback(f::Function, args::Tuple{Matrix{Float64}, Matrix{Float64}})
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/interface.jl:40
 [12] gradient(f::Function, args::Tuple{Matrix{Float64}, Matrix{Float64}})
    @ Zygote ~/.julia/packages/Zygote/DkIUK/src/compiler/interface.jl:75
 [13] var"##core#346"(model#341::PairwiseFusion{typeof(+), Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}, x#342::Tuple{Matrix{Float64}, Matrix{Float64}})
    @ Main ~/.julia/packages/BenchmarkTools/7xSXH/src/execution.jl:489
 [14] var"##sample#347"(::Tuple{PairwiseFusion{typeof(+), Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}, Tuple{Matrix{Float64}, Matrix{Float64}}}, __params::BenchmarkTools.Parameters)
    @ Main ~/.julia/packages/BenchmarkTools/7xSXH/src/execution.jl:495
 [15] _run(b::BenchmarkTools.Benchmark, p::BenchmarkTools.Parameters; verbose::Bool, pad::String, kwargs::Base.Pairs{Symbol, Integer, NTuple{4, Symbol}, NamedTuple{(:samples, :evals, :gctrial, :gcsample), Tuple{Int64, Int64, Bool, Bool}}})
    @ BenchmarkTools ~/.julia/packages/BenchmarkTools/7xSXH/src/execution.jl:99
 [16] #invokelatest#2
    @ ./essentials.jl:801 [inlined]
 [17] #run_result#45
    @ ~/.julia/packages/BenchmarkTools/7xSXH/src/execution.jl:34 [inlined]
 [18] run(b::BenchmarkTools.Benchmark, p::BenchmarkTools.Parameters; progressid::Nothing, nleaves::Float64, ndone::Float64, kwargs::Base.Pairs{Symbol, Integer, NTuple{5, Symbol}, NamedTuple{(:verbose, :samples, :evals, :gctrial, :gcsample), Tuple{Bool, Int64, Int64, Bool, Bool}}})
    @ BenchmarkTools ~/.julia/packages/BenchmarkTools/7xSXH/src/execution.jl:117
 [19] #warmup#54
    @ ~/.julia/packages/BenchmarkTools/7xSXH/src/execution.jl:169 [inlined]
 [20] warmup(item::BenchmarkTools.Benchmark)
    @ BenchmarkTools ~/.julia/packages/BenchmarkTools/7xSXH/src/execution.jl:168
 [21] top-level scope
    @ ~/.julia/packages/BenchmarkTools/7xSXH/src/execution.jl:393
 [22] top-level scope
    @ ~/.julia/packages/CUDA/qAl31/src/initialization.jl:52
```

The forward pass works just fine, so not sure what's going wrong. I'll try with a generated function and see if that makes a difference

darsnack: I would recommend design specifically for the case where `x` is a tuple, then adding a method that turns non-tuples into tuples for the other cases.

theabhirath: So I've used a generated function - it's pretty much the same as the one in Lux, but that one stops at a slightly different stage (it stops _after_ the combination of the output and the next input, meaning N layers actually end up requiring N + 1 inputs unlike what's advertised). The gradient works now, but it's consistently about 6 µs off the pace from Lux when I benchmark it.

theabhirath:
The other question I had was regarding the return value. Currently this function returns only the final output. Is there a way to write a return expression that can return the entire range of the y-values (i.e. $[y_1 \ldots y_n]$)? I tried interpolating in different ways but I couldn't quite figure out how to do it 😅

src/layers/basic.jl
CarloLucibello: Also let's try to avoid generated function if possible, we have such an aestethically pleasant codebase...

The problem is array mutation. Maybe use comprehension instead?

theabhirath: Not quite sure which array is being mutated...do you mean `outputs`? Writing that as a comprehension seems confusing to me, given that every next element of `outputs` is using the previous element, which means that writing a one-line initialisation statement seems...tricky. I can't quite see how to do it 😅 The generated function does seem to help with AD though

CarloLucibello: > Not quite sure which array is being mutated...do you mean outputs? 


yes.

Something like this scheme should work:
```julia
julia> f(i, h) = i * h
g (generic function with 1 method)

julia> function test()
           h = 1
           [(h = f(i, h); h) for i=1:4]
       end
test (generic function with 1 method)

julia> test()
4-element Vector{Int64}:
  1
  2
  6
 24
```

CarloLucibello: And as @darsnak suggested
```julia
(m::PairwiseFusion)(x) = m(ntuple(i -> x, m.layers))

function (m::PairwiseFusion)(x::Tuple)
   ...
```

theabhirath: Okay, so I tried that, and it benchmarks more than 2x slower:

## Without generated function

```julia
function (m::PairwiseFusion)(x::Tuple)
  nlayers = length(m.layers)
  nx = length(x)
  if nx != nlayers
    throw(ArgumentError("PairwiseFusion with $nlayers layers takes $nlayers inputs, but got $nx inputs"))
  end
  out(::Val{1}) = m.layers[1](x[1])
  out(::Val{i}) where i = m.layers[i](m.connection(out(Val(i - 1)), x[i]))
  outputs = [out(Val(i)) for i in 1:nlayers]
  return outputs
end
```

### Benchmark

```julia
julia> model = PairwiseFusion(+,  Dense(1, 30),  Dense(30, 10));

julia> x = (rand(1, 10), rand(30, 10));

julia> @benchmark Zygote.gradient(p -> sum($model(p)[1]), $x)
 Range (min … max):  50.833 μs …   7.276 ms  ┊ GC (min … max): 0.00% … 98.10%
 Time  (median):     55.209 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   60.031 μs ± 174.896 μs  ┊ GC (mean ± σ):  7.05% ±  2.40%

            ▄█▇▇▇▃▁
  ▁▁▃▄▅▇▆▇▇█████████▆▅▅▄▄▃▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  50.8 μs         Histogram: frequency by time         70.8 μs <

 Memory estimate: 53.98 KiB, allocs estimate: 258.
```

## With generated function

```julia
function (m::PairwiseFusion)(x::T) where {T}
  lx = length(x)
  N = length(m.layers)
  if T <: Tuple && lx != N
    throw(ArgumentError("PairwiseFusion with $N sub-layers can take one input or $N inputs, but got $lx inputs"))
  end
  applypairwisefusion(m.layers, m.connection, x)
end

@generated function applypairwisefusion(layers::NamedTuple{names}, connection, x::T) where {names, T}
  N = length(names)
  y_symbols = [gensym() for _ in 1:(N + 1)]
  getinput(i) = T <: Tuple ? :(x[$i]) : :x
  calls = [:($(y_symbols[N + 1]) = $(getinput(1)))]
  append!(calls,
            [:($(y_symbols[i]) = layers[$i]($(y_symbols[N + 1]));
               $(y_symbols[N + 1]) = connection($(y_symbols[i]), $(getinput(i + 1)))) 
             for i in 1:N - 1])
  push!(calls, :($(y_symbols[N]) = layers[$N]($(y_symbols[N + 1]))))
  push!(calls, :(return $(y_symbols[N])))
  return Expr(:block, calls...)
end
```

### Benchmark:

```julia
julia> model = PairwiseFusion(+,  Dense(1, 30),  Dense(30, 10));

julia> x = (rand(1, 10), rand(30, 10));

julia> @benchmark Zygote.gradient(p -> sum($model(p)[1]), $x)
 Range (min … max):  22.042 μs …   4.498 ms  ┊ GC (min … max): 0.00% … 97.52%
 Time  (median):     24.125 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   27.066 μs ± 107.292 μs  ┊ GC (mean ± σ):  9.55% ±  2.40%

       ▃▅▅▆▆▆▇▄█▆▅▃▂
  ▁▂▃▅███████████████▇▆▅▄▃▃▃▃▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  22 μs           Histogram: frequency by time         31.8 μs <

 Memory estimate: 28.23 KiB, allocs estimate: 138.
```

CarloLucibello: or just
```julia
h = x[1]
out(i) = i ===1 ? m.layers[1](h) : m.layers[i](m.connection(h, x[i]))
outputs = [(h = out(i); h) for i in 1:nlayers]
```
if performance is the same

theabhirath:
> > So currently, this layer returns only the final output. Is there a way to write a return expression that can return the entire range of the y-values (i.e. [y1…yn])? I tried interpolating in different ways but I couldn't quite figure out how to do it 😅
> 
> @theabhirath see https://github.com/avik-pal/Lux.jl/blob/0ca5a265b7ef8c6d3da2b2dfacfefc56a39f5163/src/layers/basic.jl#L329

That did it, thank you! 😄

theabhirath:
Bump on this? I've added a very basic test (one that occurred to me immediately, I'll find some more and add them later). It'll help me complete some of the Metalhead models that have been stalling (Res2Net, for example)

src/layers/basic.jl
```diff
@@ -510,16 +510,109 @@ end
...
+```julia
+y = x
+for i in 1:N
+    y = connection(x, layers[i](y))
+end
+```
+
+## Returns
+
+`PairwiseFusion` returns a tuple of length `N` with the output of each fusion ((`y1`, `y2`, ..., `yN`) in the example above).
```
ToucheSir, line 567:
```suggestion
A tuple of length `N` with the output of each fusion ((`y1`, `y2`, ..., `yN`) in the example above).
```

src/layers/basic.jl
```diff
@@ -510,16 +510,109 @@ end
...
+  names = ntuple(i -> Symbol("layer_$i"), length(layers))
+  return PairwiseFusion(connection, NamedTuple{names}(layers))
+end
+
+function (m::PairwiseFusion)(x::T) where {T}
+  lx = length(x)
+  N = length(m.layers)
+  if T <: Tuple && lx != N
+    throw(ArgumentError("PairwiseFusion with $N sub-layers can take one input or $N inputs, but got $lx inputs"))
+  end
```
ToucheSir, line 584: Can you toss this in a function and mark it `@non_differentiable`?

src/layers/basic.jl
```diff
@@ -510,16 +510,109 @@ end
...
+
+@generated function applypairwisefusion(layers::NamedTuple{names}, connection, x::T) where {names, T}
+  N = length(names)
+  y_symbols = [gensym() for _ in 1:(N + 1)]
+  getinput(i) = T <: Tuple ? :(x[$i]) : :x
+  calls = [:($(y_symbols[N + 1]) = $(getinput(1)))]
+  append!(calls,
+            [:($(y_symbols[i]) = layers[$i]($(y_symbols[N + 1]));
+               $(y_symbols[N + 1]) = connection($(y_symbols[i]), $(getinput(i + 1)))) 
+             for i in 1:N - 1])
```
ToucheSir, line 596:
```suggestion
  for i in 1:N - 1
    push!(calls, quote
      $(y_symbols[i]) = layers[$i]($(y_symbols[N + 1]))
      $(y_symbols[N + 1]) = connection($(y_symbols[i]), $(getinput(i + 1)))
    end)
  end
```
For clarity.

src/layers/basic.jl
```diff
@@ -510,16 +510,109 @@ end
...
+end
+```
+
+2. Any other kind of input:
+
+```julia
+y = x
+for i in 1:N
+    y = connection(x, layers[i](y))
+end
```
ToucheSir, line 562: This will get a test, I presume?

PR #1973, opened 2022-05-27
Use `var` to speed up normalisation

mcabbott:
In [this comparison](https://discourse.julialang.org/t/ann-lux-jl-explicitly-parameterized-neural-networks-in-julia/81689/5), I think that [one line](https://github.com/FluxML/Flux.jl/blob/5f17f1c32fc1cdc2c65a6f5c583f4a2734238dca/src/layers/normalise.jl#L198) accounts for Flux's extra memory use, compared to [Lux's version](https://github.com/avik-pal/Lux.jl/blob/main/src/layers/normalize.jl#L147-L148) with `var`. (Perhaps that wasn't supported earlier?). This PR fixes it:
```julia
julia> x = rand(rng, Float32, 128, 1000);

julia> @btime gradient(model -> sum(model(x)), model);
  4.906 ms (1305 allocations: 29.19 MiB)  # before
  3.136 ms (1219 allocations: 23.32 MiB)  # after

julia> v, re = destructure(model);

julia> @btime gradient(v -> sum(re(v)(x)), v);
  5.044 ms (1562 allocations: 29.47 MiB)  # before
  3.306 ms (1476 allocations: 23.60 MiB)  # after
```
compared to Lux, same machine same size:
```julia
julia> @btime gradient(p -> sum(Lux.apply(model, x, p, st)[1]), ps);
  4.069 ms (2679 allocations: 23.41 MiB)

julia> ca = ComponentArray(ps);

julia> @btime gradient(p -> sum(Lux.apply(model, x, p, st)[1]), ca);
  4.327 ms (3061 allocations: 24.62 MiB)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.13.2"
+version = "0.13.3"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -195,7 +195,7 @@ function _norm_layer_forward(
     σ² = reshape(l.σ², stats_shape)
   else # trainmode or testmode without tracked stats
     μ = mean(x; dims=reduce_dims)
-    σ² = mean((x .- μ).^2; dims=reduce_dims)
+    σ² = var(x; mean=μ, dims=reduce_dims, corrected=false)
     if l.track_stats
       _track_stats!(l, x, μ, σ², reduce_dims) # update moving mean/std
     end


ToucheSir (approved):
I think this was even discussed at some point but fell through the cracks. LGTM.

CarloLucibello:
Does the gradient propagate to the keyword argument? I think that was the problem

mcabbott:
> Does the gradient propagate to the keyword argument? I think that was the problem

The gradient of the keyword is correctly zero, I think:
```julia
julia> gradient(x3, m3) do x, m
         var(x; mean=m)
       end
([-0.07914716919668174, 0.04399098883554592, 0.035156180361135936], nothing)

julia> ForwardDiff.gradient(x3) do x
         var(x; mean=m3)
       end
3-element Vector{Float64}:
 -0.07914716919668174
  0.04399098883554592
  0.035156180361135936

julia> ForwardDiff.derivative(m3) do m
         var(x3; mean=m)
       end
-1.1102230246251565e-16
```
Xref https://github.com/FluxML/Zygote.jl/pull/478

mcabbott:
You could save a few more copies here by making some `μ, σ2 = mean_var(x)` which computes both gradients. Or a rule for `_norm_layer_forward(x, μ, σ², ϵ)`, or better, fuse these two.

cossio:
The centered second moment `<(x - m)^2>` has a minimum when `m` coincides with the mean, `m = <x>`, so the gradient is correctly zero in this case.

mcabbott:
Here's a gist: https://gist.github.com/mcabbott/57befcf926b839e5e528ace38f018a66

tl;dr is that 2nd derivatives where you compute the mean one line above are fine. If you supply the mean from completely outside, then my head hurts, it's some overparameterised 2nd order tangent story.

CarloLucibello:
Ok, thanks, so I don't understand why but we are good

PR #1976, opened 2022-05-27
Piratical errors for two mistakes

### LINKED ISSUE #1974: Make `gpu(::DataLoader)` work or error loudly if it doesn't
This has come up a couple times, most recently in https://discourse.julialang.org/t/training-with-flux-jl-on-the-gpu-causes-argumenterror-cannot-take-the-cpu-address-of-a-cuarray/81751/2. RFC on what the desired approach is.

mcabbott:
Closes #1974, closes https://github.com/FluxML/Optimisers.jl/issues/76 

Both commit piracy, as MLUtils doesn't depend on Functors, and Optimisers doesn't depend on Zygote. Wouldn't be so hard to fix #1974 without piracy, but this is simple.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -7,11 +7,14 @@ using MacroTools: @forward
 
 @reexport using NNlib
 using MLUtils
-import Optimisers: trainable, destructure  # before v0.13, Flux owned these functions
+import Optimisers: Optimisers, trainable, destructure  # before v0.13, Flux owned these functions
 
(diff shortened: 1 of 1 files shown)


ToucheSir:
I suggested this and am naturally in favour of it, but if there are objections now is the time :)

CarloLucibello:
Seems the simplest thing to do

CarloLucibello (approved):

PR #1977, opened 2022-05-27
Show `using Flux` before BSON `@load`

JeffFessler:
To a novice, it is not obvious that `Flux` module must be in scope before one can `@load` a model saved with `BSON.@save`.  This small documentation change makes that requirement explicit.

### DIFF
--- a/docs/src/saving.md
+++ b/docs/src/saving.md
@@ -24,6 +24,8 @@ julia> @save "mymodel.bson" model
 Load it again:
 
 ```jldoctest saving
+julia> using Flux # must precede @load
+
 julia> using BSON: @load
 
 julia> @load "mymodel.bson" model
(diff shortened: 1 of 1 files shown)


docs/src/saving.md
```diff
@@ -24,6 +24,8 @@ julia> @save "mymodel.bson" model
 Load it again:
 
 ```jldoctest saving
+julia> using Flux # must precede @load
```
ToucheSir, line 27:
```suggestion
julia> using Flux # must be imported beforehand
```
Bikeshedding wording. May be a irrelevant and all users know that they can run code between `using Flux` and `@load ...`

JeffFessler: good point - the word "precede" might be read as "immediately precede"
but the word `import` might also mislead because `import Flux` will not work (I believe).
so i've tried a different rewording.  feel free to bikeshed further :)

ToucheSir (approved):
Looks good!

PR #1978, opened 2022-05-27
Update docstrings of `basic.jl` and `conv.jl`

Saransh-cpp:
Hi everyone! Super excited to work on Flux's documentation as a JSoC student! This PR begins the first phase of my work, mainly auditing the documentation, fixing the outdated documentation, adding missing docstrings, and adding doctests to every docstring.

Summarising this PR, I have added missing docstrings, updated the existing ones, and added some doctests in the following files -
- `basics.jl`
- `conv.jl`

There are some minor changes in `normalise.jl` but I will be taking that complete file with some other files in the next PR. Or I can update all the files in this PR altogether, whichever method is easier to review :)

cc: @DhairyaLGandhi and all other community members!

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -50,6 +50,36 @@ end
 
 (c::Chain)(x) = applychain(c.layers, x)
 
+"""
+    applychain(layers, x)
+
+Calculates the forward results of the complete chain provided as a `Tuple`, `AbstractVector`,
+or a `NamedTuple` of layers with `x` as model input. Users are encouraged to call a chain
+instead of using this function directly.
+
+# Examples
+
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
+julia> using Flux: applychain
+
+julia> c = Chain(Dense(10 => 5, σ), Dense(5 => 2), softmax);
+
+julia> applychain(c.layers, rand(10))  # will output a 2 element vector as our chain has 2 neurons in the final output layer
+2-element Vector{Float64}:
+ 0.5101459322326873
+ 0.4898540677673126
+
+julia> applychain([Dense(10 => 5, σ), Dense(5 => 2), softmax], rand(10))  # will output a 2 element vector as our chain has 2 neurons in the final output layer
+2-element Vector{Float64}:
+ 0.5101459322326873
+ 0.4898540677673126
+
+julia> c(rand(10))  # encouraged method
+2-element Vector{Float64}:
+ 0.4861734115447846
+ 0.5138265884552153
+```
+"""
 @generated function applychain(layers::Tuple{Vararg{<:Any,N}}, x) where {N}
   symbols = vcat(:x, [gensym() for _ in 1:N])
   calls = [:($(symbols[i+1]) = layers[$i]($(symbols[i]))) for i in 1:N]
@@ -65,6 +95,8 @@ function applychain(layers::AbstractVector, x)  # type-unstable path, helps comp
   x
 end
 
+# Overrides Base.getindex to obtain a `Chain` with selected layers
+# from a `Chain` passed to it.
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
   Chain(NamedTuple{keys(c)[i]}(Tuple(c.layers)[i]))
@@ -73,6 +105,8 @@ function Base.show(io::IO, c::Chain)
   _show_layers(io, c.layers)
   print(io, ")")
 end
+
+# supporting methods for the overriden `Base.show` function
 _show_layers(io, layers::Tuple) = join(io, layers, ", ")
 _show_layers(io, layers::NamedTuple) = join(io, ["$k = $v" for (k, v) in pairs(layers)], ", ")
 _show_layers(io, layers::AbstractVector) = (print(io, "["); join(io, layers, ", "); print(io, "]"))
@@ -85,10 +119,35 @@ _show_layers(io, layers::AbstractVector) = (print(io, "["); join(io, layers, ", 
 """
     activations(c::Chain, input)
 
-Calculate the forward results of each layers in Chain `c` with `input` as model input.
+Calculates the forward results of each layers in Chain `c` with `input` as model input.
+
+# Examples
+
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
+julia> using Flux: activations
+
+julia> c = Chain(Dense(10 => 5, σ), Dense(5 => 2), softmax);
+
+julia> activations(c, rand(10))  # will output a tuple of 3 lists (with length = 5, 2, and 2) as our chain has 3 layers
+([0.3274892431795043, 0.5360197770386552, 0.3447464835514667, 0.5273025865532305, 0.7513168089280781], [-0.3533774181890544, -0.010937055274926138], [0.4152168057978045, 0.5847831942021956])
+```
 """
 activations(c::Chain, input) = extraChain(Tuple(c.layers), input)
 
+"""
+    extraChain(fs::Tuple, x)
+
+Calculates the forward results of each layer provided in a `Tuple` with `x` as model input.
+
+# Examples
+
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
+julia> using Flux: extraChain
+
+julia> extraChain((Dense(10 => 5, σ), Dense(5 => 2), softmax), rand(10))  # will output a tuple of 3 lists (with length = 5, 2, and 2) as our chain has 3 layers
+([0.3274892431795043, 0.5360197770386552, 0.3447464835514667, 0.5273025865532305, 0.7513168089280781], [-0.3533774181890544, -0.010937055274926138], [0.4152168057978045, 0.5847831942021956])
+```
+"""
 function extraChain(fs::Tuple, x)
   res = first(fs)(x)
   return (res, extraChain(Base.tail(fs), res)...)
@@ -98,7 +157,7 @@ extraChain(::Tuple{}, x) = ()
 
 """
     Dense(in => out, σ=identity; bias=true, init=glorot_uniform)
-    Dense(W::AbstractMatrix, [bias, σ])
+    Dense(weights, bias=true, σ=identity)
 
 Create a traditional fully connected layer, whose forward pass is given by:
 
@@ -254,7 +313,7 @@ See Goodfellow, Warde-Farley, Mirza, Courville & Bengio "Maxout Networks"
 See also [`Parallel`](@ref) to reduce with other operators.
 
 # Examples
-```
+```jldoctest
 julia> m = Maxout(x -> abs2.(x), x -> x .* 3);
 
 julia> m([-2 -1 0 1 2])
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -50,36 +50,9 @@ end
 
 (c::Chain)(x) = applychain(c.layers, x)
 
-"""
-    applychain(layers, x)
-
-Calculates the forward results of the complete chain provided as a `Tuple`, `AbstractVector`,
-or a `NamedTuple` of layers with `x` as model input. Users are encouraged to call a chain
-instead of using this function directly.
-
-# Examples
-
-```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
-julia> using Flux: applychain
-
-julia> c = Chain(Dense(10 => 5, σ), Dense(5 => 2), softmax);
-
-julia> applychain(c.layers, rand(10))  # will output a 2 element vector as our chain has 2 neurons in the final output layer
-2-element Vector{Float64}:
- 0.5101459322326873
- 0.4898540677673126
-
-julia> applychain([Dense(10 => 5, σ), Dense(5 => 2), softmax], rand(10))  # will output a 2 element vector as our chain has 2 neurons in the final output layer
-2-element Vector{Float64}:
- 0.5101459322326873
- 0.4898540677673126
-
-julia> c(rand(10))  # encouraged method
-2-element Vector{Float64}:
- 0.4861734115447846
- 0.5138265884552153
-```
-"""
+# Calculates the forward results of the complete chain provided as a `Tuple`, `AbstractVector`,
+# or a `NamedTuple` of layers with `x` as model input. Users are encouraged to call a chain
+# instead of using this function directly.
 @generated function applychain(layers::Tuple{Vararg{<:Any,N}}, x) where {N}
   symbols = vcat(:x, [gensym() for _ in 1:N])
   calls = [:($(symbols[i+1]) = layers[$i]($(symbols[i]))) for i in 1:N]
@@ -134,20 +107,7 @@ julia> activations(c, 1)
 """
 activations(c::Chain, input) = extraChain(Tuple(c.layers), input)
 
-"""
-    extraChain(fs::Tuple, x)
-
-Calculates the forward results of each layer provided in a `Tuple` with `x` as model input.
-
-# Examples
-
-```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
-julia> using Flux: extraChain
-
-julia> extraChain((Dense(10 => 5, σ), Dense(5 => 2), softmax), rand(10))  # will output a tuple of 3 lists (with length = 5, 2, and 2) as our chain has 3 layers
-([0.3274892431795043, 0.5360197770386552, 0.3447464835514667, 0.5273025865532305, 0.7513168089280781], [-0.3533774181890544, -0.010937055274926138], [0.4152168057978045, 0.5847831942021956])
-```
-"""
+# Calculates the forward results of each layer provided in a `Tuple` with `x` as model input.
 function extraChain(fs::Tuple, x)
   res = first(fs)(x)
   return (res, extraChain(Base.tail(fs), res)...)
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -68,8 +68,6 @@ function applychain(layers::AbstractVector, x)  # type-unstable path, helps comp
   x
 end
 
-# Overrides Base.getindex to obtain a `Chain` with selected layers
-# from a `Chain` passed to it.
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
   Chain(NamedTuple{keys(c)[i]}(Tuple(c.layers)[i]))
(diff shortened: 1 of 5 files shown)


src/layers/basic.jl
```diff
@@ -50,6 +50,36 @@ end
 
 (c::Chain)(x) = applychain(c.layers, x)
 
+"""
+    applychain(layers, x)
```
ToucheSir, line 54: This method is effectively private. I like the docstring, but that should be noted (alongside a warning that it could change/disappear at any time).

src/layers/basic.jl
```diff
@@ -50,6 +50,36 @@ end
...
+julia> applychain([Dense(10 => 5, σ), Dense(5 => 2), softmax], rand(10))  # will output a 2 element vector as our chain has 2 neurons in the final output layer
+2-element Vector{Float64}:
+ 0.5101459322326873
+ 0.4898540677673126
+
+julia> c(rand(10))  # encouraged method
+2-element Vector{Float64}:
+ 0.4861734115447846
+ 0.5138265884552153
+```
```
ToucheSir, line 81: Instead of adding examples here, I would recommend just noting that this is used internally by `Chain` and directing users to its docs.

src/layers/basic.jl
```diff
@@ -85,10 +121,35 @@ _show_layers(io, layers::AbstractVector) = (print(io, "["); join(io, layers, ",
...
+
+# Examples
+
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
+julia> using Flux: activations
+
+julia> c = Chain(Dense(10 => 5, σ), Dense(5 => 2), softmax);
+
+julia> activations(c, rand(10))  # will output a tuple of 3 lists (with length = 5, 2, and 2) as our chain has 3 layers
+([0.3274892431795043, 0.5360197770386552, 0.3447464835514667, 0.5273025865532305, 0.7513168089280781], [-0.3533774181890544, -0.010937055274926138], [0.4152168057978045, 0.5847831942021956])
```
ToucheSir, line 134:
```suggestion
```jldoctest
julia> using Flux: activations

julia> c = Chain(x -> x + 1, x -> x * 2, x -> x ^ 3);

julia> activations(c, 1)
(2, 4, 64)
```
A simpler, deterministic example can be terser and reduce the need for filtering out doctests because of randomness.

src/layers/basic.jl
```diff
@@ -85,10 +121,35 @@ _show_layers(io, layers::AbstractVector) = (print(io, "["); join(io, layers, ",
...
+Calculates the forward results of each layer provided in a `Tuple` with `x` as model input.
+
+# Examples
+
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
+julia> using Flux: extraChain
+
+julia> extraChain((Dense(10 => 5, σ), Dense(5 => 2), softmax), rand(10))  # will output a tuple of 3 lists (with length = 5, 2, and 2) as our chain has 3 layers
+([0.3274892431795043, 0.5360197770386552, 0.3447464835514667, 0.5273025865532305, 0.7513168089280781], [-0.3533774181890544, -0.010937055274926138], [0.4152168057978045, 0.5847831942021956])
+```
```
ToucheSir, line 151: Same comments here as on `applychain`.

src/layers/basic.jl
```diff
@@ -85,10 +121,35 @@ _show_layers(io, layers::AbstractVector) = (print(io, "["); join(io, layers, ",
 """
     activations(c::Chain, input)
 
-Calculate the forward results of each layers in Chain `c` with `input` as model input.
+Calculates the forward results of each layers in Chain `c` with `input` as model input.
```
ToucheSir, line 124:
```suggestion
Like calling a `Chain`, but saves the result of each layer as an output.
```
Might benefit from more workshopping on top of this.

src/layers/conv.jl
```diff
@@ -19,10 +19,16 @@ the padding to be chosen such that the input and output sizes agree
 When `stride≠1`, the output size equals `ceil(input_size/stride)`.
 
 See also [`Conv`](@ref), [`MaxPool`](@ref).
+
+# Examples
+```jldoctest
+julia> Conv((2,2), 1 => 1, pad=SamePad())
+Conv((2, 2), 1 => 1, pad=(1, 0, 1, 0))  # 5 parameters
```
ToucheSir, line 26: :+1: I don't think same padding on even-sized kernels is super intuitive, so this is great. Would you consider adding another example for stride > 1?

src/layers/conv.jl
```diff
@@ -69,6 +76,10 @@ Keywords to control initialization of the layer:
 * `bias` - The initial bias vector is all zero by default. Trainable bias can be disabled entirely
   by setting this to `false`, or another vector can be provided such as `bias = randn(Float32, out)`.
 
+Convolutional layer can also be constructed with given weights and biases.
+While constructung a convolutional layer by providing weights and biases, the layer accepts the
+same keywords (and has the same defaults) as the `Conv((4,4), 3 => 7, relu)` method.
```
ToucheSir, line 81:
```suggestion
Convolutional layers can also be constructed with given weights and biases. This constructor accepts the
same keywords (and has the same defaults) as the `Conv((4,4), 3 => 7, relu)` method.
```
"given" feels a tad non-descriptive, but I can't think of a concise way to describe this. "manual construction"? "directly constructed"?

src/layers/conv.jl
```diff
@@ -154,6 +155,8 @@ channels from `in` to `out`.
 
 Accepts the keyword `init` (default: `glorot_uniform`) to control the sampling
 distribution.
+
+This is internally used by the [`Conv`](@ref) layer but can also be used independently.
```
ToucheSir, line 159: An example showing inputs -> output size would be great here.

src/layers/conv.jl
```diff
@@ -209,6 +215,10 @@ Note that `pad=SamePad()` here tries to ensure `size(output,d) == size(x,d) * st
 Parameters are controlled by additional keywords, with defaults
 `init=glorot_uniform` and `bias=true`.
 
+ConvTranspose layer can also be constructed with given weights and biases.
+While constructung a ConvTranspose layer by providing weights and biases, the layer accepts the
+same keywords (and has the same defaults) as the `ConvTranspose((4,4), 3 => 7, relu)` method.
```
ToucheSir, line 220: Same comments here as for the Conv constructor.

src/layers/conv.jl
```diff
@@ -350,6 +354,10 @@ specifying the size of the convolutional kernel;
 Parameters are controlled by additional keywords, with defaults
 `init=glorot_uniform` and `bias=true`.
 
+CrossCor layer can also be constructed with given weights and biases.
+While constructung a CrossCor layer by providing weights and biases, the layer accepts the
+same keywords (and has the same defaults) as the `CrossCor((4,4), 3 => 7, relu)` method.
```
ToucheSir, line 359: ref. Conv + ConvTranspose comments.

src/layers/conv.jl
```diff
@@ -186,6 +189,8 @@ function Base.show(io::IO, l::Conv)
   print(io, ")")
 end
 
+# helper function to  print additional options and keyword arguments for `Conv`
+# and friends when using `Base.show`
```
ToucheSir, line 193: I think most of this prose can be substituted by a small example showing which part of the conv display code this handles. That or a test. The function name and surrounding context can handle the rest.

src/layers/conv.jl
```diff
@@ -642,6 +644,7 @@ function Base.show(io::IO, m::MaxPool)
   print(io, ")")
 end
 
+# helper functions for printing `Conv` (and family) layers
```
ToucheSir, line 647: Don't really think this comment needs to exist.
```suggestion
```

src/layers/normalise.jl
```diff
@@ -256,13 +256,13 @@ that will be used to renormalize the input in test phase.
...
-  Dense(64 => 10),
-  BatchNorm(10),
-  softmax)
+```jldoctest
+julia> m = Chain(
+        Dense(28^2 => 64),
+        BatchNorm(64, relu),
+        Dense(64 => 10),
+        BatchNorm(10),
+        softmax);
```
ToucheSir, line 265: Why the formatting change?

ToucheSir (commented):
Thanks, my comments below:

src/layers/basic.jl
```diff
@@ -65,14 +95,20 @@ function applychain(layers::AbstractVector, x)  # type-unstable path, helps comp
...
   Chain(NamedTuple{Base.keys(c)[i]}(Tuple(c.layers)[i]))
+
+# Overrides Base.show to print a `Chain` passed to it.
 function Base.show(io::IO, c::Chain)
   print(io, "Chain(")
   _show_layers(io, c.layers)
   print(io, ")")
 end
+
+# supporting methods for the overriden `Base.show` function
```
ToucheSir, line 111: All of these comments are unnecessary.

src/layers/basic.jl
mcabbott: I think no docstring is better. The only reason this wasn't completely replaced with `foldl` or `Base.afoldl` was some tests on speed & TTFG. 

If you want to rename it `_applychain` to stress that it's internal, that would be fine.

src/layers/conv.jl
mcabbott: Can you expand this a little, to show the output sizes with this layer, compared to the same without the SamePad, on the same `x`?

src/layers/conv.jl
mcabbott: I would rather mark this as internal. It's unlikely to change but nothing outside of this file should be using it.

src/layers/conv.jl
mcabbott: I would just remove this comment. The function's name seems pretty clear no? It's obviously internal, and is used only in this file.

src/layers/basic.jl
Saransh-cpp: Thanks! Looks much better!

src/layers/conv.jl
Saransh-cpp: This method is available in the user-facing documentation manual of `Flux`. Should I remove it from there to mark it as internal?

src/layers/conv.jl
```diff
@@ -92,27 +103,7 @@ julia> Conv((1,1), 3 => 7; pad = (20,10,0,0))(xs) |> size
...
-"""
-    Conv(weight::AbstractArray, [bias, activation; stride, pad, dilation])
-
-Constructs a convolutional layer with the given weight and bias.
-Accepts the same keywords (and has the same defaults) as the `Conv((4,4), 3 => 7, relu)`
-method.
-
-# Examples
-```jldoctest
 julia> weight = rand(3, 4, 5);
```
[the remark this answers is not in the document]
Saransh-cpp: This example and the associated docstring don't render into Flux's user manual. A workaround could be listing both the constructors in the manual by providing their signatures in the `.md` file. A related issue - https://github.com/JuliaDocs/Documenter.jl/issues/396

src/layers/conv.jl
mcabbott: Oh that's awful, I didn't realise. My vote would be to remove it from there, but not to rename it (or at least not without a depwarn) in case someone is in fact relying on it.

src/layers/conv.jl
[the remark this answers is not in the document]
mcabbott: Ah ok. 

Does adding the signature help? It's possible that something like this would work:
````
```@docs
Conv
Conv(weight::AbstractArray)
```
````
For instance `sqrt` at the REPL shows a second method for matrices, and it is shown in a different page of the docs:
https://raw.githubusercontent.com/JuliaLang/julia/master/stdlib/LinearAlgebra/docs/src/index.md

File: src/layers/basic.jl [hunk 16]
```diff
@@ -48,17 +48,20 @@ end
 
 @functor Chain
 
-(c::Chain)(x) = applychain(c.layers, x)
+(c::Chain)(x) = _applychain(c.layers, x)
 
-@generated function applychain(layers::Tuple{Vararg{<:Any,N}}, x) where {N}
+# Calculates the forward results of the complete chain provided as a `Tuple`, `AbstractVector`,
+# or a `NamedTuple` of layers with `x` as model input. Users are encouraged to call a chain
+# instead of using this function directly.
```
mcabbott, line 55:
```suggestion
```
I think the code says this pretty clearly. 

If you want a comment, it should be a reference to the issues about TTFG and `foldl` etc, which may help future readers.

src/layers/conv.jl
```diff
@@ -19,10 +19,33 @@ the padding to be chosen such that the input and output sizes agree
...
+
+# Examples
+```jldoctest
+julia> xs = rand(Float32, 100, 100, 3, 50);  # a batch of image
+
+julia> layer = Conv((2,2), 3 => 7, pad=SamePad())
+Conv((2, 2), 3 => 7, pad=(1, 0, 1, 0))  # 91 parameters
+
+julia> layer(xs) |> size  # notice how the dimensions would stay same as "same" padding is applied
+(100, 100, 7, 50)
```
mcabbott, line 31:
```suggestion
julia> xs = rand(Float32, 100, 100, 3, 50);  # a batch of images

julia> layer = Conv((2,2), 3 => 7, pad=SamePad())
Conv((2, 2), 3 => 7, pad=(1, 0, 1, 0))  # 91 parameters

julia> layer(xs) |> size  # notice how the dimensions stay the same with this padding
(100, 100, 7, 50)
```

src/layers/conv.jl
```diff
@@ -19,10 +19,33 @@ the padding to be chosen such that the input and output sizes agree
...
+Conv((2, 2), 3 => 7)  # 91 parameters
+
+julia> layer2(xs) |> size  # the output dimension changes as the padding was not "same"
+(99, 99, 7, 50)
+
+julia> layer3 = Conv((2,2), 3 => 7, stride=2, pad=SamePad())
+Conv((2, 2), 3 => 7, pad=(1, 0, 1, 0), stride=2)  # 91 parameters
+
+julia> layer3(xs) |> size  # output size = `ceil(input_size/stride)` = 50
+(50, 50, 7, 50)
```
mcabbott, line 43: Could this example be engineered not to have the same `pad=(1, 0, 1, 0)` again?

src/layers/conv.jl
```diff
@@ -342,6 +381,7 @@ end
 
 """
     CrossCor(filter, in => out, σ=identity; stride=1, pad=0, dilation=1, [bias, init])
+    CrossCor(weight::AbstractArray, [bias, activation; stride, pad, dilation])
 
 Standard cross convolutional layer. `filter` is a tuple of integers
```
Saransh-cpp, line 386:
```suggestion
Standard cross correlation layer. `filter` is a tuple of integers
```
Is this supposed to say correlation, or is convolution correct?

ToucheSir: Good catch, it should say correlation.

src/layers/conv.jl
```diff
@@ -379,8 +422,24 @@ end
 """
     CrossCor(weight::AbstractArray, [bias, activation; stride, pad, dilation])
 
-Constructs a layer with the given weight and bias arrays.
-Accepts the same keywords as the `CrossCor((4,4), 3 => 7, relu)` method.
+Constructs a CrossCor layer with the given weight and bias.
+Accepts the same keywords (and has the same defaults) as the `CrossCor((5,5), 3 => 6, relu)` method.
```
ToucheSir, line 426:
```suggestion
Accepts the same keywords and has the same defaults as `CrossCor((5,5), 3 => 6, relu)`.
```
I would trim these bits off the other Conv* constructor doc strings as well.

Saransh-cpp:
Moved the normalisation layers to #1995

src/layers/conv.jl
```diff
@@ -343,13 +381,17 @@ end
...
+Standard cross correlation layer. `filter` is a tuple of integers
 specifying the size of the convolutional kernel;
 `in` and `out` specify the number of input and output channels.
 
 Parameters are controlled by additional keywords, with defaults
 `init=glorot_uniform` and `bias=true`.
 
+CrossCor layer can also be manually constructed by passing in weights and
+biases. This constructor accepts the layer accepts the same keywords (and has
+the same defaults) as the `CrossCor((4,4), 3 => 7, relu)` method.
```
ToucheSir, line 393:
```suggestion
the same defaults) as `CrossCor(k::NTuple{N,Integer}, ch::Pair{<:Integer,<:Integer}, σ; ...)`.
```
Better to use the signature so that users know which constructor to look for.

src/layers/conv.jl
```diff
@@ -244,8 +267,24 @@ _channels_out(l::ConvTranspose) = size(l.weight)[end-1]*l.groups
 """
     ConvTranspose(weight::AbstractArray, [bias, activation; stride, pad, dilation, groups])
 
-Constructs a layer with the given weight and bias arrays.
-Accepts the same keywords as the `ConvTranspose((4,4), 3 => 7, relu)` method.
+Constructs a ConvTranspose layer with the given weight and bias.
+Accepts the same keywords and has the same defaults as `ConvTranspose((4,4), 3 => 7, relu)`.
```
ToucheSir, line 271:
```suggestion
Accepts the same keywords and has the same defaults as `ConvTranspose(k::NTuple{N,Integer}, ch::Pair{<:Integer,<:Integer}, σ; ...)`.
```
I'm not sure if `@ref` works here, but if it does I would recommend using it to refer to the other constructor.

src/layers/conv.jl
```diff
@@ -19,10 +19,33 @@ the padding to be chosen such that the input and output sizes agree
 When `stride≠1`, the output size equals `ceil(input_size/stride)`.
 
 See also [`Conv`](@ref), [`MaxPool`](@ref).
+
+# Examples
+```jldoctest
+julia> xs = rand(Float32, 100, 100, 3, 50);  # a batch of images
+
+julia> layer = Conv((2,2), 3 => 7, pad=SamePad())
```
ToucheSir, line 27: These layer variables are sometimes called `layer`, sometimes `c__`, sometimes `lay`, etc. There should be a consistent naming scheme for them all (at the very least for all the conv-like layers).

src/layers/conv.jl
```diff
@@ -379,8 +421,24 @@ end
...
+julia> bias = zeros(5);
+
+julia> lay = CrossCor(weight, bias, relu)
+CrossCor((3,), 4 => 5, relu)  # 65 parameters
+
+julia> lay(randn(100, 4, 64)) |> size
+(98, 5, 64)
+
+julia> Flux.params(lay) |> length
+2
```
ToucheSir, line 440: I don't really think this is necessary?
```suggestion
```

ToucheSir (approved):
Ok, this looks to be in a good place. I'm going to give it another day or so and then merge :)

Saransh-cpp:
Is this ready to be merged?

ToucheSir:
Yes it is, sorry it fell through the cracks!

PR #1983, opened 2022-06-02
`PairwiseFusion` layer, take 2

theabhirath:
I messed up the commits on #1971 a little bit, so this one should close #1971. If merged, this will also close #1989.

---

This PR implements a `PairwiseFusion` layer similar to the one in [Lux](https://github.com/avik-pal/Lux.jl/blob/0ca5a265b7ef8c6d3da2b2dfacfefc56a39f5163/src/layers/basic.jl#L335-L388). This might be especially useful for implementing models like [Res2Net](https://github.com/FluxML/Metalhead.jl/pull/123), since the forward pass for the bottleneck layer is quite similar to this.

## PR Checklist
- [x] Add pretty-printing similar to `Parallel`
- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -67,7 +67,7 @@ end
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
-  Chain(NamedTuple{Base.keys(c)[i]}(Tuple(c.layers)[i]))
+  Chain(NamedTuple{keys(c)[i]}(Tuple(c.layers)[i]))
 function Base.show(io::IO, c::Chain)
   print(io, "Chain(")
   _show_layers(io, c.layers)
@@ -566,35 +566,37 @@ end
 
 `PairwiseFusion` returns a tuple of length `N` with the output of each fusion ((`y1`, `y2`, ..., `yN`) in the example above).
 """
-struct PairwiseFusion{F, T<:Union{Tuple, NamedTuple}}
+struct PairwiseFusion{F, T <: NamedTuple}
   connection::F
   layers::T
 end
 
-PairwiseFusion(connection, layers...) = PairwiseFusion(connection, layers)
-function PairwiseFusion(connection; kw...)
-  layers = NamedTuple(kw)
-  if :layers in keys(layers) || :connection in keys(layers)
-    throw(ArgumentError("a Parallel layer cannot have a named sub-layer called `connection` or `layers`"))
-  end
-  isempty(layers) && return Parallel(connection, ())
-  return PairwiseFusion(connection, layers)
+function PairwiseFusion(connection, layers...)
+  names = ntuple(i -> Symbol("layer_$i"), length(layers))
+  return PairwiseFusion(connection, NamedTuple{names}(layers))
 end
 
 function (m::PairwiseFusion)(x::T) where {T}
-  nlayers = length(m.layers)
-  getinput(i) = T <: Tuple ? x[i] : x
-  if T <: Tuple
-    nx = length(x)
-    if nx != nlayers
-      throw(ArgumentError("PairwiseFusion with $nlayers layers takes $nlayers inputs, but got $nx inputs"))
-    end
+  lx = length(x)
+  N = length(m.layers)
+  if T <: Tuple && lx != N
+    throw(ArgumentError("PairwiseFusion with $N sub-layers can take one input or $N inputs, but got $lx inputs"))
   end
-  outputs = [m.layers[1](getinput(1))]
-  for i in 2:nlayers
-    push!(outputs, m.layers[i](m.connection(getinput(i), outputs[i - 1])))
-  end
-  return outputs
+  applypairwisefusion(m.layers, m.connection, x)
+end
+
+@generated function applypairwisefusion(layers::NamedTuple{names}, connection, x::T) where {names, T}
+  N = length(names)
+  y_symbols = [gensym() for _ in 1:(N + 1)]
+  getinput(i) = T <: Tuple ? :(x[$i]) : :x
+  calls = [:($(y_symbols[N + 1]) = $(getinput(1)))]
+  append!(calls,
+            [:($(y_symbols[i]) = layers[$i]($(y_symbols[N + 1]));
+               $(y_symbols[N + 1]) = connection($(y_symbols[i]), $(getinput(i + 1)))) 
+             for i in 1:N - 1])
+  push!(calls, :($(y_symbols[N]) = layers[$N]($(y_symbols[N + 1]))))
+  push!(calls, :(return $(y_symbols[N])))
+  return Expr(:block, calls...)
 end
 
 @functor PairwiseFusion
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -595,18 +595,23 @@ end
                $(y_symbols[N + 1]) = connection($(y_symbols[i]), $(getinput(i + 1)))) 
              for i in 1:N - 1])
   push!(calls, :($(y_symbols[N]) = layers[$N]($(y_symbols[N + 1]))))
-  push!(calls, :(return $(y_symbols[N])))
+  push!(calls, :(return tuple($(Tuple(y_symbols[1:N])...))))
   return Expr(:block, calls...)
 end
 
 @functor PairwiseFusion
 
 Base.getindex(m::PairwiseFusion, i) = m.layers[i]
-Base.getindex(m::PairwiseFusion, i::AbstractVector) = PairwiseFusion(m.connection, m.layers[i])
 Base.getindex(m::PairwiseFusion{<:Any, <:NamedTuple}, i::AbstractVector) =
   PairwiseFusion(m.connection, NamedTuple{keys(m)[i]}(Tuple(m.layers)[i]))
 
 Base.keys(m::PairwiseFusion) = keys(getfield(m, :layers))
+
+function Base.show(io::IO, m::PairwiseFusion)
+  print(io, "PairwiseFusion(", m.connection, ", ")
+  _show_layers(io, m.layers)
+  print(io, ")")
+end
 
 """
     Embedding(in => out; init=randn)
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -538,7 +538,7 @@ end
 
 This layer behaves differently based on input type:
 
-1. If input `x` is a tuple of length `N`, matching the number of `layers`, 
+1. If input `x` is a tuple of length N (or the input is `xs` with N `x`'s), matching the number of `layers`, 
   then each layer receives a new input `x[i]` combined with the previous output `y[i-1]` using `connection`.
   Thus `(y1, y2, y3) = PairwiseFusion(connection, layer1, layer2, layer3)((x1, x2, x3))`
   may be drawn as:
@@ -567,7 +567,7 @@ end
 
 ## Returns
 
-A tuple of length `N` with the output of each fusion ((`y1`, `y2`, ..., `yN`) in the example above).
+A tuple of length N with the output of each fusion ((`y1`, `y2`, ..., `yN`) in the example above).
 """
 struct PairwiseFusion{F, T<:Union{Tuple, NamedTuple}}
   connection::F
@@ -597,6 +597,7 @@ function (m::PairwiseFusion)(x::T) where {T}
   _pairwise_check(x, m.layers, T)
   applypairwisefusion(m.layers, m.connection, x)
 end
+(m::PairwiseFusion)(xs...) = m(xs)
 
 @generated function applypairwisefusion(layers::Tuple{Vararg{<:Any,N}}, connection, x::T) where {N, T}
   y_symbols = [gensym() for _ in 1:(N + 1)]
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -582,11 +582,15 @@ function PairwiseFusion(connection; kw...)
 end
 
 function (m::PairwiseFusion)(x::T) where {T}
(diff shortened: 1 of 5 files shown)


src/layers/basic.jl
```diff
@@ -510,16 +510,116 @@ end
...
+x1 --> layer1 --> y1
+                  |
+                  |--> connection --> layer2 --> y2
+                  |                              |
+                  x2                             |--> connection --> layer3 --> y3
+                                                 |                              |
+                                                 x3                             |--> connection --> y4
+                                                                                |
+                                                                                x4
+```
```
mcabbott, line 536: Can this diagram be clearer? It looks like `connection` gets one input. I also think it should have next to it the corresponding code, something like:
````
With 3 layers, it takes 3 inputs and returns 3 outputs.
`(y1, y2, y3) = PairwiseFusion(con, lay1, lay2, lay3)((x1, x2, x3))`
may be drawn as:
```
x1 → lay1 → y1 ↘
                 con → lay2 → y2 ↘
            x2 ↗                   con → lay3 → y3
                              x3 ↗ 
```
or written
```julia
y1 = lay1(x1)
y2 = lay2(con(x2, y1))
y3 = lay3(con(x3, y2))
```
````
The loops below also seem a bit misleading, as they keep only one `y`, not all intermediates.

File: src/layers/basic.jl [hunk 2]
```diff
@@ -510,16 +510,114 @@ end
...
   _show_layers(io, m.layers)
   print(io, ")")
 end
 
+"""
+    PairwiseFusion(connection, layers...)
+
+## Arguments
+
+- `connection`: Takes 2 inputs and combines them
```
CarloLucibello, line 528:
```suggestion
- `connection`: A function taking 2 inputs and combining them into a single output 
```

src/layers/basic.jl
```diff
@@ -510,16 +510,114 @@ end
...
+y₁ = x
+for i in 1:N
+    yᵢ₊₁ = connection(x, layers[i](yᵢ))
+end
+```
+
+## Returns
+
+A tuple of length `N` with the output of each fusion ((`y1`, `y2`, ..., `yN`) in the example above).
+"""
```
CarloLucibello, line 566: Would be good to add a concrete example

darsnack:
Despite the performance difference, do we feel good about a generated function? Why just this layer and not all the control flow layers like `Chain` or `Parallel`? One of the goals of the Julia + Flux is that you should be able to just write a loop, have it be fast, and have it be differentiable. Needing to unroll any loops seems clearly sub-optimal to me.

A couple questions related to performance:
- Is the performance difference coming from the backwards pass? Hopefully the loop and generated function are not too far off on the forwards pass.
- Is this consistent across Zygote versions/Julia versions (including 1.5)? Somewhere in the Zygote 0.6 and/or Julia 1.6 transition, there was a noticeable degradation. I know I have said this in several places without providing hard data, but it was measurable enough that we got issues on the Zygote about it, we saw it in Metalhead test times, and my own anecdotal experience. Hopefully I have time after I defend next week to benchmark these things (I feel guilty about complaining without doing the work 😄).
- Does using a `Vector` help the loop-version performance like it does for `Chain`?

theabhirath:
Well, the performance on the forward pass is almost exactly the same. The backwards pass, though...well.

Using a generated function:
```julia
julia> model = PairwiseFusion(+,  Dense(1, 30),  Dense(30, 10));

julia> x = (rand(1, 10), rand(30, 10));

julia> @benchmark $model($x)
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.158 μs … 265.258 μs  ┊ GC (min … max):  0.00% … 98.56%
 Time  (median):     1.746 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   2.387 μs ±  13.244 μs  ┊ GC (mean ± σ):  28.64% ±  5.12%

   ▅▁                 ▁▃█▄▃▁
  ▄██▃▂▂▂▁▁▁▁▁▂▂▂▂▂▂▄▅██████▇▆▅▄▃▄▄▃▃▃▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  1.16 μs         Histogram: frequency by time        2.69 μs <

 Memory estimate: 12.05 KiB, allocs estimate: 7.

julia> @benchmark Zygote.gradient(p -> sum($model(p)[1]), $x)
BenchmarkTools.Trial: 10000 samples with 8 evaluations.
 Range (min … max):  3.229 μs … 330.734 μs  ┊ GC (min … max):  0.00% … 97.61%
 Time  (median):     4.875 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   6.764 μs ±  24.378 μs  ┊ GC (mean ± σ):  27.72% ±  7.54%

   ▃▁            ▆█▄▅▅
  ▃██▃▂▃▂▂▃▄▅▆▆▆███████▇▆▄▃▃▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▃
  3.23 μs         Histogram: frequency by time        9.07 μs <

 Memory estimate: 34.39 KiB, allocs estimate: 41.
```

Using standard for-loops:
```julia
julia> @benchmark $model($x)
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.575 μs … 292.387 μs  ┊ GC (min … max):  0.00% … 98.56%
 Time  (median):     2.225 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   2.927 μs ±  14.270 μs  ┊ GC (mean ± σ):  24.72% ±  5.02%

   ▁               ▄█▇▇▅▁
  ▃██▅▅▄▃▃▃▂▂▃▃▃▃▄▇██████▇▆▅▆▅▅▄▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂ ▃
  1.58 μs         Histogram: frequency by time        3.53 μs <

 Memory estimate: 12.30 KiB, allocs estimate: 13.

julia> @benchmark Zygote.gradient(p -> sum($model(p)[1]), $x)
 Range (min … max):  46.250 μs …   4.285 ms  ┊ GC (min … max): 0.00% … 96.76%
 Time  (median):     48.541 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   53.598 μs ± 135.555 μs  ┊ GC (mean ± σ):  8.18% ±  3.19%

         ▁▅▄█▅▆▂▂
  ▁▁▂▃▄▆▆████████▇▅▅▃▃▃▃▄▃▃▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  46.2 μs         Histogram: frequency by time         57.7 μs <

 Memory estimate: 51.61 KiB, allocs estimate: 399.
```

Julia version:
```julia
julia> versioninfo()
Julia Version 1.9.0-DEV.695
Commit d84f8901ae (2022-06-01 05:24 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin21.5.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, apple-m1)
  Threads: 4 on 4 virtual cores
Environment:
  JULIA_NUM_THREADS = 4
```

This is Flux 0.13.3, Zygote 0.6.40. I will have to check for other Julia versions - but anything before 1.7 might not be the best benchmarked on my computer because it won't be native, so Rosetta might interfere with performance.

theabhirath:
I've also opened #1989 so that if there is an obvious performance trick that I'm missing there, someone can help me fix it. It'll also help in people benchmarking independently on their respective machines as well

ToucheSir:
> Despite the performance difference, do we feel good about a generated function? Why just this layer and not all the control flow layers like `Chain` or `Parallel`?

To be fair, we did switch `Chain` over to a generated function for exactly this reason. `Parallel` doesn't need it because mapreduce is flexible enough to handle all of its cases, but after playing around with a number of variations (including `foldl`), a generated function came out on top for `Chain`.

For that reason, my vote is to get this in first and then work on getting rid of `@generated`. One idea I've been mulling over is a throwback to the old `applychain` model of recursively folding over a tuple, but with liberal use of `@nospecialize` to avoid the crushing compile time penalty that comes with. Unclear whether that will have better TTFG than a generated function, so I'd not count them out until we know for sure.

darsnack:
Okay this is embarrassing but I didn't realize we already switched `Chain` over. I somehow thought we ended on slightly varied recursive implementation + Vector. I should have checked the current code instead of my memory of PRs, my bad.

In which case I'm inclined to agree with you, if we're already using generated functions in the short term, then no need to not merge this.

src/layers/basic.jl
```diff
@@ -510,16 +510,114 @@ end
...
+struct PairwiseFusion{F, T <: NamedTuple}
+  connection::F
+  layers::T
+end
+
+function PairwiseFusion(connection, layers...)
+  names = ntuple(i -> Symbol("layer_$i"), length(layers))
+  return PairwiseFusion(connection, NamedTuple{names}(layers))
+end
```
mcabbott, line 575: Do we want standard names? Parallel and Chain don't do this, they allow just a Tuple too.

src/layers/basic.jl
```diff
@@ -510,16 +510,114 @@ end
...
+  push!(calls, :($(y_symbols[N]) = layers[$N]($(y_symbols[N + 1]))))
+  push!(calls, :(return tuple($(Tuple(y_symbols[1:N])...))))
+  return Expr(:block, calls...)
+end
+
+@functor PairwiseFusion
+
+Base.getindex(m::PairwiseFusion, i) = m.layers[i]
+Base.getindex(m::PairwiseFusion{<:Any, <:NamedTuple}, i::AbstractVector) =
+  PairwiseFusion(m.connection, NamedTuple{keys(m)[i]}(Tuple(m.layers)[i]))
```
mcabbott, line 611: If the type always has names, then no need for `<:NamedTuple` dispatch. 

If it doesn't, then one more method is needed for `i=1:2` to return another PairwiseFusion.

src/layers/basic.jl
```diff
@@ -510,16 +510,114 @@ end
...
+end
+
+function _pairwise_check(lx, N, T)
+  if T <: Tuple && lx != N
+    throw(ArgumentError("PairwiseFusion with $N sub-layers can take one input or $N inputs, but got $lx inputs"))
+  end
+end
+ChainRulesCore.@non_differentiable _pairwise_check(lx, N, T)
+
+function (m::PairwiseFusion)(x::T) where {T}
```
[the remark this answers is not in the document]
darsnack: The first suggestion seems okay. The second would make the layer less usable if the sub-layers are custom layers that accept something other than the type restriction that we provide. Presumably, in most cases, the sub-layers should be appropriately restricted to throw an error?

src/layers/basic.jl
theabhirath: Pushed changes that should take care of this

src/layers/basic.jl
theabhirath: This should now happen with the latest changes

src/layers/basic.jl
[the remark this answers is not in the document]
theabhirath: I've allowed the first. Not sure about the second...

theabhirath:
I rebased, not merged, but git seems to have weird duplicated docs merges that are partially squashed 🤦🏽‍♂️ Not sure what happened

theabhirath:
Bump?

ToucheSir (approved):
No more comments from me, will merge tonight barring any objections :)

src/layers/basic.jl
mcabbott: How about:
```
julia> PairwiseFusion(vcat, x->x.+1, x->x.+2, x->x.^3)(2)
(3, [5, 4], [125, 64, 8])

julia> PairwiseFusion(vcat, x->x.+1, x->x.+2, x->x.^3)((2, 10, 20))
(3, [5, 12], [125, 1728, 8000])
```

src/layers/basic.jl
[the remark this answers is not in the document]
mcabbott: I don't see the method allowing this:
```
julia> PairwiseFusion(vcat, x->x.+1, x->x.+2, x->x.^3)(2, 10, 20)
ERROR: MethodError: no method matching (::PairwiseFusion{typeof(vcat), Tuple{var"#130#133", var"#131#134", var"#132#135"}})(::Int64, ::Int64, ::Int64)
Closest candidates are:
  (::PairwiseFusion)(::T) where T at REPL[11]:1
```
Agree that the second is less obvious. In particular it rules out many easy readme examples.

test/layers/basic.jl
```diff
@@ -350,3 +350,19 @@ end
   @test Flux.destructure(m1)[2](z1)[1].weight ≈ Flux.destructure(m1v)[2](z1)[1].weight
   # Note that Flux.destructure(m1v)[2](z) has a Chain{Tuple}, as does m1v[1:2]
 end
+
+@testset "PairwiseFusion" begin
+  x = (rand(1, 10), rand(30, 10))
+  layer = PairwiseFusion(+,  Dense(1, 30),  Dense(30, 10))
```
mcabbott, line 356: Testing with `+` means you don't check whether the `x` is the 1st or the 2nd argument. Maybe some tests should have `vcat`?

src/layers/basic.jl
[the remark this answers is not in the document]
theabhirath: Whoops, yeah, I think I'd missed pushing that - should be fixed now

test/layers/basic.jl
theabhirath: I've added a couple of tests with `vcat` - this was also a nice opportunity to add a test with just the single input `x`

PR #1995, opened 2022-06-11
Update docstrings in `upsample.jl`, `recurrent.jl`, and `normalise.jl` 

Saransh-cpp:
Added missing docstrings (in the manual), updated the existing ones, and added doctests in the following files -

- `normalise.jl`
- `recurrent.jl`
- `upsample.jl`

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -67,7 +67,7 @@ Custom RNGs are only supported on the CPU.
 Does nothing to the input once [`Flux.testmode!`](@ref) is `true`.
 
 # Examples
-```jldoctest; filter = r"[+-]?(?:(?:[0-9])(?:\\.\\d+)?)|(?:1)(?:\\.0+)?"
+```julia
 julia> m = Chain(Dense(2 => 2), Dropout(1))
 Chain(
   Dense(2 => 2),                        # 6 parameters
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -131,15 +131,15 @@ Does nothing to the input once [`testmode!`](@ref) is true.
 ```jldoctest
 julia> using Statistics
 
-julia> x = randn(20,1);
+julia> x = randn(1000,1);
 
-julia> m = Chain(Dense(20 => 10, selu), AlphaDropout(0.5));
+julia> m = Chain(Dense(1000 => 1000, selu), AlphaDropout(0.2));
 
 julia> Flux.trainmode!(m);
 
 julia> y = m(x);
 
-julia> isapprox(std(x), std(y), atol=0.6)
+julia> isapprox(std(x), std(y), atol=0.2)
 true
 ```
 """
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -87,7 +87,7 @@ Chain(
   Dropout(0.5),
 )
 
-julia> Flux.trainmode!(m);  # activating the layer without actually training it
+julia> Flux.trainmode!(m);
 
 julia> m([1, 2])  # drops neurons with a probability of 0.5
 2-element Vector{Float32}:
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -77,13 +77,13 @@ julia> y = m([1]);
 julia> y == [0]
 true
 
-julia> m = Chain(Dense(1 => 1), Dropout(0.5));
+julia> m = Chain(Dense(1000 => 1000), Dropout(0.5));
 
 julia> Flux.trainmode!(m);
 
-julia> y = m([1]);
+julia> y = m(ones(1000));
 
-julia> m[2].p - 0.5 <= count(i->(i == 0), y) <= m[2].p + 0.5  # number of zeros can be 0 or 1
+julia> isapprox(count(==(0), y) / length(y), 0.5, atol=0.1)
 true
 ```
 """
@@ -139,7 +139,7 @@ julia> Flux.trainmode!(m);
 
 julia> y = m(x);
 
-julia> isapprox(std(x), std(y), rtol=0.6)
+julia> isapprox(std(x), std(y), atol=0.6)
 true
 ```
 """
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -67,7 +67,7 @@ Custom RNGs are only supported on the CPU.
 Does nothing to the input once [`Flux.testmode!`](@ref) is `true`.
 
 # Examples
-```jldoctest
+```jldoctest; filter = r"[+-]?(?:(?:[0-9])(?:\\.\\d+)?)|(?:1)(?:\\.0+)?"
 julia> m = Chain(Dense(2 => 2), Dropout(1))
 Chain(
   Dense(2 => 2),                        # 6 parameters
@@ -136,14 +136,10 @@ remain the same as before.
 Does nothing to the input once [`testmode!`](@ref) is true.
 
 # Examples
-```jldoctest
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
 julia> x = randn(20,1);
 
-julia> m = Chain(Dense(20 => 10, selu), AlphaDropout(0.5))
-Chain(
-  Dense(20 => 10, selu),                # 210 parameters
-  AlphaDropout{Float64, Random.TaskLocalRNG}(0.5, nothing, Random.TaskLocalRNG()),
-)
+julia> m = Chain(Dense(20 => 10, selu), AlphaDropout(0.5));
 
 julia> Flux.trainmode!(m);
 
@@ -212,7 +208,7 @@ using the [`Scale`](@ref) layer.
 See also [`BatchNorm`](@ref), [`InstanceNorm`](@ref), [`GroupNorm`](@ref), and [`normalise`](@ref).
 
 # Examples
-```jldoctest
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
 julia> xs = rand(3, 3, 3, 2);  # a batch of 2 3X3X3 images
 
 julia> m = LayerNorm(3);
@@ -423,7 +419,7 @@ that will be used to renormalize the input in test phase.
 in previous Flux versions (< v0.12).
 
 # Examples
-```jldoctest
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
 julia> xs = rand(3, 3, 3, 2);  # a batch of 2 3X3X3 images
 
 julia> m = InstanceNorm(3);
@@ -521,7 +517,7 @@ If `track_stats=true`, accumulates mean and var statistics in training phase
 that will be used to renormalize the input in test phase.
 
 # Examples
-```jldoctest
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
 julia> xs = rand(3, 3, 4, 2);  # a batch of 2 3X3X4 images
 
 julia> m = GroupNorm(4, 2);
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -196,7 +196,7 @@ See also [`BatchNorm`](@ref), [`InstanceNorm`](@ref), [`GroupNorm`](@ref), and [
 ```jldoctest
 julia> using Statistics
 
-julia> xs = rand(3, 3, 3, 2);  # a batch of 2 3X3X3 images
+julia> xs = rand(3, 3, 3, 2);  # a batch of 2 images, each having 3 channels
 
 julia> m = LayerNorm(3);
 
@@ -312,7 +312,7 @@ Use [`testmode!`](@ref) during inference.
 ```julia
 julia> using Statistics
 
-julia> xs = rand(3, 3, 3, 2);  # a batch of 2 3X3X3 images
+julia> xs = rand(3, 3, 3, 2);  # a batch of 2 images, each having 3 channels
 
 julia> m = BatchNorm(3);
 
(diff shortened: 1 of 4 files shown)


Saransh-cpp:
I tried a lot of filters, but I can't get the doctests for the Dropout layer to pass. The primary cause for this failure is the presence of `0.0` and `-0.0` in the output, which the doctests can't handle.

For example, a regex pattern available here - https://regex101.com/r/1KtxKj/1 - correctly handles the `0.0` and `-0.0` case, but it fails while running the doctests.

I have converted the `jldoctest` block to a `julia` block for the moment, but I will revert it back if I find something.

Saransh-cpp:
Additionally, should the `dropout` function stay in the manual or should it be removed (internal API?)? If it stays, should I add examples for it too?

ToucheSir:
Instead of trying to match the output, you can add a check that the number of zeros  is equal to the dropout rate +- some margin of error.

`dropout` is 100% an internal function.

src/layers/normalise.jl
```diff
@@ -105,6 +134,29 @@ The AlphaDropout layer ensures that mean and variance of activations
...
+1.097500619939126
+
+julia> Flux.std(y)  # maintains the standard deviation of the input
+1.1504012188827453
+
+julia> Flux.mean(x)  # maintains the mean of the input
+-0.3217018554158738
+
+julia> Flux.mean(y)
+-0.2526866470385106
```
ToucheSir, line 158: These should be unprefixed and imported from `Statistics` if not already.

src/layers/normalise.jl
```diff
@@ -154,6 +206,27 @@ If `affine=true`, it also applies a learnable shift and rescaling
...
+julia> Flux.std(xs[:, :, :, 1])
+0.28713812337208383
+
+julia> Flux.std(y[:, :, :, 1])  # normalises each image (or all channels in an image)
+1.018993632693022
+
+julia> Flux.std(xs[:, :, :, 2])
+0.22540260537916373
+
+julia> Flux.std(y[:, :, :, 2])  # normalises each image (or all channels in an image)
```
ToucheSir, line 227: Ditto here.

src/layers/normalise.jl
```diff
@@ -339,6 +417,27 @@ that will be used to renormalize the input in test phase.
...
+
+# Examples
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
+julia> xs = rand(3, 3, 3, 2);  # a batch of 2 3X3X3 images
+
+julia> m = InstanceNorm(3);
+
+julia> y = m(xs);
+
+julia> Flux.std(xs[:, :, 1, 1])  # original standard deviation of the first channel of image 1
```
ToucheSir, line 429: Instead of using comments and having to filter out the random numbers, why not compare directly against the thing in the comment? e.g: `std(y[:, :, 1, 1]) \approx 1 && std(y[:, :, 1, 1]) \notapprox std(xs[:, :, 1, 1])` (can't find the symbols right now)

src/layers/recurrent.jl
```diff
@@ -63,28 +63,97 @@ in the background. `cell` should be a model of the form:
...
+julia> rnn(2) 
+2
+
+julia> rnn(3)
+3
+
+julia> rnn.state
+5
+
+julia> rnn.(1:10)  # apply to a sequence
```
ToucheSir, line 83: This uses broadcasting and it should not.

src/layers/recurrent.jl
```diff
@@ -63,28 +63,97 @@ in the background. `cell` should be a model of the form:
...
+
+julia> rnn([3])
+1-element Vector{Int64}:
+ 3
+
+julia> rnn.state
+1×1 Matrix{Int64}:
+ 5
+
+julia> rnn(reshape(1:10, 1, 1, :))  # apply to a sequence of (features, batch, time)
```
ToucheSir, line 121: Consider showing the output size and contents separately. That way you can use something like `vec` for the latter to save a lot of vertical space.

src/layers/recurrent.jl
```diff
@@ -107,8 +176,28 @@ Base.show(io::IO, m::Recur) = print(io, "Recur(", m.cell, ")")
...
-```julia
-rnn.state = hidden(rnn.cell)
+
+    rnn.state = hidden(rnn.cell)
+
+# Examples
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
+julia> r = RNN(1 => 1);
+
+julia> a = Vector{Float32}([1])
```
ToucheSir, line 186:
```suggestion
julia> a = ones(Float32, 1)
```

src/layers/upsample.jl
```diff
@@ -75,9 +75,20 @@ end
...
+Pixel shuffling layer with upscale factor `r`. Usually used for generating higher
+resolution images while upscaling them.
  
 See [`NNlib.pixel_shuffle`](@ref).
+
+# Examples
+```jldoctest
+julia> p = PixelShuffle(2);
+
+julia> xs = rand(2, 2, 4, 1); # an image with 4 channels having 2X2 pixels in each channel
```
ToucheSir, line 87: For this, I think it would actually be good to feed a fixed input through and display how the output changes because the operation is not terribly intuitive.

Saransh-cpp:
Thank you for the review, @ToucheSir!

> These should be unprefixed and imported from Statistics if not already.

Should I import Statistics in every doctest or should I import it only in the first doctest and group the other doctests together?

> Instead of using comments and having to filter out the random numbers, why not compare directly against the thing in the comment? e.g: std(y[:, :, 1, 1]) \approx 1 && std(y[:, :, 1, 1]) \notapprox std(xs[:, :, 1, 1]) (can't find the symbols right now)

Yes, this would make the doctests much cleaner. However, I could not use the `\approx` symbol here due to the default low tolerance. Should I use the `isapprox` function with `atol` = some quantity (or perhaps `rtol`) here?

> For this, I think it would actually be good to feed a fixed input through and display how the output changes because the operation is not terribly intuitive.

I did this initially (with random values, I really need to stop using random values), but then the extra vertical space taken changed my mind. I'll revert it back!

src/layers/recurrent.jl
```diff
@@ -63,28 +63,86 @@ in the background. `cell` should be a model of the form:
...
+julia> rnn.state
+1×1 Matrix{Int64}:
+ 5
+
+julia> vec = rnn(reshape(1:10, 1, 1, :));  # apply to a sequence of (features, batch, time)
+
+julia> size(vec)
+(1, 1, 10)
+
+julia> vec
```
ToucheSir, line 110: Just to clarify, by "vec" I meant literally calling `vec` on the RNN output to get a `Vector` (easier to print). The variable name was fine before :)

src/layers/normalise.jl
```diff
@@ -65,6 +65,27 @@ Specify `rng` to use a custom RNG instead of the default.
...
+
+# Examples
+```jldoctest
+julia> m = Chain(Dense(1 => 1), Dropout(1));
+
+julia> Flux.trainmode!(m);
+
+julia> y = m([1]);
+
+julia> count(i->(i == 0), y) == m[2].p  # number of zeros == 1
```
ToucheSir, line 77:
```suggestion
julia> y == [0]
```
Expresses more or less the same thing.

src/layers/normalise.jl
```diff
@@ -65,6 +65,27 @@ Specify `rng` to use a custom RNG instead of the default.
...
+true
+
+julia> m = Chain(Dense(1 => 1), Dropout(0.5));
+
+julia> Flux.trainmode!(m);
+
+julia> y = m([1]);
+
+julia> m[2].p - 0.5 <= count(i->(i == 0), y) <= m[2].p + 0.5  # number of zeros can be 0 or 1
```
ToucheSir, line 86:
```suggestion
julia> y = m(ones(1000));

julia> isapprox(count(==(0), y) / length(y), 0.5, atol=0.1)
```
I'm not sure if this tolerance is enough, so you may have to do some tweaking.

src/layers/normalise.jl
```diff
@@ -105,6 +126,22 @@ The AlphaDropout layer ensures that mean and variance of activations
...
+
+julia> x = randn(20,1);
+
+julia> m = Chain(Dense(20 => 10, selu), AlphaDropout(0.5));
+
+julia> Flux.trainmode!(m);
+
+julia> y = m(x);
+
+julia> isapprox(std(x), std(y), rtol=0.6)
```
ToucheSir, line 142: Note that `rtol == atol` for `isapprox` when the inputs are scalar numbers.

src/layers/recurrent.jl
Saransh-cpp: Oh, my bad😅

src/layers/normalise.jl
ToucheSir: I stand corrected. `rtol` != `atol`, but the `norm` computation that would happen while comparing arrays in `rtol*max(norm(x), norm(y))` is a no-op for scalars.

Saransh-cpp: Oh, yes, I just tested it. `atol` works fine with the given tolerance, tried a bunch of times, no errors!

src/layers/normalise.jl
```diff
@@ -154,6 +191,23 @@ If `affine=true`, it also applies a learnable shift and rescaling
 using the [`Scale`](@ref) layer.
 
 See also [`BatchNorm`](@ref), [`InstanceNorm`](@ref), [`GroupNorm`](@ref), and [`normalise`](@ref).
+
+# Examples
+```jldoctest
+julia> using Statistics
+
+julia> xs = rand(3, 3, 3, 2);  # a batch of 2 3X3X3 images
```
ToucheSir, line 199: this is not quite correct. Recall that Flux uses WHCN layout for conv layers, which means the third 3 is actually the number of channels. For a 3D input, you'd have to add another 3 up front. i.e. 3x3x3x3x2.

src/layers/normalise.jl
```diff
@@ -154,6 +191,23 @@ If `affine=true`, it also applies a learnable shift and rescaling
...
+
+julia> m = LayerNorm(3);
+
+julia> y = m(xs);
+
+julia> isapprox(std(y[:, :, :, 1]), 1, atol=0.1) && std(xs[:, :, :, 1]) != std(y[:, :, :, 1])
```
ToucheSir, line 205: I feel like there's a way to make this terser using the `dims` kwarg of `std`, but I've not tried to write it out.

src/layers/recurrent.jl
```diff
@@ -107,8 +146,28 @@ Base.show(io::IO, m::Recur) = print(io, "Recur(", m.cell, ")")
...
 
 Assuming you have a `Recur` layer `rnn`, this is roughly equivalent to:
-```julia
-rnn.state = hidden(rnn.cell)
+
+    rnn.state = hidden(rnn.cell)
+
+# Examples
+```jldoctest; filter = r"[+-]?([0-9]*[.])?[0-9]+"
+julia> r = RNN(1 => 1);
```
ToucheSir, line 154: It would be interesting to set the weight and bias to fixed values (even `[1.0f0]` might work, but I'm sure you could come up with something better) and show deterministic outputs much like you have for the `accum` example above. Really hammer home the iterative nature of the RNN interface

Saransh-cpp: I tried initializing an RNN with weights by passing them in through the `RNNCell`'s constructor, but for some reason I cannot get it to work. Could you please let me know how can I set the weights to a fixed value?

ToucheSir: Can you shared what you did? That should work.

src/layers/normalise.jl
ToucheSir: I just tested and it seems both support multiple dimensions for `dims`? What issues were you running into there?

Saransh-cpp: I don't remember what I was trying, but yes, passing a vector to `dims` works. With some help from discourse I got it working! Thank you!

src/layers/recurrent.jl
Saransh-cpp: Not as clearly as shown by the `accum` example. I looked up in the documentation but couldn't figure out how to use `RNNCell`s to create an example similar to the one using `accum`. But, this example does show that `reset!` resets the state to `state0` of `RNNCell` rather than the initial `state` passed to `Recur`.

ToucheSir: You can choose a slightly easier to interpret activation function for the cell. I think sigmoid is a little hard to visualize, but there may be something better than just `identity`.

Saransh-cpp: `relu` might look good here. I have updated the example to use the weights instead of just setting them to 0 -

```julia
julia> r = Flux.RNNCell(relu, ones(1,1), zeros(1,1), ones(1,1), zeros(1,1));

julia> y = Flux.Recur(r, ones(1,1));

julia> y.state
1×1 Matrix{Float64}:
 1.0

julia> y(ones(1,1))  # relu(1*1 + 1)
1×1 Matrix{Float64}:
 2.0

julia> y.state
1×1 Matrix{Float64}:
 2.0

julia> Flux.reset!(y)
1×1 Matrix{Float64}:
 0.0

julia> y.state
1×1 Matrix{Float64}:
 0.0
```

ToucheSir: Looks good!

ToucheSir (approved):
Failure on Nightly is from ComponentArrays and can be safely ignored. Thanks for your patient work on this @Saransh-cpp !

Saransh-cpp:
🎉🎉

Issue #1949, opened 2022-04-22
Ada or ADA


KronosTheLate:
Because doing both
![image](https://user-images.githubusercontent.com/61620837/164649822-efe4c0ca-5f3e-424c-a88a-19533e5de405.png)

Seems like a bad idea...

ToucheSir:
This is the consequence of organic growth :P. For now, a PR adding a type alias would be welcome.

cossio:
I suggest to follow PyToch capitalizations (https://pytorch.org/docs/stable/optim.html):

Adam, AdaMax, Adadelta, AdamW, etc.

mcabbott:
Similar issue solved at https://github.com/FluxML/Optimisers.jl/pull/78, should be easy to copy over.

mcabbott:
Closed by #1996

PR #1996, opened 2022-06-13
replace ADAM with Adam and its variants thereof

Karthik-d-k:
This PR is towards solving good first issue #1949, wherein name changes are done for the following mentioned `Optimisers` -->
- ADAM         -->    Adam
- NADAM      -->    NAdam
- ADAMW      -->    AdamW
- RADAM      -->    RAdam
- OADAM      -->    OAdam
- ADAGrad    -->    AdaGrad
- ADADelta   -->    AdaDelta

`Xref` FluxML/Optimisers.jl#78

### Open Points : 
1. In `Xref` FluxML/Optimisers.jl#78, `@deprecate` is used to export optimisers with old names, but i coudn't able to figure out where to add these stuff (if needed), will update this PR once i get this info from reviewers.
2. Please let me know if there are any more changes that needs to be done, as this is `My First PR` 😄 

### PR Checklist

- [x] Tests (updated w/ new names)
- [ ] Entry in NEWS.md
- [x] Documentation (updated w/ new names)

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -73,10 +73,10 @@ GRUCell(in::Integer, out::Integer; kw...) = GRUCell(in => out; kw...)
 GRUv3Cell(in::Integer, out::Integer; kw...) = GRUv3Cell(in => out; kw...)
 
 # Optimisers with old naming convention
-@deprecate ADAM Adam
-@deprecate NADAM NAdam
-@deprecate ADAMW AdamW
-@deprecate RADAM RAdam
-@deprecate OADAM OAdam
-@deprecate ADAGrad AdaGrad
-@deprecate ADADelta AdaDelta
+Base.@deprecate_binding ADAM Adam
+Base.@deprecate_binding NADAM NAdam
+Base.@deprecate_binding ADAMW AdamW
+Base.@deprecate_binding RADAM RAdam
+Base.@deprecate_binding OADAM OAdam
+Base.@deprecate_binding ADAGrad AdaGrad
+Base.@deprecate_binding ADADelta AdaDelta
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -72,7 +72,7 @@ LSTMCell(in::Integer, out::Integer; kw...) = LSTMCell(in => out; kw...)
 GRUCell(in::Integer, out::Integer; kw...) = GRUCell(in => out; kw...)
 GRUv3Cell(in::Integer, out::Integer; kw...) = GRUv3Cell(in => out; kw...)
 
-# Deprecate Optimisers with old naming convention
+# Optimisers with old naming convention
 @deprecate ADAM Adam
 @deprecate NADAM NAdam
 @deprecate ADAMW AdamW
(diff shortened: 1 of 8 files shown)


ToucheSir:
Fixing CI failures aside, we'd also need deprecations set up for the old constructors and for both old and new to be exported (such that existing code doesn't break).

ToucheSir:
https://github.com/FluxML/Flux.jl/blob/master/src/deprecations.jl is the place for deprecations.

mcabbott:
I think the reason deprecations are a bit more tricky here is that these are defined in their own sub-module, exported into Flux

https://github.com/FluxML/Flux.jl/blob/master/src/optimise/Optimise.jl

and then exported from there to the outside world:

https://github.com/FluxML/Flux.jl/blob/master/src/Flux.jl#L32-L35

My guess is that we should put `@deprecate` statements inside the Optimise sub-module, and will also need temporary `export` statements for the old names in Flux, in [src/deprecations.jl](https://github.com/FluxML/Flux.jl/blob/master/src/deprecations.jl)

Karthik-d-k:
@ToucheSir , @mcabbott I forgot to convert `ADADelta` to its correct form `AdaDelta`, hence the CI failures.
Now, it is fixed and CI tests are all through.
I have added deprecations in `deprecations.jl` and it worked without any trouble.

I haven't updated any docs warning the users to use new naming conventions, any ideas on how to accomplish this ?
I'm thinking of adding `Notes` saying that to use new names instead of old ones in `Flux.optimisers` documentation !!

ToucheSir:
AtomicGraphNets and NeuralPDE failures do not look to be related, but the aliases generated by `@deprecate` are not playing well with https://github.com/darsnack/ParameterSchedulers.jl/blob/main/src/ParameterSchedulers.jl#L86. Does anyone know how we can make the deprecated name a type as well?

src/deprecations.jl
```diff
@@ -71,3 +71,12 @@ LSTMCell(in::Integer, out::Integer; kw...) = LSTMCell(in => out; kw...)
...
 GRUv3Cell(in::Integer, out::Integer; kw...) = GRUv3Cell(in => out; kw...)
+
+# Optimisers with old naming convention
+@deprecate ADAM Adam
+@deprecate NADAM NAdam
+@deprecate ADAMW AdamW
+@deprecate RADAM RAdam
+@deprecate OADAM OAdam
+@deprecate ADAGrad AdaGrad
+@deprecate ADADelta AdaDelta
```
ToucheSir, line 82:
```suggestion
Base.@deprecate_binding ADAM Adam
Base.@deprecate_binding NADAM NAdam
Base.@deprecate_binding ADAMW AdamW
Base.@deprecate_binding RADAM RAdam
Base.@deprecate_binding OADAM OAdam
Base.@deprecate_binding ADAGrad AdaGrad
Base.@deprecate_binding ADADelta AdaDelta
```

ToucheSir:
Got a tip from Slack to use `@deprecate_binding` instead, hopefully this works.

ToucheSir (approved):
And successful. Let's get this merged.

Karthik-d-k:
Thanks for taking time to fix issues on behalf of me 😊 🚀

PR #1998, opened 2022-06-14
Miscellaneous docstring additions and fixes

### LINKED ISSUE #1990: [Discussion]: doctests, docstrings, documentation manual, and unclear internal API (for newcomers)
Hello! I am opening this issue for further discussions on docstrings, doctests, making internal API visible (by prepending `_`), and updating the manual. Before I open a PR, I think it would be best to discuss the documentation here. I hope this is okay as I couldn't find a `discussions` tab in `Flux`'s repository.

## onhot.jl
`onehot.jl` misses doctests for the structs returned by `onehotbatch` and `onehot`. I am not sure if these structs are supposed to be user-facing? The docstring of the struct `Embedding` uses `OneHotMatrix` directly, but other than that, I couldn't find any other use cases. Should I add some doctests to these structs? I can also add them to the manual if they are supposed to be user-facing.

Additionally, right now if a user tries to access the documentation of `OneHotVector` and `OneHotMatrix`, they are presented with the documentation (with the signature type) of `OneHotArray`. After reading Flux's code, the reason becomes obvious, but I don't think every user will want to dive into Flux's code. Should these structs have their independent docstrings?

## utils.jl
`utils.jl` has a few functions with no doctests, but it isn't clear if they are user-facing or not. From my previous PRs I have learned that the internal API doesn't need doctests, so should I prepend a `_` to these functions, or should I add doctests to them (if they are user-facing)? The functions -

**Not present in the manual** -
- `rng_from_array`
- `ones32`
- `zeros32`
- `create_bias`

**Present in the manual** -
- `rand32`
- `randn32`
- `throttle`

## functor.jl
The exact same case as that of `utils.jl` -

**Present in the manual** -
- `testmode!`
- `trainmode!`
- `f32`
- `f64`

Additionally, I think `cpu` and `gpu` are user-facing, but their APIs aren't documented in the manual.

## optimisers.jl
Currently, none of the optimisers have doctests, and I am not sure how to add them without excessive repetition. For example, I came up with the following doctest for the `Descent` optimiser -
```
julia> f(x) = 4x + 2;

julia> x = hcat(0:5...); y = f.(x);

julia> model = Dense(1 => 1);

julia> opt = Descent(0.01);

julia> loss(x, y) = Flux.Losses.mse(model(x), y);

julia> ps = Flux.params(model)  # initialised parameters
Params([Float32[-1.1358223;;], Float32[0.0]])

julia> loss(x, y)  # loss with initial parameters
297.14438f0

julia> gs = gradient(ps) do
                  loss(x, y)
              end;

julia> Flux.Optimise.update!(opt, ps, gs) # the parameters will be updated, and the loss should ideally go down

julia> ps
Params([Float32[-0.09425485;;], Float32[0.29679117]])

julia> loss(x, y)
191.4279f0
```
This does look a bit extra, but I am not sure how else I can document it properly. My main concern is that this would have to be repeated for all the optimisers, if it is added. A workaround can be initializing the same variables only once -
```julia
"""
    Descent(η = 0.1)
...
\```jldoctest optimisers
julia> f(x) = 4x + 2;

julia> x = hcat(0:5...); y = f.(x);

julia> model = Dense(1 => 1);

julia> opt = Descent(0.01);

julia> loss(x, y) = Flux.Losses.mse(model(x), y);

julia> ps = Flux.params(model)  # initialised parameters
Params([Float32[-1.1358223;;], Float32[0.0]])

julia> loss(x, y)  # loss with initial parameters
297.14438f0

julia> gs = gradient(ps) do
                  loss(x, y)
              end;

julia> Flux.Optimise.update!(opt, ps, gs) # the parameters will be updated, and the loss should ideally go down

julia> ps
Params([Float32[-0.09425485;;], Float32[0.29679117]])

julia> loss(x, y)
191.4279f0
\```
"""
mutable struct Descent <: AbstractOptimiser
   ...
"""
    Momentum(η = 0.01, ρ = 0.9)
...
\```jldoctest optimisers
julia> opt = Momentum(0.03, 0.99);

julia> ps = Flux.params(model)
Params([Float32[-0.09425485;;], Float32[0.29679117]])

julia> loss(x, y)
191.4279f0

julia> gs = gradient(ps) do
                  loss(x, y)
       end;

julia> Flux.Optimise.update!(opt, ps, gs) # the parameters will be updated

julia> ps  # updated parameters
Params([Float32[2.4130664;;], Float32[1.013122]])

julia> loss(x, y)
31.88943f0
\```
"""
mutable struct Momentum <: AbstractOptimiser
   ...
```
I am not sure if I should proceed ahead with this.

Suggestions from everyone are welcome :)

### LINKED ISSUE #1993: [Bug]: Swapped `alpha` and `beta` in `tversky` loss?
The `tversky` loss has 2 parameters, $\alpha$ and $\beta$, and `Flux` internally calculates the value of $\alpha$ as 1 - $\beta$. The loss is defined as `1 - tversky index` -

$$1 - \frac{True Positives}{True Positives + \alpha \* False Positives + \beta \* False Negatives}$$

> the Tversky index is defined as:
S(P, G; α, β) = |P G| / (|P G| + α|P \ G| + β|G \ P|) (2)
where α and β control the magnitude of penalties for FPs and FNs, respectively.

`Flux` implements it as -
```julia
1 - sum(|y .* ŷ| + 1) / (sum(y .* ŷ + β*(1 .- y) .* ŷ + (1 - β)*y .* (1 .- ŷ)) + 1)
```

Code -
```julia
num = sum(y .* ŷ) + 1
den = sum(y .* ŷ + β * (1 .- y) .* ŷ + (1 - β) * y .* (1 .- ŷ)) + 1
1 - num / den
```

Notice how the term `(1 .- y) .* ŷ` (False Positives, I hope I am not wrong) is multiplied by $\beta$, whereas it should be multiplied with $\alpha$ (which is 1 - $\beta$). Similarly, the term `y .* (1 .- ŷ)` is multiplied with $\alpha$ (that is 1 - $\beta$), whereas it should be multiplied with $\beta$.

This makes the loss function behave in a manner opposite to its documentation. For example - 
```julia

julia> y = [0, 1, 0, 1, 1, 1];

julia> ŷ_fp = [1, 1, 1, 1, 1, 1];  # 2 false positive -> 2 wrong predictions

julia> ŷ_fnp = [1, 1, 0, 1, 1, 0];  # 1 false negative, 1 false positive -> 2 wrong predictions

julia> Flux.tversky_loss(ŷ_fnp, y)
0.19999999999999996

julia> Flux.tversky_loss(ŷ_fp, y)  # should be smaller than tversky_loss(ŷ_fnp, y), as FN is given more weight
0.21875
```
Here the loss for `ŷ_fnp, y` should have been larger than the loss for `ŷ_fp, y` as the loss should give more weight or penalize the False Negatives (default $\beta$ is 0.7; hence it should give more weight to FN), but the exact opposite happens.

Changing the implementation of the loss -
```julia
julia> y = [0, 1, 0, 1, 1, 1];

julia> ŷ_fp = [1, 1, 1, 1, 1, 1];  # 2 false positive -> 2 wrong predictions

julia> ŷ_fnp = [1, 1, 0, 1, 1, 0];  # 1 false negative, 1 false positive -> 2 wrong predictions

julia> Flux.tversky_loss(ŷ_fnp, y)
0.19999999999999996

julia> Flux.tversky_loss(ŷ_fp, y)  # should be smaller than tversky_loss(ŷ_fnp, y), as FN is given more weight
0.1071428571428571
```
which looks right.

Is this a bug, or am I missing something? Would be happy to create a PR if it is a bug!

Saransh-cpp:
Added missing docstrings (in the manual), updated the existing ones, and added doctests in the following files -

- `stateless.jl`
- `losses/functions.jl`
- `utils.jl`

In addition, there are various miscellaneous docstring fixes, and I have also prepended an `_` to the internal API. This PR should close #1990.

This PR also closes #1993, but I am not 100% sure if the current implementation is wrong. I will revert the changes back if everything is correct.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -10,7 +10,7 @@ _dropout_shape(s, dims) = tuple((i ∉ dims ? 1 : si for (i, si) ∈ enumerate(s
 _dropout_kernel(y::T, p, q) where {T} = y > p ? T(1 / q) : T(0)
 
 """
-    dropout([rng = _rng_from_array(x)], x, p; dims=:, active=true)
+    dropout([rng = rng_from_array(x)], x, p; dims=:, active=true)
 
 The dropout function. If `active` is `true`,
 for each input, either sets that input to `0` (with probability
@@ -34,7 +34,7 @@ function dropout(rng, x, p; dims=:, active::Bool=true)
   y = dropout_mask(rng, x, p, dims=dims)
   return x .* y
 end
-dropout(x, p; kwargs...) = dropout(_rng_from_array(x), x, p; kwargs...)
+dropout(x, p; kwargs...) = dropout(rng_from_array(x), x, p; kwargs...)
 
 dropout_mask(rng::CUDA.RNG, x::CuArray, p; kwargs...) = _dropout_mask(rng, x, p; kwargs...)
 dropout_mask(rng, x::CuArray, p; kwargs...) =
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -10,7 +10,7 @@ _dropout_shape(s, dims) = tuple((i ∉ dims ? 1 : si for (i, si) ∈ enumerate(s
 _dropout_kernel(y::T, p, q) where {T} = y > p ? T(1 / q) : T(0)
 
 """
-    dropout([rng = rng_from_array(x)], x, p; dims=:, active=true)
+    dropout([rng = _rng_from_array(x)], x, p; dims=:, active=true)
 
 The dropout function. If `active` is `true`,
 for each input, either sets that input to `0` (with probability
@@ -34,7 +34,7 @@ function dropout(rng, x, p; dims=:, active::Bool=true)
   y = dropout_mask(rng, x, p, dims=dims)
   return x .* y
 end
-dropout(x, p; kwargs...) = dropout(rng_from_array(x), x, p; kwargs...)
+dropout(x, p; kwargs...) = dropout(_rng_from_array(x), x, p; kwargs...)
 
 dropout_mask(rng::CUDA.RNG, x::CuArray, p; kwargs...) = _dropout_mask(rng, x, p; kwargs...)
 dropout_mask(rng, x::CuArray, p; kwargs...) =
@@ -51,7 +51,7 @@ end
 ChainRulesCore.@non_differentiable dropout_mask(::Any, ::Any, ::Any)
 
 """
-    Dropout(p; dims=:, rng = rng_from_array())
+    Dropout(p; dims=:, rng = _rng_from_array())
 
 Dropout layer.
 
@@ -96,9 +96,9 @@ mutable struct Dropout{F,D,R<:AbstractRNG}
   active::Union{Bool, Nothing}
   rng::R
 end
-Dropout(p, dims, active) = Dropout(p, dims, active, rng_from_array())
+Dropout(p, dims, active) = Dropout(p, dims, active, _rng_from_array())
 
-function Dropout(p; dims=:, rng = rng_from_array())
+function Dropout(p; dims=:, rng = _rng_from_array())
   @assert 0 ≤ p ≤ 1
   Dropout(p, dims, nothing, rng)
 end
@@ -121,7 +121,7 @@ function Base.show(io::IO, d::Dropout)
 end
 
 """
-    AlphaDropout(p; rng = rng_from_array())
+    AlphaDropout(p; rng = _rng_from_array())
 
 A dropout layer. Used in
 [Self-Normalizing Neural Networks](https://arxiv.org/abs/1706.02515).
@@ -155,8 +155,8 @@ mutable struct AlphaDropout{F,R<:AbstractRNG}
     new{typeof(p), typeof(rng)}(p, active, rng)
   end
 end
-AlphaDropout(p, active) = AlphaDropout(p, active, rng_from_array())
-AlphaDropout(p; rng = rng_from_array()) = AlphaDropout(p, nothing, rng)
+AlphaDropout(p, active) = AlphaDropout(p, active, _rng_from_array())
+AlphaDropout(p; rng = _rng_from_array()) = AlphaDropout(p, nothing, rng)
 
 @functor AlphaDropout
 trainable(a::AlphaDropout) = (;)
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -51,7 +51,7 @@ end
 ChainRulesCore.@non_differentiable dropout_mask(::Any, ::Any, ::Any)
 
 """
-    Dropout(p; dims=:, rng = _rng_from_array())
+    Dropout(p; dims=:, rng = default_rng_value())
 
 Dropout layer.
 
@@ -96,9 +96,9 @@ mutable struct Dropout{F,D,R<:AbstractRNG}
   active::Union{Bool, Nothing}
   rng::R
 end
-Dropout(p, dims, active) = Dropout(p, dims, active, _rng_from_array())
+Dropout(p, dims, active) = Dropout(p, dims, active, default_rng_value())
 
-function Dropout(p; dims=:, rng = _rng_from_array())
+function Dropout(p; dims=:, rng = default_rng_value())
   @assert 0 ≤ p ≤ 1
   Dropout(p, dims, nothing, rng)
 end
@@ -121,7 +121,7 @@ function Base.show(io::IO, d::Dropout)
 end
 
 """
-    AlphaDropout(p; rng = _rng_from_array())
+    AlphaDropout(p; rng = default_rng_value())
 
 A dropout layer. Used in
 [Self-Normalizing Neural Networks](https://arxiv.org/abs/1706.02515).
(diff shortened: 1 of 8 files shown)


Saransh-cpp:
cc: @DhairyaLGandhi

src/layers/normalise.jl
```diff
@@ -10,7 +10,7 @@ _dropout_shape(s, dims) = tuple((i ∉ dims ? 1 : si for (i, si) ∈ enumerate(s
 _dropout_kernel(y::T, p, q) where {T} = y > p ? T(1 / q) : T(0)
 
 """
-    dropout([rng = rng_from_array(x)], x, p; dims=:, active=true)
+    dropout([rng = _rng_from_array(x)], x, p; dims=:, active=true)
```
mcabbott, line 13: Maybe this shouldn't mention this internal function? If you aren't looking at the source, you don't need to know it exists:
```suggestion
    dropout([rng], x, p; dims=:, active=true)
```
Alternatively, if it's a function you may need to know about & overload for your `x`, then it should be left as `rng_from_array` and included in the manual.

src/layers/normalise.jl
```diff
@@ -51,7 +51,7 @@ end
 ChainRulesCore.@non_differentiable dropout_mask(::Any, ::Any, ::Any)
 
 """
-    Dropout(p; dims=:, rng = rng_from_array())
+    Dropout(p; dims=:, rng = _rng_from_array())
```
mcabbott, line 54: Same comment:
```suggestion
    Dropout(p; dims=:, [rng])
```

src/layers/normalise.jl
```diff
@@ -121,7 +121,7 @@ function Base.show(io::IO, d::Dropout)
 end
 
 """
-    AlphaDropout(p; rng = rng_from_array())
+    AlphaDropout(p; rng = _rng_from_array())
```
mcabbott, line 124:
```suggestion
    AlphaDropout(p; [rng])
```

src/utils.jl
```diff
@@ -461,33 +461,40 @@ zeros32(dims::Integer...) = Base.zeros(Float32, dims...)
...
 
 """
     rand32([rng], size...)
-    randn32([rng], size...)
 
-Return an `Array{Float32}` of the given `size`, filled like `rand` or `randn`.
+Return an `Array{Float32}` of the given `size`, filled like `rand`.
 When the size is not provided, `rand32(rng::AbstractRNG)` returns a function.
 """
 rand32(dims::Integer...) = Base.rand(Float32, dims...)
```
mcabbott, line 482: Should this also call `_rng_from_array()`?

src/layers/normalise.jl
Saransh-cpp: Interesting. The docstring does redirect users to the `Dropout` construct, so should we be concerned with users using this function?

If users are indeed using the `dropout` function, then it makes sense to treat `rng_from_array` as a public function and add it to the manual.

Saransh-cpp: The majority of docstrings use the following notation for specifying the default value as `rng_from_array`-
```
[rng=GLOBAL_RNG]
``` 
Example -
```
    kaiming_uniform([rng=GLOBAL_RNG], size...; gain = √2) -> Array
    kaiming_uniform([rng]; kw...) -> Function
```
```
    glorot_normal([rng=GLOBAL_RNG], size...; gain = 1) -> Array
    glorot_normal([rng]; kw...) -> Function
```
etc.

I think these should also be changed. Now that I think of it, maybe we should keep it as `rng_from_array` and add it to the manual.

ToucheSir: One point against mentioning `rng_from_array` in the docstring is that these functions don't actually use it! For most cases users will encounter it will behave as if it were, but there are enough edge cases that I'd prefer not to unintentionally mislead about expected behaviour.

src/utils.jl
Saransh-cpp: Should `_rng_from_array()` be added in the following way? -
```julia
rand32(dims::Integer...) = Base.rand(_rng_from_array(), Float32, dims...)
```

src/layers/normalise.jl
Saransh-cpp: Ah, so should I keep it private, and update the docstring as `[rng]`?

ToucheSir: I stand corrected, it seems the functions in question were updated very recently to use it. Don't see much hurt in documenting `rng_from_array` and just noting it's an internal function, but you may want to wait for others to chime in. Another idea would be to call the zero-arg `rng_from_array` something else and document that. The current name is a bit of a misnomer anyways because there is no array involved in that method...

Saransh-cpp: Oh yes, something like `default_rng_value()`? Then maybe add both of them to the manual?

Saransh-cpp:
I have renamed `rng_from_array()` (with no arguments) to `default_rng_value()`, and added both of them to the manual.

test/losses.jl
```diff
@@ -163,8 +163,8 @@ y = [1.0 0.5 0.3 2.4]
 end
 
 @testset "tversky_loss" begin
-  @test Flux.tversky_loss(ŷ, y) ≈ -0.06772009029345383
-  @test Flux.tversky_loss(ŷ, y, β=0.8) ≈ -0.09490740740740744
+  @test Flux.tversky_loss(ŷ, y) ≈ 0.028747433264887046
+  @test Flux.tversky_loss(ŷ, y, β=0.8) ≈ 0.050200803212851364
```
ToucheSir, line 167: How did this change come about? It seems very drastic given the inputs have not changed.

Saransh-cpp: I have tweaked the implementation here - https://github.com/FluxML/Flux.jl/pull/1998/files#diff-0ff18e54a53b74bc46260164231d6265a35569aa9a367cedda9f8baee211df42R534-R538.

See #1993. I'll revert the changes if the original implementation is correct. (Should have dealt with this in a separate PR, my bad.)

ToucheSir: Yes, it would be best to revert for this PR. I'll follow up in #1993.

Saransh-cpp: Reverted!

ToucheSir:
Thanks @Saransh-cpp !

PR #2005, opened 2022-06-23
Improve type stability of LayerNorm and Dropout

ToucheSir:
These two layers made use of explicit or implicit control flow (e.g. default keyword argument values) which Zygote does not like. This PR is essentially a set of small hacks to work around that.

Any ideas on how to avoid `return_type` in `_dropout` would be much appreciated, but for now it seems to work.

TODO benchmarks.

### PR Checklist

- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -2,9 +2,14 @@ istraining() = false
 
 ChainRulesCore.rrule(::typeof(istraining)) = true, _ -> (NoTangent(),)
 
-_isactive(m) = isnothing(m.active) ? istraining() : Bool(m.active)
+_isactive(m) = Bool(something(m.active, istraining()))
 
-ChainRulesCore.@non_differentiable _isactive(::Any)
+# Avoids instabilities from differentiating through getproperty(m, :active)
+function ChainRulesCore.rrule(::typeof(_isactive), m)
+  training, _ = rrule(istraining)
+  _isactive_pullback(_) = (NoTangent(), NoTangent())
+  return Bool(something(m.active, training)), _isactive_pullback
+end
 
 _dropout_shape(s, ::Colon) = size(s)
 _dropout_shape(s, dims) = tuple((i ∉ dims ? 1 : si for (i, si) ∈ enumerate(size(s)))...)
@@ -57,7 +62,7 @@ end
 
 function (pb::DropoutPullback)(dy)
   dx = pb.project(_apply_mask(dy, pb.mask))
-  return (NoTangent(), NoTangent(), dx, NoTangent())
+  return (NoTangent(), NoTangent(), dx, NoTangent(), NoTangent(), NoTangent())
 end
 
 _apply_mask(x, ::Nothing) = x
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -2,7 +2,9 @@ istraining() = false
 
 ChainRulesCore.rrule(::typeof(istraining)) = true, _ -> (NoTangent(),)
 
-_isactive(m) = isnothing(m.active) ? istraining() : m.active
+_isactive(m) = isnothing(m.active) ? istraining() : Bool(m.active)
+
+ChainRulesCore.@non_differentiable _isactive(::Any)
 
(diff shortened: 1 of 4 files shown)


src/layers/normalise.jl
```diff
@@ -31,26 +33,50 @@ automatically managed using the [`Dropout`](@ref) layer instead of the
 
 The [`Dropout`](@ref) layer is what you should use in most scenarios.
 """
-function dropout(rng, x, p; dims=:, active::Bool=true)
-  active || return x
-  y = dropout_mask(rng, x, p, dims=dims)
-  return x .* y
-end
+dropout(rng, x, p; dims=:, active::Bool=true) = _dropout(rng, x, p; dims, active)
```
CarloLucibello, line 36:
```suggestion
dropout(rng, x, p; dims=:, active::Bool=true) = _dropout(rng, x, p, dims, active)
```

ToucheSir: Thanks. There's probably a few more things to fix before this is ready, so I'm integrating them locally before tossing it to CI.

ToucheSir:
TTFG timings using the following snippet:
<details>

<summary>Test code</summary>

```julia
using Metalhead, Flux, Zygote
using Metalhead: ChannelLayerNorm

model = ConvNeXt(:tiny; inchannels=1, nclasses=1).layers
# ChannelLayerNorm isn't type stable yet (for the same reason as LayerNorm wasn't),
# So remove it for this demo
model = fmap(Returns(identity), model; exclude=Base.Fix2(isa, ChannelLayerNorm))

# display(model); println()

loss(m, x) = sum(m(x))

inputs = randn(Float32, 32, 32, 1, 1)
# @time loss(model, inputs)
# @time loss(model, inputs)

loss_grad(m, x) = gradient((m, x) -> loss(m, x), m, x)

@time loss_grad(model, inputs)
# @time loss_grad(model, inputs)
```

</details>

```
julia> @time loss_grad(model, inputs)
 34.835647 seconds (87.12 M allocations: 4.701 GiB, 3.14% gc time, 99.38% compilation time) # 0.13.3
 30.679322 seconds (78.88 M allocations: 4.300 GiB, 3.46% gc time, 98.96% compilation time) # this PR
```
Replacing the `Chain{Vector}` with a `Chain{Tuple}` creates a larger gap:
```
julia> @time loss_grad(model, inputs)
 79.846248 seconds (98.87 M allocations: 5.243 GiB, 1.68% gc time, 99.67% compilation time) # 0.13.3
 63.024710 seconds (79.23 M allocations: 4.245 GiB, 1.92% gc time, 99.45% compilation time) # this PR
 52.838056 seconds (70.81 M allocations: 3.745 GiB, 1.98% gc time, 99.60% compilation time) # this PR + Zygote#1248
```

ToucheSir:
For kicks, here is Diffractor with https://github.com/JuliaDiff/ChainRules.jl/pull/644 and a _tuple_ Chain (vec Chain doesn't work):
```julia
julia> @time loss_grad(model, inputs)
 30.442982 seconds (92.61 M allocations: 4.148 GiB, 3.18% gc time, 89.07% compilation time)
```
Re-enabling `ChannelLayerNorm` adds but ~1s to the total. This is faster than Zygote _with_ vec Chain!

theabhirath:
Does Diffractor already work with most Flux models (or at least those with built-in layers)? I was under the impression that it wasn't there yet 😅

ToucheSir:
Not OOTB, which is why that ChainRules PR is required.

chengchingwen:
@ToucheSir Could you try running the layer norm gradient with gpu? I have try that manual broadcast fusion before but `CUDA.time` said it actually allocated more gpu memory

ToucheSir:
You're right, it allocates one more time for over 2x the memory overhead. I also found this out the hard way recently while trying to fuse the RNN cell kernels for https://github.com/FluxML/Flux.jl/pull/2023, but forgot about the change here.

darsnack:
Any updates on this (like benchmarks after unfusing)?

CarloLucibello:

PR #2011, opened 2022-06-27
RFC: add issue forms

ToucheSir:
Reference on GH docs: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,0 +1,74 @@
+name: Bug report
+description: File a bug report
+title: '[Bug]: '
+labels:
+  - bug
+body:
+  - type: markdown
+    attributes:
+      value: >-
+        Thank you for taking the time to file an issue. Please see the
+        instructions below so that we can most effectively help you.
+
+
+        - Make sure you're reporting *a bug*; for general questions, please use
+        Discourse, Slack or Zulip.
+
+
+        - Always ensure you're using the latest version of the following
+        packages:
+          - Flux
+          - [Zygote](https://github.com/FluxML/Zygote.jl)
+          - [NNlib](https://github.com/FluxML/NNlib.jl)
+          - [CUDA](https://github.com/JuliaGPU/CUDA.jl)
+          - [MLUtils](https://github.com/JuliaML/MLUtils.jl)
+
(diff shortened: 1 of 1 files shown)


ToucheSir:
Linked to rendered template: https://github.com/FluxML/Flux.jl/blob/bde0d43c3d1cdb3f430f3e5ad9303a28acefcf65/.github/ISSUE_TEMPLATE/bug-report.yml

Live version to play with: https://github.com/ToucheSir/issue-form-example/issues/new/choose

If this format makes sense, we can add similar a similar template for feature requests.

mcabbott (commented):
Something like this might be nice. But I wonder if this version comes off a little too harsh, and is quite a wall of text to push at people. We do want to hear about problems. We don't at present have a large problem with too many issues.

Two worthwhile goals are, I think:
* Encouraging people to simplify a MWE, especially if they can isolate which package is at fault. It may not be obvious to all that Flux is so far from being a monolith. Zygote bugs which can be reproduced without Flux are easier to look into. (Moving the issue to a different repository is very easy for us.)
* Reminding people to try the latest version, and explaining how. Pkg can be pretty confusing. But I think `] activate --temp` is the magic command to know, to test on everything latest.

Saransh-cpp (commented):
A nice thing here would be to add a `config.yml` file. See [this](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser). Something like this -

```yaml
blank_issues_enabled: true 
contact_links:
  - name: Julia's Slack
    url: fix-me
    about: How do I do 'x' with Flux?
```

Saransh-cpp:
Asking about the OS and the Julia version would be nice too!

ToucheSir:
The template looks a lot less intimidating once rendered. However, I stopped iterating on it because the editing experience is rather clunky (no live preview, among other things).

One simpler alternative is to clone the [CUDA.jl issue templates](https://github.com/JuliaGPU/CUDA.jl/tree/master/.github/ISSUE_TEMPLATE), which are much shorter. This issue form was actually inspired by those.

ToucheSir:
Since this was last posted, we added org-wide templates which are both less onerous to develop and to fill out.

PR #2014, opened 2022-07-01
Make `Dropout` docs a little more user friendly

theabhirath:
Currently, the `Dropout` docs redirects users to `Flux.dropout`, which has slightly more confusing details than most users need to deal with. This PR remedies that by shifting some of the explanation from `Flux.dropout` to `Dropout`.


ToucheSir:
Looks like that docstring needs a bit of general TLC, but this looks good.

ToucheSir (approved):

PR #2016, opened 2022-07-05
A new linear regression tutorial

Saransh-cpp:
This PR -
- Adds a new "Getting Started" section in the docs which I will be working on for the next couple of weeks
- Moves `basics.md` and `overview.md` to the "Getting Started" section. Currently, there are no changes in the files, but they will be changed soon.
- Adds a Linear Regression example to the "Getting Started" section

For the linear regression guide -
- Should I add the backpropagation step in the first half, or should I leave it to the `Flux.Optimise.update!` step?
- This guide overlaps with a lot of other textual information present but scattered in Flux's docs. These other texts will also be updated or moved to a better place soon. Basics.md, for example, does something very similar with dummy data, and the current "Getting Started" guide does something similar but with pre-defined weights.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


Saransh-cpp:
The first thought I had was to add a hyperlink on the website's `Getting Started` page, which would redirect the user to the docs' `Getting Started` page (similar to the `Ecosystem` page). 

Should I also add the same tutorial (once approved) to the website, or should we link the section?

avik-pal:
I think there is a consensus to not use `Flux.params` API any longer. We should not introduce it into new tutorials being written.

Saransh-cpp:
Thank you for the suggestion! I took some time to go through `Optimisers.jl`, and I am assuming that the traditional `Flux` training method should be replaced with `Optimisers.jl`?

mcabbott:
Agree we should move away from the whole weird `Params` story. 

But perhaps this linear regression example should delay introducing Optimisers.jl  as long as possible. For simple gradient descent, just writing out something like this:
```
dLdW, _, _ = gradient(loss, W, x, y)
W .= W .- 0.1 .* dLdW
```
might be a better level than immediately introducing optimiser state etc. It's unfortunate that this has quite a few moving parts, unlike `train!`'s apparent simplicity (although really `train!` hides a lot & this is also confusing). Maybe Optimisers.jl ought to be introduced along with an explanation that adding momentum helps, so that you know why there is a state? 

Even something like this seems OK to me, pretty explicit, and makes you understand why you are about to see a tool for walking over the arrays:
```
m = Dense(1 => 1)
for step in 1:10
  dLdm, _, _ = gradient(loss, m, x, y)
  
  m.weight .= m.weight .- 0.1 .* dLdm.weight
  m.bias .= m.bias .- 0.1 .* dLdm.bias
end
```

One more comment. In such tutorials, things like `params = Flux.params(W, b)` seem super-confusing. It would be nice to choose variable names which are very clearly things you've chosen, not features of Flux. `flux_model` is good.

(Link to rendered version: https://github.com/Saransh-cpp/Flux.jl/blob/linear-regression/docs/src/getting_started/linear_regression.md )

Saransh-cpp:
Agreed, we should keep the linear regression example simple. I will update the tutorial to show the gradient descent algorithm in action.

I have been working on the logistic regression example locally and will update that with the same.

Saransh-cpp:
I tried writing out the algorithm using `gradient` but I am not sure if I am doing this right, especially the `MLDatasets` example -

```julia
model(W, b, x) = @. W*x + b


function loss(W, b, x, y)
     ŷ = model(W, b, x)
     sum((y .- ŷ).^2) / length(x)
end

function train_model()
    dLdW, dLdb, _, _ = gradient(loss, W, b, x, y)
    
    @. W = W - 0.1 * dLdW
    @. b = b - 0.1 * dLdb
end
```

Saransh-cpp:
I'll update this with the new `train!` definition once #2029 is merged. Right now this PR does not use `train!` in any way.

Saransh-cpp:
Test: @ModelZookeeper commands

docs/src/getting_started/linear_regression.md
```diff
@@ -0,0 +1,488 @@
...
+
+## A simple linear regression model
+Let us start by building a simple linear regression model. This model would be trained on the data points of the form `(x₁, y₁), (x₂, y₂), ... , (xₙ, yₙ)`. In the real world, these `x`s can have multiple features, and the `y`s denote a label. In our example, each `x` has a single feature; hence, our data would have `n` data points, each point mapping a single feature to a single label.
+
+Importing the required `Julia` packages -
+
+```jldoctest linear_regression_simple
+julia> using Flux
+
+julia> using Plots
```
mcabbott, line 13:
```suggestion
julia> using Flux, Plots
```
Like this you get one prompt to install them, not two, if you don't have them.

docs/src/getting_started/linear_regression.md
```diff
@@ -0,0 +1,488 @@
...
+
+
+```jldoctest linear_regression_simple
+julia> x = x .* reshape(rand(Float32, 61), (1, 61));
+```
+
+Visualizing the final data -
+
+```jldoctest linear_regression_simple
+julia> plot(reshape(x, (61, 1)), reshape(y, (61, 1)), lw = 3, seriestype = :scatter, label = "", title = "Generated data", xlabel = "x", ylabel= "y");
```
mcabbott, line 52: Shorter, maybe clearer?
```suggestion
julia> plot(vec(x), vec(y), lw = 3, seriestype = :scatter, label = "", title = "Generated data", xlabel = "x", ylabel= "y");
```

docs/src/getting_started/linear_regression.md
```diff
@@ -0,0 +1,488 @@
...
+julia> x_train_n = Flux.normalise(x_train);
+
+julia> std(x_train_n)
+1.0000843694328236
+```
+
+The standard deviation is now close to one! Our data is ready!
+
+### Model
+We can now directly use `Flux` and let it do all the work internally! Let's define a model that takes in 13 inputs (13 features) and gives us a single output (the label). We will then pass our entire data through this model in one go, and `Flux` will handle everything for us! Remember, we could have declared a model in plain `Julia` as well. The model will have 14 parameters, 13 weights, and one bias.
```
mcabbott, line 310:
```suggestion
We can now directly use `Flux` and let it do all the work internally! Let's define a model that takes in 13 inputs (13 features) and gives us a single output (the label). We will then pass our entire data through this model in one go, and `Flux` will handle everything for us! Remember, we could have declared a model in plain `Julia` as well. The model will have 14 parameters: 13 weights and 1 bias.
```

docs/src/getting_started/linear_regression.md
```diff
@@ -0,0 +1,488 @@
...
+After getting familiar with the basics of `Flux` and `Julia`, we moved ahead to build a machine learning model for a real dataset. We repeated the exact same steps, but this time with a lot more features and data points, and by harnessing `Flux`'s full capabilities. In the end, we developed a training loop that was smarter than the hardcoded one and ran the model on our normalised dataset to conclude the tutorial.
+
+## Copy-pastable code
+### Dummy dataset
+```julia
+using Flux
+using Plots
+
+# data
+x = hcat(collect(Float32, -3:0.1:3)...)
```
mcabbott, line 402: It's nice to encourage people to try this out. I fear that nobody will find this section until they have done plenty of copying pieces above. 

And that they will tend to drift out of sync. (Maybe this section should at least run as a smoke-test.)

Should we consider making these pages with Pluto, so that they are just a static version of something you can download and run?

mcabbott (commented):
I had some comments on an earlier version... perhaps useful but not a full review.

Saransh-cpp:
Thanks for the review, @mcabbott! Pluto sounds good! We can get rid of the copy-paste section if this is converted to a Pluto notebook (or file, not sure, will go through it in detail).

(`SciML` uses this copy-paste section at the top, but this code was too lengthy to be placed at the top)

Saransh-cpp:
@mcabbott, [JuliaManifolds/Manopt.jl](https://github.com/JuliaManifolds/Manopt.jl) renders `Pluto` notebooks in the `Documenter` documentation using the following `make.jl` contents - https://github.com/JuliaManifolds/Manopt.jl/blob/master/docs/make.jl#L1-L106.

This does look a bit hacky, and it also distorts the documentation when a user opens one of the rendered `Pluto` notebooks -
## Sidebar
### Normal page
image
### Rendered Pluto notebook
image

## Settings
### Normal page
image
### Rendered Pluto notebook
image

Is there a legitimate way to render `Pluto` notebooks in `Documenter`'s documentation? This hack does not look good to me. Alternatively, we could keep the section as it is and at the top link a `Pluto` notebook that can be downloaded for following along.

Here is how the converted `Pluto` notebook looks like - https://saransh-cpp.github.io/assets/pluto/linear_regression.jl.html

Note: I am just redirecting users to the `html` page generated by `Pluto` from here - https://saransh-cpp.github.io/blog/ (this will be removed once this PR is merged to avoid duplicate pages on the web).

Saransh-cpp:
Currently, https://fluxml.ai/Flux.jl/stable/models/overview/ does something similar but is not as extensive as this guide. Should it be removed, or should it also be put under the "Getting Started" guide with a better title?

ToucheSir:
What if anything do we lose if it's removed? I think it would be nice to do so, but any material in it which isn't covered elsewhere would need a new home.

Saransh-cpp:
I think the following text can be used at the top of this guide. The rest of this page is definitely a subset of this guide.

---

Flux is a pure Julia ML stack that allows you to build predictive models. Here are the steps for a typical Flux program:

- Provide training and test data
- Build a model with configurable parameters to make predictions
- Iteratively train the model by tweaking the parameters to improve predictions
- Verify your model

Under the hood, Flux uses a technique called automatic differentiation to take gradients that help improve predictions. Flux is also fully written in Julia so you can easily replace any layer of Flux with your own code to improve your understanding or satisfy special requirements.

Here's how you'd use Flux to build and train the most basic of models, step by step.

mcabbott:
> Currently, https://fluxml.ai/Flux.jl/stable/models/overview/ does something similar but is not as extensive as this guide. Should it be removed, or

One virtue of that page is that it's much shorter. 

I like this PR's story, it's a nice ground-up explanation. But if you have met some of this before, and want a faster path to seeing how to write it in Julia's syntax and using Flux's pieces, then perhaps you prefer the older one.

Not sure what the ideal way to organise this material is...

Saransh-cpp:
> But if you have met some of this before, and want a faster path to seeing how to write it in Julia's syntax and using Flux's pieces, then perhaps you prefer the older one.

Yes, this makes sense. Maybe converting it into a "Quickstart" page under the "Getting Started" section?

mcabbott:
In addition to the main manual

https://fluxml.ai/Flux.jl/stable/models/overview/

we also have some quite nice tutorials here:

https://fluxml.ai/tutorials.html

How do we make these all findable, and where does this new page go?

The main docs also seem a slightly awkward combination of introduction and reference. It's possible that the "Building models" heading should be split in two? Not sure. 

Screenshot 2022-08-29 at 15 12 58

Saransh-cpp:
> In addition to the main manual
>
> https://fluxml.ai/Flux.jl/stable/models/overview/
>
> we also have some quite nice tutorials here:
>
> https://fluxml.ai/tutorials.html
>
> How do we make these all findable, and where does this new page go?

While drafting the "Getting Started" section, I wanted to include only the guides that will get a user started with flux. The website tutorials should be the ones introducing something that a user doesn't find themselves engaged with when they begin with ML/DL or an ML/DL package, for example, GANs and Transfer Learning. I think the `DataLoader` tutorial should be moved to the `MLUtils` page and the [Deep Learning with Flux - A 60 Minute Blitz (September 2020)](https://fluxml.ai/tutorials/2020/09/15/deep-learning-flux.html) can be added to the "Getting Started" section.

Ideally, there should be a "Tutorials" heading on the docs' sidebar which should redirect users to the website's tutorials page. Similarly, the website's getting started page should redirect users to the docs' getting started section.

A nice [infographic](https://documentation.divio.com/#about-the-system) -

image

IMO, the `DataLoader` example is a "How-To Guide", the "Getting Started" guides are more inclined towards "Explanations", and the website tutorials are "Tutorials".

mcabbott:
I was looking at Turing's documentation a bit. They clearly face the same problem, that there are many levels at which you can introduce something. 

One thing I do like is that they have a compact example which basically does what the library does, up front, in one code block. Anyone can copy and run this in one go. Experts who know another package / language can see roughly what the notation is etc. Beginners will be mystified as to how it's working, but at least you get a plot which shows something worked.

My first attempt at such an example is here: https://gist.github.com/mcabbott/a61930793c5a064e10411ec427a6377a  . Whether this ought to be instead of, or in addition to, existing text I don't know.

Saransh-cpp:
> One thing I do like is that they have a compact example which basically does what the library does, up front, in one code block. Anyone can copy and run this in one go. Experts who know another package / language can see roughly what the notation is etc. Beginners will be mystified as to how it's working, but at least you get a plot which shows something worked.

Yes! These are the how-to guides. I find them extremely helpful as well!

I am not sure where these small tutorials should be added?

Saransh-cpp:
Hmm, where should this tutorial go now? We do have a new Getting Started section, so should this go in there (along with other tutorials)?

ToucheSir:
Good question. PyTorch and JAX have tutorial sections for this, but at present we've outsourced that to the site. Maybe we could add one and pull a couple of website tutorials in for a later PR (but link to them for now)? That would also help prevent bitrot of the tutorials on the site.

Saransh-cpp:
So would that mean shifting the tutorial section to the `Documenter` website completely? Maybe I can shift the existing section before adding this guide to the same. We could still keep the website navbar link, but instead, link it to the new section in the `Documenter` site (like the `ecosystem` page)?

ToucheSir:
I like that idea. It might be easier to create that section in this PR and do a follow-up for migrating the other tutorials.

Saransh-cpp:
This PR now creates a new "tutorials" section. I will migrate the website's tutorials here in the follow-up PRs.

mcabbott:
Note that there's already a tutorials section (on master) so simplest to add this there?

Saransh-cpp:
> Note that there's already a tutorials section (on master) so simplest to add this there?

Ah, I missed it because of how low it is in the sidebar. Maybe we should move it above the "Performance Tips" section?

docs/make.jl
```diff
@@ -1,19 +1,19 @@
...
     # strict = [:cross_references,],
     pages = [
         "Getting Started" => [
             "Welcome" => "index.md",
-            "Quick Start" => "models/quickstart.md",
-            "Fitting a Line" => "models/overview.md",
-            "Gradients and Layers" => "models/basics.md",
+            "Quick Start" => "getting_started/quickstart.md",
+            "Fitting a Line" => "getting_started/overview.md",
+            "Gradients and Layers" => "getting_started/basics.md",
```
mcabbott, line 16: Can I suggest leaving files where they are, until sure? I think moving them may break some links elsewhere. And we may re-organise this into Guide / Reference. 

(Adding `id` & linking by that, not heading name nor file name, seems like the right solution.)

Saransh-cpp: All the references associated with these pages now use `id`s! I have also reverted back the structural changes.

mcabbott (approved):
Let's do it. Tutorials is the right section, IMO.

PR #2021, opened 2022-07-16
Add a logistic regression example to the Getting Started section

Saransh-cpp:
This PR adds a new Logistic Regression example to the "Getting Started" section.

Note: This PR contains the changes from the linear regression PR (without some of these changes, the tests would fail). This PR should be merged after the linear regression example, and I will make sure to sync the changes before this PR is merged.

Rendered guide - https://github.com/Saransh-cpp/Flux.jl/blob/logistic-regression/docs/src/getting_started/logistic_regression.md

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/tutorials/logistic_regression.md
+++ b/docs/src/tutorials/logistic_regression.md
@@ -2,14 +2,14 @@
 
 The following page contains a step-by-step walkthrough of the logistic regression algorithm in Julia using Flux. We will then create a simple logistic regression model without any usage of Flux and compare the different working parts with Flux's implementation. 
 
-Let's start by importing the required `Julia` packages!
+Let's start by importing the required Julia packages.
 
 ```jldoctest logistic_regression
 julia> using Flux, Statistics, MLDatasets, DataFrames, OneHotArrays
 ```
 
 ## Dataset
-Let's start by importing a dataset from `MLDatasets.jl`! We will use the `Iris` dataset that contains the data of three different `Iris` species. The data consists of 150 data points (`x`s), each having four features. Each of these `x` is mapped to `y`, the name of a particular `Iris` specie.
+Let's start by importing a dataset from MLDatasets.jl. We will use the `Iris` dataset that contains the data of three different `Iris` species. The data consists of 150 data points (`x`s), each having four features. Each of these `x` is mapped to `y`, the name of a particular `Iris` specie.
 
 ```jldoctest logistic_regression
 julia> Iris()
@@ -22,7 +22,7 @@ dataset Iris:
 julia> x, y = Iris(as_df=false)[:];
 ```
 
-Our next step would be to convert this data into a form that can be fed to a machine learning model. The `x` values are arranged in a matrix and thus don't need any alterations, but the labels must be one hot encoded. [Here](https://discourse.julialang.org/t/all-the-ways-to-do-one-hot-encoding/64807) is a great discourse thread on different techniques that can be used to one hot encode data with or without using any external `Julia` package.
+Our next step would be to convert this data into a form that can be fed to a machine learning model. The `x` values are arranged in a matrix and thus don't need any alterations, but the labels must be one hot encoded. [Here](https://discourse.julialang.org/t/all-the-ways-to-do-one-hot-encoding/64807) is a great discourse thread on different techniques that can be used to one hot encode data with or without using any external Julia package.
 
 ```jldoctest logistic_regression
 julia> y_r = reshape(y, (150, 1));
@@ -34,7 +34,7 @@ julia> custom_y_onehot = unique(y_r) .== permutedims(y_r)
  0  0  0  0  0  0  0  0  0  0  0  0  0     1  1  1  1  1  1  1  1  1  1  1  1
 ```
 
-This same operation can also be performed using [OneHotArrays](https://github.com/FluxML/OneHotArrays.jl)' `onehotbatch` function! We will use both of these outputs parallelly to show how intuitive `FluxML` is!
+This same operation can also be performed using [OneHotArrays](https://github.com/FluxML/OneHotArrays.jl)' `onehotbatch` function. We will use both of these outputs parallelly to show how intuitive FluxML is!
 
 ```jldoctest logistic_regression
 julia> flux_y_onehot = onehotbatch(y_r, ["Iris-setosa", "Iris-versicolor", "Iris-virginica"])
@@ -45,7 +45,7 @@ julia> flux_y_onehot = onehotbatch(y_r, ["Iris-setosa", "Iris-versicolor", "Iris
  0  0  0  0  0  0  0  0  0  0  0  0  0     1  1  1  1  1  1  1  1  1  1  1  1
 ```
 
-Our data is ready! The next step would be to build a classifier for the same.
+Our data is ready. The next step would be to build a classifier for the same.
 
 ## Building a model
 
@@ -72,7 +72,7 @@ julia> W = rand(Float32, 3, 4);
 julia> b = [0.0f0, 0.0f0, 0.0f0];
 ```
 
-Now our model can take in the complete dataset and predict the class of each `x` in one go! But, we need to ensure that our model outputs the probabilities of an input belonging to the respective classes. As our model has three outputs, each would denote the probability of the input belonging to a particular class.
+Now our model can take in the complete dataset and predict the class of each `x` in one go. But, we need to ensure that our model outputs the probabilities of an input belonging to the respective classes. As our model has three outputs, each would denote the probability of the input belonging to a particular class.
 
 We will use an activation function to map our outputs to a probability value. It would make sense to use a `softmax` activation function here, which is defined mathematically as -
 
@@ -80,7 +80,7 @@ We will use an activation function to map our outputs to a probability value. It
 σ(\vec{x}) = \frac{\\e^{z_{i}}}{\\sum_{j=1}^{k} \\e^{z_{j}}}
 ```
 
-The `softmax` function scales down the outputs to probability values such that the sum of all the final outputs equals `1`. Let's implement this in `Julia`!
+The `softmax` function scales down the outputs to probability values such that the sum of all the final outputs equals `1`. Let's implement this in Julia.
 
 ```jldoctest logistic_regression
 julia> custom_softmax(x) = exp.(x) ./ sum(exp.(x), dims=1)
@@ -116,7 +116,7 @@ julia> sum(custom_model(W, b, x), dims=1)
 
 Every output value is between `0` and `1`, and every column adds to `1`!
 
-Let's convert our `custom_model` to a `Flux` model. `Flux` provides the users with a very elegant API that almost feels like writing your code!
+Let's convert our `custom_model` to a Flux model. Flux provides the users with a very elegant API that almost feels like writing your code!
 
 Note, all the `flux_*` variables in this tutorial would be general, that is, they can be used as it is with some other similar-looking dataset, but the `custom_*` variables will remain specific to this tutorial.
 
@@ -128,9 +128,9 @@ Chain(
 )
 ```
 
-A [`Dense(4 => 3)`](@ref Dense) layer denotes a layer with four inputs (four features in every data point) and three outputs (three classes or labels). This layer is the same as the mathematical model defined by us above! Under the hood, `Flux` too calculates the output using the same expression! But, we don't have to initialize the parameters ourselves this time, instead `Flux` does it for us.
+A [`Dense(4 => 3)`](@ref Dense) layer denotes a layer with four inputs (four features in every data point) and three outputs (three classes or labels). This layer is the same as the mathematical model defined by us above. Under the hood, Flux too calculates the output using the same expression, but we don't have to initialize the parameters ourselves this time, instead Flux does it for us.
 
-The `softmax` function provided by `NNLib.jl` is re-exported by `Flux`, which has been used here. Lastly, `Flux` provides users with a `Chain` struct which makes stacking layers seamless!
+The `softmax` function provided by NNLib.jl is re-exported by Flux, which has been used here. Lastly, Flux provides users with a `Chain` struct which makes stacking layers seamless.
 
 A model's weights and biases can be accessed as follows -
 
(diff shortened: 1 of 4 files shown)


Saransh-cpp:
TODO: switch to `OneHotArrays.jl`

Saransh-cpp:
Rebased and fixed the doctests!

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,377 @@
...
+
+```jldoctest logistic_regression
+julia> custom_model(W, b, x) |> size
+(3, 150)
+```
+
+It works! Let's check if the `softmax` function is working.
+
+```jldoctest logistic_regression
+julia> all(custom_model(W, b, x) .< 1.0f0) && all(custom_model(W, b, x) .> 0.0f0)
```
mcabbott, line 109: This can be simpler:
```suggestion
julia> all(0 .<= custom_model(x) .<= 1)
```

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,377 @@
+# Logistic Regression
+
+The following page contains a step-by-step walkthrough of the logistic regression algorithm in `Julia` using `Flux`! We will then create a simple logistic regression model without any usage of `Flux` and compare the different working parts with `Flux`'s implementation.
```
mcabbott, line 3: A minor style change I'd make globally is not to write package names as code. And perhaps to put a few less `!`s!
```suggestion
The following page contains a step-by-step walkthrough of the logistic regression algorithm in Julia using Flux. We will then create a simple logistic regression model without any usage of Flux and compare the different working parts with Flux's implementation. 
```

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,377 @@
...
+ 0  0  0  0  0  0  0  0  0  0  0  0  0     0  0  0  0  0  0  0  0  0  0  0  0
+ 0  0  0  0  0  0  0  0  0  0  0  0  0     1  1  1  1  1  1  1  1  1  1  1  1
+```
+
+This same operation can also be performed using [OneHotArrays](https://github.com/FluxML/OneHotArrays.jl)' `onehotbatch` function! We will use both of these outputs parallelly to show how intuitive `FluxML` is!
+
+```jldoctest logistic_regression
+julia> flux_y_onehot = onehotbatch(y_r, ["Iris-setosa", "Iris-versicolor", "Iris-virginica"])
+3×150×1 OneHotArray(::Matrix{UInt32}) with eltype Bool:
+[:, :, 1] =
```
mcabbott, line 42: Should this use just `y` to get `3×150` without training dim?

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,377 @@
...
+```
+
+The `softmax` function scales down the outputs to probability values such that the sum of all the final outputs equals `1`. Let's implement this in `Julia`!
+
+```jldoctest logistic_regression
+julia> custom_softmax(x) = exp.(x) ./ sum(exp.(x), dims=1)
+custom_softmax (generic function with 1 method)
+```
+
+The implementation looks straightforward enough! Note that we specify `dims=1` in the `sum` function to calculate the sum of probabilities across columns. Remember, we will have a `3X150` matrix (predicted `y`s) as the output of our model, where each column would be an output of a corresponding input.
```
mcabbott, line 90: I never know whether "across columns" means "perpendicular", or "comparing different columns", rather than "within one colunm, applied to all". Maybe this avoids that:
```suggestion
The implementation looks straightforward enough! Note that we specify `dims=1` in the `sum` function to calculate the sum of probabilities in each column. Remember, we will have a 3×150 matrix (predicted `y`s) as the output of our model, where each column would be an output of a corresponding input.
```

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,377 @@
...
+           custom_logitcrossentropy(ŷ, y)
+       end;
+
+julia> custom_loss(W, b, x, custom_y_onehot)
+1.1714406827505623
+```
+
+The loss function works!
+
+`Flux` provides us with many minimal yet elegant loss functions. In fact, the `custom_logitcrossentropy` defined above has been taken directly from `Flux`, the only difference being the functionalities provided by `Flux` is much more general. For example, you can use the `Flux.logitcrossentropy` with any dataset, but `custom_logitcrossentropy` is specific to this tutorial and would give weird results if used otherwise.
```
mcabbott, line 168: What does "more general" mean here?

Things you could say are that it includes some sanity-checks on the shapes, and that you are more likely to make a typo.

But I think that otherwise this is in fact identical, and a function you write isn't a 2nd class citizen. That's a feature.

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,377 @@
...
+The following page contains a step-by-step walkthrough of the logistic regression algorithm in `Julia` using `Flux`! We will then create a simple logistic regression model without any usage of `Flux` and compare the different working parts with `Flux`'s implementation. 
+
+Let's start by importing the required `Julia` packages!
+
+```jldoctest logistic_regression
+julia> using Flux, Statistics, MLDatasets, DataFrames, OneHotArrays
+```
+
+## Dataset
+Let's start by importing a dataset from `MLDatasets.jl`! We will use the `Iris` dataset that contains the data of three different `Iris` species. The data consists of 150 data points (`x`s), each having four features. Each of these `x` is mapped to `y`, the name of a particular `Iris` specie.
```
mcabbott, line 12: Maybe mention that this will download it the first time.

Actually I get an error, it wants me to load DataFrames. Maybe the tutorial should too?
```
julia> using MLDatasets

julia> Iris()
This program has requested access to the data dependency Iris.
which is not currently installed. It can be installed automatically, and you will not see this message again.

Dataset: The Iris dataset
Website: https://archive.ics.uci.edu/ml/datasets/Iris



Do you want to download the dataset from ["https://archive.ics.uci.edu/ml/machine-learning-databases/iris/iris.data"] to "/Users/me/.julia/datadeps/Iris"?
[y/n]
y
ERROR: Add `import DataFrames` or `using DataFrames` to your code to unlock this functionality.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] assert_imported
    @ ~/.julia/packages/MLDatasets/A3giY/src/require.jl:30 [inlined]
  [3] require_import
    @ ~/.julia/packages/MLDatasets/A3giY/src/require.jl:37 [inlined]
  [4] getproperty
    @ ~/.julia/packages/MLDatasets/A3giY/src/require.jl:23 [inlined]
  [5] read_csv_asdf(path::String; kws::Base.Pairs{Symbol, Int64, Tuple{Symbol}, NamedTuple{(:header,), Tuple{Int64}}})
    @ MLDatasets ~/.julia/packages/MLDatasets/A3giY/src/io.jl:16
```

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,377 @@
...
+
+```jldoctest logistic_regression
+julia> Iris()
+dataset Iris:
+  metadata   =>    Dict{String, Any} with 4 entries
+  features   =>    150×4 DataFrame
+  targets    =>    150×1 DataFrame
+  dataframe  =>    150×5 DataFrame
+
+julia> x, y = Iris(as_df=false)[:];
```
mcabbott, line 22: It's a little opaque what `[:]` is doing here. Is there any way to explain easily how you are meant to know to do that? 

`Iris().features` gives a DataFrame, whose columns are 
```
julia> names(Iris().features)
4-element Vector{String}:
 "sepallength"
 "sepalwidth"
 "petallength"
 "petalwidth"
```
And
```
julia> y
1×150 Matrix{InlineStrings.String15}:
 "Iris-setosa"  "Iris-setosa"  …  "Iris-virginica"

julia> x |> summary
"4×150 Matrix{Float64}"
```

Maybe most importantly, what exactly are x & y and what do they mean? How do they relate to flowers & why might we expect to be able to predict one from the other?

docs/src/tutorials/logistic_regression.md
Saransh-cpp: Yes, it has been loaded above. I will add a note about the dataset being downloaded the first time.

docs/src/tutorials/logistic_regression.md
Saransh-cpp: The official example of the dataset uses `[:]` to retrieve the x and y values - https://juliaml.github.io/MLDatasets.jl/stable/datasets/misc/#MLDatasets.Iris.

I will explain the dataset a bit.

docs/src/tutorials/logistic_regression.md
Saransh-cpp: Using `y` gives me a matrix of size `3x1x150` -
```julia
julia> flux_y_onehot = onehotbatch(y, ["Iris-setosa", "Iris-versicolor", "Iris-virginica"]) |> summary
"3×1×150 OneHotArray(::Matrix{UInt32}) with eltype Bool"
```

ToucheSir: I might've missed this, but why not use `y_r = vec(y)` or `vec(y)` directly?

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,396 @@
...
+We now move to the `accuracy` metric and run it with the untrained `custom_model`.
+
+```jldoctest logistic_regression; filter = r"[+-]?([0-9]*[.])?[0-9]+(f[+-]*[0-9])?"
+julia> custom_accuracy(W, b, x, y) = mean(custom_onecold(custom_model(W, b, x)) .== y_r);
+
+julia> custom_accuracy(W, b, x, y)
+0.3333333333333333
+```
+
+Alternatively, we can use the functionalities provided by Flux to define the `flux_accuracy` function.
```
ToucheSir, line 282:
```suggestion
We could also have used Flux's built-in functionality to define this accuracy function.
```

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,396 @@
...
+           custom_y_cold = Vector{String}(undef, size(custom_y_onehot)[2])
+           for i = 1:size(custom_y_onehot)[2]
+               if mxidx[i].I[1] == 1
+                   custom_y_cold[i] = "Iris-setosa"
+               elseif mxidx[i].I[1] == 2
+                   custom_y_cold[i] = "Iris-versicolor"
+               elseif mxidx[i].I[1] == 3
+                   custom_y_cold[i] = "Iris-virginica"
+               end
+           end
```
ToucheSir, line 232: Could this use `findfirst` instead of the if statement? The vectorized version of `findfirst` might be able to replace the loop too.

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,396 @@
...
+
+```jldoctest logistic_regression; filter = r"[+-]?([0-9]*[.])?[0-9]+(f[+-]*[0-9])?"
+julia> @show flux_accuracy(x, y);
+flux_accuracy(x, y) = 0.98
+
+julia> flux_loss(flux_model, x, flux_y_onehot)
+0.6952386604624324
+```
+
+The accuracy went up! The loss went down! Everything worked again!
```
ToucheSir, line 386:
```suggestion
We see a very similar final loss and accuracy.
```
Not 100% on this, but the thought is to re-emphasize how functionally similar the two approaches are.

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,396 @@
...
+```
+
+Here, `W` is the weight matrix, `b` is the bias vector, ``\eta`` is the learning rate, ``\frac{dL}{dW}`` is the derivative of the loss function with respect to the weight, and ``\frac{dL}{db}`` is the derivative of the loss function with respect to the bias.
+
+The derivatives are calculated using an Automatic Differentiation tool, and Flux uses [`Zygote.jl`](https://github.com/FluxML/Zygote.jl) for the same. Since Zygote.jl is an independent Julia package, it can be used outside of Flux as well! Refer to the documentation of Zygote.jl for more information on the same.
+
+Our first step would be to obtain the gradient of the loss function with respect to the weights and the biases. Flux re-exports Zygote's `gradient` function; hence, we don't need to import Zygote explicitly to use the functionality.
+
+```jldoctest logistic_regression
+julia> dLdW, dLdb, _, _ = gradient(custom_loss, W, b, x, custom_y_onehot);
```
ToucheSir, line 309: Thoughts on either closing over `x` and `custom_y_onehot` in an anonymous function to avoid having to `_, _ = ...`, or adding a brief sentence explaining why we're ignoring certain gradients?

ToucheSir (commented):
Let's get this landed :). Just a few small comments from me.

docs/src/tutorials/logistic_regression.md
Saransh-cpp: Oh yes, that should work!

docs/src/tutorials/logistic_regression.md
Saransh-cpp: i did manage to vectorize the loop and get rid off the if statements using just `argmax`!

docs/src/tutorials/logistic_regression.md
```diff
@@ -0,0 +1,387 @@
...
+
+Now we can write a function that calculates the indices of the maximum element in each column, and maps them to a class name.
+
+```jldoctest logistic_regression
+julia> function custom_onecold(custom_y_onehot)
+           max_el = string.([x[1] for x in Tuple.(argmax(custom_y_onehot, dims=1))])
+           max_el[max_el.=="1"] .= "Iris-setosa"
+           max_el[max_el.=="2"] .= "Iris-versicolor"
+           max_el[max_el.=="3"] .= "Iris-virginica"
+           vec(max_el)
```
ToucheSir, line 224:
```suggestion
           max_idx = [x[1] for x in argmax(custom_y_onehot; dims=1)]
           classes = ["Iris-setosa", "Iris-versicolor", "Iris-virginica"]
           return classes[max_idx]
```
Is what I was thinking of. It may help to extract the classes into a separate `const` variable if they're used elsewhere.

Saransh-cpp: That worked! Thanks!

ToucheSir (approved):
Thanks for your patience @Saransh-cpp, let's get this merged.

PR #2023, opened 2022-07-20
Use muladd for LSTM cell matmuls

ToucheSir:
This appears to help significantly on CPU, ref. https://discourse.julialang.org/t/slow-lstm-on-gpu-in-flux/84228. I have not yet benchmarked GPU performance, but it is likely a wash. It would be great to have a single function that does `x1 * y1 + x2 * y2`, does one exist in LinearAlgebra?

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -295,7 +295,7 @@ end
 
 function (m::LSTMCell{A,V,<:NTuple{2,AbstractMatrix{T}}})((h, c), x::Union{AbstractVecOrMat{T},OneHotArray}) where {A,V,T}
   b, o = m.b, size(h, 1)
-  g = m.Wi*x .+ m.Wh*h .+ b
+  g = muladd(m.Wi, x, muladd(m.Wh, h, b))
   input, forget, cell, output = multigate(g, o, Val(4))
   c′ = @. sigmoid_fast(forget) * c + sigmoid_fast(input) * tanh_fast(cell)
   h′ = @. sigmoid_fast(output) * tanh_fast(c′)


mcabbott:
Great if `muladd` helps something, I think I tried for some other layers and was disappointed.

It is a very simple `out .= z; mul!(out, A, B)` which you could easily extend to allow `(A*B + C*D) .+ z`, but as far as I know nobody has packaged that up.

mcabbott:
Timing as in this gist: https://gist.github.com/JLDC/66c8eed6c36cb9cda85ff6404284d841 (from Discourse thread above) I get CPU:
```
julia> main()  # 2nd run:
181.671379 seconds (273.90 k allocations: 49.164 GiB, 90.20% gc time)  # before
141.337037 seconds (257.10 k allocations: 34.292 GiB, 87.77% gc time)  # after

julia> main()  # on a different computer! Julia 1.8, first run
 46.989183 seconds (2.30 M allocations: 49.275 GiB, 36.97% gc time, 11.01% compilation time)  # before
 35.523793 seconds (2.27 M allocations: 34.402 GiB, 14.94% gc time, 14.46% compilation time)  # after
```
GPU, inserting `|> gpu` and `CUDA.@time`. Quite noisy, best of a few each, same device:
```
julia> main()
# before
  1.994243 seconds (2.26 M CPU allocations: 213.138 MiB, 7.88% gc time) (61.50 k GPU allocations: 49.165 GiB, 36.31% memmgmt time)
# after
  1.800311 seconds (2.20 M CPU allocations: 213.963 MiB, 4.43% gc time) (53.70 k GPU allocations: 34.287 GiB, 17.56% memmgmt time)

julia> CUDA.device()
CuDevice(3): Tesla P100-PCIE-16GB
```

mcabbott (approved):
So we should do this.

And probably write a `mulmuladd` which calls `mul!` twice.

PR #2024, opened 2022-07-22
Remove CTC loss (moved to NNlib)

mcabbott:
A step towards not needing to load CUDA.jl. This needs https://github.com/FluxML/NNlibCUDA.jl/pull/55 (and also https://github.com/FluxML/NNlib.jl/pull/426 which is tagged).

Tests are copied there, could leave them here for now, or delete.


mcabbott:
This should be ready to go.

CarloLucibello (approved):

Issue #1544, opened 2021-03-23
OneHotArrays.jl?


mschauer:
Would it make sense for one-hot-arrays to live outside Flux?

DhairyaLGandhi:
I think that would be the case if we could make it more generic and better tested. I'm assuming the use cases for such a package are generally narrow though.

mschauer:
I don't think it's too narrow, just skimming https://juliahub.com/ui/CodeSearch?q=onehot&u=all&t=all

darsnack:
Yeah, I was thinking the same thing working on a personal project where I wanted one-hot arrays but not the rest of Flux. I agree with @DhairyaLGandhi that we would want some more comprehensive coverage and testing. I think the underlying array implementation is generic enough, but I'm talking about defining functions like `argmax`, etc. We may not have all the possible functions defined sensibly right now.

mcognetta:
I, too wish it lived outside flux. If I put together a roadmap for migrating it could there be a review process for it?

darsnack:
Absolutely, I will review any PRs that migrate the code. The destination repo will still need to be owned by Flux since it is integral to our codebase. I mostly haven't worked on this because of a lack of time, not a lack of interest.

ParadaCarleton:
> I think that would be the case if we could make it more generic and better tested. I'm assuming the use cases for such a package are generally narrow though.

Judging by [this](https://discourse.julialang.org/t/all-the-ways-to-do-one-hot-encoding/64807) thread there'd be a lot of interest in it outside of MLJ.

mcognetta:
Thanks for pinging this. Actually I have a PR nearly ready. It needs to be done in two stages though, and I'd need a core maintainer to help. 

How should we move forward?

PS sorry for the delay, it has been a hectic few months.

ToucheSir:
Feel free to link the repo you've been working on once you feel it's in a state to review. Once that's registered as a package, we can be far more leisurely about deciding how to integrate it into Flux.

darsnack:
Yeah happy to provide core maintainer help. Just ping me where needed.

TLipede:
I wouldn't want to step on any toes, but I'm happy to contribute to moving this to its own package if that would be helpful. I can appreciate that time can get away, and I luckily have a little bandwidth to work on this at the moment.

darsnack:
[...]

PR #2025, opened 2022-07-23
using OneHotArrays

mcabbott:
This removes `onehot.jl` since the package is now registered, https://github.com/JuliaRegistries/General/pull/64647

Tests ~~could be~~ removed too.

Maybe docs need though?

Closes https://github.com/FluxML/Flux.jl/issues/1544


mcabbott:
GPU test failure is real, and happens without Flux:
```julia
julia> using CUDA, OneHotArrays, Zygote

julia> CUDA.allowscalar(false)

julia> y = onehotbatch(ones(3), 1:2) |> cu
2×3 OneHotMatrix(::CuArray{UInt32, 1, CUDA.Mem.DeviceBuffer}) with eltype Bool:
 1  1  1
 ⋅  ⋅  ⋅

julia> gA = rand(3, 2) |> cu
3×2 CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}:
 0.0501317  0.408421
 0.446921   0.0276333
 0.805853   0.55874

julia> gradient(A -> sum(A * y), gA)[1] isa CuArray
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore are only permitted from the REPL for prototyping purposes.
If you did intend to index this array, annotate the caller with @allowscalar.
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] assertscalar(op::String)
    @ GPUArraysCore ~/.julia/packages/GPUArraysCore/rSIl2/src/GPUArraysCore.jl:78
  [3] getindex
    @ ~/.julia/packages/GPUArrays/gok9K/src/host/indexing.jl:9 [inlined]
  [4] getindex
    @ ~/.julia/packages/GPUArrays/gok9K/src/host/indexing.jl:30 [inlined]
  [5] gather!(dst::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, src::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, idx::CuArray{UInt32, 1, CUDA.Mem.DeviceBuffer})
    @ NNlib ~/.julia/packages/NNlib/uDZz6/src/gather.jl:27
  [6] gather
    @ ~/.julia/packages/NNlib/uDZz6/src/gather.jl:77 [inlined]
  [7] *(A::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, B::OneHotMatrix{UInt32, 2, CuArray{UInt32, 1, CUDA.Mem.DeviceBuffer}})
    @ OneHotArrays ~/.julia/packages/OneHotArrays/Moo4n/src/linalg.jl:10
  [8] rrule
    @ ~/.julia/packages/ChainRules/oot29/src/rulesets/Base/arraymath.jl:40 [inlined]
  [9] rrule
    @ ~/.julia/packages/ChainRulesCore/16PWJ/src/rules.jl:134 [inlined]
 [10] chain_rrule
    @ ~/.julia/packages/Zygote/IoW2g/src/compiler/chainrules.jl:217 [inlined]
 [11] macro expansion
    @ ~/.julia/packages/Zygote/IoW2g/src/compiler/interface2.jl:0 [inlined]
 [12] _pullback(::Zygote.Context, ::typeof(*), ::CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}, ::OneHotMatrix{UInt32, 2, CuArray{UInt32, 1, CUDA.Mem.DeviceBuffer}})
    @ Zygote ~/.julia/packages/Zygote/IoW2g/src/compiler/interface2.jl:9
```
Does not happen with the version of onehot within flux, why?
```
julia> using Flux: onehotbatch

julia> gradient(A -> sum(A * y), gA)[1] isa CuArray
true

(@v1.7) pkg> st Flux
      Status `~/.julia/environments/v1.7/Project.toml`
  [587475ba] Flux v0.13.4
```

mcabbott:
GPU test failure seems to be this:
```julia
julia> using CUDA, OneHotArrays, NNlibCUDA

julia> CUDA.allowscalar(false)

julia> x = [1, 3, 2];

julia> y = onehotbatch(x, 0:3)
4×3 OneHotMatrix(::Vector{UInt32}) with eltype Bool:
 ⋅  ⋅  ⋅
 1  ⋅  ⋅
 ⋅  ⋅  1
 ⋅  1  ⋅

julia> y2 = onehotbatch(x |> cu, 0:3)
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore are only permitted from the REPL for prototyping purposes.
If you did intend to index this array, annotate the caller with @allowscalar.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] assertscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/rSIl2/src/GPUArraysCore.jl:78
 [3] getindex
   @ ~/.julia/packages/GPUArrays/gok9K/src/host/indexing.jl:9 [inlined]
 [4] iterate
   @ ./abstractarray.jl:1144 [inlined]
 [5] iterate
   @ ./abstractarray.jl:1142 [inlined]
 [6] _onehotbatch(data::CuArray{Int64, 1, CUDA.Mem.DeviceBuffer}, labels::NTuple{4, Int64})
   @ OneHotArrays ~/.julia/packages/OneHotArrays/Moo4n/src/onehot.jl:87
 [7] onehotbatch(::CuArray{Int64, 1, CUDA.Mem.DeviceBuffer}, ::UnitRange{Int64})
   @ OneHotArrays ~/.julia/packages/OneHotArrays/Moo4n/src/onehot.jl:84
 [8] top-level scope
```

mcabbott:
Downstream failure for Transformers seems to be related:
```
Precompiling project...
[469](https://github.com/FluxML/Flux.jl/runs/7489381487?check_suite_focus=true#step:6:470)
  ✗ Transformers
[470](https://github.com/FluxML/Flux.jl/runs/7489381487?check_suite_focus=true#step:6:471)
  0 dependencies successfully precompiled in 13 seconds (102 already precompiled)
[471](https://github.com/FluxML/Flux.jl/runs/7489381487?check_suite_focus=true#step:6:472)
1 dependency errored. To see a full report either run `import Pkg; Pkg.precompile()` or load the package
[472](https://github.com/FluxML/Flux.jl/runs/7489381487?check_suite_focus=true#step:6:473)
     Testing Running tests...
[473](https://github.com/FluxML/Flux.jl/runs/7489381487?check_suite_focus=true#step:6:474)
WARNING: both PrimitiveOneHot and Flux export "OneHotArray"; uses of it in module Basic must be qualified
[474](https://github.com/FluxML/Flux.jl/runs/7489381487?check_suite_focus=true#step:6:475)
ERROR: LoadError: UndefVarError: OneHotArray not defined
[475](https://github.com/FluxML/Flux.jl/runs/7489381487?check_suite_focus=true#step:6:476)
```

chengchingwen:
Personally, I would prefer not having `OneHotArray` exported from Flux. And usually you won't need that either because the most-used api for onehot encoding is `onehot` and `onehotbatch`.

chengchingwen:
Just out of Curiosity, do we really need a dependency on OneHotArrays.jl directly? It seems that none of code/functions in Flux explicitly need OneHotArray. It could be a complete separate package and people need that just `using OneHotArrays` directly, though that would be breaking change.

mcabbott:
Ok. I think nothing was exported before, so for now this PR shouldn't call `@reexport`, not sure why I put that initially.

And longer term, indeed, there's no strong reason for Flux to depend on this. Maybe Flux@0.14 can simply drop it?

chengchingwen:
> And longer term, indeed, there's no strong reason for Flux to depend on this. Maybe Flux@0.14 can simply drop it?

Then do we need a deprecate warning for accessing those function from Flux in the next patch release?

mcabbott:
CI tells me that the `Embedding` layer has special methods for OneHotArrays. So it can't trivially be removed.

Possibly those methods be changed to dispatch on `AbstractMatrix{<:Bool}` & call generic `*` etc, but not this PR.

chengchingwen:
Also these:

https://github.com/FluxML/Flux.jl/blob/b8bdc2dd71857e94ece17d50fc250a2718bf8d43/src/layers/recurrent.jl#L203

I think `OneHotArray` should overload `*` and `NNlib.gather` so we can just call those functions.

Saransh-cpp:
`OneHotArrays` should also be added as a doc dependency or the API reference would not show up in `Flux`'s docs (missing docstring is a warning, hence the test won't error out). We could also remove the page completely, given that we are not re-exporting the package, but there should be some reference to the package in `Flux`'s docs. Maybe adding it to `ecosystem.md` if we decide to remove the page?

mcabbott:
Sorting out docs in another PR sounds great. 

Not so clear whether it wants to be included like NNlib / MLUtils or pushed out to ecosystem.md; maybe that depends on whether Flux@0.14 is going to load it at all. 

Then the goal of this PR is only to remove the code, so that we don't have two versions current -- e.g. https://github.com/FluxML/Flux.jl/pull/1959 happened after the package was created, which is confusing.

darsnack (approved):
I think with a rebase this should be good to go. The longer we wait, the more things will depend on the Flux-internal version (e.g. #2031).

The downstream errors appear unrelated (FastAI and Metalhead for sure). AtomicGraphNets.jl has not run CI for 2 months, but the error appears to be related to the SciML stack.

darsnack:
I just ran the AtomicGraphNets.jl test locally against the current release. They throw the same errors, so we can safely ignore those. @rkurchin you may want to look into those.

PR #2027, opened 2022-07-27
mark `stop`, `skip`, `@epochs` as deprecated

mcabbott:
Replaces #1913 with the simpler plan to remove these from Flux 0.14.

Closes #1913

### DIFF
--- a/src/optimise/train.jl
+++ b/src/optimise/train.jl
@@ -38,6 +38,9 @@ struct SkipException <: Exception end
 Call `Flux.skip()` in a callback to indicate when a callback condition is met.
 This will trigger the train loop to skip the current data point and not update with the calculated gradient.
 
+!!! note
+    `Flux.skip()` will be removed from Flux 0.14
+
 # Examples
 ```julia
 cb = function ()
@@ -60,6 +63,9 @@ struct StopException <: Exception end
 
 Call `Flux.stop()` in a callback to indicate when a callback condition is met.
 This will trigger the train loop to stop and exit.
+
+!!! note
+    `Flux.stop()` will be removed from Flux 0.14. It should be replaced with `break` in an ordinary `for` loop.
 
 # Examples
 ```julia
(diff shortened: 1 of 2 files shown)


Saransh-cpp:
Should the occurrences of `stop`, `skip`, and `@epochs` be removed from the docs, or should this wait till they are removed (`v0.14`)? The doctests are failing due to the deprecation warnings. Should the doctests be removed, or should they be updated with the warnings?

mcabbott:
That's a good question. I think that:
* For the dep-warns to be of any use, I guess this must be merged into a 0.13 release.  
* The removal of both the code and the docs should be in 0.14, thus a different PR
* Possibly the last 0.13 docs should have a warning note added too? That could be this PR.
* To keep tests passing, maybe this PR should also remove some `jldoctest` annotations.

mcabbott:
@Saransh-cpp maybe you know, are the built docs previewed somewhere? Is this `!!! note` thing going to work?

Saransh-cpp:
@mcabbott the previews are only generated if a PR is created from a branch belonging to the root `Flux` repository (restrictions from `Documenter.jl`). If the PR is from a branch belonging to the root repository, adding the `documentation` tag to the PR will trigger the `DocBot` to comment the generated URL.

Saransh-cpp:
`!!! note` works! Tried this locally!

mcabbott:
Ready to merge?

src/optimise/train.jl
```diff
@@ -46,6 +49,9 @@ end
 ```
 """
 function skip()
+  Base.depwarn("""Flux.skip() will be removed from Flux 0.14.
+                  It never worked as described,
```
ToucheSir, line 53:
```suggestion
```

src/optimise/train.jl
```diff
@@ -150,6 +164,8 @@ hello
 ```
 """
 macro epochs(n, ex)
+  Base.depwarn("""The macro `@epochs` will be removed from Flux 0.14.
+                  Just write an ordinary for loop.""", Symbol("@epochs"), force=true)
```
ToucheSir, line 168:
```suggestion
                  As an alternative, you can write a simple `for i in 1:epochs` loop.""", Symbol("@epochs"), force=true)
```

ToucheSir (commented):
A bit of wordsmithing, but otherwise good to go

ToucheSir (approved):

PR #2028, opened 2022-07-27
RFC: add a supertype to layers

### LINKED ISSUE #2044: Macro to display model struct the way Flux does

mcabbott:
This proposes to gives Flux's layer types various supertypes.

One reason to like this is that it simplifies the use of `show`. If you have the same supertype as `Chain`, you will be unfolded at top level like it is. No mystery functions to overload. Closes https://github.com/FluxML/Flux.jl/pull/1932, closes #2044

Edit: [568af9b](https://github.com/FluxML/Flux.jl/pull/2028/commits/568af9bafe045a933e92ccd17c1f7da2e86a4f41) goes further: We can define `functor` for this abstract type, eliminating the need to call `@functor`. (It's a pretty mysterious macro, and you can't `@macroexpand` it to have a look.) We can also define `trainable` for some abstract types; maybe that's also less mysterious.

Another is this: `Flux.gpu` and `CUDA.cu` both recursively move things, the latter via Adapt not Functors. Which means `cu` does not preserve tied weights. But if we can rely on the the shared arrays both living within a struct whose type we own (like Chain) then we can convert `cu` to something Functors-based at that boundary. (It's enough for one outer layer to do this -- using weird Dense-like layers marked only with `@functor` within a Chain is fine.)

Note that this supertype is entirely optional. The PR does not change the fact that `functor` is how Flux walks models, etc, and so it does not restrict how you can customise things. It only aims to make it easy to opt-in to the standard behaviours, without a zoo of weird macros.

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -103,7 +103,6 @@ function Dropout(p; dims=:, rng = rng_from_array())
   Dropout(p, dims, nothing, rng)
 end
 
-@functor Dropout
 function (a::Dropout)(x)
   _isactive(a) || return x
   return dropout(a.rng, x, a.p; dims=a.dims, active=true)
@@ -156,7 +155,6 @@ end
 AlphaDropout(p, active) = AlphaDropout(p, active, rng_from_array())
 AlphaDropout(p; rng = rng_from_array()) = AlphaDropout(p, nothing, rng)
 
-@functor AlphaDropout
 function (a::AlphaDropout)(x::AbstractArray{T}) where T
   _isactive(a) || return x
   p = a.p
@@ -219,8 +217,6 @@ function LayerNorm(size::Tuple{Vararg{Int}}, λ=identity; affine::Bool=true, ϵ:
 end
 LayerNorm(size::Integer...; kw...) = LayerNorm(Int.(size); kw...)
 LayerNorm(size_act...; kw...) = LayerNorm(Int.(size_act[1:end-1]), size_act[end]; kw...)
-
-@functor LayerNorm
 
 (a::LayerNorm)(x) = a.diag(normalise(x, dims=1:length(a.size), ϵ=a.ϵ))
 
@@ -348,8 +344,6 @@ function BatchNorm(chs::Int, λ=identity;
             nothing, chs)
 end
 
-@functor BatchNorm
-
 function (BN::BatchNorm)(x)
   @assert size(x, ndims(x)-1) == BN.chs
   N = ndims(x)
@@ -437,8 +431,6 @@ function InstanceNorm(chs::Int, λ=identity;
             nothing, chs)
 end
 
-@functor InstanceNorm
-
 function (l::InstanceNorm)(x)
   @assert ndims(x) > 2
   @assert size(x, ndims(x)-1) == l.chs
@@ -514,8 +506,6 @@ mutable struct GroupNorm{F,V,N,W} <: PartialTrainLayer{(:β, :γ)}
   active::Union{Bool, Nothing}
   chs::Int # number of channels
 end
-
-@functor GroupNorm
 
 function GroupNorm(chs::Int, G::Int, λ=identity;
               initβ=zeros32, initγ=ones32,
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -90,7 +90,7 @@ julia> isapprox(count(==(0), y) / length(y), 0.5, atol=0.1)
 true
 ```
 """
-mutable struct Dropout{F,D,R<:AbstractRNG}
+mutable struct Dropout{F,D,R<:AbstractRNG} <: NoTrainLayer
   p::F
   dims::D
   active::Union{Bool, Nothing}
@@ -104,8 +104,6 @@ function Dropout(p; dims=:, rng = rng_from_array())
 end
 
 @functor Dropout
-trainable(a::Dropout) = (;)
-
 function (a::Dropout)(x)
   _isactive(a) || return x
   return dropout(a.rng, x, a.p; dims=a.dims, active=true)
@@ -146,7 +144,7 @@ julia> isapprox(std(x), std(y), atol=0.2)
 true
 ```
 """
-mutable struct AlphaDropout{F,R<:AbstractRNG}
+mutable struct AlphaDropout{F,R<:AbstractRNG} <: NoTrainLayer
   p::F
   active::Union{Bool, Nothing}
   rng::R
@@ -159,8 +157,6 @@ AlphaDropout(p, active) = AlphaDropout(p, active, rng_from_array())
 AlphaDropout(p; rng = rng_from_array()) = AlphaDropout(p, nothing, rng)
 
 @functor AlphaDropout
-trainable(a::AlphaDropout) = (;)
-
 function (a::AlphaDropout)(x::AbstractArray{T}) where T
   _isactive(a) || return x
   p = a.p
@@ -209,7 +205,7 @@ julia> isapprox(std(y, dims=1:3), ones(1, 1, 1, 2), atol=0.1) && std(y, dims=1:3
 true
 ```
 """
-struct LayerNorm{F,D,T,N} <: SimpleLayer
+struct LayerNorm{F,D,T,N} <: PartialTrainLayer{(:diag,)}
   λ::F
   diag::D
   ϵ::T
@@ -322,7 +318,7 @@ julia> isapprox(std(m(xs)), 1, atol=0.1) && std(xs) != std(m(xs))
 true
 ```
 """
-mutable struct BatchNorm{F,V,N,W} <: SimpleLayer
+mutable struct BatchNorm{F,V,N,W} <: PartialTrainLayer{(:β, :γ)}
   λ::F  # activation function
   β::V  # bias
   γ::V  # scale
@@ -353,7 +349,6 @@ function BatchNorm(chs::Int, λ=identity;
 end
 
 @functor BatchNorm
-trainable(bn::BatchNorm) = hasaffine(bn) ? (β = bn.β, γ = bn.γ) : (;)
 
 function (BN::BatchNorm)(x)
   @assert size(x, ndims(x)-1) == BN.chs
@@ -412,7 +407,7 @@ julia> isapprox(std(y, dims=1:2), ones(1, 1, 3, 2), atol=0.2) && std(y, dims=1:2
 true
 ```
 """
-mutable struct InstanceNorm{F,V,N,W} <: SimpleLayer
+mutable struct InstanceNorm{F,V,N,W} <: PartialTrainLayer{(:β, :γ)}
   λ::F  # activation function
   β::V  # bias
   γ::V  # scale
@@ -443,7 +438,6 @@ function InstanceNorm(chs::Int, λ=identity;
 end
 
 @functor InstanceNorm
-trainable(in::InstanceNorm) = hasaffine(in) ? (β = in.β, γ = in.γ) : (;)
 
 function (l::InstanceNorm)(x)
   @assert ndims(x) > 2
@@ -506,7 +500,7 @@ julia> isapprox(std(y[:, :, 3:4, 2]), 1, atol=0.1) && std(xs[:, :, 3:4, 2]) != s
 true
 ```
 """
-mutable struct GroupNorm{F,V,N,W} <: SimpleLayer
+mutable struct GroupNorm{F,V,N,W} <: PartialTrainLayer{(:β, :γ)}
   G::Int  # number of groups
   λ::F  # activation function
   β::V  # bias
@@ -522,7 +516,6 @@ mutable struct GroupNorm{F,V,N,W} <: SimpleLayer
 end
 
 @functor GroupNorm
-trainable(gn::GroupNorm) = hasaffine(gn) ? (β = gn.β, γ = gn.γ) : (;)
 
 function GroupNorm(chs::Int, G::Int, λ=identity;
               initβ=zeros32, initγ=ones32,
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -209,7 +209,7 @@ julia> isapprox(std(y, dims=1:3), ones(1, 1, 1, 2), atol=0.1) && std(y, dims=1:3
 true
 ```
 """
-struct LayerNorm{F,D,T,N}
+struct LayerNorm{F,D,T,N} <: SimpleLayer
   λ::F
   diag::D
   ϵ::T
@@ -322,7 +322,7 @@ julia> isapprox(std(m(xs)), 1, atol=0.1) && std(xs) != std(m(xs))
 true
 ```
 """
-mutable struct BatchNorm{F,V,N,W}
+mutable struct BatchNorm{F,V,N,W} <: SimpleLayer
   λ::F  # activation function
   β::V  # bias
   γ::V  # scale
@@ -412,7 +412,7 @@ julia> isapprox(std(y, dims=1:2), ones(1, 1, 3, 2), atol=0.2) && std(y, dims=1:2
 true
 ```
 """
-mutable struct InstanceNorm{F,V,N,W}
+mutable struct InstanceNorm{F,V,N,W} <: SimpleLayer
   λ::F  # activation function
   β::V  # bias
   γ::V  # scale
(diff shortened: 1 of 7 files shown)


src/layers/types.jl
```diff
@@ -0,0 +1,99 @@
...
+abstract type AbstractLayer end
+
+if VERSION > v"1.9-"
+  function Functors.functor(::Type{T}, x) where {T<:AbstractLayer}
+    namedtuple(x), Splat(paramerterlesstype(T))  # avoids warnings: `splat(x)` is deprecated, use `Splat(x)` instead.
+  end
+else
+  function Functors.functor(::Type{T}, x) where {T<:AbstractLayer}
+    namedtuple(x), Base.splat(paramerterlesstype(T))
+  end
```
mcabbott, line 34: This method for Flux.functor is slightly slower than the ones written by `@functor`. Perhaps `@generated` can make it faster. Although my guess is that doing anything nontrivial with `fmap` will make this wash out, e.g. 28.454 μs -> 29.071 μs here:
```julia
julia> @btime Flux.functor($(BatchNorm(3)))
  min 402.290 ns, mean 890.385 ns (5 allocations, 224 bytes)  # before
  min 1.333 μs, mean 1.371 μs (6 allocations, 256 bytes)      # after
((λ = identity, β = Float32[0.0, 0.0, 0.0], γ = Float32[1.0, 1.0, 1.0], μ = Float32[0.0, 0.0, 0.0], σ² = Float32[1.0, 1.0, 1.0], ϵ = 1.0f-5, momentum = 0.1f0, affine = true, track_stats = true, active = nothing, chs = 3), Splat(BatchNorm))

julia> @btime Flux.f32($(Chain(Chain(BatchNorm(3), Dense(3=>3)), Chain(BatchNorm(3), Dense(3=>3)))));
  min 15.250 μs, mean 28.454 μs (123 allocations, 5.72 KiB)  # before
  min 27.416 μs, mean 29.071 μs (138 allocations, 5.75 KiB)  # after
```

ToucheSir: I think it would make sense to use ConstructionBase here. That way types with non-standard reconstruction needs (e.g. needing to preserve certain type params) are more likely to be supported.

mcabbott: My idea was that the supertype should match the standard use of `@functor T`.

If you have anything nonstandard, you can call the macro or overload `functor` yourself, exactly as before this PR. Introducing yet another package seems more confusing to me, how do you know which to overload in what circumstances?

More narrowly, the performance issue appears to be this:
```
julia> @btime @inferred Flux.paramerterlesstype($(Vector{Int}))
  min 1.796 μs, mean 1.911 μs (12 allocations, 640 bytes)
Array
```
Similar code at e.g. https://github.com/SciML/MultiScaleArrays.jl/blob/f82cbc35bdbd06d8a0ee40bb3c9d42a9d09607c4/src/shape_construction.jl#L11 and at https://github.com/cstjean/QuickTypes.jl/blob/54871cf2fd7f77441b74fcaea6310c2c7b38e248/src/QuickTypes.jl#L27  has the same overhead.

mcabbott: Performance fixed in ba61a9a

src/layers/types.jl
```diff
@@ -0,0 +1,91 @@
...
+function Functors.functor(::Type{T}, x) where {T<:AbstractLayer}
+  if @generated
+    F = fieldnames(T)
+    args = map(sy -> :(getfield(x, $(QuoteNode(sy)))), F)
+    C = Base.typename(T).name  # constructor
+    recon = VERSION > v"1.9-" ? :(Splat($C)) : :(Base.splat($C))
+    :((NamedTuple{$F}(($(args...),)), $recon))
+  else
+    # Getting this parameterless type takes about 2μs, every time:
+    namedtuple(x), Base.splat(Base.typename(T).wrapper)
```
ericphanson, line 36: I guess one option here is to use `ConstructionBase.constructorof` so that this can be customized in problematic cases (if there are any with layers... not really sure since the discussion in https://github.com/JuliaLang/julia/pull/46213 is kinda hard to follow). But I guess we can't do that in the generated path, and we don't want diverging behavior.

ericphanson (commented):
I like this (after arguing for macro in https://github.com/FluxML/Flux.jl/issues/2044 😄). One comment is it might be helpful to put the subtype declarations in the docstring for each function, e.g.
```julia
"""
    Chain(layers...) <: ContainerLayer
```
and to document the abstract types in https://github.com/FluxML/Flux.jl/blob/master/docs/src/models/layers.md.

src/layers/types.jl
mcabbott: My argument against ConstructionBase is this:

* At present the path for "easy" structs is `@functor MyStruct`, which makes assumptions about the default constructor. 

* If you have a weird struct, you must write a method for `functor` which understands it. But this is discouraged.

With the supertype, you should automatically get the same behaviour as `@functor MyStruct`. 

But Functors is still handling recursion and re-building, and you can still supply a more specific method for `functor`. It is in this case that a the supertype method using ConstructionBase would provide an alternative way to customise behaviour. But now you have to know about two different recurse-and-rebuild libraries, instead of one. And because we own the supertype, it's fair to assume that you are writing this for Flux, it isn't some large pre-existing thing which already has ConstructionBase methods written... I mean that's fine, you can still use Flux, but not via the supertype.

(I would actually probably favour ditching Functors, and simply demanding that all Flux models use structs with default constructors. Unlike Zygote it has no ambition to apply to arbitrary code. But that's for another day, this PR adds a friendly path without changing how other paths work.)

src/layers/types.jl
```diff
@@ -0,0 +1,91 @@
...
+only to shield types which would otherwise be trainable, such as arrays of floats.
+
+Also (like any `AbstractLayer`) removes the need for `@functor`,
+and (like `SimpleLayer`) tells `show` not to unfold further.
+"""
+abstract type PartialTrainLayer{which} <: SimpleLayer end
+
+function Optimisers.trainable(layer::PartialTrainLayer{which}) where {which}
+  NamedTuple{which}(map(sy -> getfield(layer, sy), which))
+end
```
mcabbott, line 84: This type is the odd one out, I'm not entirely sure we should have it.

One, if we wish to have other traits which identify some subset of fields, the way `trainable` does, then if they are independent they cannot share the same type hierarchy, and so are a bit second-class. Maybe that's ok, trainability is pretty important. 

Two, it doesn't cover all existing cases of `trainable`, e.g. we want `mutable struct Recur{T,S} <: ContainerLayer` for printing, but also have `trainable(a::Recur) = (; cell = a.cell)`.

src/layers/recurrent.jl
```diff
@@ -189,7 +188,7 @@ end
 
 # Vanilla RNN
 
-struct RNNCell{F,A,V,S}
+struct RNNCell{F,A,V,S} <: SimpleLayer  # or should it be PartialTrainLayer{(:Wi, :Wh, :b)}?
   σ::F
   Wi::A
```
mcabbott, line 193: Here I'm also not sure. Should `state0` be trainable?

src/layers/normalise.jl
```diff
@@ -352,9 +344,6 @@ function BatchNorm(chs::Int, λ=identity;
             nothing, chs)
 end
 
-@functor BatchNorm
-trainable(bn::BatchNorm) = hasaffine(bn) ? (β = bn.β, γ = bn.γ) : (;)
```
mcabbott, line 356: Here `PartialTrainLayer{(:β, :γ)}` fixes which fields are trainable, permanently. That's OK, since `β = nothing` when it's not trainable, so it'll be ignored. It improves type-stability although I doubt this matters at all.

mcabbott (commented):
Yes to putting the description of these into the docs. 

I'm not sure whether they should be in layer docstrings. IMO these should aim at people calling the layer, for which it doesn't matter at all. Whereas to write a new one you should first do `@less Dense(1 => 1)`. 

Or look at the right bit of the manual... I think we had some examples of defining new structs, and probably those should also grow to mention supertypes.

src/layers/types.jl
```diff
@@ -0,0 +1,91 @@
...
+    namedtuple(x), Base.splat(Base.typename(T).wrapper)
+  end
+end
+
+function namedtuple(x::T) where T
+  F = fieldnames(T)
+  NamedTuple{F}(map(sy -> getfield(x, sy), F))
+end
+
+Adapt.adapt_structure(to, layer::AbstractLayer) = fmap(x -> adapt(to, x), layer)
```
mcabbott, line 45: Can be just this
```suggestion
Adapt.adapt_structure(to, layer::AbstractLayer) = fmap(adapt(to), layer)
```
but check the version bound on Adapt

src/layers/types.jl
ToucheSir: My argument is _for_ ConstructionBase, but not in Flux. Rather, Functors should use it instead of [rolling its own thing](https://github.com/FluxML/Functors.jl/blob/master/src/functor.jl#L19). The fallback behaviour will be identical (call `T`), but it's more likely a library will need to depend on ConstructionBase anyhow for compat with Accessors etc. instead of on Functors (which is why we see piracy).

That or nix Functors as Michael said, but then the argument is even stronger IMO. You can still get away with a default constructor, but if you want something fancy this is but one game left in town.

darsnack:
Since this has now been put on the next milestone, I should chime in. I personally do not like having a super-type. My reasons:
- It precludes someone from using their own type hierarchy. There are real examples of this: [InvertibleNetworks.jl](https://github.com/slimgroup/InvertibleNetworks.jl) and [Mill.jl](https://github.com/CTUAvastLab/Mill.jl). Or any code where we write functions of the form `f: model -> model'` will reasonably want to dispatch on the model type. Depending on what exactly `f` and `model` are, using a type hierarchy could be very useful here.
- It's a sharp change from Flux's approach to this problem in the past. If we were getting something of important value here, then it might be worth it. But it seems like this is one implementation that is subjectively more convenient than a macro.
- One positive for this approach is that it is more natural to Julia users than a macro. I agree in general with this for types, but I don't think it holds as much weight in our case. Any user wanting to implement a custom layer is going to have to look at the documentation...these types aren't so intuitive that they should naturally come to mind. I don't see how learning that you need to `@functor MyLayer` is so much worse at this point.
- I agree that `@functor` is weird and confusing. I think Flux ought to provide something more friendly like `@layer` which does the Functor stuff but also `show` stuff and more.
- The current type hierarchy—`PartialTrainLayer{which} <: SimpleLayer <: AbstractLayer`—is already getting messy. Julia's type system just isn't good for hanging all this information on, IMO (see the never-ending traits discussions). What happens for a layer which `<: ContainerLayer` but also has trainable parameters of its own?
- A super-type makes adopting the defaults easy, but it doesn't avoid needing to know and override `trainable`, `show`, etc. The non-default cases are still just as annoying. I'd rather work on a solution that makes those cases better. Hiding the default case is not so hard of a problem.
- I don't see us ever moving away from something Functors-like. We can nix Functors, but we will still have the need to walk and recurse nested structs. Having multiple levels/ways of opting into being a Flux layer seems confusing to me.

It seems like the main motivation for this PR originally (though it has grown) is to make `_big_show` easier to opt-into. Once we adopt a type hierarchy, it will be hard to change our minds. We should think carefully about what it should be and what functionality should hang on top of it. If default `show` is what's at stake here, then a more minimal change might be easier for right now. Two options:
- Define something like `@layer` which does `@functor` + `show`. This seems like it has no downsides. We mask the confusing `@functor`, and we can always introduce types on top of it later. If we want to remove it in favor of types, it seems easier to deprecate.
- Introduce just `SimpleLayer` and `ContainerLayer` for the purpose of `show` only. Leave `trainable`, etc. as is for now.

mcabbott:
I agree the present state of this PR is a sort-of "maximial" look at what's possible, to see what clashes result, and [as noted](https://github.com/FluxML/Flux.jl/pull/2028#discussion_r951464456) I'm not sure that including `trainable` is a great idea. The "minimal" one would be just `SimpleLayer` and `ContainerLayer`.

Figuring out whether things clash with packages is one goal here. It looks at first glance that InvertibleNetworks.jl has a [type hierarchy](https://github.com/slimgroup/InvertibleNetworks.jl/blob/a94c9487e697b872fe446917795467ae330dc73e/src/utils/neuralnet.jl#L4-L13) which could subtype Flux.AbstractLayer, but perhaps no finer, not sure, anyone know more details? Mill.jl I'm less sure -- anyone know? Note that both packages mentioned do load Flux.

I also agree that some kind of `Flux.@layer` macro is an alternative option. Maybe a smaller change from `@functor` / `@treelike`. (Maybe someone can make a PR to try it out and see what problems arise; `@functor` is a weird macro with `eval` but perhaps it need not be.) To be useful for `show` it would somehow need to take options to make the Chain-like / Dense-like distinction, default to Dense-like perhaps. Maybe it ought to take options to control `trainable` too.

Unlike a supertype, the new macro could be *obligatory*. The idea of translating `cu` to an `fmap` walk at the outermost layer would then be easy. 

Macros are always a bit opaque. You could argue that FluxML has probably gone a little too far in avoiding types, e.g. `train!` took 4 arguments of `::Any`, good luck remembering... but now `methods(Flux.train!)` makes it easier not to mess up. Although it's certainly possible to go too far the other way.

(Don't read too much into the milestone, just trying to separate future ideas from bugfixes a bit.)

ericphanson:
I don’t think a macro that defines a show method should be obligatory because then you will get method overwrites if you define your own show method (which is one nice thing about the abstract type- it’s a fallback then).

darsnack:
Good point. The messy thing about macros will be the order in which stuff is called. Something like
```julia
struct MyLayer
    foo
end

Base.show(..., ::MyLayer) = ...

@layer MyLayer
```
seems like a simple error to run into. The current `@functor` only defines `functor` which is likely not to run into this issue.

I think with something like https://github.com/FluxML/Functors.jl/issues/41, we remove `trainable` etc. completely. Then we are left with `show` which can be opt-in:
```
@layer MyLayer prettyprint=:simple
```
I was going to prototype the Functors issue anyways, so I can also test out the `@layer` stuff.

Another option is to make `@layer` apply on the struct definition itself. This will ensure that any methods that are defined later like a custom `show` will happen after the default.

Alternatively, introducing the hierarchy with just `SimpleLayer` and `ContainerLayer` only for non-parameter related stuff like `show` seems reasonable to me. At least there will be a clean demarcation between Functors stuff and other stuff.

mcabbott:
Yes the macro needs at least a way to choose Chain-like vs. Dense-like printing. It it not only methods for `show` that do this, but possibly one option should apply no methods to `show`. I don't think we should ever rely on order of overwriting methods.

Macros which act on structs seem a step more opaque. And also more complicated to write, e.g. how should it interact with `@kwdef`?

src/layers/show.jl
```diff
@@ -55,16 +54,11 @@ _show_children(m::Maxout) = m.layers
 _show_children(p::Parallel) = (p.connection, p.layers...)
 _show_children(f::PairwiseFusion) = (f.connection, f.layers...)
```
mcabbott, line 55: These methods are a further complication to any attempt to make a `show` easier to opt into. If you claim (through a macro or a type) that your layer is a Chain-like container, do you imply that you want any Tuple within it to be splatted, i.e. that you have a constructor like `Parallel(f, layers...) = Parallel(f, layers)`? Or is that still an obscure function you ought to overload?

mcabbott: BTW the `@layer` PR automates this: 
https://github.com/FluxML/Flux.jl/pull/1932/files#diff-08261c6ab5a9b75d3f9bca6a6f93b32fefe4e22e91ffcfdfd8d88dad08c3a498R61-R68

PR #2031, opened 2022-08-02
Add `EmbeddingBag`

mcognetta:
Add `EmbeddingBag`, a slight generalization of `Embedding` which allows for embedding multiple items at once and performing a reduction on them. See: [PyTorch's implementation](https://pytorch.org/docs/stable/generated/torch.nn.EmbeddingBag.html).

This PR implements PyTorch's `input`/`offset` embedding as well as scalar, vector, vector of vector, matrix, and `OneHotVector`/`OneHotMatrix` input types.

`EmbeddingBag` is an outstanding feature in #1431.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md


mcognetta:
There are two things to improve:

1) add support for `padding_idx` (and in `Embedding`)
2) reduce allocations (some of the methods don't operate entirely inplace on a workspace matrix)

src/layers/basic.jl
```diff
@@ -692,3 +692,84 @@ end
 function Base.show(io::IO, m::Embedding)
   print(io, "Embedding(", size(m.weight, 2), " => ", size(m.weight, 1), ")")
 end
+
+"""
+    EmbeddingBag(in => out, reduction=Statistics.mean; init=randn)
```
CarloLucibello, line 697:
```suggestion
    EmbeddingBag(in => out, reduction=mean; init=randn)
```

test/layers/basic.jl
```diff
@@ -311,6 +311,63 @@ import Flux: activations
...
+      @test_throws BoundsError emb_bag([1,2,3,4,5,6], [0,12])
+
+      # docstring example
+      @test emb_bag([1,2,3,4,5,6,7,8,9,10], [0,4,5,7]) ≈ emb_bag([[1,2,3,4], [5], [6,7], [8,9,10]])
+
+      # multiple bags (input as a vector of vectors)
+      x = [rand(1:vocab_size, 3) for _ in 1:4]
+      y = emb_bag(x)
+      z = reduce(hcat, reduction.(emb.(x), dims=2))
+      @test y isa Matrix{Float32}
```
CarloLucibello, line 346: should a vector of vecs input correspond to a vector of vecs output instead

mcognetta: Oh, interesting point. I will defer to you + other maintainers.

mcognetta: I can see it both ways. But to me the most compelling argument is for keeping it as a matrix: you may want to do vector of vectors to get a matrix when you don't have the same number of classes per bag (since we don't have `padding_idx` there is no way to do this).

mcognetta: Oh whoops. There is a way to do that ofc... The PyTorch input/offset style.

mcognetta: How about the following: make vector of vecs return a vector of vectors and add a one liner in the docs about converting a vector of vectors to pytorch format.

Would that be acceptable?

src/layers/basic.jl
```diff
@@ -692,3 +692,84 @@ end
...
+function (m::EmbeddingBag)(inputs::AbstractVector, offsets::AbstractVector)
+    offsets[1] == 0 || throw(ArgumentError("`offsets` must begin with 0."))
+    out = zeros(eltype(m.weight), size(m.weight, 1), length(offsets))
+    start = firstindex(inputs)
+    for i in eachindex(offsets[1:end-1])
+        out[:, i] = m(inputs[start:offsets[i+1]])
+        start = offsets[i+1]+1
+    end
+    out[:, end] = m(inputs[offsets[end]+1:end])
+    out
```
ToucheSir, line 757: Is the plan to add rrules to make this differentiable? I see just the PyTorch CPU-side implementation is [quite involved](https://github.com/pytorch/pytorch/blob/4300f64ad000c95aa0ae0e1a6985c2fe08203a1a/aten/src/ATen/native/EmbeddingBag.cpp), but perhaps most of that complexity is already handled by Julia.

mcognetta: I do not have a good intuition for when we need custom `rrules`, since many layers don't. Is it the assignment operator?


The PyTorch implementation was confusing to me, it looks like a lot of platform specific code. The main stuff that adds complexity is the presence of the sparse and padding index parameters, which are not supported here.

DhairyaLGandhi:
Avoiding the mutation would be ideal.

src/layers/basic.jl
```diff
@@ -692,3 +692,84 @@ end
...
+end
```
darsnack, line 758: I think this should be differentiable out of the box and avoid the need for `rrule`s.

```suggestion
function (m::EmbeddingBag)(inputs::AbstractVector, offsets::AbstractVector)
    offsets[1] == 0 || throw(ArgumentError("`offsets` must begin with 0."))
    start = firstindex(inputs)
    slices = [view(inputs, start:offsets[i + 1])
              for i in eachindex(offsets[1:end - 1])]
    
    return m(slices)
end
```

src/layers/basic.jl
```diff
@@ -692,3 +692,84 @@ end
...
+(m::EmbeddingBag)(bags::AbstractMatrix) = reduce(hcat, m.(eachcol(bags)))
+
+function (m::EmbeddingBag)(x::OneHotVector{T,L}) where {T,L}
+  size(m.weight, 2) == L || throw(DimensionMismatch("Matrix column must correspond with OneHot size: $(size(m.weight, 2)) != $L"))
+  return m(onecold(x))
+end
+function (m::EmbeddingBag)(x::OneHotMatrix{T,L}) where {T,L}
+  size(m.weight, 2) == L || throw(DimensionMismatch("Matrix column must correspond with OneHot size: $(size(m.weight, 2)) != $L"))
+  return m(LinearAlgebra.Transpose(onecold(x)))
+end
```
darsnack, line 771: I think these will work. It will avoid you duplicating checks, and it should automatically use the fast paths.

```suggestion
(m::EmbeddingBag)(x::OneHotVector) = m.weight * x
(m::EmbeddingBag)(x::OneHotMatrix) = m.reduction(m.weight * x); dims = 2)
```

darsnack (changes requested):
A couple suggestions that you'll want to verify.

src/layers/basic.jl
mcognetta: Thanks. Did something similar to this.

src/layers/basic.jl
mcognetta: Done.

mcognetta:
I made a few updates following the suggestions presented above. In particular, I have changed the `inputs`/`offsets` method to be more Julian. I originally wanted it to be identical to the PyTorch version, but that was 0-indexed. I changed it to be 1 indexed and updated the docs to more clearly explain how that input should be used. I also updated the implementation so it is non-mutating.

It seems there is one last comment that is unresolved: https://github.com/FluxML/Flux.jl/pull/2031#discussion_r935736267

> should a vector of vecs input correspond to a vector of vecs output instead

I am open to suggestions on this from the maintainers. To me, the best solution is:

> make vector of vecs return a vector of vectors and add a one liner in the docs about converting a vector of vectors to pytorch format.

Thanks for the helpful reviews so far, hopefully we can get this merged soon!

src/layers/basic.jl
```diff
@@ -692,3 +692,74 @@ end
 function Base.show(io::IO, m::Embedding)
   print(io, "Embedding(", size(m.weight, 2), " => ", size(m.weight, 1), ")")
 end
+
+"""
+    EmbeddingBag(in => out, reduction=mean; init=randn)
```
darsnack, line 697:
```suggestion
    EmbeddingBag(in => out, reduction=mean; init=Flux.randn32)
```

src/layers/basic.jl
```diff
@@ -692,3 +692,74 @@ end
...
+
+A lookup table that stores embeddings of dimension `out` for a vocabulary of size 
+`in`. Similar to [`Embedding`](@ref) but can take multiple inputs in a "bag". The
+embeddings of these are then reduced to a single embedding based on `reduction`.
+Typically, `reduction` is `mean`, `sum`, or `maximum`.
```
darsnack, line 702:
```suggestion
A lookup table that stores embeddings of dimension `out` for a vocabulary of size 
`in`. Similar to [`Embedding`](@ref) but can take multiple inputs in a "bag,"
then reduced all outputs to a single embedding based on `reduction`.
Typically, `reduction` is `mean`, `sum`, or `maximum`. 
```

src/layers/basic.jl
```diff
@@ -692,3 +692,74 @@ end
...
+`in`. Similar to [`Embedding`](@ref) but can take multiple inputs in a "bag". The
+embeddings of these are then reduced to a single embedding based on `reduction`.
+Typically, `reduction` is `mean`, `sum`, or `maximum`. 
+
+This layer is often used to store word embeddings and retrieve them using indices. 
+The inputs can take several forms:
+  - A scalar := single bag with a single item
+  - A vector := single bag with multiple items
+  - A matrix := multiple bags with multiple items (each column is a bag)
+  - A vector of vectors: multiple bags with multiple items (each vector is a bag)
```
darsnack, line 709:
```suggestion
  - A vector of vectors: multiple bags with multiple items (each inner vector is a bag)
```

darsnack (changes requested):
This mostly looks good with a couple of minor docstring changes.

The main outstanding question I have is the utility of input/offset style input. What's the value of specifying inputs in this way? It seems very confusing in comparison to a vector of vectors. Is there an upstream layer that will specify an input to `EmbeddingBag` in this way? If not, can we just do away with this option?

As far as the vector of vectors issue, I would go with the output that is most natural for the downstream layer that receives the output of an `EmbeddingBag`.

mcognetta:
Thanks for the comments. I'll review them tomorrow.

> The main outstanding question I have is the utility of input/offset style input. What's the value of specifying inputs in this way? It seems very confusing in comparison to a vector of vectors. Is there an upstream layer that will specify an input to `EmbeddingBag` in this way? If not, can we just do away with this option?

I included it specifically for feature parity with Pytorch. I agree that it is cumbersome compared to the vector of vectors input, but I think it has utility in that you aren't messing with essentially ragged tensors. And it's quite easy to build them sequentially.

 I think it should be kept just for completeness.


> As far as the vector of vectors issue, I would go with the output that is most natural for the downstream layer that receives the output of an `EmbeddingBag`.

I will keep it as is then (so that it returns a matrix, not a vector of vectors).

darsnack:
> I included it specifically for feature parity with Pytorch. I agree that it is cumbersome compared to the vector of vectors input, but I think it has utility in that you aren't messing with essentially ragged tensors. And it's quite easy to build them sequentially.
>
> I think it should be kept just for completeness.

Sure, this is okay. I do think the examples in the docstring need to be expanded then to demonstrate the difference between all the input types (especially showing the input/offset case). Do we know where this format is preferred? It might also be good to mention when each style is useful in the docstring.

src/layers/basic.jl
```diff
@@ -692,3 +692,74 @@ end
...
+    start = firstindex(inputs)
+    newoffsets = vcat(offsets, [lastindex(inputs)])
+    slices = [inputs[offsets[i]:(i+1 > lastindex(offsets) ? end : offsets[i+1]-1)] for i in eachindex(offsets)]
+
+    return m(slices)
+end
+(m::EmbeddingBag)(idx::Integer) = m.weight[:, idx]
+(m::EmbeddingBag)(bag::AbstractVector) = vec(m.reduction(NNlib.gather(m.weight, bag), dims=2))
+(m::EmbeddingBag)(bags::AbstractVector{<:AbstractVector}) = reduce(hcat, m.(bags))
+(m::EmbeddingBag)(bags::AbstractMatrix) = reduce(hcat, m.(eachcol(bags)))
```
darsnack, line 758: After reading the PyTorch docstring, it seems the main advantage of this layer is memory efficiency. So, shouldn't these be `mapreduce` instead of a broadcast to achieve the same feature?

mcognetta: Unfortunately, `mapreduce(f, hcat, collection)` is not optimized. But yes, I agree. I will add a todo for when specialized `mapreduce` functions are added. See: https://discourse.julialang.org/t/different-performance-between-reduce-map-and-mapreduce/85149 and https://github.com/JuliaLang/julia/issues/31137.

```julia
julia> (m::EmbeddingBag)(bags::AbstractVector{<:AbstractVector}) = reduce(hcat, m.(bags))
julia> (m::EmbeddingBag)(bags::AbstractMatrix) = reduce(hcat, m.(eachcol(bags)))

julia> test(m::EmbeddingBag, bags::AbstractVector{<:AbstractVector})  = mapreduce(m, hcat, bags)
julia> test(m::EmbeddingBag, bags::AbstractMatrix) = mapreduce(m, hcat, eachcol(bags))
julia> e = Flux.EmbeddingBag(100=>64)
julia> bags = [[rand(1:100) for _ in 1:3] for _ in 1:1000]
julia> @btime e(bags);
  709.630 μs (14004 allocations: 2.16 MiB)

julia> @btime test(e, bags);
  14.700 ms (15935 allocations: 124.18 MiB)
```

mcognetta:
> > I included it specifically for feature parity with Pytorch. I agree that it is cumbersome compared to the vector of vectors input, but I think it has utility in that you aren't messing with essentially ragged tensors. And it's quite easy to build them sequentially.
> > I think it should be kept just for completeness.
> 
> Sure, this is okay. I do think the examples in the docstring need to be expanded then to demonstrate the difference between all the input types (especially showing the input/offset case). Do we know where this format is preferred? It might also be good to mention when each style is useful in the docstring.

I've expanded the documentation and added notes on the input/offset input type.

src/layers/basic.jl
```diff
@@ -692,3 +692,126 @@ end
...
+
+julia> model = Flux.EmbeddingBag(vocab_size => embed_size)
+EmbeddingBag(1000 => 4)  # 4_000 parameters
+
+julia> bags = [[1, 200, 25, 789], [2, 5, 10, 999]];
+
+julia> bags_mtx = [1 2; 200 5; 25 10; 789 999];
+
+julia> model(bags) |> summary
+"4×2 Matrix{Float32}"
```
mcabbott, line 733: I think this example should not use 4 with two meanings. How about `26 => 10`, embedding letters in 10 dims?

src/layers/basic.jl
```diff
@@ -692,3 +692,126 @@ end
...
+
+@functor EmbeddingBag
+
+EmbeddingBag((in, out)::Pair{<:Integer, <:Integer}, reduction::Function = mean; init = randn32) = EmbeddingBag(init(out, in), reduction)
+EmbeddingBag(weight) = EmbeddingBag(weight, mean)
+
+function (m::EmbeddingBag)(inputs::AbstractVector, offsets::AbstractVector)
+    offsets[firstindex(offsets)] == 1 || throw(ArgumentError("`offsets` must begin with 1."))
+    start = firstindex(inputs)
+    newoffsets = vcat(offsets, [lastindex(inputs)])
```
mcabbott, line 799: This only accepts `AbstractVector{<:Integer}`, and doesn't seem to use `start`.
```suggestion
function (m::EmbeddingBag)(inputs::AbstractVector{<:Integer}, offsets::AbstractVector{<:Integer})
    first(offsets) == 1 || throw(ArgumentError("`offsets` must begin with 1."))
    newoffsets = vcat(offsets, lastindex(inputs))
```
It seems to be allowing for OffsetArrays, but nothing else in Flux will work with these. Can it be simplified much if you put `LinearAlgebra.require_one_based_indexing(inputs, offsets)` somewhere?

src/layers/basic.jl
```diff
@@ -692,3 +692,126 @@ end
...
+embeddings of these are then reduced to a single embedding based on `reduction`.
+Typically, `reduction` is `mean`, `sum`, or `maximum`. 
+
+This layer is often used to store word embeddings and retrieve them using indices. 
+The inputs can take several forms:
+  - A scalar := single bag with a single item
+  - A vector := single bag with multiple items
+  - A matrix := multiple bags with multiple items (each column is a bag)
+  - A vector of vectors := multiple bags with multiple items (each vector is a bag)
+  - An input vector and offset vector := Explained below.
```
mcabbott, line 710: Maybe clearer to show what "input" means. And since the 2-argument form gets a long description, maybe it does not need to be a bullet point too:
```suggestion

After constructing `eb = EmbeddingBag(26 => 10, mean)`, the layer accepts the following inputs as `eb(x)`:
  - `x::Int` is a single bag with a single item. `mean` is not called, and this is identical to an `Embedding` layer with the same weights. Returns a vector. 
  - `x::Vector` is single bag with multiple items. `mean` is called once. Returns a vector.
  - `x::Matrix` and `xs::Vector{<:Vector} = eachcol(x)` are identical, both multiple bags with multiple items. `mean` is called once per column. Returns a matrix.

The layer may also be called with two vectors, `eb(xs, off)`. This is interpreted as `length(off)-1` bags, the first of which is `xs[1:off[2]-1]`, then `xs[off[2]:off[3]-1]`, and so on. These need not have the same length. Returns a matrix.
```
This 2-argument method is really just a variant of the `Vector{<:Vector}` one, with a utility function to turn offsets into slices. Maybe that function should have a name, and a docstring which describes it more clearly (as there are quite a few edge case).

(Avoid "offset vector" as that already has a meaning. And I'm not sure the first of two inputs should be called "input".)

src/layers/basic.jl
```diff
@@ -692,3 +692,126 @@ end
 function Base.show(io::IO, m::Embedding)
   print(io, "Embedding(", size(m.weight, 2), " => ", size(m.weight, 1), ")")
 end
+
+"""
+    EmbeddingBag(in => out, reduction=mean; init=randn)
+
+A lookup table that stores embeddings of dimension `out` for a vocabulary of size 
+`in`. Similar to [`Embedding`](@ref) but can take multiple inputs in a "bag". The
```
mcabbott, line 700: How fixed is this "bag" terminology? It seems awful... can it be "MeanEmbedding" or "ReducedEmbedding" or "EmbeddingGroups" or something?

src/layers/basic.jl
```diff
@@ -692,3 +692,126 @@ end
...
+
+julia> scalar_bag = 5 # just a single bag of one item
+5
+
+julia> model(scalar_bag);
+
+julia> single_bag = [1, 2, 2, 4]; # one bag several items
+
+julia> model(single_bag) |> summary
+"8-element Vector{Float32}"
```
mcabbott, line 753: There are 2 example blocks, maybe not both needed? 

And maybe this can be compacted a bit, some things don't need to be given variable names:
```suggestion
julia> model = Flux.EmbeddingBag(26 => 10, mean)  # vocab size 26, output 10 dims
EmbeddingBag(26 => 10)  # 260 parameters

julia> model(5) |> summary  # one item
"10-element Vector{Float32}"

julia> model([1, 2, 2, 4]) |> summary  # one bag
"10-element Vector{Float32}"

julia> model([1, 2, 4]) ≈ mean(model.([1, 2, 3]))
```

src/layers/basic.jl
mcabbott: The really big memory cost is going to be the gradient of `gather`. For every column / vector, `∇gather_src` is going to allocate like a copy of the weights. 

https://github.com/FluxML/NNlib.jl/blob/6f74fad0a2a24e3594fc5229cc515fa25e80f877/src/gather.jl#L80

One could write a more efficient combined rule for this. Or add some thunks to the one in NNlib & wait for AD to learn to exploit them.

src/layers/basic.jl
mcognetta: Personally, I like the terminology (since it is like an embedding multiset) and it is reasonably well established ([PyTorch](https://pytorch.org/docs/stable/generated/torch.nn.EmbeddingBag.html), [Tensorflow](https://www.tensorflow.org/addons/api_docs/python/tfa/layers/EmbeddingBag)). Of course, I am not tied to it though.

src/layers/basic.jl
mcognetta: I kept one block specifically for the `input`/`offsets` example, since that is complex and one to demonstrate each of the input types. I will make it more concise though.

mcabbott: Yes, that (if it stays) could happily be a second block below.

src/layers/basic.jl
```diff
@@ -802,7 +802,7 @@ function (m::EmbeddingBag)(inputs::AbstractVector, offsets::AbstractVector)
     return m(slices)
 end
 (m::EmbeddingBag)(idx::Integer) = m.weight[:, idx]
-(m::EmbeddingBag)(bag::AbstractVector) = vec(m.reduction(NNlib.gather(m.weight, bag), dims=2))
+(m::EmbeddingBag)(bag::AbstractVector{<:Integer}) = vec(m.reduction(NNlib.gather(m.weight, bag), dims=2))
```
mcabbott, line 805: One implementation thought. Maybe this ought to be something like `NNlib.gather(mean, m.weight, bag)` to fuse the indexing and the accumulation.

This could be added later as an optimisation, but allowing arbitrary reductions instead of just `mean` may make this harder. Are functions other than `mean` useful? If not, then restricting now may make upgrading later easier.

mcognetta: People often use `sum`, but I kept `mean` as the default since that is PyTorch's default (I think you can see a trend here haha).

mcabbott: I guess the optimisation I suggest could easily dispatch to slow-and-simple versions (as before) for all but a few functions, mean/sum. So perhaps no need to limit it now.

src/layers/basic.jl
```diff
@@ -692,3 +692,84 @@ end
...
+        out[:, i] = m(inputs[start:offsets[i+1]])
+        start = offsets[i+1]+1
+    end
+    out[:, end] = m(inputs[offsets[end]+1:end])
+    out
+end
+(m::EmbeddingBag)(idx::Integer) = m.weight[:, idx]
+(m::EmbeddingBag)(bag::AbstractVector) = vec(m.reduction(NNlib.gather(m.weight, bag), dims=2))
+(m::EmbeddingBag)(bags::AbstractVector{<:AbstractVector}) = reduce(hcat, m.(bags))
+(m::EmbeddingBag)(bags::AbstractMatrix) = reduce(hcat, m.(eachcol(bags)))
```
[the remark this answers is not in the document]
mcognetta: Done.

src/layers/basic.jl
mcognetta: This can be done after this PR, right?

mcabbott: Yes. I just mean these concerns will dwarf the `hcat` cost. (Even on the forward pass, the thing you make to call `mean` on it will also be much larger.)

src/layers/basic.jl
mcognetta: Done.

I split this into a `_splitat` function with a docstring and some tests and renamed `input`/`offset` to `data`/`offsets`.

src/layers/basic.jl
mcognetta: I removed `start` which was unused. I don't think it can be simplified much, with that check, as `first`/`lastindex` etc will do basically what we would otherwise do by hand.

mcognetta:
Is there anything else necessary for this PR? Maybe some improvements with mapreduce can be done, but are not possible right now.

src/layers/basic.jl
```diff
@@ -692,3 +692,107 @@ end
...
+
+julia> model(data, offsets) |> summary
+"8×3 Matrix{Float32}"
+
+julia> model(Flux.OneHotVector(2, vocab_size)) |> summary # single bag with one item
+"8-element Vector{Float32}"
+
+julia> model(Flux.OneHotMatrix([2, 3, 5, 7], vocab_size)) |> summary # 4 bags, each with one item
+"8×4 Matrix{Float32}"
+```
```
mcabbott, line 770: An attempt to write this in a way which introduces complexity gradually...

Examples are such that you can see the averaging.

I don't think the split locations should be called `offset`s in 1-based land, since the first is 1 but not offset by 1 from the start. I also think it's much too confusing to have an example in the text before you see the basic idea.

````suggestion
A lookup table that stores embeddings of dimension `out` for a vocabulary of size `in`.
Differs from [`Embedding`](@ref) in that, instead of acting on a single vocabulary index,
it always acts a vector of indices which it call a "bag".
Their individual embedding vectors are reduced to one, using `mean` or some other function.

Instead of acting on one "bag", such as `x::Vector{Int}`, the layer can also act on several:

* Acting on a vector of "bags", it produces a matrix whose columns are the reduced vectors.
  More generally on `x::Array{Vector{Int}}`, its output is of size `(out, size(x)...)`.

* Any higher-rank array of integers is interpreted as a collection of "bags" each along the first dimension.
  Thus the output is `mapslices(e, x; dims=1)` when `e::EmbeddingBag` and `x::Array{Int,N}`.
  This is more efficient, but requires that all "bags" have the same length.

* A vector of "bags" may also be produced by splitting a vector of indices at specified points.
  For this case the layer takes two inputs, both vectors of integers. See details below.

The "bag" may equivalently be represented as a `OneHotMatrix`. A collection of these,
or one higher-rank `OneHotArray`, again produce a stack of embeddings. See details below.

# Examples
```
julia> vocab_size = 26;  # embed into 3 dimensions, with non-random vectors:

julia> eb = EmbeddingBag(vocab_size => 3, init=Flux.identity_init(gain=100))
EmbeddingBag(26 => 3)  # 78 parameters

julia> eb([2])  # one bag of 1 item
3-element Vector{Float32}:
   0.0
 100.0
   0.0

julia> eb([3,3,1])  # one bag of 3 items, one mean embedding
3-element Vector{Float32}:
 33.333332
  0.0
 66.666664

julia> eb([[3,1,3], [2,1]])  # two bags
3×2 Matrix{Float32}:
 33.3333  50.0
  0.0     50.0
 66.6667   0.0

julia> eb([1 1 1 1; 1 2 3 4])  # 4 bags each of 2 items, eachcol([1 1 1 1; 1 2 3 4])
3×4 Matrix{Float32}:
 100.0  50.0  50.0  50.0
   0.0  50.0   0.0   0.0
   0.0   0.0  50.0   0.0

julia> eb(rand(1:26, 10, 5, 5)) |> size  # 25 bags each of 10 items
(3, 5, 5)
```

Another way to specify "many bags of many items" is to provide a vector `data` (each in `1:in`)
and a vector `at` stating where to split that up into "bags".
The first bag starts with `data[at[1]]`, the second at `data[at[2]]`, and so on, 
with no overlaps and nothing left out (thus it requires `at[1]==1`).

```julia
julia> data = [11, 1, 12, 2, 13, 3, 14];

julia> Flux._splitat(data, [1, 4]) |> println  # internal function, makes data[1:3], data[4:end]
[[11, 1, 12], [2, 13, 3, 14]]

julia> eb(data, [1, 4])  # two bags, of 3 and 4 items
3×2 Matrix{Float32}:
 33.3333   0.0
  0.0     25.0
  0.0     25.0
```

Finally, each bag may also be also be represented as a [`OneHotMatrix`](@ref OneHotArrays.onehotbatch).

```julia
julia> eb(onehotbatch("bba", 'a':'z'))  # same as [2,2,1], one bag of 3 items
3-element Vector{Float32}:
 33.333332
 66.666664
  0.0

julia> eb([onehotbatch("bba", 'a':'z'), onehotbatch("cc", 'a':'z')])  # two bags
3×2 Matrix{Float32}:
 33.3333    0.0
 66.6667    0.0
  0.0     100.0
```
````

**Review by mcabbott — CHANGES_REQUESTED** [review]
I had a look over this. I am still dismayed by the need for 5 distinct bullet points to explain what this thing does, which suggests to me that something is not quite right. Worse it's really 7 because the PR's 5 do not describe its behaviour on onehot arrays.

`Embedding` has *one* rule, it returns an array `(out, size(x)...)`, and the same if you do `onehotbatch(x)` first. It always wants integer indices (and `size(::Int) = ()` for just one). What's the corresponding simple rule?

I think `EmbeddingBag` should always take vectors of integers, called "bags". Given any collection of such vectors `vs`, its output is an array `(out, size(vs)...)`. A single vector is a trivial collection, `size(vs) == ()`. An array of integers `x` is always sliced, `vs = eachslice(x, dims=Tuple(2:ndims(x)))`, hence `(out, size(x)[2:end]...)` follows. 

Then `EmbeddingBag(in=>out)(3)` should be an error, I think, otherwise it breaks the pattern & that's confusing. You can use Embedding. (Maybe there should be constructors like `Embedding(::EmbeddingBag)`?)

For onehot arrays, surely a OneHotMatrix is a bag. Then I think it must stand in for a vector of integers in all circumstances: Just one, a vector of OneHotMatrix, and an N-dim OneHotArray.

Despite saying "always sliced" above, the actual implementation should *not* slice if it doesn't have to. What I put in the suggestion should work for any `x::Array{Int,N}` and is much faster than the PR.

File: test/layers/basic.jl [hunk 18]
```diff
@@ -311,6 +311,86 @@ import Flux: activations
...
+
+    for reduction in [sum, Statistics.mean, maximum]
+      vocab_size, embed_size = 10, 4
+      emb_bag = Flux.EmbeddingBag(vocab_size => embed_size, reduction)
+      emb = Flux.Embedding(emb_bag.weight)
+      @test size(emb_bag.weight) == (embed_size, vocab_size)
+
+      # scalar bag
+      @test emb_bag(2) ≈ emb_bag.weight[:,2]
+      @test emb_bag(3) ≈ emb(3)
```
mcabbott, line 347:
```suggestion
      @test_throws ErrorException emb_bag(2)
```

test/layers/basic.jl
```diff
@@ -311,6 +311,86 @@ import Flux: activations
...
+    offsets_too_large = [1, 5, 11]
+
+    @test_throws ArgumentError  Flux._splitat(data, offsets_non_monotonic)
+    @test_throws ArgumentError  Flux._splitat(data, offsets_non_sorted)
+    @test_throws ArgumentError  Flux._splitat(data, offsets_non_one)
+    @test_throws ArgumentError  Flux._splitat(data, offsets_too_large)
+
+    for reduction in [sum, Statistics.mean, maximum]
+      vocab_size, embed_size = 10, 4
+      emb_bag = Flux.EmbeddingBag(vocab_size => embed_size, reduction)
```
mcabbott, line 341:
```suggestion
    @testset for reduction in [sum, Statistics.mean, maximum]
      vocab_size, embed_size = 10, 4
      emb_bag = EmbeddingBag(vocab_size => embed_size, reduction)
```

test/layers/basic.jl
```diff
@@ -311,6 +311,86 @@ import Flux: activations
...
+      y = emb_bag(x)
+      z = reduce(hcat, reduction.(emb.(xvec), dims=2))
+      @test y ≈ emb_bag(xvec)
+      @test y ≈ z
+
+      # one hot bags. should be identical to Embedding, since the bags
+      # are of size 1.
+      @test emb_bag(Flux.OneHotVector(3, vocab_size)) ≈ emb_bag.weight[:,3]
+      @test emb_bag(Flux.OneHotVector(4, vocab_size)) ≈ emb(Flux.OneHotVector(4, vocab_size))
+      @test_throws DimensionMismatch emb_bag(Flux.OneHotVector(3, 1000))
```
mcabbott, line 384:
```suggestion
      # a one-hot matrix is a bag, but a one-hot vector is not.
      @test_throws ErrorException emb_bag(Flux.OneHotVector(3, vocab_size))
```

test/layers/basic.jl
```diff
@@ -311,6 +311,86 @@ import Flux: activations
...
+
+      x2 = Flux.OneHotMatrix(rand(1:vocab_size, 3), vocab_size)
+      y2 = emb_bag(x2)
+      z2 = emb(x2)
+      @test y2 isa Matrix{Float32}
+      @test y2 ≈ z2
+      @test_throws DimensionMismatch emb_bag(Flux.OneHotMatrix(1:5, 1000))
```
mcabbott, line 391:
```suggestion
      i2 = rand(1:vocab_size, 3)
      x2 = Flux.OneHotMatrix(i2, vocab_size)
      y2 = emb_bag(x2)
      z2 = emb(i2)
      @test y2 isa Matrix{Float32}
      @test y2 ≈ mean(z2, dims=2)
      @test_throws DimensionMismatch emb_bag(Flux.OneHotMatrix(1:5, 1000))
```

mcabbott (commented):
I think these are the test changes needed to match the above.

mcognetta:
Thanks for the comments, I'll check it soon and hopefully we can move forward.

test/layers/basic.jl
```diff
@@ -311,6 +311,81 @@ import Flux: activations
...
+    offsets_too_large = [1, 5, 11]
+
+    @test_throws ArgumentError  Flux._splitat(data, offsets_non_monotonic)
+    @test_throws ArgumentError  Flux._splitat(data, offsets_non_sorted)
+    @test_throws ArgumentError  Flux._splitat(data, offsets_non_one)
+    @test_throws ArgumentError  Flux._splitat(data, offsets_too_large)
+
+    @testset for reduction in [sum, Statistics.mean, maximum]
+      vocab_size, embed_size = 10, 4
+      emb_bag = EmbeddingBag(vocab_size => embed_size, reduction)
```
mcabbott, line 341: CI says this isn't exported, I didn't notice...
```suggestion
      emb_bag = Flux.EmbeddingBag(vocab_size => embed_size, reduction)
```

test/layers/basic.jl
```diff
@@ -311,6 +311,81 @@ import Flux: activations
...
+
+      # a one-hot matrix is a bag, but a one-hot vector is not.
+      @test_throws ErrorException emb_bag(Flux.OneHotVector(3, vocab_size))
+
+      i2 = rand(1:vocab_size, 3)
+      x2 = Flux.OneHotMatrix(i2, vocab_size)
+      y2 = emb_bag(x2)
+      z2 = emb(i2)
+      @test y2 isa Matrix{Float32}
+      @test y2 ≈ mean(z2, dims=2)
```
mcabbott, line 385:
```suggestion
      @test y2 isa Vector{Float32}
      @test y2 ≈ vec(mean(z2, dims=2))
```

test/layers/basic.jl
```diff
@@ -311,6 +311,81 @@ import Flux: activations
...
+
+      # a one-hot matrix is a bag, but a one-hot vector is not.
+      @test_throws ErrorException emb_bag(Flux.OneHotVector(3, vocab_size))
+
+      i2 = rand(1:vocab_size, 3)
+      x2 = Flux.OneHotMatrix(i2, vocab_size)
+      y2 = emb_bag(x2)
+      z2 = emb(i2)
+      @test y2 isa Vector{Float32}
+      @test y2 ≈ vec(mean(z2, dims=2))
```
mcabbott, line 385: One more typo...
```suggestion
      @test y2 ≈ vec(reduction(z2, dims=2))
```

src/layers/basic.jl
```diff
@@ -692,3 +692,108 @@ end
...
+EmbeddingBag(weight) = EmbeddingBag(weight, mean)
+
+function (m::EmbeddingBag)(data::AbstractVector, offsets::AbstractVector)
+  return m(_splitat(data, offsets))
+end
+(m::EmbeddingBag)(inds::AbstractArray{<:Integer}) = dropdims(m.reduction(Embedding(m.weight)(inds), dims=2), dims=2)
+(m::EmbeddingBag)(ind::Integer) = error("EmbeddingBag expects an array of indices, not just one")
+
+(m::EmbeddingBag)(hot::AbstractArray{Bool}) = dropdims(m.reduction(Embedding(m.weight)(hot), dims=2), dims=2)
+(m::EmbeddingBag)(hot::AbstractVector{Bool}) = error("EmbeddingBag not defined for a one-hot vector")
```
mcognetta, line 789: This seems to be too general of a type restriction. For example, I could define a `MultiHot <: AbstractVector{Bool}`, that succinctly  encodes a bag with fixed `k` elements (in fact, this was one of my original use cases for `EmbeddingBags`), and then if index `i` is true, it should be included in the bag.

src/layers/basic.jl
```diff
@@ -692,3 +692,108 @@ end
...
+  weight::W
+  reduction::F
+end
+
+@functor EmbeddingBag
+
+EmbeddingBag((in, out)::Pair{<:Integer, <:Integer}, reduction::Function = mean; init = randn32) = EmbeddingBag(init(out, in), reduction)
+EmbeddingBag(weight) = EmbeddingBag(weight, mean)
+
+function (m::EmbeddingBag)(data::AbstractVector, offsets::AbstractVector)
```
mcognetta, line 782: minor nit: this can be reduced to just `(m::EmbeddingBag)(data::AbstractVector, offsets::AbstractVector) = m(_splitat(data, offsets))` to save lines and fit the style of the other functions.

I will make this change later if there are no objections.

mcognetta (commented):
Left a few comments and will make the changes based on responses.

CarloLucibello:
What is the status of this PR?

mcabbott:
Status from my side is that I thought I understood exactly how it ought to work, and perhaps rudely pushed hard to get it done before I forgot... and now I've mostly forgotten :(  But should perhaps revive!

mcognetta:
Yes I lost the thread also. I'm happy to wrap it up this weekend, if the proposed changes are what we want.

src/layers/basic.jl
```diff
@@ -716,3 +716,143 @@ Embedding((in, out)::Pair{<:Integer, <:Integer}; init = randn32) = Embedding(ini
...
+
+"""
+    _splitat(data::AbstractVector, offsets::AbstractVector{Int})
+
+Splits a vector of data into a vector of vectors based on offsets. Each offset
+specifies the next sub-vectors starting index in the `data` vector. In otherwords,
+the `data` vector is chuncked into vectors from `offsets[1]` to `offsets[2]` (not including the element at `offsets[2]`), `offsets[2]` to `offsets[3]`, etc.
+The last offset specifies a bag that contains everything to the right of it.
+
+The `offsets` vector must begin with `1` and be monotonically increasing. The last element of `offsets` must be at most `length(data)`.
```
mcabbott, line 729: Shorter and clearer?
````suggestion
    _splitat(data::AbstractVector, at::AbstractVector{Int})

Partitions `data` into a vector of `length(at)` vectors.

Each index `i in at` specifies that an inner vector starts with `data[i]`.
These indices must be strictly increasing, and start at `1`.
The last inner vector ends with `data[end]`.

### Example
```jldoctest
julia> Flux._splitat('A':'Z', [1, 3])
2-element Vector{StepRange{Char, Int64}}:
 'A':1:'B'
 'C':1:'Z'
```
````

src/layers/basic.jl
mcabbott: This is a possible encoding. Dispatch on such a type specifically is not forbidden by this method.

So far, I think every other use of one-hot arrays behaves identically if you `collect` it. This is why I think it makes sense to define these methods for `AbstractArray{Bool}`. Another boolean type with a different meaning cannot also have this property that `collect` doesn't change the result.

CarloLucibello (approved):

PR #2034, opened 2022-08-06
fix array constructor rrule

### LINKED ISSUE #2033: Wrong rrule dispatch for Array constructor
This would dispatch on all call to parametric type because `typeof(Array)` is `UnionAll`.

https://github.com/FluxML/Flux.jl/blob/c4837f74a767f9ed9a0919626792460b93ee7995/src/functor.jl#L118

It should be `::Core.Typeof(Array)` or `::Type{Array}`.

chengchingwen:
fix #2033


chengchingwen:
I'm just fixing that issue, but the line seems to be type piracy?

DhairyaLGandhi (approved):
Type piracy - maybe not with this fix? We extend it as an `rrule`. If CUDA gets a ChainRulesCore dep then we can move it there i suppose.

chengchingwen:
Yeah, I personally think that rule should be defined in CUDA, but that's a separate issue.

PR #2035, opened 2022-08-08
Fix the last remaining 404 errors

Saransh-cpp:
The only thing keeping me from making the cross reference check strict -
```julia
strict = [:cross_references]
```
is the failing `isnumeric` hyperlink (https://fluxml.ai/Flux.jl/stable/training/optimisers/#Optimisers.destructure), which does not have a docstring in `Optimisers.jl`.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -4,6 +4,7 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 MLUtils = "f1d291b0-491e-4a28-83b9-f70985020b54"
 NNlib = "872c559c-99b0-510c-b3b7-b6c96a88d5cd"
+OneHotArrays = "0b1bfda6-eb8a-41d2-88d8-f5af5cad476f"
 Optimisers = "3bd65402-5787-11e9-1adc-39752487f4e2"
 
 [compat]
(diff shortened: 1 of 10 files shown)


ToucheSir:
Nice! @mcabbott should we slap a docstring onto `isnumeric` or scrub the dangling reference to it in https://github.com/FluxML/Optimisers.jl/blob/9b75dbb0d80e4e383eab279345a70d4b567721ea/src/destructure.jl#L8?

mcabbott:
Could go either way. I made https://github.com/FluxML/Optimisers.jl/pull/104 to see how it looks.

Saransh-cpp:
I will fix the reference here once https://github.com/FluxML/Optimisers.jl/pull/104 is merged!

mcabbott (approved):

Saransh-cpp:
Temporarily fixed the references for the API of `OneHotArrays.jl`. See #2038.

ToucheSir:
If you wouldn't mind looking into the doctest failure, we can get this merged.

Saransh-cpp:
- The regex pattern is getting out of hand 😅 but now it captures `Float32` values too (which were causing the CI to fail).
- `Documenter.jl` was outdated and was not allowing `strict` to be of the type `Vector{Symbol}`.

I think a new release for `Optimisers.jl` is required for the tests to pass. The CI is still not picking up the docstring for `Optimisers.isnumeric`.

ToucheSir:
> I think a new release for `Optimisers.jl` is required for the tests to pass. The CI is still not picking up the docstring for `Optimisers.isnumeric`.

Ok, I was wondering why it worked given that wasn't tagged! Have another try after https://github.com/JuliaRegistries/General/pull/66169.

Saransh-cpp:
> Have another try after https://github.com/JuliaRegistries/General/pull/66169.

Everything works now!

ToucheSir (approved):
Great!

PR #2041, opened 2022-08-18
Add ability to filter `loadmodel!` recursion

darsnack:
This adds a keyword predicate function `recursable` to `loadmodel!`. The leaves of the source and destination models are filtered like `filter(recursable, ldsts)` before recursing. This allows the user to specify leaves that should get skipped.

A common use-case for this feature is when you are trying to load weights to/from a model with some optional non-trainable layer. For example, ResNets might have `Dropout` layers. We want to load the saved weights from these models into an equivalently structured ResNet modulo `Dropout`. This happens in Metalhead.jl.

### PR Checklist

- [x] Tests are added
- [x] ~~Entry in NEWS.md~~

### DIFF
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -81,7 +81,7 @@ however, attempting to copy a non-zero array to an inactive parameter will throw
 Likewise, copying a `src` value of `false` to any `dst` array is valid,
 but copying a `src` value of `true` will error.
 """
-function loadmodel!(dst, src; filter = Returns(true), cache = Base.IdSet())
+function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
   ldsts = _filter_children(filter, functor(dst)[1])
   lsrcs = _filter_children(filter, functor(src)[1])
   (keys(ldsts) == keys(lsrcs)) ||
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -33,7 +33,7 @@ _filter_children(f, children::NamedTuple) =
 _filter_children(f, children) = filter(f, children)
 
 """
-    loadmodel!(dst, src; filter = _ -> true)
+    loadmodel!(dst, src)
 
 Copy all the parameters (trainable and non-trainable) from `src` into `dst`.
 
@@ -42,9 +42,6 @@ and calling `copyto!` on parameter arrays or throwing an error when there is a m
 Non-array elements (such as activation functions) are not copied and need not match.
 Zero bias vectors and `bias=false` are considered equivalent
 (see extended help for more details).
-
-Specify the predicate function `filter` to control what is recursed.
-A child node `x` in either `dst` and `src` is skipped when `filter(x) == false`.
 
 # Examples
 ```julia
@@ -66,19 +63,6 @@ false
 
 julia> iszero(dst[2].bias)
 true
-
-julia> src = Chain(Dense(5 => 2), Dropout(0.2), Dense(2 => 1))
-Chain(
-  Dense(5 => 2),                        # 12 parameters
-  Dropout(0.2),
-  Dense(2 => 1),                        # 3 parameters
-)                   # Total: 4 arrays, 15 parameters, 348 bytes.
-
-julia> Flux.loadmodel!(dst, src; filter = x -> !(x isa Dropout)) # skips loading Dropout
-Chain(
-  Dense(5 => 2, tanh),                  # 12 parameters
-  Dense(2 => 1),                        # 3 parameters
-)                   # Total: 4 arrays, 15 parameters, 316 bytes.
 ```
 
 # Extended help
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -33,7 +33,7 @@ _filter_children(f, children::NamedTuple) =
 _filter_children(f, children) = filter(f, children)
 
 """
-    loadmodel!(dst, src)
+    loadmodel!(dst, src; filter = _ -> true)
 
 Copy all the parameters (trainable and non-trainable) from `src` into `dst`.
 
@@ -43,9 +43,12 @@ Non-array elements (such as activation functions) are not copied and need not ma
 Zero bias vectors and `bias=false` are considered equivalent
 (see extended help for more details).
 
+Specify the predicate function `filter` to control what is recursed.
+A child node `x` in either `dst` and `src` is skipped when `filter(x) == false`.
+
 # Examples
 ```julia
-julia> dst = Chain(Dense(Flux.ones32(2, 5, tanh)), Dense(2 => 1; bias = [1f0]))
+julia> dst = Chain(Dense(Flux.ones32(2, 5), Flux.ones32(2), tanh), Dense(2 => 1; bias = [1f0]))
 Chain(
   Dense(5 => 2, tanh),                  # 12 parameters
   Dense(2 => 1),                        # 3 parameters
@@ -63,6 +66,19 @@ false
 
 julia> iszero(dst[2].bias)
 true
+
+julia> src = Chain(Dense(5 => 2), Dropout(0.2), Dense(2 => 1))
+Chain(
+  Dense(5 => 2),                        # 12 parameters
+  Dropout(0.2),
+  Dense(2 => 1),                        # 3 parameters
+)                   # Total: 4 arrays, 15 parameters, 348 bytes.
+
+julia> Flux.loadmodel!(dst, src; filter = x -> !(x isa Dropout)) # skips loading Dropout
+Chain(
+  Dense(5 => 2, tanh),                  # 12 parameters
+  Dense(2 => 1),                        # 3 parameters
+)                   # Total: 4 arrays, 15 parameters, 316 bytes.
 ```
 
 # Extended help
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -28,6 +28,10 @@ _tie_check(dst, src) = true
 
 _bool_tie_check(dst, src) = true
 
+_filter_children(f, children::NamedTuple) =
+  NamedTuple(filter(kv -> f(kv[2]), pairs(children)))
+_filter_children(f, children) = filter(f, children)
+
 """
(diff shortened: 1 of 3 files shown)


darsnack:
@theabhirath will this help with the loading issues in Metalhead? It would be used like:
```julia
loadmodel!(dst, src; recursable = x -> !(x isa Dropout))
```

theabhirath:
I just tried it out for loading from a `ResNet` with `Dropout` into one without, and I can confirm this works! This will help enormously for these use cases when we have to load models for transfer learning

ToucheSir (commented):
Seems reasonable to me outside of some name bikeshedding. How about `{expand|traverse}_child`?

mcabbott:
Re the bikeshed, maybe`loadmodel!(dst, src; ignore = x -> (x isa Dropout))`? Assuming that you would often ignore only a few things, since you must explore all the outer containers etc. Or `skip`, `omit`... can't be `exclude` as `fmap` uses that differently.

darsnack:
I like `filter` best because it hints at the implementation. I agree that you are likely to ignore only a few things, so if we prefer the logic to be specified that way, then `ignore` seems best.

src/loading.jl
```diff
@@ -77,9 +81,9 @@ however, attempting to copy a non-zero array to an inactive parameter will throw
 Likewise, copying a `src` value of `false` to any `dst` array is valid,
 but copying a `src` value of `true` will error.
 """
-function loadmodel!(dst, src; cache = Base.IdSet())
-  ldsts, _ = functor(dst)
-  lsrcs, _ = functor(src)
+function loadmodel!(dst, src; filter = Returns(true), cache = Base.IdSet())
+  ldsts = _filter_children(filter, functor(dst)[1])
+  lsrcs = _filter_children(filter, functor(src)[1])
```
mcabbott, line 86: Right now it works by field name only, and thus allows you to load from "plain" types:
```julia
julia> model = Chain(Dense(2 => 1, tanh), Dense(1 => 1, bias=false));

julia> data = fmapstructure(x -> x isa AbstractArray ? zero(x) : nothing, model)
(layers = ((weight = Float32[0.0 0.0], bias = Float32[0.0], σ = nothing), (weight = Float32[0.0;;], bias = nothing, σ = nothing)),)

julia> Flux.loadmodel!(model, data);

julia> model[1].weight
1×2 Matrix{Float32}:
 0.0  0.0
```
What does filtering do here? 

One answer would be to apply the test only to `dst`, which is assumed to be a Flux model with useful types. 

Another would be to apply it to both, and take `||`.

darsnack: The source could contain types that we want to skip too. I don't quite see the issue with your example, since my intention for `filter` was to be very permissive, i.e. you can filter for a specific object id (`x -> x !=== y`). In other words, whoever passes in `filter` should write the predicate knowing what `model` and `data` are.

mcabbott: At present a model and its "plain" version are equivalent for `loadmodel!`, and no matter what the types, the fields must *always* match. I guess my initial take is that this property ought to be preserved. You may skip the loading of some parts, but may not perform surgery on the trees. 

I think the present state makes them inequivalent, as after filtering, the fields of dst will no longer match, but the same mask is not applied to src. 

Now I see you have a test where you skip Dropout, and the source & dst trees don't match. This seems to me like it'll be a juggling act to make something with just the right branches missing. (They are numbered fields of a Chain/Tuple, not named fields.)

darsnack: I agree, it is a complicated juggling act, and that was the reason I was hesitant to even add the feature. But the use case described and in the tests is exactly why I created the PR, and it is a reasonable/common case.

I don't think there is any great way to make juggling the trees convenient and straightforward. The implementation in the PR was what I thought would keep the code of `loadmodel!` simple and the intuition behind the mechanism still easy to understand.

mcabbott: One thought is this. Since no parameters are ever loaded from `Dropout`, perhaps it should always be ignored. 

That is, when the recursion encounters a Tuple or a Vector (no field names) then it always filters out all structs which contain no parameters, and demands that the rest match. 

Would this agree with the present behaviour in all cases? If so, it could be a non-breaking upgrade to make the case you want just work. 

What should happen then with structs whose fields *do* have names, it the next question. But possibly they can also just be ignored. (Clearly "contain no parameters" means all their children too.)

ToucheSir: Given that `isleaf` is literally implemented as `children(x) == ()`, is there any risk of false positives from this approach?

darsnack: Ultimately, what you're describing is still filtering the children before recursing but with a fixed predicate. So, I still feel that the person calling `loadmodel!` must understand what the filtered tree looks like, and there is still juggling going on (only we're responsible for it now).

Here's a scenario I just thought about: one model has `BatchNorm` with no affine transformation (no parameters) and the other does not have `BatchNorm`. Maybe we want to skip `BatchNorm` in this case. On the other hand, two models with `BatchNorm` but no affine transformation would want to load state like tracked mean/variances from one model to the other. How do you differentiate these cases for skipping `BatchNorm`?

My thought is that loading is a rather tricky thing to get right, so it is better not to do too much guessing and be permissive about what is loadable.

darsnack: > Given that isleaf is literally implemented as children(x) == (), is there any risk of false positives from this approach?

I think, yes, but you could specially ignore this case for filtering.

mcabbott: Something like `(x = 1.0,)` has children but, in my scheme, would be filtered out of a Tuple as it contains no parameters to load. Likewise `(x = (1.0,),)`. 

I guess `false` would need careful handling. Something like `Flux.loadmodel!(([0.01;;], [0.222], [0.03;;]), ([111;;], false, [333;;]))` should still work.

mcabbott: BatchNorm is a nice example. Right now `Flux.loadmodel!(BatchNorm(1, affine=true), BatchNorm(1, affine=false))` does not work. Filtering one of those out (for having no parameters) would lead to a perhaps more confusing error, but it's hard to picture it leading to doing the wrong thing.

ToucheSir: > Something like `(x = 1.0,)` has children but, in my scheme, would be filtered out of a Tuple as it contains no parameters to load. Likewise `(x = (1.0,),)`.

Depending on context, that seems pretty important! There is perhaps a philosophical discussion to be had about whether e.g. Dropout rates should be loadable (or saveable), but we should provide clear and consistent behaviour one way or the other.

mcabbott: Sure. Right now scalars are never copied, only arrays:
```
julia> Flux.loadmodel!(
         Parallel(vcat, Dense(1=>1), Dropout(0.11)),
         (connection = hcat, layers = (Dense([0;;]), Dropout(0.99))),
       )
Parallel(
  vcat,
  Dense(1 => 1),                        # 2 parameters  (all zero)
  Dropout(0.11),
) 
```

darsnack: Right, the original PR already made a call here to only handle arrays.

ToucheSir: I wonder if the filter callback should take both src and dest values.

mcabbott: > filter both BatchNorms (so that the tree structure remains aligned) but still be able to load the other weights

This is not covered by my rule above. (Nor by the present code.) I can see how this might be desirable, but what's the formalised rule for it?

darsnack: It would be covered by the present code via `loadmodel!(dst, src; filter = x -> !(x isa BatchNorm)`, right?

mcabbott: Sorry, I meant it's an error on tagged Flux. (Thus non-breaking to change.) I agree this PR's filter would work there. It would not work if one of them was `fmapstructure`'d.

darsnack: > I wonder if the filter callback should take both src and dest values.

This would be trickier with the predicate style callback when the list of destination and source children are not the same length (e.g. one has `Dropout` while the other does not). I think to make this work and not get messy, you'd have to allow the callback to the filtering itself (at which point it could be licensed to _anything_). Though I agree the ability to make a decision to skip based on both destination and source would be more powerful.

darsnack: > Thus non-breaking to change.

Specifically on this point, I think since the `filter` keyword did not exist on tagged Flux, and the default value is `_ -> true`, even this PR adds these features in a non-breaking manner, i.e. the example will still error unless the user specifies `filter`.

> My question, since we're wondering whether do-what-I-want is possible, is whether there's a nice automatic rule which could see this for you. Some layers are known never to change the size (BatchNorm, Dropout) and so could conceivably be auto-ignore-when-mismatched. But I'm not sure the exact rule.

Yes, I think this would be good to handle automatically, but maybe we make this rule (whatever it should be), a default value for `filter` in a breaking update? Then there's always the option to turn it off or override it.

ToucheSir:
The lack of this is affecting Metalhead CI, right? Can we treat `filter =` as a de-facto private interface to unblock that while discussion continues on a more optimal solution?

darsnack:
So just remove it from the docstring and NEWS?

darsnack:
Done

src/loading.jl
```diff
@@ -77,9 +81,9 @@ however, attempting to copy a non-zero array to an inactive parameter will throw
 Likewise, copying a `src` value of `false` to any `dst` array is valid,
 but copying a `src` value of `true` will error.
 """
-function loadmodel!(dst, src; cache = Base.IdSet())
-  ldsts, _ = functor(dst)
-  lsrcs, _ = functor(src)
+function loadmodel!(dst, src; filter = Returns(true), cache = Base.IdSet())
```
ToucheSir, line 84: `Returns` is not available on `1.6`. We'd need either to change to 
```suggestion
function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
```
Or use Compat (which may be useful elsewhere?)

ToucheSir (approved):

Issue #2006, opened 2022-06-24
Deprecate `track_stats=true` for `GroupNorm` and `InstanceNorm`


ToucheSir:
We discussed this on Slack, and it appears other libraries do not track running stats for either. Nor do any papers appear to make use of a running average. Thus the complexity feels very much not worth it, and doing it by default creates unnecessary inconsistencies with the rest of the ecosystem.

The rough deprecation plan for this is as follows:
1. Emit a warning when users pass `track_stats=true`, with a pointer to use `track_stats=false` since the former will go away in the next breaking Flux release. This is the "good first issue" part.
2. For that breaking release, remove the tracking entirely and leave `track_stats` as a vestigial kwarg. Alternatively, we could remove it completely.

akahard2dj:
Hello! Julia Libraty contirbution is first time in my life!
Should I scarach this one?

ToucheSir:
If this is your first PR to _any_ Julia library, it may be a little tricky. Feel free to take a shot though :)

ToucheSir:
This was fixed in #2042, but it was overzealous in the specification and erroneously included `InstanceNorm`. #2149 fixes that.

PR #2042, opened 2022-08-20
Mark `track_stats=true` as deprecated

akahard2dj:
First job of  the issue #2006. 
`track_stats`  will be removed next breaking Flux release.
First job is only focus to warn that `track_stats` arguments.

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -529,6 +529,10 @@ function GroupNorm(chs::Int, G::Int, λ=identity;
               affine=true, track_stats=false,
               ϵ=1f-5, momentum=0.1f0)
 
+if track_stats
+  @warn("`track_stats=true` will be removed. Please use `track_stats=false` instead")
+end
+
   chs % G == 0 || error("The number of groups ($(G)) must divide the number of channels ($chs)")
 
   β = affine ? initβ(chs) : nothing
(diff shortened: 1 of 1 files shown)


src/layers/normalise.jl
```diff
@@ -431,6 +431,10 @@ function InstanceNorm(chs::Int, λ=identity;
                     affine=false, track_stats=false,
                     ϵ=1f-5, momentum=0.1f0)
 
+  if track_stats
+    @warn("`track_stats=true` will be removed. Please use `track_stats=false` instead")
+  end
```
mcabbott, line 436: Should this use `Base.depwarn`? Maybe changed to `force=true` on the last 0.13 release (otherwise it only prints during tests etc.)
```suggestion
  if track_stats
    Base.depwarn("`track_stats=true` will be removed from InstanceNorm in Flux 0.14. The default value is `track_stats=false`, which will work as before.", :InstanceNorm)
  end
```

akahard2dj: It does not be triggered `Base.depwarn` when I use. So I use a `@warn` for triggering a deprecation message. Could you give some tips?

ToucheSir: You may need to call `Base.depwarn("...", :InstanceNorm; force=true)` as suggested above.

akahard2dj: @ToucheSir Thanks for sharing your valuable tip. `force=flag` is really work properly.

mcabbott (approved):
Maybe worth adding a comment in https://github.com/FluxML/Flux.jl/blob/master/src/deprecations.jl  so that when someone reviews that to see what can be removed, they remember this? 

Otherwise looks good, thanks!

ToucheSir (approved):

PR #2046, opened 2022-08-22
Better docs for reexported packages

### LINKED ISSUE #2038: [Discussion]: documentation for `@reexport`ed and `import`ed (or `using`) packages
The `Flux` namespace imports a lot of packages - some of them have their separate pages in `Flux`'s documentation, whereas some of them don't, and currently, there isn't a single rule for such packages.

The packages available under `Flux`'s namespace - 
- `Optimisers.jl`: has a separate section in `Flux`'s docs and a stand-alone documentation
- `NNlib.jl`: has a separate page in `Flux`'s docs and should have stand-alone documentation soon (see https://github.com/FluxML/NNlib.jl/pull/431)
- `OneHotArrays.jl`: recently added, `Flux`'s docs have not been updated, stand-alone docs are not deployed, is blocking #2035
- `MLUtils.jl`: has a separate page in `Flux`'s docs and a stand-alone documentation
- `Functors.jl`: has a separate page in `Flux`'s docs and a stand-alone documentation
- `Zygote.jl`: has stand-alone documentation and bits and pieces scattered over `Flux`'s documentation

I don't think we should remove these separate pages from Flux's documentation, but these pages should not do the job of the documentation of these independent packages.

We can either have separate pages detailing some of the functionalities provided by the package that would be of interest to Flux users (for example see the latest commit of #2035) or display the API reference of this package (for example, see https://fluxml.ai/Flux.jl/stable/models/nnlib/).

How should such documentation exist? Open for discussion :)

Saransh-cpp:
Closes #2038 

This PR -
- makes the page titles more understandable and more welcoming for newcomers.
- syncs the API reference of `NNlib` with https://github.com/FluxML/NNlib.jl/pull/431

There is a tutorial (or more precisely, a getting-started/how-to guide) on `DataLoader` available here - https://fluxml.ai/tutorials/2021/01/21/data-loader.html. Should this be shifted to the new "Getting Started" section?

_Originally posted by @Saransh-cpp in https://github.com/FluxML/Flux.jl/issues/2038#issuecomment-1222478949_

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -18,12 +18,12 @@ makedocs(
             "Loss Functions" => "models/losses.md",
             "Regularisation" => "models/regularisation.md",
             "Advanced Model Building" => "models/advanced.md",
-            "NNlib" => "models/nnlib.md",
-            "Functors" => "models/functors.md"
+            "Neural Network primitives from NNlib.jl" => "models/nnlib.md",
+            "Functor from Functors.jl" => "models/functors.md"
          ],
          "Handling Data" => [
-             "One-Hot Encoding" => "data/onehot.md",
-             "MLUtils" => "data/mlutils.md"
+             "One-Hot Encoding with OneHotArrays.jl" => "data/onehot.md",
+             "Working with data using MLUtils.jl" => "data/mlutils.md"
          ],
          "Training Models" => [
              "Optimisers" => "training/optimisers.md",
--- a/docs/src/data/mlutils.md
+++ b/docs/src/data/mlutils.md
@@ -1,4 +1,4 @@
-# MLUtils.jl
+# Working with data using MLUtils.jl
 
 Flux re-exports the `DataLoader` type and utility functions for working with
 data from [MLUtils](https://github.com/JuliaML/MLUtils.jl).
--- a/docs/src/data/mlutils.md
+++ b/docs/src/data/mlutils.md
@@ -22,6 +22,7 @@ Below is a non-exhaustive list of such utility functions.
 
 ```@docs
 MLUtils.unsqueeze
+MLUtils.flatten
 MLUtils.stack
 MLUtils.unstack
 MLUtils.numobs
--- a/docs/src/data/mlutils.md
+++ b/docs/src/data/mlutils.md
@@ -7,6 +7,8 @@ data from [MLUtils](https://github.com/JuliaML/MLUtils.jl).
 
 `DataLoader` can be used to handle iteration over mini-batches of data.
 
+`Flux`'s website has a [dedicated tutorial](https://fluxml.ai/tutorials/2021/01/21/data-loader.html) on `DataLoader` for more information. 
+
 ```@docs
 MLUtils.DataLoader
 ```
@@ -20,7 +22,6 @@ Below is a non-exhaustive list of such utility functions.
 
 ```@docs
 MLUtils.unsqueeze
-MLUtils.flatten
 MLUtils.stack
 MLUtils.unstack
 MLUtils.numobs
--- a/docs/src/models/functors.md
+++ b/docs/src/models/functors.md
@@ -1,6 +1,6 @@
-# Functor from Functors.jl
+# Recursive transformations from Functors.jl
 
-Flux makes use of the [Functors.jl](https://github.com/FluxML/Functors.jl) to represent many of the core functionalities it provides.
+Flux models are deeply nested structures, and [Functors.jl](https://github.com/FluxML/Functors.jl) provides tools needed to explore such objects, apply functions to the parameters they contain, and re-build them.
 
 Functors.jl is a collection of tools designed to represent a [functor](https://en.wikipedia.org/wiki/Functor_(functional_programming)). Flux makes use of it to treat certain structs as functors. Notable examples include the layers that Flux defines.
 
--- a/docs/src/models/functors.md
+++ b/docs/src/models/functors.md
@@ -1,4 +1,4 @@
-# Functors.jl
+# Functor from Functors.jl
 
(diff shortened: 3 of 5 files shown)


Saransh-cpp:
TODO:

- [ ] Add a tutorial link for `Functors.jl`.
- [ ] Add a tutorial link for `DataLoader.jl` or move the tutorial to the new "Getting Started" guide.

docs/src/models/functors.md
```diff
@@ -1,4 +1,4 @@
-# Functors.jl
+# Functor from Functors.jl
 
 Flux makes use of the [Functors.jl](https://github.com/FluxML/Functors.jl) to represent many of the core functionalities it provides.
```
mcabbott, line 4: This seems a bit vague to me... first, this use of "functor" belongs to a tiny nice of functional programming, I don't think it's worth putting any emphasis on that here, it's just some package with a weird name. 

Second, I don't know what "many of the core functionalities" means. I wrote something but maybe giving examples (params, training, gpu) would be nice too.

```suggestion
# Recursive transformations Functors.jl

Flux models are deeply nested structures, and [Functors.jl](https://github.com/FluxML/Functors.jl) provides tools needed to explore such objects, apply functions to the parameters they contain, and re-build them.

```

I can't make suggestions below this, but (IMO) it might also be worth separating the list of functions according to level of obscurity. Maybe `@functor`, `functor`, `isleaf`, `fmap` should be in one block, the rest in another? Or just `@functor`, `fmap`  as the top? Not sure, maybe it's too messy to make such a division.

Saransh-cpp: The number of docstrings looks too few for a further division; should I still divide them?

ToucheSir: I would think of and describe functors like you would the module system of any other library. Because while it is a general-purpose library, that's what we use it for in Flux. One way to do this could be to show practical examples of using `@functor` etc. to define layers, and then explain what is happening along with why things are done this way. If that last part is too much, just the examples and a link to the advanced model building page could suffice.

docs/src/data/mlutils.md
```diff
@@ -1,4 +1,4 @@
-# MLUtils.jl
+# Working with data using MLUtils.jl
```
mcabbott, line 1: I can't comment below, but `flatten` appears on this page (as it should), and also here:

https://fluxml.ai/Flux.jl/latest/models/layers/#Flux.flatten

Should it be removed?

docs/src/models/functors.md
```diff
@@ -1,15 +1,17 @@
...
+# Recursive transformations from Functors.jl
 
-Flux makes use of the [Functors.jl](https://github.com/FluxML/Functors.jl) to represent many of the core functionalities it provides.
+Flux models are deeply nested structures, and [Functors.jl](https://github.com/FluxML/Functors.jl) provides tools needed to explore such objects, apply functions to the parameters they contain, and re-build them.
 
-Functors.jl is a collection of tools designed to represent a [functor](https://en.wikipedia.org/wiki/Functor_(functional_programming)). Flux makes use of it to treat certain structs as functors. Notable examples include the layers that Flux defines.
+More precisely, using the `Functors.@functor` macro allows `Flux` layers to access additional functionalities, like collecting parameters or scaling them to the GPU.
+
+Functors.jl is a collection of tools designed to represent a [functor](https://en.wikipedia.org/wiki/Functor_(functional_programming)). Flux makes use of it to treat certain structs as functors. Notable examples include the layers that Flux defines. The basic usage of `Functors.jl` can be found [here](https://fluxml.ai/Functors.jl/stable/#Basic-Usage-and-Implementation).
```
mcabbott, line 7: I still think this is too many words. Maybe it could be cut like this, possibly integrated with the paragraph above.
```suggestion
New layers should be annotated using the `Functors.@functor` macro. This will enable [`params`](@ref) to see the parameters inside, and [`gpu`](@ref) to move them to the GPU.

`Functors.jl` has its own [notes on basic usage](https://fluxml.ai/Functors.jl/stable/#Basic-Usage-and-Implementation) for more details.
```
This will change when we ditch Params, and if we do #/1932 or #2028, so perhaps not worth too much care just now.

mcabbott (approved):
This looks like a good step forwards, thanks!

docs/make.jl
```diff
@@ -18,12 +18,12 @@ makedocs(
             "Loss Functions" => "models/losses.md",
             "Regularisation" => "models/regularisation.md",
             "Advanced Model Building" => "models/advanced.md",
-            "NNlib" => "models/nnlib.md",
-            "Functors" => "models/functors.md"
+            "Neural Network primitives from NNlib.jl" => "models/nnlib.md",
+            "Functor from Functors.jl" => "models/functors.md"
```
mcabbott, line 22: One more thought. Should Zygote.jl be among the packages which gets a sidebar heading?

docs/src/data/mlutils.md
Saransh-cpp: Ah, yes! Thanks!

Saransh-cpp: Correction: [`MLUtils.unsqueeze`](https://fluxml.ai/Flux.jl/latest/data/mlutils/#MLUtils.unsqueeze) cross-references `MLUtils.flatten`, and the doctests fail if I remove `MLUtils.flatten`'s reference from the docs.

mcabbott:
Let's merge this and refine more if we need to, later. Thanks!

Saransh-cpp:
Thanks for the review! I missed the `Zygote.jl` suggestion, but I can add it in another PR.

PR #2051, opened 2022-08-28
Add extra test for params

christiangnrd:
Adding the test suggested by @mcabbott in pull-request #2048 to prevent future regression.

If there are any more cases that you think should be added, let me know.

I made a separate pull-request because I feel that this test should be added regardless of the stats of #2048.

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -264,7 +264,7 @@ end
   r[1] = r
   @test size.(params(r)) == [(5, 10), (5, 5), (5,), (5, 1)]
 
-  # Ensure proper behaviour of the functor recursion
+  # Ensure functor explores inside Transpose but not SubArray
   m = (x = view([1,2,3]pi, 1:2), y = transpose([4 5]pi))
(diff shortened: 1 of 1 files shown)


test/utils.jl
```diff
@@ -263,6 +263,10 @@ end
   r = Any[nothing,m]
   r[1] = r
   @test size.(params(r)) == [(5, 10), (5, 5), (5,), (5, 1)]
+
+  # Ensure proper behaviour of the functor recursion
```
mcabbott, line 267:
```suggestion
  # Ensure functor explores inside Transpose but not SubArray
```

mcabbott (approved):
I don't love `size.(Flux.params(m))` as I'm not sure the order is guaranteed, but lots of other tests above do the same.

Maybe the comment can be clearer too, otherwise fine, thanks!

christiangnrd:
I used `size.(Flux.params(m))` because all the other tests do the same. I agree about the comment being vague, l wrote it as specific as I could given my current familiarity with the code :)

PR #1005, opened 2020-01-25
Added WeightNorm

bhvieira:
In light of #993 I wanted to create a WeightNorm constructor that could operate over any layer, param and dim.

This proved harder than I thought, but with the help of @chengchingwen and @mcabbott I could finally make this work.

So the execution is simply this:

1. User calls `WeightNorm` on a layer
2. We catch the param they want to normalize, and keep its direction and magnitude in a new type, `WeightNormWeight`
3. We redefine several operations over this type (so it operates like an Array) while Zygote can transparently see that it's actually a direction and magnitude
4. We reconstruct the layer using the `functor` functionality with the substitution in place

The catch is that it is not ready though.

~~- [ ] Make it work with `NNlib.conv`~~ (I really want to make this work, but now I noticed it's not only Conv that will be laborious to fix, simple stuff like RNNs will also require some iterations to get working simply because they hide their params inside cells)
- [x] Remove dead code (I'm pretty sure some things I did there should not be needed, specially regarding the `Base` array operations. But I couldn't subtype `WeightNormWeight` into `AbstractArray` either
- [x] Make it accept more than one `weight` at a time, potentially with differing `dim` as well.
- [x] Add tests

So, here's a first try at least. I could throw an error to `Conv`, but I want to make it work and perhaps it'll be better to wait out until everything is right.

Inputs welcome!

PR #2053, opened 2022-08-29
Add `WeightNorm` layer

### LINKED ISSUE #942: Feature request: Weight normalization

mcabbott:
RFC as to whether this looks right, and whether the approach is a good one.

Needs tests.

Edit: earlier attempts are #993 (Dense only) and #1005. I had completely forgotten this, but the approach here is similar to @CarloLucibello's suggestion in https://github.com/FluxML/Flux.jl/pull/1005#issuecomment-582973027. (But I'm not sure why the forward pass is unchanged in the example there.)

Closes #942

#### Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/models/layers.md
+++ b/docs/src/models/layers.md
@@ -76,6 +76,8 @@ AlphaDropout
 LayerNorm
 InstanceNorm
 GroupNorm
+WeightNorm
+WeightNorm(::Dense)
 ```
 
 ### Testmode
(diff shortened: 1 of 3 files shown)


bhvieira:
I like the approach, would just advise to do performance tests as well. Due to compiler optimizations, I guess/hope this doesn't incur too much penalization. Also, would this work with recurrent layers? Perhaps define a new dispatch when it finds a `Recur`. That was also a big blocker in #1005, but I don't believe it should block merging this

CarloLucibello:
Done in #2550

PR #2054, opened 2022-08-29
Make params non-differentiable (Closes #2040 & #2048)

### LINKED ISSUE #2040: Pull request #2007 causes Flux.params() calls to not get cached

christiangnrd:
This is the new pull request I mentioned in #2048 to allow `params()` calls from within `gradient()` to be cached by making `params()` non-differentiable. This would close issue #2040, and supersede pull request #2048.

Might be worth turning on downstream test like in the other pr.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -88,7 +88,7 @@ function params(m...)
   return ps
 end
 
-# Allows caching of the parameters when params is called within gradient()
+# Allows caching of the parameters when params is called within gradient() to fix #2040.
 @non_differentiable params(m...)
 
 struct FluxCUDAAdaptor end
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -88,6 +88,9 @@ function params(m...)
   return ps
 end
 
+# Allows caching of the parameters when params is called within gradient()
+@non_differentiable params(m...)
+
 struct FluxCUDAAdaptor end
 adapt_storage(to::FluxCUDAAdaptor, x) = CUDA.cu(x)
 adapt_storage(to::FluxCUDAAdaptor, x::Zygote.FillArrays.AbstractFill) = CUDA.cu(collect(x))


christiangnrd:
@ToucheSir can you enable the downstream tests for this pr?

ToucheSir:
FastAI failure is innocuous and known. Metalhead I think too (cc @darsnack to confirm it's just the ResNet weights loading). GeometricFlux I'm not sure about, but there doesn't appear to be any AD calls in https://github.com/FluxML/GeometricFlux.jl/blob/master/test/layers/graphlayers.jl. @yuehhua can you confirm?

yuehhua:
Failure in GeometricFlux is not related to AD and this PR could go. I will check the error further.

src/functor.jl
```diff
@@ -88,6 +88,9 @@ function params(m...)
   return ps
 end
 
+# Allows caching of the parameters when params is called within gradient()
```
ToucheSir, line 91: Can you add a quick mention of https://github.com/FluxML/Flux.jl/issues/2040 here? Otherwise this looks good to go!

ToucheSir (approved):

src/functor.jl
christiangnrd: Just pushed with edited comment.

ToucheSir:
Thanks!

mcabbott:
Unfortunately this breaks the use of `params` in explicit mode:
```julia
julia> m = (x=[1,2.0], y=[3.0]);

julia> gradient(m -> (sum(norm, Flux.params(m))), m)
((x = [0.4472135954999579, 0.8944271909999159], y = [1.0]),)  # before, Flux@0.13.5
(nothing,)  # after, Flux@0.13.6

julia> gradient(() -> (sum(norm, Flux.params(m))), Flux.params(m))
Grads(...)

julia> ans[m.y]  # unchanged
1-element Vector{Float64}:
 1.0
```

PR #2055, opened 2022-08-29
Leftover changes from #2046

Saransh-cpp:
@mcabbott, I was about to make these changes in #2046, but it was merged. I have added them in this follow-up PR -

* remove linear_regression.jl (committed by mistake)
* @ToucheSir's comments
* Update Functor.jl's sidebar title

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


docs/make.jl
```diff
@@ -19,7 +19,7 @@ makedocs(
             "Regularisation" => "models/regularisation.md",
             "Advanced Model Building" => "models/advanced.md",
             "Neural Network primitives from NNlib.jl" => "models/nnlib.md",
-            "Functor from Functors.jl" => "models/functors.md"
+            "Recursive transformations from Functors.jl" => "models/functors.md"
          ],
          "Handling Data" => [
              "One-Hot Encoding with OneHotArrays.jl" => "data/onehot.md",
```
mcabbott, line 25: Sorry if I rushed it. At least this way it's easy to see the preview, https://fluxml.ai/Flux.jl/dev/ 

I wonder whether all the sidebar titles should be just the package name, not the description? This would make a bit of a distinction between internal and external docs, and would fit on one line each. 

Screenshot 2022-08-29 at 14 29 53

mcabbott:
While fiddling, is there a better way to organise this bit of the docs? It feels a bit like random things are dumped down here. 

Screenshot 2022-08-29 at 14 54 48

There has got to be a better way to organise this material, but I'm not exactly sure what what the right top-level headings are. 

Under "utilities" this groups:
* All the init functions do need to live somewhere. It's a long list. Maybe it should be a subsection near to NNlib? (Maybe NNlib should be two sections, in fact -- one to list the activation functions which you use, another for the backend functions your layers use?) 
* "Model Building" means `outputsize`. It's nice but nobody will find it here. 
* "Model Abstraction" just means Flux.modules. Since this is almost `fcollect` should it live with Functors.jl, or are those sections exclusively functions from the package?
* "Changing the type" should this live with initialisaton, or live with Functors.jl? 
* "Callback helpers" are very random. 

The top-level heading "datasets" has two links. Can it just move to "ecosystem"? 

Also can that be "The Flux Ecosystem" not the Julia one? It's things somewhat connected to Flux, and is a nice list.

Can "community" heading be deleted... it could move to a sub-heading on the very first page, and/or under "ecosystem"?

mcabbott (approved):

PR #2059, opened 2022-09-06
Fix the failing CI build

christiangnrd:
It seems like Julia 1.9 will add a `stack` method, so this pr specifies which `stack` is to be used during tests.

I ran the tests locally on 1.8 and the github master branch and it seems to have fixed the failing CI builds on nightly.

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -1,7 +1,7 @@
 using Flux
 using Flux: throttle, nfan, glorot_uniform, glorot_normal,
              kaiming_normal, kaiming_uniform, orthogonal, truncated_normal,
-             sparse_init, identity_init, stack, unstack, batch, unbatch,
+             sparse_init, identity_init, unstack, batch, unbatch,
              unsqueeze, params, loadparams!, loadmodel!
 using StatsBase: var, std
 using Statistics, LinearAlgebra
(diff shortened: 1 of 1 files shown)


christiangnrd:
Boy is it nice to see a green check mark.

test/utils.jl
```diff
@@ -326,14 +326,14 @@ end
 
 @testset "Stacking" begin
   x = randn(3,3)
-  stacked = stack([x, x], dims=2)
+  stacked = Flux.MLUtils.stack([x, x], dims=2)
```
ToucheSir, line 329:
```suggestion
  stacked = MLUtils.stack([x, x], dims=2)
```
We'd like to avoid reaching into modules to get dependencies which are directly accessible in the project, primarily because Flux may stop importing MLUtils (or any other dep) at the top level at some point. This, doing the same below and adding an import for MLUtils should do the trick.

ToucheSir (commented):
One small thing and this should be good to go :)

test/utils.jl
christiangnrd: I've made the changes and all tests passed.

christiangnrd:
Off-topic but what's left until 0.13.6?

ToucheSir (approved):
Thanks!

darsnack:
> Off-topic but what's left until 0.13.6?

Nothing, we're planning that release imminently.

PR #2063, opened 2022-09-21
Add a dark mode version of logo

Saransh-cpp:
- dark mode version of logo
- no sidebar title
- dynamic rendering of dark/light logo in the README file

See https://github.com/FluxML/Zygote.jl/pull/1306

image

image

The white Flux logo looks a bit weird to me, but maybe because I am used to seeing the black Flux logo?

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


darsnack:
Maybe a white outline on the arrow is better than a totally white arrow?

mcabbott (approved):

ToucheSir:
> Maybe a white outline on the arrow is better than a totally white arrow?

I concur. At the very least, it would be nice to have distinct shades for the arrow and outline.

mcabbott:
Shall we try this out? Seems like a step forwards. 

I'm not too sure how @Saransh-cpp  made this, but I guess replacing a colour is easier than adding new borders of a different colour.

Saransh-cpp:
> I guess replacing a colour is easier than adding new borders of a different colour.

Yes! It took me some time (not a graphic designer 😬), but here are the designs (see them in GitHub's dark mode) -

IMG_0616
IMG_0620

Saransh-cpp:
Note: The arrow in current Flux logo has a greyish outline.

mcabbott:
I actually think this looks fine. Maybe there's a design where the text & arrow have boundaries like the "layers" do, but this sounds like a lot of work, and not immediately obvious it would be better. I think you'd still want the bulk of the text white (in dark mode) to match the other text. 

Before this PR: 
Screenshot 2022-09-27 at 10 13 05

After:
Screenshot 2022-09-27 at 10 12 52Screenshot 2022-09-27 at 10 15 06

Saransh-cpp:
`conda-forge` does something similar with their logo, but the current one looks good to me as well - 

image
image
image

Saransh-cpp:
Also, now that I look at it, maybe we should always use the old logo in README? Flux uses the old logo everywhere (for example - FluxML's GitHub logo/display picture), and most of them cannot be made dynamic.

image

PR #2064, opened 2022-09-22
Fix a few crossrefs + update Zygote's page

Saransh-cpp:
- Anything beyond an `h2` (`##`) does not appear in the sidebar
- Fixed a few API references for Zygote
- Fixed a few internal cross-references

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/models/losses.md
+++ b/docs/src/models/losses.md
@@ -21,7 +21,7 @@ loss(ŷ, y, agg=x->mean(w .* x))    # weighted mean
 loss(ŷ, y, agg=identity)           # no aggregation.
 ```
 
-### Function listing
+## Function listing
 
 ```@docs
 Flux.Losses.mae
--- a/docs/src/models/losses.md
+++ b/docs/src/models/losses.md
@@ -21,7 +21,7 @@ loss(ŷ, y, agg=x->mean(w .* x))    # weighted mean
 loss(ŷ, y, agg=identity)           # no aggregation.
 ```
 
-## Function listing
+### Function listing
 
 ```@docs
 Flux.Losses.mae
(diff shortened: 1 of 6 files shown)


docs/src/models/losses.md
```diff
@@ -21,7 +21,7 @@ loss(ŷ, y, agg=x->mean(w .* x))    # weighted mean
 loss(ŷ, y, agg=identity)           # no aggregation.
 ```
 
-### Function listing
+## Function listing
```
mcabbott, line 24: My thinking with H3 was that it's nice to separate the listing from the introductory blurb, but it doesn't really need an entry on the side-bar if there's just one heading. 


Screenshot 2022-09-22 at 18 32 02

docs/src/training/zygote.md
```diff
@@ -36,4 +37,17 @@ To manually supply the gradient for one function, you should define a method of
...
+
+Zygote provides a user with some friendly utility functions.
+All the utility functions are available in [Zygote's documentation](https://fluxml.ai/Zygote.jl/latest/utils/).
+
+```@docs
+Zygote.hessian
+Zygote.hessian_reverse
+Zygote.diaghessian
+Zygote.withjacobian
+Zygote.withgradient
```
mcabbott, line 52: I think my vote would be to include `withgradient` alongside gradient. Then this section would just be about second derivatives, so it can have a heading more useful than "utilities". 

However, these are very limited, Zygote over Zygote never works that well. So perhaps it needs disclaimers here. 

And perhaps it wants explanation of how to use hessian with destructure, since oherwise it can't act on a Flux model parameters?

docs/src/models/losses.md
mcabbott:
```suggestion
### Function Listing
```

docs/src/training/zygote.md
mcabbott: (Such a Hessian example is now in `destructure`'s own page)

mcabbott (approved):

Saransh-cpp:
Yes, went through the page and decided to remove this section!

PR #2065, opened 2022-09-22
Make `rng_from_array` non differentiable

### LINKED ISSUE #2062: Make `rng_from_array` non-differentiable
Using this like:
```julia
function mylayer(x)
    rng = rng_from_array(x)
    ...
end
```
will try to propagate gradients through which we don't want.

Saransh-cpp:
Fixes #2062 

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -46,6 +46,8 @@ The current defaults are:
 rng_from_array(::AbstractArray) = default_rng_value()
 rng_from_array(::CuArray) = CUDA.default_rng()
 
+@non_differentiable rng_from_array(::Any)
+
 if VERSION >= v"1.7"
   @doc """
       default_rng_value()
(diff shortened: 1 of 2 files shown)


ToucheSir (approved):
Thanks! Could you add a quick test which uses this with `gradient`?

Saransh-cpp:
How can I check whether the `gradient` is propagating through `rng_from_array`? I have something like right now - 
```julia
@test "rng_from_array" begin
  m(x) = (rand(rng_from_array(x)) * x)[1]
  gradient(m, ones(2))
end
```
but I am not sure if this is testing anything concrete?

ToucheSir:
That should be fine. We just need something that will complain if the non-differentiability breaks down at some point.

Saransh-cpp:
Thanks, @ToucheSir! This should work.

ToucheSir (approved):
LGTM

PR #2066, opened 2022-09-26
Speed up `show` for gpu models

mcabbott:
The present `show` methods have about a 20s startup delay when the model is on the GPU. This comes from the checks like `any(isnan, x)` which print friendly warnings. Perhaps we should remove them, to save startup time?

This PR replaces them with an "(on GPU)" annotation, like so:
```
julia> m
Chain(
  Dense(2 => 3, σ),                     # 9 parameters
  Dense(3 => 2),                        # 8 parameters  (some NaN)
  NNlib.softmax,
)                   # Total: 4 arrays, 17 parameters, 324 bytes.

julia> m |> gpu
Chain(
  Dense(2 => 3, σ),                     # 9 parameters (on GPU)
  Dense(3 => 2),                        # 8 parameters (on GPU)
  NNlib.softmax,
)                   # Total: 4 arrays, 17 parameters, 576 bytes.
```
Maybe that ends up quite noisy for bigger models. For this reason I didn't make it some bright colour.

### DIFF
--- a/src/layers/show.jl
+++ b/src/layers/show.jl
@@ -113,7 +113,10 @@ underscorise(n::Integer) =
   join(reverse(join.(reverse.(Iterators.partition(digits(n), 3)))), '_')
 
 function _nan_show(io::IO, x)
-  if !isempty(x) && _all(iszero, x)
+  if any(y -> y isa Zygote.AbstractGPUArray, x)
+    # These friendly warnings take 10-20 sec to compile the first time, for models on GPU. 
+    printstyled(io, " (on GPU)", color=:light_black)
+  elseif !isempty(x) && _all(iszero, x)
     printstyled(io, "  (all zero)", color=:cyan)
   elseif _any(isnan, x)
     printstyled(io, "  (some NaN)", color=:red)


darsnack:
I think it might be confusing for users. Someone used to `show` on the CPU might interpret the absence of NaN messages as no NaNs.

Is it possible to make this method more "official" with knobs to turn on/off the details? Then the default `show` calls this with the details turned off.

mcabbott:
Yes this is my concern too.

I wonder a bit how useful these things really are, how often do people get NaN? Maybe typically through the loss becoming infinite etc? 

Can we pre-compile GPU stuff? It just seems a little crazy to wait 20s for the printing code to warm up...

ToucheSir:
> Can we pre-compile GPU stuff?

We could definitely try using SnoopPrecompile. Based on https://github.com/JuliaLang/julia/issues/46296 and https://github.com/JuliaLang/julia/pull/46373 though, that may have to wait until the next 1.7 and 1.8 point releases.

CarloLucibello:

PR #2036, opened 2022-08-08
Revamp `basic.md` (put it under the `getting_started` section)

Saransh-cpp:
This PR revamps `basic.md` and renames it to "Neural Networks from scratch". Most of the content is intact, but I have cleaned up the code and removed the bits and pieces which are not supposed to be used now.

Note: This PR contains the changes from the linear regression PR and the logistic regression PR (without some of these changes, the tests would fail). This PR should be merged after both of these PRs. I will make sure to sync the changes before this PR is merged.

Lastly, I am not sure if this guide should contain a training section. The title talks only about constructing Neural Networks, but I can add a training function and showcase `Optimisers.jl`.

Rendered guide - https://github.com/Saransh-cpp/Flux.jl/blob/revamp-basic/docs/src/getting_started/neural_nets_from_scratch.md

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md
- [ ] Documentation, if applicable

PR #2068, opened 2022-09-27
Add a quick start example, and change some headings

mcabbott:
This changes the start of the docs, so that "Getting Started" has (after the index page) two paths: A fast one (for people who know about neural nets, just not Flux) and a slow one (introducing linear regressing, parameters, etc). The slow one was called "Overview" before. The fast one is new. 

Re-reading "Overview", it's nicer than I remembered. I'm not entirely sure it should live in a different section to the next one, "Basics" (which I call "Gradients and Layers" here). Thoughts? Maybe the second "Basics" one should also move into "Getting Started"? (I do think they need better titles.)

I'm not precisely sure how this intersects with #2036, #2021, #2016. Those may replace text in some sections, to make the slow path even more detailed, or to upgrade the "Basics" section. Or maybe we want a tutorials section in the docs, which replaces the one in the website?


ToucheSir:
If you port this to a branch on the Flux repo, that should trigger the PR preview workflow.

mcabbott:
Oh that's what I forgot. Is this possible while keeping the same PR open? Probably not.

Issue #2049, opened 2022-08-25
Make `create_bias` a public API?


ToucheSir:
This was recently marked private in https://github.com/FluxML/Flux.jl/pull/1998, but looking at a recent downstream test run (my mistake for not enabling them on that PR) it appears that [downstream libraries](https://github.com/FluxML/Flux.jl/runs/8008495389?check_suite_focus=true) are relying on it. If we want to dissuade them from continuing to use it, perhaps we could push a hotfix with non-underscore `create_bias` marked deprecated? Otherwise we should just restore it and render the docstring, I think.

Saransh-cpp:
The tests are also failing due to the `rng_from_array()` -> `default_rng_value()` renaming -
```
config = small: Error During Test at /home/runner/work/Flux.jl/Flux.jl/downstream/test/convnets.jl:332
  Got exception outside of a @test
  MethodError: no method matching rng_from_array()
  Closest candidates are:
    rng_from_array(::CUDA.CuArray) at ~/work/Flux.jl/Flux.jl/src/utils.jl:47
    rng_from_array(::AbstractArray) at ~/work/Flux.jl/Flux.jl/src/utils.jl:46
  Stacktrace:
    [1] DropPath(p::Float64)
      @ Metalhead.Layers ~/work/Flux.jl/Flux.jl/downstream/src/layers/drop.jl:159
    [2] convnextblock(planes::Int64, drop_path_rate::Float64, layerscale_init::Float32)
      @ Metalhead ~/work/Flux.jl/Flux.jl/downstream/src/convnets/convnext.jl:14
    [3] (::Metalhead.var"#187#188"{Float32, Vector{Int64}, LinRange{Float64, Int64}, Int64})(j::Int64)
      @ Metalhead ./none:0
    [4] iterate
      @ ./generator.jl:47 [inlined]
    [5] collect
      @ ./array.jl:787 [inlined]
    [6] convnext(depths::Vector{Int64}, planes::Vector{Int64}; drop_path_rate::Float64, layerscale_init::Float32, inchannels::Int64, nclasses::Int64)
      @ Metalhead ~/work/Flux.jl/Flux.jl/downstream/src/convnets/convnext.jl:60
    [7] #ConvNeXt#191
      @ ~/work/Flux.jl/Flux.jl/downstream/src/convnets/convnext.jl:99 [inlined]
    [8] ConvNeXt
[...]

PR #2069, opened 2022-09-27
Add a quick start example, and change some headings

mcabbott:
[Replaces #2068]

This wants to change the start of the docs, so that "Getting Started" has (after the index page) two paths: A fast one (for people who know about neural nets, just not Flux) and a slow one (introducing linear regressing, parameters, etc). The slow one was called "Overview" before. 

The fast one is new. I'm not sure it's the optimal example, maybe MNIST can be done about as compactly? It aims to fit on one screen & show you all the main moving parts.

Re-reading "Overview", it's nicer than I remembered. I'm not entirely sure it should live in a different section to the next one, "Basics" (which I call "Gradients and Layers" here). So I moved the second "Basics" one into "Getting Started" too. (I do think they need better titles.)

I'm not precisely sure how this intersects with https://github.com/FluxML/Flux.jl/pull/2036, https://github.com/FluxML/Flux.jl/pull/2021, https://github.com/FluxML/Flux.jl/pull/2016. Those may replace text in some sections, to make the slow path even more detailed, or to upgrade the "Basics" section. Or maybe we want a tutorials section in the docs, which replaces the one in the website?


ToucheSir:
I like the proposed structure here. Could the "Fitting a Line" page mention "linear regression" somewhere to contextualize the problem and help with keyword searches? Haven't looked at the other docs changes yet.

mcabbott:
One unhappy feature is that we have a bit of a random mix of sections describing stuff in words, and sections which are more API listings. 

Maybe the ought to be marked somehow, something like this?

Screenshot 2022-09-27 at 10 02 01

Perhaps we also need to take a pass and move some of the more scattered docstrings from "descriptive" sections to "listing" sections.

docs/src/index.md
```diff
@@ -1,26 +1,31 @@
...
-* **Performance is key**. Flux integrates with high-performance AD tools such as [Zygote.jl](https://github.com/FluxML/Zygote.jl) for generating fast code. Flux optimizes both CPU and GPU performance. Scaling workloads easily to multiple GPUs can be done with the help of Julia's [GPU tooling](https://github.com/JuliaGPU/CUDA.jl) and projects like [DaggerFlux.jl](https://github.com/DhairyaLGandhi/DaggerFlux.jl).
-* **Play nicely with others**. Flux works well with Julia libraries from [data frames](https://github.com/JuliaComputing/JuliaDB.jl) and [images](https://github.com/JuliaImages/Images.jl) to [differential equation solvers](https://github.com/JuliaDiffEq/DifferentialEquations.jl), so you can easily build complex data processing pipelines that integrate Flux models.
+* **Play nicely with others**. Flux works well with Julia libraries from [images](https://github.com/JuliaImages/Images.jl) to [differential equation solvers](https://github.com/JuliaDiffEq/DifferentialEquations.jl), so you can easily build complex data processing pipelines that integrate Flux models.
```
mcabbott, line 7: In fact https://github.com/JuliaData/JuliaDB.jl is long-dead. Should anything else be highlighted? 

My vote is to make this intro shorter if possible, it's not all that helpful.

docs/src/models/basics.md
```diff
@@ -211,7 +211,7 @@ m = Chain(x -> x^2, x -> x+1)
 m(5) # => 26
 ```
 
-## Layer helpers
+## Layer Helpers
 
 Flux provides a set of helpers for custom layers, which you can enable by calling
```
mcabbott, line 216: In fact only one helper. Maybe this section made more sense long ago. Maybe it should include `create_bias` alla #2049?

docs/src/index.md
ToucheSir: Does it make any sense to `s/diffeq/sciml/g` for this point?

docs/src/assets/flux.css
```diff
@@ -100,10 +100,17 @@ article pre {
...
   border-radius: 10px 0px 0px 10px;
+  /* negative spacing removes indent of code within blocks at top level */
   margin-left: -1em;
   margin-right: -2em;
 }
 
+article .docstring pre {
+  /* cancel negative spacing inside a docstring, as it collides with outer box */
+  margin-left: 0em;
+  margin-right: 0em;
```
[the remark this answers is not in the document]
ToucheSir: I'm all for removing the negative spacing. At the very least, it should be opt-in instead of opt-out.

Come to think of it, is there a need for this stylesheet at all? Looking through some SciML package docs, any deviations from the default Documenter theme seem minimal and thus the maintenance overhead not very justifiable.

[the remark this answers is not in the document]
mcabbott: I do think it's nice if distinct packages look a little different, if only to help you remember which tab is which... here this seems to mean being a bit green. 

Would be better if all FluxML docs pulled the style sheet from one place though. ChainRules et al moved theirs to a package https://github.com/JuliaDiff/DocThemeIndigo.jl but I'm not volunteering to set that up. 

The one big visual change which would be nice would be to unify the website and the docs, like e.g. https://turing.ml/v0.21/docs/using-turing/quick-start . Maybe we can steal things from  https://github.com/TuringLang/turinglang.github.io

docs/src/index.md
mcabbott: Probably! Link does still work.

docs/src/assets/flux.css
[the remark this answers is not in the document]
mcabbott: [03eba13](https://github.com/FluxML/Flux.jl/pull/2069/commits/03eba13e44c72d74f6d43eeb0a733c1295eb3ad3) removes all negative spacing, looks fine.

docs/src/models/basics.md
mcabbott: Done in [98b4cd8](https://github.com/FluxML/Flux.jl/pull/2069/commits/98b4cd8fb72dfa27f4c2261a80c4c787365cefe3)

mcabbott:
Shall we do this? It's not done but seems like a step forwards. 

Biggish changes not mentioned in the top message are:
* The "advanced.md" page about custom layers is moved to the tutorials. It's a bit of a random selection of examples, and perhaps could be tidied up, but I didn't try. 
* New page for destructure, plus Flux.modules, as they didn't fit well elsewhere


I think the next steps after this are 
* to re-organise the training / regularization / optimisers docs. But these are coupled to the move away from implicit parameters. 
* to start moving other tutorials here from the website. Coupled to work on the website.

mcabbott:
> Adding 📚 to mark sections which contain API listings.

An alternative here would be to re-order sections into Guide / Reference. Maybe better? But right now the division isn't quite clean enough, IMO... the optimisers / training sections especially are a bit tangled. 

Screenshot 2022-10-10 at 20 27 26

Saransh-cpp (approved):
Thank you for this, @mcabbott! I do like the idea of separating out the API reference from the guide, but this can be taken up in another PR (if this PR is getting too large). Some minor comments below -

mcabbott:
I think the hurdles to separating Guide / Reference are
* that the order of reference starts to seem completely random, is there a better way? maybe moving all the "foreign" pages to "Reference, II" section would help?
* that quite a few sections currently are half in each camp, training / optimisers etc. But these will change for 0.14 anyway.

Issue #821, opened 2019-07-28
Optimizer handling of infinite loss


ChrisRackauckas:
Ref https://github.com/JuliaDiffEq/DiffEqFlux.jl/issues/70 . As discussed at JuliaCon, nicer strategies can be used so that if you get an infinite loss the training won't just explode. You can discard the point and take another random draw and things like that. @oxinabox @pkofod probably know more, since I know Optim.jl and other libraries handle this.

(@freemin7)

oxinabox:
Just dropping it feels like a risky move, but I think it might actually be the right one.
under SGD random minibatch is just sampling the loss space.
If most of it is not a pole then we can just sample again

There are loss functions that have a degenerate case of `-Inf` when you are doing very well,
(e.g. anything given by maximizing an expectated value parameter, and dividing by a dispersion parameter.)

OTOH, this is a bit risky since this could also be a result of a user mistake in the loss function definition.

MikeInnes:
Dropping everything on `Inf` seems pretty expensive (we'd have to check for the presence of `Inf` across all gradients). But just treating `Inf` as `0` would be easy to do.

oxinabox:
Not gradients, losses.
Different.

MikeInnes:
If it's really losses you're concerned about, that's independent of the optimisers, which only ever see gradients. It's less clear what the action item is here since normally it's up to the user to decide what to do with a loss. We could perhaps add a check in `train!`, but that'd only apply if you used that, and this seems like a relatively advanced use case.

freemin7:
I used DiffEqFlux to estimate parameters of an differential equation whose solution has a finite escape in some areas of the parameter space.
This means 3 things:
 - any good ODE solver will fail to integrate the solution past the singularity in the solution 
 - the only knowledge gained by this result is: This value is bad and there is nothing that can be done.
 - The difference between training data and solution is infinite in some areas and not defined in others.

So even from a mathematical point of view there is no well behaved way to handle this. If i currently return an infinity in my loss function DiffEqFlux (and the underlying machinery) throws an error halting the training.

It would be really helpful if there was way to tell Flux to ignore this loss and possibly stay away from this area (optional but nice to have in my case where bad parameters form a region). I am not here to debate whether this should be a +Inf or a NaN or something similar.

If i had to give an action item it would be:
 - Give the  basic control over the optimizer from the loss function.

My proposal would look like this:
[...]

PR #2070, opened 2022-09-27
Stop training on Inf/NaN loss

### LINKED ISSUE #1981: Warn on `NaN` loss
Would it not be convenient, particularily for users with less experienced with artificial neural networks, to print a warning when trying to update parameters with a loss of NaN? This could save a lot of wasted resources.

mcabbott:
Closes #1981, by altering `train!` so that when it encounters an infinite / NaN loss, it ~~prints a warning~~ throws an error, and stops. It stops before updating the model, because such an update will usually make everything NaN.

Not sure it should stop, in fact. Maybe it should skip that datapoint & continue?

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/test/optimise.jl
+++ b/test/optimise.jl
@@ -91,7 +91,7 @@ end
   m = Dense(1 => 1)
   m.weight .= 0
   CNT = 0
-  Flux.train!(Flux.params(m), 1:100, Descent(0.1)) do i
+  @test_throws DomainError Flux.train!(Flux.params(m), 1:100, Descent(0.1)) do i
     CNT += 1
     (i == 51 ? NaN32 : 1f0) * sum(m([1.0]))
   end
--- a/test/optimise.jl
+++ b/test/optimise.jl
@@ -87,6 +87,18 @@ end
   Flux.train!(loss, Flux.params(r), (r,), Descent())
 end
 
+@testset "Stop on NaN" begin
+  m = Dense(1 => 1)
+  m.weight .= 0
+  CNT = 0
+  Flux.train!(Flux.params(m), 1:100, Descent(0.1)) do i
+    CNT += 1
+    i > 50 ? NaN32 : sum(m([1.0]))
+  end
+  @test m.weight[1] ≈ -5  # did not corrupt weights
+  @test CNT == 51  # stopped early
+end
+
 @testset "ExpDecay" begin
 
   @testset "Sanity Check" begin
--- a/test/optimise.jl
+++ b/test/optimise.jl
@@ -93,7 +93,7 @@ end
   CNT = 0
   Flux.train!(Flux.params(m), 1:100, Descent(0.1)) do i
     CNT += 1
-    i == 51 ? NaN32 : sum(m([1f0]))
+    (i == 51 ? NaN32 : 1f0) * sum(m([1.0]))
   end
   @test CNT == 51  # stopped early
   @test m.weight[1] ≈ -5  # did not corrupt weights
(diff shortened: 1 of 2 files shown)


test/optimise.jl
```diff
@@ -87,6 +87,18 @@ end
...
+@testset "Stop on NaN" begin
+  m = Dense(1 => 1)
+  m.weight .= 0
+  CNT = 0
+  Flux.train!(Flux.params(m), 1:100, Descent(0.1)) do i
+    CNT += 1
+    i == 51 ? NaN32 : sum(m([1f0]))
+  end
+  @test CNT == 51  # stopped early
+  @test m.weight[1] ≈ -5  # did not corrupt weights
```
mcabbott, line 99: Somewhat surprisingly, current Flux doesn't actually produce NaN weights, but I don't know where it skips this update:
```
julia> let
       m = Dense(1 => 1)
                m.weight .= 0
                CNT = 0
                Flux.train!(Flux.params(m), 1:100, Descent(0.1)) do i
                  CNT += 1
                  i == 51 ? NaN32 : sum(m([1.0]))
                end
                m.weight, CNT
       end
(Float32[-9.900002;;], 100)
```

mcabbott: My fault, that iteration just has zero gradient. A more realistic one would be something like this:
```
julia> let
       m = Dense(1 => 1)
                m.weight .= 0
                CNT = 0
                Flux.train!(Flux.params(m), 1:100, Descent(0.1)) do i
                  CNT += 1
                  (i == 51 ? NaN32 : 1f0) * sum(m([1.0]))
                end
                m.weight, CNT
       end
(Float32[NaN;;], 100)
```

ToucheSir:
It would be good to have some way for user code to tell training has stopped prematurely. Not sure what that would look like: return value, callback, flag, etc?

mcabbott:
One way would be to upgrade this to an error. It does depend a little on how complex/comprehensive we think `train!` ought to be.

I found https://github.com/FluxML/Flux.jl/issues/821, in which all options were discussed. And the idea was to write `isfinite(l) || Flux.skip()`, except that this got documented as something to do in a callback, which never worked, it happens too late... and nobody noticed because this was sufficiently confusing? So my opinion is that anything that complicated should be an ordinary `for` loop.

ToucheSir:
Raising `StopException` from the inner loop doesn't seem like the worst idea. Unlike calling `skip` or `stop` in callbacks, this would be able to intercept things in time. I guess the only question is whether making `StopException` part of the public API for this purpose is palatable.

mcabbott:
`StopException` leads directly to `break`, so I think there's nothing gained by implementing an auto-check that way compared to this PR's current code?

The point of the exception was, I now see, was that you could throw it from inside the loss function passed to gradient, and still get to `break`. But documenting it as part of the callback was a mistake.

ToucheSir:
I missed that `StopException` was still being caught by the inner loop of `train!`. In that case, it'd have to be another exception type. Whether adding one just for this makes sense is a good question.

mcabbott:
Ah OK, so you weren't proposing to throw and catch an error. It could just throw a DomainError, that seems the closest Base type.

ToucheSir (approved):
LGTM, but if @darsnack or @CarloLucibello wouldn't mind having a once-over that would be great.

KronosTheLate:
Thanks for taking the issue I raised so seriously, it is so nice to encounter active developers who listen to user requests. Thanks for your work <3

PR #2072, opened 2022-09-30
Export `Embedding`

mcognetta:
Exporting `Embedding` from `layers`.


ToucheSir (approved):
I see no reason not to do this unless _maybe_ we want to figure out https://github.com/FluxML/Flux.jl/pull/1656 first? Will merge in a couple of days if there are no objections.

CarloLucibello:
I think this is fine, Transformers.jl does not export `Embedding` so hopefully we won't generate conflicts (although it does export `Embed`).

PR #2073, opened 2022-10-02
Relax `RNN`/`LSTM`/`GRUCell` internal matrix type restrictions

mcognetta:
Recurrent cells currently require that the internal matrices `W_h`, `W_i`, etc. be exactly the same type. This is unnecessarily restrictive. This PR relaxes the type requirements, so that the internal weight matrices can be of any type.

I will briefly motivate this change here. The most obvious case that this would be used in is when dealing with sparse internal matrices. If `W_h` is sparse, `W_i` must also be a sparse type, but it is possible that `W_i` would be extremely dense (basically a dense embedding matrix) and then the performance would suffer greatly. However, there is another case that has come up in my work where this restriction basically doubles the parameter count of a model unnecessarily.

Suppose I have a model:

```julia
m = Chain(Embedding(3=>5), RNN(5=>5))
```

The embedding weight matrix is `5x3` but the `RNNCell` *also* contains a `5x5` matrix (`W_i`) designed to project the input vector into the hidden embedding space. However, since the recurrent definition is something like `h = σ.(Wi*x .+ Wh*h .+ b)`, there is no non-linearity between `Embedding(x)` and the multiplication with `W_i`, so this is just a linear transformation. In other words, having both of these matrices is redundant. 

```julia
julia> a = rand(5, 5);

julia> e = Flux.Embedding(3=>5)
Embedding(3 => 5)   # 15 parameters

julia> z = a*e.weight;

julia> r1 = Flux.Recur(Flux.RNNCell(tanh, a, rand(5, 5), zeros32(5), zeros32(5,1)))
Recur(
  RNNCell(5 => 5, tanh),                # 60 parameters
)         # Total: 4 trainable arrays, 60 parameters,
          # plus 1 non-trainable, 5 parameters, summarysize 672 bytes.

julia> r2 = Flux.Recur(Flux.RNNCell(tanh, z, rand(5, 5), zeros32(5), zeros32(5,1)))
Recur(
  RNNCell(3 => 5, tanh),                # 50 parameters
)         # Total: 4 trainable arrays, 50 parameters,
          # plus 1 non-trainable, 5 parameters, summarysize 592 bytes.

julia> r1(e(1))
5-element Vector{Float64}:
 -0.935882363105291
 -0.13412698209804408
 -0.9471803427440432
 -0.4815452514782611
  0.049400737502303525

julia> r2(Float32[1, 0, 0])
5-element Vector{Float64}:
 -0.935882363105291
 -0.13412698209804408
 -0.9471803427440432
 -0.4815452514782611
  0.049400737502303525
```

In this case, after the type relaxation, we could instead have a model like `Flux.Recur(Flux.RNNCell(tanh, rand(5, 5), Identity(5), rand(5), rand(5, 1))` and it would work just as well. 

### TODO
- [ ] Add docs for this to the other cell types (currently, I only added it for `RNNCell`).

### PR Checklist
- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -189,10 +189,10 @@ end
 
 # Vanilla RNN
 
-struct RNNCell{F,A,V,S}
+struct RNNCell{F,I,H,V,S}
   σ::F
-  Wi::A
-  Wh::A
+  Wi::I
+  Wh::H
   b::V
   state0::S
 end
@@ -200,7 +200,7 @@ end
 RNNCell((in, out)::Pair, σ=tanh; init=Flux.glorot_uniform, initb=zeros32, init_state=zeros32) =
   RNNCell(σ, init(out, in), init(out, out), initb(out), init_state(out,1))
 
-function (m::RNNCell{F,A,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {F,A,V,T}
+function (m::RNNCell{F,I,H,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {F,I,H,V,T}
   Wi, Wh, b = m.Wi, m.Wh, m.b
   σ = NNlib.fast_act(m.σ, x)
   h = σ.(Wi*x .+ Wh*h .+ b)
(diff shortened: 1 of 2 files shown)


ToucheSir (approved):
Makes sense, thanks!

PR #2074, opened 2022-10-02
Add GH action format check files with `JuliaFormatter`

Saransh-cpp:
Most of the Julia packages I have seen lack a proper style guide for their codebase. I feel `JuliaFormatter` coupled with a clever workflow can help us maintain a clean codebase. This PR adds a bot-like workflow that fixes the formatting issues associated with a PR and pushes the fixed files in the same branch.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,9 @@
+style = "sciml"
+whitespace_in_kwargs = true
+format_docstrings = true
+always_for_in = true
+join_lines_based_on_source = true
+separate_kwargs_with_semicolon = false
+always_use_return = true
+margin = 92
+indent = 4
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# activated JuliaFormatter
+bf30ed78e89f9a652d885d1848a1d41f414dffc8
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,62 +1,67 @@
-using Documenter, Flux, NNlib, Functors, MLUtils, BSON, Optimisers, OneHotArrays, Zygote, ChainRulesCore
-
+using Documenter, Flux, NNlib, Functors, MLUtils, BSON, Optimisers, OneHotArrays, Zygote,
+      ChainRulesCore
 
 DocMeta.setdocmeta!(Flux, :DocTestSetup, :(using Flux); recursive = true)
 
-makedocs(
-    modules = [Flux, NNlib, Functors, MLUtils, BSON, Optimisers, OneHotArrays, Zygote, ChainRulesCore, Base],
-    doctest = false,
-    sitename = "Flux",
-    # strict = [:cross_references,],
-    pages = [
-        "Getting Started" => [
-            "Welcome" => "index.md",
-            "Quick Start" => "models/quickstart.md",
-            "Fitting a Line" => "models/overview.md",
-            "Gradients and Layers" => "models/basics.md",
-        ],
-        "Building Models" => [
-            "Built-in Layers 📚" => "models/layers.md",
-            "Recurrence" => "models/recurrence.md",
-            "Activation Functions 📚" => "models/activation.md",
-            "NNlib.jl 📚 (`softmax`, `conv`, ...)" => "models/nnlib.md",
+makedocs(modules = [
+             Flux,
+             NNlib,
+             Functors,
+             MLUtils,
+             BSON,
+             Optimisers,
+             OneHotArrays,
+             Zygote,
+             ChainRulesCore,
+             Base,
          ],
-         "Handling Data" => [
-             "MLUtils.jl 📚 (`DataLoader`, ...)" => "data/mlutils.md",
-             "OneHotArrays.jl 📚 (`onehot`, ...)" => "data/onehot.md",
+         doctest = false,
+         sitename = "Flux",
+         # strict = [:cross_references,],
+         pages = [
+             "Getting Started" => [
+                 "Welcome" => "index.md",
+                 "Quick Start" => "models/quickstart.md",
+                 "Fitting a Line" => "models/overview.md",
+                 "Gradients and Layers" => "models/basics.md",
+             ],
+             "Building Models" => [
+                 "Built-in Layers 📚" => "models/layers.md",
+                 "Recurrence" => "models/recurrence.md",
+                 "Activation Functions 📚" => "models/activation.md",
+                 "NNlib.jl 📚 (`softmax`, `conv`, ...)" => "models/nnlib.md",
+             ],
+             "Handling Data" => [
+                 "MLUtils.jl 📚 (`DataLoader`, ...)" => "data/mlutils.md",
+                 "OneHotArrays.jl 📚 (`onehot`, ...)" => "data/onehot.md",
+             ],
+             "Training Models" => [
+                 "Training" => "training/training.md",
+                 "Regularisation" => "models/regularisation.md",
+                 "Loss Functions 📚" => "models/losses.md",
+                 "Optimisation Rules 📚" => "training/optimisers.md",  # TODO move optimiser intro up to Training
+                 "Callback Helpers 📚" => "training/callbacks.md",
+                 "Zygote.jl 📚 (`gradient`, ...)" => "training/zygote.md",
+             ],
+             "Model Tools" => [
+                 "GPU Support" => "gpu.md",
+                 "Saving & Loading" => "saving.md",
+                 "Shape Inference 📚" => "outputsize.md",
+                 "Weight Initialisation 📚" => "utilities.md",
+                 "Flat vs. Nested 📚" => "destructure.md",
+                 "Functors.jl 📚 (`fmap`, ...)" => "models/functors.md",
+             ],
+             "Performance Tips" => "performance.md",
+             "Flux's Ecosystem" => "ecosystem.md",
+             "Tutorials" => [  # TODO, maybe
+                 "Custom Layers" => "models/advanced.md",  # TODO move freezing to Training
+             ],
          ],
-         "Training Models" => [
-             "Training" => "training/training.md",
-             "Regularisation" => "models/regularisation.md",
-             "Loss Functions 📚" => "models/losses.md",
-             "Optimisation Rules 📚" => "training/optimisers.md",  # TODO move optimiser intro up to Training
-             "Callback Helpers 📚" => "training/callbacks.md",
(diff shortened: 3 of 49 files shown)


ToucheSir:
Perhaps the title would be better with `s/lint/(auto)format/`? That is, unless you're planning on adding some actions for static analysis tools as well.

On formatting, we should use the same style across all FluxML repos. Just copying https://github.com/FluxML/Metalhead.jl/blob/master/.JuliaFormatter.toml should be enough, I think.

Saransh-cpp:
> Perhaps the title would be better with s/lint/(auto)format/? That is, unless you're planning on adding some actions for static analysis tools as well.

Are you referring to the PRs title or commit message of the bot's commit? (Should it be - `autoformat: format jl files with JuliaFormatter`?)

> On formatting, we should use the same style across all FluxML repos. Just copying https://github.com/FluxML/Metalhead.jl/blob/master/.JuliaFormatter.toml should be enough, I think.

Thanks! I will add this in!

Saransh-cpp:
This works now!

Working on a PR created from the base repository - https://github.com/Saransh-cpp/Flux.jl/pull/53
Working on a PR created from a fork - https://github.com/Saransh-cpp/Flux.jl/pull/55

---

### Note
1. This will work once this PR is merged (cannot test it before due to security reasons, explained below).
2. To accomplish this, the repository secrets must contain a personal access token (PAT) of an account with write access to FluxML. The token should be added to repository secrets with the name - "AUTO_LINT_PAT".

---

### Why is PAT required?
A bot commit without a PAT does not trigger workflows. This means that the reformatted files will not go through the conventional CI, which can potentially be dangerous. Let's say a bug in `JuliaFormatter` rewrites a file with some broken syntax -> this broken formatting update will not be detected by the CI as the CI will not run on the bot commit.

---

### Is this a security concern?
Since this workflow runs on pull_request_target, the PAT itself cannot be leaked. I have successfully tested this here -

https://github.com/Saransh-cpp/Flux.jl/pull/55 (see the files changed in the last commit - I tried printing out the PAT that, in a sense, is an attempt to hack FluxML's private tokens) -> https://github.com/Saransh-cpp/Flux.jl/actions/runs/3175153155/jobs/5172838261 (see how the "HACK" step was not executed at all).

This shows that GitHub will run the "Lint" workflow according to the instructions written in the base `FluxML` repository and not the fork repository.

Due to this reason, we cannot test this action on this PR (as `pull_request_merged` does not run the workflow using the instructions written in a fork). For more reference - https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

> Having said that, mixing pull_request_target with an explicit PR checkout is not always vulnerable. The workflow may, for example:
>
> - Reformat and commit the code
> - Checkout both base and head repositories and generate a diff
> - Run grep on the checked out source.
> 
> Generally speaking, when the PR contents are treated as passive data, i.e. not in a position of influence over the build/testing process, it is safe. But the repository owners must be extra careful not to trigger any script that may operate on PR controlled contents like in the case of npm install.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,35 @@
...
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
+jobs:
+  JuliaFormatter:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
```
DilumAluthge, line 17: > Combining pull_request_target workflow trigger with an explicit checkout of an untrusted PR is a dangerous practice that may lead to repository compromise.

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

DilumAluthge: You already quoted this section above:

> But the repository owners must be extra careful not to trigger any script that may operate on PR controlled contents like in the case of npm install.

This is the problem. Combining `pull_request_target` with a checkout is a huge footgun. Can it be done safely? Potentially, yes. But it is so easy to make a mistake here.

DilumAluthge: My recommendation would be to not go with this approach, because it is too easy to end up compromising the repository.

Instead, if the goal is to enforce a common format across all repositories in the FluxML organization, I would recommend:

1. Have a CI job that checks if the source code is formatted correctly, and errors if any formatting problems are detected.
2. Use the `needs` construct to force all other CI jobs to wait for the successful completion of the formatting check.
3. If the formatting check does not succeed, do not run any other CI jobs.

DilumAluthge: Actually, in my above comment, change "potentially compromise" to "compromise". This PR already compromises the repository.

I won't point out the line that introduces the compromise. Does anyone see what it is?

Saransh-cpp: Thanks, @DilumAluthge! Good thing I cross-posted this in `ci-dev`. I agree, this workflow is potentially risky. If someone merges a PR that changes the workflows then Flux would be at a security risk. Leaving out the PAT completely is equally risky as that would allow the GitHub actions bot to commit anything without the CI running on it.

> This PR already compromises the repository.

Could you let me know? Is it that one can write a script to leak the PAT in another workflow? I tried this in one of my PRs, but GitHub does not allow the user to access the PAT - 

```
      - name: POTENTIAL HACK
         run: echo ${{ secrets.PAT }}
```

Screenshot 2022-10-04 at 7 49 00 PM

ToucheSir:
After another read-through, I agree with @DilumAluthge's concerns and raise one of my own. If the bot is appending formatting commits to PR branches, how are contributors meant to continue working on their feature branches from another checkout without running into conflicts? https://github.com/FluxML/Flux.jl/pull/2074#discussion_r986453790 seems like a more reasonable approach and is already used in projects like https://github.com/JuliaDiff/ChainRulesCore.jl/blob/main/.github/workflows/format.yml. To ensure code is formatted in the remote branch, we could also consider a pre-commit hook.

> Are you referring to the PR's title or the commit message of the bot's commit?

Both. Your proposed rename looks fine.

Saransh-cpp:
A `pre-commit` hook would ideally solve all of the problems. I will look into how one develops that and how pre-commit.ci can support `JuliaFormatter`. For the time being, I think the suggestion of adding an action that fails on wrong formatting sounds best.

Saransh-cpp:
> is already used in projects like https://github.com/JuliaDiff/ChainRulesCore.jl/blob/main/.github/workflows/format.yml.

`ChainRulesCore.jl` tries to do something similar but by using an existing tool - [reviewdog](https://github.com/marketplace/actions/reviewdog-suggester-suggest-any-code-changes-based-on-diff-with-reviewdog). As expected, review dog requires a PAT too - see [this](https://github.com/JuliaDiff/ChainRulesCore.jl/actions/runs/2923309119/jobs/4661741155) CI run -
```
reviewdog: This GitHub token doesn't have write permission of Review API [1], 
so reviewdog will report results via logging command [2] and create annotations similar to
github-pr-check reporter as a fallback.
[1]: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target, 
[2]: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#logging-commands
```

But printing the `diff` in the CI is useful too. I think I will use `review dog` without the commenting thing for `FluxML` too. (@DilumAluthge could you confirm if it is safe to pass in PAT in the `review dog` workflow?)

mcabbott:
FWIW the formatter on ChainRulesCore.jl seems more irritating than helpful. It clutters actual review with lots of comments telling you that `[1,2,3]` is wrong. Code by different authors disagreeing on minor things like that seems far down the list of style problems -- e.g. there are some easy-to-read straightforward tests, and some where it's a really clever puzzle to figure out what the test is doing.

An ideal auto-formatter that results in *less* time spent thinking about spacing might be nice. But we spend very little. And a half-baked one which creates *more* fiddling seems much worse than where we are now. 

Other concerns here are:
* If we adopt 4 spaces, has someone figured out how to preserve git blame, including on the website?
* Doing so will mean every open PR needs a pretty big rebase. I presume these would be smaller if it's set to two spaces.
* Note that it edits examples to e.g. write `m = Upsample(; scale = (2, 3))`, I'm not sure we want that as the documented thing to type in.
* It also seems to want a pretty short line length, e.g. [this](https://github.com/FluxML/Flux.jl/pull/2074/files#diff-08261c6ab5a9b75d3f9bca6a6f93b32fefe4e22e91ffcfdfd8d88dad08c3a498R123-R152) is a lot of lines for a simple thing.

ToucheSir:
IMO the most important part of an autoformatter is that it enforces `format(file)` is the identity operation for all checked-in source files. Especially on FluxML repos, it can get quite frustrating to make a small change, then spend 2-3 minutes reverting or ignoring parts of the diff because you accidentally (or intentionally) bumped the format shortcut (`git add -p` helps but does not solve the problem).

> If we adopt 4 spaces, has someone figured out how to preserve git blame, including on the website?

GitHub [supports](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) using `.git-blame-ignore-revs` for this. Metalhead uses this post https://github.com/FluxML/Metalhead.jl/pull/168.

Saransh-cpp:
TODO
- [ ] add `.git-blame-ignore-revs`
- [ ] discuss formatting options

Saransh-cpp:
Looks better now. Should the formatting options be discussed in the next community call? Here is the current configuration -

```toml
style = "sciml"
whitespace_in_kwargs = true
format_docstrings = true
always_for_in = true
join_lines_based_on_source = true
separate_kwargs_with_semicolon = true
always_use_return = true
margin = 92
indent = 4
```

Saransh-cpp:
`separate_kwargs_with_semicolon` is `false` now. See #2086.

ToucheSir:
I think it would be better if we didn't have to rely on pre-commit (which is a heavy dependency and requires a Python interpreter). A solution which works with pure Julia would be better and should be doable, someone just has to write it.

The other blocker is the JuliaFormatter latency from the command line is really only acceptable on 1.9+. If the binary distribution story can be figured out before 1.9 becomes ubiquitous, then we can consider that route instead.

Saransh-cpp:
I hadn't factored in adding pre-commit as a dependency.

Trying out JuliaFormatter after 1.9 is out of Beta sounds great. I think it would be better to create a new PR when 1.9 is out. Resolving conflicts in this PR is getting more and more difficult with every update in Flux's codebase 😅

PR #2078, opened 2022-10-05
Add `@autosize`

mcabbott:
The macro replaces `_` with an inferred input size:
```julia
julia> @autosize (2,3,4) Chain(Dense(_ => 5, tanh), Flux.flatten, Maxout(() -> Dense(_ => 3), 2))
Chain(
  Dense(2 => 5, tanh),                  # 15 parameters
  Flux.flatten,
  Maxout(
    Dense(15 => 3),                     # 48 parameters
    Dense(15 => 3),                     # 48 parameters
  ),
)                   # Total: 6 arrays, 111 parameters, 892 bytes.

julia> ans(rand(2,3,4)) |> size
(3, 4)
```

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/docs/src/outputsize.md
+++ b/docs/src/outputsize.md
@@ -27,7 +27,7 @@ The function `outputsize` works by passing a "dummy" array into the model, which
 It should work for all layers, including custom layers, out of the box.
 
 An example of how to automate model building is this:
-```jldoctest; output = false
+```jldoctest; output = false, setup = :(using Flux)
 """
     make_model(width, height, [inchannels, nclasses; layer_config])
 
--- a/docs/src/outputsize.md
+++ b/docs/src/outputsize.md
@@ -27,7 +27,7 @@ The function `outputsize` works by passing a "dummy" array into the model, which
 It should work for all layers, including custom layers, out of the box.
 
 An example of how to automate model building is this:
-```julia
+```jldoctest; output = false
 """
     make_model(width, height, [inchannels, nclasses; layer_config])
 
@@ -57,6 +57,10 @@ end
 m = make_model(28, 28, 3, layer_config = [9, 17, 33, 65])
 
 Flux.outputsize(m, (28, 28, 3, 42)) == (10, 42) == size(m(randn(Float32, 28, 28, 3, 42)))
+
+# output
+
+true
 ```
 
 Alternatively, using the macro, the definition of `make_model` could end with:
--- a/docs/src/outputsize.md
+++ b/docs/src/outputsize.md
@@ -5,17 +5,18 @@ of arrays that layers will recieve, without doing any computation.
 This is especially useful for convolutional models, where the same [`Conv`](@ref) layer
 accepts any size of image, but the next layer may not. 
 
-The higher-level one is a macro [`@autosize`](@ref) which acts on the code defining the layers,
-and replaces each appearance of `_` with the relevant size. A simple example might be:
+The higher-level tool is a macro [`@autosize`](@ref) which acts on the code defining the layers,
+and replaces each appearance of `_` with the relevant size. This simple example returns a model
+with `Dense(845 => 10)` as the last layer:
 
 ```julia
 @autosize (28, 28, 1, 32) Chain(Conv((3, 3), _ => 5, relu, stride=2), Flux.flatten, Dense(_ => 10))
 ```
 
-The size may be provided at runtime, like `@autosize (sz..., 1, 32) Chain(Conv(`..., but all the
+The input size may be provided at runtime, like `@autosize (sz..., 1, 32) Chain(Conv(`..., but all the
 layer constructors containing `_` must be explicitly written out -- the macro sees the code as written.
 
-This relies on a lower-level function [`outputsize`](@ref Flux.outputsize), which you can also use directly:
+This macro relies on a lower-level function [`outputsize`](@ref Flux.outputsize), which you can also use directly:
 
 ```julia
 c = Conv((3, 3), 1 => 5, relu, stride=2)
@@ -53,7 +54,9 @@ function make_model(width, height, inchannels = 1, nclasses = 10;
   return Chain(conv_layers..., Flux.flatten, last_layer)
 end
 
-make_model(28, 28, 3, layer_config = [8, 17, 33, 65])
+m = make_model(28, 28, 3, layer_config = [9, 17, 33, 65])
+
+Flux.outputsize(m, (28, 28, 3, 42)) == (10, 42) == size(m(randn(Float32, 28, 28, 3, 42)))
 ```
 
 Alternatively, using the macro, the definition of `make_model` could end with:
--- a/docs/src/outputsize.md
+++ b/docs/src/outputsize.md
@@ -1,47 +1,72 @@
 # Shape Inference
 
-To help you generate models in an automated fashion, [`Flux.outputsize`](@ref) lets you 
-calculate the size returned produced by layers for a given size input.
-This is especially useful for layers like [`Conv`](@ref).
+Flux has some tools to help generate models in an automated fashion, by inferring the size
+of arrays that layers will recieve, without doing any computation. 
+This is especially useful for convolutional models, where the same [`Conv`](@ref) layer
+accepts any size of image, but the next layer may not. 
 
-It works by passing a "dummy" array into the model that preserves size information without running any computation.
-`outputsize(f, inputsize)` works for all layers (including custom layers) out of the box.
-By default, `inputsize` expects the batch dimension,
-but you can exclude the batch size with `outputsize(f, inputsize; padbatch=true)` (assuming it to be one).
+The higher-level one is a macro [`@autosize`](@ref) which acts on the code defining the layers,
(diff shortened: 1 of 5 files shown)


src/outputsize.jl
```diff
@@ -168,3 +172,165 @@ for (fn, Dims) in ((:conv, DenseConvDims),)
...
+@functor LazyLayer
+
+function striplazy(x)
+  fs, re = functor(x)
+  re(map(striplazy, fs))
+end
+striplazy(l::LazyLayer) = l.layer == nothing ? error("should be initialised!") : l.layer
+
+# Could make LazyLayer usable outside of @autosize, for instance allow Chain(@lazy Dense(_ => 2))?
+# But then it will survive to produce weird structural gradients etc.
```
ToucheSir, line 324: Could we force users to call `recursive_striplazy(model, input_size)` or something before using an incrementally constructed network like this? Maybe define a `rrule` which throws an error?

mcabbott: `striplazy` should be fully recursive. We could make a function that calls this after `outputsize` & returns the model. And indeed an `rrule` would be one way to forbid you not to strip the model before using it for real. 

I suppose the other policy would just be to allow these things to survive in the model. As long as you never change it, and don't care about the cost of the `if` & type instability, it should work? 

But any use outside of `@autosize` probably needs another macro... writing `Flux.LazyLayer("", x -> Dense(size(x,1) => 10), nothing)` seems sufficiently obscure that perhaps it's OK to say that's obviously at own risk, for now? `@autosize` can be the only API until we decide if we want more.

ToucheSir: I agree. Recursive lazy layer removal + explicitly using `LazyLayer` for incremental construction seems more than enough for now.

docs/src/outputsize.md
```diff
@@ -1,47 +1,72 @@
...
+end
+
+make_model(28, 28, 3, layer_config = [8, 17, 33, 65])
+```
+
+Alternatively, using the macro, the definition of `make_model` could end with:
 
-  # the input dimension to Dense is programatically calculated from
-  #  width, height, and nchannels
-  return Chain(conv_layers..., Dense(prod(conv_outsize) => nclasses))
```
mcabbott, line 41: I edited this example a bit, partly to make it shorter, but also to correct two bugs -- a wrong variable name, and a missing `flatten` step. With the old code:
```julia
julia> function make_model(width, height, inchannels, nclasses;
                           layer_config = [16, 16, 32, 32, 64, 64])
         # construct a vector of conv layers programmatically
         conv_layers = [Conv((3, 3), inchannels => layer_config[1])]
         for (infilters, outfilters) in zip(layer_config, layer_config[2:end])
           push!(conv_layers, Conv((3, 3), infilters => outfilters))
         end

         # compute the output dimensions for the conv layers
         # use padbatch=true to set the batch dimension to 1
         conv_outsize = Flux.outputsize(conv_layers, (width, height, nchannels); padbatch=true)

         # the input dimension to Dense is programatically calculated from
         #  width, height, and nchannels
         return Chain(conv_layers..., Dense(prod(conv_outsize) => nclasses))
       end
make_model (generic function with 3 methods)

julia> m = make_model(28, 28, 1, 10)
ERROR: UndefVarError: nchannels not defined
Stacktrace:
 [1] make_model(width::Int64, height::Int64, inchannels::Int64, nclasses::Int64; layer_config::Vector{Int64})
   @ Main ./REPL[216]:11
 [2] make_model(width::Int64, height::Int64, inchannels::Int64, nclasses::Int64)
   @ Main ./REPL[216]:1
 [3] top-level scope
   @ REPL[217]:1
 [4] top-level scope
   @ ~/.julia/packages/CUDA/DfvRa/src/initialization.jl:52

julia> nchannels = 1
1

julia> m = make_model(28, 28, 1, 10)
Chain(
  Conv((3, 3), 1 => 16),                # 160 parameters
  Conv((3, 3), 16 => 16),               # 2_320 parameters
  Conv((3, 3), 16 => 32),               # 4_640 parameters
  Conv((3, 3), 32 => 32),               # 9_248 parameters
  Conv((3, 3), 32 => 64),               # 18_496 parameters
  Conv((3, 3), 64 => 64),               # 36_928 parameters
  Dense(16384 => 10),                   # 163_850 parameters
)                   # Total: 14 arrays, 235_642 parameters, 922.711 KiB.

julia> Flux.outputsize(m, (28, 28, 1, 32))
┌ Error: layer Dense(16384 => 10), index 7 in Chain, gave an error with input of size (16, 16, 64, 32)
└ @ Flux ~/.julia/dev/Flux/src/outputsize.jl:107
ERROR: DimensionMismatch: matrix A has dimensions (10,16384), matrix B has dimensions (16,32768)
Stacktrace:
```

If not a `jldoctest`, would be good to ensure the code at least runs... does documenter have an easy way to do this?

src/outputsize.jl
```diff
@@ -168,3 +172,162 @@ for (fn, Dims) in ((:conv, DenseConvDims),)
...
+# But then it will survive to produce weird structural gradients etc. 
+
+function ChainRulesCore.rrule(l::LazyLayer, x)
+  l(x), _ -> error("LazyLayer should never be used within a gradient. Call striplazy(model) first to remove all.")
+end
+function ChainRulesCore.rrule(::typeof(striplazy), m)
+  striplazy(m), _ -> error("striplazy should never be used within a gradient")
+end
+
+params!(p::Params, x::LazyLayer, seen = IdSet()) = error("LazyLayer should never be used within params(m). Call striplazy(m) first.")
```
mcabbott, line 322: I made some errors to prevent the use of this LazyLayer within a real model. So far the macro is the only API, and it should always remove it before returning.

docs/src/outputsize.md
ToucheSir: Looks like it does: https://juliadocs.github.io/Documenter.jl/stable/man/doctests/

mcabbott: The remaining CI failure is from this code block, with "ERROR: UndefVarError: SamePad not defined".

That's pretty odd, as the use `using Flux` everywhere else.

ToucheSir: Is there isolation on the code block or page level? I don't see any imports in that markdown file.

mcabbott: I thought this line https://github.com/FluxML/Flux.jl/blob/master/docs/make.jl#L4 was what made every other doctest work. But perhaps it matters that they are all in source files, not in docs?

mcabbott: Then I read the docs. That line only affects docstrings in the code. Now fixed.

ToucheSir (approved):

test/outputsize.jl
```diff
@@ -142,16 +142,103 @@ end
...
+  img = [28, 28];
+  m = @autosize (img..., 1, 32) Chain(              # size is only needed at runtime
+         Chain(c = Conv((3,3), _ => 5; stride=2, pad=SamePad()),
+               p = MeanPool((3,3)),
+               b = BatchNorm(_),
+               f = Flux.flatten),
+         Dense(_ => _÷4, relu, init=Flux.rand32),   # can calculate output size _÷4
+         SkipConnection(Dense(_ => _, relu), +),
+         Dense(_ => 10),
+      ) |> gpu                                      # moves to GPU after initialisation
```
ToucheSir, line 223: @mcabbott I missed this in the review, but GPU tests are failing because the `|>` here is binding more tightly than `@autosize` and thus the model isn't actually moved onto the GPU. Is there anything we can do about that other than adding more parens? Whatever changes made would affect the `@autosize` docstring as well, since it shares this example.

mcabbott: Oh no I thought I checked this, sorry. It's not working but the binding is what I expected:
```
julia> (@autosize (1,2) Dense(_, 3) |> f64).bias
3-element Vector{Float32}:
 0.0
 0.0
 0.0

julia> :(@autosize (1,2) Dense(_, 3) |> f64) |> dump
Expr
  head: Symbol macrocall
  args: Array{Any}((4,))
    1: Symbol @autosize
    2: LineNumberNode
      line: Int64 1
      file: Symbol REPL[16]
    3: Expr
      head: Symbol tuple
      args: Array{Any}((2,))
        1: Int64 1
        2: Int64 2
    4: Expr
      head: Symbol call
      args: Array{Any}((3,))
        1: Symbol |>
        2: Expr
          head: Symbol call
          args: Array{Any}((3,))
            1: Symbol Dense
            2: Symbol _
            3: Int64 3
        3: Symbol f64
```
It seems the `gpu` walk is taking place too early. Maybe I never implemented my scheme to delay it.

mcabbott: My scheme was I think aimed at `adapt`. There you can grab the function being mapped, and replace the layer's maker function with one composed with this. But for Functors.jl I think that's impossible. 

So we should just make it an error. And remove this use from the docs. Call `gpu` once it's done, i.e. with brackets.

ToucheSir: Sounds good to me.

Issue #1986, opened 2022-06-02
Deprecate Flux.Optimisers and implicit parameters in favour of Optimisers.jl and explicit parameters


CarloLucibello:
So all the things are in place and we can get rid of the current pattern using implicit params:

```julia
using Flux
ps = Flux.params(model)
opt = Flux.Optimise.ADAM()
gs = gradient(() -> loss(model(x), y), ps)
Flux.Optimise.update!(opt, ps, grads)
```
to the one using explicit parameters and [Optimisers.jl](https://github.com/FluxML/Optimisers.jl)
```julia
using Flux, Optimisers
opt_state = Optimisers.setup(Optimisers.Adam(), model)
∇model = gradient(m -> loss(m(x), y), model)[1]
opt_state, model = Optimisers.update!(opt_state, model, ∇model)
## or the non-mutating
# state, model = Optimisers.update(opt_state, model, ∇model)
```

### Code
- [ ] Upgrade or remove `train!(loss, ::Params, data, ::AbstractOptimiser)`:
- [ ] A replacement for iterating over `Flux.params(m)`... or just fix it, broken in #2054:
```
julia> gradient(m -> (sum(norm, Flux.params(m))), (x=[1,2.0], y=[3.0]))
(nothing,)
```

### Documentation
- [x] Rewrite https://fluxml.ai/Flux.jl/stable/training/optimisers/ -- done in https://github.com/FluxML/Flux.jl/pull/2114
- [ ] Better documentation for [Optimisers.jl](https://github.com/FluxML/Optimisers.jl)
  - [ ] working with custom models https://github.com/FluxML/Optimisers.jl/issues/84

### Examples
- [ ] Port model zoo examples -- [tag "update"](https://github.com/FluxML/model-zoo/issues?q=label%3Aupdate+explicit)
- [ ] Help porting downstream libraries and check there are no surprises
  - [ ] GraphNeuralNetworks.jl

@mcabbott @ToucheSir @darsnack feel free to add to this

ToucheSir:
Thoughts on copying or moving parts of https://github.com/FluxML/Flux.jl/blob/v0.13.3/docs/src/training/optimisers.md#scheduling-optimisers over to the Optimisers.jl docs?

darsnack:
Xref https://github.com/darsnack/ParameterSchedulers.jl/issues/34

DrChainsaw:
[...]

PR #2082, opened 2022-10-13
Add explicit `train!`, unify `update!`, and auto-translate the two `Adam`s

mcabbott:
This is a non-breaking alternative to #2029, and part of #1986's goal to kill off Flux.Optimise. 

It adds an explicit `train!(loss, model, data, opt)`, which uses Optimisers.jl inside, and needs `opt = Flux.setup(Adam(), model)` before use. Tries to provide a friendly upgrade path:
* `setup` accepts the old `Adam()`, which is unchanged & can still be used with implicit `train!`. I hope this makes it easy to translate working code, as new & old will work in the same session. Then v0.14 will export the new `Adam()` instead, but code won't change.
* If you forget to call `setup`, it will still train, but will warn you if state is being lost.
* Most wrong mixtures of implicit and explicit arguments should produce helpful errors. Also an error is `d in data` which isn't a Tuple. This avoids the weird `batchmaybe` thing, which sometimes splats and sometimes doesn't. 
* `Flux.update! === Optimisers.update!`, so you can't use the wrong one. (At present there are two `setup`s.)

Needs doc updates, but probably not this PR. I think this would free us to remove every mention of implicit parameters from the docs, during 0.13.

Then 0.14 can delete Flux.Optimise and the `train!(..., ::Params, ...)` function completely.

This `train!` would like to have methods like `mse(m, x, y) = mse(m(x), y)` for all the loss functions, to allow `train!(mse, model, data, opt)` rather than defining a trivial wrapper function every time. Not this PR though. (Now #2090.)

The meaning of `data` is still the same as before -- it's an iterator usually over tuples, which are usually splatted into the loss. This is (IMO) a confusing feature of `train!`, and perhaps the implicit / explicit break is the time to fix that too. One possibility would be to take arrays not an iterator `train!(mse, model, X, Y; opt)`:
```julia
function train!(loss, model, data...; opt, batchsize=nothing, kw...)
  if batchsize != nothing
    for d in DataLoader(data; batchsize, kw...)
      g, _... = gradient(loss, model, d...)  # same arguments order as train!
      update!(opt, model, g)
    end
  else
    g, _... = gradient(loss, model, data...)  # exact same arguments as train!
    update!(opt, model, g)
... 
```

~~It also adds an explicit way of changing the AD used, via `@train_autodiff`. RFC, I guess. Tests for it run on Tracker & Yota.~~ Removed in [bbc0f85](https://github.com/FluxML/Flux.jl/pull/2082/commits/bbc0f856695dd0fccd06264d279dcfc3ea90caac), for now, to make things more orthogonal. The macro was the same as [this bit of 2029](https://github.com/FluxML/Flux.jl/pull/2029/files#diff-97456b64e767b66e23147a996bb3613529be282661182504b54e713350acf108R109-R154).

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


src/train.jl
```diff
@@ -0,0 +1,142 @@
...
+Note that the built-in loss functions accept 3 arguments, allowing for instance
+`train!(Flux.Losses.mse, model, data, opt)` instead of defining `loss3` as above.
+
+!!! note
+    This method has significant changes from the one in Flux ≤ 0.13:
+    * It now takes the `model` itself, not the result of [`Flux.params`](@ref).
+      (This is to move away from Zygote's "implicit" parameter handling, with `Grads`.)
+    * Instead of `loss` being a function which typically accepts two arguments
+      (the input `x` and expected output `y` from each element of `data`)
+      now it should typically accept three, the first of which is the `model` itself.
```
darsnack, line 97: We never restricted specifically to 2 arguments (and we don't seem to restrict to 3 now either). I think the change is
- <= v0.13: `loss` accepts as many arguments as `length(first(data))`
- \> v0.13: `loss` accepts at least 1 argument, the model, and can accept additional `N` additional arguments where `N = length(first(data))`

I think the distinction is important, since for things like language models, `length(first(data)) == 1` (conceivably), and the loss handles taking the single data argument and turning it into a supervised problem.

darsnack (commented):
What do you think about merging this and #2083? With the latest changes, they are orthogonal, so it would be rebasing #2083 to replace `explicit_withgradient` once this is merged.

mcabbott:
> What do you think about merging this and https://github.com/FluxML/Flux.jl/pull/2083? With the latest changes, they are orthogonal

Yes. I simplified this to just the `train!`-API story, kicking the question of whether or how you might select an AD to use down the road. 

I have comments on #2083 which I should tidy up and post there. 

I think the major question here is whether, when making a new and incompatible method for `train!`, we should
* Do the minimal thing as here, and keep `data` as something which iterates tuples, and 4th positional argument is the optimiser state. Or:
* Do more substantial changes at the same time, e.g. to avoid having 4 positional arguments.

mcabbott:
> Do more substantial changes at the same time

So the confusing features of `train!` are:
1. It takes 4 positional arguments in a hard-to-remember order
2. It doesn't really do much more than a loop with `gradient` + `update!`
3. It splats the data into the loss, sometimes (or in this PR, always, else error)

One alternative would be to change it to 
```
train!(loss, model, data...; state, kw...)
```
which calls `gradient(loss, model, data...)` -- no mystery order. And then to build in DataLoader, so that `train!(loss, model, data...; state, batchsize = 32)` breaks up `data` for you. Each batch is the same shape as the original, though, so this does not introduce confusion about argument order. Since the keyword `state` (or something?) is required, this method should not be easily confused with the old one.

src/train.jl
```diff
@@ -0,0 +1,142 @@
...
+
+julia> model.bias  # was zero, mutated by Flux.train!
+1-element Vector{Float32}:
+ 10.190001
+
+julia> opt  # mutated by Flux.train!
+(weight = Leaf(Momentum{Float64}(0.1, 0.9), Float32[-2.018 3.027]), bias = Leaf(Momentum{Float64}(0.1, 0.9), Float32[-10.09]), σ = ())
+```
+"""
+function setup(rule::Optimisers.AbstractRule, model)
```
darsnack, line 46: I am hesitant to create a function in the Flux namespace that clashes with Optimisers.jl. It is hard enough already to keep track of where "Flux functions" actually come from.

Why not extend `Optimisers.setup` for `Flux.Optimise.AbstractOptimiser` and remove the mutability check? I am guessing this is to guard against immutable models since `train!` does not return the model?

src/train.jl
```diff
@@ -0,0 +1,142 @@
...
+
+function _rule_to_state(model, rule::Optimisers.AbstractRule)
+  state = setup(rule, model)
+  @gensym warn_id
+  name = typeof(rule).name.name
+  fmap(state, exclude = x -> x isa Optimisers.Leaf) do leaf
+    leaf.state isa Nothing ||  @warn """Optimiser $name has state which will be discarded after `train!` finishes.
+                                        Please run `opt = Flux.setup($name(), model)` and pass this `opt` to `train!`.""" leaf maxlog=1 _id=warn_id
+    leaf
+  end
```
darsnack, line 136: This check is only relevant when `AbstractOptimiser` was originally passed in and upgraded right?

darsnack (commented):
My only hesitancy around this PR involve choices made to keep `train!` as close to the old one as possible. So, do not require users to capture the return value, etc.

I would prefer returning both the model and the state from `train!`. In v0.13.X, we do not need to document this change to the behavior. We silently upgrade to full Optimisers.jl support and throw deprecation warnings whenever `Params` and `AbstractOptimiser` are used. Then some versions down the line, we kill support for implicit.

Also, is the a reason for a separate `Train` submodule?

src/train.jl
mcabbott: No, or at least that's not the intention. 

If you do not call `setup` at all, and directly call `train!`, then this helpfully runs `setup` in the background. If you are using `Optimisers.Descent()` then this is actually fine, there is no meaningful state and so you don't have to care. For almost any other optimiser, you will get one epoch of training, and a warning that the state is being discarded, and how to fix this.

I think this is a not-so-uncommon mistake to make, calling `train!(...., Adam())` with a fresh Adam in every call. In the present setup, we can't detect this and warn, as it looks just like the first epoch should.

src/train.jl
mcabbott: I thought I did (in addition) extend `Optimisers.setup` that way, but in fact I made an error. Can change it. 

Yes the guard against immutable models is the point. All Flux models are assumed mutable right now, and this just makes the check explicit.

We can of course make `train!` return the model. But this isn't enough, as you *also* have to re-do your code to keep not discard the returned model. It's a bit awkward.

mcabbott:
> choices made to keep train! as close to the old one as possible. 

See suggestions above for how we might change it. But one of the goals is making the transition as easy as possible, which argues for keeping the weird `data` splat thing. 

> So, do not require users to capture the return value, etc.

How could you require this? You could encourage it, but since code without it will continue to work, it seems tricky to keep saying "you really ought to...".

> I would prefer returning both the model and the state from train!.

Note that there is no need to return the state. This is guaranteed to be mutated, even for a model of immutable arrays. We could change `update!` to return just one thing, which would be less confusing (one less thing to remember the order of!) but perhaps a confusing change, and would have to differ from `update` no-bang.

> We silently upgrade to full Optimisers.jl support and throw deprecation warnings whenever Params and AbstractOptimiser are used. Then some versions down the line, we kill support for implicit.

Yes, that's precisely what this PR amis to do. It makes claims that 0.14 will remove this. But how soon that is, we can see. 

> Also, is the a reason for a separate Train submodule?

No very strong one, there's a Losses module. And an Optimise module which (1) is a terrible near-clash, and (2) contains neatly everything which the no-more-implicit change will delete entirely.

src/train.jl
mcabbott: BTW, I picture Flux 0.14 deleting Flux.Optimise and exporting Adam etc from Optimisers.jl. 

Code that goes `using Flux; opt = Flux.setup(Adam(), model); train!(loss, model, data, opt)` will work equally well on 0.13 and 0.14. You don't have to load Optimisers.jl yourself at all, and all will be safe.

If you do load Optimisers.jl yourself and use its functions, then you have opted into the `model, _ = update!(opt, model, grad)` thing where you are supposed to get back the new model.

darsnack: > Code that goes using Flux; opt = Flux.setup(Adam(), model); train!(loss, model, data, opt) will work equally well on 0.13 and 0.14.

I guess the question is should this be the case? I do think there should be a patch release of v0.13.X that accepts `Optimisers.Adam`, etc. and upgrades `Flux.Optimise.Adam` with a warning. This will allow `train!` to work like quoted above too. But in v0.14, I was expecting that we force people to start using `model = train!(...)`. Previously, `train!` and `update!` worked similarly (mutating optimizers and model), and we could say `train!` is "just" a loop. Diverging how they work seems worse than a minor code refactor on a breaking release. Especially given people will get warnings from before.

src/train.jl
darsnack: Okay, so then for any case where `model !isa Params`, we expect the output of `setup` instead of a `AbstractOptimiser` or `AbstractRule`. We add this warning as a feature to `train!` to remind people to do `model = train!(...)`. We can't enforce this behavior, but we can warn.

mcabbott: > for any case where model !isa Params, we expect the output of setup instead of a AbstractOptimiser or AbstractRule. 

Yes. Explicit params goes with `setup`.

> We add this warning as a feature to train! to remind people to do model = train!(...)

This seems completely orthogonal. Once you have inserted `setup`, but still expect `train!` to mutate, you cannot get a warning about not calling `setup`.

src/train.jl
mcabbott: Because of this difficulty, and because Flux models do actually only have mutable parameter arrays for the foreseeable future, my proposal is roughy that we accept this:
* Let everything `Flux.Train` require (and ideally check) mutability, as `Flux.Optimise` does now.
* Everything `Optimisers.xxx` allow for immutability. Calling those functions is how you opt in to less-mutable world.

> Diverging how they work

By this you mean that `train!` does not return the model, and `update!` does. Part of me still wishes that `train!` would return the loss but it's not important. It could be made to return the model for the sake of symmetry. We could have a `Flux.update!` which returns `nothing` for the sake of symmetry, but shouldn't since we gain nothing. 

My proposal is I think that Flux's docs simply ignore the fact that explicit `update!` returns something not nothing, it's a mutating function with a `!`. We can revisit this if ImmutableArrays ever actually land.

darsnack: Okay, let's forget my suggestions about the warnings, as I agree about the orthogonality.

One option here is to return the model from `train!` which would allow for immutable models to work. Mutable models still don't need to capture the return value to work. So, we don't force people to do `model = train!(...)`. And we still have `Flux.setup` here to work in the reverse direction: warn if any leaf is immutable.

darsnack: But I agree that if we are adding `Flux.setup`, then this seems like something that can be revisited later too.

mcabbott: One thing we should consider is changing Optimisers. Maybe its present function should be called `update!!` as that's something of a convention for "tries to mutate but may fail". 

Then in Optimisers@0.3, we can introduce a new function `update!` which demands mutability, fails on immutable parameters. And that's the one we identify with Flux's function.

src/train.jl
```diff
@@ -0,0 +1,142 @@
...
+    This method has significant changes from the one in Flux ≤ 0.13:
+    * It now takes the `model` itself, not the result of [`Flux.params`](@ref).
+      (This is to move away from Zygote's "implicit" parameter handling, with `Grads`.)
+    * Instead of `loss` being a function which typically accepts two arguments
+      (the input `x` and expected output `y` from each element of `data`)
+      now it should typically accept three, the first of which is the `model` itself.
+    * `data` must iterate tuples, otherwise you get an error.
+      (Previously non-tuple types were not splatted into the loss. 
+      Pass in `((d,) for d in data)` to simulate this.)
+    * `opt` should be the result of [`Flux.setup`](@ref). Using an optimiser
```
CarloLucibello, line 101: Does Documenter need [`Flux.Train.setup`](@ref) instead?

mcabbott: Maybe. I just tried, and I think neither this new `train!` nor `setup` appear in the docs at present. That section needs to be re-worked for explicit parameters.

src/train.jl
```diff
@@ -0,0 +1,142 @@
...
+    d isa Tuple || error("""train! expects as data an iterator producing tuples, but got $(typeof(d)).
+                            Pass it `((d,) for d in data)`, or use `gradient` and `update!` for more control.""")
+    # l, (g, _...) = explicit_withgradient(loss, model, d...)  # BTW this un-thunks gradient w.r.t. data. Could avoid that
+    l, (g, _...) = explicit_withgradient(m -> loss(m, d...), model)
+    isfinite(l) || throw(DomainError("loss function returned $l, stopping training"))
+    opt, model = Optimisers.update!(opt, model, g)
+    push!(losses, l)
+    @logprogress Base.haslength(data) ? i/length(data) : nothing
+  end
+  return losses  # Not entirely sure returning losses is a good idea, as it may conflict with later returning immutable models alla Optimisers.jl
```
CarloLucibello, line 120: I think that instead of returning all the losses, which is a bit arbitrary, we should return the model and allow for passing a callback instead. For flexibility, the callback could take a named tuple input with the following payload: 
```julia
cb((; model, opt, data=d, step=i, loss=l, gradient=g))
``
`

Comment (mcabbott) [reply src/train.jl]: Maybe returning the loss isn't a great idea... removed now.

I agree that if you wanted a callback, then explicitly passing it stuff would be better than the current global story. But I think my vote is to remove this? If you just write a `for` loop then you can do all of this and more, in completely normal Julia code.

File: src/train.jl (hunk 2 above)
Comment (CarloLucibello) [reply src/train.jl:43]: I think `train!` should guarantee mutation, this is the widespread julian convention. We can have a `train` and a `train!!` for non-mutating and mutate-if-possible versions.
In that case, whether it returns the model or not hasn't great relevance. Base functions such as `replace!` and `map!` return the mutated input. Maybe just for REPL usage convenience? In our case returning the model in the repl would just be an annoyance I guess.

Comment (mcabbott) [reply src/train.jl:43]: Yes, the present one returns `nothing`, which as you say means you don't get a screenful of stuff, and also serves as a reminder that it mutates the model. 

I think I'd be happiest if `update!` did the same. I understand the attraction of `state, model = update(state, model, grad)` but IMO it's a pain to remember the order, and `update!` is now in a weird place where it does guarantee to mutate the `state`, but not the model.

File: src/train.jl (hunk 5 above)
Comment (CarloLucibello) [reply src/train.jl]: agree, and in case callback support can be added later if someone pushes for it.
One argument in favor could be that `train!` now supports callback, some people maybe have built their code/workflow around that and we don't want to disrupt it too much.

File: src/train.jl [hunk 6]
```diff
@@ -0,0 +1,136 @@
...
+                            For more control use a loop with `gradient` and `update!`.""")
+  losses = Float32[]
+  @withprogress for (i,d) in enumerate(data)
+    d isa Tuple || error("""train! expects as data an iterator producing tuples, but got $(typeof(d)).
+                            Pass it `((d,) for d in data)`, or use `gradient` and `update!` for more control.""")
+    # l, (g, _...) = explicit_withgradient(loss, model, d...)  # BTW this un-thunks gradient w.r.t. data. Could avoid that
+    l, (g, _...) = explicit_withgradient(m -> loss(m, d...), model)
+    isfinite(l) || throw(DomainError("loss function returned $l, stopping training"))
+    opt, model = Optimisers.update!(opt, model, g)
+    push!(losses, l)
```
CarloLucibello, line 112:
```suggestion
    push!(losses, l)
```
and also the `losses` definition above

src/train.jl
```diff
@@ -0,0 +1,136 @@
...
+"""
+function train!(loss, model, data, opt; cb = nothing)
+  isnothing(cb) || error("""train! does not support callback functions.
+                            For more control use a loop with `gradient` and `update!`.""")
+  losses = Float32[]
+  @withprogress for (i,d) in enumerate(data)
+    d isa Tuple || error("""train! expects as data an iterator producing tuples, but got $(typeof(d)).
+                            Pass it `((d,) for d in data)`, or use `gradient` and `update!` for more control.""")
+    # l, (g, _...) = explicit_withgradient(loss, model, d...)  # BTW this un-thunks gradient w.r.t. data. Could avoid that
+    l, (g, _...) = explicit_withgradient(m -> loss(m, d...), model)
```
CarloLucibello, line 109: If we consider this function to be scrutinized by many people which will then move on to define their own training loop, maybe better to be really explicit and do
```suggestion
    l, (g, _...) = Zygote.withgradient(m -> loss(m, d...), model)
```

src/train.jl
mcabbott: Yes. I guess that's an argument for running the old-style callbacks still? Regard this as a legacy function to ease the transition? Not sure.

src/train.jl
mcabbott: Oops, done.

src/train.jl
mcabbott: Should probably go.

Earlier commits had a mechanism like `@train_autodiff Tracker` which replaced this function, as a global switch of what AD to use. (Turing does something like this.) But... maybe this too I think is better done by writing a loop and calling `Tracker.withgradient` or whatever.

CarloLucibello:
I would keep the `batchmemaybe(d)` thing. It is not uncommon for datasets to yield named tuples or dictionaries (e.g. hugging face does that) and wrapping dataloaders into generators is ugly and annoying. And also because of the let's try to disrupt as little as possible argument.

src/train.jl
CarloLucibello: we could still support (with a deprecation message) the old-style callback wrapping a try catch block around every cb:
```
wrap_callback(cb)
  nt -> begin
    try 
      cb(nt)
    catch e
      if e isa MethodError
       @warn "...." 
      end
      cb()
   end
end
```

mcabbott: I could be persuaded to add back the old-style callback handling, and even old-style `batchmaybe`, in the name of making upgrades easier. 

I agree that a new more advanced callback mechanism could be added in a non-breaking way. But this PR does too much already. 

I'd even really rather add the old-style things in another PR.

CarloLucibello: ok

CarloLucibello:
conditional on fixing the failing test LGTM

src/train.jl
mcabbott: I've removed the "3 arguments" bit, to say just "Instead of `loss` being a function which accepts only the data, now it must also accept the `model` itself, as the first argument."

src/train.jl
mcabbott: Doc update is #2114, will need to be rebased on this & then checked.

test/train.jl
```diff
@@ -0,0 +1,93 @@
...
+  # Explicit gradient
+  gs = gradient(marg -> marg(x), m)
+  @test gs isa Tuple
+  @test_throws ErrorException Flux.update!(Flux.Adam(), Flux.params(m), gs) # friendly
+  @test_throws ErrorException Flux.update!(Flux.Adam(), Flux.params(m), gs[1]) # friendly
+  @test_throws ErrorException Flux.update!(Flux.Adam(), m, gs)  # friendly
+  @test_throws ErrorException Flux.update!(Flux.Adam(), m, gs[1])  # friendly
+  s = Flux.setup(Adam(), m)
+  @info "ignore this warning, just testing an upgrade path:"
+  Flux.update!(s, m, gs)  # Chain + Tuple can be unambiguously sorted out
```
mcabbott, line 80: Most recent commits add some friendly errors to most ways you could use `update!` wrong, by mixing up implicit & explicit bits, or forgetting to call `setup`.

docs/src/models/overview.md
```diff
@@ -112,40 +114,28 @@ julia> predict.bias
...
-julia> parameters = Flux.params(predict)
-Params([Float32[0.9066542], Float32[0.0]])
-```
-
-These are the parameters Flux will change, one step at a time, to improve predictions. At each step, the contents of this `Params` object changes too, since it is just a collection of references to the mutable arrays inside the model: 
-
-```jldoctest overview
-julia> predict.weight in parameters, predict.bias in parameters
-(true, true)
-```
```
mcabbott, line 127: Also new, remove all `params` from "overview".

CarloLucibello (approved):

PR #2083, opened 2022-10-14
Add support for AD backends and explicit optimizers

darsnack:
This is another approach in a similar vein to #2029 and #2082. The primary goal of this PR is to focus on the use of alternate AD backends, since this is necessary for explicit mode training. Note that while this PR does add explicit support to `train!`, it does not tackle transitioning the optimizers to using only Optimisers.jl. As such, this PR could be merged and #2082 put on top of it (so it is not a complete replacement for the other PRs).

The changes allow Flux to be used with explicit mode by passing `Flux.Optimise.ZygoteExplicitBackend()` to `train!`:
```julia
model = Chain(Dense(2 => 1, relu), Dense(1 => 1))
opt = Optimisers.Momentum(1e0)
optstate = Optimisers.setup(opt, model)
ad = Flux.Optimise.ZygoteExplicitBackend()

for epoch in 1:10
    @info "Epoch $epoch"
    optstate, model = Flux.train!(ad, model, data, optstate) do model, x, y
        Flux.Losses.logitbinarycrossentropy(model(x), y)
    end
end
```
From the user's perspective, the approach taken here is to explicitly (pun not intended) require the correct things to be passed to train (the AD backend and optimizer state tree). If there's a mismatch, then errors are thrown. The default backend is the implicit mode for Zygote. Since AbstractDifferentiation already has backends for other ADs, a user can load the corresponding AD and run the code they want:
```julia
model = Chain(Dense(2 => 1, relu), Dense(1 => 1))
opt = Optimisers.Momentum(1e0)
optstate = Optimisers.setup(opt, model)
ad = AD.TrackerBackend()

for epoch in 1:10
    @info "Epoch $epoch"
    optstate, model = Flux.train!(ad, model, data, optstate) do model, x, y
        Flux.Losses.logitbinarycrossentropy(model(x), y)
    end
end
```

The main change here is to stop using `Zygote.gradient` and use `AD.gradient` (where `AD === AbstractDifferentiation`) instead. Even though AbstractDifferentiation.jl supports Zygote, it really only supports the explicit mode as a special case of ChainRules.jl compatible reverse mode ADs. This PR wraps this and does instead:
1. Define `ZygoteImplicitBackend` and `ZygoteExplicitBackend` as AbstractDifferentiation backends with the appropriate primitives defined. Both wrap `AD.ZygoteBackend` but a clear type for each allows Flux to specialize dispatch.
2. Define `AD.gradient` for both of the above to get around https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/63#issuecomment-1225959150.
3. Add `Optimisers.update` and `Optimisers.update!` for `Flux.AbstractOptimiser` (removing the old `Flux.update!`)
4. Minor changes to `train!` to utilize everything above correctly. Namely, constructing the loss correctly for implicit vs. explicit gradients.
5. The following needs to be defined for Tracker.jl support:
   ```julia
   # this is to work around AD.TrackerBackend only supporting vectors of params
   AD.gradient(::AD.TrackerBackend, f, xs...) = Tracker.withgradient(f, xs...).grad
   ```
   This can be submitted as a PR to AbstractDifferentiation.jl (or Tracker.jl can officially adopt AbstractDifferentiation.jl). I tried adding it directly to Flux, but seemed to run into issues with Tracker being `@require`-ed in AbstractDifferentiation.jl.

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


src/optimise/gradients.jl
```diff
@@ -0,0 +1,23 @@
...
+    core_backend::T
+end
+ZygoteImplicitBackend() = ZygoteImplicitBackend(AD.ZygoteBackend())
+
+AD.@primitive pullback_function(ad::ZygoteImplicitBackend, f, x::Zygote.Params) =
+    AD.pullback_function(ad.core_backend, f, x)
+
+# this is a hack to get around
+# https://github.com/JuliaDiff/AbstractDifferentiation.jl/issues/63#issuecomment-1225959150
+AD.gradient(::ZygoteImplicitBackend, f, x::Zygote.Params) = Zygote.gradient(f, x)
```
ToucheSir, line 11: Could this be `value_and_gradient` to support changes like https://github.com/FluxML/Flux.jl/pull/2070?

src/optimise/train.jl
```diff
@@ -12,18 +8,23 @@ according to optimizer `opt`  and the gradients `gs` (the gradient `g`).
...
-function update!(opt::AbstractOptimiser, xs::Params, gs)
+function Optimisers.update!(opt::AbstractOptimiser, xs::Params, gs)
   for x in xs
     isnothing(gs[x]) && continue
     update!(opt, x, gs[x])
   end
+
+  return opt, xs
 end
+Optimisers.update(opt::AbstractOptimiser, xs::Params, gs) = update!(opt, xs, gs)
```
ToucheSir, line 27: Needs a defensive copy? Or does https://github.com/FluxML/Optimisers.jl/blob/master/src/interface.jl#L57-L58 handle that automatically?

src/optimise/train.jl
```diff
@@ -120,16 +131,15 @@ The callback can call [`Flux.stop`](@ref) to interrupt the training loop.
...
   n = (itrsz == Base.HasLength()) || (itrsz == Base.HasShape{1}()) ? length(data) : 0
   @withprogress for (i, d) in enumerate(data)
     try
-      gs = gradient(ps) do
-        loss(batchmemaybe(d)...)
-      end
-      update!(opt, ps, gs)
+      _loss = _build_loss(ad, loss, batchmemaybe(d))
+      gs = _gradient_only(AD.gradient(ad, _loss, model))
+      optstate, model = update(optstate, model, gs)
```
ToucheSir, line 142:
```suggestion
      optstate, model = update!(optstate, model, gs)
```
I think we can get away with this by default.

src/optimise/train.jl
darsnack: I opted to remove it, since a defensive copy of `Params` is tricky. And with the change you suggested below, we don't really need this anymore.

src/optimise/gradients.jl
darsnack: Not quite, because it runs into the issue you mentioned in the link above the code. I could define both `gradient` and `value_and_gradient` to essentially block out AbstractDifferentiation until they sort out the primitives issues.

darsnack: Alternatively, it might make sense to have `Flux.gradient` and `Flux.withgradient` that defaults to `AD.gradient` and `AD.value_and_gradient`. Right now, `Flux.gradient(f, xs...)` wouldn't default to `ZygoteImplicitBackend`. Defining our own method would allow us to do this.

mcabbott:
I wrote some comments which I should post:

Complicated Flux models currently use `gradient` and `update!`, and I think we agree this will still be the case in future. The role of `train!` is presently fairly simple uses where you don't want to customise much.

Without `train!`, at present:

* To select a different AD, you can use a different package's `gradient`. You have to load (say) Diffractor, and surely you need to know that `Diffractor.gradient` or `Yota.grad` exist -- a module and a function.

* To select a different mode of Zygote, you call a different method: `gradient(fn, args...)` vs. `gradient(f0, ::Params)`. (The corresponding methods `update!(opt, m, g)` vs `update!(opt, ::Params, ::Grads)` are not yet the same function, but could be.)

If I understand right, this PR wants to introduce a *different* mechanism for changing what AD is used, and what mode. You would do both by passing a token to `train!`. My objections are:

* 4 positional arguments is already a lot, easy to get wrong. 5 in a particular order is... one more.

* And one more object you need to import from somewhere, give a name to, and keep track of. Now three: a module, a function, and a special token.

* It means `train!` has a different mechanism for choosing the AD (and style) to `gradient` (etc). You could fix this by also extending the same mechanism to a new `Flux.gradient`, or hope everyone starts using Abst.Diff.'s function. But it's unlikely you that won't have to know about `Yota.gradient`, so it's still an additional thing to know, another layer of API.

* With Zygote alone, this mechanism lets you use the same signature `train!(ad, loss3, model, data, state)` for both implicit and explicit mode. That might be a less weird way to use implicit mode than the present `train!(loss0, ::Params, data, opt)`, and would enable us to remove the old `::Params` signature from the docs. But why use it at all? If you are changing the syntax, why not move fully to explicit mode in one go? Or is there a limitation to Zygote's explicit mode that makes it desirable to keep using `Params`, internally?

darsnack:
Answering the more specific comments separately:

> 4 positional arguments is already a lot, easy to get wrong. 5 in a particular order is... one more.

Alternatively, `train!` could not accept the backend as an argument, and simply construct `ZygoteImplicitBackend()` internally and be the same. Or don't make the AD backend a positional argument. I only wrote it this way for consistency with `AD.gradient(ad, ...)`.

> And one more object you need to import from somewhere, give a name to, and keep track of. Now three: a module, a function, and a special token.

Well if you don't care to switch ADs, then there is nothing to know or remember. I think this version is easier to learn and use than knowing to overload `explicit_withgradient` correctly.

> It means train! has a different mechanism for choosing the AD (and style) to gradient (etc). You could fix this by also extending the same mechanism to a new Flux.gradient, or hope everyone starts using Abst.Diff.'s function. But it's unlikely you that won't have to know about Yota.grad, so it's still an additional thing to know, another layer of API.

The point is to coalesce around `AD.gradient` which is really the only thing that will work consistently across packages and users. FluxML owns Zygote.jl (implicit and explicit) and Tracker.jl. Just having those two buy into AbstractDifferentiation.jl would cover most ADs that people use in ML. Any ChainRules-compatible reverse mode AD gets supported by default, so Yota and Diffractor should work too. Explicit buy-in from those two would be nice though.

> With Zygote alone, this mechanism lets you use the same signature train!(ad, loss3, model, data, state) for both implicit and explicit mode. That might be a less weird way to use implicit mode than the present train!(loss0, ::Params, data, opt), and would enable us to remove the old ::Params signature from the docs. But why use it at all? If you are changing the syntax, why not move fully to explicit mode in one go? Or is there a limitation to Zygote's explicit mode that makes it desirable to keep using Params, internally?

Unless we plan on killing implicit mode within Zygote itself, I don't see a reason to forbid it. There doesn't need to be `train!` API for it, but the beauty of this PR is that `train!` can support it without us ever mentioning or promoting it.

ToucheSir:
@darsnack and I had a productive but unfortunately too short discussion on this last ML call, so putting some follow-up thoughts here.

Both #2082 and #2083 seek to have training with implicit and explicit params use roughly the same interface. Same function arity, same number of return values, etc. Given we're planning on removing `train!(loss, ps::Params, data, opt::AbstractOptimiser)` though, is there any reason to do this? It seems some of the contention around e.g. new vs old Optimizer compatibility would be resolved by adding a deprecation warning in the implicit params `train!` like so:
```
train!(loss, ps::Params, data, opt::AbstractOptimiser) is deprecated and will be removed in Flux v0.14.

Going forward, please use the new train! interface:

using Flux.Optimisers: Optimisers

optstate = Optimisers.setup(Optimisers.Descent(), model) 
optstate, model = Flux.train!((model, x, y) -> loss_fn(model(x), y), model, data, optstate)
```

This means no back and forth conversions. The `train!` methods with old and new optimizers (and implicit vs explicit params) are now mutually exclusive. We could even consider adding dep warnings to the old optimizer constructors themselves pushing users towards using new-style train with Optimisers.jl rules. As long as new style `train!` can handle any workflow that old `train!` could, we could change it to have whatever interface we want.

mcabbott:
> Unless we plan on killing implicit mode within Zygote itself, I don't see a reason to forbid it

The reason to drop it entirely from 0.14 is that this lets us delete Flux.Optimise, and all of its duplicate code for how Adam works, etc.

Keeping a path for implicit parameters while having only one Adam definition (in Optimisers.jl) means writing some new code to make corresponding `IdDict`s. Nobody much liked this in  https://github.com/FluxML/Flux.jl/pull/2029. Maybe it could be done in other ways, using the same state tree from `setup`.

Other uses of Zygote can of course use it as they wish. If ripping implicit parameters out of Zygote completely led to some substantial improvement, then that could be considered as a major change. But realistically nobody is going to get around to trying.

> The `train!` methods with old and new optimizers (and implicit vs explicit params) are now mutually exclusive. We could even consider adding dep warnings 

Yes, this is now the present state after #2082. There are doc notes but no dep warnings. They could be added in last version of 0.13? Should be in `train!` & `update!` I think.

> this version is easier to learn and use than knowing to overload explicit_withgradient correctly.

Note that such overloading was never a proposed API. The initial proposal #2082 (now removed) was that there be a macro `@train_ad Tracker` which globally changes what Flux uses, by some internal mechanism. This is similar to what Turing does. Global switches have their downsides, but they do avoid this one-more-named-thing problem with passing special tokens around.

> packages like FluxTraining.jl or FastAI.jl need to invent their own logic for swapping ADs

Yes. If they do settle on some nice high-level scheme, then one future for `train!` could be that it should be a junior version of those, which follows the same scheme. 

The other possible future is that `train!` is a legacy function, only for very simple training loops & to make upgrading old code easier. The question of whether it should take no/the same/more callbacks is related.

PR #2084, opened 2022-10-15
Simplify `Embedding`

mcabbott:
Embedding has some special code for OneHotMatrix which (1) will break with latest changes, and (2) doesn't allow higher-rank arrays the way that "index" input does:
```
julia> Embedding(26 => 10)(rand(1:26)) |> size
(10,)

julia> Embedding(26 => 10)(rand(1:26, 2,3,4,5)) |> size
(10, 2, 3, 4, 5)

julia> Embedding(26 => 10)(onehot(rand(1:26), 1:26)) |> size
(10,)

julia> Embedding(26 => 10)(onehotbatch(rand(1:26, 2), 1:26)) |> size  # can't go further
(10, 2)
```
So this PR simplifies & adds reshape. 

I did this after forgetting that #1656 exists, some overlap. This PR does not attempt to fix `outputsize`. Some other changes there have already happened elsewhere.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -654,25 +654,32 @@ The input to the layer can be a vocabulary index in `1:in`, an array of indices,
 or the corresponding [`onehot encoding`](@ref OneHotArrays.onehotbatch).
 
 For indices `x`, the result is of size `(out, size(x)...)`, allowing several batch dimensions.
-For one-hot `x`, the result is of size `(out, size(x)[2:end]...)`.
+For one-hot `ohx`, the result is of size `(out, size(ohx)[2:end]...)`.
 
 # Examples
 ```jldoctest
-julia> vocab_size, embed_size = 1000, 4;
+julia> emb = Embedding(26 => 4, init=Flux.identity_init(gain=22))
+Embedding(26 => 4)  # 104 parameters
 
-julia> model = Flux.Embedding(vocab_size => embed_size)
-Embedding(1000 => 4)  # 4_000 parameters
+julia> emb(2)  # one column of e.weight (here not random!)
+4-element Vector{Float32}:
+  0.0
+ 22.0
+  0.0
+  0.0
 
-julia> vocab_idxs = [1, 722, 53, 220, 3];
+julia> emb([3, 1, 20, 14, 4, 15, 7])  # vocabulary indices, in 1:26
+4×7 Matrix{Float32}:
+  0.0  22.0  0.0  0.0   0.0  0.0  0.0
+  0.0   0.0  0.0  0.0   0.0  0.0  0.0
+ 22.0   0.0  0.0  0.0   0.0  0.0  0.0
+  0.0   0.0  0.0  0.0  22.0  0.0  0.0
 
-julia> x = Flux.onehotbatch(vocab_idxs, 1:vocab_size); summary(x)
-"1000×5 OneHotMatrix(::Vector{UInt32}) with eltype Bool"
-
-julia> model(x) |> summary
-"4×5 Matrix{Float32}"
-
-julia> model(vocab_idxs) == model(x)
+julia> ans == emb(Flux.onehotbatch("cat&dog", 'a':'z', 'n'))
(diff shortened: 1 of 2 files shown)


src/layers/basic.jl
```diff
@@ -684,10 +694,9 @@ Embedding((in, out)::Pair{<:Integer, <:Integer}; init = randn32) = Embedding(ini
 (m::Embedding)(x::AbstractVector) = NNlib.gather(m.weight, x)
 (m::Embedding)(x::AbstractArray) = reshape(m(vec(x)), :, size(x)...)
 
-function (m::Embedding)(x::Union{OneHotVector{T,L}, OneHotMatrix{T,L}}) where {T,L}
-  size(m.weight, 2) == L || throw(DimensionMismatch("Matrix column must correspond with OneHot size: $(size(m.weight, 2)) != $L"))
-  return m(onecold(x))
-end
+(m::Embedding)(x::AbstractVector{Bool}) = m.weight * x  # usually OneHotVector
+(m::Embedding)(x::AbstractMatrix{Bool}) = m.weight * x  # usually OneHotMatrix
```
mcabbott, line 698: These could instead call `Flux.onecold`. The result will differ on e.g. `[true, true, false]`, not sure we care too much either way?

darsnack: For performance in the one hot case? If it's `onecold`-compatible, then folks should use `OneHotArray` for performance. At least with `*`, we do the mathematically expected operation.

mcabbott: For `OneHotArray` these should be identical, right? Result and performance.

For a one-hot BitArray, the results will agree. I would guess that onecold is faster but haven't checked. 

For a generic BitArray, I'm not sure which is mathematically expected really. I think you're saying that `*` is.

darsnack: Yes, what you wrote is what I meant re: performance. I was adding that in the one-hot bit array case, we can direct people to `OneHotArray` if their concern is performance.

Yeah whenever I've come across this type of operation in papers, I see it written as `*`. There's an implicit assumption that `x` is one-hot, so maybe `onecold` could be better here if it were made to error for `[true, true, false]`, etc. But I think silently choosing the first "hot" index is wrong.

mcabbott: Yes. Mixing two embedding vectors seems less wrong. But probably nobody ever hits this & it's just a way to decouple from OneHotArray types. I don't think we should document that boolean indexing is an option.

src/layers/basic.jl
```diff
@@ -644,32 +644,42 @@ function Base.show(io::IO, m::PairwiseFusion)
...
+  0.0   0.0  0.0  0.0  22.0  0.0  0.0
+
+julia> ans == emb(Flux.onehotbatch("cat&dog", 'a':'z', 'n'))
 true
+
+julia> emb(rand(1:26, (10, 1, 12))) |> size  # three batch dimensions
+(4, 10, 1, 12)
 ```
 """
 struct Embedding{W}
```
mcabbott, line 685: #1656 also restricted this type:
```suggestion
struct Embedding{W<:AbstractMatrix}
```
At the moment, you can do this, which doesn't make much sense:
```
julia> Flux.Embedding(3)
Embedding(1 => 1)

julia> ans(1)
ERROR: MethodError: no method matching getindex(::Int64, ::Colon, ::Int64)
```

darsnack: Let's make this change

src/layers/basic.jl
darsnack: So I think we are happy with the current implementation in the PR?

mcabbott:
The "more" is #2088, really. Will merge when green.

src/layers/basic.jl
mcabbott: Yes I think so. 

I see we had a very similar discussion in https://github.com/FluxML/Flux.jl/pull/1656#discussion_r669612039 BTW, I forgot... but same conclusion.

PR #2088, opened 2022-10-16
Make `outputsize` work with `Embedding`

mcabbott:
Like https://github.com/FluxML/Flux.jl/pull/1656 this wants to make `outputsize(Embedding(3 => 4), (5,)) == (4, 5)`. That is, it thinks the size referred to by `outputsize` should be the size of the array of vocabulary indices, not the size of the one-hot representation. 

But rather than overload indexing or `gather` (as here https://github.com/FluxML/Flux.jl/pull/1656/files#diff-0dfa3b94337acdaa714025f5198f6907e6a50a59aac03ba1230fbcb681126da2R172) this just adds methods to `(::Embedding)`. I think that's least likely to cause surprises. If indexing shows up elsewhere, we can decide then whether to extend. 

Restricting `(m::Embedding)(x::AbstractArray{<:Integer})` also seems like the right thing to do, error right away on non-integer input.

### DIFF
--- a/src/outputsize.jl
+++ b/src/outputsize.jl
@@ -288,8 +288,10 @@ is needed to make `@autosize (2,3,4) Dense(_ => 5)` return
 """
 autosizefor(::Type, x::AbstractArray) = size(x, max(1, ndims(x)-1))
 autosizefor(::Type{<:Dense}, x::AbstractArray) = size(x, 1)
-autosizefor(::Type{<:Embedding}, x::AbstractArray) = size(x, 1)
 autosizefor(::Type{<:LayerNorm}, x::AbstractArray) = size(x, 1)
+
+autosizefor(::Type{<:Embedding}, x::AbstractArray) = error(
+  "@autosize Embeeding(_ => n) cannot work, as this _ is the size of the vocabulary, not an array size")
 
 _replaceunderscore(e, s) = e === :_ ? s : e
 _replaceunderscore(ex::Expr, s) = Expr(ex.head, map(a -> _replaceunderscore(a, s), ex.args)...)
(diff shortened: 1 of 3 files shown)


src/outputsize.jl
```diff
@@ -157,6 +163,9 @@ end
 
 ## fixes for layers that don't work out of the box
 
+(m::Embedding)(x::Nil) = similar(m.weight, Nil, size(m.weight, 1))
+(m::Embedding)(x::AbstractArray{Nil}) = similar(m.weight, Nil, size(m.weight, 1), size(x)...)
```
darsnack, line 167: Should these warn about the limitations? `outputsize` can be used in sanity checking contexts, so I think a user would be willing to accept a little noise.

src/outputsize.jl
```diff
@@ -87,6 +87,12 @@ DimensionMismatch("Input channels must match! (7 vs. 3)")
 julia> outputsize([Dense(10 => 4), Dense(4 => 2)], (10, 1)) # Vector of layers becomes a Chain
 (2, 1)
 ```
+
+Limitations:
+* `Embedding` accepts either integers or one-hot arrays, and `ohx = onehotbatch(x, ...)`
+  has one more dimension than `x`. Here `outputsize` uses `size(x)`.
+* At present `outputsize` does not work with recurrent layers,
+  `outputsize(RNN(2 => 3), (2, 1))` gives an error. This is a bug.
```
ToucheSir, line 95:
```suggestion
  `outputsize(RNN(2 => 3), (2, 1))` gives an error. See https://github.com/FluxML/Flux.jl/pull/1755 for more.
```
I'm not up to speed on any `Embedding` changes, but a link reference here would be nice.

ToucheSir:
Would this landing obsolete https://github.com/FluxML/Flux.jl/pull/1656?

CarloLucibello:

PR #2093, opened 2022-10-24
Use the standard Documenter.jl local redirect

ChrisRackauckas:
That way it stays up to date.


ToucheSir:
Fixing all the legacy weirdness with the Flux site, one PR at a time :P

Issue #2045, opened 2022-08-22
v0.13.5 breaks Flux.train! on a custom type


simonmandlik:
Hi, in [Mill.jl](https://github.com/CTUAvastLab/Mill.jl) we have a special matrix type we call [PostImputingMatrix](https://github.com/CTUAvastLab/Mill.jl/blob/master/src/special_arrays/postimputing_matrix.jl), which fills in a vector trainable parameters ψ when they are missing in the input data. In short:

```julia
julia> using Flux, Mill

julia> A = Mill.PostImputingMatrix(reshape(1:9, 3, 3))
3×3 PostImputingMatrix{Int64, Base.ReshapedArray{Int64, 2, UnitRange{Int64}, Tuple{}}, Vector{Int64}}:
W:
 1  4  7
 2  5  8
 3  6  9

ψ:
 0
 0
 0

julia> B = Mill.maybehotbatch([1, missing, 3], 1:3)
3×3 MaybeHotMatrix with eltype Union{Missing, Bool}:
  true  missing    ⋅
   ⋅    missing    ⋅
   ⋅    missing   true

julia> A * B
3×3 Matrix{Int64}:
 1  0  7
 2  0  8
 3  0  9
```

For efficiency, this requires several some [custom `rrule`s](https://github.com/CTUAvastLab/Mill.jl/blob/23b5f66d6677d2cc3e15fec5f06ad8604732013c/src/special_arrays/postimputing_matrix.jl#L100-L119) as well as [`opt`ing out of default rules for Base.*](https://github.com/CTUAvastLab/Mill.jl/blob/23b5f66d6677d2cc3e15fec5f06ad8604732013c/src/special_arrays/postimputing_matrix.jl#L71).

When taking parameters and computing gradients, in `Flux@v0.13.4` it works as follows:

```julia
julia> Flux.params(A)
Params([PostImputingMatrix(W = [1 4 7; 2 5 8; 3 6 9], ψ = [0, 0, 0])])

julia> gradient(() -> sum(A*B), Flux.params(A)).grads
IdDict{Any, Any} with 2 entries:
  3×3 PostImputingMatrix => (W = [1 0 1; 1 0 1; 1 0 1], ψ = [1, 1, 1])
  :(Main.A)              => (W = [1 0 1; 1 0 1; 1 0 1], ψ = [1, 1, 1])
```

[...]

PR #2098, opened 2022-11-02
Safer gradients (by copying before mutating) & less piracy (by removing ArrayInterface)

mcabbott:
#1613 added a check to work around the fact that Zygote may return an immutable array as a gradient, and Flux.Optimise likes to mutate them. However, such mutation isn't actually safe, as there is no guarantee that arrays in the gradient do not alias each other. So this PR inserts a copy instead.

#1613 also added a dep on ArrayInterface. This was supposed to provide a safe function for turning immutable arrays into mutable ones. But in reality, ArrayInterface is pretty experimental, and its piracy causes the following problem, noticed today:
```
using Flux, StaticArrays; setindex((@SMatrix [1 0 ; 1 1]), 5, 1, 1)

ERROR: MethodError: setindex(::SMatrix{2, 2, Int64, 4}, ::Int64, ::Int64, ::Int64) is ambiguous. Candidates:
 setindex(a::StaticArray, x, inds...) in StaticArrays at /home/jonathan/.julia/packages/StaticArrays/PUoe1/src/deque.jl:198
 setindex(x::AbstractMatrix, v, i::Int64, j::Int64) in ArrayInterface at /home/jonathan/.julia/packages/ArrayInterface/R0AhD/src/ArrayInterface.jl:199
```
Flux does commit some piracy, but should aim to commit less. And especially not in wildly unrelated functions it has no reason to be anywhere near. 

The extra copy here may cost some speed. We could instead preserve (legacy, unsafe) behaviour in 99% of cases by just testing `Union{Array, CuArray}`. Deleting Flux.Optimise soon should be a goal.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,10 +1,9 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.13.7"
+version = "0.13.8"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
-ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
@@ -27,7 +26,6 @@ Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
 Adapt = "3.0"
-ArrayInterface = "3.1, 4, 5, 6"
 CUDA = "3"
 ChainRulesCore = "1.12"
 Functors = "0.3"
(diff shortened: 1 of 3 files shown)


ToucheSir (approved):
I wonder if this inadvertently fixes https://github.com/FluxML/Flux.jl/issues/2045 as well. Either way LGTM.

mcabbott:
Timing this with [ResNet(18)](https://fluxml.ai/Optimisers.jl/dev/#Usage-with-[Flux.jl](https://github.com/FluxML/Flux.jl)) as an example gives about what you'd expect
```
)         # Total: 62 trainable arrays, 11_689_512 parameters,
          # plus 40 non-trainable, 9_600 parameters, summarysize 44.654 MiB.

julia> @time Flux.train!(Flux.params(model), [image], Flux.Adam()) do x
           sum(model(x))
       end
  3.872156 seconds (156.71 k allocations: 976.264 MiB, 53.12% gc time, 40.96% compilation time)  # before
  3.807214 seconds (156.79 k allocations: 1020.857 MiB, 52.45% gc time, 45.07% compilation time)  # after

julia> 1020.857 - 976.264
44.59299999999996

julia> @time Optimisers.update!(state, model, ∇model);
  0.069692 seconds (3.76 k allocations: 305.172 KiB)

julia> @time Optimisers.update(state, model, ∇model);  # this copies model + adam's 2 momenta
  0.325004 seconds (6.62 k allocations: 134.385 MiB, 81.20% gc time)
```

PR #2099, opened 2022-11-03
Remove type restrictions on normalization layers' scale and bias

darsnack:
Right now, the normalization scale and bias are required to be the same type. This is an unnecessary restriction that isn't there for the weight and biases of other layers.

### PR Checklist

- [x] ~~Tests are added~~
- [x] ~~Entry in NEWS.md~~

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -322,10 +322,10 @@ julia> isapprox(std(m(xs)), 1, atol=0.1) && std(xs) != std(m(xs))
 true
 ```
 """
-mutable struct BatchNorm{F,V,U,N,W}
+mutable struct BatchNorm{F,B,S,N,W}
   λ::F  # activation function
-  β::V  # bias
-  γ::U  # scale
+  β::B  # bias
+  γ::S  # scale
   μ::W     # moving mean
   σ²::W    # moving var
   ϵ::N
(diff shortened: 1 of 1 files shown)


src/layers/normalise.jl
```diff
@@ -322,10 +322,10 @@ julia> isapprox(std(m(xs)), 1, atol=0.1) && std(xs) != std(m(xs))
 true
 ```
 """
-mutable struct BatchNorm{F,V,N,W}
+mutable struct BatchNorm{F,V,U,N,W}
   λ::F  # activation function
   β::V  # bias
-  γ::V  # scale
+  γ::U  # scale
```
ToucheSir, line 328: I don't know where the original type param names came from, but would it make sense to harmonize them with the actual parameter names? e.g. `B,S` instead of `U,V`.

mcabbott:
Should we consider wrapping these two in a Scale sub-layer?

ToucheSir:
It's complicated because backends like cuDNN fuse in the affine transform as well. One idea would be to extract the `scale` and `bias` params from the sub-layer in the BN/IN/GN forward pass, but that feels a tad hacky.

ToucheSir (approved):
LGTM!

CarloLucibello:

PR #2101, opened 2022-11-07
Support Vararg Chain (Chain of Parallel)

### LINKED ISSUE #2100: Chain(Parallel(...), ...)
### Describe the potential feature

When combining multiple inputs (or an embedding + inputs), one can combine them at the end, with `Parallel(+, Chain(...), Chain(...))`, but one can't combine them earlier: `Chain(Parallel(...), Dense(...))`. That code runs, but when called, there is no `(::Chain)(args...)` method. Only the [unary one](https://github.com/FluxML/Flux.jl/blob/b2b0748c0b37b8610d9b6358dff8a501b75b80ab/src/layers/basic.jl#L51).

In other words, if Chain is function composition, it should support 

```julia
julia> (sin ∘ cos ∘ +)(1, 2)
-0.8360218615377305
```

### Motivation

Helps with multi-modal inputs / embeddings.

### Possible Implementation

```julia
_applychain(layers::Tuple, args...) = _apply_chain(layers[2:end], layers[1](args...))
```
? https://github.com/FluxML/Flux.jl/pull/1809 has some background on why the generated function was needed for `_applychain`. If the above one-liner isn't zygote-friendly (... I really don't know), then I could modify the generated function.

cstjean:
Closes #2100

As mentionned in https://github.com/FluxML/Flux.jl/issues/2100#issuecomment-1305399770, this will break any code using `Chain()` as the identity function. I need a decision whether this is acceptable, or if I should special-case it.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -48,18 +48,21 @@ end
 
 @functor Chain
 
-(c::Chain)(x) = _applychain(c.layers, x)
+(c::Chain)(inputs...) = _applychain(c.layers, inputs...)
 
-@generated function _applychain(layers::Tuple{Vararg{<:Any,N}}, x) where {N}
-  symbols = vcat(:x, [gensym() for _ in 1:N])
-  calls = [:($(symbols[i+1]) = layers[$i]($(symbols[i]))) for i in 1:N]
-  Expr(:block, calls...)
+@generated function _applychain(layers::Tuple{Vararg{<:Any,N}}, inputs...) where {N}
+  symbols = [gensym() for _ in 1:N]
+  calls = [:($(symbols[i]) = layers[$i]($(symbols[i-1]))) for i in 2:N]
+  Expr(:block,
+       :($(symbols[1]) = layers[1](inputs...)),
+       calls...)
 end
 
-_applychain(layers::NamedTuple, x) = _applychain(Tuple(layers), x)
+_applychain(layers::NamedTuple, inputs...) = _applychain(Tuple(layers), inputs...)
 
-function _applychain(layers::AbstractVector, x)  # type-unstable path, helps compile times
-  for f in layers
+function _applychain(layers::AbstractVector, inputs...)  # type-unstable path, helps compile times
+  x = layers[1](inputs...)
+  for f in @view(layers[2:end])
     x = f(x)
   end
   x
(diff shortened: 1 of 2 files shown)


ToucheSir:
This reminded me that we discussed multi-input `Chain`s just over 2 years ago, see discussion starting [here](https://julialang.zulipchat.com/#narrow/stream/237432-ml-contributers/topic/FluxTrax.3F/near/215036677). One unresolved issue is whether users will expect `Chain(twoargfn, twoargfn)` to work if `Chain(twoargfn)` does. It does not feel like we have enough information to know whether it is safe to splat the output of one layer into another.

darsnack:
This was also discussed in great detail in #1698 where there was a desire to remove `(m::Parallel)(xs::Tuple) = m(xs...)`. There I suggested that this PR was necessary to make that happen.

A couple of details that are relevant from that discussion:
- the Julia convention for multiple outputs is a `Tuple` with explicit splatting to downstream functions
- A `Splat` layer is an option for convenient explicit splatting that avoids needing a `(m)(xs::Tuple) = m(xs...)` for every multi-arg layer

The issue (#2100) is a little ambiguous. You can insert `Parallel` at the end or the middle, and it "just" works. Only inserting at the very start of a `Chain` breaks the `Vararg` variant. If that's what's missing, then a simpler change:
```julia
(m::Chain)(xs...) = _applychain(m.layers, xs)
```
would work. This still keeps `Chain` as 1-in-1-out from passing arguments between layers, but optionally auto-tuplifies a `Vararg` input.

cstjean:
I'm not sure how to read your message: is it "we don't know yet whether something like this PR ought to be merged?" Or is it, "this PR is fine, since it's just the first layer, but it's not clear what to do about multi-argument in the middle layers"?

Is there some debate about what the first layer ought to do, when provided with multiple arguments? 

If not this PR, what would be the standard way of using an Embedding along with other inputs? I've been scratching my head and couldn't figure it out. Of course I could have `Parallel(+, inputs=Chain(...), embed=Chain(Embedding(...)))`, but that severely constrains the type of model I can build (no interaction between the embedding and the inputs).

cstjean:
> (m::Parallel)(xs::Tuple) = m(xs...)

Is this documented anywhere? If not, I can make a PR. It's a crucial part for combining embeddings and inputs.

In any case, then I agree with either the status quo, or the

```julia
(m::Chain)(xs...) = _applychain(m.layers, xs)
```

proposal.

darsnack:
> Is this documented anywhere?

Surprisingly the `Tuple <=> Vararg` equivalence is not in the docstring for `Parallel`. It should be.

My initial comment was just to surface prior discussions on the topic. Having had the change to review those discussions in more detail, I think we can condense to two options:
1. `Chain` is MIMO (i.e. it accepts `Vararg` and always splats the intermediate arguments). In this option, multiple input-outputs are automatically handled.
2. `Chain` is SISO (status quo). Multiple arguments are explicitly handled via one of two sub-options:
    a. Each multiple argument layer must add a `(m)(xs::Tuple)` dispatch
    b. We have `Splat` to make it explicit via the user instead of the layer author

In the case of (2), the `(m::Chain)(xs...) = _applychain(m.layers, xs)` method can be added for convenience.

(2a) is what we currently have, and I am okay with all options. Maybe the other maintainers can weigh in.

darsnack:
@cstjean were you planning on opening another PR with the convenience method discussed above?

cstjean:
Sure, I can do that.

PR #2110, opened 2022-11-11
Remove `train!` from quickstart example

mcabbott:
As suggested in #2104, this replaces `train!` with an explicit loop in the quickstart example. Also steals the idea from there about showing logging code.

It does not completely scrap the example, which otherwise tries to make quite a few points carefully & concisely. Not sure it fits on one screen anymore, though.


MilesCranmer:
LGTM! Although I would suggest you just remove the README example as it is intimidating and unclear. If the goal is to get the user to the docs, I would just leave the API a surprise, so they see it in the best light.

docs/src/models/quickstart.md
```diff
@@ -18,26 +18,36 @@ model = Chain(Dense(2 => 3, tanh), BatchNorm(3), Dense(3 => 2), softmax)
...
+            # Evaluate model and loss inside gradient context:
+            y_hat = model(x)
+            Flux.crossentropy(y_hat, y)  # could use just sum(abs2, y_hat .- y)
+        end
+        # Use the gradient to update the model's parameters (and momentum):
+        Flux.update!(opt, pars, grad)
+        # Logging code, outside gradient context:
+        push!(losses, loss)
+    end
+    if isinteger(log2(epoch))
```
MilesCranmer, line 43: A bit too non-traditional to use in a quickstart. Users are going to be pattern-matching based on their experience with PyTorch/JAX, so better to keep things in traditional form. I would just do something simple like `epoch == 1 || (epoch - 1) % 50 == 0`.

mcabbott: Maybe. Or they paste it in and learn something:
```
after epoch 1, loss is 0.7743482
after epoch 2, loss is 0.68434924
after epoch 4, loss is 0.650178
after epoch 8, loss is 0.49720597
after epoch 16, loss is 0.30286682
after epoch 32, loss is 0.20718074
after epoch 64, loss is 0.15182666
after epoch 128, loss is 0.13036121
after epoch 256, loss is 0.17226303
after epoch 512, loss is 0.11514424
```
I did `if div(epoch, 100) == 0` first, but then you see nothing, it looks like the loss isn't changing.

MilesCranmer: Up to you, but so many DL examples just use `epoch % num == 0` that people are conditioned to wait for training to complete, so I wouldn't worry about it. I would just try to use as few functions as possible so people can match their knowledge of PyTorch/JAX to Flux.jl as quickly as possible.

> Or they paste it in and learn something:

They already have a lot to learn from a quickstart example; best not to overwhelm them with functions they don't need yet.

MilesCranmer:
For the start example, I would some syntax changes to make things more idiomatic to newcomers to Flux.jl/Julia who might have previous DL experience (probably in Python). In due time they will learn `map`, `...`, `eachcol`, but I think they don't need it all at once.

```julia
# Generate some data for the XOR problem: vectors of length 2, as columns of a matrix:
noisy = rand(Float32, 2, 1000)                                    # 2×1000 Matrix{Float32}
truth = map(col -> xor(col...), eachcol(noisy .> 0.5))            # 1000-element Vector{Bool}

# Define our model, a multi-layer perceptron with one hidden layer of size 3:
model = Chain(Dense(2 => 3, tanh), BatchNorm(3), Dense(3 => 2), softmax)

# The model encapsulates parameters, randomly initialised. Its initial output is:
out1 = model(noisy)                                               # 2×1000 Matrix{Float32}
```

into this:

```julia
# Generate some data for the XOR problem: vectors of length 2, as columns of a matrix:
noisy = rand(Float32, 2, 1000)                                    # 2×1000 Matrix{Float32}
truth = [xor(noisy[1, i] > 0.5, noisy[2, i] > 0.5) for i=1:1000]  # 1000-element Vector{Bool}

# Define our model, a multi-layer perceptron with one hidden layer of size 16:
model = Chain(
    Dense(2 => 16), tanh,
    Dense(16 => 16), tanh,
    Dense(16 => 2), softmax
)

# The model encapsulates parameters, randomly initialised. Its initial output is:
out1 = model(noisy)                                               # 2×1000 Matrix{Float32}
```

(keeping the activation outside the layer is more idiomatic, showing it inside makes me confused about what I can put inside `Chain`. They'll learn later on that they can do this though.)

docs/src/models/quickstart.md
```diff
@@ -18,26 +18,36 @@ model = Chain(Dense(2 => 3, tanh), BatchNorm(3), Dense(3 => 2), softmax)
...
 out1 = model(noisy)                                               # 2×1000 Matrix{Float32}
 
-# To train the model, we use batches of 64 samples:
-mat = Flux.onehotbatch(truth, [true, false])                      # 2×1000 OneHotMatrix
-data = Flux.DataLoader((noisy, mat), batchsize=64, shuffle=true);
-first(data) .|> summary                                           # ("2×64 Matrix{Float32}", "2×64 Matrix{Bool}")
+# To train the model, we use batches of 64 samples, and one-hot encoding:
+target = Flux.onehotbatch(truth, [true, false])                   # 2×1000 OneHotMatrix
+loader = Flux.DataLoader((noisy, target), batchsize=64, shuffle=true);
+first(loader) .|> summary                                         # ("2×64 Matrix{Float32}", "2×64 Matrix{Bool}")
```
MilesCranmer, line 24: More minimization of syntax:
```suggestion
example_x, example_y = first(loader)
summary(example_x), summary(example_y)   # ("2×64 Matrix{Float32}", "2×64 Matrix{Bool}")
```

ToucheSir:
I think it best not to put the activation functions outside of their associated layers, because Julia assigns special meanings to some of them beyond `fn(::Array) == fn.(::Array)`:

```julia
julia> x = rand(2, 2)
2×2 Matrix{Float64}:
 0.479567  0.200181
 0.719861  0.0244788

julia> tanh(x)
2×2 Matrix{Float64}:
 0.409203  0.178483
 0.641832  0.00344357

julia> tanh.(x)
2×2 Matrix{Float64}:
 0.445897  0.197549
 0.616823  0.0244739
```

docs/src/models/quickstart.md
MilesCranmer: (I would use `epoch == 1 || epoch % num == 0` so that they see the loss on the first epoch - then it's clear it is going down)

MilesCranmer:
> softmax however should stay outside, because it only comes as a "vectorized" function (makes no sense to apply element-wise). This also clues users into the difference between activation functions and functional layers.

Good point! Okay fine with me to keep them inside.

docs/src/models/quickstart.md
mcabbott: The new version of DataLoader will print this nicely, and perhaps the comment is not needed (or can reflect that).

MilesCranmer:
I would implement this change though:

```diff
- truth = map(col -> xor(col...), eachcol(noisy .> 0.5))            # 1000-element Vector{Bool}
+ truth = [xor(noisy[1, i] > 0.5, noisy[2, i] > 0.5) for i=1:1000]  # 1000-element Vector{Bool}
```

`map`, `...`, `eachcol`, `.>` - just too many things to learn in a single line. (In all honesty, I didn't even know `eachcol` myself until now, despite developing packages in Julia for a couple of years...!)

mcabbott:
Worse, the model proposed doesn't run:
```
julia> # The model encapsulates parameters, randomly initialised. Its initial output is:
       out1 = model(noisy)                                               
ERROR: DimensionMismatch: matrix is not square: dimensions are (16, 1000)
Stacktrace:
 [1] checksquare
   @ ~/.julia/dev/julia/usr/share/julia/stdlib/v1.9/LinearAlgebra/src/LinearAlgebra.jl:239 [inlined]
```
`tanh` acting on a matrix already means something. (This is why I was unhappy about making the others auto-broadcast.)

MilesCranmer:
> `tanh` acting on a matrix already means something. (This is why I was unhappy about making the others auto-broadcast.)

Scary indeed! Maybe a warning should be printed if `tanh` or `exp` is passed nakedly to `Chain`? (since it would be really weird for a user to do this and actually mean it?) If I had set the batchsize equal to the number of features, it wouldn't have errored and I might never have realized this.

README.md
```diff
@@ -18,12 +18,12 @@
...
 Flux is an elegant approach to machine learning. It's a 100% pure-Julia stack, and provides lightweight abstractions on top of Julia's native GPU and AD support. Flux makes the easy things easy while remaining fully hackable.
 
-Works best with [Julia 1.8](https://julialang.org/downloads/) or later. Here's a simple example to try it out:
+Works best with [Julia 1.8](https://julialang.org/downloads/) or later. Here's a very short example to try it out:
 ```julia
 using Flux  # should install everything for you, including CUDA
 
 x = hcat(digits.(0:3, base=2, pad=2)...) |> gpu  # let's solve the XOR problem!
-y = Flux.onehotbatch(xor.(eachrow(x)...), 0:1) |> gpu
+y = Flux.onehotbatch(xor.(eachrow(cpu(x))...), 0:1) |> gpu
```
MilesCranmer, line 26: Why `cpu(x) |> gpu`? Are some of these functions not implemented on GPU?

mcabbott: This gives scalar indexing warnings if run on the GPU.

MilesCranmer: Maybe just move them both to the GPU when creating the dataloader?

mcabbott: I tried a few things, but they are all ugly, or don't work. Maybe simpler to give up on squeezing `|> gpu` into this model.

MilesCranmer: Sounds good to me. What will explicit parameters do?

mcabbott: Something like this, no more `params`:
```julia
model = Chain(Dense(2 => 3, sigmoid), BatchNorm(3), Dense(3 => 2))
optim = Flux.setup(Adam(0.1, (0.7, 0.95)), model)

for _ in 1:100
    grad = gradient(m -> Flux.logitcrossentropy(m(x), y), model)[1]
    Flux.update!(optim, model, grad)  # this changes model & optim
end
```

MilesCranmer: Very clean! Multiple dispatch rules 🚀

docs/src/models/quickstart.md
```diff
@@ -34,15 +34,13 @@ opt = Flux.Adam(0.01)      # will store optimiser momentum, etc.
 for epoch in 1:1_000
     losses = []
     for (x, y) in loader
-        loss, grad = Flux.withgradient(pars) do
+        loss, grad = Zygote.withgradient(pars) do
```
MilesCranmer, line 37: Why this change? For ease of starting I think the fewer modules the better. Even if the desire is that they learn they could take gradient with an arbitrary package, I think that could just be introduced later.

README.md
mcabbott: I couldn't find a version of this model which was both compact and clear. So I gave up & replaced it with a curve-fitting problem, [918fc0b](https://github.com/FluxML/Flux.jl/pull/2110/commits/918fc0b359b8d836a5cc4d6bf1a2591553e99e4b). 

But IDK, maybe we should give up & go back to having no example in the readme.

README.md
```diff
@@ -18,23 +18,27 @@
 
 Flux is an elegant approach to machine learning. It's a 100% pure-Julia stack, and provides lightweight abstractions on top of Julia's native GPU and AD support. Flux makes the easy things easy while remaining fully hackable.
 
-Works best with [Julia 1.8](https://julialang.org/downloads/) or later. Here's a simple example to try it out:
+Works best with [Julia 1.8](https://julialang.org/downloads/) or later. Here's a very short example to try it out:
 ```julia
-using Flux  # should install everything for you, including CUDA
+using Flux, Plots
+data = [([x], x-cbrt(x)) for x in range(-2, 2, 100)]
```
MilesCranmer, line 24:
```suggestion
data = [([x], x - x^2) for x in range(-2, 2, 100)]
```

Something simpler (In all honesty I have never seen `cbrt` before... wasn't sure what it was)

README.md
```diff
@@ -18,23 +18,27 @@
...
+loss(x,y) = abs2(model(x) - y)
+optim = Flux.Adam()
+for epoch in 1:1000
+  Flux.train!(loss, Flux.params(model), data, optim)
+end
 
-Flux.train!(mloss, Flux.params(model), data, optim)  # updates model & optim
-
-all((softmax(model(x)) .> 0.5) .== y)  # usually 100% accuracy.
+plot(x -> x-cbrt(x), -2, 2, legend=false)
```
MilesCranmer, line 34:
```suggestion
plot(x -> x - x^3, -2, 2, legend=false)
```

README.md
```diff
@@ -18,23 +18,27 @@
...
 
-x = hcat(digits.(0:3, base=2, pad=2)...) |> gpu  # let's solve the XOR problem!
-y = Flux.onehotbatch(xor.(eachrow(x)...), 0:1) |> gpu
-data = ((Float32.(x), y) for _ in 1:100)  # an iterator making Tuples
+model = Chain(Dense(1 => 10, tanh), Dense(10 => 1), only)
 
-model = Chain(Dense(2 => 3, sigmoid), BatchNorm(3), Dense(3 => 2)) |> gpu
-optim = Adam(0.1, (0.7, 0.95))
-mloss(x, y) = Flux.logitcrossentropy(model(x), y)  # closes over model
+loss(x,y) = abs2(model(x) - y)
```
MilesCranmer, line 28:
```suggestion
loss(x,y) = (model(x) - y) ^ 2
```

README.md
MilesCranmer: I like the 
```
    grad = gradient(m -> Flux.logitcrossentropy(m(x), y), model)[1]
    Flux.update!(optim, model, grad)  # this changes model & optim
```
example a lot. Why not just show this one?

README.md
mcabbott: I guess the point is to make a pretty curve which isn't obviously trivial to fit:

Screenshot 2022-11-24 at 21 19 24

Does it matter if the function isn't immediately familiar?

README.md
MilesCranmer: (or the equivalent, but with the MSE loss)

README.md
MilesCranmer: Ah, sorry, I edited my comment to be `x - x^3`. I agree `x - x^2` looks way too simple.

MilesCranmer: (When I read `cbrt` I wasn't sure if the output was scalar or not in the example - whereas `^` is more clear to me)

mcabbott: Now with `2x-x^3` 
Screenshot 2022-11-24 at 21 43 33

mcabbott: > wasn't sure if the output was scalar or not in the example

This example is a little odd in that it takes a trivial vector input, and returns a scalar. As it makes the loss nice and simple. Maybe that introduces confusion though? E.g. 

https://discourse.julialang.org/t/input-to-neural-network/90770

MilesCranmer: I was going to suggest that it take two features as input instead, like `f(x, y) = 2x - y^3` over $(-2:2) \times (-2:2)$. However the downside is it would be harder to visualize (I think a heatmap is good too, though).

mcabbott: Yes. I think if this readme thing exists at all, it needs to be super-simple. It must confirm that Flux is running on your system, and solve a vaguely NN-like problem (not linear regression).

I think the present XOR one invites too much time spent figuring it out, when people should rather go into the docs.

mcabbott: (The rule BTW around here is that PRs need one approval to merge, not necc. from anyone with rights.)

docs/src/models/quickstart.md
CarloLucibello: `isinteger(log2(epoch))` is quite autoesplicative, and on the other hand `epoch % 50 == 0` is so well known and similar to that in other languages that we don't have to teach it

mcabbott: Now deleted, maybe it's better to save a few lines and save distraction.

docs/src/models/quickstart.md
mcabbott: There's now just a comment, without any code like `.|> summary`

docs/src/models/quickstart.md
```diff
@@ -6,45 +6,54 @@ If you haven't, then you might prefer the [Fitting a Straight Line](overview.md)
...
-model = Chain(Dense(2 => 3, tanh), BatchNorm(3), Dense(3 => 2), softmax)
+model = Chain(
+    Dense(2 => 3, tanh),   # activation function inside layer
+    BatchNorm(3),
+    Dense(3 => 2),
+    softmax) |> gpu        # move model to GPU, if available
 
 # The model encapsulates parameters, randomly initialised. Its initial output is:
-out1 = model(noisy)                                               # 2×1000 Matrix{Float32}
+out1 = model(noisy |> gpu) |> cpu                                 # 2×1000 Matrix{Float32}
```
mcabbott, line 23: Recent commit runs things on the GPU, as that seems worth showing off. (Even though this is actually slower). One quirk is that `model(noisy |> gpu) |> cpu` is a bit noisy, but maybe not so confusing to figure out.

MilesCranmer: Yeah this looks fine to me.

MilesCranmer (approved):
LGTM!

Couple minor comments which I'll leave up to you:

- I'd either (1) upload images to a separate docs-specific repo, or (2) upload the image file directly to GitHub rather than the git repo (you can do this by drag-and-dropping an image into the markdown), so it doesn't weight down the git history with binary files. 
- Use a slightly large neural net in the README, so that the fit is better:
```julia
model = Chain(Dense(1 => 23, tanh), Dense(23 => 23, tanh), Dense(23 => 1, bias=false), only)
```
does the trick.

mcabbott:
> either (1) upload images to a separate docs-specific repo, or (2) upload the image file directly to GitHub rather than the git repo (you can do this by drag-and-dropping an image into the markdown), so it doesn't weight down the git history 

This is a good point. I have unfortunately committed sins to the tune of 1.5MB in #2125 , and if we want more tutorials in here we probably need a plan. This PR is 4% of that (the larger image is already here) so my inclination is to kick the can down the road for now.

MilesCranmer:
Sounds good to me! (I'm a sinner as well; I recently realized some of my repositories were getting large because of it and have started slowly changing this habit)

mcabbott:
Cool, many thanks for taking a look. 

If you have a minute, your take on #2114 would also be welcomed.

PR #2114, opened 2022-11-19
Re-write training docs

mcabbott:
This re-writes `training.md` to move from implicit to explicit, and creates a new page `train_api.md` to hold the docstrings. For now they live together but the goal is that they can live in different sections, "narrative vs. lookup". 

Also updates basic.md to introduce explicit first.

In both of these, it puts implicit things in `!!! note` boxes nearby. I start to think that's better than having a separate page for implicit-to-explicit changes.

It expects #2082 in order for `update!` and `Adam` to work with explicit gradients, and for the new `train!`.

After this I think regularisation.md needs some re-working. It could be a home for some of what's in   optimisers.md (except the listing of rules), basically all the more advanced topics after this. But not this PR.


mcabbott:
I think this is about ready. 

It could use another set of eyeballs to see whether anything from the old docs has been lost, or made more confusing.

docs/src/training/training.md
```diff
@@ -1,210 +1,387 @@
...
+direction in various clever ways.
+Most require some memory of the gradients from earlier steps, rather than always
+walking straight downhill -- [`Momentum`](@ref Flux.Optimise.Momentum) is the simplest.
+The function [`setup`](@ref Flux.Train.setup) creates the necessary storage for this, for a particular model.
+It should be called once, before training, and returns a tree-like object which is the
+first argument of `update!`. Like this:
+
+```julia
+# Initialise momentum 
+opt = Flux.setup(Momentum(0.01, 0.9), model)
```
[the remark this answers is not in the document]
mcabbott: Changed to state everywhere in [11e4825](https://github.com/FluxML/Flux.jl/pull/2114/commits/11e4825cbc1894de67da8055f1bffa2fbf3988c2)

README.md
```diff
@@ -25,10 +25,9 @@ data = [([x], 2x-x^3) for x in -2:0.1f0:2]
 
 model = Chain(Dense(1 => 23, tanh), Dense(23 => 1, bias=false), only)
 
-mloss(x,y) = (model(x) - y)^2
-optim = Flux.Adam()
+state = Flux.setup(Adam(), model)
```
CarloLucibello, line 28: `state` is too generic, should we call this `optstate` or just `opt`?

docs/src/models/basics.md
```diff
@@ -29,35 +29,77 @@ julia> gradient(f, [2, 1], [2, 0])
...
-Machine learning often can have *hundreds* of parameters, so Flux lets you work with collections of parameters, via the `params` functions. You can get the gradient of all parameters used in a program without explicitly passing them in.
+Machine learning often can have *hundreds* of parameter arrays.
+Instead of passing them to `gradient` individually, we can store them together in a structure.
+The simplest example is a named tuple, created by the following syntax:
 
 ```jldoctest basics
-julia> x = [2, 1];
+julia> nt = (a = [2, 1], b = [2, 0], c = abs2);
+
+julia> g(x::NamedTuple) = sum(abs2, x.a .- x.b);
```
CarloLucibello, line 41: Was this meant to be
```suggestion
julia> g(x::NamedTuple) = sum(x.c, x.a .- x.b);
```
?

README.md
mcabbott: I just changed it from `opt`, in the name of emphasising that it's not the same thing as before: https://github.com/FluxML/Flux.jl/pull/2114#discussion_r1037581442

mcabbott: Now I don't like any of them. I changed it from `opt` after trying to straighten out someone's code, which was a muddle of old & new... so much old code has `opt = Adam()` etc.

But I agree "state of what?" is a problem. 

Maybe the main training page should have something long & explicit... at least while it has all these "how to upgrade" info boxes.

ToucheSir: Looking around at what terminology others use:
- Optax uses [`opt_state`](https://github.com/deepmind/optax), which is probably where I got the idea from.
- The lower-level, function forms of PyTorch Optimizers make users pass in every piece of state separately ([e.g.](https://github.com/pytorch/pytorch/blob/9eccfedca246a8b2c81d8f33c0664fa66a73636a/torch/optim/adam.py#L281)). Their class-based Optimizer API just follows the same [`state_dict`](https://pytorch.org/docs/stable/generated/torch.optim.Optimizer.state_dict.html#torch.optim.Optimizer.state_dict) based paradigm that layers do, i.e. `opt.state_dict()["state"]`.
- The only other library I could think of with this kind of FP-inspired Optimizer API is Axon, which uses the even more verbose [`optimizer_state`](https://hexdocs.pm/axon/Axon.Optimizers.html).

mcabbott: Thanks for looking these up. I've changed it to `opt_state` now.

docs/src/models/basics.md
mcabbott: I don't remember. Maybe I thought `sum(f, stuff)` is slightly obscure and adding a reference to a function makes is worse? 

Maybe the value of `x.c` should just be `tanh` or `42` or something, all that matters is that it has no gradient.

mcabbott: For now I made it tanh.

Are there other concerns before merging?

CarloLucibello (approved):

Issue #2091, opened 2022-10-20
Iteration over `params(m)` in explicit mode gives no gradient


mcabbott:
The use of `params` within explicit-mode gradients has been broken by #2054 . Sadly there were no tests of this:
```julia
julia> m = (x=[1,2.0], y=[3.0]);

julia> gradient(m -> (sum(norm, Flux.params(m))), m)
((x = [0.4472135954999579, 0.8944271909999159], y = [1.0]),)  # before, Flux@0.13.5
(nothing,)  # after, Flux@0.13.6

julia> gradient(() -> (sum(norm, Flux.params(m))), Flux.params(m))
Grads(...)

julia> ans[m.x]  # unchanged
2-element Vector{Float64}:
 0.4472135954999579
 0.8944271909999159
 ```

(This form really is very annoying. Suggestions are fine but required textboxes seem to be solving a problem we didn't have. Also, it seems to break links to issues, like #2054 above. Edited to delete the boilerplate.)

ToucheSir:
Is there a smart rule we could write which avoids #2054? I think the main considerations would be how map `trainable` pullbacks to the original tree structure and the final params vector. Maybe maintaining an auxiliary tree structure with indices at the leaves?

mcabbott:
This is surely possible, someone just has to write one more Functors thing which builds this gradient tree, as the `rrule` for `params(m)`.

PR #2118, opened 2022-11-21
Narrower version of `@non_differentiable params`

mcabbott:
This replaces #2054 with a narrower definition, to fix https://github.com/FluxML/Flux.jl/issues/2091

Exploits https://github.com/FluxML/Zygote.jl/pull/1248 to essentially apply `@no_grad` only to implicit mode. 

@christiangnrd could you check that this doesn't re-introduce #2040? I am not getting consistent times but not sure why.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -38,7 +38,7 @@ ProgressLogging = "0.1"
 Reexport = "0.2, 1.0"
 SpecialFunctions = "1.8.2, 2.1.2"
 StatsBase = "0.33"
-Zygote = "0.6.34"
+Zygote = "0.6.49"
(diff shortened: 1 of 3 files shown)


christiangnrd:
#2040 is not reintroduced with this commit.

Cheers!

christiangnrd:
Seems like #2123 is what fixes the failing test ad it was merged shortly after the CI was triggered.

mcabbott:
Thanks! It passed initially, except for what was probably just buildkite having a bad day... now rebased to try again.

ToucheSir (approved):

PR #2119, opened 2022-11-21
allow non-tuple data in the new train!

CarloLucibello:
Follow up to #2082 with ~two~ changes to the new `train!`:
- allows for non-tuple elements in the data iterators (same as the `batchmemaybe` function of the old `train!`)
according to  https://github.com/FluxML/Flux.jl/pull/2082#issuecomment-1319559519
- ~support callbacks~

I can factor out the second change if deemed controversial. Edit: removed the callback addition.

### DIFF
--- a/src/train.jl
+++ b/src/train.jl
@@ -56,8 +56,8 @@ end
 
 Uses a `loss` function and training `data` to improve the `model`'s parameters
 according to a particular optimisation rule `opt`. Iterates through `data` once,
-evaluating `loss(model, d...)` for each `d` in `data` in case of tuple iterates, 
-and `loss(model, d)` otherwise.
+evaluating for each `d in data` either `loss(model, d...)` if `d isa Tuple`,
+or else   `loss(model, d)` for other `d`.
 
 For example, with these definitions...
 ```
@@ -97,11 +97,11 @@ It adds only a few features to the loop above:
       But any code can be included in the above `for` loop.
 """
 function train!(loss, model, data, opt; cb = x -> nothing)
-  cb = old_cb_deprecation(cb)
   cb = runall(cb)
+  @show cb
   @withprogress for (i,d) in enumerate(data)
-    ds = batchmemaybe(d)
-    l, gs = Zygote.withgradient(m -> loss(m, ds...), model)
+    d_splat = d isa Tuple ? d : (d,)
+    l, gs = Zygote.withgradient(m -> loss(m, d_splat...), model)
     cb((; model, data=d, opt, step=i, loss=l, gradient=gs[1]))
     if !isfinite(l)
       throw(DomainError("Loss is $(l) on data item $i, stopping training"))
@@ -112,33 +112,14 @@ function train!(loss, model, data, opt; cb = x -> nothing)
 end
 
 # This method let you use Optimisers.Descent() without setup, when there is no state
-function train!(loss, model, data, rule::Optimisers.AbstractRule; cb=x->nothing)
+function train!(loss, model, data, rule::Optimisers.AbstractRule; cb = x -> nothing)
   train!(loss, model, data, _rule_to_state(model, rule); cb)
 end
 
-batchmemaybe(x) = tuple(x)
-batchmemaybe(x::Tuple) = x
-
 call(f, xs...) = f(xs...)
 runall(f) = f
-runall(fs::AbstractVector) = x -> foreach(call, fs, x)
 
-old_cb_deprecation(f::AbstractVector) = [old_cb_deprecation(f) for f in f]
-
-function old_cb_deprecation(f)
-  return x -> begin
-    try
-      f(x)
-    catch e
-      if e isa MethodError
-        @warn "Callback functions must accept a named tuple argument. See the docs for `train!`."
-        f()
-      else
-        rethrow(e)
-      end
-    end
-  end
-end
+runall(fs::AbstractVector) = x -> [f(x) for f in fs]
 
 function _rule_to_state(model, rule::Optimisers.AbstractRule)
   state = setup(rule, model)
--- a/src/train.jl
+++ b/src/train.jl
@@ -57,7 +57,7 @@ end
 Uses a `loss` function and training `data` to improve the `model`'s parameters
 according to a particular optimisation rule `opt`. Iterates through `data` once,
 evaluating for each `d in data` either `loss(model, d...)` if `d isa Tuple`,
-or else   `loss(model, d)` for other `d`.
+or else `loss(model, d)` for other `d`.
 
 For example, with these definitions...
 ```
--- a/src/train.jl
+++ b/src/train.jl
@@ -61,7 +61,7 @@ or else   `loss(model, d)` for other `d`.
 
 For example, with these definitions...
 ```
-data = [(x1, y1), (x2, y2), (x3, y3)]  
+data = [(x1, y1), (x2, y2), (x3, y3)]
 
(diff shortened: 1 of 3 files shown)


src/train.jl
```diff
@@ -88,32 +89,53 @@ It adds only a few featurs to the loop above:
       (This is to move away from Zygote's "implicit" parameter handling, with `Grads`.)
     * Instead of `loss` being a function which accepts only the data,
       now it must also accept the `model` itself, as the first argument.
-    * `data` must iterate tuples, otherwise you get an error.
-      (Previously non-tuple types were not splatted into the loss. 
-      Pass in `((d,) for d in data)` to simulate this.)
+    * If `data` iterates over tuples, these will be splatted when passed to `loss`.
+      If you want to avoid the splatting, you can pass `((d,) for d in data)` instead.
```
mcabbott, line 93: This should not be in the list of changes
```suggestion
```

src/train.jl
```diff
@@ -52,15 +52,16 @@ function setup(rule::Optimisers.AbstractRule, model)
...
 
 """
-    train!(loss, model, data, opt)
+    train!(loss, model, data, opt; cb=nothing)
 
 Uses a `loss` function and training `data` to improve the `model`'s parameters
 according to a particular optimisation rule `opt`. Iterates through `data` once,
-evaluating `loss(model, d...)` for each `d` in data.
+evaluating `loss(model, d...)` for each `d` in `data` in case of tuple iterates, 
+and `loss(model, d)` otherwise.
```
mcabbott, line 60: I think this is pretty hard to understand if you don't already know about this weird thing, what's a tuple iterator? Maybe:
```suggestion
evaluating for each `d in data` either `loss(model, d...)` if `d isa Tuple`,
or else   `loss(model, d)` for other `d`.
```

src/train.jl
```diff
@@ -88,32 +89,53 @@ It adds only a few featurs to the loop above:
...
-  isnothing(cb) || error("""train! does not support callback functions.
-                            For more control use a loop with `gradient` and `update!`.""")
+  cb = old_cb_deprecation(cb)
+  cb = runall(cb)
   @withprogress for (i,d) in enumerate(data)
-    d isa Tuple || error("""train! expects as data an iterator producing tuples, but got $(typeof(d)).
-                            Pass it `((d,) for d in data)`, or use `gradient` and `update!` for more control.""")
-    l, gs = Zygote.withgradient(m -> loss(m, d...), model)
+    ds = batchmemaybe(d)
+    l, gs = Zygote.withgradient(m -> loss(m, ds...), model)
```
mcabbott, line 104: Clearer to directly do what this does? Any maybe the name should not echo `gs` which is sort-of plural...
```suggestion
    d_splat = d isa Tuple ? d : (d,)
    l, gs = Zygote.withgradient(m -> loss(m, d_splat...), model)
```

src/train.jl
```diff
@@ -88,32 +89,53 @@ It adds only a few featurs to the loop above:
...
 
 # This method let you use Optimisers.Descent() without setup, when there is no state
-function train!(loss, model, data, rule::Optimisers.AbstractRule)
-  train!(loss, model, data, _rule_to_state(model, rule))
+function train!(loss, model, data, rule::Optimisers.AbstractRule; cb=nothing)
+  train!(loss, model, data, _rule_to_state(model, rule); cb)
+end
+
+batchmemaybe(x) = tuple(x)
+batchmemaybe(x::Tuple) = x
```
mcabbott, line 120: Can we kill this batchmaybe function? I know a few people found it pretty confusing. And it has nothing to do with batching. 
```suggestion
```

mcabbott (commented):
I think this maybe-splat behaviour is a little odd, but see the argument for keeping it in the name of making upgrades easier. Comments below are only on this.

Besides the headline change, this PR not only re-introduces callbacks, but introduces a larger more complicated callback scheme. Do we want this, or to push people towards writing a `for` loop?

CarloLucibello:
> Besides the headline change, this PR not only re-introduces callbacks, but introduces a larger more complicated callback scheme. Do we want this, or to push people towards writing a for loop?

I simplified the scheme, compared to the old scheme it is more useful and has the same complexity. 
I can factor out the callback change into another PR. Or also ditch it, I'm not sure myself it is something we want to do.

mcabbott:
I agree that a callback which gets some details is more useful; the old-style ones where you wrap the global model seem more in keeping with the implicit style. 

Passing this as one NamedTuple is probably a better choice than splatting it as keywords, as your function can then take what it needs.

One downside is that it does mean many more things need names, in code not just in docs. The present list is `(model, data, opt, step, loss, gradient)`, that's a lot. And is "opt" an optimiser or a state or a tree, the documentation isn't completely consistent... but so far its name never matters much.

Maybe it would be clearer to call this `train!(loss, m, data, opt; callback)` with a different keyword. Then there's less confusion between the old and new things.  The old `cb` is a pretty cryptic name; it could survive for now to accept a zero-arg function with a depwarn.

darsnack:
Regarding callbacks, I think it's more of a design/community issue than anything else. The proposed callback scheme does provide all the state in the loop to the user. But there is already a difficulty: as written, the callback happens before the optimizer update, wouldn't `opt` and `model` be more useful after the update? I see two options:

1. We provide a callback as proposed here but called at the end of each loop iteration. We take a hard stance on declining to add more entry points into the loop. We do not provide built-in callbacks. This feature is described as a "quick and dirty" callback for when you do not want to refactor your code.
2. We delete callbacks completely. I don't think the old style (close over globals) callbacks should remain as we transition to explicit. They are incompatible with immutable models, and even though most models are/will be mutable, I think it is confusing to have a schism in behavior.

This is why I describe it as a community issue, because the issue is preventing feature creep. (2) does this neatly whereas (1) will always have users requesting more.

mcabbott:
Re before/after, worth remembering the `Flux.skip` debacle, where apparently it was so confusing and so code-golfed that it got documented in a way which didn't actually skip anything, and tests didn't notice.

Whereas `test() && continue` in a loop does exactly what it looks like it'll do.

CarloLucibello:
> But there is already a difficulty: as written, the callback happens before the optimizer update, wouldn't opt and model be more useful after the update? 

my reasoning was that executing the callback before the update you are able to implement things such as gradient clipping

mcabbott:
For simple clipping we of course have [ClipGrad](https://fluxml.ai/Optimisers.jl/dev/api/#Optimisers.ClipGrad). 

Do we have good examples of callbacks being useful in the wild?

Every single use in the model zoo appears to be to print the loss, with `throttle`. So are all the ones I found on discourse in 5 minutes. 

https://github.com/FluxML/model-zoo/search?q=cb
https://github.com/FluxML/Metalhead.jl/issues/62#issue-651615425
https://discourse.julialang.org/search?q=flux%20cb

We could just build that into `train!` as an option. The goal of #2120 is to make the same thing easy in "procedural" code.

src/train.jl
```diff
@@ -56,11 +56,12 @@ end
 
 Uses a `loss` function and training `data` to improve the `model`'s parameters
 according to a particular optimisation rule `opt`. Iterates through `data` once,
-evaluating `loss(model, d...)` for each `d` in data.
+evaluating for each `d in data` either `loss(model, d...)` if `d isa Tuple`,
+or else   `loss(model, d)` for other `d`.
```
darsnack, line 60:
```suggestion
or else `loss(model, d)` for other `d`.
```

darsnack (approved):

Issue #2105, opened 2022-11-09
Documentation overhaul roadmap


darsnack:
**This issue is a WIP. I am starting from the template discussed on call, we need to break this into smaller steps.**

# Flux documentation restructure

It should surprise no one that Flux's current documentation is not very good. It might surprise people that the "whole suite" of docs does have reasonably good coverage—that speaks to the poor organization of the documentation. Despite being written down, most information is an ordeal to find.

With this in mind, we consolidated a list of steps to improve this situation. These steps have been discussed by maintainers and contributors on various Julia platforms, but this issue will now serve as a central tracking for the re-haul.

- [ ] **Restructure the Flux.jl documentation**
      The current doc model was designed at a time when Flux's ethos was to demonstrate how _simple_ neural networks in Julia can be. The current package offers a lot more than a simple list of layers and tracker-based AD. The proposed organization here is a commonly used format (notably in Python fast.ai and our own SciML). See [below](#new-documentation-structure) for a description of this format.
- [ ] **Deprecate website tutorials and model zoo in favor of documentation**
- [ ] **Detailed documentation for NNlib.jl, FluxTraining.jl, etc.**

### New documentation structure

**Home / Getting Started:** Briefly describe the package and quick installation instructions. Provide links to (a) the entry point for an absolute beginner, (b) the ecosystem page, (c) contributing and bug reporting resources.

[**Ecosystem:**](https://fluxml.ai/Flux.jl/stable/ecosystem/) This page is currently way down at the bottom, and it needs to be prominently featured. This should immediately make clear what it is the Flux provides (layers + simple utilities) and what it does not / what it borrows (datasets, data loading, optimizers, schedules, etc.). A diagram like [MLJ's](https://github.com/alan-turing-institute/MLJ.jl/blob/dev/material/MLJ_stack.svg) would be useful.

**Tutorials:** These should be complete, runnable examples that users can use to get started quickly and learn by doing. *The model zoo examples should be tutorials.* [Literate.jl](https://fredrikekre.github.io/Literate.jl/v2/documenter/) offers the ability to generate markdown for consumption by Documenter.jl. Tutorials should leverage this feature.
[...]

PR #2115, opened 2022-11-19
Move doc sections to "guide" + "reference"

mcabbott:
This is one take on how we might aim to organise the docs. No new text, just moving headings around. 

The goal is to separate it into
* A guide you could read straight through, and encounter each idea once. Basics first, more specialised later.
* A reference section, to look up all the weird loss functions / layers / etc, mostly just docstrings
* A tutorials section, into which we move what's now on the website, and maybe the model zoo. Each should solve some problem, ideally data -> model -> results. Fine if they overlap. 

This seems like a non-awful way to present things, which can be achieved without re-writing more than we have to. The sections which most awkwardly straddle guide/reference are the training/optimisation/regularisation ones, which need re-working for explicit params anyway. #2114 is a start. 

One downside of this is that some things like `onehot` and `outputsize` don't really appear in the "guide" half at all. Maybe they ought to be squeezed in somewhere? 

### Edit, more comments:

~~Maybe "Fitting a Line" and "Quickstart" also straddle the guide/tutorial boundary a bit. Could think about moving them there.~~ Maybe I like the choice of fast/slow intro. I see now that this was what the original overview PR envisaged: https://github.com/FluxML/Flux.jl/pull/1579#issue-860485363 (naming exactly xor as what the quickstart should target!)

Maybe the guide pages should be consolidated a bit -- "performance tips" doesn't say much, can it be rolled into "GPU & performance tips"? "Regularisation" also doesn't say much ... #2114  wants to make it just a section.

Not sure what order these sections ought to be in. One complaint in https://github.com/FluxML/Flux.jl/issues/2105 is "Ecosystem...  is currently way down at the bottom". This could move up, although we can't put everything first... what exactly deserves to go last? 

Do tutorials come after reference? Perhaps #2125 was merged in haste, as these tutorials aren't actually much good... maybe one or two should be updated & the rest trashed. The versions on the model zoo have stayed more up to date, maybe people should just read those -- the added text on these versions is often more confusing than helpful.

PR #2125, opened 2022-11-25
Move 5 tutorials from fluxml.github.io

mcabbott:
This moves 5 pages from https://fluxml.ai/tutorials/ to the docs at https://fluxml.ai/Flux.jl/dev/, each of which builds & trains some model. Xref https://github.com/FluxML/Flux.jl/issues/2105#issue-1442666143 about moving them.

I have made no attempt to update them, the main change is replacing the blog's title + author format with an H1 and an `!!! info` block.

I have not checked that they run. Nor does this PR make any attempt to make CI run them. 

They should certainly be updated to explicit parameters etc. (I presume that by having them in the docs, we are undertaking to keep them up to date, rather than being fixed-in-time blog posts.) But better to make the update a separate PR.


docs/make.jl
```diff
@@ -42,7 +42,14 @@ makedocs(
              "Functors.jl 📚 (`fmap`, ...)" => "models/functors.md",
          ],
          "Tutorials" => [
+             # Roughly in order of increasing complexity? Not chronological.
             "Linear Regression" => "tutorials/linear_regression.md",
+            "Julia & Flux: 60 Minute Blitz" => "tutorials/2020-09-15-deep-learning-flux.md",
```
ToucheSir, line 47: What do you think about adding a :construction: or some admonition in the tutorial itself so that users know this might be out of date? Or is the hope to have those sorted out before these hit a stable release?

mcabbott: Haven't seen one of those in a while! I don't think these are terribly out of date in fact, nothing looks like it clearly won't run.

docs/src/tutorials/2021-10-08-dcgan-mnist.md
```diff
@@ -0,0 +1,356 @@
...
+# Get the image filenames from the folder
+img_paths = readdir(folder, join=true)
+# Load all the images as an array
+images = load.(img_paths)
+# Join all the images in the array to create a matrix of images
+gif_mat = cat(images..., dims=3)
+save("./output.gif", gif_mat)
+```
+
+![](https://fluxml.ai/assets/tutorialposts/2021-10-08-dcgan-mnist/output.gif)
```
ToucheSir, line 348: We might as well move the assets over so that everything is in one place. `@ref` links can be cleaned up later.

docs/make.jl
mcabbott: Found a few places they needed to import `params`

docs/src/tutorials/2021-10-08-dcgan-mnist.md
mcabbott: I tried a bit locally, and gave up... IDK why I couldn't get relative links to work. Note that they are broken in the current website version too.

docs/make.jl
mcabbott: Re updating, some of these link to the model zoo:

https://github.com/FluxML/model-zoo/blob/master/vision/mlp_mnist/mlp_mnist.jl

https://github.com/FluxML/model-zoo/blob/master/vision/conv_mnist/conv_mnist.jl

https://github.com/FluxML/model-zoo/blob/master/vision/dcgan_mnist/dcgan_mnist.jl

IDK what the plan ought to be for keeping things in sync.

docs/src/tutorials/2021-10-08-dcgan-mnist.md
ToucheSir: Ok, no use fighting them for a quick PR. I suspect the website links were also a victim of the Franklin migration.

docs/make.jl
ToucheSir: If we make them literate files, then copying over the model zoo files and deleting them from that repository would work. In the meantime, things can stay as they are.

Saransh-cpp (approved):
Thanks, @mcabbott!!

I'll look into using Literate.jl for the tutorials.

Saransh-cpp:
Also, should the Dataloader example go in a new "How to" sections in Flux's docs?

Just saw #2115, which aims to separate out the references from the guide. Should we have a single guide section (as suggested by @mcabbott) and move all the "How to" guides under this heading?

docs/src/tutorials/2021-02-07-convnet.md
```diff
@@ -0,0 +1,260 @@
...
+ 
+   ```julia
+     anynan(x) = any(y -> any(isnan, y), x)
+   ```
+ 
+* `accuracy` computes the accuracy of our ConvNet:
+ 
+   ```julia
+     accuracy(x, y, model) = mean(onecold(cpu(model(x))) .== onecold(cpu(y)))
+   ```
```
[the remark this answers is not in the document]
mcabbott: Thanks I missed that.

mcabbott:
> should the Dataloader example go in a new "How to" sections

IDK. What else goes in there?

Lots of the reference sections do start with some text, to supplement the docstrings. The logical place to add something sounds like the MLUtils page. But probably not the exact text, since it's long and doesn't say much? 

https://fluxml.ai/Flux.jl/stable/destructure/
https://fluxml.ai/Flux.jl/stable/data/onehot/

Alternatively, a bigger "data handling" or something tutorial, which introduces how to read CSV files and image files and so on, as well as data-handling tools, might be nice to have if someone wants to write it.

PR #2127, opened 2022-11-27
Allow ForwardDiff in BatchNorm's track_stats

### LINKED ISSUE #2122: ForwardDiff + destructure is different from Zygote, on a model with BatchNorm
### Package Version

Optimisers v0.2.10, ForwardDiff v0.10.33, Flux v0.13.7

### Julia Version

1.8

### OS / Environment

OS

### Describe the bug

The example from Optimisers using ForwardDiff compared with the output from Zygote is unfortunately not the same.

### Steps to Reproduce

```Julia
using ForwardDiff  # an example of a package which only likes one array
using Flux
using Random
using Optimisers
Random.seed!(123)

model = Chain(  # much smaller model example, as ForwardDiff is a slow algorithm here
          Conv((3, 3), 3 => 5, pad=1, bias=false), 
          BatchNorm(5, relu), 
          Conv((3, 3), 5 => 3, stride=16),
        )
image = rand(Float32, 224, 224, 3, 1);
@show sum(model(image));

loss(m, x) = sum(m(x))

rule = Optimisers.Adam(0.001f0,  (0.9f0, 0.999f0), 1.1920929f-7)

flat, re = Flux.destructure(model)
st = Optimisers.setup(rule, flat)  # state is just one Leaf now

∇flat = ForwardDiff.gradient(flat) do v
    loss(re(v), image) # re(v), rebuild a new object like model
end

st, flat = Optimisers.update(st, flat, ∇flat)
@show loss(re(flat),image);
```

and Zygote version

```Julia
using Flux
using Random
Random.seed!(123)

model = Chain(  # much smaller model example, as ForwardDiff is a slow algorithm here
          Conv((3, 3), 3 => 5, pad=1, bias=false), 
          BatchNorm(5, relu), 
          Conv((3, 3), 5 => 3, stride=16),
        )
image = rand(Float32, 224, 224, 3, 1);
@show sum(model(image));

loss(m, x) = sum(m(x))

opt = Flux.Adam(0.001f0,  (0.9f0, 0.999f0), 1.1920929f-7)
θ = Flux.params(model)
grads = Flux.gradient(θ) do 
    loss(model, image)
end

Flux.update!(opt, θ, grads)
@show loss(model, image);
```

### Expected Results

```
sum(model(image)) = -0.33076355f0
loss(model, image) = -5.064876f0
```

### Observed Results

```
sum(model(image)) = -0.33076355f0
loss(re(flat), image) = -7.7023053f0
```

### Relevant log output

_No response_

mcabbott:
Fixes #2122 in the narrowest way, by explicitly removing duals. 

Does not cause ForwardDiff to put the layer into training mode. 

Has a test, which ought to survive even if some more elaborate solution is found.

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -275,8 +275,8 @@ function _track_stats!(
   μnew = vec(N ∈ reduce_dims ? μ : mean(μ, dims=N))
   σ²new = vec(N ∈ reduce_dims ? σ² : mean(σ², dims=N))
 
-  bn.μ = res_mtm .* bn.μ .+ mtm .* μnew
-  bn.σ² = res_mtm .* bn.σ² .+ mtm .* (m / (m - one(V))) .* σ²new
+  bn.μ .= value.(res_mtm .* bn.μ .+ mtm .* μnew)
+  bn.σ² .= value.(res_mtm .* bn.σ² .+ mtm .* (m / (m - one(V))) .* σ²new)
   return nothing
 end
 
--- a/test/layers/normalisation.jl
+++ b/test/layers/normalisation.jl
@@ -1,5 +1,5 @@
 using Flux, Test, Statistics
-using Zygote: pullback
+using Zygote: pullback, ForwardDiff
 
 evalwgrad(f, x...) = pullback(f, x...)[1]
 
@@ -463,3 +463,13 @@ end
   m1 = Dropout(0.5)
   @test Zygote.hessian_reverse(sum∘m1, [1.0,2.0,3.0]) == zeros(3, 3)
 end
+
+@testset "ForwardDiff" begin
+  bn = BatchNorm(3)
+  @test ForwardDiff.jacobian(bn, rand(Float32, 3, 4)) isa Matrix{Float32}
+  iszero(bn.μ)  # true, but ideally it would automatically choose trainmode
+  Flux.trainmode!(bn)
+  @test ForwardDiff.jacobian(bn, rand(Float32, 3, 4)) isa Matrix{Float32}
+  @test !iszero(bn.μ)
+end
+
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -11,6 +11,7 @@ import Optimisers: Optimisers, trainable, destructure  # before v0.13, Flux owne
 
 using Zygote, ChainRulesCore
 using Zygote: Params, @adjoint, gradient, pullback, @nograd
+using Zygote.ForwardDiff: value
 export gradient
 
 # Pirate error to catch a common mistake. (Internal function `base` because overloading `update!` is more likely to give ambiguities.)


src/layers/normalise.jl
```diff
@@ -275,8 +275,8 @@ function _track_stats!(
   μnew = vec(N ∈ reduce_dims ? μ : mean(μ, dims=N))
   σ²new = vec(N ∈ reduce_dims ? σ² : mean(σ², dims=N))
 
-  bn.μ = res_mtm .* bn.μ .+ mtm .* μnew
-  bn.σ² = res_mtm .* bn.σ² .+ mtm .* (m / (m - one(V))) .* σ²new
+  bn.μ .= value.(res_mtm .* bn.μ .+ mtm .* μnew)
```
mcabbott, line 278: Also updates `bn.μ` in-place, rather than replacing the field. We know it's mutable as the constructor made it so.

CarloLucibello: can we leave a comment explaining why `value` is there?

mcabbott: Good point, done.

mcabbott:
Bump?

test/layers/normalisation.jl
```diff
@@ -462,4 +462,18 @@ end
...
   @test Zygote.hessian_reverse(sum∘m1, [1.0,2.0,3.0]) == zeros(3, 3)
+
+  m2 = Chain(BatchNorm(3), sum)
+  @test Zygote.hessian_reverse(m2, Float32[1 2; 3 4; 5 6]) == zeros(Float32, 6, 6)
+end
+
+@testset "ForwardDiff" begin
+  bn = BatchNorm(3)
+  @test ForwardDiff.jacobian(bn, rand(Float32, 3, 4)) isa Matrix{Float32}
+  iszero(bn.μ)  # true, but ideally it would automatically choose trainmode
```
ToucheSir, line 473: If this isn't `@test`ed, could it be a part of the comment?

mcabbott: It could, but does it matter enough to be worth finding this branch again?

ToucheSir: It's just weird to see what should be a statement with zero side effects kicking around in a test. That's why I wanted to make sure I wasn't missing anything (e.g. the test is that it doesn't throw or there was a missing `@test` at the front).

mcabbott: Maybe only I do this... it's something which you can look at, but doesn't really deserve to be either test_broken nor a test.

mcabbott:
```suggestion
  # iszero(bn.μ)  # is true. But ideally would not be, if Flux would automatically choose trainmode
```

ToucheSir (approved):

PR #2133, opened 2022-12-11
Delete bors.toml

CarloLucibello:


ToucheSir:
I think we'd also need to unregister the app from this org, correct?

CarloLucibello:
ok, I uninstalled bors from the org

ToucheSir (approved):

PR #2146, opened 2022-12-29
MultiHeadAttention implementation

CarloLucibello:
~For the time being, just porting the rather basic implementation in [Metalhead.jl](https://github.com/FluxML/Metalhead.jl/blob/master/src/layers/attention.jl).~
~In the future (this PR or future ones), we will want to rely on [NeuralAttentionlib.jl](https://chengchingwen.github.io/NeuralAttentionlib.jl/dev/) for the heavy lifting.~
~Possibly, we will want to add all the flexibility of [Transformers.jl](https://chengchingwen.github.io/Transformers.jl/dev/basic/#Transformers.Basic.MultiheadAttention-NTuple{4,%20Int64}) and [pytorch](https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html#multiheadattention) implementations.~

Implementation of a (cross-)attention layer, inspired by pytorch, Transformers.jl and [flax](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.MultiHeadDotProductAttention.html?highlight=multihead#flax.linen.MultiHeadDotProductAttention)

The implementation relies on the recently added `NNlib.dot_product_attention`. 

cc @chengchingwen @theabhirath @darsnack @mcabbott 

Fix https://github.com/FluxML/Metalhead.jl/issues/141

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### Old Post
At the moment, there are two implementations: 
- one based on Tullio.jl
- one offloading the computation to NeuralAttentionlib.jl

We will have to pick one. I tested a few cases and they yield the same outputs.
NeuralAttentionlib.jl seems to be more performant, probably thanks to the handwritten rules.

I can add a "native" implementation using permutedims and batched_mul.


mcabbott:
Thanks for making a start. Some questions, perhaps not all immediate, are:

1. How complicated a combined thing should be called a layer here? This one is a much bigger block than anything else, where you stack dropout etc. by hand. 
2. Is the goal to provide the basic idea, in a readable way, or the fastest and most flexible possible implementation?
3. What deps should Flux (or perhaps NNlib) accept? We finally extricated ourselves from some of them...

Re implementations & performance, well I'm obviously pleased how well Tullio does on the CPU, compared to permutedims etc, at least at this size. It is easy to fuse the scaling of `q` into multiplication & save a copy. It might be possible to fuse more. It might be optimal to provide gradients rather than use its own. Although, see point 3.

Most of the time is spent in `softmax`. Partly because `maximum` is surprisingly slow (due to careful NaN handling?). `@turbo exp` may help a lot (but needs care as it doesn't like size-1 dimensions).

On GPU, I presume that it will be very hard to beat [cudnnMultiHeadAttnForward](https://github.com/JuliaGPU/CUDA.jl/blob/master/lib/cudnn/src/multiheadattn.jl) . But have timed nothing. (Nor have I looked closely at what cases that does & doesn't handle.)

<details>

```julia
julia> dim, len, batch_size, num_heads = 128, 8, 128, 32;
julia> mha = MultiHeadAttention(dim, num_heads);
julia> x = rand(Float32, (dim, len, batch_size));
julia> mha(x, impl=:native) ≈ mha(x, impl=:tullio)
true

# Timing the layer

julia> @btime mha(x, impl=:native, proj=false);
  min 7.193 ms, mean 8.043 ms (152 allocations, 8.26 MiB)

julia> @btime mha(x, impl=:tullio, proj=false);
  min 3.317 ms, mean 3.701 ms (161 allocations, 6.26 MiB)
  min 3.196 ms, mean 3.721 ms (167 allocations, 5.76 MiB)  with scale absorbed

julia> @btime mha(x, impl=:nalib, proj=false);
  min 7.070 ms, mean 7.926 ms (139 allocations, 6.76 MiB)

# Timing pieces

julia> @btime copy($x);  # so each copy is 1/2 MB
  min 13.083 μs, mean 48.775 μs (2 allocations, 512.06 KiB)

julia> mha.qkv_proj   # this step is always done, could be fused?
QKVProj(Dense(128 => 128; bias=false), Dense(128 => 128; bias=false), Dense(128 => 128; bias=false))

julia> @btime mha.qkv_proj(x,x,x);  # half the memory, but not much time.
  min 200.375 μs, mean 500.531 μs (26 allocations, 3.00 MiB)

julia> @btime mha.attn_drop(x);  # dropout appears to be cheap
  min 132.096 ns, mean 139.195 ns (1 allocation, 48 bytes)

julia> z = rand(Float32, 8, 8, 32, 128);  # size of input to softmax

julia> @btime softmax($z);
  min 2.754 ms, mean 2.883 ms (12 allocations, 1.25 MiB)

julia> @btime maximum($z; dims=1);  # this is a fairly big component
  min 767.625 μs, mean 801.111 μs (6 allocations, 128.25 KiB)

julia> @btime @fastmath reduce(max, $z; dims=1, init=0.0f0);
  min 78.125 μs, mean 97.439 μs (3 allocations, 128.11 KiB)

julia> @btime _softmax($z);  # an attempt...
  min 2.785 ms, mean 2.964 ms (9 allocations, 1.25 MiB)  # with reduce(max), why not better?
  min 659.542 μs, mean 825.112 μs (9 allocations, 1.25 MiB)  # with @turbo, in fact wrong

# Without softmax nor qkv_proj

julia> 7.193 - 2.754 - 0.2  # native
4.239

julia> 3.317 - 2.754 - 0.2 # tullio
0.36300000000000016

julia> 7.070 - 2.754 - 0.2  # nalib
4.1160000000000005
```

</details>

CarloLucibello:
> 1. How complicated a combined thing should be called a layer here? This one is a much bigger block than anything else, where you stack dropout etc. by hand.

The paper "Attention is all you need" and all deep learning frameworks are quite consistent in what MultiHeadAttention is, all those components are needed. I can remove the dropout on the outputs, I don't remember where I got that from, and is absent in pytorch, keras, flax.

Pytorch goes also much further and implements the whole [transformer model](https://pytorch.org/docs/stable/generated/torch.nn.Transformer.html)

> 2. Is the goal to provide the basic idea, in a readable way, or the fastest and most flexible possible implementation?

for me, it would be flexibility > readability > performance

> 3. What deps should Flux (or perhaps NNlib) accept? We finally extricated ourselves from some of them...

Being the overall performance for the layer not so different (thanks to softmax) I think we can go with `native` and avoid carrying on additional dependencies

CarloLucibello:
I can keep only the "native" implementation, and leave the tullio contractions as comments that help understand what is going on and also as pointers for implementing more performant similar operations.

I will also try to hook cudnn.

chengchingwen:
The goal of NeuralAttenionlib is to provide flexible attention variants without sacrificing the performance (though I'm mainly focus on GPU performance, no CPU).

> On GPU, I presume that it will be very hard to beat [cudnnMultiHeadAttnForward](https://github.com/JuliaGPU/CUDA.jl/blob/master/lib/cudnn/src/multiheadattn.jl) . But have timed nothing. (Nor have I looked closely at what cases that does & doesn't handle.)

If anyone is interested in beating cudnnMultiHeadAttnForward, there is a [FlashAttention](https://github.com/HazyResearch/flash-attention) that fuse almost all operation in attention into one cuda kernel, which use the WMMA api.

> How complicated a combined thing should be called a layer here? This one is a much bigger block than anything else, where you stack dropout etc. by hand.

IMPO I think Flux really need to setup a paradigm for building model. Like in Pytorch, all the predefined layer object is not needed. The user can always inheritance the `torch.nn.Module` and use `torch.functional` to define the forward method. While in Flux, we are more restrict to the predefined Layers and hence the question (what should be viewed as a layer).

mcabbott:
Re functional things, some [discussion here](https://github.com/FluxML/Fluxperimental.jl/discussions/2) but no PRs yet. 

For the attention story, I think what you're asking for is more like what we have for `conv` right? i.e. some `selfattention` function which lives in NNlib and does the work (and is overloaded to call the CUDA implementation) while the Flux layer handles initialisation etc. (We should do that to `batchnorm` too.)

For CPU implementations, note that `softmax` is the majority of the time only in the Tullio case (and I did not time gradients). https://github.com/FluxML/NNlib.jl/pull/450 makes it a little faster. Now that we have package extensions, having `using LoopVectorization` load a faster path for `softmax!(::Array, ...)` is probably something we should do. Perhaps it should similarly load a faster overload for the functional `selfattention(::Array...)`.

Thanks for the link to FlashAttention, from a very quick skim, interesting they can avoid memory quadratic in sequence length. And surprising how much time they say pytorch spends on mask/softmax/dropout.

ToucheSir:
> The user can always inheritance the `torch.nn.Module` and use `torch.functional` to define the forward method.

Where do you feel the current approach of `@functor` ≈ `torch.nn.Module` and NNlib ≈ `torch.nn.functional` falls short here? If anything I would've thought we do better than PyTorch because one doesn't have to rely on the entirety of Flux to write custom layers, but maybe there are gaps in functionality/documentation I'm missing.

chengchingwen:
I'm not sure if we really need a "factory", but I do think we doesn't separate functionality thoroughly enough. There are many stuff that should be defined in NNlib but currently live in Flux (e.g. dropout, normalization, etc.). Ideally (IMPO) every forward definition of a layer should be almost a single call of a function from NNlib, so yes, just like what we have for `conv`.

The attention story is more complicated in NAlib. It is not only designed with the functional style in mind, but also tries to be flexible enough to support different attention variants. For example:

```julia
generic_multihead_qkv_attention(
    weighted_sum_mixing,
    normalized_score(softmax) $
    masked_score(GenericMaskOp(), mask) $
    scalar_relative_position_embedding(t5_bucketed_position_id(n_bucket, max_distance), position_bias) $
    dot_product_score,
    head, q, k, v
)
```

This is an implementation of a multi-head attention with relative position embedding. By given different kinds of `mask`, it can also become other variant like a `causal` or `local` one.

ToucheSir:
> There are many stuff that should be defined in NNlib but currently live in Flux (e.g. dropout, normalization, etc.). Ideally (IMPO) every forward definition of a layer should be almost a single call of a function from NNlib, so yes, just like what we have for `conv`.

100% agreed. https://github.com/FluxML/NNlib.jl/pull/452 is a big step in that direction. I think we can set a de-facto policy that any future Flux layers should follow the criteria you've outlined as well.

chengchingwen:
> On the attention op composition example, the ideal would be that a user can mix and match bits like you've shown, but end up with something close to the performance of Flash Attention (e.g. memory efficient, fusion where applicable). It's not clear to me what that would look like, but it certainly seems within the realm of possibility.

We can always dispatch to optimized specialization of a composition, but I'm not sure if it's possible to have a composable flash attention since that might require composability at the kernel level.

> How to represent this in a layer API is just as interesting. Do we stick with a fixed "default" MHA layer as mentioned above, or can we allow some composition at that level too?

I store an [attention op inside the attention layer](https://github.com/chengchingwen/Transformers.jl/blob/0.2/src/layers/layer.jl#L575) and define [types for each composition](https://github.com/chengchingwen/Transformers.jl/blob/0.2/src/layers/attention_op.jl#L181).

ToucheSir:
I don't think it's impossible, but it would be tricky. Separate host-side functions should always be the fallback, but I feel we could eventually come up with something like Base's broadcast types which allows for kernel composition (probably would involve some generated functions).

chengchingwen:
Another issue is the gradient computation. Even the gradient of the broadcast fusion is hard to compute, I don't think it would easier for attention fusion. I was actually hoping Enzyme could help with this, but we need to have the fused kernel beforehand.

ToucheSir:
I'd be pleasantly surprised if Enzyme could synthesize a fast gradient for such complex kernels. Definitely been playing around with it for fused broadcasting however. Speculatively, a halfway solution could be to define something like https://github.com/JuliaGPU/GemmKernels.jl where we create gradient building blocks and have the library combine them together. Anyhow, none of that ought to hold up this PR, just a nice brainstorm :)

chengchingwen:
> Anyhow, none of that ought to hold up this PR, just a nice brainstorm :)

Sure, I have no opinion on this PR since Transformers.jl would be using its own implementation.

CarloLucibello:
any feedback? This should be essentially ready

test/test_utils.jl
```diff
@@ -82,3 +82,30 @@ function gpu_autodiff_test(
...
+function test_grad_equal(g1::AnyCuArray{T}, g2::Array{T}; atol=1e-4) where T 
+    @test Array(g1) ≈ g2 atol=atol
+end
+
+function test_grad_equal(g1::T1, g2::T2) where {T1 <: NamedTuple, T2 <: NamedTuple}
+    @test fieldnames(T1) == fieldnames(T2)
+    for f in fieldnames(T1)
+        test_grad_equal(g1[f], g2[f])
+    end
+end
```
ToucheSir, line 111: This looks to be already covered by the existing utils such as `check_grad`. The tests above should use those instead.

src/layers/attention.jl
```diff
@@ -0,0 +1,121 @@
...
+  ## [k_in] = [k_in_dim, kv_len, batch_size] 
+  ## [v_in] = [v_in_dim, kv_len, batch_size]
+  q = mha.q_proj(q_in)  # [q] = [qk_dim, q_len, batch_size]
+  k = mha.k_proj(k_in)  # [k] = [qk_dim, kv_len, batch_size] 
+  v = mha.v_proj(v_in)  # [v] = [v_dim, kv_len, batch_size]
+  x, α = NNlib.dot_product_attention(q, k, v, bias; mha.nheads, mask, fdrop=mha.attn_drop)
+  x = mha.out_proj(x)
+  # [x] = [out_dim, q_len, batch_size]
+  # [α] = [kv_len, q_len, nheads, batch_size]
+  return withscores ? (x, α) : x
```
CarloLucibello, line 120: For the sake of type stability, maybe we should always return both `(x, α)`. 
In most applications the scores are not needed, but sometimes they are.
And this is not the kind of layer that you put in a `Chain` in any case, 
it will be part of blocks, so having extra outputs is not problematic

CarloLucibello: always returning  (x, α) now

ToucheSir (approved):
Doesn't look like there's much left to say here, shall we do this?

CarloLucibello:
letsdoit!

PR #2150, opened 2023-01-03
Move `dropout` to NNlib

mcabbott:
Needs https://github.com/FluxML/NNlib.jl/pull/454

```julia
julia> model = Chain(Dense(10=>100), Dropout(0.5)); data = rand(Float32, 10, 100);

julia> @btime $model($data);
  min 4.536 μs, mean 11.995 μs (4 allocations, 78.22 KiB)
  min 4.692 μs, mean 14.468 μs (4 allocations, 78.22 KiB)

julia> @btime gradient(m -> sum(m($data)), $model);
  min 45.750 μs, mean 77.092 μs (176 allocations, 251.69 KiB)  # before
  min 32.125 μs, mean 55.435 μs (110 allocations, 209.89 KiB)  # after

julia> trainmode!(model);

julia> @btime $model($data);
  min 14.500 μs, mean 25.182 μs (17 allocations, 156.67 KiB)
  min 10.375 μs, mean 24.783 μs (6 allocations, 117.33 KiB)   # after
```


src/layers/normalise.jl
```diff
@@ -1,114 +1,86 @@
...
 end
-Dropout(p, dims, active) = Dropout(p, dims, active, default_rng_value())
+Dropout(p::Real, dims, active) = Dropout(p, dims, active, default_rng_value())
 
-function Dropout(p; dims=:, rng = default_rng_value())
-  @assert 0 ≤ p ≤ 1
+function Dropout(p::Real; dims=:, rng = default_rng_value())
+  0 ≤ p ≤ 1 || throw(ArgumentError("Dropout expexts 0 ≤ p ≤ 1"))
   Dropout(p, dims, nothing, rng)
 end
```
mcabbott, line 73: Since `Dropout` is mutable, you can conceivably do `d = Dropout(0.0); d.p = 0.1`. Is this supported?

If not, then the constructor could do this:
```
p == 0 && return identity
p == 1 && return zero
```
That would mean that some model construction function which happens to have `p==0` will do the right thing, without imposing a branch etc.

ToucheSir: I don't know how to feel about this. @darsnack might be able to give a more definite answer because we ran into this question a couple times during the Metalhead rework.

src/deprecations.jl
```diff
@@ -174,6 +174,18 @@ function update!(opt::Optimise.AbstractOptimiser, ::Params, grads::Union{Tuple,
     """)
 end
 
+
+function dropout(rng, x, p; dims=:, active::Bool=true)
+  if active
+    NNlib.dropout(rng, x, p; dims)
+  else
+    Base.depwarn("Flux.dropout(...; active=false) is deprecated. Please branch outside the function, or call dropout(x, 0) if you must.", :dropout)
```
mcabbott, line 182: I made this to preserve the `active` keyword. Not entirely sure whether use of that was supported outside Flux. 

The exported function is this one, not the NNlib one, and I think it lacks a docstring at present.

ToucheSir: I added it to https://github.com/FluxML/NNlib.jl/pull/452 because backends need it, but to my knowledge none do for dropout so this seems fine.

src/layers/normalise.jl
mcabbott: [6c84a6c](https://github.com/FluxML/Flux.jl/pull/2150/commits/6c84a6cf2c6b737be39f8beba41acec6dd8e220c) adds this for `Dropout(0)`, which you cannot mutate, but not `Dropout(0.0)`. Not very sure that's desirable.

src/deprecations.jl
mcabbott: You think it might be OK to just go without the deprecation?

ToucheSir: Not sure. A quick JuliaHub search didn't turn up anything, but I've enabled downstream tests in case we want to check.

src/layers/normalise.jl
```diff
@@ -1,111 +1,85 @@
...
-function dropout(rng, x, p; dims=:, active::Bool=true)
-  active || return x
-  y = dropout_mask(rng, x, p, dims=dims)
-  return x .* y
-end
-dropout(x, p; kwargs...) = dropout(rng_from_array(x), x, p; kwargs...)
-
-dropout_mask(rng::CUDA.RNG, x::CuArray, p; kwargs...) = _dropout_mask(rng, x, p; kwargs...)
-dropout_mask(rng, x::CuArray, p; kwargs...) =
-  throw(ArgumentError("x isa CuArray, but rng isa $(typeof(rng)). dropout_mask only support CUDA.RNG for CuArrays."))
```
mcabbott, line 38: How much do we care about preserving this error?

With the present state of NNlib, the RNG is ignored:
```julia
julia> using Random; Random.seed!(0);

julia> dropout(Random.default_rng(), CUDA.rand(2,7), 0.3)
2×7 CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}:
 0.0823338  0.0  0.0     0.0  0.289135  0.886469  1.30285
 0.152207   0.0  1.0032  0.0  0.23335   0.0       0.0

julia> using Random; Random.seed!(0);

julia> dropout(Random.default_rng(), CUDA.rand(2,7), 0.3)
2×7 CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}:
 1.2269    0.0  0.0     0.0  1.30776  0.874458  0.17939
 0.760811  0.0  1.3571  0.0  1.37249  0.0       0.0
```
vs Array case:
```
julia> using Random; Random.seed!(0);

julia> dropout(rand(2,7), 0.3)
2×7 Matrix{Float64}:
 0.650341  0.0      0.0423538  0.0  0.12421  0.0       1.28538
 0.782346  1.34369  1.06706    0.0  0.0      0.167127  0.0

julia> using Random; Random.seed!(0);

julia> dropout(rand(2,7), 0.3)
2×7 Matrix{Float64}:
 0.650341  0.0      0.0423538  0.0  0.12421  0.0       1.28538
 0.782346  1.34369  1.06706    0.0  0.0      0.167127  0.0
```

mcabbott: NNlib + NNlibCUDA aren't set up to make inserting such a check trivial, without duplicating code. 

It's possible that this ought to be viewed as a CUDA bug, that it should error instead of running `rand!(rng, CuArray)` ignoring `rng`? Maybe too disruptive.

Maybe CUDA should do something other than ignore --- for instance it could hash the components of Xoshiro and use that as the seed; this would not update the original RNG though...
```
julia> Random.default_rng() |> copy |> dump
Xoshiro
  s0: UInt64 0x3278a8849972e116
  s1: UInt64 0x10d9ea4d886b539b
  s2: UInt64 0x5f94d0fd98860389
  s3: UInt64 0x0e550adea2e2c23f

julia> CUDA.default_rng()|> dump
CUDA.RNG
  seed: UInt32 0x29d0efbc
  counter: UInt32 0x0000002c
```

mcabbott: Done.

mcabbott:
Downstream:
* AtomicGraphNets unrelated
* Metalhead ` loadmodel!(dst::Tuple{Chain{` "but the structures do not match.", unclear where the mismatch is & whether it's related. (Can that have a better error message?)
* GeometricFlux fails as its `∇indexed_softmax` says "UndefVarError: within_grad not defined" from [here](https://github.com/FluxML/GeometricFlux.jl/blob/c306989f3aeb25d41e0fe00858880caf56d1e3f5/src/operation.jl#L46). Maybe https://github.com/FluxML/NNlib.jl/pull/434 should have deprecated that or maybe it was obviously internal.

ToucheSir:
Metalhead load failure is known and can be ignored. Adding a better error message is tricky because `loadmodel!` is general-purpose, but that's not my area of expertise.

For GeometricFlux, I missed that the NNlib PR renamed the function. A courtesy deprecation seems fine here.

src/deprecations.jl
mcabbott: Now 0e396a6 removes this entirely. At least to see if anything changes.

src/utils.jl
```diff
@@ -36,32 +36,12 @@ epseltype(x) = eps(float(eltype(x)))
...
-rng_from_array(::AbstractArray) = default_rng_value()
-rng_from_array(::CuArray) = CUDA.default_rng()
-
-@non_differentiable rng_from_array(::Any)
-
-if VERSION >= v"1.7"
-  default_rng_value() = Random.default_rng()
-else
-  default_rng_value() = Random.GLOBAL_RNG
-end
```
mcabbott, line 55: I don't understand why this exists, instead of `Random.default_rng()`. It was added in https://github.com/FluxML/Flux.jl/pull/1849  and re-named in  https://github.com/FluxML/Flux.jl/pull/1998 . 

This function sounds a bit like it makes a special token to say "default, no option provided" and we could allow only that token to be moved to GPU's default. But no, we accept the output of `Random.default_rng()` here, and in that case disobey your choice because it happened to match the default:

https://github.com/FluxML/Flux.jl/blob/ee78ce3cefb027228413f8edace3c0385139d786/src/functor.jl#L100-L107

Which isn't a terrible policy, but then there's no need for this function. Random provides this for us.

ToucheSir: I don't remember either, hence https://github.com/FluxML/NNlib.jl/pull/454#discussion_r1061042380.

mcabbott: Ah right, sorry. 

Anyway, maybe not this PR's problem. We can see about something cleaner elsewhere.

src/layers/normalise.jl
```diff
@@ -1,111 +1,85 @@
...
-`dims` keyword. e.g. `Dropout(p; dims = 3)` will randomly zero out entire channels on WHCN input
-(also called 2D dropout). This is used as a regularisation, i.e. it reduces overfitting during 
-training.
+While training, it sets each input to `0` (with probability `p`)
+or else scales it by `1 / (1 - p)`, using the [`NNlib.dropout`](@ref) function.
+While testing, it has no effect.
 
-In the forward pass, this layer applies the [`Flux.dropout`](@ref) function. See that for more
-details.
+By defaul the mode will switch automatically, but it can also
```
CarloLucibello, line 14:
```suggestion
By default the mode will switch automatically, but it can also
```

src/utils.jl
darsnack: Sorry for the absence on my part, but the issue was that pre-1.6, `Base.default_rng()` exists but returns a new `MersenneTwister` instead of `GLOBAL_RNG`. The latter is what is actually used as the default RNG when a user doesn't specify one. So this existed to be consistent with what users expect.

darsnack: I didn't look into the sequence generated, so I didn't spot the practical (non)-difference that you are highlighting for the single-threaded case.

But yes, based on my reading from when this code was inserted into Flux, pre-1.7, all random generation functions used `GLOBAL_RNG` which was shared across threads. Calling `default_rng()` within a thread would create a `MersenneTwister` with the thread id used as the seed.

On Julia 1.6:
```
julia> using Random

julia> function rng_default()
       rng = Random.default_rng()
       @show Threads.threadid(), rng, rand(rng)
       end
rng_default (generic function with 1 method)

julia> function rng_global()
       rng = Random.GLOBAL_RNG
       @show Threads.threadid(), rng, rand(rng)
       end
rng_global (generic function with 1 method)

julia> Random.seed!(42)
MersenneTwister(42)

julia> Threads.@threads for i = 1:3
       rng_default()
       end
(Threads.threadid(), rng, rand(rng)) = (1, MersenneTwister(42, (0, 1002, 0, 1)), 0.5331830160438613)
(Threads.threadid(), rng, rand(rng)) = (3, MersenneTwister(0x58441c6282d70c2ce98e81ef10db8713, (0, 1002, 0, 1)), 0.2614004171338713)
(Threads.threadid(), rng, rand(rng)) = (2, MersenneTwister(0x782312cddc8165942860ed1e998ede22, (0, 1002, 0, 1)), 0.8547272913751691)

julia> Random.seed!(42)
MersenneTwister(42)

julia> Threads.@threads for i = 1:3
       rng_global()
       end
(Threads.threadid(), rng, rand(rng)) = (1, Random._GLOBAL_RNG(), 0.5331830160438613)
(Threads.threadid(), rng, rand(rng)) = (2, Random._GLOBAL_RNG(), 0.6812787857760749)
(Threads.threadid(), rng, rand(rng)) = (3, Random._GLOBAL_RNG(), 0.3439025677665648)
```

darsnack: On Julia 1.8 which ditched `GLOBAL_RNG`:
```
julia> Random.seed!(42)
TaskLocalRNG()

julia> Threads.@threads for i = 1:3
       rng_default()
       end
(Threads.threadid(), rng, rand(rng)) = (2, TaskLocalRNG(), 0.07326257546968751)
(Threads.threadid(), rng, rand(rng)) = (1, TaskLocalRNG(), 0.980853662971136)
(Threads.threadid(), rng, rand(rng)) = (3, TaskLocalRNG(), 0.5850242776598632)

julia> Random.seed!(42)
TaskLocalRNG()

julia> Threads.@threads for i = 1:3
       rng_default()
       end
(Threads.threadid(), rng, rand(rng)) = (3, TaskLocalRNG(), 0.5850242776598632)
(Threads.threadid(), rng, rand(rng)) = (1, TaskLocalRNG(), 0.980853662971136)
(Threads.threadid(), rng, rand(rng)) = (2, TaskLocalRNG(), 0.07326257546968751)
```

darsnack: And just to clarify what I mean by "users expect": in pre-1.7, the "default" RNG was always `GLOBAL_RNG` in Base and in the ecosystem. `Base.default_rng()` exists and we can use it, but then the RNG being used will not always like up with my mental model: that the RNG used should be the same as `rand()`. I guess I didn't realize that in the single-threaded case, this difference is only in the printed type and not a practical one.

src/layers/normalise.jl
darsnack: We discussed this during the ML community call. The decision is to keep things simple at the level of Flux/NNlib, i.e. we should return `Dropout` and not optimize it away. These kinds of optimization could be part of a tool that cuts and stitches the model back together with optimized parts (e.g. in Fluxperimental.jl).

darsnack:
We discussed this on call and resolved the outstanding default RNG issue (i.e. leave unchanged). I commented on other issues with the decision on call. Other than those discussions, this PR appears ready.

mcabbott:
Ok. In its present state this PR doesn't change the RNG handling.

Looking briefly again, I wonder if we should remove the branch `if _isactive(a, x) && a.p != 0` here (which is visible to Zygote) in favour of `dropout(a.rng, x, a.p * _isactive(a, x); dims=a.dims)`. Haven't timed it but https://github.com/FluxML/NNlib.jl/issues/462 made me more worried about branches.

mcabbott:
One more thought. Should we make test mode visible? It's a bit weird that the model has invisible state. Perhaps we should add a keyword to the constructor, and print this:
```
julia> Flux.trainmode!(m)
Chain(  
  Dense(2 => 3),                        # 9 parameters
  Dropout(0.4, active=true),
)
```
With more effort, we could make `show` check `_isactive` and write say `# train mode, 9 parameters` but not sure that's better.

darsnack:
> Perhaps we should add a keyword to the constructor

I agree, and a keyword seems the best option to me.

ToucheSir:
We could make the `active` keyword behave like `{train,test}mode!`, where users can pass `:auto` or `nothing`. That would also help with display.

mcabbott:
Yes. Is `active` the right name? This is only used internally at present. Maybe it should follow `{train,test}mode!` and be called `train=true` for the active state?

ToucheSir:
I recall doing a quick survey of this at some point, but not where I posted the results. PyTorch uses `training` as [inherited](https://pytorch.org/docs/stable/generated/torch.nn.Module.html?highlight=module#torch.nn.Module) from `torch.nn.Module`. TF relies on a [`training` kwarg](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dropout) at _call time_ which the framework automagically sets. Flax has a [`deterministic` field](https://flax.readthedocs.io/en/latest/guides/dropout.html?highlight=deterministic) which it expects users to set themselves and provides no automatic functionality for.

mcabbott:
Thanks for digging. Maybe my vote is to leave this for now, and do it later to BatchNorm too.

src/layers/normalise.jl
```diff
@@ -1,112 +1,80 @@
...
-Dropout(p, dims, active) = Dropout(p, dims, active, default_rng_value())
+Dropout(p::Real, dims, active) = Dropout(p, dims, active, default_rng_value())
 
-function Dropout(p; dims=:, rng = default_rng_value())
-  @assert 0 ≤ p ≤ 1
+function Dropout(p::Real; dims=:, rng = default_rng_value())
+  0 ≤ p ≤ 1 || throw(ArgumentError("Dropout expexts 0 ≤ p ≤ 1, got p = $p"))
+  if p isa Integer  # Dropout(0)
+    return p==0 ? identity : zero
+  end
```
darsnack, line 70: I think on call we decide against this even for the `Integer` case

src/layers/normalise.jl
```diff
@@ -1,112 +1,80 @@
...
   active::Union{Bool, Nothing}
   rng::R
 end
-Dropout(p, dims, active) = Dropout(p, dims, active, default_rng_value())
+Dropout(p::Real, dims, active) = Dropout(p, dims, active, default_rng_value())
 
-function Dropout(p; dims=:, rng = default_rng_value())
-  @assert 0 ≤ p ≤ 1
+function Dropout(p::Real; dims=:, rng = default_rng_value())
+  0 ≤ p ≤ 1 || throw(ArgumentError("Dropout expexts 0 ≤ p ≤ 1, got p = $p"))
```
darsnack, line 67:
```suggestion
  0 ≤ p ≤ 1 || throw(ArgumentError("Dropout expects 0 ≤ p ≤ 1, got p = $p"))
```

src/layers/normalise.jl
```diff
@@ -1,112 +1,80 @@
...
 """
-mutable struct Dropout{F,D,R<:AbstractRNG}
+mutable struct Dropout{F<:Real,D,R<:AbstractRNG}
   p::F
   dims::D
   active::Union{Bool, Nothing}
   rng::R
 end
-Dropout(p, dims, active) = Dropout(p, dims, active, default_rng_value())
+Dropout(p::Real, dims, active) = Dropout(p, dims, active, default_rng_value())
```
darsnack, line 64: Not sure if this is intentional but the error checking seems to only apply to the keyword based constructor.

src/layers/normalise.jl
mcabbott: Ah ok:
```suggestion
```

src/layers/normalise.jl
```diff
@@ -1,112 +1,80 @@
...
-  y = rand!(rng, similar(x, realfptype, _dropout_shape(x, dims)))
-  y .= _dropout_kernel.(y, p, 1 - p)
-  return y
-end
-
-# TODO move this to NNlib
-ChainRulesCore.@non_differentiable dropout_mask(::Any, ::Any, ::Any)
-
 """
     Dropout(p; dims=:, rng = default_rng_value())
```
mcabbott, line 5: This can also be shortened -- how the lack of `dims` is implemented isn't important:
```suggestion
    Dropout(p; [dims, rng])
```

src/layers/normalise.jl
```diff
@@ -1,112 +1,80 @@
...
 
-In the forward pass, this layer applies the [`Flux.dropout`](@ref) function. See that for more
-details.
+By default the mode will switch automatically, but it can also
+be controlled manually via [`Flux.testmode!`](@ref).
 
-Specify `rng` to use a custom RNG instead of the default.
-Custom RNGs are only supported on the CPU.
+By default every input is treated independently. The `dims` keyword
+instead takes a random choice only along that dimension.
```
mcabbott, line 18:
```suggestion
By default every input is treated independently. With the `dims` keyword,
instead it takes a random choice only along that dimension.
```

src/layers/normalise.jl
mcabbott: I think that's the only "public" one. I have no idea why we have this 3-arg constructor `Dropout(p, dims, active)`, e.g. Functors will use the 4-arg one. Maybe it was in case someone was relying on it from before the `rng` field was added?

darsnack: Yeah that's my recollection.

mcabbott:
Maybe this is ready?

ToucheSir (approved):

PR #2152, opened 2023-01-06
Use NNlib's `within_gradient`

### LINKED ISSUE #2122: ForwardDiff + destructure is different from Zygote, on a model with BatchNorm

mcabbott:
Uses https://github.com/FluxML/NNlib.jl/pull/434

Planned to fix issue  #2122, but doesn't! This is because ForwardDiff's chunk mode runs the model several times, and each of them updates the history within BatchNorm. Sounds pretty tricky to solve.

But with https://github.com/FluxML/Tracker.jl/pull/136, BatchNorm should work with Tracker. Ideally we'd have multi-AD tests... not here though.


src/deprecations.jl
```diff
@@ -86,6 +86,17 @@ Base.@deprecate_binding Data Flux false "Sub-module Flux.Data has been removed.
 
 @deprecate rng_from_array() default_rng_value()
 
+function istraining()
+  Base.depwarn("Flux.istraining() is deprecated, use NNlib.within_gradient(x) instead", :istraining)
+  false
+end
+ChainRulesCore.rrule(::typeof(istraining)) = true, _ -> (NoTangent(),)
+
+function _isactive(m)
```
CarloLucibello, line 95: Maybe we can avoid deprecating this by passing to `within_gradient` one of the trainables of `m`?

mcabbott: This is possible, but not sure we need do anything. It's an internal function... not sure if there are any users outside Flux?

CarloLucibello: yeah I guess doesn't matter

CarloLucibello (approved):

PR #2156, opened 2023-01-08
Match layer output to weights

### LINKED ISSUE #1565: Recurrent cell `eltype` restriction breaks `outputsize`

### LINKED ISSUE #1972: Opaque error caused by Float64 input to RNN
The simplest example in the Flux documentation is
```
using Flux
rnn = RNN(3,5)
rnn(rand(Float32, 3))
```
However, if you write
`rnn(rand(3))`
you get the opaque error message
```
ERROR: MethodError: no method matching (::Flux.RNNCell{typeof(tanh), Matrix{Float32}, Vector{Float32}, Matrix{Float32}})(::Matrix{Float32}, ::Vector{Float64})
Closest candidates are:
  (::Flux.RNNCell{F, A, V, var"#s371"} where var"#s371"<:AbstractMatrix{T})(::Any, ::Union{AbstractVector{T}, AbstractMatrix{T}, Flux.OneHotArray}) where {F, A, V, T} at /home/xxx/.julia/packages/Flux/BPPNj/src/layers/recurrent.jl:103
Stacktrace:
 [1] (::Flux.Recur{Flux.RNNCell{typeof(tanh), Matrix{Float32}, Vector{Float32}, Matrix{Float32}}, Matrix{Float32}})(x::Vector{Float64})
   @ Flux ~/.julia/packages/Flux/BPPNj/src/layers/recurrent.jl:47
 [2] top-level scope
   @ REPL[4]:1
 [3] top-level scope
   @ ~/.julia/packages/CUDA/nYggH/src/initialization.jl:52
```
This occurs on both Julia 1.6 and 1.7. I realize Float32 is used in Flux for speed, but if RNN can only support Float32 I think the error message should be clearer especially considering expressions like `rand`, `ones`, `zeros`, and `[1.0]` all default to Float64.

mcabbott:
The performance tips page spends ages talking about how not to accidentally promote to Float64. Despite which, I think some of the tutorials still do so accidentally, and are many times slower than they need to be. 

Perhaps we should just automatically fix the problem, and convert to the element type of the weights? Are there any real uses where you'd want to rely on `*` promoting the weights (or the input) for you, rather than explicitly setting the types?

Fixes #1972

Does not break this example: https://github.com/FluxML/Flux.jl/pull/1755#issuecomment-961678653

<details>

```julia
julia> Flux.train!(loss_gpu, θ_gpu, [(X_gpu, Y_gpu)], opt_gpu)

julia> loss_cpu(X, Y)
-10.739283f0

julia> loss_gpu(X_gpu, Y_gpu)
-10.739282f0

julia> rnn_gpu.
cell   state
julia> rnn_gpu.state
3×15 CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}:
 -0.896594   -0.879953  -0.546474  -0.513224  …  -0.92431   -0.00636907  -0.205964  -0.584286
  0.887761    0.803234   0.543203   0.416374      0.810287  -0.0428312    0.295843  -0.0679076
 -0.0956036  -0.38699    0.485399  -0.874913      0.775657   0.655159    -0.781047  -0.871242

julia> rnn.state
3×15 Matrix{Float32}:
 -0.896594   -0.879953  -0.546474  -0.513224  …  -0.92431   -0.0063691  -0.205965  -0.584287
  0.887761    0.803234   0.543203   0.416374      0.810287  -0.0428311   0.295843  -0.0679076
 -0.0956035  -0.38699    0.485399  -0.874913      0.775657   0.655159   -0.781047  -0.871242
```

</details>

Also fixes #1565, by an explicit method for Recur.

Originally part of #2137

[Tutorial](https://github.com/FluxML/Flux.jl/blob/master/docs/src/tutorials/2021-10-14-vanilla-gan.md#training):

<details>

With minor fixes to run on tagged Flux, time the first epoch (on CPU):
```julia
julia> @time for x in train_loader
           # - Flatten the images from 28x28xbatchsize to 784xbatchsize
           real_data = Flux.flatten(x);

           # Train the discriminator
           noise = randn(latent_dim, size(x)[end]) |> gpu
           fake_data = generator(noise)
           loss_dscr = train_dscr!(discriminator, real_data, fake_data)
           # loss_sum_dscr += loss_dscr

           # Train the generator
           loss_gen = train_gen!(discriminator, generator)
           # loss_sum_gen += loss_gen
       end
749.946580 seconds (949.68 k allocations: 98.895 GiB, 94.96% gc time, 0.21% compilation time)
```

With this PR:
```julia
┌ Warning: Layer with Float32 parameters got Float64 input.
│   The input will be converted, but any earlier layers may be very slow.
│   layer = Dense(100 => 256, #13)  # 25_856 parameters
│   summary(x) = "100×128 Matrix{Float64}"
└ @ Flux ~/.julia/dev/Flux/src/layers/stateless.jl:77
269.454910 seconds (908.34 k allocations: 30.413 GiB, 92.92% gc time, 0.60% compilation time)
```

</details>

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -200,10 +200,11 @@ end
 RNNCell((in, out)::Pair, σ=tanh; init=Flux.glorot_uniform, initb=zeros32, init_state=zeros32) =
   RNNCell(σ, init(out, in), init(out, out), initb(out), init_state(out,1))
 
-function (m::RNNCell{F,I,H,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {F,I,H,V,T}
+function (m::RNNCell{F,I,H,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{<:AbstractFloat},OneHotArray}) where {F,I,H,V,T}
   Wi, Wh, b = m.Wi, m.Wh, m.b
   σ = NNlib.fast_act(m.σ, x)
-  h = σ.(Wi*x .+ Wh*h .+ b)
+  xT = _match_eltype(m, T, x)
+  h = σ.(Wi*xT .+ Wh*h .+ b)
   return h, reshape_cell_output(h, x)
 end
 
@@ -305,9 +306,10 @@ function LSTMCell((in, out)::Pair;
   return cell
 end
 
-function (m::LSTMCell{I,H,V,<:NTuple{2,AbstractMatrix{T}}})((h, c), x::Union{AbstractVecOrMat{T},OneHotArray}) where {I,H,V,T}
+function (m::LSTMCell{I,H,V,<:NTuple{2,AbstractMatrix{T}}})((h, c), x::Union{AbstractVecOrMat{<:AbstractFloat},OneHotArray}) where {I,H,V,T}
   b, o = m.b, size(h, 1)
-  g = muladd(m.Wi, x, muladd(m.Wh, h, b))
+  xT = _match_eltype(m, T, x)
+  g = muladd(m.Wi, xT, muladd(m.Wh, h, b))
   input, forget, cell, output = multigate(g, o, Val(4))
   c′ = @. sigmoid_fast(forget) * c + sigmoid_fast(input) * tanh_fast(cell)
   h′ = @. sigmoid_fast(output) * tanh_fast(c′)
@@ -376,9 +378,10 @@ end
 GRUCell((in, out)::Pair; init = glorot_uniform, initb = zeros32, init_state = zeros32) =
   GRUCell(init(out * 3, in), init(out * 3, out), initb(out * 3), init_state(out,1))
 
-function (m::GRUCell{I,H,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {I,H,V,T}
+function (m::GRUCell{I,H,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{<:AbstractFloat},OneHotArray}) where {I,H,V,T}
   Wi, Wh, b, o = m.Wi, m.Wh, m.b, size(h, 1)
-  gxs, ghs, bs = multigate(Wi*x, o, Val(3)), multigate(Wh*h, o, Val(3)), multigate(b, o, Val(3))
+  xT = _match_eltype(m, T, x)
+  gxs, ghs, bs = multigate(Wi*xT, o, Val(3)), multigate(Wh*h, o, Val(3)), multigate(b, o, Val(3))
   r, z = _gru_output(gxs, ghs, bs)
   h̃ = @. tanh_fast(gxs[3] + r * ghs[3] + bs[3])
   h′ = @. (1 - z) * h̃ + z * h
@@ -444,9 +447,10 @@ GRUv3Cell((in, out)::Pair; init = glorot_uniform, initb = zeros32, init_state = 
   GRUv3Cell(init(out * 3, in), init(out * 2, out), initb(out * 3),
             init(out, out), init_state(out,1))
 
-function (m::GRUv3Cell{I,H,V,HH,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {I,H,V,HH,T}
+function (m::GRUv3Cell{I,H,V,HH,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{<:AbstractFloat},OneHotArray}) where {I,H,V,HH,T}
   Wi, Wh, b, Wh_h̃, o = m.Wi, m.Wh, m.b, m.Wh_h̃, size(h, 1)
-  gxs, ghs, bs = multigate(Wi*x, o, Val(3)), multigate(Wh*h, o, Val(2)), multigate(b, o, Val(3))
+  xT = _match_eltype(m, T, x)
+  gxs, ghs, bs = multigate(Wi*xT, o, Val(3)), multigate(Wh*h, o, Val(2)), multigate(b, o, Val(3))
   r, z = _gru_output(gxs, ghs, bs)
   h̃ = tanh_fast.(gxs[3] .+ (Wh_h̃ * (r .* h)) .+ bs[3])
   h′ = @. (1 - z) * h̃ + z * h
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -57,3 +57,47 @@ true
   σ = std(x, dims=dims, mean=μ, corrected=false)
   return @. (x - μ) / (σ + ϵ)
 end
+
+"""
+    _match_eltype(layer, ::Type{T}, x)
+    _match_eltype(layer, x)
+
+This internal function corrects most layer input to match the type of the weights.
+The second method uses `T = eltype(layer.weight)`.
+
+It solves a common performance bug: Before, accidentally supplying `Float64` input,
+or an activation function which produces `Float64`, would silently run the
+entire forward pass in this precision.
+"""
+_match_eltype(layer, ::Type{T}, x::AbstractArray{T}) where {T} = x
+
+# A common mistake, print a friendly warning, and fix it:
+function _match_eltype(layer, ::Type{Float32}, x::AbstractArray{Float64})
+  # This warning is the only reason this needs to take the layer.
+  @warn "Layer with Float32 parameters got Float64 input.
+  The input will be converted, but any earlier layers may be very slow." layer summary(x) maxlog=1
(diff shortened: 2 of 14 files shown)


ToucheSir:
I've been putting off reading about how other DL libraries handle promotion and/or mixed precision because it's hardly a riveting read, but does anyone have observations to offer there? Thus far, it seems like [PyTorch's promotion rules](https://pytorch.org/docs/stable/tensor_attributes.html#type-promotion-doc) roughly match ours and AMP training would be mostly orthogonal to this.

mcabbott:
Many thanks for digging. 

It sounds like Pytorch is conflicted between wanting roughly Flux/Julia's present rules (always widen), and instead wanting errors on mismatch (assumed to be a mistake), but at present has a mixture of the two. 

This PR does a 3rd thing, fixing mistakes by assuming the weights to be definitive. With a warning, which I hope may be about as good as an error for debugging, and less unfriendly. It would be easy to make a switch like `CUDA.allow_scalar` to let you choose an error.

Re mixed precision, this was my understanding of the basic idea: A copy of the model is run entirely in Float16, weights, data & gradients. (Possibly with the loss function doing some calculation in higher precision.) 

Using Float16 gradients to update Float32 weights should now be trivial with Optimisers.jl, which is in fact a nice advantage compared to Flux's old implicit way, which I think would be confused by the two copies of each weight array have different `objectid`s.

mcabbott:
Doctests pass after using #2157 some places.

src/layers/recurrent.jl
```diff
@@ -200,10 +200,11 @@ end
 RNNCell((in, out)::Pair, σ=tanh; init=Flux.glorot_uniform, initb=zeros32, init_state=zeros32) =
   RNNCell(σ, init(out, in), init(out, out), initb(out), init_state(out,1))
 
-function (m::RNNCell{F,I,H,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{T},OneHotArray}) where {F,I,H,V,T}
+function (m::RNNCell{F,I,H,V,<:AbstractMatrix{T}})(h, x::Union{AbstractVecOrMat{<:AbstractFloat},OneHotArray}) where {F,I,H,V,T}
```
ToucheSir, line 203: Is there any risk of us having to handle `Dual`s or such here?

ToucheSir (commented):
I'm on the fence about the complexity-performance tradeoff this brings, but at some point the only way we'll know is to try it. Just one comment on the RNN changes.

src/layers/recurrent.jl
mcabbott: Since I forgot, the intention here is that an `RNNCell{F,I,H,V,<:AbstractMatrix{Float32}}` should accept also accept `Matrix{Float64}` etc, which we are sure that `_match_eltype` will convert to Float32, and hence not promote the type of the hidden state. It does not accept `Matrix{Nil}`, nor dual numbers.

For the old code to accept `Dual`, the weight matrix would have to also be duals with the same tag parameter. It's hard to imagine how that would happen. So I don't think that this will break anything.

mcabbott:
Re complexity, it does seem a little sad to move from this:
```julia
(a::Dense)(x::AbstractVecOrMat) = (a.σ).(a.weight * x .+ a.bias)
```
to this opaque thing (after other PRs too):
```julia
  _size_check(a, x, 1 => size(a.weight, 2))
  xT = _match_eltype(a, x)
  NNlib.bias_act!(a.σ., a.weight * xT, a.bias)
end
```
But each change here does serve a real purpose. Maybe we should leave more comments in the source about why. For more complicated layers, the source was never all that self-explanatory, so there's less to lose.

src/layers/recurrent.jl
ToucheSir: Correct me if I'm wrong, but wouldn't the weights be Duals when doing nested diff with Zygote + ForwardDiff? Or would the tags differ there too?

ToucheSir:
This complexity was inevitable at some point. We can always consider encapsulating some of it into NNlib functions in the future.

src/layers/recurrent.jl
ToucheSir: Would it be easy to just revert this part of the PR if we run into issues? If so, we can at least try it.

mcabbott: Sure. Could have precisely old and new signatures, calling the same internal function.

ToucheSir (approved):

src/layers/stateless.jl
```diff
@@ -57,3 +57,47 @@ true
...
+  convert(AbstractArray{Float32}, x)
+end
+
+# Allow OneHot to reach specialisation of * etc:
+_match_eltype(layer, ::Type, x::OneHotLike) = x
+
+# Other floats, and integers, silently fix.
+function _match_eltype(layer, ::Type{T}, x::AbstractArray{<:Union{AbstractFloat, Integer}}) where {T}
+  convert(AbstractArray{T}, x)
+end
```
ChrisRackauckas, line 88: This is not correct for all types, for example TrackedArray and Array{TrackedReal}. This is leading to a good amount of downstream breakage. Is there a reason this wasn't made to be in a breaking update?

ToucheSir: Changing `T<:Union{AbstractFloat,Integer}` as well would likely be sufficient.

> This is leading to a good amount of downstream breakage

Can you provide some links? We made a point of double-checking the reverse CI results for DiffEqFlux, NeuralPDE and OperatorLearning, so if they have insufficient coverage then that's very concerning.

ToucheSir: Is there a reason the relevant test groups for SciMLSensitivity weren't added alongside https://github.com/FluxML/Flux.jl/blob/master/.github/workflows/Downstream.yml#L28-L30? I don't think anyone actively searches for new downstream tests unless they're in the same org.

mcabbott: > Is there a reason this wasn't made to be in a breaking update?

Yes, the reason is that there is no plan to stop supporting code which used to work.

This is a bug, and (once again) it turns out that people rely on Flux in ways which aren't tested. Besides adding more downstream tests, it would be extremely helpful if someone who needs Tracker could write up some small self-contained set of Tracker tests, to include here.

ChrisRackauckas: Is there an issue with the self-contained tracker test above?

mcabbott: Doesn't test Conv, Dropout, BatchNorm. Doesn't test any mixtures of tracked and not tracked. Relies on Optimisers.jl. Does not contain `@test`. Is not a PR to `/test/...`. Anything else?

ChrisRackauckas: > Which nobody is obliged to work on. But flippant suggestions that the MWE is somehow equivalent to upgraded tests get flippant responses.

I was just asking if that MWE was a good start to making a test or whether you were seeing the issue as something different. All you had to say was that it would be good if it checked a few more layer types (and remove the restructure? Is that the Optimisers part you were mentioning?) And I will include that in the PR, along with adding SciMLSensitivity core 1-5 as downstream tests, and a package extension to fix the Tracker case (though this really needs a new array primitive, so I was testing the promote eltype stuff in Array interface and patching all odd array types). But at least I understand now why I get so many DMs scared of contributing to Flux...

ChrisRackauckas: > Yes, the reason is that there is no plan to stop supporting code which used to work.

The purpose of the PR is to change the output of every call that was doing such mixed precision. It changes the element type and the result of a lot of things, and explains some other tests which broke (SciMLDocs). I think that would have at least been discussed as a breaking change which is why I was confused that I didn't find such a discussion. Changing something that was 32 but precision to 64 bit is borderline, but changing a function call from 64 to 32 is going to give a different output. I mean it's fine if the answer is that the results are not guaranteed to that level and therefore it's not breaking, or sorry we didn't catch it in any test so let's improve downstream testing, but I just didn't know where to find the conversation.

mcabbott: Yea it's a bug, we thought about and tested dual numbers, but apparently failed to think about Tracker. Won't be hard to fix but won't happen today. Tracker lies about element types so a bit of care will be needed to get it right & test to ensure it cannot go wrong again.

Accidental promotion to Float64 was the most common Flux performance bug, and the one the docs spend pages trying to teach you to avoid. I was trying to re-write these to be clearer but really it's seems crazy to spend so much text teaching about a footgun we can simply remove. The other half of this footgun was accidental promotion of gradients, which is harder to see, and similarly solved some time ago.

When we discussed this, nobody could picture a real reason to run a model with low-precision weights and high-precision data. For high-precision results, it is of course no problem to run a model with high-precision weights.

> changes the element type and the result of a lot of things

What does "a lot of things" mean?

PR #2167, opened 2023-01-25
Remove piratical array conversion rule

ToucheSir:
Seeing what breaks here. If nothing, we should drop it entirely.


ToucheSir:
GPU CI looks good and I realized too soon downstream CI doesn't have GPU access. @chengchingwen given you authored https://github.com/FluxML/Flux.jl/pull/2034, would you be affected by this removal? cc also @yuehhua since I see you mentioned #2033 in a PR.

chengchingwen:
#2034 just tried to fix the wrong dispatch rule, so I'm totally fine with the entire removal. But I think @yuehhua used that in his package.

yuehhua:
I just checked it and it doesn't break anything in my side. It's fine to drop it entirely. Thanks for remind.

CarloLucibello (approved):

Issue #2171, opened 2023-02-01
On some recent Flux experiences


mcabbott:
Apparently @owainkenwayucl was trying out Flux, and @giordano was helping him out. 

https://twitter.com/owainkenway/status/1620771011863121921

https://github.com/owainkenwayucl/JuliaML/blob/main/Fashion/simple.jl

It's useful to see what problems newcomers run into. Especially people new to Julia.

* Scope issues like `global a_sum_ = a_sum_ + 1` are weird. Flux's tutorials tend to define many functions to put everything in local scope... maybe too many... but for this common use, perhaps Flux ought to have an `accuracy` function instead of every tutorial rolling its own?

* Wrong array dimensions give pretty confusing errors. Perhaps Flux layers should catch more of them, instead of waiting for `*` etc. Some made-up examples (but examples from the wild might be different):

```julia
julia> Conv((3,3),3=>4)(rand(10,10,3))  # reasonable to try, maybe it should just work
ERROR: DimensionMismatch: Rank of x and w must match! (3 vs. 4)

julia> Conv((3,3),3=>4)(rand(10,10,2,1))  # error could print out which Conv layer, and size of input, etc.
ERROR: DimensionMismatch: Input channels must match! (2 vs. 3)

julia> Dense(2,3)(rand(4))  # error could be from Dense, there is no matrix called A in user code
ERROR: DimensionMismatch: second dimension of A, 2, does not match length of x, 4
```

* Perhaps we can encourage more use of `outputsize` for understanding array sizes. There could be some way to get an overview of the whole model, like this:  https://flax.readthedocs.io/en/latest/getting_started.html#view-model-layers . The default `show` doesn't know the input size, at present, so it can't tell you all of this. One idea would be to give Chain a mutable field in which to store the most recent input size?

skyleaworlder:
It seems that storing input size is a necessary way. Chain itself is a good place for both points: clear responsibilities and uncomplicated design.

In flax, all operators push their input & output size to a list (`call_info_stack`) which is included by a global thread local object (`_DynamicContext`). If `tabulate` called, flax would go to fetch the last record from `call_info_stack` and format its properties. This is another choice.

ToucheSir:
[...]

PR #2176, opened 2023-02-03
Add friendly size check

mcabbott:
Partly addresses #2171, by catching more size mistakes early enough to print out what layer is the problem:
```julia
julia> Conv((3,3),3=>4)(rand(10,10,2,1))  # say which Conv layer, etc.
ERROR: DimensionMismatch: layer Conv((3, 3), 3 => 4) expects size(x, 3) == 3, but got x = 10×10×2×1 Array{Float64, 4}

julia> Dense(2,3)(rand(4))  # there is no matrix called A in user code
ERROR: DimensionMismatch: layer Dense(2 => 3) expects size(x, 1) == 2, but got x = 4-element Vector{Float64}
```

### DIFF
--- a/src/outputsize.jl
+++ b/src/outputsize.jl
@@ -62,7 +62,7 @@ which should work out of the box for custom layers.
 If `m` is a `Tuple` or `Vector`, its elements are applied in sequence, like `Chain(m...)`.
 
 # Examples
-```julia-repl
+```jldoctest
 julia> using Flux: outputsize
 
 julia> outputsize(Dense(10 => 4), (10,); padbatch=true)
@@ -80,7 +80,7 @@ julia> outputsize(m, (10, 10, 3, 64))
 (6, 6, 32, 64)
 
 julia> try outputsize(m, (10, 10, 7, 64)) catch e println(e) end
-DimensionMismatch("layer Conv((3, 3), 3 => 16) expects size(x, 3) == 3, but got x = 10×10×7×64 Array{Flux.NilNumber.Nil, 4}")
+DimensionMismatch("layer Conv((3, 3), 3 => 16) expects size(input, 3) == 3, but got 10×10×7×64 Array{Flux.NilNumber.Nil, 4}")
 
 julia> outputsize([Dense(10 => 4), Dense(4 => 2)], (10, 1)) # Vector of layers becomes a Chain
 (2, 1)
--- a/src/outputsize.jl
+++ b/src/outputsize.jl
@@ -80,8 +80,6 @@ julia> outputsize(m, (10, 10, 3, 64))
 (6, 6, 32, 64)
 
 julia> try outputsize(m, (10, 10, 7, 64)) catch e println(e) end
-┌ Error: layer Conv((3, 3), 3 => 16), index 1 in Chain, gave an error with input of size (10, 10, 7, 64)
-└ @ Flux ~/.julia/dev/Flux/src/outputsize.jl:107
 DimensionMismatch("layer Conv((3, 3), 3 => 16) expects size(x, 3) == 3, but got x = 10×10×7×64 Array{Flux.NilNumber.Nil, 4}")
 
 julia> outputsize([Dense(10 => 4), Dense(4 => 2)], (10, 1)) # Vector of layers becomes a Chain
@@ -97,19 +95,6 @@ nil_input(pad::Bool, s::Tuple{Vararg{Integer}}) = pad ? fill(nil, (s...,1)) : fi
 nil_input(pad::Bool, multi::Tuple{Vararg{Integer}}...) = nil_input.(pad, multi)
 nil_input(pad::Bool, tup::Tuple{Vararg{Tuple}}) = nil_input(pad, tup...)
 
-function outputsize(m::Chain, inputsizes::Tuple{Vararg{Integer}}...; padbatch=false)
-  x = nil_input(padbatch, inputsizes...)
-  for (i,lay) in enumerate(m.layers)
-    try
-      x = lay(x)
-    catch err
-      str = x isa AbstractArray ? "with input of size $(size(x))" : ""
-      @error "layer $lay, index $i in Chain, gave an error $str"
-      rethrow(err)
-    end
-  end
-  return size(x)
-end
 
 """
     outputsize(m, x_size, y_size, ...; padbatch=false)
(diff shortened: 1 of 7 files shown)


src/outputsize.jl
```diff
@@ -80,9 +80,9 @@ julia> outputsize(m, (10, 10, 3, 64))
 (6, 6, 32, 64)
 
 julia> try outputsize(m, (10, 10, 7, 64)) catch e println(e) end
-┌ Error: layer Conv((3, 3), 3=>16), index 1 in Chain, gave an error with input of size (10, 10, 7, 64)
-└ @ Flux ~/.julia/dev/Flux/src/outputsize.jl:114
-DimensionMismatch("Input channels must match! (7 vs. 3)")
+┌ Error: layer Conv((3, 3), 3 => 16), index 1 in Chain, gave an error with input of size (10, 10, 7, 64)
+└ @ Flux ~/.julia/dev/Flux/src/outputsize.jl:107
+DimensionMismatch("layer Conv((3, 3), 3 => 16) expects size(x, 3) == 3, but got x = 10×10×7×64 Array{Flux.NilNumber.Nil, 4}")
```
mcabbott, line 85: `outputsize(::Chain, ...)` has code to print `@error` on mismatches, before throwing the underlying error. But perhaps we should now just delete that, if the error itself is clear enough?

test/outputsize.jl
```diff
@@ -144,7 +144,6 @@ end
   @test outputsize(m, (32, 32, 3); padbatch=true) == (32, 32, 3, 1)
   m2 = LayerNorm(3, 2)
   @test outputsize(m2, (3, 2)) == (3, 2) == size(m2(randn(3, 2)))
-  @test outputsize(m2, (3,)) == (3, 2) == size(m2(randn(3, 2)))
```
mcabbott, line 147: Allowing `LayerNorm(3, 2)(rand(3))` to work, and make something of size `3,2`, seems pretty weird. Surely not intentional, it jut happened that an implementation with broadcasting allowed this.

`Scale` still does, but `LayerNorm` now calls `_size_check` and does not.

test/cuda/layers.jl
```diff
@@ -115,7 +115,7 @@ dropout_layers = [Dropout, AlphaDropout]
 gpu_gradtest("Dropout", dropout_layers, r, 0.5f0; test_cpu = false) # dropout is not deterministic
 
 layer_norm = [LayerNorm]
-gpu_gradtest("LayerNorm 1", layer_norm, rand(Float32, 28,28,3,4), 1, test_cpu = false) #TODO fix errors
+gpu_gradtest("LayerNorm 1", layer_norm, rand(Float32, 28,28,3,4), 28, test_cpu = false) #TODO fix errors
```
mcabbott, line 118: This GPU test seems to be the only one of this behaviour:
```
julia> LayerNorm(1)([1 2 3; 4 5 6])
2×3 Matrix{Float64}:
 -0.999993  -0.999993  -0.999993
  0.999993   0.999993   0.999993
```
Is this supported or not?

At present the PR changes it to an error:
```
julia> LayerNorm(1)([1 2 3; 4 5 6])
ERROR: DimensionMismatch: layer LayerNorm(1) expects size(input, 1) == 1, but got 2×3 Matrix{Int64}
Stacktrace:
 [1] _size_check(layer::LayerNorm{typeof(identity), Flux.Scale{typeof(identity), Vector{Float32}, Vector{Float32}}, Float32, 1}, x::Matrix{Int64}, ::Pair{Int64, Int64})
   @ Flux ~/.julia/dev/Flux/src/layers/basic.jl:194
```

src/outputsize.jl
skyleaworlder: I think it's okay. BTW, I think the fatal problem is that it cannot process recursive Chain as expected. Once a mismatch event in nested Chain happens, previous `outputsize` only simply prints all the inner layers of the outmost Chain layer (

mcabbott: Yes. It's possible that we were thinking about a more elaborate recursive `outputsize` at that point... or just that we weren't thinking...

test/cuda/layers.jl
ToucheSir: Blame shows https://github.com/FluxML/Flux.jl/commit/717ad9328da73facb287660053982f3c6d7792df#diff-d3a447abb172d95ef564228a96b95372358974a7a2b4f4d5b13e7a36ae235922R43 with a revision (that shouldn't have changed the overall behaviour) in https://github.com/FluxML/Flux.jl/commit/30cb0a84395126f3fd7891b97afb5cd43f1433f1. Unless @CarloLucibello recalls the history behind this, I agree it should be an error.

src/outputsize.jl
ToucheSir: The extra wayfinding within `Chain` seems nice, but if we go with the `trace` suggestion in https://github.com/FluxML/Flux.jl/issues/2171#issuecomment-1416824986 I guess it becomes redundant?

mcabbott: We can keep the Chain method, but it's only used for the outermost Chain. I think that usually if your model is one Chain, the individual layers are probably recognisable enough? 

Without doing the full trace, another option would be to have a method `(c::Chain)(x::Array{Nil})` instead of `outputsize(m::Chain, ...`. That would apply to every Chain even deeply nested. But IDK how useful its output is... maybe in complex models it just ends up printing pages of noise.

ToucheSir (approved):
Anything left to do here?

src/outputsize.jl
```diff
@@ -80,9 +80,7 @@ julia> outputsize(m, (10, 10, 3, 64))
 (6, 6, 32, 64)
 
 julia> try outputsize(m, (10, 10, 7, 64)) catch e println(e) end
-┌ Error: layer Conv((3, 3), 3=>16), index 1 in Chain, gave an error with input of size (10, 10, 7, 64)
-└ @ Flux ~/.julia/dev/Flux/src/outputsize.jl:114
-DimensionMismatch("Input channels must match! (7 vs. 3)")
+DimensionMismatch("layer Conv((3, 3), 3 => 16) expects size(x, 3) == 3, but got x = 10×10×7×64 Array{Flux.NilNumber.Nil, 4}")
```
mcabbott, line 83: Maybe without the warning this can be made a doctest. It may not have the latest error message...

PR #2177, opened 2023-02-04
cuda 4.0 compat

CarloLucibello:
supersedes #2172

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -206,9 +206,6 @@ end
 function check_use_cuda()
   if use_cuda[] === nothing
     use_cuda[] = CUDA.functional()
-    if use_cuda[] && !CUDA.has_cudnn()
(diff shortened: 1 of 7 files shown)


ToucheSir (approved):

src/functor.jl
```diff
@@ -206,9 +206,6 @@ end
 function check_use_cuda()
   if use_cuda[] === nothing
     use_cuda[] = CUDA.functional()
-    if use_cuda[] && !CUDA.has_cudnn()
```
maleadt, line 209: Dropping `has_cudnn` isn't correct; CUDNN.jl still has a `has_cudnn` that should be checked. CUDNN not being available can happen when using a toolkit that isn't supported, or when using an unsupported platform. In the future, CUDNN.jl will just fail to init in this scenario.

PR #2141, opened 2022-12-26
accept integer labels in (logit)crossentropy

CarloLucibello:
A convenience method that avoids the onehot transformation of the labels.
Pytorch supports this [as well](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html#torch.nn.CrossEntropyLoss). 

Performance seems to be approximately the same on both CPU and GPU:
```julia
using Flux, BenchmarkTools, Flux.Losses, CUDA

function perf(c, n)
    labels = rand(1:c, n)
    y = Flux.onehotbatch(labels, 1:c)
    ŷ = randn(Float32, c, n)
    
    labelsgpu = labels |> gpu
    ygpu = y |> gpu
    ŷgpu = ŷ |> gpu

    println("with ŷ")
    @btime logitcrossentropy($ŷ, $y); 
    @btime gradient(ŷ -> logitcrossentropy(ŷ, $y), $ŷ);
    
    println("with labels")
    @btime logitcrossentropy($ŷ, $labels);
    @btime gradient(ŷ -> logitcrossentropy(ŷ, $labels), $ŷ);
    
    println("with ŷ - gpu")
    @btime CUDA.@sync logitcrossentropy($ŷgpu, $ygpu); 
    @btime CUDA.@sync gradient(ŷ -> logitcrossentropy(ŷ, $ygpu), $ŷgpu);
    
    println("with labels - gpu")
    @btime CUDA.@sync logitcrossentropy($ŷgpu, $labelsgpu);
    @btime CUDA.@sync gradient(ŷ -> logitcrossentropy(ŷ, $labelsgpu), $ŷgpu);
    
    return nothing
end

perf(10, 128)
# with ŷ
#   14.648 μs (10 allocations: 13.17 KiB)
#   27.381 μs (19 allocations: 35.39 KiB)
# with labels
#   13.716 μs (16 allocations: 9.88 KiB)
#   41.338 μs (119 allocations: 25.22 KiB)
# with ŷ - gpu
#   46.107 μs (163 allocations: 8.52 KiB)
#   109.656 μs (414 allocations: 24.17 KiB)
# with labels - gpu
#   42.620 μs (125 allocations: 6.23 KiB)
#   117.972 μs (375 allocations: 19.61 KiB)

perf(100, 128)
# with ŷ
#   121.148 μs (12 allocations: 103.02 KiB)
#   212.059 μs (25 allocations: 304.92 KiB)
# with labels
[...]

PR #2181, opened 2023-02-05
Propose accuracy functions

skyleaworlder:
### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

About #2171 | accuracy function

I simply define 3 accuracy function for multi-class and multi-label problem.

```jl
julia> typeof(test_x)
Matrix{Float32} (alias for Array{Float32, 2})

julia> typeof(test_y)
Vector{Int64} (alias for Array{Int64, 1})

julia> multiclass_accuracy(model, test_x, test_y);
```

```jl
julia> typeof(test_x)
Matrix{N0f8} (alias for Array{FixedPointNumbers.Normed{UInt8, 8}, 2})

julia> typeof(test_y)
OneHotMatrix{UInt32, Vector{UInt32}} (alias for OneHotArray{UInt32, 1, 2, Array{UInt32, 1}})

julia> multiclass_accuracy(model, test_x, test_y);

julia> multilabel_accuracy(model, test_x, test_y);
```

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -1,0 +1,58 @@
+using OneHotArrays: OneHotMatrix, onecold, onehotbatch
+using Statistics: mean
+
+"""
+    multiclass_accuracy(model, test_x, test_y::Vector{T})
+
+Used to calculate accuracy of multi-class problem.
+Test set consists of `test_x` and `test_y`. `size(test_x)[end]` and `size(test_y)[end]`
+equal to the number of samples. `test_x` contains batch of samples, and `test_y` is a Vector,
+which contains the position of the largest element of the output (`findmax(output)`)
+obtained by feeding `test_x` into the ideal model.
+
+e.g.
+
+```jldoctest
+julia> typeof(test_x)
+Matrix{Float32} (alias for Array{Float32, 2})
+
+julia> typeof(test_y)
+Vector{Int64} (alias for Array{Int64, 1})
+
+julia> multiclass_accuracy(model, test_x, test_y);
+```
+"""
+multiclass_accuracy(model, test_x, test_y::Vector{T}) where {T} = mean(onecold(model(test_x)) .== test_y)
+
+"""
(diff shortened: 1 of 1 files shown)


CarloLucibello:
I'm not sure we want to add accuracy metrics in Flux, but if we do they should take the signature `metric(yhat, y)` and not `metric(model, x, y)`

ToucheSir:
I think we should add them somewhere officially supported, since we've seen a lot of people end up rolling their own less-than-optimal versions. The question is where? Do we have the capacity to revive https://github.com/JuliaML/MLMetrics.jl?

mcabbott:
We should think more about the right signatures. 

Matching the categorical functions in `Flux.Losses` suggests:
```
accuracy(yhat::AbstractMatrix, y::OneHotMatrix)         # like crossentropy
accuracy(yhat::AbstractMatrix, y::AbstractVector{Bool}) # like binarycrossentropy
```
If the first class are expanded to take labels instead (like #2141) then  
```
accuracy(yhat::AbstractMatrix, y::AbstractVector, labels=1:size(yhat,1))
```
If all loss functions accept the model as 1st argument (#2090) then perhaps you want `accuracy(m, x, y) = accuracy(m(x), y)`. But that loss signature is mostly because it would be convenient to pass to `train!`, how useful is it here?

Matching the input of `train!` is another possibility. Having made a DataLoader with batches, this function could perhaps iterate over it for you:
```
accuracy(model, data) = mean(accuracy(model(x), y) for (x,y) in data)
```

Those seem the obvious reference points if this lives within Flux. If it lives in OneHotArrays, then perhaps the `binarycrossentropy`-like signature is out of scope.

skyleaworlder:
I use `multilabel_accuracy(model, test_x, test_y)` because I think that the point of the semantics of this function signature is more like a model than test dataset. If it were data, there would be some libraries that could have already done well for it.

If https://github.com/JuliaML/MLMetrics.jl is enough, applying it in docs / tutorials is also a good choice, which can also help new users reduce work in testing model.

CarloLucibello:

PR #2132, opened 2022-12-09
[WIP] Move CUDA support to a package extension

IanButterworth:
Uses [new package extensions](https://github.com/JuliaLang/julia/pull/47695) (due to be in julia 1.9), to make CUDA functionality optional.
On julia versions that don't have package extensions Flux will work as before.

I've only done the obvious things here to get loading working and haven't run tests locally.
I took the approach of maintaining file names of where functions came from to make it easier to cross-reference. It could all be consolidated if preferred.

updated:
```julia
julia> @time using Flux
  5.154103 seconds (5.45 M allocations: 381.721 MiB, 2.06% gc time, 99.50% compilation time: 37% of which was recompilation)

julia> x = rand(2)
2-element Vector{Float64}:
 0.8415303775713626
 0.26837175189895346

julia> gpu(x)
[ Info: The GPU functionality is being called via `Flux.gpu` but `NNlibCUDA` must be loaded to access GPU functionality
2-element Vector{Float64}:
 0.8415303775713626
 0.26837175189895346

julia> @time using NNlibCUDA
  7.160284 seconds (20.20 M allocations: 1.142 GiB, 3.41% gc time, 42.59% compilation time: 82% of which was recompilation)

julia> gpu(x)
┌ Info: The GPU function is being called but the GPU is not accessible.
└ Defaulting back to the CPU. (No action is required if you want to run on the CPU).
2-element Vector{Float64}:
 0.8415303775713626
 0.26837175189895346
```

For reference, Flux master on julia master
```
julia> @time using Flux
 12.092398 seconds (25.81 M allocations: 1.526 GiB, 3.02% gc time, 65.40% compilation time: 57% of which was recompilation)
```

Helps with https://github.com/FluxML/Flux.jl/issues/1961

cc. @mcabbott 

[Edit: closes #2155]

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md
- [ ] Documentation, if applicable

PR #2182, opened 2023-02-06
CUDA code refactoring

CarloLucibello:
This moves all cuda related code inside the (already existing) CUDAint submodule. 
Now the Flux module doesn't have a direct `using CUDA` statement.

Also
- reintroduces the `has_cudnn` check per comment https://github.com/FluxML/Flux.jl/pull/2177#discussion_r1096579632 (cc @maleadt)
- removes the v0.12 deprecations
- factorized the old `train!` compatibility methods in their own file `deprecations_train.jl` in order to clean up `deprecations.jl`

I'm sorry it predates a bit #2132, but maybe better to start this refactoring now while waiting for julia v1.9 to be released.


CarloLucibello:
Merge?

mcabbott:
What's gained by moving things around, now instead of once we figure out extensions (https://github.com/FluxML/Flux.jl/pull/2132)? 

`has_cudnn` could be separate. 

Removing deprecations should ideally happen right before 0.14.

The train deprecations would probably also go then. What's gained by moving them around?

PR #2186, opened 2023-02-11
RFC: add `gpu(::DataLoader)`

mcabbott:
A common pattern in the model zoo is to copy individual batches to the GPU by hand. This proposes instead making it easy to wrap a `DataLoader` in a `CuIterator`, if CUDA is available, and otherwise do nothing. Then this pair should be all you need:
```
data |> gpu
model |> gpu
```

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -214,6 +214,55 @@ function check_use_cuda()
 end
 ChainRulesCore.@non_differentiable check_use_cuda()
 
+"""
+    gpu(data::DataLoader)
+
+Wraps ths given `DataLoader` in a `CuIterator`, which will
+move each batch of data to the GPU as required.
+
+(If no GPU is available, it does nothing.)
+
+# Example
+
+```julia-repl
+julia> dl = Flux.DataLoader((x = Flux.ones32(2,10),), batchsize=3)
+4-element DataLoader(::NamedTuple{(:x,), Tuple{Matrix{Float32}}}, batchsize=3)
+  with first element:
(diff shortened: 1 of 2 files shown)


ToucheSir:
Considered this before, but one holdup is that `CuIterator` will try to `unsafe_free!` everything in the batch regardless if it's an array or not. We might want to change that before doing this.

mcabbott:
Yes. Also there is this bug:
```
julia> Flux.DataLoader(Flux.ones32(2,10)) |> CuIterator |> first |> typeof
Matrix{Float32} (alias for Array{Float32, 2})

julia> Flux.DataLoader((Flux.ones32(2,10),),) |> CuIterator |> first |> typeof
Tuple{CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}}
```

Whatever types it accepts, it should surely use the same `map` / `adapt` for both creating and freeing CuArrays... but perhaps ` adapt(CuArray, x)` is tricky?

https://github.com/JuliaGPU/CUDA.jl/blob/5525c15eb5d21df6a9d86f07c554566b2018bff5/src/iterator.jl#L21-L29

Since Flux gives an error right now, it could test that the particular DataLoader is safe, and error on construction if not.

mcabbott:
One thing https://github.com/JuliaGPU/CUDA.jl/pull/1769 won't do is that, while `gpu` converts to Float64 like `cu`, the `CuIterator` does not.

Possibly we should just build something locally which is equivalent, and would work for AMD too.

PR #2194, opened 2023-02-21
fixed BSON loadmodel! documentation and added a test case

### LINKED ISSUE #2191: loadmodel! example in the docs doesn't work
The docs (http://fluxml.ai/Flux.jl/stable/saving/) say you can load a model with : 

`model = loadmodel!(model, @load("mymodel.bson"))` 

but `@load("mymodel.bson")` returns nothing.

I think it should be something like `loadmodel!(model, load("mymodel.bson")[:model])`. Also it looks like this case is not tested (can't find it in test/utils.jl).

Flux v0.13.11 & BSON v0.3.6

jonathanBieler:
Fixes #2191

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md


ToucheSir (approved):
Thanks!

PR #2196, opened 2023-02-22
reintroduce flatten

### LINKED ISSUE #2195: Flattern removal
Hi there, the removal of Flatten in version 0.13.13 means that models trained and saved from legacy versions of Flux are not loaded properly (the flatten layer is completely gone). Is there a way to ensure the flatten layer is still imported? The models loaded correctly under version 0.12.8. The models were imported from JLD2 files, and the entire model, except the flatten layer was missing, without any way to recover it.

CarloLucibello:
fix #2195

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -187,6 +187,8 @@ function update!(opt::Optimise.AbstractOptimiser, ::Params, grads::Union{Tuple, 
     """)
 end
 
+# https://github.com/FluxML/Flux.jl/issues/2195
+const flatten = MLUtils.flatten
 
 # v0.14 deprecations
 


mcabbott:
By the way, did we make the right choice with building reshape into Dense? We could have built in `flatten` instead, i.e. regard the `d = max(ndims(x), 2)` dimension as the batch, always, and all others as feature dims. I suspect this use is much more common than having several batch dimensions.

CarloLucibello:
>  I suspect this use is much more common than having several batch dimensions.

well in transformers you have shape `[embed_dim, seq_length, batch_dim]` and there are MLP blocks that you want to apply considering the last 2 dims as batch but preserving the shape

PR #2197, opened 2023-02-28
fix various deprecation warnings

ExpandingMan:
I have fixed some deprecation warnings which were driving me nuts.  There was a simple warning from a use of the deprecated macro `Zygote.@nograd`.  The rest were caused by a tricky issue with the use of `Vararg`.  When one writes a function signature with arguments like `::Tuple{Vararg{<:T}}` one gets the deprecation warning
```
WARNING: Wrapping `Vararg` directly in UnionAll is deprecated (wrap the tuple instead).
```
The solution is apparently to replace the `<:T` expressions with `T`, which is equivalent due to tuples being covariant.  This property of tuples is horribly obscure, so I've added comments explaining what is going on.

### PR Checklist

- [x] Tests are added (all changes on existing code, no new tests should be needed)
- [x] Entry in NEWS.md
- [x] Documentation, if applicable (not applicable)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # Flux Release Notes
 
+## v0.13.14
+* Fixed various deprecation warnings, from `Zygone.@nograd` and `Vararg`.
+
 ## v0.13.13
 * Added `f16` which changes precision to `Float16`, recursively.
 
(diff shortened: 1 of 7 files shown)


ExpandingMan:
I'm assuming nightly is not required to pass since I don't think those errors are related.

skyleaworlder:
> I'm assuming nightly is not required to pass since I don't think those errors are related.

Yes. This error is of julia core instead of this PR.

mcabbott (approved):

PR #2201, opened 2023-03-04
don't import losses in Flux's namespace

CarloLucibello:


ToucheSir:
Do we have dep warnings for the losses which will no longer be available? If not I don't think this needs to be part of a major release.

mcabbott:
What's gained by doing this?

It will certainly break lots of code, [e.g. this](https://github.com/FluxML/model-zoo/search?q=Flux.logitcrossentropy).

Flux is otherwise quite permissive,e.g.  you are not required to notice that `Flux.onehotbatch`, `Flux.flatten` and `Flux.destructure` all come from outside packages. Is it a good idea to require you to know of the existence of this internal sub-module?

src/Flux.jl
```diff
@@ -67,7 +67,6 @@ include("outputsize.jl")
 export @autosize
 
 include("losses/Losses.jl")
-using .Losses # TODO: stop importing Losses in Flux's namespace in v0.12
```
CarloLucibello, line 70:
```suggestion
using .Losses # TODO: stop importing Losses in Flux's namespace in v0.12

```

CarloLucibello:
let's keep it then, I just removed the comment

mcabbott (approved):

CarloLucibello:
argh, I should have edited the title before merging 😭

PR #2203, opened 2023-03-05
Use consistent spelling for optimise

jeremiahpslewis:
Flux.jl uses alternately optimize and optimise, in various parts of the project, which can be a bit tricky, particularly as the JuMP.jl ecosystem seems to have chosen optimize. Not sure whether it would make sense to support both spellings, but in the meantime it seems best to be internally consistent.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,5 @@
 # Flux Release Notes
 
-* Make spelling of optimise consistent throughout the project.
 
 ## v0.13.14
 * Fixed various deprecation warnings, from `Zygone.@nograd` and `Vararg`.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # Flux Release Notes
 
+* Make spelling of optimise consistent throughout the project.
+
 ## v0.13.14
 * Fixed various deprecation warnings, from `Zygone.@nograd` and `Vararg`.
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -45,7 +45,7 @@ been removed in favour of MLDatasets.jl.
 * Fixed [AlphaDropout](https://github.com/FluxML/Flux.jl/pull/1781)
 
 ## v0.12.8
-* Optimized inference and gradient calculation of OneHotMatrix[pr](https://github.com/FluxML/Flux.jl/pull/1756)
+* Optimised inference and gradient calculation of OneHotMatrix[pr](https://github.com/FluxML/Flux.jl/pull/1756)
 
 ## v0.12.7
 * Added support for [`GRUv3`](https://github.com/FluxML/Flux.jl/pull/1675)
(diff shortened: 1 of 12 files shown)


bhvieira:
I'm in favor of "optimiz*" everywhere.
"Z" is much more common worldwide, while "s" is an alternative spelling that also exists everywhere.
This change would require a deprecation warning until the next major release only. And perhaps, in the future a warning can be thrown if people try the old names. Better than keeping aliases imo.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # Flux Release Notes
 
+* Make spelling of optimise consistent throughout the project.
```
mcabbott, line 3: Not sure anyone needs to find out which versions are before/after this:
```suggestion
```

mcabbott (approved):
This seems OK, it's a minor spelling fix to docs. There is no way that the docs are ever likely to be fully consistent about using UK spelling, although I do think that's the convention.

Trying to avoid code which requires one spelling seems like a good prinicple. E.g. we should kill `Flux.normalise`. 

Am I correct in thinking that the only place where Flux's code cares about the spelling of "optimiser" is the package name Optimisers.jl ? Renaming that would be a much bigger step. Perhaps one we should have contemplated before registering it -- I mean to some other name completely, both to avoid words whose spelling varies, and to avoid the (temporary) clash with the Flux.Optimise sub-module.

docs/src/gpu.md
```diff
@@ -138,12 +138,12 @@ In order to train the model using the GPU both model and the training data have
...
    ```julia
    train_loader = Flux.DataLoader((xtrain, ytrain), batchsize = 64, shuffle = true)
-   # ... model, optimizer and loss definitions
+   # ... model, optimiser and loss definitions
    for epoch in 1:nepochs
        for (xtrain_batch, ytrain_batch) in train_loader
            x, y = gpu(xtrain_batch), gpu(ytrain_batch)
            gradients = gradient(() -> loss(x, y), parameters)
-           Flux.Optimise.update!(optimizer, parameters, gradients)
+           Flux.Optimise.update!(optimiser, parameters, gradients)
```
mcabbott, line 146: This ought to be re-written in gradients parameters anyway, in which case the variable will be `opt_state`

docs/src/saving.md
```diff
@@ -129,10 +129,10 @@ revert to an older copy of the model if it starts to overfit.
...
+Note that to resume a model's training, you might need to restore other stateful parts of your training loop. Possible examples are stateful optimisers (which usually utilize an `IdDict` to store their state), and the randomness used to partition the original data into the training and validation sets.
 
 You can store the optimiser state alongside the model, to resume training
-exactly where you left off. BSON is smart enough to [cache values](https://github.com/JuliaIO/BSON.jl/blob/v0.3.4/src/write.jl#L71) and insert links when saving, but only if it knows everything to be saved up front. Thus models and optimizers must be saved together to have the latter work after restoring.
+exactly where you left off. BSON is smart enough to [cache values](https://github.com/JuliaIO/BSON.jl/blob/v0.3.4/src/write.jl#L71) and insert links when saving, but only if it knows everything to be saved up front. Thus models and optimisers must be saved together to have the latter work after restoring.
```
mcabbott, line 135: To be re-written for explicit gradients

docs/src/tutorials/2021-02-07-convnet.md
```diff
@@ -145,7 +145,7 @@ function train(; kws...)
        return logitcrossentropy(ŷ, y)
    end
   
-   # Train our model with the given training set using the Adam optimizer and
+   # Train our model with the given training set using the Adam optimiser and
```
mcabbott, line 148: This file to be deleted in favour of the model zoo, perhaps https://github.com/FluxML/model-zoo/pull/394

src/optimise/optimisers.jl
```diff
@@ -45,7 +45,7 @@ end
 """
     Momentum(η = 0.01, ρ = 0.9)
 
-Gradient descent optimizer with learning rate `η` and momentum `ρ`.
+Gradient descent optimiser with learning rate `η` and momentum `ρ`.
```
mcabbott, line 48: This file will be deleted in 0.14

PR #2205, opened 2023-03-08
dummy implementation of flatten

### LINKED ISSUE #2195: Flattern removal

### LINKED ISSUE #2204: `UndefVarError: #flatten not defined` with existing model
Has there been an non-backwards-compatible change in the Flux internals regarding the use of CNNs? We are regularly running tests in Trixi.jl with existing, older CNN/FFNN models, and now all of a sudden some tests fail during BSON loading with what I assume is due to some change in the Flux.jl API:
```
  LoadError: UndefVarError: #flatten not defined
  
  Some of the types have been truncated in the stacktrace for improved reading. To emit complete information
  in the stack trace, evaluate `TruncatedStacktraces.VERBOSE[] = true` and re-run the code.
  
  Stacktrace:
    [1] #31
      @ ~/.julia/packages/BSON/73cTU/src/extensions.jl:21 [inlined]
    [2] (::Base.BottomRF{BSON.var"#31#32"})(acc::Module, x::String)
      @ Base ./reduce.jl:81
    [3] _foldl_impl(op::Base.BottomRF{BSON.var"#31#32"}, init::Module, itr::Vector{Any})
      @ Base ./reduce.jl:62
    [4] foldl_impl
      @ ./reduce.jl:48 [inlined]
    [5] mapfoldl_impl
      @ ./reduce.jl:44 [inlined]
    [6] _mapreduce_dim
      @ ./reducedim.jl:327 [inlined]
    [7] #mapreduce#731
      @ ./reducedim.jl:322 [inlined]
    [8] #reduce#733
      @ ./reducedim.jl:371 [inlined]
    [9] resolve(fs::Vector{Any}, init::Module)
      @ BSON ~/.julia/packages/BSON/73cTU/src/extensions.jl:21
   [10] (::BSON.var"#35#36")(d::Dict{Symbol, Any}, init::Module)
[...]
```
Full error message:
https://github.com/trixi-framework/Trixi.jl/actions/runs/4361886357/jobs/7626199321#step:7:4552

Can someone quickly confirm that there was such a change and would happen to know how to fix it?

CarloLucibello:
Unfortunately, the workaround #2196 doesn't work. I can't get BSON to recognize a method transferred to another package, so we have to define our own `flatten` in order to load old bson checkpoints.

Fix #2204 fix #2195


ToucheSir:
Docs CI seems to be failing.

docs/make.jl
```diff
@@ -1,7 +1,9 @@
 using Documenter, Flux, NNlib, Functors, MLUtils, BSON, Optimisers, OneHotArrays, Zygote, ChainRulesCore, Plots, MLDatasets, Statistics, DataFrames
 
 
-DocMeta.setdocmeta!(Flux, :DocTestSetup, :(using Flux); recursive = true)
+DocMeta.setdocmeta!(Flux, :DocTestSetup, 
+    :(using Flux; using Flux: flatten);
```
CarloLucibello, line 5: This didn't work. @ToucheSir you know why?

ToucheSir: No idea, unfortunately. I rarely touch docs stuff.

darsnack: You could try adding `using Flux: flatten` to the `jldoctest` heading directly in `stateless.jl`. And reverting this change.

CarloLucibello: I don't understand. Which jldoctest heading?

ToucheSir (approved):

PR #2212, opened 2023-03-17
Create dependabot.yml

CarloLucibello:
This allows us to get updates for GitHub actions automatically. After merging this, we should also enable other Dependabot actions in "Settings -> Code security and analysis -> Dependabot alerts" and "... -> Dependabot security updates".

After merging this, Dependabot will prepare PRs with updates of the versions of GitHub actions we use such as.


ToucheSir (approved):
> updates of the versions of GitHub actions we use such as.

We're all on tenterhooks waiting for the list :smile:

Issue #2185, opened 2023-02-09
explicit differentiation for RNN gives wrong results


alerem18:
i tried to implement a RNN MODEL to classify Mnist Dataset but i get an accuracy around 40-50% even with running it for more than 20 epochs, while in pytorch, i'll get an accuracy upto 90% after just 4-5 epochs

here is my code:
```julia
using Flux
using Flux: onehotbatch, onecold, params, gradient
using MLDatasets: MNIST
using Base.Iterators: partition
using TensorCast
using Statistics: mean
using Random: shuffle

#---------------------------------- DATA -------------------------------------
DATA_TRAIN = MNIST.traindata(Float32)
DATA_TEST = MNIST.testdata(Float32)

#-------------------------------- PREPROCESS DATA ------------------------------
@cast x_train[j][i, k] := DATA_TRAIN[1][i, j, k] # reshape to vector of size 28 with matrix of size 28 x 60000
@cast x_test[j][i, k] := DATA_TEST[1][i, j, k] # reshape to vector of size 28 with matrix of size 28 x 10000

# create onehotbatch for train label
y_train = onehotbatch(DATA_TRAIN[2], 0:9)
y_test = DATA_TEST[2]

#------------------------------ CONSTANTS ---------------------------------------
INPUT_DIM = size(x_train[1], 1)
OUTPUT_DIM = 10 # number of classes
LR = 0.001 # learning rate
EPOCHS = 100
BATCH_SIZE = 1000
TOTAL_SAMPLES = size(x_train[1], 2)

#--------------------------------- BUILD MODEL -----------------------------------
struct RnnModel
  rnn
  fc
end

Flux.@functor RnnModel

# pass input thorough MODEL
function (m::RnnModel)(input_data)

  # warmup rnn
  [m.rnn(x) for x ∈ input_data[1:end - 1]]

  # pass latest layer to fc layer
  m.fc(m.rnn(input_data[end]))
end

# build MODEL
model = RnnModel(
  Chain(RNN(INPUT_DIM, 128), relu, RNN(128, 64), relu, RNN(64, 32), relu),
  Chain(Dense(32, OUTPUT_DIM), softmax)
)

#----------------------------- HELPER FUNCTIONS --------------------------------------
loss_fn(x, y) = Flux.Losses.logitcrossentropy(model(x), y)
function accuracy(x, y)
  Flux.reset!(model)
  mean(onecold(model(x), 0:9) .== y)
end

[...]

PR #2215, opened 2023-03-17
manual gradient checks for RNN - implicit and explicit gradients

jeremiedb:
This adds gradient tests for RNN in relation to https://github.com/FluxML/Flux.jl/issues/2185

Note that for implicit gradient mode, gradients successfully pass all tests on all Julia versions. 
Implicit mode gradients only fail on Julia >= 1.7 when in REPL (and gradient isnt' call from within a function).

For explicit mode gradient (new Optimisers.jl), all gradients fail on Julia >= 1.7. 
On Julia v1.6, all gradients other than `state0` are correct. The correct `state0` gradient is actually assigned to Recur's `state` rather than to the cell's `state0`. 

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/test/layers/recurrent.jl
+++ b/test/layers/recurrent.jl
@@ -73,137 +73,122 @@ end
 
 # Ref FluxML/Flux.jl#1209 1D input
 @testset "BPTT-1D" begin
-    seq = [rand(Float32, 2) for i = 1:3]
-    for r ∈ [RNN]
-        rnn = r(2 => 3)
-        Flux.reset!(rnn)
-        grads_seq = gradient(Flux.params(rnn)) do
-            sum([rnn(s) for s in seq][3])
-        end
-        Flux.reset!(rnn)
-        bptt = gradient(
-            Wh -> sum(
-                tanh.(
-                    rnn.cell.Wi * seq[3] +
-                    Wh *
-                    tanh.(
-                        rnn.cell.Wi * seq[2] +
-                        Wh *
-                        tanh.(rnn.cell.Wi * seq[1] + Wh * rnn.cell.state0 + rnn.cell.b) +
-                        rnn.cell.b
-                    ) +
-                    rnn.cell.b
-                ),
-            ),
-            rnn.cell.Wh,
-        )
-        @test grads_seq[rnn.cell.Wh] ≈ bptt[1]
+  seq = [rand(Float32, 2) for i = 1:3]
+  for r ∈ [RNN,]
+    rnn = r(2 => 3)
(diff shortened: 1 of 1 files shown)


test/layers/recurrent.jl
```diff
@@ -1,123 +1,209 @@
...
+        @test_broken ∇Wi ≈ grads[:Wi]
+        @test_broken ∇Wh ≈ grads[:Wh]
+        @test_broken ∇b ≈ grads[:b]
+        @test_broken ∇state0 ≈ grads[:state0]
+    end
+end
+
 # Ref FluxML/Flux.jl#1209 1D input
 @testset "BPTT-1D" begin
-  seq = [rand(Float32, 2) for i = 1:3]
```
ToucheSir, line 5: As much as it pains me to do so, we should probably leave all this formatting alone and tackle it in one fell swoop with a formatting PR. Otherwise it messes up git blame.

ToucheSir (commented):
Did you discover anything in your test creation which might give us a lead on where the incorrect grads are coming from?

test/layers/recurrent.jl
jeremiedb: I've reverted to original format. It's a bit of a second reflex for me to call the auto-formatter while writing code!

jeremiedb:
> Did you discover anything in your test creation which might give us a lead on where the incorrect grads are coming from?

Sorry no new insights at the moment. For explicit mode `state0` issue, I suspect some "alias" from https://github.com/FluxML/Flux.jl/blob/1ca15f35ea35dea8548ffeeefea8a4c22dad413c/src/layers/recurrent.jl#L179.

ToucheSir (approved):
Thanks!

mcabbott:
Would it be easy to add explicit errors to the wrong ones? E.g. by overloading `Zygote.pullback(::Context{true}, ...)` where the [struct has a flag](https://github.com/FluxML/Zygote.jl/blob/1f079016d2c59d57c6c92262bdc79067828a25db/src/compiler/interface.jl#L8-L10) to indicate implicit mode.

ToucheSir:
Unfortunately it's the explicit mode path which is the broken one.

mcabbott:
Sure, but we can dispatch on that too? Haven't tried & not sure whether there's a point at which this could be attached.

ToucheSir:
We can, but we have to find that point first. And if we do, it's likely we'll be able fo fix the bug then without having to put up a "this is broken" sign.

PR #2219, opened 2023-03-22
fix some documenter warnings

CarloLucibello:
Fixes some, but we still have too many warnings.

One of the main issues, is that it seems impossible to tell Documenter that we would like to report some docstrings from other packages for convenience, but it shouldn't check that all docstrings from those peripheral packages are present.

We want to check that all docstrings are present only for exported methods of Flux and NNlib I guess.


docs/make.jl
```diff
@@ -65,6 +63,21 @@ makedocs(
         assets = ["assets/flux.css"],
         prettyurls = get(ENV, "CI", nothing) == "true"
     ),
+    clean = true, 
+    doctest = false,
```
Saransh-cpp, line 67: Why have the doctests been turned off in Flux? I've been away from Flux for a while; will need to dig up git to find the relevant issue/PR. I remember setting up a reliable system for doctests (along with warning checks) for Flux and other neighboring packages, but now all of them are set to false (at least in Flux) 😬 

This is not related to this PR, just general curiosity, thank you for fixing these!

Saransh-cpp:
> One of the main issues, is that it seems impossible to tell Documenter that we would like to report some docstrings from other packages for convenience, but it shouldn't check that all docstrings from those peripheral packages are present.

Yes! It is in fact impossible to do so.

docs/make.jl
ToucheSir: They haven't? Pretty sure this is just WIP work on the PR.

CarloLucibello: I left `doctest = false` unchanged, I was focused on fixing other issues. I can set it to `true` and see what happens.

Saransh-cpp: That was indeed an error on my side. The doctests run on a different step of the workflow independently.

docs/make.jl
```diff
@@ -1,15 +1,21 @@
...
     recursive = true)
 
+DocMeta.setdocmeta!(NNlib, :DocTestSetup, :(using NNlib); recursive=true)
+DocMeta.setdocmeta!(Functors, :DocTestSetup, :(using Functors); recursive=true)
+DocMeta.setdocmeta!(MLUtils, :DocTestSetup, :(using MLUtils); recursive=true)
+DocMeta.setdocmeta!(BSON, :DocTestSetup, :(using BSON); recursive=true)
+DocMeta.setdocmeta!(Optimisers, :DocTestSetup, :(using Optimisers); recursive=true)
+DocMeta.setdocmeta!(OneHotArrays, :DocTestSetup, :(using OneHotArrays); recursive=true)
+DocMeta.setdocmeta!(Zygote, :DocTestSetup, :(using Zygote); recursive=true)
+DocMeta.setdocmeta!(ChainRulesCore, :DocTestSetup, :(using ChainRulesCore, LinearAlgebra); recursive=true)
```
CarloLucibello, line 14: this should be doable within a loop but I'm so weak with quotes/macros that I failed

docs/make.jl
CarloLucibello: With `doctest=true` I hit a bunch of errors, most due to missing symbols, some to actual output changes (maybe different random numbers in recent julia versions). 

I think that for the time being I will just stop here, leaving `doctest=false` as it is. The per-module `:DocTestSetup`  that I added i already a good start for fixing the remaining issues. The rest is left for future PRs (maybe @Saransh-cpp  wants to take over?).

CarloLucibello: On second thought, the way @Saransh-cpp set things up is the correct one and the only one possible:
we have to set `doctest=true` in makedocs, otherwise is going to check docstrings we don't own,
and then one has to call the function `doctest(Flux)` to test only Flux docstrings. 

So the only change I made is to remove the separate CI step doing `doctest(Flux)` and do it in `docs/make.jl` instead, so that doctests can be run locally with ease.

docs/src/tutorials/linear_regression.md
```diff
@@ -126,7 +126,7 @@ A [`Dense(1 => 1)`](@ref Dense) layer denotes a layer of one neuron with one inp
 
 ```jldoctest linear_regression_simple; filter = r"[+-]?([0-9]*[.])?[0-9]+(f[+-]*[0-9])?"
 julia> flux_model.weight, flux_model.bias
-(Float32[1.1412252], Float32[0.0])
+(Float32[-1.2678515;;], Float32[0.0])
```
CarloLucibello, line 129: the reason why doctest were failing on recent julia version is because of this different representation for matrices with one element, and not because of the different numerical output (which I now see is filtered)

Saransh-cpp: Yes, I am not very sure why the doctests ran on `1.6`, but I think maybe because `1.6` is the LTS release.

.github/workflows/ci.yml
```diff
@@ -71,20 +71,12 @@ jobs:
...
             Pkg.instantiate()'
-      - run: |
-          julia --color=yes --project=docs/ -e '
-            using Flux
-            # using Pkg; Pkg.activate("docs")
-            using Documenter
-            using Documenter: doctest
-            DocMeta.setdocmeta!(Flux, :DocTestSetup, :(using Flux); recursive=true)
-            doctest(Flux)'
       - run: julia --project=docs docs/make.jl
```
Saransh-cpp, line 80: Will make it easier to look at the logs and identify potential errors.
```suggestion
      - run: julia --color=yes --project=docs docs/make.jl
```

docs/make.jl
Saransh-cpp: There will always be a tradeoff. Building docs locally will take more time now as the command will run the doctests too, but running the doctests locally will be much easier now.

I don't have a strong preference here. One can always comment out the `doctest(Flux)` line if they want to only build the docs. On second thoughts, this approach looks much better!

Saransh-cpp:
> do you know if there is a corresponding issue in Documenters.jl? I think this is worth a feature request upstream.

I think there was some discussion about this on slack, but I don't remember there being a feature request for this upstream.

CarloLucibello:
@Saransh-cpp you can hit approve if satisfied with the changes

Saransh-cpp (approved):
Thanks for this, @CarloLucibello!

(I think this might require approval from someone else too. I might not have the appropriate approver role.)

PR #2222, opened 2023-03-31
Print the state of Dropout etc.

mcabbott:
As discussed starting here https://github.com/FluxML/Flux.jl/pull/2150#issuecomment-1406875399 this prints an indication when layers like `Dropout` are set to train/test mode. But not in the default automatic state:
```julia
julia> m
Chain(
  Dense(2 => 3),                        # 9 parameters
  Dropout(0.4),
  BatchNorm(3),                         # 6 parameters, plus 6
)         # Total: 4 trainable arrays, 15 parameters,
          # plus 2 non-trainable, 6 parameters, summarysize 512 bytes.

julia> trainmode!(m)
Chain(
  Dense(2 => 3),                        # 9 parameters
  Dropout(0.4, active=true),
  BatchNorm(3, active=true),            # 6 parameters, plus 6
)         # Total: 4 trainable arrays, 15 parameters,
          # plus 2 non-trainable, 6 parameters, summarysize 513 bytes.
```
Adds the keyword printed to the constructor too. 

Needs tests.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -5,7 +5,7 @@ import Functors: Functors, @functor, functor, fmap, isleaf
 using SparseArrays: AbstractSparseArray
 
 """
-    testmode!(m, mode = true)
+    testmode!(m, inactive = true)
 
 Set a layer, or all layers in a model, to test mode.
 This disables the effect of [`Dropout`](@ref), and similar layers.
@@ -13,9 +13,9 @@ This disables the effect of [`Dropout`](@ref), and similar layers.
 _Note_: if you manually set a model into test mode, you need to manually place
 it back into train mode during training phase.
 
-Possible values of optional 2nd argument `mode` are:
+Possible values of optional 2nd argument `inactive` are:
 - `true` for testing
-- `false` for training, same as [`trainmode!`](@ref)
+- `false` for training, same as [`trainmode!`](@ref)`(m)`
 - `:auto` or `nothing` for Flux to detect training automatically.
 
 # Example
@@ -34,23 +34,23 @@ julia> trainmode!(d, :auto)  # back to default
 Dropout(0.3)
 ```
 """
-testmode!(m, mode = true) = (foreach(x -> testmode!(x, mode), trainable(m)); m)
+testmode!(m, inactive = true) = (foreach(x -> testmode!(x, inactive), trainable(m)); m)
 
 """
-    trainmode!(m, mode = true)
+    trainmode!(m, active = true)
 
 Set a layer, or all layers in a model, to training mode.
-Opposite to [`testmode!`](@ref) (i.e. `trainmode!(m, mode) == testmode!(m, !mode)`).
+Opposite to [`testmode!`](@ref) (i.e. `trainmode!(m, active) == testmode!(m, !active)`).
 
 _Note_: if you manually set a model into train mode, you need to manually place
 it into test mode during testing phase.
 
-Possible values of optional 2nd argument `mode` are:
+Possible values of optional 2nd argument `active` are:
 - `true` for training
 - `false` for testing
 - `:auto` or `nothing` for Flux to detect training automatically
 """
-trainmode!(m, mode = true) = mode isa Bool ? testmode!(m, !mode) : testmode!(m, mode)
+trainmode!(m, active = true) = active isa Bool ? testmode!(m, !active) : testmode!(m, active)
 
 function params!(p::Params, x, seen = IdSet())
   if x isa AbstractArray{<:Number} && Functors.isleaf(x)
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -49,7 +49,7 @@ trainmode!(m, ::Nothing) = testmode!(m, nothing)  # why do we have so much API?
     testmode!(model, inactive)
 
 This two-argument method is largely internal. It recurses into the `model`,
-and until a method like `testmode!(d::Dropout, mode)` alters the activity of a layer.
+and until a method like `testmode!(d::Dropout, inactive)` alters the activity of a layer.
 
 Possible values of  `inactive` are:
 - `true` for testing, i.e. `active=false`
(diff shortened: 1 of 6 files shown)


darsnack (approved):

src/functor.jl
```diff
@@ -5,36 +5,52 @@ import Functors: Functors, @functor, functor, fmap, isleaf
...
+- `:auto` or `nothing` for Flux to detect training automatically.
+
+# Example
+
+```jldoctest
+julia> d = Dropout(0.3)
+Dropout(0.3)
+
+julia> testmode!(d)   # dropout is now always disabled
+Dropout(0.3, active=false)
```
mcabbott, line 28: Are we happy with the name `active`? This existed as a field name, but not previously exposed. 

`testmode!` and `trainmode!` both had a positional argument called `mode` with opposite meanings. I made these `active` + `inactive` to match.

ToucheSir: I find the double negative a little confusing but can't come up with a better word. For a future PR, would it make sense to add a third `setactive!` function and move the true/false/auto/nothing handling logic to that? Then `trainmode!` and `testmode!` lose their second arg and become shortcuts for `setactive!(model, <true|false>)`.

mcabbott: It's awful that they both do everything. It would be OK if either accepted `:auto`, but never true/false. Maybe that's a deprecation goal? 

There could also be a 3rd function, but two is already a lot. Or the 3rd could replace both, but that's more churn.

mcabbott: But more immediately, your enum suggestion could read `Dropout(0.5; mode=:test)` and `Dropout(0.5; mode=:train)`. That has the advantage of always being one type. It's a little more indirect -- it tells you what the layer is intended for, not what it does.

ToucheSir: > It would be OK if either accepted :auto, but never true/false. Maybe that's a deprecation goal?

That's a good idea!

mcabbott: The tricky thing is that `trainmode!(m, ::Bool)` recurses to itself, and is what's overloaded by layers. Presumably some layers in packages may overload this too.

Deprecating that method and changing the recursion to use something else means that we will break any packages which rely on it.

ToucheSir: Good point. Maybe 3 functions isn't that bad after all then if it makes the deprecation path easier. PyTorch has `.train()`, `.eval()` and `module.training = {true,false}`.

src/functor.jl
```diff
@@ -5,36 +5,64 @@ import Functors: Functors, @functor, functor, fmap, isleaf
...
+and until a method like `testmode!(d::Dropout, inactive)` alters the activity of a layer.
 
-Possible values include:
-- `true` for training
-- `false` for testing
-- `:auto` or `nothing` for Flux to detect the mode automatically
+Possible values of  `inactive` are:
+- `true` for testing, i.e. `active=false`
+- `false` for training, same as [`trainmode!`](@ref)`(m)`
+- `:auto` or `nothing` for Flux to detect training automatically.
```
ToucheSir, line 57: Do we already have a note about overloading for custom layers or should that live here?

test/layers/normalisation.jl
```diff
@@ -26,6 +26,11 @@ evalwgrad(f, x...) = pullback(f, x...)[1]
     y = evalwgrad(m, x)
     @test count(a->a==0, y) > 50
 
+    # Keyword active=false
+    m2 = Dropout(0.9; active=false, rng_kwargs...)
+    y2 = evalwgrad(m2, x)
+    @test count(a->a==0, y2) == 0
```
ToucheSir, line 32:
```suggestion
    @test count(iszero, y2) == 0
```

test/layers/normalisation.jl
```diff
@@ -73,6 +78,9 @@ evalwgrad(f, x...) = pullback(f, x...)[1]
       @test cpu(m).rng === only(values(rng_kwargs))
     end
   end
+
+  @test Dropout(0.5; active=true).active === true
+  @test_throws Exception Dropout(0.5; active=:something_else)
```
ToucheSir, line 83:
```suggestion
  @test_throws ArgumentError Dropout(0.5; active=:something_else)
```
Same for below.

src/functor.jl
mcabbott: I don't think this is mentioned elsewhere. But do we like it enough to keep it? The whole thing feels very sloppy, it accepts 4 different types & may or may not give errors from random internal places...

test/layers/normalisation.jl
mcabbott: I think I went back on this, and just constrained the types accepted for the keyword to Union{Bool,Nothing}, with only the Bool case public. So it's no longer an ArgumentError.

src/functor.jl
```diff
@@ -5,36 +5,64 @@ import Functors: Functors, @functor, functor, fmap, isleaf
...
+julia> d = Dropout(0.3)
+Dropout(0.3)
+
+julia> testmode!(d)   # dropout is now always disabled
+Dropout(0.3, active=false)
+
+julia> trainmode!(d)  # dropout is now always enabled
+Dropout(0.3, active=true)
+
+julia> trainmode!(d, :auto)  # back to default
```
mcabbott, line 32:
```suggestion
julia> testmode!(d, :auto)  # back to default
```

src/functor.jl
ToucheSir: I think I chose a poor place to attach this comment. The question was how we document how users can overload this behaviour for their own layers. i.e. "overload `testmode!(l::MyLayer, [mode])` if you want to have `trainmode!` and `testmode!` work for `MyLayer`.

darsnack:
We discussed this PR on call. Here is our review:
- Add a doc section [here](https://fluxml.ai/Flux.jl/stable/models/advanced/) explaining overloading `testmode!` for stochastic layers
- Add forward pointers to doc section from docstring
- Some discussion of future steps:
    - Have a single `setactive!(m, mode)` and just `trainmode!(m)` / `testmode!(m)`
    - This would be breaking so post-PR

mcabbott:
OK last commit adds a note that this is the layer API but it may change later. It does not edit the "advanced" docs page, as that's a bit of a mess and this PR has many things already. This is not a new feature here, it's just that revising it brought attention.

PR #2226, opened 2023-04-11
Float32 warning bug fix

ashwani-rathee:
- [x] For removing Float32 warning in readme example

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Flux is an elegant approach to machine learning. It's a 100% pure-Julia stack, a
 Works best with [Julia 1.8](https://julialang.org/downloads/) or later. Here's a very short example to try it out:
 ```julia
 using Flux, Plots
-data = [([x], 2x-x^3) for x in -2:0.1f0:2]
+data = [([Float32(x)], Float32(2x-x^3)) for x in -2:0.1f0:2]
 
 model = Chain(Dense(1 => 23, tanh), Dense(23 => 1, bias=false), only)
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ for epoch in 1:1000
 end
 
 plot(x -> 2x-x^3, -2, 2, legend=false)
-scatter!(-2:0.1f0:2, [model([x]) for x in -2:0.1f0:2])
+scatter!(-2:0.1:2, [model([x]) for x in -2:0.1f0:2])
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,7 +21,7 @@ Flux is an elegant approach to machine learning. It's a 100% pure-Julia stack, a
 Works best with [Julia 1.8](https://julialang.org/downloads/) or later. Here's a very short example to try it out:
 ```julia
 using Flux, Plots
-data = [([x], 2x-x^3) for x in -2:0.1f0:2]
+data = [([Float32(x)], Float32(2x-x^3)) for x in -2:0.1f0:2]
```
ToucheSir, line 24: I don't think this is the right line or the right change:
```julia-repl
julia> [([x], 2x-x^3) for x in -2:0.1f0:2] |> summary
"41-element Vector{Tuple{Vector{Float32}, Float32}}"
```

ToucheSir: More likely the problem is at https://github.com/FluxML/Flux.jl/blob/master/README.md?plain=1#L34, which does not use f32 literals.

ToucheSir: `-2:0.1f0:2` note the `f0`

mcabbott: Needs to be the one in the comprehension though -- it's the x being fed to the model which matters, not the plotting.

ToucheSir: Yes, I should've been more specific. The precision of the first arg to `scatter` doesn't matter here and can stay as the default Float64.

ashwani-rathee: Still something missing in this?

ToucheSir:
```
scatter!(-2:0.1:2, [model([x]) for x in -2:0.1f0:2])
```

ToucheSir: Sure, you could do `scatter(-2:0.1f0:2, ...)` as well. The only change necessary to avoid the warning is to ensure the _second_ range is Float32.

README.md
```diff
@@ -31,7 +31,7 @@ for epoch in 1:1000
 end
 
 plot(x -> 2x-x^3, -2, 2, legend=false)
-scatter!(-2:0.1f0:2, [model([x]) for x in -2:0.1f0:2])
+scatter!(-2:0.1:2, [model([x]) for x in -2:0.1f0:2])
```
mcabbott, line 34: Another option here would be this:
```suggestion
scatter!(x -> model([x]), -2:0.1f0:2)
```
This avoids having to repeat the `-2:0.1:2` bit. The line above also defines such a function.

mcabbott (approved):

Issue #2214, opened 2023-03-17
buildkite failure on julia v1.6


CarloLucibello:
Lately, the buildkite gpu tests on julia v1.6 are failing, see
https://buildkite.com/julialang/flux-dot-jl/builds/3755#0186f06e-ec5d-4ddc-a1ea-3662edd19d53
as an example. 
The error message is super long, truncated, and show a lot of `CartesianIndex`. It's hard to understand why this happens.

ToucheSir:
I think that was a one-off spurious failiure, because I don't see it in any previous runs or on master. If the latest CI run completes successfully, we can probably close this.

ToucheSir:
Ok, it happened again on a different PR: https://buildkite.com/julialang/flux-dot-jl/builds/3764#0186f1c1-c5df-4ebc-9a5d-9f0c24ee4969. I'm guessing something changed on the CUDA.jl side because we shouldn't have done anything on the FluxML side?

christiangnrd:
This might be caused by GPUArrays 8.6.4 that was released yesterday.

CarloLucibello:
now also v1.8 is failing

CarloLucibello:
Failure of the day is
https://buildkite.com/julialang/flux-dot-jl/builds/3802#018728a6-e273-41ed-b265-322836527ac7

CarloLucibello:
The problem is a CUDNN failure in a Float16 test on julia v1.6

ToucheSir:
Strange that we'd suddenly get `CUDNN_STATUS_BAD_PARAM` and only on 1.6...

PR #2234, opened 2023-04-21
Remove type restrictions for recurrent cells

darsnack:
The type restrictions on the recurrent cell input only existed to match the input and weight types. Now, we have `_match_eltype`, so we don't need such strong restrictions. This simplifies the type signature so that weird types like `Nil` can also pass through (and `_match_eltype` already support handling these).

A nice benefit is that we don't need to have a special `outputsize` rule for recurrent cells anymore.

### PR Checklist

- [x] ~~Tests are added~~ (already exist)
- [x] ~~Entry in NEWS.md~~

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -200,7 +200,7 @@ end
 RNNCell((in, out)::Pair, σ=tanh; init=Flux.glorot_uniform, initb=zeros32, init_state=zeros32) =
   RNNCell(σ, init(out, in), init(out, out), initb(out), init_state(out,1))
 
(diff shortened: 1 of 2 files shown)


ToucheSir (approved):
Buildkite failure is https://github.com/FluxML/Flux.jl/issues/2214#issuecomment-1507999752 and known, nightly failure is the usual Zygote breaking every new Julia release.

darsnack:
I merging since the Metalhead failure is known and downstream. What's with AtomicGraphNets? Why does it take so long to run the tests?

ToucheSir:
I'm not sure. It's been on and off stable over the years, so if CI on master there isn't passing (as is the case now) I wouldn't count it as a blocker.

PR #2235, opened 2023-04-21
Fix Conv transfer to AMDGPU

pxl-th:
Previously only this flipped conv weights:
```julia
Conv((3, 3), 3 => 3) |> gpu
```
but not this (or anything more complex):
```julia
Chain(Conv((3, 3), 3 => 3)) |> gpu
```

To fix this, I modified `exclude` function passed to `fmap` both for `CPU -> GPU` & `GPU -> CPU` transfer (so that weights are flipped back).

- Update AMDGPU compat to latest to include MIOpen synchronization fix: https://github.com/JuliaGPU/AMDGPU.jl/pull/415

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/ext/AMDGPUExt/functor.jl
+++ b/ext/AMDGPUExt/functor.jl
@@ -31,19 +31,19 @@ end
 # Same for CPU -> GPU & GPU -> CPU movements.
 # Note, that gradients are also flipped.
 
-function Flux._isleaf(::X) where {
-    X <: Union{Flux.Conv{N, M, F, A, V}, Flux.ConvTranspose{N, M, F, A, V}}
-} where {N, M, F, A <: ROCArray, V}
-    return true
-end
+const FLUX_CONV{M} = Union{
+    Flux.Conv{<:Any, <:Any, <:Any, <:M, <:Any},
+    Flux.ConvTranspose{<:Any, <:Any, <:Any, <:M, <:Any}}
+const CPU_CONV = FLUX_CONV{Array}
+const AMD_CONV = FLUX_CONV{ROCArray}
+
+_conv_basetype(c::C) where C <: Conv = Conv
+_conv_basetype(c::C) where C <: ConvTranspose = ConvTranspose
+
+Flux._isleaf(::AMD_CONV) = return true
 
 _exclude(x) = _isleaf(x)
-
-function _exclude(::X) where {
-    X <: Union{Conv{N, M, F, A, V}, ConvTranspose{N, M, F, A, V}}
-} where {N, M, F, A <: Array, V}
-    true
-end
+_exclude(::CPU_CONV) = true
 
 function _amd(x)
     check_use_amdgpu()
@@ -53,12 +53,9 @@ end
 
 # CPU -> GPU
 
-_conv_basetype(c::Type{C}) where C <: Conv = Conv
-_conv_basetype(c::Type{C}) where C <: ConvTranspose = ConvTranspose
-
-function Adapt.adapt_structure(to::FluxAMDAdaptor, m::C) where C <: Union{Flux.Conv, Flux.ConvTranspose}
+function Adapt.adapt_structure(to::FluxAMDAdaptor, m::CPU_CONV)
     flipped_weight = reverse(m.weight; dims=ntuple(i -> i, ndims(m.weight) - 2))
-    _conv_basetype(C)(
+    _conv_basetype(m)(
         Adapt.adapt(to, m.σ),
         Adapt.adapt(to, flipped_weight),
         Adapt.adapt(to, m.bias),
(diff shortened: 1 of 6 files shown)


src/functor.jl
```diff
@@ -352,3 +325,42 @@ function gpu(::FluxAMDAdaptor, x)
...
+end
+
+function _cpu_exclude()
+    !(GPU_BACKEND == "AMD" && AMDGPU_LOADED[]) && return _isleaf
+
+    function is_amd_conv(x)
+        is_conv = typeof(x) <: Union{Conv, ConvTranspose}
+        (is_conv && (typeof(x.weight) <: CUDA.AbstractGPUArray)) || _isleaf(x)
+    end
+    return is_amd_conv
```
ToucheSir, line 365: Instead of adding this conditional, we should overload `_isleaf` for `Conv` on AMDGPU.

pxl-th: Done!

ext/AMDGPUExt/functor.jl
```diff
@@ -26,23 +26,37 @@ function ChainRulesCore.rrule(
...
-    fmap(x -> Adapt.adapt(FluxAMDAdaptor(), x), x; exclude=_isleaf)
-end
-
 # Since MIOpen supports only cross-correlation as convolution,
 # for the actual convolution, we flip horizontally and vertically the weights.
 # Same for CPU -> GPU & GPU -> CPU movements.
 # Note, that gradients are also flipped.
 
+function Flux._isleaf(::X) where {
+    X <: Union{Flux.Conv{N, M, F, A, V}, Flux.ConvTranspose{N, M, F, A, V}}
```
ToucheSir, line 35: One more thing now that the diff is a little clearer for me. We should consider defining an alias for `Union{Flux.Conv{<:Any, <:Any, <:Any, <:ROCArray, <:Any}, Flux.ConvTranspose{<:Any, <:Any, <:Any, <:ROCArray, <:Any}}` to use in all of these functions. There's no need to specialize on the other type parameters, and not doing so could save us some latency.

ext/AMDGPUExt/functor.jl
```diff
@@ -52,31 +66,30 @@ function adapt_storage(to::FluxAMDAdaptor, m::C) where C <: Union{Conv, ConvTran
 end
 
 # Don't adapt again.
-function adapt_storage(
-    to::FluxAMDAdaptor, m::Conv{N, M, F, A, V},
-) where {N, M, F, A <: ROCArray, V}
+
+function Adapt.adapt_structure(to::FluxAMDAdaptor, m::Conv{N, M, F, A, V}) where {N, M, F, A <: ROCArray, V}
```
ToucheSir, line 70: Having an alias would allow us to dedup these pairs of methods from lines 70-108. I want to say the `_amd` overload for Conv(Transpose) is also no longer required with the `_isleaf` overload, but worth a check.

pxl-th: Thanks! Indeed, `_amd` for `Conv` is no longer required.

ext/AMDGPUExt/functor.jl
```diff
@@ -26,70 +26,51 @@ function ChainRulesCore.rrule(
...
+    Flux.ConvTranspose{<:Any, <:Any, <:Any, <:M, <:Any}}
+const CPU_CONV = FLUX_CONV{Array}
+const AMD_CONV = FLUX_CONV{ROCArray}
+
+_conv_basetype(c::C) where C <: Conv = Conv
+_conv_basetype(c::C) where C <: ConvTranspose = ConvTranspose
 
-_conv_basetype(c::Type{C}) where C <: Conv = Conv
-_conv_basetype(c::Type{C}) where C <: ConvTranspose = ConvTranspose
+Flux._isleaf(::AMD_CONV) = return true
```
ToucheSir, line 43:
```suggestion
Flux._isleaf(::AMD_CONV) = true
```

ext/AMDGPUExt/functor.jl
```diff
@@ -26,70 +26,51 @@ function ChainRulesCore.rrule(
...
-end
-
-function adapt_storage(
-    to::FluxAMDAdaptor, m::ConvTranspose{N, M, F, A, V},
-) where {N, M, F, A <: ROCArray, V}
-    return m
-end
 
-_amd(m::Union{Conv, ConvTranspose}) = adapt_storage(FluxAMDAdaptor(), m)
+Adapt.adapt_structure(to::FluxAMDAdaptor, m::AMD_CONV) = return m
```
ToucheSir, line 67:
```suggestion
Adapt.adapt_structure(to::FluxAMDAdaptor, m::AMD_CONV) = m
```

ext/AMDGPUExt/functor.jl
```diff
@@ -26,70 +26,51 @@ function ChainRulesCore.rrule(
...
 
-# CPU -> GPU
+const FLUX_CONV{M} = Union{
+    Flux.Conv{<:Any, <:Any, <:Any, <:M, <:Any},
+    Flux.ConvTranspose{<:Any, <:Any, <:Any, <:M, <:Any}}
+const CPU_CONV = FLUX_CONV{Array}
+const AMD_CONV = FLUX_CONV{ROCArray}
+
+_conv_basetype(c::C) where C <: Conv = Conv
+_conv_basetype(c::C) where C <: ConvTranspose = ConvTranspose
```
ToucheSir, line 41:
```suggestion
_conv_basetype(::Conv) = Conv
_conv_basetype(::ConvTranspose) = ConvTranspose
```

ToucheSir (approved):
A couple of small touch-ups, but otherwise LGTM

pxl-th:
Done! Once this is merged, can we also tag a release?
I'd like to specify a compat bound for one of the packages.

ToucheSir:
I want to see if the Julia v1 CUDA failures on this PR propagate to the main branch. They shouldn't be caused by this test suite because it never runs on CI (we should add that), so it'd be good to figure out what's going on before tagging.

pxl-th:
The one with Float16 or the one at the very end?

ToucheSir:
The Float16 one. I'm not sure why it's seemingly spread from just the 1.6 job to the 1.x one as well...

pxl-th:
Right before error there is this Info...
```julia
[ Info: Don't mind the following error, it's for testing purpose.
[ Info: Testing GPU Support
Misc. Float16: Error During Test at /var/lib/buildkite-agent/builds/gpuci-10/julialang/flux-dot-jl/test/cuda/layers.jl:332
```

pxl-th:
> You mean running the CUDA tests locally works?

Yes. And now it always succeeds. Can't reproduce the error for some reason...

ToucheSir:
I was able to repro consistently yesterday on 1.6 and 1.8. Will try to look into it over the next couple of days.

PR #2239, opened 2023-04-25
add Flux.state(x)

CarloLucibello:
This implements the idea expressed multiple times that we should have something like `fmapstructure` for grabbing the internal state. The aim is to provide a less fragile solution than bson for serialization. 

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


ToucheSir:
Related discussion: https://github.com/FluxML/Functors.jl/pull/56.

test/utils.jl
```diff
@@ -613,6 +613,23 @@ end
       @test ∇p ≈ destructure(∇m)[1]
     end
   end
+
+  @testset "state" begin
+    m1 = Chain(Dense(10, 5), Parallel(+, Dense(Flux.ones32(2, 5), false), Dense(5 => 2)))
```
mcabbott, line 618: I think this saves Dropout activity state, and BatchNorm's momenta, which would be worth testing. These aren't trainable parameters but are state.

src/loading.jl
```diff
@@ -101,3 +101,32 @@ function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
...
+    else
+      return x isa Union{Number, AbstractArray} ? x : nothing
+    end
+  else
+    return valuemap(c -> state(c; full), Functors.children(x))
+  end
+end
+
+valuemap(f, x) = map(f, x)
+valuemap(f, x::Dict) = Dict(k => f(v) for (k, v) in x)
```
mcabbott, line 132: Perhaps `_valuemap` to discourage wider use?

src/loading.jl
```diff
@@ -101,3 +101,32 @@ function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
...
+basic containers (e.g. named tuples, tuples, arrays, and dictionaries).
+
+If `full` is `false` (default), then only arrays and scalar original leaves are used as leaf values in the return, 
+with the other leaves being replaced by `nothing`.
+
+This method is particularly useful for saving and loading models, since it doesn't
+require the user to specify the model type.
+The returned state, can be passed to `loadmodel!` to restore the model.
+"""
+function state(x; full=false)
```
mcabbott, line 119:
```suggestion
function state(x; full::Bool=false)
```

docs/src/models/saving.md
```diff
@@ -0,0 +1,6 @@
+# Saving and Loading
+
+```julia
+Flux.loadparams!
```
mcabbott, line 4: This is the deprecated one, no?

docs/make.jl
```diff
@@ -35,6 +35,7 @@ makedocs(
             "Gradients -- Zygote.jl" => "training/zygote.md",
             "Batching Data -- MLUtils.jl" => "data/mlutils.md",
             "OneHotArrays.jl" => "data/onehot.md",
+            "Saving and Loading" => "models/saving.md"
```
mcabbott, line 38: Should this be its own page for two commands? They could live with "Flat vs. Nested" which also has `modules`.

mcabbott (commented):
Unlike fmapstructure, this recursion doesn't know about shared arrays. Does this matter?

If it doesn't, or we don't want it, that's an argument for it being here not in Functors.jl

Bikeshedding the name, is `state` specific enough? We have `opt_state` with which this should never be confused. Since it pairs with `loadmodel!(m, x)`, should its name echo that... like `modelstate(m)`? Or "bare" or "Base" or "tree"? (Honestly a tree if it ignores identity.)

CarloLucibello:
> Unlike fmapstructure, this recursion doesn't know about shared arrays. Does this matter?

it doesn't need to since it doesn't transform (except for returning nothing).
If needed, we can provide in the future serializers and desarializers handling shared arrays.

> Bikeshedding the name, is state specific enough? We have opt_state with which this should never be confused. Since it pairs with loadmodel!(m, x), should its name echo that... like modelstate(m)? Or "bare" or "Base" or "tree"? (Honestly a tree if it ignores identity.)

This is the common pattern that I expect for this method:
```julia
model = ...
opt_state = Optimisers.setup(Adam(), model)
model_state = Flux.state(model)
BSON.@save "checkpoint.bson" opt_state model_state
```
It reads nice and clear as it, I think less nice with `modelstate`, `bare`, `Base` or `tree`.
I would actually rename `loadmodel!` to `loadstate!`.

CarloLucibello:
What should we `keep` by default in the state? Since the main purpose is serialization, at the very least we should exclude anonymous functions. This way, we could recommend JLD2.jl as the main serialization solution, since it is more robust and maintained compared to BSON.jl  (cc @JonasIsensee)

test/loading.jl
```diff
@@ -0,0 +1,227 @@
...
+@testset "loadmodel!(dst, src) with BSON" begin
+  m1 = Chain(Dense(Float32[1 2; 3 4; 5 6], Float32[7, 8, 9]), Dense(3 => 1))
+  m2 = Chain(Dense(Float32[0 0; 0 0; 0 0], Float32[0, 0, 0]), Dense(3 => 1))
+  @test m1[1].weight != m2[1].weight
+  mktempdir() do dir
+    BSON.@save joinpath(dir, "test.bson") m1
+    m2 = Flux.loadmodel!(m2, BSON.load(joinpath(dir, "test.bson"))[:m1])
+    @test m1[1].weight == m2[1].weight
+  end
+end
```
CarloLucibello, line 191: everything above in this file is just moved from `test/utils.jl`. Sorry for diff noise but I would like the `test/` structure to match the `src/` as close as possible.

test/utils.jl
```diff
@@ -420,182 +420,6 @@ end
...
-    m1 = Chain(Dense(Float32[1 2; 3 4; 5 6], Float32[7, 8, 9]), Dense(3 => 1))
-    m2 = Chain(Dense(Float32[0 0; 0 0; 0 0], Float32[0, 0, 0]), Dense(3 => 1))
-    @test m1[1].weight != m2[1].weight
-    mktempdir() do dir
-      BSON.@save joinpath(dir, "test.bson") m1
-      m2 = Flux.loadmodel!(m2, BSON.load(joinpath(dir, "test.bson"))[:m1])
-      @test m1[1].weight == m2[1].weight
-    end
-  end
-
```
CarloLucibello, line 598: everything removed from this file is now (unchanged) in `test/loading.jl`

theabhirath:
Not sure if this PR is completely ready yet, but I just tried to save weights using `Flux.state` (using BSON.jl, admittedly) and `loadmodel!` blows up with both memory and time usage. For only a ResNet-18 model, I get close to 3.5 GB of memory being used and about 2 whole minutes for the model to load. https://github.com/FluxML/Metalhead.jl/pull/225#issuecomment-1521096235 leads me to believe this might be something that needs to be addressed from the `loadmodel!` side of things? Fixing this sort of storing and loading of weights without storing anything more than that needs to be stored (read: RNGs etc. when we store the entire Flux model as is the docs' convention now) is one of the major things that can improve pretrained model support in Metalhead, so I am very invested 😅

CarloLucibello:
Thanks for the investigation! Yes, the problem surely is with `loadmodel!` although I don't know the cause. So that shouldn't block this PR, but needs to be investigated.

test/loading.jl
darsnack: Yeah that's an artifact from when loading was in utils.jl. Should have been switched earlier.

darsnack:
The current implementation is equivalent to `fmapstructure(identity, m; exclude = keep, cache = nothing)` (`keep` might need to change slightly). My point is we should not write yet another `fmap`-like loop somewhere in the ecosystem. I worry about maintaining all these different loops when decisions about `trainable`, `children`, walks, etc. are made. So unless it warrants a custom loop like `loadmodel!` or Optimisers.jl, then we should try to implement it with Functors.

Re: what's kept. `loadmodel!` presently ignores functions, RNGs, etc. So those are obvious candidates not to save. It does understand non-trainable state. I believe it assume the `active` field of the destination model.

darsnack:
> One thing I can picture wanting to configure is this: Instead of getting all state, get only the trainable parameters, as a similar nested structure. That cannot be controlled by keep, but would instead want children to be replaced by trainable. No idea whether loadmodel! will at present understand such a truncated tree, but it could be made to.

The current solution in `loadmodel!` is the `filter` keyword to ignore the truncated sub-tree, then manually handle it if necessary. Perhaps a simple filter for the use-case you describe is if some sentinel value is written at that node in the stored tree.

CarloLucibello:
So what I'm proposing is that this whole PR becomes
```julia
state_arrays(x) = fmapstructure(x -> x isa AbstractArray ? x : nothing, x)
```

darsnack:
Given what `loadmodel!` understands, I think your proposal is the correct thing to do right now. Minor change might be using `Optimisers.numeric` instead of checking just for arrays.

ToucheSir:
Skimming through some prior art and possible inspiration:
- PyTorch has [`named_parameters()`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.named_parameters), which flattens the model tree but implicitly tracks structure in the names. Objax does something similiar with [`vars()`](https://objax.readthedocs.io/en/latest/objax/objax.html#objax.Module).
- TF has an interesting combination of [`get_config()`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#get_config) and [`get_weights()`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer#get_weights). The names are so-so, but I like the idea of having a way of accessing both even if we usually only save the latter.
- The well-known JAX-based frameworks like Flax or Haiku already store parameters outside the callable model structure, but they generally use `params` or `variables` to refer to them ([e.g.](https://flax.readthedocs.io/en/latest/api_reference/flax.training.html#flax.training.train_state.TrainState)).

When it comes to serialization, I think our approach would be closest to how JAX encourages serializing PyTrees. A 1-1 match would require keeping around non-arrays (while excluding stuff like RNGs) and adding metadata tags for what type each node in the tree was created from. For now though, we can get away with far less metadata.

CarloLucibello:
I think in pytorch the closest thing would be [.dict_state()](https://pytorch.org/tutorials/recipes/recipes/what_is_state_dict.html)

For our state I would prefer an opt-in approach for the leaf types that get into the state. We can make the contract that future (non-breaking) releases  can modify the returned state by allowing for more types to get in, but never less. 

The implementation could be
```julia
state(x) = fmapstructure(x -> _keep(x) ? x : missing, x)

_keep(x::TOKEEP) = true
_keep(x) = false
```
Then it is not clear where we should set the bar. These are possible candidates
1. `TOKEEP = AbstractArray`
2. `TOKEEP = Union{AbstractArray, Number}`
3. `TOKEEP = Union{AbstractArray, Number, Nothing, AbstractString, Symbol}`

In the future we can expose some interface for customization.

CarloLucibello:
In the end I went with `TOKEEP = Union{AbstractArray, Number, Nothing, AbstractString, Symbol}`, I don't see any harm in saving a few more simple types. 

I also rewrote the "Saving and Loading" docs to encourage the serialization of the state and discourage the serialization of the model struct.

ToucheSir:
I'm generally in favour of this now. Thoughts on JLD2 vs the Serialization stdlib? I know the former sometimes requires maintenance for new Julia versions, but the latter has its own issues.

JonasIsensee:
> 

Maybe this is a good moment for me to chime in. (also since I've been tagged the second time ;) )

From what I can tell, you've decided to largely restrict yourself to storing "plain" data such as `String`s, Floats, Ints, and Arrays of such. In that case, JLD2 output matches regular HDF5 output. (Great, because you can use the inspection tools and load everything independently of julia)

Now, concerning `Nothing`, `Symbol`, `Missing`, etc. (and do you really have/need `<:AbstractString` that are not `String` ?).
These do not exist a priori in HDF5 but JLD2 will happily encode them. Given the stage of Julia development, I don't think these will ever break either. (except for the strings?)

Lastly, I would very much recommend building your own `serializetojld2` function in order to unroll the different named fields
(citing from above)
```
julia> s = Flux.state(Dropout(0.2))
(p = 0.2, dims = nothing, active = nothing, rng = Random.TaskLocalRNG())
```
into separate jld2 datasets e.g.: 
```
function writestatetojld2(fname, nt)
     jldopen(fname) do f
         wsession = JLD2.JLDWriteSession() # objid tracker for identity preservation
        for (k,v) in pairs(nt)
            write(f, string(k), v, wsession)
        end
     end
end
```
and an analog for loading.
While this is a bit more work on your side, it allows for much more graceful / gradual failing when some datasets (e.g. RNG states) fail to reconstruct.

src/loading.jl
```diff
@@ -104,3 +106,65 @@ function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
...
+
+julia> m2 = Chain(Dense(1, 2, tanh), Dense(2, 1));
+
+julia> s = Flux.state(m1)
+layers = ((weight = Float32[-0.56867087; 1.229064;;], bias = Float32[0.0, 0.0], σ = nothing), (weight = Float32[0.23323897 -0.5561147], bias = Float32[0.0], σ = nothing)),)
+
+julia> Flux.loadmodel!(m2, s);
+
+julia> m2[1].weight == m1[1].weight
+true
```
mcabbott, line 143: IMO it would be worth showing some more complicated layers. Perhaps separately after the simple one is best. 
```suggestion
julia> m2[1].weight
2×1 Matrix{Float32}:
 1.0
 1.0

julia> Flux.state(trainmode!(Dropout(0.2)))  # stores p & activity, but not RNG state
(p = 0.2, dims = missing, active = true, rng = missing)

julia> Flux.state(BatchNorm(1))  # stores non-trainable arrays μ, σ²
(λ = missing, β = Float32[0.0], γ = Float32[1.0], μ = Float32[0.0], σ² = Float32[1.0], ϵ = 1.0f-5, momentum = 0.1f0, affine = true, track_stats = true, active = nothing, chs = 1)
```

mcabbott (commented):
Some minor comments below, but I think this basically looks great.

> (and do you really have/need `<:AbstractString` that are not `String` ?)

This would be super-easy to standardise on String to be safe.

We could surely also convert all Symbols to String, and have `loadmodel!` convert back when it sees that the original model has a Symbol there.

> `Nothing`, `Symbol`, `Missing`, etc.

This PR uses `missing` for unsaved fields. I think another obvious candidate would be `()`. Quick tests are that `loadmodel!` understands this fine. (This is what `Optimisers.setup` uses.)

At present `nothing` is meaningful as the default automatic test/train state of `Dropout`, etc. We could consider changing them all to something like `mode = :auto / :train / :test`. Converting so say a string `"nothing"` might be tricky for `loadmodel!` to reverse, as these mutable structs take Union{Nothing,Bool}, but perhaps someone else does `Union{Nothing,String}` or something.

ToucheSir:
I don't think the plan is to depend on any serialization libraries in Flux itself. At best this would be handled by extensions or left to users.

docs/src/saving.md
```diff
@@ -98,43 +56,91 @@ In longer training runs it's a good idea to periodically save your model, so tha
...
 Chain(
   Dense(10 => 5, relu),                 # 55 parameters
   Dense(5 => 2),                        # 12 parameters
-  NNlib.softmax,
 )                   # Total: 4 arrays, 67 parameters, 524 bytes.
 
 julia> evalcb = throttle(30) do
-         # Show loss
-         @save "model-checkpoint.bson" model
+          jldsave("model-checkpoint.jld2", model_state = Flux.state(m))
```
mcabbott, line 68: This doctest fails to load JLD2

Note also that it's written for old-style `train!` callbacks. Not urgent but would be worth revising to remove that.

src/loading.jl
```diff
@@ -104,3 +106,65 @@ function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
...
+
+julia> Flux.loadmodel!(m2, JLD2.load("checkpoint.jld2", "model_state"))
+```
+"""
+state(x) = Functors.fmapstructure(x -> _state_keep(x) ? x : missing, x)
+
+const STATE_TYPES = Union{AbstractArray, Number, Nothing, AbstractString, Symbol}
+
+_state_keep(x::STATE_TYPES) = true
+_state_keep(x) = false
```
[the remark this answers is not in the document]
CarloLucibello: > _state(x::AbstractString) = String(x)

I would avoid potentially expensive conversions for the time being. In case that should be done at serialization time.

NEWS.md
```diff
@@ -1,18 +1,20 @@
 # Flux Release Notes
 
+<<<<<<< HEAD
```
darsnack, line 3: Missed git stuff

src/loading.jl
```diff
@@ -104,3 +106,71 @@ function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
...
+
+julia> Flux.loadmodel!(m2, JLD2.load("checkpoint.jld2", "model_state"))
+```
+"""
+state(x) = Functors.fmapstructure(_state, x)
+
+const STATE_TYPES = Union{AbstractArray, Number, Nothing, AbstractString, Symbol}
+
+_state(x::STATE_TYPES) = x
+_state(x) = missing
```
darsnack, line 176: Does it make more sense for this to be `(;)` like @mcabbott suggested? Our sentinel value leaks out to the external API because it affects what `state` returns. If someone uses this function for anything other than directly saving/loading (not entirely uncommon), then a custom model with `missing` fields will have a confusing looking `state`. I think `(;)` is clearer and also handled automatically by other Functors function like `loadmodel!`.

darsnack (commented):
It looks good to go, but one minor point worth discussing since changing it would require a breaking release. I think this got brought up previously but went undiscussed.

src/loading.jl
mcabbott: `()` is even simpler than `(;)`. The one case I tried worked fine in `loadmodel!`.

CarloLucibello: annoys me a little bit that `()` could mean either `()` or something not part of the state but in practice it doesn't matter. 

Another option, circumventing the problem, is to just prune those dangling nodes.  I'll give it a try, seems the cleanest solution.

CarloLucibello: So we can apply 
```julia
prune_missing(x) = x

prune_missing(nt::NamedTuple) =
  (; (k => prune_missing(v) for (k,v) in pairs(nt) if !ismissing(v))...)

prune_missing(d::Dict) =
  Dict(k => prune_missing(v) for (k,v) in pairs(d) if !ismissing(v))
```
before returning the state. We still need a sentinel value for tuples though, just shrinking the couple is too ambiguous. 
Are we ok with this mixed situation? I think I am, the state would look much cleaner. 
If yes, is it ok at this point for the rarely showing sentinel to be `missing` or should i change it to `()`?

CarloLucibello: In the end I also replaced the remaining missings with `()`.  `missing` conveys the feeling that "something is lost" which is not appropriate here

CarloLucibello:
Should be ready for merge

darsnack (approved):
Just a couple more tests, but looks ready

mcabbott:
I'm a little unsure of this prune-the-tree idea. I agree it makes for a tidier state to look at. 

But it adds complexity for anything consuming it: Instead of knowing that field names will always match, and some will be `::Tuple{}`, now you need to also handle the field names being a subset. You can't use `map(f, ::NamedTuple, ::NamedTuple)`, and I think you can't use `fmap(f, model, state)`. Yet you are not free from knowing about a special token for empty spaces at all, since this is still needed for when `x::Tuple`.

And, in addition, accidentally using the wrong saved state is less likely to give you an informative error.

In the AD world, Zygote always makes a NamedTuple for all fields. ChainRules allows some to be missing via Tangent, and IMO this is a mistake, it adds complexity without much gain. (Maybe for Dict it makes sense to allow missing entries, as they aren't free.)

darsnack:
You are right, I had thought a missing key would be cleaner and the same as not recursing a node, but it is not. We have `Functors.NoChildren` which is exactly `()`. It may be ugly to use `NoChildren` for a missing leaf, but it is more robust than deleting the leaf.

darsnack (dismissed):

CarloLucibello:
I reverted the change, so state is not dropping fields anymore, and `()` is for non-state leaves. 

I preserved the change I made to `loadmodel!` though, I think it is good for it to be able to accept a subtree as a source.
If needed, we can add an option `strict=true` to it forcing all fields to be present in the source. Future PR material possibly.

darsnack:
Why not just put that change in the future PR that also adds `strict`? I can see both sides for accepting subtrees and not. The current `loadmodel!` forces you to use `filter` (albeit undocumented) to handle the subtree case. For saving/loading, I think it is better to be overly strict with tooling for handling edges. The `filter` keyword is a weak implementation of [Orbax's transforms](https://github.com/google/orbax/blob/main/docs/checkpoint.md#transformations). Orbax might be guidance for us, because Jax/Flax most closely matches our approach to layers.

CarloLucibello:
I'm not using `filter`, `filter` could be removed in fact. I'm simply iterating over the source children

CarloLucibello:
merge like this?

ToucheSir:
Isn't `filter` still needed for Metalhead? I confess I haven't followed the implementation here very closely.

darsnack:
I am not saying your change is using `filter`. I am saying `filter` is an existing mechanism for handling a partial copy of a subtree by skipping it and doing it manually. All I'm suggesting is that the change to `loadmodel!` is no longer required to make this PR work, and that we should do it in another PR where there is time to discuss it. Isn't it breaking anyways? Something that previously intentionally errored does not error anymore.

CarloLucibello:
restored previous behaviour for `loadmodel!`

mcabbott (commented):
Late but...

PR #1980, opened 2022-05-30
Added Common GPU Workflows in Docs

lfenzo:
Following the discussion in #1974, this is a very rough draft of the "Common GPU Workflows", feel free to suggest any changes or additions to this new section in the docs.

I also took the liberty of reformatting the `docs/make.jl` to make the pages and titles a bit easier to read.

PR #2240, opened 2023-04-25
gpu(::DataLoader), take II

mcabbott:
Closes https://github.com/FluxML/Flux.jl/pull/2186 , by replacing CuIterator with much the same thing, but owned by Flux.

The advantage of this is that we can control exactly what it does, to match `gpu(x)`. This means converting to Float32, and also working with AMDGPU arrays (and whatever we add next).

Needs tests.

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -136,18 +136,32 @@ Some of the common workflows involving the use of GPUs are presented below.
 
 In order to train the model using the GPU both model and the training data have to be transferred to GPU memory. This process can be done with the `gpu` function in two different ways:
 
-1. Iterating over the batches in a [DataLoader](@ref) object transferring each one of the training batches at a time to the GPU. 
+1. Iterating over the batches in a [DataLoader](@ref) object transferring each one of the training batches at a time to the GPU. Done hand, this might look like this:
    ```julia
    train_loader = Flux.DataLoader((xtrain, ytrain), batchsize = 64, shuffle = true)
-   # ... model, optimiser and loss definitions
-   for epoch in 1:nepochs
+   # ... model definition, optimiser setup
+   for epoch in 1:epochs
        for (xtrain_batch, ytrain_batch) in train_loader
-           x, y = gpu(xtrain_batch), gpu(ytrain_batch)
-           gradients = gradient(() -> loss(x, y), parameters)
-           Flux.Optimise.update!(optimiser, parameters, gradients)
+           x = gpu(xtrain_batch)
+           y = gpu(ytrain_batch)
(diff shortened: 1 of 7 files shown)


docs/src/gpu.md
```diff
@@ -155,26 +170,14 @@ In order to train the model using the GPU both model and the training data have
...
    gpu_train_loader = Flux.DataLoader((xtrain, ytrain) |> gpu, batchsize = 32)
    ```
-   Note that both `gpu` and `cpu` are smart enough to recurse through tuples and namedtuples. Another possibility is to use [`MLUtils.mapsobs`](https://juliaml.github.io/MLUtils.jl/dev/api/#MLUtils.mapobs) to push the data movement invocation into the background thread:
+   Note that both `gpu` and `cpu` are smart enough to recurse through tuples and namedtuples.
+   Another possibility is to use [`MLUtils.mapsobs`](https://juliaml.github.io/MLUtils.jl/dev/api/#MLUtils.mapobs) to push the data movement invocation into the background thread:
    ```julia
    using MLUtils: mapobs
    # ...
    gpu_train_loader = Flux.DataLoader(mapobs(gpu, (xtrain, ytrain)), batchsize = 16)
    ```
```
mcabbott, line 179: I saw this while updating the docs. Does it work? `mapobs` seems to obey strange rules:
```julia
julia> data = (a=[1 2 3; 4 5 6], b=[7,8,9]);

julia> mapobs(pr, data) |> collect
x = (a = [1, 4], b = 7)
x = (a = [2, 5], b = 8)
x = (a = [3, 6], b = 9)
3-element Vector{Any}:
 (a = [1, 4], b = 7)
 (a = [2, 5], b = 8)
 (a = [3, 6], b = 9)

julia> DataLoader(mapobs(pr, data); batchsize=2) |> collect
x = (a = [1 2; 4 5], b = [7, 8])
x = (a = [3; 6;;], b = [9])
2-element Vector{NamedTuple{(:a, :b), Tuple{Matrix{Int64}, Vector{Int64}}}}:
 (a = [1 2; 4 5], b = [7, 8])
 (a = [3; 6;;], b = [9])
```
But if it does work, this might be the obvious thing for `gpu(::DataLoader)` to do.

mcabbott: Why it's written up as a method of moving the entire dataset to the GPU is confusing. Seems to have been there from the start, and nobody noticed?

https://github.com/FluxML/Flux.jl/pull/1980/files#diff-813fc68aeb136c3af32fae13632adf27c24229ce9b56b151d13b836db8a7aa2dR117-R122

PR #2241, opened 2023-04-25
fix perf issue loadmodel!

CarloLucibello:
Fix performance issue mentioned in https://github.com/FluxML/Flux.jl/pull/2239#issuecomment-1521446746

When loading from a flux model into another everything is fine:
```julia
using Metalhead, Flux

m1 = ResNet(18)
m2 = ResNet(18)
@time Flux.loadmodel!(m2, m1) # warmup
@time Flux.loadmodel!(m2, m1)
#  0.003388 seconds (23.01 k allocations: 2.157 MiB) # fast on both master and this PR
```
When loading from a similar struct such as nested named tuples instead (as proposed in #2239) perf on master is very bad:
```julia
using Functors

function state(x)
    if Functors.isleaf(x)
        return x
    else
        return map(state, Functors.children(x))
    end
end

s = state(m1)
@time Flux.loadmodel!(m2, s) # warmup
@time Flux.loadmodel!(m2, s)
# 69.507712 seconds (1.06 G allocations: 40.744 GiB, 4.74% gc time) # master
# 0.002409 seconds (6.85 k allocations: 1.109 MiB)                               # this PR
```

The perf problem was due to huge strings being created at each recursion step as part of error messages.

@theabhirath 

~Also removed the `filter` argument, which was not documented nor used in practice.~


theabhirath:
`filter` is important for Metalhead. The motivation is described in https://github.com/FluxML/Flux.jl/pull/2041

src/loading.jl
```diff
@@ -81,21 +82,22 @@ however, attempting to copy a non-zero array to an inactive parameter will throw
 Likewise, copying a `src` value of `false` to any `dst` array is valid,
 but copying a `src` value of `true` will error.
 """
-function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
-  ldsts = _filter_children(filter, functor(dst)[1])
-  lsrcs = _filter_children(filter, functor(src)[1])
```
darsnack, line 86: This is featured we added to avoid recursing parts of the tree that do not match but also don't need to. See #2041. Is this the source of the performance issue? If not, it should be reverted.

darsnack (changes requested):

darsnack:
I don't have an issue with that, but I recall that last time there was some bikeshedding and we opted to keep it private. If it's still an issue, no need to let it hold up this PR.

darsnack (approved):

PR #2245, opened 2023-04-27
`gpu(::DataLoader)`, take III

mcabbott:
Simpler variant that just calls `DataLoader(mapobs(gpu, data), ...)`.

What this misses compared to more complex CuIterator-like things is that it does not call `finalize` afterwards. But perhaps that doesn't matter since each call of the model will allocate so much more, and also not `finalize` explicitly.

Closes #2240, closes #2186

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md

### DIFF
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -126,7 +126,7 @@ julia> x |> cpu
 
 In order to train the model using the GPU both model and the training data have to be transferred to GPU memory. Moving the data can be done in two different ways:
 
-1. Iterating over the batches in a [`DataLoader`](@ref) object transferring each one of the training batches at a time to the GPU. This is recommended for large datasets. Done hand, it might look like this:
+1. Iterating over the batches in a [`DataLoader`](@ref) object transferring each one of the training batches at a time to the GPU. This is recommended for large datasets. Done by hand, it might look like this:
    ```julia
    train_loader = Flux.DataLoader((X, Y), batchsize=64, shuffle=true)
    # ... model definition, optimiser setup
@@ -139,7 +139,7 @@ In order to train the model using the GPU both model and the training data have 
        end
    end
    ```
-   Rather than write this every time, there is a method of `gpu(::DataLoader)` which does it for you:
+   Rather than write this out every time, you can just call `gpu(::DataLoader)`:
    ```julia
    gpu_train_loader = Flux.DataLoader((X, Y), batchsize=64, shuffle=true) |> gpu
    # ... model definition, optimiser setup
@@ -161,7 +161,7 @@ In order to train the model using the GPU both model and the training data have 
        for (x, y) in gpu_train_loader
            # ...
    ```
-   Here `(X, Y) |> gpu` applies [`gpu`](@ref) to both arrays -- it recurses into not just tuples, as here, but also whole Flux models.
+   Here `(X, Y) |> gpu` applies [`gpu`](@ref) to both arrays, as it recurses into structures.
 
 ## Saving GPU-Trained Models
 
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -128,12 +128,12 @@ In order to train the model using the GPU both model and the training data have 
 
 1. Iterating over the batches in a [DataLoader](@ref) object transferring each one of the training batches at a time to the GPU. This is recommended for large datasets. Done hand, it might look like this:
    ```julia
-   train_loader = Flux.DataLoader((xtrain, ytrain), batchsize = 64, shuffle = true)
+   train_loader = Flux.DataLoader((X, Y), batchsize=64, shuffle=true)
    # ... model definition, optimiser setup
    for epoch in 1:epochs
-       for (xtrain_batch, ytrain_batch) in train_loader
-           x = gpu(xtrain_batch)
-           y = gpu(ytrain_batch)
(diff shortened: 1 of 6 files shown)


docs/src/gpu.md
```diff
@@ -122,61 +122,48 @@ julia> x |> cpu
...
-Some of the common workflows involving the use of GPUs are presented below.
-
-### Transferring Training Data
+## Transferring Training Data
 
-In order to train the model using the GPU both model and the training data have to be transferred to GPU memory. This process can be done with the `gpu` function in two different ways:
+In order to train the model using the GPU both model and the training data have to be transferred to GPU memory. Moving the data can be done in two different ways:
 
-1. Iterating over the batches in a [DataLoader](@ref) object transferring each one of the training batches at a time to the GPU. 
+1. Iterating over the batches in a [`DataLoader`](@ref) object transferring each one of the training batches at a time to the GPU. This is recommended for large datasets. Done hand, it might look like this:
```
CarloLucibello, line 129: "done hand"?

docs/src/gpu.md
```diff
@@ -122,61 +122,48 @@ julia> x |> cpu
...
-   ```
-   Note that both `gpu` and `cpu` are smart enough to recurse through tuples and namedtuples. Another possibility is to use [`MLUtils.mapsobs`](https://juliaml.github.io/MLUtils.jl/dev/api/#MLUtils.mapobs) to push the data movement invocation into the background thread:
-   ```julia
-   using MLUtils: mapobs
-   # ...
-   gpu_train_loader = Flux.DataLoader(mapobs(gpu, (xtrain, ytrain)), batchsize = 16)
-   ```
-
-3. Wrapping the `DataLoader` in [`CUDA.CuIterator`](https://cuda.juliagpu.org/stable/usage/memory/#Batching-iterator) to efficiently move data to GPU on demand:
+   Rather than write this every time, there is a method of `gpu(::DataLoader)` which does it for you:
```
CarloLucibello, line 142:
```suggestion
   Rather than write this every time, the method `gpu(::DataLoader)` does it for you:
```

docs/src/gpu.md
```diff
@@ -122,61 +122,48 @@ julia> x |> cpu
...
+   # ... model definition, optimiser setup
+   for epoch in 1:epochs
+       for (x, y) in gpu_train_loader
+           grads = gradient(m -> loss(m, x, y), model)
+           Flux.update!(opt_state, model, grads[1])
        end
    end
    ```
+   This is equivalent to `DataLoader(MLUtils.mapobs(gpu, (X, Y)); keywords...)`.
+   Something similar can also be done with [`CUDA.CuIterator`](https://cuda.juliagpu.org/stable/usage/memory/#Batching-iterator), `gpu_train_loader = CUDA.CuIterator(train_loader)`. However, this only works with a limited number of data types: `first(train_loader)` should be a tuple (or `NamedTuple`) of arrays.
```
CarloLucibello, line 154: Here we could hint at using `mapobs` to transform the dataset into something CUIterator compatible. Could be a short example like
```julia
train_loader = mapobs(preprocess_transform, train_loader)
gpu_train_loader = CUDA.CuIterator(train_loader)
```
 Also, mention when CuIterator should be preferred over gpu?

docs/src/gpu.md
```diff
@@ -122,61 +122,48 @@ julia> x |> cpu
...
-   Note that this works with a limited number of data types. If `iterate(train_loader)` returns anything other than arrays, approach 1 or 2 is preferred.
+2. Transferring all training data to the GPU at once before creating the `DataLoader`. This is usually performed for smaller datasets which are sure to fit in the available GPU memory.
+   ```julia
+   gpu_train_loader = Flux.DataLoader((X, Y) |> gpu, batchsize = 32)
+   # ...
+   for epoch in 1:epochs
+       for (x, y) in gpu_train_loader
+           # ...
+   ```
+   Here `(X, Y) |> gpu` applies [`gpu`](@ref) to both arrays -- it recurses into not just tuples, as here, but also whole Flux models.
```
CarloLucibello, line 164:
```suggestion
   Here `(X, Y) |> gpu` applies [`gpu`](@ref) to both arrays -- it recurses into not just tuples, as here, but also nested data structures.
```

docs/src/gpu.md
mcabbott:
```suggestion
1. Iterating over the batches in a [`DataLoader`](@ref) object transferring each one of the training batches at a time to the GPU. This is recommended for large datasets. Done by hand, it might look like this:
```

docs/src/gpu.md
mcabbott:
```suggestion
   Here `(X, Y) |> gpu` applies [`gpu`](@ref) to both arrays, as it recurses into structures.
```

docs/src/gpu.md
mcabbott: Is it preferred? This PR takes the line that it's not... it's tricky about what types, and finalize doesn't matter. So it's mentioned here in case people are already using it. 

If finalize does matter, then we should do https://github.com/FluxML/Flux.jl/pull/2240 instead.

docs/src/gpu.md
mcabbott:
```suggestion
   Rather than write this out every time, you can just call `gpu(::DataLoader)`:
```

PR #2252, opened 2023-05-02
Remove greek-letter keyword from `normalise`

mcabbott:
This case missed in #2139


mcabbott:
Apparently this is our test of this function (besides the docstring):

https://github.com/FluxML/Flux.jl/blob/8a5b9774f0d6b254f60c88ca07a984997a016e75/test/layers/stateless.jl#L12

ToucheSir (approved):

Issue #2237, opened 2023-04-23
Separate nightly CI from posted CI badge


ToucheSir:
Inspired from a discussion on Slack about how other packages handle the same. Flux frequently has to deal with this because new nightly releases often break Zygote, but Flux itself is rarely affected. There seem to be two main approaches to accomplishing this:

1. Allow the nightly CI job to fail, e.g. https://github.com/JuliaSmoothOptimizers/SolverTest.jl/pull/6. This is simple to implement but does risk missing real failures if one doesn't dig beyond the big green checkmark.
2. Separate nightly CI into its own job and show the non-nightly results badge, e.g. https://github.com/JuliaManifolds/Manifolds.jl/blob/master/.github/workflows/nightly.yml. This is more work but will still fail loudly for PRs.

A related question is whether we want to do the same for downstream tests. There are some known downstreams which are quite flaky, but since we don't include any downstream results in the README I'd assume this is less of a problem.

PR #2254, opened 2023-05-06
[CI] Separate julia-nightly from common matrix generator

skyleaworlder:
About #2237 

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### Screenshot

The corresponding run is  [here](https://github.com/skyleaworlder/Flux.jl/actions/runs/4901856571).

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,24 +13,29 @@ jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.allow_failure }}
     strategy:
       fail-fast: false
       matrix:
         version:
           - '1.6' # Replace this with the minimum Julia version that your package supports.
           - '1'
-          - 'nightly'
-        os:
-          - ubuntu-latest
-        arch:
-          - x64
+        os: [ubuntu-latest]
+        arch: [x64]
+        allow_failure: [false]
         include:
+          - os: ubuntu-latest
+            version: 'nightly'
+            arch: x64
+            allow_failure: true
           - os: windows-latest
             version: '1'
             arch: x64
+            allow_failure: false
           - os: macOS-latest
             version: '1'
             arch: x64
+            allow_failure: false
     steps:
       - uses: actions/checkout@v3
       - uses: julia-actions/setup-julia@v1


skyleaworlder:
The current problem I feel confused is:

In this PR, I declare "ubuntu-latest&nightly" as "allow_failure" in "include", which means:

```yaml
matrix:
        version:
          - '1.6' # Replace this with the minimum Julia version that your package supports.
          - '1'
        os: [ubuntu-latest]
        arch: [x64]
        allow_failure: [false]
        include:
          - os: ubuntu-latest
            version: 'nightly'
            arch: x64
            allow_failure: true
          - os: windows-latest
            version: '1'
            arch: x64
            allow_failure: false
          - os: macOS-latest
            version: '1'
            arch: x64
            allow_failure: false
```

While as [GitHub Actions | Using Workflows | Workflow Syntax for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude) said, I think it equals to:

```yaml
matrix:
        version:
          - '1.6' # Replace this with the minimum Julia version that your package supports.
          - '1'
          - 'nightly'
        os: [ubuntu-latest]
        arch: [x64]
        allow_failure: [false]
        include:
          - version: 'nightly'
            allow_failure: true
          - os: windows-latest
            version: '1'
            arch: x64
            allow_failure: false
          - os: macOS-latest
            version: '1'
            arch: x64
            allow_failure: false
```

But it failed [here](https://github.com/skyleaworlder/Flux.jl/actions/runs/4901653312).

image

I wonder if I made some non-obvious but low-level mistakes. 😕

ToucheSir:
Thanks! Instead of adding an extra `allow_failure` field to each configuration, could we not check if `matrix.version` is `nightly` directly in the `continue-on-error` condition?

skyleaworlder:
> Thanks! Instead of adding an extra `allow_failure` field to each configuration, could we not check if `matrix.version` is `nightly` directly in the `continue-on-error` condition?

Originally I think it would work well. But now I'm not sure about that, due to the unexpected expression parse error mentioned above. (Maybe it's related to some complicated use cases) But I'll try.

skyleaworlder:
Additionally, it seems that we can only make the badge show "passing" with a given branch, like these: [![][action-img]][action-url] (I use the branch "ci" of my own repo)

But It looks like we cannot show "passing" for a specific pipeline in the Actions, which we make it "passing" falsely with `allow_failure`, which seems to be a bug : [![][action-ci]][action-ci-url] (Here I use the CI.yml, which is supposed to show "passing")

image


[action-img]: https://github.com/skyleaworlder/Flux.jl/actions/workflows/ci.yml/badge.svg?branch=ci
[action-url]: https://github.com/skyleaworlder/Flux.jl/tree/ci
[action-ci]: https://github.com/skyleaworlder/Flux.jl/actions/workflows/ci.yml/badge.svg
[action-ci-url]: https://github.com/skyleaworlder/Flux.jl/actions/workflows/ci.yml

ToucheSir:
Yup, that's what I was driving at with "This is simple to implement but does risk missing real failures if one doesn't dig beyond the big green checkmark." in the issue.

skyleaworlder:
After that, I followed the second suggestion in issue 2237.

Although the badge is still displaying "failing" ([![][action-ci]][action-ci-url]) both here and in README.md, this is because GitHub Actions Badge always show the status of default branch.

We can see:

image

image


[action-ci]: https://github.com/FluxML/Flux.jl/actions/workflows/ci.yml/badge.svg
[action-ci-url]: https://github.com/FluxML/Flux.jl/actions/workflows/ci.yml

ToucheSir (approved):
Good work!

skyleaworlder:
As issue 2237 expected:

> Separate nightly CI into its own job and show the non-nightly results badge.

Yesterday I just want to check if the badge is able to show the correct status, so I didn't hurry to update the badge in README. (Now you can see [CI](https://github.com/FluxML/Flux.jl/actions/workflows/ci.yml) and [Nightly CI](https://github.com/FluxML/Flux.jl/actions/workflows/nightly-ci.yml))

So could you reopen this PR, I don't want to open another PR, which is a kind of pollution. 🥲

ToucheSir:
No it's not your fault but quite intentional from me. I wasn't sure if we wanted to show nightly CI in the README so I didn't bother asking for it in the PR. We can pick up the discussion in #2237.

PR #2281, opened 2023-06-29
fix doc of `PairwiseFusion`

ctarn:
according to the implementation: https://github.com/FluxML/Flux.jl/blob/24b1eb2f0c7de93e36d47833ae47a1b473f31f80/src/layers/basic.jl#L637, the `connection` of `PairwiseFusion` is called as `connection(y1, x2)` instead of `connection(x2, y1)`.


vmpyr (approved):

darsnack (approved):

darsnack:
Thanks!

PR #2289, opened 2023-07-14
Update description of `trainable` in "advanced.md"

mcabbott:
This sounded like it suggested `@functor Affine (W,)` was much the same as `trainable`, but it really isn't. 

"advanced.md" in the manual, 

old: http://fluxml.ai/Flux.jl/stable/models/advanced/#Customising-Parameter-Collection-for-a-Model

new: https://github.com/FluxML/Flux.jl/blob/mcabbott-patch-2/docs/src/models/advanced.md 

(Maybe we should aim to kill this whole file, or re-write it to something nicer.)

### DIFF
--- a/docs/src/models/advanced.md
+++ b/docs/src/models/advanced.md
@@ -36,34 +36,38 @@ For an intro to Flux and automatic differentiation, see this [tutorial](https://
 
 Taking reference from our example `Affine` layer from the [basics](@ref man-basics).
 
-By default all the fields in the `Affine` type are collected as its parameters, however, in some cases it may be desired to hold other metadata in our "layers" that may not be needed for training, and are hence supposed to be ignored while the parameters are collected. With Flux, it is possible to mark the fields of our layers that are trainable in two ways.
-
-The first way of achieving this is through overloading the `trainable` function.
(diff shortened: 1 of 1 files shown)


docs/src/models/advanced.md
```diff
@@ -36,34 +36,38 @@ For an intro to Flux and automatic differentiation, see this [tutorial](https://
...
 
-```julia
-Flux.@functor Affine (W,)
+```julia-repl
+julia> Flux.params(Affine(true, [10, 11, 12.0]))
+Params([])
 ```
 
-However, doing this requires the `struct` to have a corresponding constructor that accepts those parameters.
+It is also possible to further restrict what fields are seen by writing `@functor Affine (W,)`. However, this is not recommended. This requires the `struct` to have a corresponding constructor that accepts those parameters. And the contentes `gpu`
```
darsnack, line 70:
```suggestion
It is also possible to further restrict what fields are seen by writing `@functor Affine (W,)`. However, this is not recommended. This requires the `struct` to have a corresponding constructor that accepts only `W` as an argument, and the ignored fields will not be seen by functions like `gpu` (which is usually undesired).
```

mcabbott: Thanks, I forgot to finish the sentence!

darsnack (approved):

PR #2299, opened 2023-07-28
Lecun Normal added

vortex73:
Lecun Normal Initializer added
### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


vortex73:
Please mention any corrections if any.

mcabbott:
Welcome!

This needs a docstring describing what it does (& linking to the paper?), and an entry in the manual. 

It also needs tests, I think somewhere there's a list of init functions which get looped over, to which this could be added.

PR #2314, opened 2023-08-19
Removing deprecated method call in `GPU_BACKEND_ORDER`.

### LINKED ISSUE #2312: Warning: `sort(d::Dict; args...)` is deprecated, use `sort!(OrderedDict(d); args...)` instead.
In precompilation, I got:
```julia
┌ Flux [587475ba-b771-5e3f-ad9e-33799f191a9c]
│  ┌ Warning: `sort(d::Dict; args...)` is deprecated, use `sort!(OrderedDict(d); args...)` instead.
│  │   caller = top-level scope at functor.jl:190
│  └ @ Core C:\Users\Hossein Pourbozorg\.julia\packages\Flux\pR3k3\src\functor.jl:190
└
```

codetalker7:
Closes #2312.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -187,16 +187,9 @@ _isbitsarray(x) = false
 _isleaf(::AbstractRNG) = true
 _isleaf(x) = _isbitsarray(x) || Functors.isleaf(x)
 
-const GPU_BACKEND_ORDER = sort(
-    Dict(
-        "CUDA" => 1,
-        "AMD" => 2,
-        "Metal" => 3,
-        "CPU" => 4,
-    ),
-    byvalue = true
-) 
-const GPU_BACKENDS = tuple(collect(keys(GPU_BACKEND_ORDER))...)
+# the order below is important
+const GPU_BACKENDS = ["CUDA", "AMD", "Metal", "CPU"]
+const GPU_BACKEND_ORDER = Dict(collect(zip(GPU_BACKENDS, 1:length(GPU_BACKENDS))))
 const GPU_BACKEND = @load_preference("gpu_backend", "CUDA")
 
 function gpu_backend!(backend::String)


ToucheSir (approved):
Nice, this is cleaner than my ideas!

prbzrg:
This isn't same as before, it should be `OrderedDict` from `OrderedCollections`.
```julia
julia> const GPU_BACKENDS = ["CUDA", "AMD", "Metal", "CPU"]
4-element Vector{String}:
 "CUDA"
 "AMD"
 "Metal"
 "CPU"

julia> const GPU_BACKEND_ORDER = Dict(collect(zip(GPU_BACKENDS, 1:length(GPU_BACKENDS))))
Dict{String, Int64} with 4 entries:
  "CUDA"  => 1
  "AMD"   => 2
  "CPU"   => 4
  "Metal" => 3

julia> const GPU_BACKEND_ORDER2 = OrderedDict(collect(zip(GPU_BACKENDS, 1:length(GPU_BACKENDS))))
OrderedDict{String, Int64} with 4 entries:
  "CUDA"  => 1
  "AMD"   => 2
  "Metal" => 3
  "CPU"   => 4

```

prbzrg:
on `Flux@0.14.2`:
```julia
julia> Flux.GPU_BACKEND_ORDER
OrderedDict{String, Int64} with 4 entries:
  "CUDA"  => 1
  "AMD"   => 2
  "Metal" => 3
  "CPU"   => 4
```

ToucheSir:
`GPU_BACKEND_ORDER` isn't a part of the public API, nor is its iteration order defined. We're not going to add a package dependency just for this.

prbzrg:
Oh, I was under the impression that the order of it matters because there was a `sort` that caused the deprecation warning.

PR #2319, opened 2023-08-25
allow get_device("Metal") and informative error messages

CarloLucibello:


test/ext_metal/get_devices.jl
```diff
@@ -1,16 +1,14 @@
...
 
-if Metal.functional()
-    @test typeof(metal_device.deviceID) <: Metal.MTLDevice 
-else
-    @test typeof(metal_device.deviceID) <: Nothing
+  @test typeof(metal_device.deviceID) <: Metal.MTLDevice 
 end
 
-# testing get_device
-if Metal.functional()
```
CarloLucibello, line 13: these tests are already run only in case Metal is functional

CarloLucibello:
bump

CarloLucibello:
Maybe I can get a review from @codetalker7

test/ext_metal/get_devices.jl
codetalker7: Yes, I just noticed this. I think this is true for tests for AMDGPU and CUDA as well. So should we remove the redundant `.functional()` check there as well (line 6 of https://github.com/FluxML/Flux.jl/blob/master/test/ext_amdgpu/get_devices.jl and line 6 of https://github.com/FluxML/Flux.jl/blob/master/test/ext_cuda/get_devices.jl)?

CarloLucibello: done

src/functor.jl
```diff
@@ -650,10 +650,10 @@ function get_device(; verbose=false)::AbstractDevice
 end
 
 """
-    Flux.get_device(backend::String, ordinal::Int = 0)::Flux.AbstractDevice
+    Flux.get_device(backend::String, idx::Int = 0)::Flux.AbstractDevice
```
ToucheSir, line 653: I tried to look around to see what the vendor libraries are calling this, and based on that I think we should stick to `ordinal` or switch to `id`. OneAPI uses device ID, CUDA uses device ordinal and device ID, and ROCm uses ID + ordinal + index but in a slightly different way (indices and ordinals are 0-indexed device #'s, while the IDs mentioned are UUIDs).

An alternative would be to double up terms in the docstring and use "ordinal ID" or "ordinal index". Either way, I think one more signal that this isn't following Julia's default one-based indexing would be good so that users aren't going off-by-one (mostly by going too high).

Otherwise, LGTM.

CarloLucibello: I moved everything to "id"

codetalker7 (approved):
LGTM too.

PR #2323, opened 2023-08-28
Adding tooling for `JuliaFormatter`.

### LINKED ISSUE #2298: move the codebase to 4 spaces indentation
Two spaces indentation has always been an anomaly in the landscape of julia packages.
Should we get rid of it?

codetalker7:
This PR is to add some format tooling to Flux. A new environment is created in the `dev` folder, which will contain all dependencies required for the tooling (currently it only contains `JuliaFormatter.jl`). A new config file for `JuliaFormatter` is added, along with a script to run `JuliaFormatter` (`dev/flux_format.jl`). Along with this, a new `JuliaFormatter` CI has been added as well. Much of this is inspired from the workflow in [ClimaCore.jl](https://github.com/CliMA/ClimaCore.jl).

TODO: 

- [x] Implement a pre-commit hook which runs the formatting script before each commit.
- [x] Do we need to add docs for this?
- [x] Do we add this to our test suite?

Closes #2298.

cc @ToucheSir @CarloLucibello @darsnack

### DIFF
--- a/.github/workflows/JuliaFormatter.yml
+++ b/.github/workflows/JuliaFormatter.yml
@@ -1,0 +1,47 @@
+on:
+  push:
+    branches:
+      - main
+      - trying
+      - staging
+    tags: '*'
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+
+jobs:
+  format:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+    steps:
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.4.0
+      with:
+        access_token: ${{ github.token }}
+
+    - uses: actions/checkout@v2.2.0
+
+    - uses: dorny/paths-filter@v2.9.1
+      id: filter
+      with:
+        filters: |
+          julia_file_change:
+            - added|modified: '**.jl'
+
+    - uses: julia-actions/setup-julia@latest
+      if: steps.filter.outputs.julia_file_change == 'true'
+      with:
+        version: 1.9
+
+    - name: Apply JuliaFormatter
+      if: steps.filter.outputs.julia_file_change == 'true'
+      run: |
+        julia --color=yes dev/flux_format.jl --verbose .
+
+    - name: Check formatting diff
+      if: steps.filter.outputs.julia_file_change == 'true'
+      run: |
+        git diff --color=always --exit-code
--- a/.github/workflows/JuliaFormatter.yml
+++ b/.github/workflows/JuliaFormatter.yml
@@ -1,9 +1,7 @@
 on:
   push:
     branches:
-      - main
-      - trying
-      - staging
+      - master
     tags: '*'
   pull_request:
     types:
@@ -17,11 +15,6 @@ jobs:
     runs-on: ubuntu-20.04
     timeout-minutes: 30
     steps:
-    - name: Cancel Previous Runs
-      uses: styfle/cancel-workflow-action@0.4.0
-      with:
-        access_token: ${{ github.token }}
-
     - uses: actions/checkout@v2.2.0
 
     - uses: dorny/paths-filter@v2.9.1
--- a/.JuliaFormatter.toml
+++ b/.JuliaFormatter.toml
@@ -1,0 +1,6 @@
+indent = 4
+margin = 80
+always_for_in = true
+whitespace_typedefs = true
+whitespace_ops_in_indices = true
+remove_extra_newlines = false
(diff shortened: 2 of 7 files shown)


codetalker7:
Just a heads up: `git` hooks in the `.git` directory aren't versioned controlled by default. Assuming that users have at least Git 2.9, we can set the `core.hooksPath` config to a custom directory in the project (say `.githooks`). To do this we'll have to probably write another script which does this whenever Flux is cloned (or installed).

Any other alternatives for managing hooks (for example [`pre-commit`](https://pre-commit.com/)) will require users to install other packages most likely.

Saransh-cpp:
See #2074 for a lot of discussion over JuliaFormatter and tools like pre-commit 🙂

ToucheSir:
Originally I had thought that one could use `.gitattributes` to work around having to install hooks as a separate step, but that turns out to not be possible. As such, we may have to compromise and make a Julia script which does tool setup + hook installation.

codetalker7:
> Originally I had thought that one could use `.gitattributes` to work around having to install hooks as a separate step, but that turns out to not be possible. As such, we may have to compromise and make a Julia script which does tool setup + hook installation. Technical setup documentation (e.g. "run this script") would then be added in CONTRIBUTING.md.

Yes, that sounds good. Also, how about adding this as a build step for Flux? So adding all these steps in `deps/build.jl` and using the `deps` folder for tooling instead of `dev` (I'm not too familiar with the pros/cons of doing this as part of the build step)?

ToucheSir:
I would double-check if build scripts can run on package install. If so, we shouldn't be using them for this.

codetalker7:
How does the `setup.jl` file look? Developers will be able to set up the hook by running `julia --project=dev dev/setup.jl`. I hope the `chmod` function from `Git.jl` works on non-Linux OS's as well. If not, I'll have to add some conditionals for each OS.

.github/workflows/JuliaFormatter.yml
```diff
@@ -0,0 +1,47 @@
+on:
+  push:
+    branches:
+      - main
```
CarloLucibello, line 4:
```suggestion
      - master
```
?

codetalker7: Ohh yes, it should be `master` instead of `main`. I'll update it.

codetalker7: Made the change; also removed the other two spurious branches, and removed a deprecated workflow.

codetalker7:
Also added formatting information to `CONTRIBUTING.md`. One question: should we have the `--verbose` option in the pre-commit hook? It prints all the formatted files when the commit is initiated.

ToucheSir:
I would say no. Does it provide any visual feedback that formatting is running right now without `--verbose`? If not we'd want to add at least a "formatting in progress" println or something like it.

ToucheSir:
Oh sorry, another thing I forgot to mention is that we should split up the commits adding tooling + CI and the commits actually reformatting the codebase. If you're comfortable doing the history manipulation to have those on the same PR go for it, just make sure to add a `.git-blame-ignore-revs` like we have [in Metalhead](https://github.com/FluxML/Metalhead.jl/blob/master/.git-blame-ignore-revs). If not, feel free to back out the reformatting and we can leave that for a future PR.

codetalker7:
> Oh sorry, another thing I forgot to mention is that we should split up the commits adding tooling + CI and the commits actually reformatting the codebase. If you're comfortable doing the history manipulation to have those on the same PR go for it, just make sure to add a `.git-blame-ignore-revs` like we have [in Metalhead](https://github.com/FluxML/Metalhead.jl/blob/master/.git-blame-ignore-revs). If not, feel free to back out the reformatting and we can leave that for a future PR.

Done. How does it look now?

ToucheSir:
It looks good, but I realized another issue, sorry. The formatting style being used is not one many contributors like. In particular, I think the splitting args across multiple lines instead of trying to wrap on as few lines as possible was a deal-breaker. Do you mind just backing out the reformat for now? I don't want us to lose out on merging the nice tooling PR while we bikeshed what the formatting style should be!

codetalker7:
> It looks good, but I realized another issue, sorry. The formatting style being used is not one many contributors like. In particular, I think the splitting args across multiple lines instead of trying to wrap on as few lines as possible was a deal-breaker. Do you mind just backing out the reformat for now? I don't want us to lose out on merging the nice tooling PR while we bikeshed what the formatting style should be!

Sure, I've removed the formatting commit. Please let me know if anything else is missing/needed. 

Also, do we want to add the formatting check to our test suite?

ToucheSir:
Do you mean to the package test suite? Probably not, I think.

codetalker7:
> Do you mean to the package test suite? Probably not, I think.

Okay 👍

CarloLucibello (approved):

PR #2328, opened 2023-09-06
rename "AMD" backend to "AMDGPU"

CarloLucibello:
Deprecate "AMD" in favor of "AMDGPU". 
"AMDGPU" is unambiguous and corresponds to the name of the related package, "AMD" instead could refer to either the cpu or the gpu. 

Additionally, this PR improves the error messages for `get_device`.

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -696,5 +696,9 @@ end
 
 # Fallback
 function get_device(::Val{D}, idx) where D
-    error("Unsupported backend: $(D). Try importing the corresponding package.")
+    if D ∈ (:CUDA, :AMDGPU, :Metal)
+        error("Unaivailable backend: $(D). Try importing the corresponding package with `using $D`.")
+    else
+        error("Unsupported backend: $(D). Supported backends are $(GPU_BACKENDS).")
+    end
 end
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -248,7 +248,7 @@ function gpu(x)
     @static if GPU_BACKEND == "CUDA"
         gpu(FluxCUDAAdaptor(), x)
     elseif GPU_BACKEND == "AMD"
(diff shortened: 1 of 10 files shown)


src/functor.jl
```diff
@@ -505,11 +508,11 @@ Base.@kwdef struct FluxCUDADevice <: AbstractDevice
 end
 
 """
-    FluxAMDDevice <: AbstractDevice
+    FluxAMDGPUDevice <: AbstractDevice
 
-A type representing `device` objects for the `"AMD"` backend for Flux.
+A type representing `device` objects for the `"AMDGPU"` backend for Flux.
 """
-Base.@kwdef struct FluxAMDDevice <: AbstractDevice
```
ToucheSir, line 512: I didn't see a binding deprecation of `FluxAMDDevice` to go along with this rename.

CarloLucibello: right, I didn't think it was exposed but indeed it is. Deprecated now.

src/functor.jl
```diff
@@ -248,7 +248,10 @@ function gpu(x)
     @static if GPU_BACKEND == "CUDA"
         gpu(FluxCUDAAdaptor(), x)
     elseif GPU_BACKEND == "AMD"
-        gpu(FluxAMDAdaptor(), x)
+        @warn "\"AMD\" backend is deprecated. Please use \"AMDGPU\" instead." maxlog=1
```
ToucheSir, line 251: Thoughts on this vs `Base.depwarn(...; force=true)`? On one hand that reads like the more appropriate function for the job and we've used it in Flux before. On the other hand, the lack of line/file info in depwarn (which `@warn` has by default) could be annoying.

ToucheSir (approved):

src/functor.jl
CarloLucibello: I would go with warn, having the line info is useful.

PR #2342, opened 2023-10-09
Use new `public` feature

mcabbott:
This marks un-exported symbols public, using a new feature of Julia 1.11. 

One use is that `Flux.mse` can be found by tab completion, etc:
```
julia> Flux.m  # <tab>
mae               make_causal_mask  maxpool
maxpool!          meanpool          meanpool!
mish              modules           mse
msle              multigate

julia> Flux.one  # Flux.o<tab>
onecold
onehot
onehotbatch
ones32
```
Not certain the list here is complete.


src/Flux.jl
```diff
@@ -69,5 +83,37 @@ include("deprecations.jl")
 include("losses/Losses.jl")
 using .Losses
 
+@compat(public, (
+  # init
+  glorot_uniform,
+  glorot_normal,
```
mcabbott, line 89: I called the macro with brackets just to allow comments like this. But eventual syntax is just `public glorot_uniform, glorot_normal` like `using`, for which this isn't possible.

ToucheSir: I will express a slight personal preference for this over the indent-only syntax on L60-62, but the best way to settle the debate might be to see what the auto-formatter outputs.

mcabbott:
The autoformatter's opinion is that with two brackets this needs four spaces:

Screenshot 2023-11-28 at 11 22 30

... and that with no brackets, no indent:

Screenshot 2023-11-28 at 11 22 48

src/Flux.jl
mcabbott: All now switched to this format.

PR #2344, opened 2023-10-13
Some small printing upgrades

mcabbott:
This is intended for https://github.com/FluxML/Fluxperimental.jl/pull/20 but probably a good idea anyway. Motivating example is something like this (where bigger arrays will print pages of numbers):
```julia
julia> struct Tmp2; x; y; end; Flux.@functor Tmp2

julia> Chain(Tmp2([Dense(2,3), randn(3,4)'], (x=1:3, y=Dense(3,4), z=rand(3))))
Chain(
  Tmp2(
    Array(
      Dense(2 => 3),                    # 9 parameters
      [0.351978391016603 0.6408681372462821 -1.326533184688648; 0.09481930831795712 
1.430103476272605 0.7250467613675332; 2.03372151428719 -0.015879812799495713 
1.9499692162118236; -1.6346846180722918 -0.8364610153059454 -1.2907265737483433],  # 12 parameters
    ),
    NamedTuple(
      1:3,                              # 3 parameters
      Dense(3 => 4),                    # 16 parameters
      [0.9666158193429335, 0.01613900990539574, 0.0205920186127464],  # 3 parameters
    ),
  ),
)                   # Total: 7 arrays, 43 parameters, 644 bytes.
```
Notice that `Array()` and `NamedTuple()` aren't actually valid syntax. Also that it prints whole arrays if they aren't inside a layer the way it expects. After this PR:
```julia
julia> Chain(Tmp2([Dense(2,3), randn(3,4)'], (x=1:3, y=Dense(3,4), z=rand(3))))
Chain(
  Tmp2(
    [
      Dense(2 => 3),                    # 9 parameters
      4×3 Adjoint,                      # 12 parameters
    ],
    (;
      x = 3-element UnitRange,          # 3 parameters
      y = Dense(3 => 4),                # 16 parameters
      z = 3-element Array,              # 3 parameters
    ),
  ),
)                   # Total: 7 arrays, 43 parameters, 644 bytes.
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.14.22"
+version = "0.14.23"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -64,6 +64,7 @@ function _applychain(layers::AbstractVector, x)  # type-unstable path, helps com
   end
   x
 end
+_show_pre_post(::Chain{<:AbstractVector}) = "Chain([", "])"  # internal method for show
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -64,7 +64,6 @@ function _applychain(layers::AbstractVector, x)  # type-unstable path, helps com
   end
   x
 end
-_show_pre_post(::Chain{<:AbstractVector}) = "Chain([", "])"  # internal method for show
 
 Base.getindex(c::Chain, i::AbstractArray) = Chain(c.layers[i])
 Base.getindex(c::Chain{<:NamedTuple}, i::AbstractArray) =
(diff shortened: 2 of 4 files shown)


ToucheSir:
Is having the `summary` instead of the actual contents important for `@compact`? If not, my preference would be to keep the more accurate printing but drop that part.

mcabbott:
The idea is that printing the values of a huge matrix isn't helpful, but showing the size is. This is what the present printing of e.g. Dense achieves. 

But `@compact` seems to often want to store the arrays themselves without some layer.

ToucheSir:
Revisiting this, it would be nice to retain some more info such as eltype if the contents aren't printed. Though it's far from perfect, could we avoid a lot of complexity by setting the `:limit` and `:compact` IOContext properties when showing these leaf nodes?

mcabbott:
Latest commit keeps eltype as suggested (or really, the first type parameter):
```julia
julia> struct Tmp2; x; y; end; Flux.@functor Tmp2

julia> Chain(Tmp2([Dense(2,3), randn(3,4)'], (x=1:3, y=Dense(3,4), z=rand(3))))
Chain(
  Tmp2(
    [
      Dense(2 => 3),                    # 9 parameters
      4×3 Adjoint{Float64,...},         # 12 parameters
    ],
    (;
      x = 3-element UnitRange{Int64},   # 3 parameters
      y = Dense(3 => 4),                # 16 parameters
      z = 3-element Vector{Float64},    # 3 parameters
    ),
  ),
)                   # Total: 7 arrays, 43 parameters, 780 bytes.
```

darsnack:
We were discussing this during our call this week. The consensus was that using `IOContext(..., :limit => true)` is preferable to a custom `summary` for arrays. This would keep things succinct while still matching Base standards/expectations.

There are times when the type (e.g. `Adjoint`) is desired. This case might be better addressed by offering something like [Flax's `tabulate`](https://flax.readthedocs.io/en/latest/api_reference/flax.linen/module.html#flax.linen.Module.tabulate).

ToucheSir (approved):

PR #2359, opened 2023-12-05
Bump AMDGPU compat to 0.8

pxl-th:
Also move AMDGPU CI to Julia 1.10 because of Navi 3 which is supported only on 1.10+.


CarloLucibello (approved):

pxl-th:
Don't have the write access to merge it, can someone merge please?

PR #2360, opened 2023-12-17
Use stable API for AMDGPU RNG conversion

ToucheSir:
The function we were calling is no longer available in the `AMDGPU.rocRAND` module. `rocrand_rng` has been around for years and is also more explicit.

Also adds tests so we can catch changes like this in future.

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md


CarloLucibello (approved):

CarloLucibello:
not related to this, but CUDA's tests are failing

PR #2369, opened 2024-01-21
Add back missing `rng_from_array(::CuArray)` method

Red-Portal:
Hi, for some reason, `rng_from_array(::CuArray)` was missing from the loading path in `FluxCUDAExt`. 
Seems like it has been left behind when migrating to extensions.
I enabled it back, and added a corresponding tests.
For some reason, the `tests/ext_cuda/utils.jl` was also not being executed despite already containing a test case.
I also added this into the corresponding `runtests.jl`.  

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md


ToucheSir (approved):
Woops, that's quite the omission. Thanks!

PR #2375, opened 2024-02-02
Allow scaling of `init` functions

mcabbott:
This allows `init = 2 * randn32` to scale the initial values.

Implemented as a struct like Fix1, in order that `2 * randn32(rng)` can also work.

Should probably apply to every function, and needs tests etc. But is this a good idea, first?

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/docs/src/utilities.md
+++ b/docs/src/utilities.md
@@ -27,6 +27,22 @@ julia> Dense(4 => 5, tanh; init=Flux.randn32(MersenneTwister(1)))
 Dense(4 => 5, tanh)  # 25 parameters
 ```
 
+All of the initialisation functions may be multiplied by a number
+to scale their output:
+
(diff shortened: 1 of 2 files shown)


ToucheSir:
My inclination is that `init = (args...) -> 2randn32(args...)` or `init = (args...) -> lmul!(2, randn32(args...))` isn't significantly more boilerplate and is a little more flexible in terms of supported operations. Perhaps documenting this pattern would be the way to go?

darsnack:
I feel the same way. A neat trick but seems like too much for something that is easily implemented as an anonymous or one line function.

mcabbott:
Ok, I think I agree this is too much. Better to stay closer to Base... and if in doubt you should make the matrix yourself.

Issue #2216, opened 2023-03-20
docs on freezing layers should be ported to the explicit syntax


CarloLucibello:
https://fluxml.ai/Flux.jl/stable/models/advanced/#Freezing-Layer-Parameters
should be reworded on the line of
https://fluxml.ai/Optimisers.jl/dev/#Frozen-Parameters

Mia-li88:
Hi! i added some changes on the docs.
Here's the PR: [https://github.com/FluxML/Optimisers.jl/pull/133](url)

mcabbott:
That file is a bit of a random grab-bag of things, I wonder how useful it is to anyone? It's headed "Customised Layers" and maybe there could be a nice description of how & why to make layers, somewhere.

Besides that, it mentions feezing, and links to the description at  http://fluxml.ai/Flux.jl/stable/training/training/#Freezing-and-Schedules  Maybe that section can just be deleted?

And it mentions `trainable`, which is a better fit (as it permanently customises layer definitions). This has a docstring at  http://fluxml.ai/Flux.jl/stable/destructure/#Optimisers.trainable  but isn't described elsewhere.

mcabbott:
Re freezing, one thing you might be tempted to use `freeze!` for is roughly transfer learning: Fixing all but the last few layers, to train only those. That will work but is quite inefficient (and the old Params equivalent was too). What you should do is only differentiate part of the model, something like this:  https://github.com/FluxML/model-zoo/blob/master/tutorials/transfer_learning/transfer_learning.jl#L117-L120

We don't say that in the docs anywhere right now. And guessing it requires quite a bit of understanding of how Zygote works. So perhaps we should... but where I don't know.

PR #2385, opened 2024-02-29
docs: improve freezing docs

isentropic:
@ToucheSir as per slack discussion

(Edit: closes https://github.com/FluxML/Flux.jl/issues/2216 )

### DIFF
--- a/.github/workflows/JuliaFormatter.yml
+++ b/.github/workflows/JuliaFormatter.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4.1.1
 
-    - uses: dorny/paths-filter@v3.0.1
+    - uses: dorny/paths-filter@v3.0.2
       id: filter
       with:
         filters: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
         if: contains(fromJson('["1", "1.9"]'), matrix.version) && matrix.os == 'ubuntu-latest'
       - uses: julia-actions/julia-processcoverage@v1
         if: contains(fromJson('["1", "1.9"]'), matrix.version) && matrix.os == 'ubuntu-latest'
-      - uses: codecov/codecov-action@v3
+      - uses: codecov/codecov-action@v4
         if: contains(fromJson('["1", "1.9"]'), matrix.version) && matrix.os == 'ubuntu-latest'
         with:
           file: lcov.info
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,13 @@
 
 See also [github's page](https://github.com/FluxML/Flux.jl/releases) for a complete list of PRs merged before each release.
 
+## v0.14.13
+* New macro `Flux.@layer` which should be used in place of `@functor`.
+  This also adds `show` methods for pretty printing.
+
+## v0.14.12
+* New `SignDecay` optimiser, like `` WeightNorm` but for L1 norm.
+
 ## v0.14.0 (July 2023)
 * Flux now requires julia v1.9 or later.
 * CUDA.jl is not a hard dependency anymore. Support is now provided through the extension mechanism, by loading `using Flux, CUDA`.
@@ -51,6 +58,7 @@ See also [github's page](https://github.com/FluxML/Flux.jl/releases) for a compl
 
 ## v0.13.6
 * Use the package [OneHotArrays.jl](https://github.com/FluxML/OneHotArrays.jl) instead of having the same code here.
+* Added [`@autosize` macro](https://github.com/FluxML/Flux.jl/pull/2078)
 
 ## v0.13.4
 * Added [`PairwiseFusion` layer](https://github.com/FluxML/Flux.jl/pull/1983)
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@ See also [github's page](https://github.com/FluxML/Flux.jl/releases) for a compl
   This also adds `show` methods for pretty printing.
 
 ## v0.14.12
-* New `SignDecay` optimiser, like `` WeightNorm` but for L1 norm.
+* New `SignDecay` optimiser, like `WeightDecay` but for L1 norm.
 
 ## v0.14.0 (July 2023)
 * Flux now requires julia v1.9 or later.
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Flux"
 uuid = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-version = "0.14.12"
+version = "0.14.13"
 
 [deps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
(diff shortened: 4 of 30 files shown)


mcabbott:
Link to version with the markdown rendered: 
https://github.com/isentropic/Flux.jl/blob/improve-freezing-docs/docs/src/models/freezing-params.md

I wonder if this ought to be filed under `docs/tutorials`. It has some overlap with advanced.md (which itself is long overdue for a re-write / removal) which is OK in the "Tutorials" section?

ToucheSir:
It also overlaps with https://fluxml.ai/Flux.jl/stable/training/training/#Freezing-and-Schedules. I do think it's a little disjointed to have docs for both layer definition tools and dedicated freezing tools on a single "freezing" page. Ideally `functor` and `trainable` would be documented alongside everything else on layer definition, but I realize that users may not realize the link between these and freezing.

isentropic:
The reason i put this one all the way in the bottom is cause it requires the reading of functors first and seeing other ways of doing things. It's more like a meta overview of how to do things in each separate case. I tried to link where possible, but Im open to suggestions

isentropic:
> It also overlaps with https://fluxml.ai/Flux.jl/stable/training/training/#Freezing-and-Schedules. I do think it's a little disjointed to have docs for both layer definition tools and dedicated freezing tools on a single "freezing" page. Ideally `functor` and `trainable` would be documented alongside everything else on layer definition, but I appreciate that users may not realize the link between these and freezing. 

I specifically replaced that section, see the file diff

ToucheSir:
> I specifically replaced that section, see the file diff

Yes, I'm aware :)

To explain why I feel this approach is disjointed by way of analogy, this would be like putting `Dropout` and `WeightDecay` on the same page because both are "regularization". Yes, you can technically achieve similar ends with both. But functionally, they are designed for different things. `@functor LayerType (fields...,)` means "I don't want Flux to do _anything_ with the other fields". Overriding `trainable` means "I don't want the optimizer/Optimisers.jl to do anything with the other fields". Using `freeze!` means "I want to dynamically control what parameters are trainable on the fly".

I think it could make sense to have a page talking about the last two, but having all three in one place doesn't make a lot of sense to me. Other than that, I think a "freezing params" page might not have enough content to justify its existence as a standalone page. If the scope were broadened slightly to working with parameters and optimization rules, that could work. Maybe leave some space to add docs on the various ways to do regularization and loss penalties.

isentropic:
The last three pages "logistic regression" "linear regression" and "custom layers" all feature somewhat a disjointed set of functionality and act like a "putting it all together" sections or like extensions of other well-defined use-cases. To me it is fine that these pages are fine to be a mixed-bag because they are disconnected from the rest of the docs and marked as "tutorials" for like more advanced use-cases and clarifications in the docs that are otherwise hard to write in linear independent chunks.

I still insist on having an independent category/page for freezing and everything that might link to it including functors and optimisers. There are examples in the wild like:
1. [lux](https://lux.csail.mit.edu/dev/manual/freezing_model_parameters) features a separate page
2. [flax](https://flax.readthedocs.io/en/latest/guides/training_techniques/transfer_learning.html#) explain how to achieve it using different ways.

I understand your sentiment, it is far from perfect right now. But I still feel like the users who read through the main docs kinda need a page or a section that would explain why and how each method is different in its own way.

I also agree that expanding the scope to include regularization and loss penalties is worth trying. Could you give me more directions here? I could try writing this up.
Thanks for comments and suggestions btw.

mcabbott:
It's not crazy to make a tutorial describing concepts that Flux thinks of as orthogonal, all together. I quite like the examples here.

One thing this could usefully expand on is post-https://github.com/FluxML/Flux.jl/pull/1932 changes... e.g. that `@layer MyLayer trainable=(w,)` is recommended in place of `@functor MyLayer` and `trainable(::MyLayer)`.

* I think the file ought to be in `docs/src/tutorials`. (Some old files like advanced.md are in slightly illogical places pending someone having time to move & redirect.)

* In addition to this file, it might be worth looking for places to add really brief cross-references. E.g. the end of the [training section about L2, weight decay etc](https://github.com/FluxML/Flux.jl/blob/master/docs/src/training/training.md#regularisation) links to `Dropout` briefly.

ToucheSir:
> The last three pages "logistic regression" "linear regression" and "custom layers" all feature somewhat a disjointed set of functionality and act like a "putting it all together" sections or like extensions of other well-defined use-cases. To me it is fine that these pages are fine to be a mixed-bag because they are disconnected from the rest of the docs and marked as "tutorials" for like more advanced use-cases and clarifications in the docs that are otherwise hard to write in linear independent chunks.

Yes, referencing a broad range of functionality works for a tutorial because there's a singular and clear end. This is less true for a "how-to guide" like this proposed freezing page and the two examples mentioned later.

I agree the custom layers page is a mess and needs to be broken up. But separating any mention of  `@functor` from its main responsibility as a layer registration tool would be a bad idea. `struct MyLayer ... end` and `@functor MyLayer` should show up on the same page.

> I still insist on having an independent category/page for freezing and everything that might link to it including functors and optimisers. There are examples in the wild like:
> 
>     1. [lux](https://lux.csail.mit.edu/dev/manual/freezing_model_parameters) features a separate page
> 
>     2. [flax](https://flax.readthedocs.io/en/latest/guides/training_techniques/transfer_learning.html#) explain how to achieve it using different ways.

I would not say they're equivalent. The Lux page really talks about one main piece of functionality, `Lux.freeze`. The equivalent Flux page would talk about `freeze!` and how to combine it with traversal helpers like `fmap`. What the Lux page does not talk about is layer _definition_ tools, which is what `@functor` is. That's left to a [separate page](https://lux.csail.mit.edu/dev/manual/interface).

The Flax page is more of a tutorial. We could use a comparable tutorial in the Flux docs, but that's a separate discussion from a how-to guide on parameter freezing.

> I understand your sentiment, it is far from perfect right now. But I still feel like the users who read through the main docs kinda need a page or a section that would explain why and how each method is different in its own way.

I think the core contention here is that I don't see `@functor Layer (fields...)` as a "method" for freezing parameters. By definition, `@functor` defines what fields are parameters. Anything not in `fields` is then clearly not a parameter, so it doesn't make sense to talk about them in the context of freezing parameters. I think it would be good to mention that one can use `@functor` to exclude certain struct fields from everything (including optimization), with a link to the layer definition section. But this should be a brief mention and not involve moving content from that page into this one. 

> I also agree that expanding the scope to include regularization and loss penalties is worth trying. Could you give me more directions here? I could try writing this up. Thanks for comments and suggestions btw.

I would just broaden the title and preamble for now and worry about adding this extra content in a follow-up.

isentropic:
I tried to incorporate your suggestions in the last commit. Please take a look at the last subsection, and do let me know if it is factually correct

isentropic:
any feedback on this @ToucheSir @mcabbott?

ToucheSir:
Do you mind rebasing so we can see what this looks like on top of https://github.com/FluxML/Flux.jl/pull/2390?

ToucheSir:
Unfortunately I don't think the rebase was successful. Usually you'd have to force push, and only the changes from your four original commits on this PR should be visible on GitHub.

PR #2386, opened 2024-02-29
test log/print sync on BK logs

IanButterworth:
Just checking for a suspected bug in BuildKite log streaming


IanButterworth:
No issue seen

PR #2387, opened 2024-03-02
Restore some support for Tracker.jl

mcabbott:
At present Tracker.jl cannot be used at all with Flux models. This is mostly because TrackedArrays lie about their eltype, but it's not so hard to work around:
```julia
julia> ta = param([1.0 2.0])
Tracked 1×2 Matrix{Float64}:
 1.0  2.0

julia> ta isa AbstractMatrix{Float64}
true

julia> eltype(ta)
TrackedReal{Float64}

julia> ta[1]
1.0 (tracked)
```
Tests are very basic, ideally we would set up some nice way of testing different ADs on all layers once e.g. Enzyme also works... but at least this is something.

Needs https://github.com/FluxML/Tracker.jl/pull/162 

Closes https://github.com/FluxML/Tracker.jl/issues/146


darsnack:
Bump the patch?

darsnack (approved):

PR #2393, opened 2024-03-10
Allow `Parallel(+, f)(x, y, z)` to work like broadcasting, and enable `Chain(identity, Parallel(+, f))(x, y, z)`

mcabbott:
At present Parallel allows multiple layers and one input, but not the reverse. This PR extends it to allow both ways... much like broadcasting in `connection((inputs .|> layers)...)`.
```julia
julia> Parallel(+, inv)(1, 2, 3)  # was an error
1.8333333333333333

julia> (1,2,3) .|> (inv,)
(1.0, 0.5, 0.3333333333333333)
```
Does this have any unintended side-effects?

### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -475,8 +475,11 @@ end
 Create a layer which passes an input array to each path in
 `layers`, before reducing the output with `connection`.
 
-Called with one input `x`, this is equivalent to `connection([l(x) for l in layers]...)`.
-If called with multiple inputs, one is passed to each layer, thus `Parallel(+, f, g)(x, y) = f(x) + g(y)`.
+Obeys the similar rules to broadcasting:
+* Called with one input `x`, this is equivalent to `connection([l(x) for l in layers]...)`.
+* With multiple `inputs` and just one layer, it is instead `connection([layer(x) for x in inputs]...)`.
+* With multiple inputs and multiple layers, one input is passed to each layer,
+  thus `Parallel(+, f, g)(x, y) = f(x) + g(y)`.
 
 Like [`Chain`](@ref), its sub-layers may be given names using the keyword constructor.
 These can be accessed by indexing: `m[1] == m[:name]` is the first layer.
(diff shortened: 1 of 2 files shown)


src/layers/basic.jl
```diff
@@ -524,23 +527,32 @@ end
...
 function (m::Parallel)(xs...)
-  _parallel_check(m.layers, xs)
-  m.connection(map(|>, xs, Tuple(m.layers))...)
+  if length(m.layers) == 1
+    f = only(m.layers)
+    m.connection(map(x -> f(x), xs)...)  # multiple arguments, one layer
+  else
+    _parallel_check(m.layers, xs)
+    m.connection(map(|>, xs, Tuple(m.layers))...)  # multiple arguments & multiple layers
+  end
```
ToucheSir, line 550: The relevant discussions are https://github.com/FluxML/Flux.jl/pull/1698#issuecomment-905030618 and https://github.com/FluxML/Flux.jl/issues/1685, I believe. As I recall the single layer + varargs combo here doesn't cause any ambiguity issues (not in the MethodError sense), so it's probably above board.

One practical concern is that Zygote will not like the branch. That could be resolved by using dispatch instead.

mcabbott: Had forgotten how wild those discussions were. But perhaps there is something there to think through.

Surely it can be implemented with dispatch, but my first attempt needed a slew of methods to solve ambiguities, hence the branch.

mcabbott:
Here's the complete run-down on where Flux does & doesn't splat at present:
```julia
julia> using Flux

julia> pr(x) = begin println("arg: ", x); x end;

julia> pr(x...) = begin println(length(x), " args: ", join(x, " & "), " -> tuple"); x end;

julia> c1 = Chain(pr, pr); ########## simple chain

julia> c1(1)
arg: 1
arg: 1
1

julia> c1((1, 2))
arg: (1, 2)
arg: (1, 2)
(1, 2)

julia> c1(1, 2)
ERROR: MethodError:
Closest candidates are:
  (::Chain)(::Any)

julia> p1 = Parallel(pr, a=pr);  ########## combiner + one layer

julia> p1(1)
arg: 1
arg: 1
1

julia> p1((1, 2))  # one 2-Tuple is NOT accepted, always splatted  --> changed by PR
ERROR: ArgumentError: Parallel with 1 sub-layers can take one input or 1 inputs, but got 2 inputs

julia> p1(1, 2)  # more obvious error  --> changed by PR
ERROR: ArgumentError: Parallel with 1 sub-layers can take one input or 1 inputs, but got 2 inputs

julia> p1((a=1, b=2))  # one NamedTuple is ok
arg: (a = 1, b = 2)
arg: (a = 1, b = 2)
(a = 1, b = 2)

julia> p1((((1,),),))  # splatted many times
arg: 1
arg: 1
1

julia> p2 = Parallel(pr, a=pr, b=pr);  ########## combiner + two layers

julia> p2(1)  # one non-tuple arg is broadcasted
arg: 1
arg: 1
2 args: 1 & 1 -> tuple
(1, 1)

julia> p2(1, 2)  # 2 args sent to 2 layers
arg: 1
arg: 2
2 args: 1 & 2 -> tuple
(1, 2)

julia> p2((1, 2))  # one tuple splatted
arg: 1
arg: 2
2 args: 1 & 2 -> tuple
(1, 2)

julia> p2((a=1, b=2))  # one NamedTuple sent to both
arg: (a = 1, b = 2)
arg: (a = 1, b = 2)
2 args: (a = 1, b = 2) & (a = 1, b = 2) -> tuple
((a = 1, b = 2), (a = 1, b = 2))

julia> p2(((1,2), ((3,4),)))  # only splatted once
arg: (1, 2)
arg: ((3, 4),)
2 args: (1, 2) & ((3, 4),) -> tuple
((1, 2), ((3, 4),))

julia> Chain(pr, p2, pr)((1, 2))  # here earlier layers cannot pass p2 two arguments
arg: (1, 2)
arg: 1
arg: 2
2 args: 1 & 2 -> tuple
arg: (1, 2)
(1, 2)
```

This PR changes the two error cases above:
```julia
julia> p1((1, 2))  # changed by PR
arg: 1
arg: 2
2 args: 1 & 2 -> tuple
(1, 2)

julia> p1(1, 2)  # changed by PR
arg: 1
arg: 2
2 args: 1 & 2 -> tuple
(1, 2)
```

You could argue that `p1((1, 2))` already has a plausible meaning, apply one layer to one input Tuple. But this use of Parallel is really just Chain (or in this order, `∘`). And it's an error at present.

I think `p1(1, 2)` has no other plausible meaning. 

The rule after this PR is:
1. `(p::Paralel)(input::Tuple)` always splats to `p(input...)`
2. return `combine((inputs .|> layers)...)`

Step 1 is unchanged, but step 2 previously allowed only broadcasting of the input. And today, I have a use where I want to broadcast the layer instead (easier than sharing it).

mcabbott:
Reading old threads... around here https://github.com/FluxML/Flux.jl/pull/2101#issuecomment-1306061980 it was agreed that adding `(c::Chain)(xs...) = c(xs)` would make sense, but there was never a PR.

That's the first MethodError in my list above. I would like this too, and perhaps should just add it here.

ToucheSir (approved):
Since it was linked here, can you quickly comment on the relationship between this and https://github.com/FluxML/Functors.jl/pull/80?

PR #2397, opened 2024-03-13
freezingdocs

isentropic:
Rewrite of the prev pull  https://github.com/FluxML/Flux.jl/pull/2385 , god i hate git
@ToucheSir

### DIFF
--- a/docs/src/tutorials/misc-model-tweaking.md
+++ b/docs/src/tutorials/misc-model-tweaking.md
@@ -1,0 +1,131 @@
+# Choosing differentiable/gpu parts of the model
+!!! note 
+    This tutorial features somewhat disconnected topics about customizing your
+    models even further. It is advised to be familiar with
+    [`Flux.@layer`](@ref), [`Flux.@functor`](@ref), [`freeze!`](@ref
+    Flux.freeze!) and other basics of Flux.
+
+Flux provides several ways of freezing, excluding from backprop entirely and
+marking custom struct fields not to be moved to the GPU
+([Functors.@functor](@ref)) hence excluded from being trained. The following
+subsections should make it clear which one suits your needs the best.
+
+## On-the-fly freezing per model instance
+Perhaps you'd like to freeze some of the weights of the model (even at
+mid-training), and Flux accomplishes this through [`freeze!`](@ref Flux.freeze!) and `thaw!`.
+
+```julia
+m = Chain(
+      Dense(784 => 64, relu), # freeze this one
+      Dense(64 => 64, relu),
+      Dense(32 => 10)
+    )
+opt_state = Flux.setup(Momentum(), m);
+
+# Freeze some layers right away
+Flux.freeze!(opt_state.layers[1])
+
+for data in train_set
+    input, label = data
+
+    # Some params could be frozen during the training:
+    Flux.freeze!(opt_state.layers[2])
+
+    grads = Flux.gradient(m) do m
+        result = m(input)
+        loss(result, label)
+    end
+    Flux.update!(opt_state, m, grads[1])
+
+    # Optionally unfreeze the params later
+    Flux.thaw!(opt_state.layers[1])
+end
+```
+
+## Static freezing per model definition
+Sometimes some parts of the model ([`Flux.@layer`](@ref)) needn't to be trained at all but these params
+still need to reside on the GPU (these params are still needed in the forward
+and/or backward pass).
+```julia
+struct MaskedLayer{T}
+    chain::Chain
+    mask::T
+end
+Flux.@layer MyLayer trainable=(chain,)
+# mask field will not be updated in the training loop
+
+function (m::MaskedLayer)(x)
+    # mask field will still move to to gpu for efficient operations:
+  return m.chain(x) + x + m.mask
+end
+
+model = MaskedLayer(...) # this model will not have the `mask` field trained
+```
+Note how this method permanently sets some model fields to be excluded from
+training without on-the-fly changing.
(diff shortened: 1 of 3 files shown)


docs/src/tutorials/misc-model-tweaking.md
```diff
@@ -0,0 +1,131 @@
...
+function (m::CustomLayer)(x)
+    result = m.chain(x) + x
+    
+    # `activation_results` are not part of the GPU loop, hence we could do
+    # things like `push!`
+    push!(m.activation_results, mean(result))
+    return result
+end
+```
+See more about this in [`Flux.@functor`](@ref)
```
ToucheSir, line 89: I think this is great. I also wonder if it'd belong better in `advanced.md`. This docs page can then have a big, prominent link to the relevant sections in that file with a one or two-sentence explanation to look there for methods of including and excluding parts of the model at definition time.

To elaborate, I still think there is utility in separating out concepts and tools which apply at definition time vs those which are dynamic and apply at "runtime". That's not to say they can't be mentioned on the same page, but they should be _introduced_ separately. One way to bridge the gap might be to retain reduced code snippets on this page alongside the link(s) I mentioned earlier. e.g.

```julia
# struct CustomLayer chain; activation_results; end
Flux.@functor CustomLayer (chain, ) # Explicitly leaving out `activation_results`
```

isentropic:
@ToucheSir @mcabbott  sorry for the time delay i got carried away. I updated the docs as you requested. Let me know your new comments. Hopefully this section could help with some of the frustrations I had for the new users

ToucheSir:
Thanks for the update. I'd say my comment at https://github.com/FluxML/Flux.jl/pull/2397#discussion_r1525712174 still applies, but if you feel strongly about it I'd recommend leaving that section out for now and we can look at merging the rest.

isentropic:
Oh I meant to incorporate your comment, maybe I couldn't understand completely what you meant. Your big "prominent links" part means like links in the subheadings directly?

- I feel like I did the ”move to advanced" part of your comments 

Could you just state more clearly what you suggested, I'll be happy to incorporate

ToucheSir:
I think what happened is a couple of file names changed, so what should've been a clear ask became slightly more confusion. Just to put it in words though, the ask is to move the content under the sections "Static freezing per model definition" and "Excluding from model definition" to https://github.com/FluxML/Flux.jl/blob/master/docs/src/guide/models/custom_layers.md. That way, we co-locate code snippets that talk about model definition instead of making users look across multiple pages for them.

To ensure users know that this examples are also useful for excluding parameters from training, the freezing docs page can then link to the two headings under `custom_layers.md`. Kind of a "see also" with maybe a couple of sentences explaining why the linked sections are relevant.

isentropic:
I tried to do this the way you suggested but it feels even more haphazard this way. `custom_layers.md` don't have any models that are defined via `struct ... end` all the models there are chains and other blocks provided by Flux, so I cannot reuse any definitions. I really tried to somehow merge it, but its just bad I feel.

Another reason is that now the file i'm contributed 'misc-model-tweaking' is partitioned again, and there is no single place to understand about the differences of various ways to freeze. The whole point seems gone, because I tried to summarize this already spread apart information into a single page, and seems like your suggestions are to split it again.

I suggest to merge it as is. Addition of a new page shouldn't hurt that much because it is very focused and specific to model freezing.

ToucheSir:
> `custom_layers.md` don't have any models that are defined via `struct ... end` all the models there are chains and other blocks provided by Flux, so I cannot reuse any definitions.

That doesn't sound like https://fluxml.ai/Flux.jl/stable/guide/models/custom_layers/. Every section on that page has an example layer defined with `struct ... end`.


> Another reason is that now the file i'm contributed 'misc-model-tweaking' is partitioned again, and there is no single place to understand about the differences of various ways to freeze.

The fundamental tension here is about what "freezing" means. In my mind, freezing parameters of a layer or model implies the ability to "thaw" those parameters later. Dynamic freezing using `freeze!` and `thaw!` does this, for example. This is also in line with the language and expectations around freezing in [PyTorch](https://discuss.pytorch.org/t/how-the-pytorch-freeze-network-in-some-layers-only-the-rest-of-the-training/7088) and [JAX](https://github.com/google/flax/discussions/1453) (which links to https://optax.readthedocs.io/en/latest/api/optimizer_wrappers.html#masked-update).

In contrast, excluding model fields from training is treated differently. In this case, the non-trainable-ness or non-parameter-ness (depending on the library) of an array is an inherent property of it. One can not thaw it, nor is it stopped from updating (as "frozen" might imply) because it can still be mutable&mdash;think BatchNorm running stats. It's illustrative that the APIs libraries use here are distinctly named and documented from the ones they have for freezing for training. See for example PyTorch's [`register_buffer`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html) or the [parameter vs state dichotomy](https://flax.readthedocs.io/en/latest/guides/flax_fundamentals/state_params.html) in Flax. Flux follows a similar philosophy with `Flux.@layer MyLayer trainable=(fields...,)`.

isentropic:
I honestly tried to do it your way, but I hated what it looked like at the end. I just thought having one more of docs is okay as long as it's all related + you can delete this one page later with minimal effort. 

I am in the shoes of the learner of the library and to the I made the PR feels cleaner and better, if you feel otherwise please change it the way you deem appropriate. 

You can take the commits partially or as a whole or close the PR all-together. At this point I don't wanna argue about a few written paragraphs anymore.

CarloLucibello:

PR #2400, opened 2024-03-18
Use LazyString in depwarn

### LINKED ISSUE #2399: 2x performance regression due to 5e80211c3302b5e7b79b4f670498f5a68af6659b
I noticed a 2x performance regression in one of my projects when updating from 0.14.12 to 0.14.13 and git-bisect points to 5e80211c3302b5e7b79b4f670498f5a68af6659b

The project in which I noticed is difficult to share but doesn't do crazy things; basically training small MLPs (8 inputs, 3 hidden layers, hidden dimension 64, 10 outputs) with `tanh` activation.

mcabbott:
Closes #2399

Maybe there are a few other cases we should do this?

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -196,8 +196,8 @@ end
 # Greek-letter keywords deprecated in Flux 0.13
 # Arguments (old => new, :function, "β" => "beta")
 function _greek_ascii_depwarn(βbeta::Pair, func = :loss, names = "" => "")
-  Base.depwarn("""function $func no longer accepts greek-letter keyword $(names.first)
-    please use ascii $(names.second) instead""", func)
+  Base.depwarn(LazyString("function $func no longer accepts greek-letter keyword ", names.first, """
+    please use ascii """, names.second, " instead"), func)
   βbeta.first
 end
 _greek_ascii_depwarn(βbeta::Pair{Nothing}, _...) = βbeta.second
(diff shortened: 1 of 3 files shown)


src/layers/macro.jl
```diff
@@ -89,7 +89,7 @@ end
 
 function _check_new_macro(x::T) where T
   Functors.isleaf(x) && return
-  Base.depwarn("This type should probably now use `Flux.@layer` instead of `@functor`: $T", Symbol("@functor"))
+  Base.depwarn(LazyString("This type should probably now use `Flux.@layer` instead of `@functor`: " T), Symbol("@functor"))
```
lassepe, line 92: Missing comma

lassepe, line 92:
```suggestion
  Base.depwarn(LazyString("This type should probably now use `Flux.@layer` instead of `@functor`:, " T), Symbol("@functor"))
```

src/deprecations.jl
```diff
@@ -196,8 +196,8 @@ end
 # Greek-letter keywords deprecated in Flux 0.13
 # Arguments (old => new, :function, "β" => "beta")
 function _greek_ascii_depwarn(βbeta::Pair, func = :loss, names = "" => "")
-  Base.depwarn("""function $func no longer accepts greek-letter keyword $(names.first)
-    please use ascii $(names.second) instead""", func)
+  Base.depwarn(LazyString("function $func no longer accepts greek-letter keyword ", names.first, """
```
mcabbott, line 199:
```suggestion
  Base.depwarn(LazyString("function ", func, " no longer accepts greek-letter keyword ", names.first, """
```

CarloLucibello (approved):

PR #2403, opened 2024-03-19
Fix dead link in docs

BioTurboNick:
Pointed to a README file in `dev` that doesn't exist.


mcabbott (approved):
Thanks. 

Sadly this needs updating, `opt.eta = next!(schedule)` will only work for old-style implicit optimisers... needs to become something like `Flux.adjust!(opt_state, eta = next!(schedule))` for new-style ones.

PR #2404, opened 2024-03-19
Improve errors for conv layers

### LINKED ISSUE #2398: Dimensions check for `Conv` is incomplete, leading to confusing error
### Motivation and description

This is the sort error message produced when a 3d array is passed to `Conv` which expects WHCD.

```
ERROR: DimensionMismatch: layer Conv((5, 5), 1 => 16, relu, pad=2) expects size(input, 2) == 1, but got 64×64×32 CUDA.CuArray{Float32, 3, CUDA.Mem.DeviceBuffer}
Stacktrace:
  [1] _size_check(layer::Flux.Conv{…}, x::CUDA.CuArray{…}, ::Pair{…})
    @ Flux C:\Users\nicho\.julia\packages\Flux\qZO6T\src\layers\basic.jl:195
```

The intended error is that the second to last dimension, which is the channel dimension, is expected to be 1 to match the input channels for the layer, but what it actually says is that the second dimension of the input is expected to be 1, and it's just not clear why that is unless you already know how it should work.

The check function currently just ensures that the input array is at least 2 dimensions. It should really check that the input array matches the kernel dimensions + 2, no? And this may be naive, but it looks like the weights array dimensions matches the expected input dimensions, and could be used for this purpose?

### Possible Implementation

_No response_

mcabbott:
Fixes #2398

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -193,7 +193,7 @@ function _size_check(layer, x::AbstractArray, (d, n)::Pair)
   d > 0 || throw(DimensionMismatch(string("layer ", layer,
     " expects ndims(input) > ", ndims(x)-d, ", but got ", summary(x))))
   size(x, d) == n || throw(DimensionMismatch(string("layer ", layer,
-    " expects size(input, $d) == $n, but got ", summary(x))))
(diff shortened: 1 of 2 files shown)


ToucheSir (approved):
We have new CI failures on master, but none of them look related

PR #2409, opened 2024-03-24
Add repeat vector

Jafagervik:
RepeatVector Layer like in [keras](https://keras.io/api/layers/reshaping_layers/repeat_vector/).

PS: I'm still new to contributing here, so if "trivial layers" like these are not requested, and rather something each user can implement on their own, then please let me know 

### PR Checklist

- [X] Tests are added
- [X] Entry in NEWS.md

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -864,3 +864,33 @@ EmbeddingBag(weight::AbstractMatrix) = EmbeddingBag(weight, mean)
 function Base.show(io::IO, m::EmbeddingBag)
   print(io, "EmbeddingBag(", size(m.weight, 2), " => ", size(m.weight, 1), ")")
 end
+
+
+"""
+    RepeatVector(n::Int)
+
+Repeat the input `n` times along the last dimension.
+
+# Examples
+```jldoctest
+julia> rv = RepeatVector(3)
+julia> rv([1, 2, 3])
+3×3 Matrix{Int64}:
+ 1  1  1
+ 2  2  2
+ 3  3  3
+```
+"""
+struct RepeatVector
+  n::Int
+end
+
+@layer RepeatVector
+
+function (rv::RepeatVector)(x::AbstractArray{T}) where {T}
+  expanded = reshape(x, (size(x)..., 1))
+  repeated = repeat(expanded, outer = (1, rv.n, 1))
+  return repeated
+end
+
+Base.show(io::IO, rv::RepeatVector) = print(io, "RepeatVector($(rv.n))")
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 See also [github's page](https://github.com/FluxML/Flux.jl/releases) for a complete list of PRs merged before each release.
 
+## v0.14.14
+* New layer `RepeatVector` which works like 
+  RepeatVector in keras
+
 ## v0.14.13
 * New macro `Flux.@layer` which should be used in place of `@functor`.
   This also adds `show` methods for pretty printing.


src/layers/basic.jl
```diff
@@ -864,3 +864,33 @@ EmbeddingBag(weight::AbstractMatrix) = EmbeddingBag(weight, mean)
...
+struct RepeatVector
+  n::Int
+end
+
+@layer RepeatVector
+
+function (rv::RepeatVector)(x::AbstractArray{T}) where {T}
+  expanded = reshape(x, (size(x)..., 1))
+  repeated = repeat(expanded, outer = (1, rv.n, 1))
+  return repeated
```
CarloLucibello, line 893: this could just be
```julia
return repeat(x, size(x)..., rv.n)
```

src/layers/basic.jl
```diff
@@ -864,3 +864,33 @@ EmbeddingBag(weight::AbstractMatrix) = EmbeddingBag(weight, mean)
 function Base.show(io::IO, m::EmbeddingBag)
   print(io, "EmbeddingBag(", size(m.weight, 2), " => ", size(m.weight, 1), ")")
 end
+
+
+"""
+    RepeatVector(n::Int)
+
+Repeat the input `n` times along the last dimension.
```
CarloLucibello, line 872: this should specify that an extra dimension is added first.

src/layers/basic.jl
```diff
@@ -864,3 +864,33 @@ EmbeddingBag(weight::AbstractMatrix) = EmbeddingBag(weight, mean)
 function Base.show(io::IO, m::EmbeddingBag)
   print(io, "EmbeddingBag(", size(m.weight, 2), " => ", size(m.weight, 1), ")")
 end
+
+
+"""
+    RepeatVector(n::Int)
```
CarloLucibello, line 870: `RepeatArray` is maybe a better name.

src/layers/basic.jl
```diff
@@ -864,3 +864,33 @@ EmbeddingBag(weight::AbstractMatrix) = EmbeddingBag(weight, mean)
...
+
+Repeat the input `n` times along the last dimension.
+
+# Examples
+```jldoctest
+julia> rv = RepeatVector(3)
```
CarloLucibello, line 876: in a jldoctest block each line should also have the corresponding output, otherwise the doctests will fail

CarloLucibello:
I would be ok with adding a layer like this, although I don't like the naming and I can't think of a really good one. 
Test should be added in the `test/` folder and the layer should be referenced in `docs/`

mcabbott:
What is this used for?

And what is the intended behaviour, on arrays of various dimensions? I'm not sure the code matches the words, nor the example given.
```julia
julia> RepeatVector(2)(rand(3)) |> size
(3, 2, 1)

julia> RepeatVector(2)(rand(3,5)) |> size
(3, 10, 1)

julia> RepeatVector(2)(rand(3,5,7)) |> size
(3, 10, 7, 1)

julia> rv = RepeatVector(3);

julia> rv([1, 2, 3])  # example says 3×3 Matrix
3×3×1 Array{Int64, 3}:
```

src/layers/basic.jl
mcabbott: That would do this:
```julia
julia> RepeatVector(2)(rand(3)) |> size
(9, 2)

julia> RepeatVector(2)(rand(3,5)) |> size
(9, 25, 2)

julia> RepeatVector(2)(rand(3,5,7)) |> size
(9, 25, 49, 2)
```
Perhaps you meant:
```julia
julia> function (rv::RepeatVector)(x::AbstractArray{T}) where {T}
          repeated = repeat(x, ntuple(_->1, ndims(x))..., rv.n)
       end

julia> RepeatVector(2)(rand(3)) |> size
(3, 2)

julia> RepeatVector(2)(rand(3,5)) |> size
(3, 5, 2)

julia> RepeatVector(2)(rand(3,5,7)) |> size
(3, 5, 7, 2)
```

ToucheSir:
The lack of clarity on how this layer should behave suggest to me that we might not want it as a built-in. Now, the good news is that Flux doesn't require a custom layer type for everything. In this case, it suffices to use a plain function/closure:
```julia
function repeatvector(x::AbstractArray, n::Int)
  expanded = reshape(x, (size(x)..., 1))
  repeated = repeat(expanded, outer = (1, rv.n, 1))
  return repeated
end
repeatvector(n::Int) = x -> repeatvector(x, n)
f = repeatvector(2)
# or, using built-in types:
f = Base.Fix2(repeatvector, n)
```

darsnack:
Going along Brian's point, even if we were to include this, I think it ought to be a function not a struct. Though I tend to lean towards not including it.

CarloLucibello:

PR #2415, opened 2024-03-27
Make sure first example in Custom Layers docs uses type parameter

BioTurboNick:


docs/src/models/advanced.md
```diff
@@ -7,8 +7,8 @@ Here we will try and describe usage of some more advanced features that Flux pro
 Here is a basic example of a custom model. It simply adds the input to the result from the neural network.
 
 ```julia
-struct CustomModel
-  chain::Chain
+struct CustomModel{T <: Chain} # Parameter to avoid type instability (see "Multiple inputs" section below)
```
ToucheSir, line 10: Mayhaps we should move the comment in that section up to here? I don't see any drawback to having it earlier, and we could even drop a link to the Julia performance tips alongside it.

CarloLucibello (approved):

Issue #2414, opened 2024-03-27
Significant time spent moving medium-size arrays to GPU, type instability


BioTurboNick:
There are occasions where `@profview` shows seemingly inordinate amount of time spent moving data to the GPU given the array size, and possibly excessive GPU memory usage? Not sure what I should be expecting. Could be related to having two array outputs rather than one? I've also see type instability reported through `@code_warntype` and Cthulhu that I'm not sure how to resolve.

Using a toy example to show the effect:

```julia
]activate --temp
]add cuDNN, CUDA, Flux

using CUDA, cuDNN, Flux, Statistics

struct Split{T1 <: Dense, T2 <: Dense}
    s1::T1
    s2::T2
    max_sources::Int
end

function Split(feature_dim::Int, max_sources::Int)
    Split(
            Dense(feature_dim => 2 * max_sources),
            Dense(feature_dim => max_sources),
            max_sources)
end

Flux.@layer Split

(m::Split)(input) = reshape(m.s1(input), 2, m.max_sources, :), m.s2(input)

function imagegen_test(batch_size)
    return randn(Float32, 2048, batch_size), (randn(Float32, 2, 20, batch_size), randn(Float32, 20, batch_size))
end

function test()
    training_batch_size = 256
    iters_per_eval = 64
    network = Split(2048, 128) |> gpu
    optimizer = Flux.Optimise.Adam(1E-4)
    opt_state = Flux.setup(optimizer, network)
    kernel_sigmas_gpu = Float32[64.0, 320.0, 640.0, 1920.0] |> gpu
    for i ∈ 1:iters_per_eval
        training_data = imagegen_test(training_batch_size) |> gpu
        Flux.train!(network, (training_data,), opt_state) do m, x, y
            θ_pred, intensity_pred = m(x)
            loss_func(θ_pred, intensity_pred, y..., kernel_sigmas_gpu)
        end
    end

    return nothing
end

pairwise_cityblock(c) =
    dropdims(sum((Flux.unsqueeze(c, 2) .- Flux.unsqueeze(c, 3)) .|> abs, dims = 1), dims = 1)

function kernel_loss(K, predicted_weights, target_weights)
    weights = [predicted_weights; -target_weights]
    embedding_loss = batched_vec(Flux.unsqueeze(weights, 1), batched_vec(K, weights))
    return dropdims(embedding_loss, dims = 1)
end

[...]

PR #2416, opened 2024-03-29
Add GPU GC comment to Performance Tips

BioTurboNick:
To help people avoid #2414 , added some commentary about GPU memory management to the performance tips section.


docs/src/performance.md
```diff
@@ -74,3 +74,7 @@ end
 
 When doing this kind of concatenation use `reduce(hcat, xs)` rather than `hcat(xs...)`.
 This will avoid the splatting penalty, and will hit the optimised `reduce` method.
+
+## Be aware of GPU memory inefficiencies
+
+Currently, GPU memory is not handled as well as system memory. If your training loop is allocating significantly on the GPU, you can quickly fill your GPU memory and the piecemeal reclamation and shuffling of data between GPU and system memory can become extremely slow. If profiling shows that a significant portion of time is spent in the `gpu` function and your data sizes are not large, this may be the cause. Running an incrememtal garbage collection manually (`GC.gc(false)`) at regular intervals can keep your GPU memory free and responsive. See other tips for CUDA memory management [here](https://cuda.juliagpu.org/stable/usage/memory/).
```
CarloLucibello, line 80:
```suggestion
Currently, GPU memory is not handled as well as system memory. If your training loop is allocating significantly on the GPU, you can quickly fill your GPU memory and the piecemeal reclamation and shuffling of data between GPU and system memory can become extremely slow. If profiling shows that a significant portion of time is spent in the `gpu` function and your data sizes are not large, this may be the cause. Running an incremental garbage collection manually (`GC.gc(false)`) at regular intervals can keep your GPU memory free and responsive. See other tips for CUDA memory management [here](https://cuda.juliagpu.org/stable/usage/memory/).
```

CarloLucibello (approved):

Issue #2096, opened 2022-11-01
gradient of `Flux.normalise` return NaN when `std` is zero


chengchingwen:
$\epsilon$ argument of `Flux.normalise` only prevent the forward value from division by zero, but there is also an division by $\sigma$ in [pullback of std](https://github.com/JuliaDiff/ChainRules.jl/blob/main/src/rulesets/Statistics/statistics.jl#L52). We might need a custom rrule for `Flux.normalise`.

```julia
julia> Zygote.gradient(x->sum(sin.(Flux.normalise(x;dims=1))), ones(3,3))
([NaN NaN NaN; NaN NaN NaN; NaN NaN NaN],)
```

mcabbott:
Xref https://github.com/JuliaML/MLUtils.jl/issues/123 (about moving & renaming) and #1992 (about NaN from batch of 1).

ToucheSir:
Other frameworks have implemented this using sqrt + var + the eps instead of using std directly.

chengchingwen:
> and #1992 (about NaN from batch of 1).

Not sure if #1992 is related. Batchnorm don't use `normalise` and this issue is not caused by batch size.

> Other frameworks have implemented this using sqrt + var + the eps instead of using std directly.

FWIW, PyTorch [layernorm add the eps to var](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cuda/layer_norm_kernel.cu#L234) and then [store the var with eps](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cuda/layer_norm_kernel.cu#L266) and [use it directly in the pullback](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/cuda/layer_norm_kernel.cu#L385). 
One quick and dirty solution can be inplace updating the `std` value with eps (with AD ignoring it).

This also brings up an issue about the error between real value and the value with eps. Since we are dividing by $(\sqrt{var} + \epsilon)$ and they are dividing by $(\sqrt{var + \epsilon})$, the resulting value could have $\epsilon$ times difference. I'm not sure which is better (IMPO it should be handled by a branch)

ToucheSir:
I think we'd have to go with $\sqrt{var + \epsilon}$ because the rule for `sqrt(x)` divides by `2*sqrt(x)`.

chengchingwen:
[...]

PR #2421, opened 2024-04-07
Epsilon change in normalise for stability

billera:
Normalise allows for an optional epsilon term aimed towards improving numerical stability. Previously the epsilon was added after computing the standard deviation of the input. The standard deviation computation involves a square root, leading to NaN's in gradients dependent on normalise when the variance is very low, and for instance LayerNorms applied to low variance inputs will result in NaN gradients. By first computing the variance and taking the square root after adding epsilon^2 (squaring to preserve scale), we prevent NaN's in gradients at low variance. See the following example with LayerNorm in the current patch.

```Julia
using Flux 
using Zygote 

ln = LayerNorm(256; eps = 1f-3)
for i in 1:10 
    x = ones(Float32, 256) .+ randn(Float32, 256) .* 10f0^(-i)
    l, gs = Zygote.withjacobian(ln, x)
    @show maximum(gs[1])
end


>>> maximum(gs[1]) = 9.44178f0
>>> maximum(gs[1]) = 95.85736f0
>>> maximum(gs[1]) = 477.4946f0
>>> maximum(gs[1]) = 910.05457f0
>>> maximum(gs[1]) = 985.8402f0
>>> maximum(gs[1]) = 995.0282f0
>>> maximum(gs[1]) = 995.9835f0
>>> maximum(gs[1]) = NaN32
>>> maximum(gs[1]) = NaN32
>>> maximum(gs[1]) = NaN32

```

We observe that while the gradients are fixed at low variance due to the epsilon addition in the denominator, this does prevent NaN's, due to the non-padded square root in the std computation. But, when using the updated normalise, these NaN's dissapear,

```Julia
>>> maximum(gs[1]) = 9.531697f0
>>> maximum(gs[1]) = 105.468056f0
>>> maximum(gs[1]) = 674.7051f0
>>> maximum(gs[1]) = 991.67163f0
>>> maximum(gs[1]) = 996.03973f0
>>> maximum(gs[1]) = 996.09314f0
>>> maximum(gs[1]) = 996.0937f0
>>> maximum(gs[1]) = 996.0937f0
>>> maximum(gs[1]) = 996.0937f0
>>> maximum(gs[1]) = 996.0937f0
```
and remain fixed to the implicitly capped value. A simple test verifying this computation's equivalence with the previous one (modulo the differences at very low standard deviations) could be added if desired.

### DIFF
--- a/src/layers/stateless.jl
+++ b/src/layers/stateless.jl
@@ -4,7 +4,7 @@
 
 Normalise `x` to mean 0 and standard deviation 1 across the dimension(s) given by `dims`.
 Per default, `dims` is the last dimension. 
-`eps` is a small term added to the denominator for numerical stability.
+`eps` is a small term added to the denominator/square root for numerical stability.
 
 # Examples
 ```jldoctest
@@ -37,8 +37,8 @@ true
 @inline function normalise(x::AbstractArray; dims=ndims(x), eps=ofeltype(x, 1e-5), ϵ=nothing)
   ε = _greek_ascii_depwarn(ϵ => eps, :InstanceNorm, "ϵ" => "eps")
   μ = mean(x, dims=dims)
-  σ = std(x, dims=dims, mean=μ, corrected=false)
-  return @. (x - μ) / (σ + ε)
+  σ² = var(x, dims=dims, mean=μ, corrected=false)
+  return @. (x - μ) / sqrt(σ² + ε^2)
 end
 
 """


src/layers/stateless.jl
```diff
@@ -4,7 +4,7 @@
 
 Normalise `x` to mean 0 and standard deviation 1 across the dimension(s) given by `dims`.
 Per default, `dims` is the last dimension. 
-`eps` is a small term added to the denominator for numerical stability.
+`eps` is a small term added to the denominator/square root for numerical stability.
```
CarloLucibello, line 7:
```suggestion
`eps` is a small term added to the variance for numerical stability.
```

CarloLucibello:
I agree with this change and pytorch does the same thing. It should considered a breaking change thous, so let's wait for when we are near v01.5 before merging.

mcabbott:
Can this have a test with input which triggers the NaN behaviour before?

Ideally testing not just the function, but also LayerNorm, maybe BatchNorm, anything which uses this internally. Then if the implementation of these layers finally gets replaced, it will be harder to lose the change.

ToucheSir:
Putting a backlink to https://github.com/FluxML/Flux.jl/issues/2096 because this work should close that.

src/layers/stateless.jl
```diff
@@ -34,10 +34,11 @@ julia> isapprox(std(y; dims=1, corrected=false), ones(1, 10), atol=1e-5)
 true
 ```
 """
-@inline function normalise(x::AbstractArray; dims=ndims(x), eps=ofeltype(x, 1e-5))
+@inline function normalise(x::AbstractArray; dims=ndims(x), eps=1f-5)
```
mcabbott, line 37: Why does this now assume Float32? Elsewhere we try to allow for Float16 too.

mcabbott (commented):
Needs news entry?

PR #2412, opened 2024-03-26
removing old deprecations and Optimise module

CarloLucibello:
Last part of #1986

PR #2422, opened 2024-04-07
documentation update

CarloLucibello:
I factored out here the part of #2412 involving only the documentation. Most of it is removing any reference to `Flux.params` and to the implicit differentiation style.

I also reorganized a bit files and folders, updated and reintroduced the old mlp tutorial, and update to the last Documenter version.

Needs https://github.com/FluxML/Optimisers.jl/pull/174 to be merged and tagged.


mcabbott:
I'm against restoring the tutorial in this form. All this text is not very helpful. You can't easily play with pieces of this to see how they work.

The model zoo has a re-write, and it could plausibly be tidied up to live here too:

https://github.com/FluxML/model-zoo/blob/master/vision/mlp_mnist/mlp_mnist.jl

I agree it would be nice if something like this were easier to find. The problem we never solved with moving model zoo models in here was that they aren't tested and tend to rot, so this means two things to check manually.

CarloLucibello:
I put away the MLP tutorial (along with an updated 60min-blitz) and created a "Model Zoo" page instead in the tutorial hopefully helping a bit the discoverability of the model zoo examples.

CarloLucibello:
merge?

mcabbott:
Would like to look over the rest but not today, sorry

Issue #1606, opened 2021-05-31
Allow BatchNorm training on CUDA with `track_stats=false`


ToucheSir:
Gathered from https://discourse.julialang.org/t/batchnorm-only-track-stats-true-supported-on-gpu/62091.

This would most likely require changes in NNlibCUDA as well. I'm not sure how interchangeable the various `cudnnBatchNormalizationForward*` functions are, so putting a pin in this until someone more knowledgeable can comment.

CarloLucibello:
Looking at [cudnn docs](https://docs.nvidia.com/deeplearning/cudnn/api/index.html#cudnnBatchNormalizationForwardTraining)
it seems that we should be able to support this: 
```
resultRunningMean, resultRunningVariance
Inputs/Outputs. Running mean and variance tensors (these have the same descriptor as the bias and scale). Both of these pointers can be NULL but only at the same time. The value stored in resultRunningVariance (or passed as an input in inference mode) is the sample variance and is the moving average of variance[x] where the variance is computed either over batch or spatial+batch dimensions depending on the mode. If these pointers are not NULL, the tensors should be initialized to some reasonable values or to 0.
```

priyamm145:
Hey @CarloLucibello ,

I would like to work on this. Can you assign this to me? 

I have no prior julia experience, but, have good cuda and C++ experience.

prikmm:
Hey @CarloLucibello ,

I would like to work on this. Can you assign this to me?

I have no prior julia experience, but, have good cuda and C++ experience.

ToucheSir:
We don't need to explicitly assign anything for people to work on them. If you're interested, feel free to file a PR and we'll start from there.

paulnovo:
Should this be closed? My PRs have merged.

CarloLucibello:
Fixed by #2427 (merged 2024-04-29). Verified on current Flux with a CUDA device:

```julia
bn = BatchNorm(3, track_stats=false) |> gpu
Flux.trainmode!(bn)
bn(CUDA.randn(Float32, 4, 4, 3, 2))          # works
Flux.gradient(m -> sum(m(x)), bn)            # works
```

Thanks @paulnovo.

PR #2427, opened 2024-04-21
Allow BatchNorm on CUDA with track_stats=False

paulnovo:
Enables BatchNorm without track_stats in training and test modes. Also, unit tests are added to ensure the CUDA implementation matches the CPU implementation.

This address #1606 but requires changes in [NNlib PR 576](https://github.com/FluxML/NNlib.jl/pull/576).
### PR Checklist

- [x] Tests are added
- [ ] Entry in NEWS.md


paulnovo:
The "buildkite/flux-dot-jl" tests appear to be failing because NNlib 0.9.13 was installed. 0.9.14 includes my related fixes to NNlib for track_stats=False. Is there a preferred way forward here? Should I increase the compat version of NNlib from 0.9.1 to 0.9.14?

ToucheSir:
Yes, please do and then this should be ready if tests pass.

ToucheSir (approved):
Thanks!

PR #2429, opened 2024-04-26
Add working downloads badge

pricklypointer:
Fixed the download count badge using https://juliapkgstats.com. The current badge is broken because the old API and PkgVisBoard is not maintained.

This PR:
- Replaces the broken badge with a new badge that uses the new API

Please review and provide any feedback. Thanks for considering this PR!

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 <div align="center">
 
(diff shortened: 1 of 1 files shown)


CarloLucibello:
Thanks!

Issue #2432, opened 2024-05-02
Dense layers with shared parameters 


pagnani:
Dear all

Suppose I have the following (trivial) encoder/decoder network. I want the decoder's parameters to be shared with the encoder (or more precisely the transpose of it):

```
using Flux, Statistics
import CUDA

function prepare_model(nin,nout,M,B; device=cpu)
    enc = Dense(nin=>nout) |> device
    dec = Dense(transpose(enc.weight))
    model = Chain(enc,dec) 
    loss(x,m) = mean(abs2,x-m(x))
    opt = Adam()
    Z = rand(Float32,nin,M) |> device
    data = Flux.DataLoader(Z,batchsize=B,shuffle=true)
    return model, data,loss,opt,Z
end
```

If I build the model on CPUs, I get the correct behavior (the 2 weights of both encoder and decoder are the same ... see the `===`)

```
julia> model_cpu, data_cpu, loss, opt, Z_cpu = prepare_model(5,6,1024,128, device=Flux.cpu);
julia> model_cpu.layers[1].weight === model_cpu.layers[2].weight'
true
```

Now if I build the model on GPUs,  as from the `prepare_model` method is doing it is ok:

```
julia> model_gpu, data_gpu, loss, opt, Z_gpu = prepare_model(5,6,1024,128, device=Flux.gpu);
julia> model_gpu.layers[1].weight=== model_gpu.layers[2].weight'
true
```
And now the problem. Suppose that now send to the GPU the model in this way:
```
julia> model_gpu_wrong = Flux.gpu(model_cpu)
Chain(
  Dense(5 => 6),                        # 36 parameters
  Dense(6 => 5),                        # 35 parameters
)                   # Total: 4 arrays, 71 parameters, 688 bytes.

julia> model_gpu_wrong.layers[1].weight === model_gpu_wrong.layers[2].weight'
false
```
So it seems that gpu-ing the entire network in one shot seems to lose track of the parameter's sharing.

Julia Version 1.10.0
Commit 3120989f39b (2023-12-25 18:01 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 40 × Intel(R) Xeon(R) CPU E5-2640 v4 @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, broadwell)
[...]

PR #2436, opened 2024-05-06
Patch `Flux._isleaf` for abstract arrays with bitstype elements

jondeuce:
Fixes https://github.com/FluxML/Flux.jl/issues/2432:

```julia
julia> x = [1.0];
julia> Flux._isleaf(x') # should be false
true # false with this PR

julia> m = (; a = x, b = x'); # tied weights
julia> mcopy = fmap(copy, m; exclude = Flux._isleaf); # copy should preserve tie
julia> mcopy.a === mcopy.b' # should be true
false # true with this PR
```

On master we have `Flux._isleaf(x) = _isbitsarray(x) || Functors.isleaf(x)`, but `_isbitsarray(x)` returns `true` for any `AbstractArray{T}` where `isbitstype(T) == true`, and so we get `_isbitsarray([1.0]') == true` and therefore `Flux._isleaf([1.0]') == true`. In the referenced issue, this breaks parameter sharing between a set of weights and their transpose when a model is moved to the gpu.

The fundamental issue is that AFAICT there is not a good way to extend `Functors.isleaf` outside of Functors.jl for abstract types which may contain children of the same abstract type. For example, here `Transpose <: AbstractArray` contains a `parent::AbstractArray` field, and so Functors.jl must overload `Functor.functor(::Transpose)` otherwise it would not be recursed into. But of course this can't be done similarly outside of Functors.jl without type piracy (hence `Flux._isleaf`).

So in order to:

1. Avoid copying all the overloaded [`Functors.functor(::AbstractArray)` methods defined here](https://github.com/FluxML/Functors.jl/blob/cfc6a608e309c64e4da0f44cd937cb9efa4fd6c7/src/base.jl#L29-L49) into Flux, and
2. Maintain the desired behaviour of treating `AbstractArray`s with bitstype elements as leaves,

I've removed `_isbitsarray` in favour of defining `_isleaf` methods directly, and special-cased `Flux._isleaf(::Union{Transpose, Adjoint, PermutedDimsArray}) = false` to match Functors.jl. The other option is to do this in Functors.jl directly, but I'm not sure if treating all bitstype arrays as leaves is desirable in general (also it's probably breaking?).

### DIFF
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -185,7 +185,7 @@ _isleaf(x) = Functors.isleaf(x)
 
 _isleaf(::AbstractArray{<:Number}) = true
 _isleaf(::AbstractArray{T}) where T = isbitstype(T)
-_isleaf(::Union{Transpose, Adjoint}) = false
+_isleaf(::Union{Transpose, Adjoint, PermutedDimsArray}) = false
 
 _isleaf(::AbstractRNG) = true
 
--- a/src/functor.jl
+++ b/src/functor.jl
@@ -181,12 +181,13 @@ julia> m.bias
 """
 cpu(x) = fmap(x -> adapt(FluxCPUAdaptor(), x), x, exclude = _isleaf)
 
-_isbitsarray(::AbstractArray{<:Number}) = true
-_isbitsarray(::AbstractArray{T}) where T = isbitstype(T)
-_isbitsarray(x) = false
+_isleaf(x) = Functors.isleaf(x)
+
+_isleaf(::AbstractArray{<:Number}) = true
+_isleaf(::AbstractArray{T}) where T = isbitstype(T)
+_isleaf(::Union{Transpose, Adjoint}) = false
 
 _isleaf(::AbstractRNG) = true
-_isleaf(x) = _isbitsarray(x) || Functors.isleaf(x)
 
 # the order below is important
 const GPU_BACKENDS = ("CUDA", "AMDGPU", "Metal", "CPU")
--- a/test/ext_cuda/cuda.jl
+++ b/test/ext_cuda/cuda.jl
@@ -109,7 +109,7 @@ end
   # This test should really not go through indirections and pull out Fills for efficiency
   # but we forcefully materialise. TODO: remove materialising CuArray here
   @test gradient(x -> sum(cpu(x)), ca)[1] isa CuArray # This involves FillArray, which should be GPU compatible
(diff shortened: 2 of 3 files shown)


ToucheSir:
Thanks for the PR! I'm a little late responding to the main issue so I'll do so here. It looks like there are still GPU-related tests failing and it's not immediately clear why, so more tweaking may be required.

Stepping back a bit, you were right to point out in https://github.com/FluxML/Flux.jl/issues/2432 that the call stack/responsibilities of various libraries here is not super clear. Let me lay out what I think a more ideal design could be, and let's see how much of that can inform this PR.

One possibility to bridge the disparate views of Flux, Functors and Adapt (which wasn't mentioned in #2432 but is a very important third player here) is to move some logic to the latter. My proposal would be to share the cache `fmap` uses to detect shared parameters with Flux's Adapt adapters. The motivation here is that Adapt has much better coverage of array wrapper types than Functors, including how to recurse into them. By continuing to consider wrapped arrays as "XPU-able" in Flux, we can decouple this part of the conversion logic and delegate it to Adapt instead of having to re-implement it.

jondeuce:
> It looks like there are still GPU-related tests failing and it's not immediately clear why, so more tweaking may be required.

I think the only failure is

https://github.com/FluxML/Flux.jl/blob/c442f0ca9ef716dfbc215f2b4422b6c34099f649/test/ext_cuda/cuda.jl#L112

which fails because previously `gradient(x -> sum(cpu(x)), ca')[1]` returned an all-ones `CuMatrix{Float32}`, but now returns an all-ones `Adjoint{Float32, CuMatrix{Float32}}`, which is a slight improvement. I'll update the test and see if anything else fails.

> One possibility to bridge the disparate views of Flux, Functors and Adapt (which wasn't mentioned in #2432 but is a very important third player here) is to move some logic to the latter. My proposal would be to share the cache `fmap` uses to detect shared parameters with Flux's Adapt adapters. The motivation here is that Adapt has much better coverage of array wrapper types than Functors, including how to recurse into them. By continuing to consider wrapped arrays as "XPU-able" in Flux, we can decouple this part of the conversion logic and delegate it to Adapt instead of having to re-implement it.

All of this sounds fantastic, and of course much more robust than this PR, which is really just meant as a stopgap to patch adjoint/transpose/etc. until a more complete fix is implemented.

CarloLucibello (approved):

PR #2442, opened 2024-05-11
Enable remaining enzyme test

wsmoses:
Simultaneously want to test if we no longer need runtime activity

@CarloLucibello @ToucheSir if this passes I am unaware of any issues of flux + enzyme


CarloLucibello (approved):

CarloLucibello:
Nice! What really matters now is gpu support

wsmoses:
@CarloLucibello do you have any gpu tests/etc. Presently I'm unaware of any GPU Flux enzyme test [and similarly unaware of any reason it would fail presently]. If you can make one / open an issue for anything which goes wrong, I can try to fix quickly

PR #2447, opened 2024-05-14
Add a "gradient zoo" page?

mcabbott:
A while back I tried to write up a docs page pointing out that Flux is AD-agnostic -- anything which produces an appropriately nested structure should work. 

This is a rough draft, help with tidying it up would be welcome.

Perhaps it's too opinionated about other packages. Certainly any opinions it offers should have clear dates attached.

Rendered version: https://github.com/mcabbott/Flux.jl/blob/grad_zoo/docs/src/tutorials/gradient_zoo.md

### DIFF
--- a/docs/src/tutorials/gradient_zoo.md
+++ b/docs/src/tutorials/gradient_zoo.md
@@ -5,6 +5,8 @@ also known as reverse-mode automatic differentiation.
 Given a model, some data, and a loss function, this answers the question
 "what direction, in the space of the model's parameters, reduces the loss fastest?"
 
+### `gradient(f, x)` interface
+
 Julia's ecosystem has many versions of `gradient(f, x)`, which evaluates `y = f(x)` then retuns `∂y_∂x`. The details of how they do this vary, but the interfece is similar. An incomplete list is (alphabetically):
 
 ```julia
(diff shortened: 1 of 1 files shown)


docs/src/tutorials/gradient_zoo.md
```diff
@@ -0,0 +1,330 @@
...
+
+### [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl)
+
+New package which works on the LLVM code which Julia compiles down to.
+
+* Allows mutation of arrays.
+
+* Long compilation times, on the first call.
+
+* Does not at present work on all Flux models, due to missing rules.
```
wsmoses, line 232: Out of curiousity, what models does it not work on atm?

docs/src/tutorials/gradient_zoo.md
```diff
@@ -0,0 +1,330 @@
...
+* Generally slower than Zygote, allocates more, and supports fewer operations.
+
+* Custom rules via its own `track` and `@grad`.
+
+
+### [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl)
+
+New package which works on the LLVM code which Julia compiles down to.
+
+* Allows mutation of arrays.
```
wsmoses, line 228: I might also add high-performance :)

docs/src/tutorials/gradient_zoo.md
```diff
@@ -0,0 +1,330 @@
...
+
+* Long compilation times, on the first call.
+
+* Does not at present work on all Flux models, due to missing rules.
+
+* Does not always handle type instability.
+
+* Custom rules by its own rules... Generally fewer such rules than Zygote, and at a lower level -- applied to `BLAS.gemm!` not `*`.
```
wsmoses, line 236: Enzyme also supports importing chainrules as well (though it is discouraged as they are usually lower-performance for sake of not understanding mutation)

docs/src/tutorials/gradient_zoo.md
```diff
@@ -0,0 +1,330 @@
...
+### Zygote over Zygote
+
+In principle this works but in practice... best start small.
+
+### ForwardDiff over Zygote
+
+Zygote.hessian is like this.
+
+### Enzyme.jl
+
```
wsmoses, line 306: Haven't used on Flux functions, but it is well tested in our CI

docs/src/tutorials/gradient_zoo.md
wsmoses: FWIW all the CPU Flux tests I know of presently pass (and please let me know any failing ones so I can fix them), including some which Zygote fails on: https://github.com/FluxML/Flux.jl/blob/d39d13b5de668fee14b08e846f452472a69fa857/test/ext_enzyme/enzyme.jl#L143

mcabbott: That's great!

Sorry this was written a few months back, and I don't recall what issues I ran into. I see many have been fixed.

PR #2449, opened 2024-05-20
Bump AMDGPU to 0.9

pxl-th:
### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md


pxl-th:
@darsnack can you please merge it? I don't have write permissions

darsnack:
The GPU test failure seems real? Can you look into that?

pxl-th:
That issue seems to come from LLVM and when having CUDA, AMDGPU and Metal in the same project.
Maybe we should add respective package only when testing for the selected GPU backend?

PR #2462, opened 2024-06-22
Add output padding for ConvTranspose

guiyrt:
This PR provides the ability to increase `ConvTranspose`'s output shape when stride > 1, implemented in the same way as in PyTorch `output_padding` (see [ConvTranspose3d](https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose3d.html) and [NaiveConvolutionTranspose3d.cpp](https://github.com/pytorch/pytorch/blob/b0ae0db8156f186eaed69b0332d8698a8dfc799a/aten/src/ATen/native/NaiveConvolutionTranspose3d.cpp#L708)). 

The purpose of this is to conserve `Conv`/`ConvTranspose` inversability. For stride = N, N>1, there are N possibilities per dimension that result in the same `Conv` output (see [what-output-padding-does-in-nn-convtranspose2d](https://stackoverflow.com/questions/67096544/what-output-padding-does-in-nn-convtranspose2d)). The added argument `outpad` controls how much the output shape is increased to match the strided `Conv` counterpart. Can be provided as an integer to equally increase all dimensions, or as a tuple for individual dimensional control).

`outpad` is invalid if any `outpad` .>= `stride`. Currently there is no verification of this when constructing `ConvTranspose`, it only breaks when called and throws **DimensionMismatch**. A similar thing happens in PyTorch, it allows constructing ConvTransposeNd with invalid `output_paddding`, but then throws **RuntimeError** before doing the deconvolution. As the verification can be done in the constructor, it makes sense to handle it there, but let me know what you think.

I tried to follow the variable naming style, but it can also be changed if you see a better fitting alternative. The same applies for other things such as argument order.

### PR Checklist

- [X] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -224,13 +224,17 @@ function _print_conv_opt(io::IO, l)
 end
 
 """
-    ConvTranspose(filter, in => out, σ=identity; stride=1, pad=0, dilation=1, [bias, init])
+    ConvTranspose(filter, in => out, σ=identity; stride=1, pad=0, outpad=0, dilation=1, [bias, init])
 
 Standard convolutional transpose layer. `filter` is a tuple of integers
 specifying the size of the convolutional kernel, while
 `in` and `out` specify the number of input and output channels.
 
 Note that `pad=SamePad()` here tries to ensure `size(output,d) == size(x,d) * stride`.
+
+To conserve [`Conv`](@ref) inversability when `stride > 1`, `outpad` can be used to increase the size
+of the output in the desired dimensions. Whereas `pad` is used to zero-pad the input,
+`outpad` only affects the output shape.
 
 Parameters are controlled by additional keywords, with defaults
 `init=glorot_uniform` and `bias=true`.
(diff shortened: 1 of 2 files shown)


CarloLucibello (approved):

CarloLucibello:
thanks!

paulnovo:
Hi @CarloLucibello and @guiyrt, I was wondering if the output padding should also be applied in the [AMDGPU implementation of conv_transpose_dims](https://github.com/FluxML/Flux.jl/blob/master/ext/FluxAMDGPUExt/conv.jl#L7-L21)?

I am new to Flux, but I noticed this when rebasing my changes to [ConvTranspose](https://github.com/FluxML/Flux.jl/pull/2463), so I thought I should ask.

pxl-th:
@paulnovo yes, currently `ConvTranspose` is broken for AMDGPU because of this:
https://buildkite.com/julialang/flux-dot-jl/builds/4674#01903fca-3e24-4906-81a2-728846992074/361-742

PR #2464, opened 2024-06-22
feat: Distributed data parallel training support

askorupka:
Support for distributed data parallel training. Inspired by https://github.com/LuxDL/Lux.jl/pull/500
This PR is still work in progress. 
PR checklist to be continued.

### PR Checklist 

- [x] Tests are added
- [x] Entry in NEWS.md

Both `MPIBackend` and `NCCLBackend` are supported.

Module can be used as in the below example (for distributed runs use `mpiexecjl --project=@. -n 3 julia distributed_MPI.jl` from your terminal, where `distributed_MPI.jl` (feel free to also use `NCCLBakend`'):
```julia
using Flux, MPI, NCCL, CUDA
using Random
using Optimisers
using Zygote
using Statistics

CUDA.allowscalar(false)

DistributedUtils.initialize(MPIBackend)
backend = DistributedUtils.get_distributed_backend(MPIBackend)
rank = DistributedUtils.local_rank(backend)

model = Chain(Dense(1 => 256, tanh), Dense(256 => 1)) |> gpu

model = DistributedUtils.synchronize!!(backend, DistributedUtils.FluxDistributedModel(model); root=0) 

x = rand(Float32, 1, 16) |> gpu
y = x .^ 3

opt = DistributedUtils.DistributedOptimizer(backend, Optimisers.Adam(0.001f0))
st_opt = Optimisers.setup(opt, model)
st_opt = DistributedUtils.synchronize!!(backend, st_opt; root=0) 

loss(model) = mean((model(x) .- y).^2)
g_ = gradient(m -> loss(m), model)[1] 
Optimisers.update!(st_opt, model, g_)

for epoch in 1:100
  global model, st_opt
  l, back = Zygote.pullback(loss, model)
  println("Epoch $epoch: Loss $l")
  g = back(one(l))[1]
  st_opt, model = Optimisers.update(st_opt, model, g)
end
```


CarloLucibello:
I suggest removing NCCL from this PR and just focusing on MPI

distributed.jl
```diff
@@ -0,0 +1,211 @@
...
+
+data = Flux.DataLoader((x, y) |> gpu, batchsize=64, shuffle=true);
+
+### approach 1
+
+for epoch in 1:16
+    Flux.train!(model, data, st_opt) do m, x, y
+      loss(m(x), y) # ERROR: ArgumentError: Illegal conversion of a CUDA.DeviceMemory to a Ptr{Float32}
+    end
+end
```
CarloLucibello, line 41: let's focus on the other approach with gradient, `train!` is a convenience function relying on `gradient`.

distributed.jl
```diff
@@ -0,0 +1,211 @@
...
+rank = DistributedUtils.local_rank(backend)
+
+model = Chain(Dense(1 => 256, tanh), Dense(256 => 1))
+
+rng = Random.default_rng()
+Random.seed!(rng, rank)
+
+model = DistributedUtils.synchronize!!(backend, DistributedUtils.FluxDistributedModel(model); root=0) 
+
+x = rand(1, 16) |> gpu
```
CarloLucibello, line 23:
```suggestion
x = rand(Float32, 1, 16) |> gpu
```

distributed.jl
```diff
@@ -0,0 +1,211 @@
...
+
+for epoch in 1:16
+    Flux.train!(model, data, st_opt) do m, x, y
+      loss(m(x), y) # ERROR: ArgumentError: Illegal conversion of a CUDA.DeviceMemory to a Ptr{Float32}
+    end
+end
+
+### approach 2
+
+gs_ = gradient(loss, model)[1] # ArgumentError: Illegal conversion of a CUDA.DeviceMemory to a Ptr{Float32}
```
CarloLucibello, line 45: this is erroring because the data has not been sent to the process' device.

It should be something like.
```
device = get_device()
model = model |> device
x = x |> device
y = y |> device
loss(m, x, y) = mean((m(x) .- y).^2)
g = gradient(m -> loss(m, x, y), model)[1]
```
We should make sure that what is returned by `get_device()` is the same device set for process by DistributedUtils.__initialize. See [here](https://github.com/LuxDL/Lux.jl/blob/1a61165fe0802eff11766f55819b27f22a6c9059/ext/LuxMPIExt.jl#L19). Since we don't have a `set_device!` function we should add one.

askorupka: ok I think I see the problem. this is connected to the discussion I've started on our discord and my doubts re reliance on external `LuxDeviceUtils` package.

I'll try to explain it here.
so the issue here is that the way `Flux.get_device()` works is that it doesn't take any arguments (https://fluxml.ai/Flux.jl/stable/guide/gpu/).
the code for some MPI methods e.g. `reduce!` is needs two devices, for both buffer subject to reduce operation as well as object after it, see e.g. (`public_api.jl` line 175)

```
function reduce!(backend::AbstractFluxDistributedBackend,
        sendbuf, recvbuf, op::F; root::Int=0) where {F}
    send_dev = LuxDeviceUtils.get_device(sendbuf)
    recv_dev = LuxDeviceUtils.get_device(recvbuf)
    if send_dev === recv_dev
        return __reduce!(backend, sendbuf, recvbuf, op, send_dev; root)
    else
        sendbuf_ = sendbuf |> recv_dev
        @warn "`sendbuf` and `recvbuf` are on different devices." maxlog=1
        __reduce!(backend, sendbuf_, recvbuf, op, recv_dev; root)
        return recvbuf
    end
end
```
because here we need to get device for both `send_dev` and `recv_dev` I couldn't use `Flux.get_device()` in case both devices would be different - as it would always return the same.
for `Flux.get_device(send_dev)` it naturally yields `MethodError: no method matching get_device(::Matrix{Float32})`.

so, shall I understand that data always should be on device defined by `Flux.get_device()` and there's no possibility that `sendbuf` will be on the different device than `recvbuf`? 
please note that this approach was took by https://github.com/avik-pal/FluxMPI.jl/blob/main/src/FluxMPI.jl
it would make life simpler. otherwise there might be a need for own implementation of `LuxDeviceUtils`.

askorupka: honestly - I can't imagine why they would be different and what would use case be but I might be missing sth

askorupka: I've tried to test the abovementioned approach (new changes pushed) but got an error when calling
`model = DistributedUtils.synchronize!!(backend, DistributedUtils.FluxDistributedModel(model); root=0) `:

```
ERROR: ArgumentError: cannot take the GPU address of inaccessible device memory.

You are trying to use memory from GPU 0 on GPU 1.
P2P access between these devices is not possible; either switch to GPU 0
by calling `CUDA.device!(0)`, or copy the data to an array allocated on device 1.
```

after I've called `CUDA.device!(0)`
got
```
This object is not a GPU array
```
called by line 73 of `FluxMPIExt.jl`
`model = model |> device` was called so I think there is something inside the leaves of model which doesn't know how to be handled/isn't properly transferred to device - and based on println looks to be tanh/identity

askorupka:
@CarloLucibello  I was able to move example forward according to your suggestions and MPI example with training works :tada: (I still need to do some cleanup tho)
details are in comments above - may be useful for you to have a look.

askorupka:
Update: both MPI and NCCL work.
please run `mpiexecjl --project=@. -n 3 julia distributed_NCCL.jl` or `mpiexecjl --project=@. -n 3 julia distributed_MPI.jl` respectively.

still in the draft state, requires some work - should be easier from now on:
- add tests
- examples
- docs
- test CUDA aware implementation

distributed.jl
askorupka: I'm basically talking to myself here but need to get this straight.
Above problems were solved by transferring corresponding objects (model, x, y) to gpu (e.g. `x |> gpu`) in example `distributed_*.jl`. This is also in line with `Lux` & `FluxMPI` implementations.

In other words, using `Flux.get_device()` instead of `gpu` as @CarloLucibello proposed was not helpful here.

Also, in case MPI implementation is not CUDA-aware (as in case of machine I'm currently working on) we need abovementioned lines of `FluxMPIExt.jl` - for different methods (`bcast`, `allreduce`, `reduce`), it's now respectively 60-83, 106-129 and 152-175.

For clarity I'm talking about what is now uncommented and defined as
```
for (aware, dType) in ((MPI_CUDA_AWARE, LuxCUDADevice), (MPI_ROCM_AWARE, LuxAMDGPUDevice))
    if !aware
        @eval begin
            function DistributedUtils.__reduce!(backend::MPIBackend, sendrecvbuf, op::F,
                    dev::$dType; root::Int) where {F}
                cdev = cpu_device()
                sendrecvbuf_ = sendrecvbuf |> cdev
                DistributedUtils.__reduce!(backend, sendrecvbuf_, op, cdev; root)
                sendrecvbuf .= dev(sendrecvbuf_)
                return sendrecvbuf
            end

            [...]
        end
    end
end
```

In case the implementation of MPI is not CUDA-aware we need to transfer the data to GPU indirectly via CPU. This block of code is doing exactly that but it requires user to set preferences correctly (default: `false` so not CUDA-aware) via `Peferences.jl` mechanism:
```julia
set_preferences!("Flux", "FluxDistributedMPICUDAAware" => true)
```
or (bash)
```bash
cat <<EOF > LocalPreferences.toml
[Flux]
FluxDistributedMPICUDAAware = true
EOF
```
```

**Review by ToucheSir — COMMENTED** [review]
This PR is a real tour de force, great work!

Assuming https://github.com/FluxML/Flux.jl/pull/2464#issuecomment-2212521646 means you're starting to wrap things up, here are a couple heads up so they don't come as a surprise for the non-draft PR review.

File: ext/FluxMPINCCLExt/FluxMPINCCLExt.jl [hunk 4]
```diff
@@ -0,0 +1,109 @@
...
+    DistributedUtils.NCCL_Initialized[] = true
+    return
+end
+
+function DistributedUtils.__get_distributed_backend(::Type{NCCLBackend})
+    unique_id = NCCL.UniqueID()  # Generate on all ranks to know the type
+    mpi_backend = DistributedUtils.__get_distributed_backend(MPIBackend)
+    buf = [unique_id.internal...]
+    DistributedUtils.bcast!(mpi_backend, buf; root=0)
+    @set! unique_id.internal = Tuple(buf)
```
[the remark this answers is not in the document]
askorupka: I've tried and got an error
```
ERROR: setfield!: immutable struct of type ncclUniqueId cannot be changed
Stacktrace:
 [1] setproperty!(x::NCCL.LibNCCL.ncclUniqueId, f::Symbol, v::NTuple{128, Int8})
 ```
will try to work around it

**Comment by askorupka** [comment]
Docs added so PR checklist completed. Ready for review :rocket:

File: docs/src/guide/gpu.md [hunk 5]
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
 ```
+
+## Distributed data parallel training
+
+Flux supports now distributed data parallel training with `DistributedUtils` module. 
+If you want to run your code on multiple GPUs, you have to install `MPI.jl` (see [docs](https://juliaparallel.org/MPI.jl/stable/usage/) for more info).
+
+```julia-repl
+julia> using Pkg; add MPI
+
```
CarloLucibello, line 396:
```suggestion
```

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+julia> using MPI
+
+julia> MPI.install_mpiexecjl()
+```
+
+Now you can run your code with `mpiexecjl --project=. -n <np> julia <filename>.jl` from CLI.
+
+You can use two backends, `MPIBackend` and `NCCBackend`. First, initialize the respective backend with `DistributedUtils.initialize`, e.g.
```
CarloLucibello, line 404:
```suggestion
You can use either the `MPIBackend` or `NCCLBackend`, the latter only if also `NCCL.jl` is loaded. First, initialize a backend with `DistributedUtils.initialize`, e.g.
```

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+
+```julia-repl
+julia> using Flux, MPI, NCCL
+
+julia> DistributedUtils.initialize(NCCLBackend)
+
+julia> backend = DistributedUtils.get_distributed_backend(NCCLBackend)
+
```
CarloLucibello, line 412:
```suggestion
```

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+```
+
+Now you can run your code with `mpiexecjl --project=. -n <np> julia <filename>.jl` from CLI.
+
+You can use two backends, `MPIBackend` and `NCCBackend`. First, initialize the respective backend with `DistributedUtils.initialize`, e.g.
+
+```julia-repl
+julia> using Flux, MPI, NCCL
+
+julia> DistributedUtils.initialize(NCCLBackend)
```
CarloLucibello, line 409: is the symbol `NCCLBackend` visible here? Maybe needs `using DistributedUtils: NCCLBacked`.

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+
+julia> backend = DistributedUtils.get_distributed_backend(NCCLBackend)
+
+NCCLBackend{Communicator, MPIBackend{MPI.Comm}}(Communicator(Ptr{NCCL.LibNCCL.ncclComm} @0x000000000607a660), MPIBackend{MPI.Comm}(MPI.Comm(1140850688)))
+```
```
CarloLucibello, line 414: make sure the whole example above runs in the REPL exactly as written

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+
+Pass your model, as well as any data to GPU device.
+```julia-repl
+julia> model = Chain(Dense(1 => 256, tanh), Dense(256 => 1)) |> gpu
+
```
CarloLucibello, line 419:
```suggestion
```

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+julia> y = x .^ 3
+1×16 CUDA.CuArray{Float32, 2, CUDA.DeviceMemory}:
+ 0.0137076  0.0362744  0.791443  0.171815  0.620854  0.668804  0.53197  0.819654  0.108651  0.179971  0.312918  0.388508  0.907292  0.00155418  0.29  0.435899
+```
+
+You can also use `DistributedUtils.DistributedDataContainer` to split the data uniformly accross processes.
+
+```julia-repl
+julia> data = DistributedUtils.DistributedDataContainer(backend, x)
+Flux.DistributedUtils.DistributedDataContainer(Float32[0.23932439 0.33102947 … 0.66191036 0.75822026], [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16])
```
CarloLucibello, line 438:
``` suggestion
julia> x = DistributedUtils.DistributedDataContainer(backend, x)
Flux.DistributedUtils.DistributedDataContainer(Float32[0.23932439 0.33102947 … 0.66191036 0.75822026], [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16])
```
should this be called `x` again?
Also, shouldn't `y` be distributed as well?

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+```julia-repl
+julia> model = DistributedUtils.synchronize!!(backend, DistributedUtils.FluxDistributedModel(model); root=0)
+Chain(
+  Dense(1 => 256, tanh),                # 512 parameters
+
+  Dense(256 => 1),                      # 257 parameters
+)                   # Total: 4 arrays, 769 parameters, 744 bytes.
+```
+
+Time to set up optimizer by using `DistributedUtils.DistributedOptimizer` and synchronize it as well.
```
CarloLucibello, line 451:
```suggestion
Time to set up an optimizer by using `DistributedUtils.DistributedOptimizer` and synchronize it as well.
```

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+st_opt = Optimisers.setup(opt, model)
+st_opt = DistributedUtils.synchronize!!(backend, st_opt; root=0) 
+```
+
+Now you can define loss and train the model.
+```julia
+loss(model) = mean((model(x) .- y).^2)
+g_ = gradient(m -> loss(m), model)[1] 
+Optimisers.update!(st_opt, model, g_)
+
```
CarloLucibello, line 464:
```suggestion
```
no need for this

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+for epoch in 1:100
+  global model, st_opt
+  l, back = Zygote.pullback(loss, model)
+  println("Epoch $epoch: Loss $l")
+  g = back(one(l))[1]
+  st_opt, model = Optimisers.update(st_opt, model, g)
```
CarloLucibello, line 470:
```suggestion
  l, grad = Zygote.withgradient(loss, model)
  println("Epoch $epoch: Loss $l")
  st_opt, model = Optimisers.update(st_opt, model, grad[1])
```

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+end
+```
+
+Remember that in order to run it on multiple GPUs you have to run from CLI `mpiexecjl --project=. -n <np> julia <filename>.jl`,
+where `--project` is path to your Julia environment and `<np>` number of processes.
```
CarloLucibello, line 475:
```suggestion
where  `<np>` is the number of processes that you want to use. The number of processes should correspond to the number of gpus + 1.
```
I'm not sure that this is correct...

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,110 @@ Flux.supported_devices
...
+If it is, set your local preference as below
+```julia
+using Preferences
+set_preferences!("Flux", "FluxDistributedMPICUDAAware" => true)
+```
+
+!!! warning "Known shortcomings"
+    
+    We don't run CUDA-aware tests so you're running it at own risk.
+    AMDGPU support is mostly experimental.
```
CarloLucibello, line 493:
```suggestion
```

docs/src/guide/gpu.md
askorupka: yes it is

docs/src/guide/gpu.md
askorupka: I'd say 
```
where  `<np>` is the number of processes that you want to use. The number of processes usually corresponds to the number of gpus.
```

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,100 @@ Flux.supported_devices
...
+for epoch in 1:100
+  global model, st_opt
+  l, grad = Zygote.withgradient(loss, model)
+  println("Epoch $epoch: Loss $l")
+  st_opt, model = Optimisers.update(st_opt, model, grad[1])
+end
+```
+
+Remember that in order to run it on multiple GPUs you have to run from CLI `mpiexecjl --project=. -n <np> julia <filename>.jl`,
+where  `<np>` is the number of processes that you want to use. The number of processes should correspond to the number of gpus + 1.
```
askorupka, line 466:
```suggestion
where  `<np>` is the number of processes that you want to use. The number of processes usually corresponds to the number of gpus.
```

docs/src/guide/gpu.md
askorupka: It doesn't have to be called `x` cause this line is just example how to distribute data uniformly across processes and it's optional.

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,100 @@ Flux.supported_devices
...
+julia> x = rand(Float32, 1, 16) |> gpu
+1×16 CUDA.CuArray{Float32, 2, CUDA.DeviceMemory}:
+ 0.239324  0.331029  0.924996  0.55593  0.853093  0.874513  0.810269  0.935858  0.477176  0.564591  0.678907  0.729682  0.96809  0.115833  0.66191  0.75822
+
+julia> y = x .^ 3
+1×16 CUDA.CuArray{Float32, 2, CUDA.DeviceMemory}:
+ 0.0137076  0.0362744  0.791443  0.171815  0.620854  0.668804  0.53197  0.819654  0.108651  0.179971  0.312918  0.388508  0.907292  0.00155418  0.29  0.435899
+```
+
+You can also use `DistributedUtils.DistributedDataContainer` to split the data uniformly accross processes.
```
askorupka, line 430:
```suggestion
You can also use `DistributedUtils.DistributedDataContainer` to split the data uniformly across processes.
```

docs/src/guide/gpu.md
askorupka: See original Lux PR (linked in the first comment) example not using it at all

docs/src/guide/gpu.md
```diff
@@ -385,3 +385,100 @@ Flux.supported_devices
 Flux.get_device
 Flux.gpu_backend!
 ```
+
+## Distributed data parallel training
+
```
CarloLucibello, line 390: Let's add a banner like this 
https://documenter.juliadocs.org/stable/showcase/#Danger-admonition
saying that distributed support is experimental and could change in the future. 
This way, if introduce changes we don't have to tag breaking version.

CarloLucibello:
The new test files should be included in `tests/runtests.jl`. 
As with other extensions, we should define in the file flags like
```julia
ENV["FLUX_TEST_DISTRIBUTED_MPI"] = "true"
ENV["FLUX_TEST_DISTRIBUTED_NCCL"] = "true"
```
and tests only conditional on them. We should test separately the MPI and NCCL backed. 

For the time being, we won't run the test on the CI because we would have to setup the MPI and NCCL stuff. We can figure out how to test on CI in a follow up PR.

Project.toml
```diff
@@ -9,14 +9,17 @@ Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MLUtils = "f1d291b0-491e-4a28-83b9-f70985020b54"
+MPI = "da04e1cc-30fd-572f-bb4f-1f8673147195"
```
CarloLucibello, line 12: this shouldn't be under deps

Project.toml
```diff
@@ -9,14 +9,17 @@ Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
...
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
+NCCL = "3fe64909-d7a1-4096-9b7d-7a0f12cf0f6b"
```
CarloLucibello, line 14: this shouldn't be under deps

CarloLucibello:
tests failure is unrelated and likely due to Enzyme. I opened an issue in https://github.com/EnzymeAD/Enzyme.jl/issues/1738

CarloLucibello (approved):

pxl-th:
Since Enzyme explicitly installs CUDA when running tests, we should avoid running them on AMDGPU/Metal CIs, until it gains support for them or switch to those backends properly.

kishore-nori:
When updating to Flux.jl latest v0.14.20 , I get the following error, which wasn't there for v0.14.19 , I am on Julia 1.10.5. I have tested it and this is a precompilation error

```julia
ERROR: LoadError: ArgumentError: Package FluxMPIExt does not have CUDA in its dependencies:
- You may have a partially installed environment. Try `Pkg.instantiate()`
  to ensure all packages in the environment are installed.
- Or, if you have FluxMPIExt checked out for development and have
  added CUDA as a dependency but haven't updated your primary
  environment's manifest file, try `Pkg.resolve()`.
- Otherwise you may need to report an issue with FluxMPIExt 
```

I think CUDA, AMDGPU should be mention here https://github.com/FluxML/Flux.jl/blob/26f5c4f708626d7e76d9de2bf8b23050d76afe97/Project.toml#L40

cc @mcabbott

askorupka:
Hello @kishore-nori, CUDA and AMDGPU are not a requirement to run the extension. You can run it also without CUDA/AMDGPU, hence those aren't in the package dependencies.

I've looked into the issue and seems I can't replicate it for version v0.14.20. 
```
julia> using Flux

julia> using MPI

julia> DistributedUtils.initialize(MPIBackend)

julia> backend = DistributedUtils.get_distributed_backend(MPIBackend)
MPIBackend{MPI.Comm}(MPI.Comm(1140850688))
```
as you can see all works properly.
From my experience I can suggest that this might be the issue with your environment after package update.
Please make sure you have MPI installed in your base environment.
It's also worth running `Pkg.instantiate()` as in the error message displayed.

kishore-nori:
update to above: the precompilation error happens only when both `Flux.jl` (0.14.20) and `MPI.jl` are in the environment, if `MPI.jl` is not there, then there is no problem. And the precompilation error complains CUDA absence, like shown above, even if CUDA is not in the environment. So, I think it's got to do with the missing deps for `FluxMPIExt` extension.

askorupka:
Hi @kishore-nori I've managed to replicate the issue, thanks for reporting it.
I'm working on the fix so that FluxMPI doesn't require CUDA anymore as we want to avoid adding too many deps.

PR #2472, opened 2024-07-26
Fix function `_size_check()`

gruberchr:
The previous check `d > 0` and the corresponding error message didn't make sense if `d` was the dimension to be checked. It would make sense to check that the difference between the available number of dimensions and the desired dimension is greater than or equal to zero, i.e., `ndims(x) - d >= 0` or `d <= ndims()` respectively.

Moreover, checking for `d > 0` is rather pointless, since all implemented checks set `d > 0`.

A corrected implementation is provided in this PR, which should fix this issue.

### DIFF
--- a/src/layers/basic.jl
+++ b/src/layers/basic.jl
@@ -191,8 +191,8 @@ Dense(W::LinearAlgebra.Diagonal, bias = true, σ = identity) =
   Scale(W.diag, bias, σ)
 
 function _size_check(layer, x::AbstractArray, (d, n)::Pair)
-  d > 0 || throw(DimensionMismatch(string("layer ", layer,
-    " expects ndims(input) > ", ndims(x)-d, ", but got ", summary(x))))
+  d <= ndims(x) || throw(DimensionMismatch(string("layer ", layer,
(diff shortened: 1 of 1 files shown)


src/layers/basic.jl
```diff
@@ -191,8 +191,8 @@ Dense(W::LinearAlgebra.Diagonal, bias = true, σ = identity) =
   Scale(W.diag, bias, σ)
 
 function _size_check(layer, x::AbstractArray, (d, n)::Pair)
-  d > 0 || throw(DimensionMismatch(string("layer ", layer,
-    " expects ndims(input) > ", ndims(x)-d, ", but got ", summary(x))))
+  d <= ndims(x) || throw(DimensionMismatch(string("layer ", layer,
```
pxl-th, line 194: To protect against typos, I'd still check if `d` is bigger than `0`.
```suggestion
  0 < d <= ndims(x) || throw(DimensionMismatch(string("layer ", layer,
```

pxl-th (approved):
Minor suggestion, other than that, it is good to go.

PR #2478, opened 2024-08-20
Run Enzyme tests only on CUDA CI machine

pxl-th:
Since Enzyme explicitly adds CUDA to its environment, run its tests only on CUDA CI machines to prevent crashing other jobs.

FYI @wsmoses

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -140,10 +140,14 @@ Random.seed!(0)
     @info "Skipping Distributed tests, set FLUX_TEST_DISTRIBUTED_MPI or FLUX_TEST_DISTRIBUTED_NCCL=true to run them."
   end
 
-  @testset "Enzyme" begin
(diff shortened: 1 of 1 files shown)


ToucheSir (approved):

wsmoses:
@pxl-th hm this is a bit weird, since Enzyme CPU tests would be equally good to check, no?

pxl-th:
Then it is probably better to run Enzyme CPU tests when CPU tests are enableed and Enzyme CUDA tests when CUDA tests are enabled.
We were getting this crashes on other backends, because Enzyme was trying to run CUDA tests on them:
https://buildkite.com/julialang/flux-dot-jl/builds/4820#01916a11-6182-4203-b661-ad9c97b96c86

ToucheSir:
Buildkite should still be running Enzyme CPU tests as well.

PR #2481, opened 2024-09-18
Update ci.yml for macos-latest to use aarch64

ViralBShah:


mcabbott (approved):

ViralBShah:
Should we add `pre` to test CI on pre-releases? Maybe for it to co-exist along with nightly or even replace nightly?

ToucheSir:
I would be for this. Flux is less sensitive to nightly changes than Zygote is, but CUDA.jl and Enzyme are very sensitive and cause frequent false positive failures on nightly.

PR #2488, opened 2024-10-04
fix: CUDA package optional for FluxMPIExt

askorupka:
Hotfix for https://github.com/FluxML/Flux.jl/pull/2464 the issue reported by @kishore-nori (https://github.com/FluxML/Flux.jl/pull/2464#issuecomment-2375891808)

### DIFF
--- a/ext/FluxMPIExt/FluxMPIExt.jl
+++ b/ext/FluxMPIExt/FluxMPIExt.jl
@@ -1,6 +1,9 @@
 module FluxMPIExt
 
-using CUDA
+if Base.find_package("CUDA") !== nothing
+    using CUDA
+end
+
 using Flux: MPIBackend, NCCLBackend, DistributedUtils,
             AbstractDevice, FluxCUDADevice, FluxAMDGPUDevice, cpu, gpu,
             get_device, MPI_CUDA_AWARE, MPI_ROCM_AWARE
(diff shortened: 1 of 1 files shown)


CarloLucibello (approved):

kishore-nori:
Thank you very much @askorupka

askorupka:
You're welcome and thanks for reporting the bug @kishore-nori !

kishore-nori:
Hi @askorupka, sorry for bugging you again on this. I have come across the same precompilation error again when `CUDA.jl` is in the package environment. And I add `MPI.jl` and `Flux.jl`. (The error doesn't occur when `CUDA.jl` is not in the package environment)

```julia
ERROR: LoadError: ArgumentError: Package FluxMPIExt does not have CUDA in its dependencies:
- You may have a partially installed environment. Try `Pkg.instantiate()`
  to ensure all packages in the environment are installed.
- Or, if you have FluxMPIExt checked out for development and have
  added CUDA as a dependency but haven't updated your primary
  environment's manifest file, try `Pkg.resolve()`.
- Otherwise you may need to report an issue with FluxMPIExt 
```

can you check if this occurs for you as well? I would suggest clearing the `Flux.jl` compilations from `.julia` folder, to reproduce the error.

ToucheSir:
I would recommend continuing discussion on #2490 since open issues are more discoverable.

PR #2492, opened 2024-10-10
handle data movement with MLDataDevices.jl

### LINKED ISSUE #2482: use MLDataDevices.jl?
I think we should rely on [MLDataDevices.jl](https://github.com/LuxDL/MLDataDevices.jl) for data movement functionality.
This would allow to reduce the codebase here and share functionality with the rest of the ecosystem. 
Moreover, MLDataDevices has more functionalities implemented.

Unfortunately, the mapping among the two is not perfect, and the function `get_device` acquires a different meaning. This is a loose map between  the two:
| Flux  | MLDataDevices |
| ------------- | ------------- |
| `get_device(; [verbose])`  | `gpu_device([device_id]; [force_gpu_usage])` |
| `get_device(backend::String, [idx])` |   `cpu_device` or `gpu_device` | 
| `gpu_backend!(backend::String)`  | `gpu_backend!(backend::String)` |
| `supported_devices()` | `supported_gpu_backends()` |
| X |  `get_device(x) -> dev::AbstractDevice ` |
| X | `functional(x::AbstractDevice)`  or `functional(::Type{<:AbstractDevice})`  |
| X | `set_device!(T::Type{<:AbstractDevice}, dev_or_id)` |
| X | `loaded(x::AbstractDevice)` or `loaded(::Type{<:AbstractDevice})` |
| X |  `default_device_rng(x::AbstractDevice)`

We should be able to deprecate the old interface in favor of the new without a sudden breaking change.

### LINKED ISSUE #2490: The dependency error about `Flux->FluxMPIExt` occurs when updating to Julia 1.11
When updating to Julia 1.11, I resolve to update all necessary changes for the installed packages. The following error occurs.
image

Further, when I instantiate all packages in Julia 1.11, it still exists, as shown in the figure below.
image

Following its hint, the `julia> err` presents the following information:
image

Are there some ways to resolve this issue? I welcome any suggestions you may have. Thanks a lot.

CarloLucibello:
Implements the move to MLDataDevices.jl. Closes #2482 

Thanks to the extended functionality given by MLDataDevices.jl we can also easily fix #2490 

The `gpu/cpu` logic could also be offloaded to MLDataDevices.jl, will do that in a followup PR.

Unfortunately this PR has become rather large also due to some unrelated changes to fix the CI since it was broken due to:
- expected arrays memory sizes as printed in some of the docstrings changed in julia v0.11
- some second derivatives tests for Zygote now fail possibly due to recent changes in Zygote or to julia v0.11
- some gradient tests involving Flux.params() (e.g. for L2 regularization) now fail. possibly due to recent changes in Zygote or to julia v0.11
- Enzyme tests seem to be completely broken (cc @wsmoses) [edit: Enzyme doesn't work on v0.11 yet]

Worth openining separate Issues to keep track of these.

### PR Checklist

- [x] Tests are added
- [x] Entry in NEWS.md


wsmoses:
I assume the error is due to Enztme not yet supporting 1.11, does it work on the LTS?

CarloLucibello:
> I assume the error is due to Enztme not yet supporting 1.11, does it work on the LTS?

ah ok, I didn't know. It works on 1.10 indeed.

wsmoses:
Maybe also adding an LTS cI run would be helpful here?

(Now that it’s released, were starting to adapt to the new intrinsics and array operators [this was sadly for us a more significant set of changes])

wsmoses:
I kind of imagine that happened for most other packages too since the release just happened the other day

askorupka (approved):

darsnack:
Should probably replace `Flux.rng_from_array` with `get_device` and `default_device_rng`.

ToucheSir:
One note for the follow-up: the MLDataDevices path does not flip conv kernels as is required for the AMDGPU backend: https://github.com/FluxML/Flux.jl/blob/aa035e9781f108e6b83fbe7a774f04d2c91f92f4/ext/FluxAMDGPUExt/functor.jl#L87-L109.

We should add this functionality before cutting any releases.

PR #2493, opened 2024-10-11
remove some v0.13 deprecations

CarloLucibello:
Remove v0.13 deprecations except for the old optimizers, which require more care and a separate PR.

This also 
- fixes some CI problems (same as #2492)
- bumps minimum julia version to 1.10 (since it also needed for #2492)
- reintroduces CI on our minimum supported julia version, which somehow was missing.
- fix load_preference for non-CUDA backend that was broken in #2492


CarloLucibello:
Docs CI error is not related to this PR and will be fixed in #2492

ToucheSir:
Should we update https://github.com/FluxML/Flux.jl/blob/master/.buildkite/pipeline.yml too? Relatedly, should we modify https://github.com/FluxML/Flux.jl/blob/master/.github/workflows/nightly-ci.yml to test on `pre` instead of `nightly`?

CarloLucibello:
I kept `nightly` since the `pre` is equivalent to 1.11 right now, but we can switch to `pre` when 1.12rc1 is out

CarloLucibello:
ready to go!

askorupka (approved):

PR #2494, opened 2024-10-11
Delete .github/workflows/Downstream.yml

CarloLucibello:
We don't really need to run these tests, and also mostly contains obsolete packages


ToucheSir:
Shouldn't we keep at least Metalhead and Transformers.jl?

darsnack:
Yeah I very much disagree. We should keep the file but possibly remove outdated packages.

PR #2498, opened 2024-10-12
simplify test machinery

CarloLucibello:
This PR introduces a single function `test_gradients` that can be used across all possible differentiation scenarios (on cpu and any gpu backend, on layer or simple functions...).


CarloLucibello:
@pxl-th I'm seeing something odd with the convolutions on AMDGPU
Could you test locally the following example? Looks like both assertions fail:

```julia

using Flux, AMDGPU, Zygote

m = Conv((2, 2), 3 => 4)
x = rand(Float32, 5, 5, 3, 2)

y, g = Zygote.withgradient(x -> sum(m(x)), x)

dev = gpu_device(force=true)    
m_gpu = m |> dev
x_gpu = x |> dev
y_gpu, g_gpu = Zygote.withgradient(x -> sum(m_gpu(x)), x_gpu)

@assert y_gpu ≈ y  atol=1e-4
@assert Array(g_gpu) ≈ g  atol=1e-4
```

pxl-th:
@CarloLucibello something broke cpu-gpu transfer for convolutions.
MIOpen only supports cross0correlation, so we need to transpose filters when moving to gpu.

This is not executed anymore:
https://github.com/FluxML/Flux.jl/blob/09a16ee0476191d6dc326d46d3e64a7265cf3d23/ext/FluxAMDGPUExt/functor.jl#L89

CarloLucibello:
Ah right, that is when using `MLDataDevices.gpu_device`. `Flux.gpu` should work fine instead. 
I'll try to fix it.

CarloLucibello:
I investigated a bit and hooking up into MLDataDevices's data transfer mechanism is not straightforward. So I propose we merge this PR as it is and proceed with a fix later. The problem is not introduced by this PR in any case, it was pre-existing and now just exposed.

Can I get an approve?

pxl-th (approved):
Yes, I can take a look at it as well tomorrow, we can fix this is a separate PR.

PR #2500, opened 2024-10-14
RNNs redesign

### LINKED ISSUE #807: Initial state in RNNs should not be learnable by default

### LINKED ISSUE #1329: Flux.destructure doesn't preserve RNN state
As pointed out here https://github.com/SciML/DiffEqFlux.jl/issues/391#issuecomment-680127253, when reconstructing `Recur`, the state values are reverted to the initial state. A simple solution would be to do

```julia
mutable struct MyRecur{T}
  cell::T
  init
  state
end

function (m::MyRecur)(xs...)
  h, y = m.cell(m.state, xs...)
  m.state = h
  return y
end

Flux.@functor MyRecur

Flux.trainable(r::MyRecur) = Flux.trainable(r.cell)
```

EDIT: This doesn't work as expected

### LINKED ISSUE #1547: Strange result with gradient
I was still confused after my [last issue](https://github.com/FluxML/Flux.jl/issues/1528), so I dove a little deeper.  Specifically, I wanted to understand how RNN training works with the loss function used [here](https://github.com/FluxML/Flux.jl/blob/master/docs/src/models/recurrence.md):

``loss(x, y) = sum((Flux.stack(m.(x),1) .- y) .^ 2)``

I tested a much simplified version of this, using a 1 -> 1 RNN cell without an activation function, and the same loss function without the square:

```
m = Flux.RNN(1, 1, x -> x)
L(x, y) = sum((Flux.stack(m.(x), 1) .- y))
```

Here's where I had a problem: when there's more than one input/output sample, what's the derivative of ``L`` with respect to ``Wi`` (``m.cell.Wi``)?  Using some semi-random values:

```
x = [[0.3], [2.5]]
y = [0.5, 1.0]
m.cell.Wi .= [0.7]   #Wi
m.cell.Wh .= [0.001] #Wh
m.cell.b .= [0.85]   #b
m.state = [0.0]     #h0, h1, h2, etc.
```

If you evaluate ``m.(x)`` or ``L(x, y)``, you get the result you expect (``[[1.06], [2.60106]]`` and ``2.16106``, respectively).  For dL/dWi, it's easy to derive by inspection that it should be x1 + x2 + Wh*x1 = 2.8003.  You could also get it by finite difference:

```
q = L(x, y)
m.cell.Wi .+= 0.01
m.state = [0.0]
r = L(x, y)
abs(q - r)/0.01 # = 2.8003
```
But when you use ``gradient``:

```
m.state = [0.0]
m.cell.Wi .= [0.7]
g = gradient(() -> L(x, y), params(m.cell.Wi))
g[m.cell.Wi] # = 2.8025
```

This result is equal to x1 + x2 + Wh\*x2 instead of x1 + x2 + Wh*x1.  Am I overlooking something, or is something weird happening here?

### LINKED ISSUE #2185: explicit differentiation for RNN gives wrong results

### LINKED ISSUE #2258: Make RNNs blocked (and maybe fixing gradients along the way)
### Motivation and description

Given #2185 and other issues caused by the current mutability of the recur interface, we should move to a more standard blocked (i.e. 3D for simple RNN) interface. This has the benefits of: 
1) cleaning the recurrent interface so it is more easily used by people coming from other packages, 
2) more easily enable workflows using convRNNs,  and 
3) potentially enable some optimizations we can handle on the Flux side (see [Lux's Recurrence](https://github.com/LuxDL/Lux.jl/blob/89ff51a83d88da531feeb41797a49cd17f7f1fb2/src/layers/recurrent.jl#L3) return_sequence=true vs false)

I have not tested how we might fix the gradients by moving to this restricted interface. But if we decide to remove the statefulness (see below) we can fix gradients as seen in https://github.com/FluxML/Fluxperimental.jl/pull/7.

### Possible Implementation

I see two ways we can do this change, one which is a wider change of the Flux chain interface and another which tries to only fix Recur. In either case, the implementation would assume the final dimension of your multi-dimensional array is the time index. For a simple RNN it would assume the dimensions of the incoming array as: Features x Batch x Time. It will produce an error if a 2d array or 1d array is passed to recur, to avoid ambiguities. 

One possible implementation is to go ahead and do the full change over to removing state from the network generally. See https://github.com/FluxML/Fluxperimental.jl/pull/7. This would overhaul large parts of the interface into chain, and could be targeted at 0.14. See the implementation done in the above PR and https://github.com/FluxML/Fluxperimental.jl/pull/5 for details. 

The second possible approach is to just first remove the loop over timesteps interface and replace with the 3d interface. This initial change restricts the interface to be 3d, but I haven't tested how we could fix gradients while maintaining mutability and statefulness in Recur. The interface/impl would likely look much like: https://github.com/FluxML/Flux.jl/blob/c9c262db1c851cc612389f86854b1987083aab25/src/layers/recurrent.jl#L184-L188

### LINKED ISSUE #2341: Flux new explicit API does not work but old implicit API works for a simple RNN
I am trying to reproduce the tutorial [A Basic RNN](https://learningjulia.com/2019/10/11/a-basic-rnn.html) using Flux.jl v0.14.6. Using the old Flux API as in the tutorial, the model can be successfully trained. The code is

```julia
using Flux

num_samples = 1000
num_epochs = 50

function generate_data(num_samples)
    train_data = [reshape(rand(Float32.(1.0:10.0), rand(2:2:10)), 2, :) for i in 1:num_samples]
    train_labels = (v -> sum(v)).(train_data)

    test_data = 2 .* train_data
    test_labels = 2 .* train_labels

    train_data, train_labels, test_data, test_labels
end

train_data, train_labels, test_data, test_labels = generate_data(num_samples)

model = Flux.RNN(2, 1, (x -> x))

function eval_model(x)
    Flux.reset!(model)
    out = [model(view(x, :, t)) for t in axes(x, 2)]
    out[end] |> first
end

loss(x, y) = abs(sum(eval_model(x) .- y))

evalcb() = @show(sum(loss.(test_data, test_labels)))

ps = Flux.params(model)

opt = Flux.ADAM(0.1)

for epoch in 1:num_epochs
    Flux.train!(loss, ps, zip(train_data, train_labels), opt, cb = Flux.throttle(evalcb, 1))
end
```

However, refractor the above code to use the new explicit API, Zygote complains:

```
ERROR: LoadError: MethodError: no method matching +(::@NamedTuple{cell::@NamedTuple{σ::Nothing, Wi::Matrix{Float32}, Wh::Matrix{Float32}, b::Vector{Float32}, state0::Nothing}, state::Matrix{Float32}}, ::Base.RefValue{Any})
```

The code is as follows:

```julia
using Flux
using Statistics

num_samples = 1000
num_epochs = 50

function generate_data(num_samples)
    train_data = [reshape(rand(Float32.(1.0:10.0), rand(2:2:10)), 2, :) for i in 1:num_samples]
    train_labels = (v -> sum(v)).(train_data)

    test_data = 2 .* train_data
    test_labels = 2 .* train_labels

    train_data, train_labels, test_data, test_labels
end

train_data, train_labels, test_data, test_labels = generate_data(num_samples)

model = Flux.RNN(2, 1, (x -> x))

function eval_model(model, x)
    # Comment following line to make it run.
    # However, in the Flux doc, the following line is required.
    Flux.reset!(model)
    out = [model(view(x, :, t)) for t in axes(x, 2)]
    out[end] |> first
end

loss(model, x, y) = abs(sum(eval_model(model, x) .- y))

opt_state = Flux.setup(Flux.ADAM(0.1), model)

for epoch in 1:num_epochs
    for (x, y) in zip(train_data, train_labels)
        train_loss, grads = Flux.withgradient(model) do m
            loss(m, x, y)
        end
        Flux.update!(opt_state, model, grads[1])
    end
    test_loss = mean(loss.(Ref(model), test_data, test_labels))
    println("Epoch $epoch, loss = $test_loss")
end

# Following codes also failed to run.
# for epoch in 1:num_epochs
#     Flux.train!(model, zip(train_data, train_labels), opt_state) do m, x, y
#         loss(m, x, y)
#     end
# end
```

Julia version info
```
julia> versioninfo()
Julia Version 1.10.0-beta2
Commit a468aa198d0 (2023-08-17 06:27 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 128 × Intel(R) Xeon(R) Platinum 8362 CPU @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, icelake-server)
  Threads: 1 on 128 virtual cores
```

CarloLucibello:
A complete rework of our recurrent layers, making them more similar to their pytorch counterpart.
This is in line with the proposal in #1365 and should allow to hook into the cuDNN machinery (future PR). 
Hopefully, this ends the infinite source of troubles that the recurrent layers have been.

- `Recur` is no more. Mutating its internal state was a source of problems for AD (#2185)
- Now `RNNCell` is exported and takes care of the minimal recursion step, i.e. a single time:
    - has forward `cell(x , h)`
    -  `x` can be of size `in` or `in x batch_size`
    - `h` can be of size `out` or `out x batch_size` 
    - returns `hnew` of size `out` or `out x batch_size` 
 - `RNN` instead takes in a (batched) sequence and a (batched) hidden state and returns the hidden state for the whole sequence:
     - has forward `rnn(x, h)`
    -  `x` can be of size `in x len` or `in x len x batch_size`
    - `h` can be of size `out` or `out x batch_size`
    - returns `hnew` of size `out x len` or `out x len x batch_size` 
-  `LSTM` and `GRU` are similarly changed.

Close #2185, close #2341, close #2258, close #1547, close #807, close #1329

Related to #1678

### PR Checklist

- [x] cpu tests
- [x] gpu tests
- [x] if hidden state not given as input, assumed to be zero 
- [x] port `LSTM` and `GRU`
- [x] Entry in NEWS.md
- [x] Remove `reset!`
- [x] Docstrings
- [x] revisit documentation
- [ ] add an option in constructors to have trainable initial state? (future PR)
- [ ] use `cuDNN` (future PR)
- [ ]  implement the `num_layers` argument for stacked RNNs (future PR)
- [ ] add dropout (future PR)
- [ ] add bidirectional option (future PR)


darsnack:
Fully agree with updating the design to be non-mutating. There are two options we've discussed in the past:
1. `y, h = cell(x, h)` like here (I guess this PR removes `y` as a return value which is fine)
2. `y, cell = cell(x)` / `y, cell = Flux.apply(cell, x)`

Option 1 is outlined in this PR so I won't say anything about it.
Option 2 is a more drastic redesign to make all layers (not just recurrent) non-mutating. Why?
- Do a design that covers stateful layers in general (e.g. norm layers) and not just recurrent cells
- Keep a nice feature of Flux's current design which is that the model contains all info: parameters, state, flags, etc.

CarloLucibello:
I thought about Option 2. On the upside, it seems a nice intermediate spot between current Flux and Lux. The downside is that the interface would seem a bit exotic to flux and pytorch users. Moreover, it would be problematic for normalization layers.

Also, we need to distinguish between normalization layers and recurrent layers.

- Normalization layer at training time update some internal buffers, within a stopgrad barrier. The buffer update has no influence on the output of the layer and the final loss. You typically apply the layer only once during the forward pass. Normalization layers are typically part of larger models (chains or custom structs). Therefore for normalization layers: 1) we haven't had the gradient computation problems we had for recurrent layers; 2) you want the layer with the updated buffer to be inserted back in your model, but this would require a mutating operation or returning a new model. 

- For recurrent layers, Option 2 would be sensible, but is it worth it? Once you adopt the perspective that a cell takes two inputs, `x` and `h`, and gives back an output, `hnew`, all problems disappear. I think we add complexity for no gain in trying to keep the state internal.

ToucheSir:
The main benefit for keeping the state "internal" or having it be part of a unified interface like `apply` would be that `Chain` works with RNNs again. Whether that's worth the extra complexity is the question. Given our priorities, I think it's best left as future work.

CarloLucibello:
I think this is ready.

CarloLucibello:
I will merge this tomorrow if there are no further comments or objections.

PR #2509, opened 2024-10-28
Adjust docs & `Flux.@functor` for Functors.jl v0.5, plus misc. depwarns

### LINKED ISSUE #2533: `Flux.params` is broken
I know it's not really supported, but `params` is more broken than it should be (on master):
```julia
julia> Flux.params([2,3])
ERROR: UndefVarError: `IdSet` not defined

julia> @eval Flux using Base: IdSet

julia> Flux.params([2,3])
Params([[2, 3]])
```
It should probably print friendly upgrade instructions, `@warn "..." maxlog=1`. People calling this are probably struggling to update old code, and we should help them along.

mcabbott:
After https://github.com/FluxML/Functors.jl/pull/51 we should stop saying that you need either `@layer` or `@functor` for Flux to find your parameters.

This PR:
* defines a new macro `Flux.@functor` which prints a depwarn & then calls `@layer`
* ~~removes the code from `@layer` which printed other depwarns~~
* ~~removes the code from `@layer` which defines `functor`~~

Fixes #2533

WIP, more doc changes needed. 

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -133,8 +133,14 @@ Zygote._pullback(::Zygote.Context{true}, ::typeof(params), m...) = params(m), _ 
 
 include("optimise/Optimise.jl") ## deprecated Module
 
-Base.@deprecate_binding Optimiser OptimiserChain
-Base.@deprecate_binding ClipValue ClipGrad
+function Optimiser(rules...)
+  @warn "`Flux.Optimiser(...)` has been removed, please call `OptimiserChain(...)`, exported by Flux from Optimisers.jl" maxlog=1
+  OptimiserChain(rules...)
+end
+function ClipValue(val)
+  @warn "`Flux.ClipValue(...)` has been removed, please call `ClipGrad(...)`, exported by Flux from Optimisers.jl" maxlog=1
+  ClipGrad(val)
+end
 
 # TODO this friendly error should go in Optimisers.jl.
 # remove after https://github.com/FluxML/Optimisers.jl/pull/181
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -93,6 +93,29 @@ function params(m...)
   return ps
 end
 
+
+"""
+    @functor MyLayer
+
+Flux used to require the use of `Functors.@functor` to mark any new layer-like struct.
+This allowed it to explore inside the struct, and update any trainable parameters within.
+Flux@0.15 removes this requirement. This is because Functors@0.5 changed ist behaviour
+to be opt-out instead of opt-in. Arbitrary structs will now be explored without special marking.
+Hence calling `@functor` is no longer required.
+
+Calling `Flux.@layer MyLayer` is, however, still recommended. This adds various convenience methods
+for your layer type, such as pretty printing, and use with Adapt.jl.
+"""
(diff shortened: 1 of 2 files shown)


CarloLucibello:
Functors v0.5 is out

CarloLucibello:
We can continue with this now that Functors v0.5 is in with #2525 
This is the last item in the v0.15 milestone

CarloLucibello:
I subsumed most of this into #2528 in order to speed up the 0.15 release. 
I think only the `Flux.@functor` macro for smoothing the transition to `@layer` is left to be implemented.

mcabbott:
It's not so much to add functionality, as to avoid this:
```
julia> Flux.@functor 1
ERROR: LoadError: UndefVarError: `@functor` not defined
```
IMO that's an extremely unfriendly upgrade path. It would be OK to just print an error message, but doing what you ask as well is pretty easy.

mcabbott:
Yea, but better to prod people to update. Hence this PR.

src/deprecations.jl
```diff
@@ -83,16 +83,48 @@ function params!(p::Zygote.Params, x, seen = IdSet())
...
+
+To collect all the parameter arrays for other purposes, use `Flux.trainables(model)`.
+"""
 function params(m...)
-  Base.depwarn("""
-  Flux.params(m...) is deprecated. Use `Flux.trainable(model)` for parameters' collection
-  and the explicit `gradient(m -> loss(m, x, y), model)` for gradient computation.
-  """, :params)
+  @warn """`Flux.params(m...)` is deprecated. Use `Flux.trainable(model)` for parameter collection,
+  and the explicit `gradient(m -> loss(m, x, y), model)` for gradient computation.""" maxlog=1
```
mcabbott, line 99: `Base.depwarn` is silent except in tests. IMO that's ideal if you are marking something deprecated *before* a breaking change, when the replacement is available.

However, my impression is that we really want people to change this, not to silently live with old code during 0.15. So I'd like something to be printed in interactive use.

Maybe the same goes for more of the depwarns in this file?

src/deprecations.jl
```diff
@@ -119,9 +153,6 @@ end
 ### v0.16 deprecations ####################
 
 
-# Enable these when 0.16 is released, and delete const ClipGrad = Optimise.ClipValue etc: 
-# Base.@deprecate_binding Optimiser OptimiserChain
-# Base.@deprecate_binding ClipValue ClipGrad
```
mcabbott, line 124: These const definitions have already been deleted:
```
julia> Flux.ClipValue
ERROR: UndefVarError: `ClipValue` not defined
Stacktrace:
 [1] getproperty(x::Module, f::Symbol)
   @ Base ./Base.jl:31
 [2] top-level scope
   @ REPL[6]:1

julia> Flux.Optimiser
ERROR: UndefVarError: `Optimiser` not defined
Stacktrace:
 [1] getproperty(x::Module, f::Symbol)
   @ Base ./Base.jl:31
 [2] top-level scope
   @ REPL[7]:1
```

CarloLucibello (approved):

PR #2510, opened 2024-10-30
GPU docs

mcabbott:
This re-writes the start of the GPU documentation page. It aims to use simpler examples, and stress that `model |> cu` just works, before talking about more exotic non-CUDA packages, and the automatic `model |> gpu`. 

Rendered MD: https://github.com/FluxML/Flux.jl/blob/gpu_docs/docs/src/guide/gpu.md

Current docs: http://fluxml.ai/Flux.jl/stable/guide/gpu/


CarloLucibello:
removing the milestone as this shouldn't be blocking

mcabbott:
The point is, in part, to think through whatever interface we're adopting by trying to explain it clearly. If it's a mess then 0.15 is when to fix it. That's why it was on the milestone.

CarloLucibello:
Can we get this done so that it doesn't delay the 0.15 release then? I would like to tag in a few days

mcabbott:
Sure. What do you think this lacks?

docs/src/guide/gpu.md
```diff
@@ -1,140 +1,150 @@
...
-m(cu(rand(10)))
+cpu(cW) isa Array{Float32, 2}
+
+model2 = cpu(c_model)  # copy model back to CPU
+model2(x)
 ```
 
-As a convenience, Flux provides the `gpu` function to convert models and data to the GPU if one is available. By default, it'll do nothing. So, you can safely call `gpu` on some data or model (as shown below), and the code will not error, regardless of whether the GPU is available or not. If a GPU library (e.g. CUDA) loads successfully, `gpu` will move data from the CPU to the GPU. As is shown below, this will change the type of something like a regular array to a `CuArray`.
+!!! compat "Flux ≤ 0.13"
+    Old versions of Flux automatically loaded CUDA.jl to provide GPU support. Starting from Flux v0.14, it has to be  loaded separately. Julia's [package extensions](https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) allow Flux to automatically load some GPU-specific code when needed.
```
CarloLucibello, line 71:
```suggestion
```

docs/src/guide/gpu.md
```diff
@@ -1,140 +1,150 @@
...
-  0.851011  -0.586942
+```julia
+gpu(x) = gpu_device()(x)
+cpu(x) = cpu_device()(x)
 ```
 
 
+## Manually selecting devices
+
+I thought there was a whole `Flux.gpu_backend!` and Preferences.jl story we had to tell??
```
CarloLucibello, line 145: `gpu_backend!` affects the return from `gpu_device` like this:

- If no GPU is available, it returns a CPUDevice object.
- If a LocalPreferences file is present, then the backend specified in the file is used. If the trigger package corresponding to the device is not loaded, then a warning is displayed. 
- If no LocalPreferences file is present, then the first working GPU with loaded trigger package is used.

This is already described in the docstring of `gpu_device`. I think we shouldn't mention `gpu_backend!` at all in this guide because it is useless in practice.

CarloLucibello:
Maybe we should put a TLDR at the top just saying something like

```julia
using CUDA # or AMDGPU or Metal
device = gpu_device()
model = model |> device
for epoch in 1:num_epochs
    for (x, y) in dataloader
        x, y = device((x, y))
        ... compute gradients and update model ...
    end
end
```

CarloLucibello:
I will finish this

mcabbott:
> put a TLDR at the top

This is what we're trying not to do. Not "here's the magic boilerplate you should copy", but instead the guide is "here's how things actually work".

PR #2521, opened 2024-11-08
invert h and x arguments in RNNs

CarloLucibello:
In order to cause less breakage and avoid confusion such as in 
https://discourse.julialang.org/t/dimension-mismatch-in-flux-jl-rnn-example/122410
this PR reinstates the previous ordering for the arguments to a cell, i.e. `cell(h, x)` instead of the `cell(x, h)` we have on master.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@ See also [github's page](https://github.com/FluxML/Flux.jl/releases) for a compl
 
 ## v0.15.0 
 * Recurrent layers have undergone a complete redesign in [PR 2500](https://github.com/FluxML/Flux.jl/pull/2500).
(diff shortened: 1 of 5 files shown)


mcabbott (approved):

CarloLucibello:
I'm not sure we should do this, I like more the `cell(x, h)` convention, and also pytorch and lux do the same. 
It is true that we have `cell(h, x)` on v0.14, but the cells are not documented and exported, they only casually appear in the recurrent layers' docstrings https://fluxml.ai/Flux.jl/stable/reference/models/layers/#Flux.RNN

MartinuzziFrancesco:
Weighing in on this, I do think that the `cell(x, h)` convention is more intuitive and, as you pointed out, in line with other libraries. Also the rnn changes in 0.15 are breaking already, so all examples should be adapted anyway I assume

PR #2524, opened 2024-11-14
Port Rotary Positional Embedding from NeuralAttentionlib.jl

mashu:
This is pretty much standard thing to do these days with MultiHeadAttention layer, I think we should have it as part of Flux.
If anyone can review it, I would be happy. Too often I miss this.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -71,7 +71,7 @@ include("functor.jl")
 
 @compat(public, (
   # from OneHotArrays.jl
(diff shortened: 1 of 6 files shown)


CarloLucibello:
Why did you close this? It would make sense to have it. Actually you could file a PR to NNlib with the function generating the rotatory embedding.

CarloLucibello:
This is in a package now
https://github.com/mashu/PositionalEmbeddings.jl

PR #2532, opened 2024-11-24
set expand option as default for `@layer`

### LINKED ISSUE #2531: Consider making the `:expand` option the default in `@layer`
It doesn't seem to be really needed, we could automatically detect if the layer contains only leaf types or is a nested container

CarloLucibello:
This introduces the new show option `:noexpand` for `@layer` and sets the default to `:expand` instead.

Fix #2531

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -27,7 +27,7 @@ export gradient
                     CUDADevice, AMDGPUDevice, MetalDevice, oneAPIDevice,
                     XLADevice,
                     # get_device, # we define get_device here for retrocompatibility
-                    # gpu_backend!, # have to define here due to https://github.com/JuliaPackaging/Preferences.jl/issues/39
+                    gpu_backend!,
                     get_device_type,
                     DeviceIterator
 
@@ -118,7 +118,7 @@ include("losses/Losses.jl")
 using .Losses
 
 include("devices.jl")
-export get_device, gpu_backend!
+export get_device
 
 # Distributed Training
 include("distributed/backend.jl")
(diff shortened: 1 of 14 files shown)


src/Flux.jl
```diff
@@ -27,7 +27,7 @@ export gradient
                     CUDADevice, AMDGPUDevice, MetalDevice, oneAPIDevice,
                     XLADevice,
                     # get_device, # we define get_device here for retrocompatibility
-                    # gpu_backend!, # have to define here due to https://github.com/JuliaPackaging/Preferences.jl/issues/39
+                    gpu_backend!,
```
CarloLucibello, line 30: unrelated change: with v0.15 we don't need to define `gpu_backend!`, we can just reexport the one from MLDataDevices

mcabbott (changes requested):
These break, and probably others. Before `@layer` there was lots of special-case code to catch all of them. 
```
julia> LayerNorm(10)
LayerNorm(10)       # 20 parameters

julia> MultiHeadAttention(64 => 1024 => 1024, nheads = 8)
MultiHeadAttention(64 => 1024 => 1024; nheads=8)  # 1_245_184 parameters
```
This PR:
```
julia> LayerNorm(10)
LayerNorm(
  identity,
  Scale(10),                            # 20 parameters
  1.0f-5,
  10,
  true,
) 
```

CarloLucibello:
Ok now we have a `:noexpand` option but `:expand` becomes the default. I remove from the codebase (src+tests) 21 `:expand`, and added 2 `:noexpand` (LayerNorm + MultiHeadAttention). 

I'm not sure only LayerNorm and MultiHeadAttention need `:noexpand`, will do more testing.

PR #2536, opened 2024-12-03
Tweak quickstart.md

mcabbott:
* I think it's too obscure to say "cuDNN.jl has to be installed". We should just write the `Pkg.add` command out.
* The line `model(noisy |> gpu) |> cpu` is doing too many things, simplify?
* I don't like the "device = ..." pattern, but Carlo seems set on it, and added it here (instead of `gpu`). This at least tidies up a bit... most importantly never re-using `x, y` for things of different type. 
* Maybe the name of that should be a verb not a noun? Like `move = get_device()`? Not done yet.

### DIFF
--- a/docs/src/guide/models/quickstart.md
+++ b/docs/src/guide/models/quickstart.md
@@ -5,22 +5,20 @@ If you have used neural networks before, then this simple example might be helpf
 If you haven't, then you might prefer the [Fitting a Straight Line](overview.md) page.
 
 ```julia
-# This will prompt if neccessary to install everything, including CUDA.
-# For CUDA acceleration, also cuDNN.jl has to be installed in your environment.
+# Install everything, including CUDA, and load packages:
+using Pkg; Pkg.add(["Flux", "CUDA", "cuDNN", "ProgressMeter"])
 using Flux, CUDA, Statistics, ProgressMeter
+device = gpu_device()  # function to move data and model to the GPU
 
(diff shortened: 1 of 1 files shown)


CarloLucibello:
> Maybe the name of that should be a verb not a noun? Like move = get_device()? Not done yet

I would stick to device, it is very familiar to pytorch users, and `x |> device` is nice, `x |> move` is not, although `move(x)` could be better than `device(x)` .  `to_device = get_device()` could be another option, bu I still preferer device, also because `get_device()` returns an XDevice object,

CarloLucibello (approved):

docs/src/guide/models/quickstart.md
```diff
@@ -5,26 +5,24 @@ If you have used neural networks before, then this simple example might be helpf
 If you haven't, then you might prefer the [Fitting a Straight Line](overview.md) page.
 
 ```julia
-# This will prompt if neccessary to install everything, including CUDA.
-# For CUDA acceleration, also cuDNN.jl has to be installed in your environment.
+# Install everything, including CUDA, and load packages:
+using Pkg; Pkg.add(["Flux", "CUDA", "cuDNN", "ProgressMeter"])
 using Flux, CUDA, Statistics, ProgressMeter
```
mcabbott, line 10:
```suggestion
using Flux, Statistics, ProgressMeter
using CUDA  # optional
```

PR #2537, opened 2024-12-04
Remove usage of global variables in linear and logistic regression tutorial training functions

christiangnrd:
Based off Slack ["thread"](https://julialang.slack.com/archives/C7LFJTXV5/p1733262651938849).

I think we should bikeshed the following:

1. Do we rename all local function variables to be different from a global variable? If so, what should they be.
2. Add return statements to the functions to make it clear what is meant to be returned. Or just a `return` to make it clear that no output is expected from the `train_<something>!()` functions.

Personally, I think we should definitely do 1., and probably do 2.

### DIFF
--- a/docs/src/tutorials/linear_regression.md
+++ b/docs/src/tutorials/linear_regression.md
@@ -214,13 +214,13 @@ The loss went down! This means that we successfully trained our model for one ep
 Let's plug our super training logic inside a function and test it again -
 
 ```jldoctest linear_regression_simple; filter = r"[+-]?([0-9]*[.])?[0-9]+(f[+-]*[0-9])?"
-julia> function train_custom_model()
-           dLdW, dLdb, _, _ = gradient(custom_loss, W, b, x, y)
-           @. W = W - 0.1 * dLdW
-           @. b = b - 0.1 * dLdb
+julia> function train_custom_model!(f_loss, weights, biases, x_train, y_train)
+           dLdW, dLdb, _, _ = gradient(f_loss, weights, biases, x_train, y_train)
+           @. weights = weights - 0.1 * dLdW
+           @. biases = biases - 0.1 * dLdb
        end;
 
-julia> train_custom_model();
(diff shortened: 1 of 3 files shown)


mcabbott:
Thanks!

I'm in favour of 1, as I remember this being confusing. What the names should be I don't know, as having too many names for the same thing also gets messy. Probably  `_W, _b, _x, _y` are too far from plain text? Are `WW, bb, xx, yy` too ugly? `W_local, b_local, ...` too long although maybe the first time?

Not sure about 2. It's a bit weird to put `return` in when it's optional. Maybe my vote would be none (as it is now) but perhaps a `# this value is returned` on the first function whose return value we do want?

christiangnrd:
What about `weights, biases, data, targets`? Long but for a tutorial I feel like clarity is more important.

mcabbott:
I'm not so worried about length, as about obscuring the relation between the locals and the globals. 

One idea is to call globals full words like `weights, biases, data, targets`, and corresponding locals shortened versions like `wei, bias, dat, targ`? Too ugly? (Some word choices work better for this than others.) Maybe `w, b, d, t` as locals still preserves the link?

christiangnrd:
> One idea is to call globals full words like weights, biases, data, targets, and corresponding locals shortened versions like wei, bias, dat, targ? Too ugly? (Some word choices work better for this than others.) Maybe w, b, d, t as locals still preserves the link?

I agreed with you until I started making the changes and realized that the globals are named that way to match with the mathematical definitions from the tutorial, so I think they should remain as-is and we change the custom training function arguments to be a slightly more verbose term that aligns with their definition. 

Coming from a CS background, seeing that `W` is passing into a function as the `weights` argument would help me remember what `W` represents

mcabbott (approved):
Ok, good point about matching the equations. Then this is fine, I think.

mcabbott:
Documentation CI was failing here for other reasons, but once fixed, this tutorial now produces a failure:

Screenshot 2024-12-04 at 14 35 36

Issue #2514, opened 2024-11-04
feature requests for RNNs


CarloLucibello:
After the redesign in #2500, here is a list of potential improvements for recurrent layers and recurrent cells

- [ ] add an option in constructors to have trainable initial state
- [ ] use cuDNN
- [ ] implement the num_layers argument for stacked RNNs
- [ ] add dropout
- [ ] add bidirectional option

MartinuzziFrancesco:
Could it be possible to add to the list the option to use different initializers for the input matrix and recurrent matrix? This is provided by both [Keras/TF](https://www.tensorflow.org/api_docs/python/tf/keras/layers/LSTMCell) and [Flax](https://flax.readthedocs.io/en/latest/api_reference/flax.nnx/nn/recurrent.html#flax.nnx.nn.recurrent.LSTMCell).

This should be as straightforward as 

```julia
function RNNCell((in, out)::Pair, σ=relu;
    kernel_init = glorot_uniform,
    recurrent_kernel_init = glorot_uniform,
    bias = true)
    Wi = kernel_init(out, in)
    U = recurrent_kernel_init(out, out)
    b = create_bias(Wi, bias, size(Wi, 1))
    return RNNCell(σ, Wi, U, b)
end
```

I can also open a quick PR on this if needed

CarloLucibello:
yes! PR welcome

MartinuzziFrancesco:
Following up on this, should we also have an option to choose the init for the bias?

CarloLucibello:
We don't do it for feedforward layers, if someone wants a non-zero bias can just change it manually in the constructor, `layer.bias .= ...`

MartinuzziFrancesco:
Ehi, a couple of questions on this features request again. Would the `initialstates` function be like the current `function (rnn::RNN)(x::AbstractVecOrMat)` but returning just the state?

would something simple like
```julia
function initialstates(rnn::RNNCell; init_state = zeros)
    state = init_state(size(rnn.Wh, 2))
    return state
end

function initialstates(lstm::LSTMCell; init_state = zeros, init_cstate = zeros)
    state = init_state(size(lstm.Wh, 2))
    cstate = init_cstate(size(lstm.Wh, 2))
    return state, cstate
end
```
suffice or were you looking for something more? Maybe more control on the type would be needed

CarloLucibello:
I would just have
```
[...]

PR #2541, opened 2024-12-05
Adding initialstates function to RNNs

MartinuzziFrancesco:
Adding `initialstates` function as per discussion in #2514 

### PR Checklist

- [ ] Tests are added
- [ ] Entry in NEWS.md
- [ ] ~Documentation, if applicable~

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -69,6 +69,8 @@ end
 
 @layer RNNCell
 
+initialstates(rnn::RNNCell) = zeros_like(rnn.Wh, size(rnn.Wh, 2))
+
 function RNNCell(
   (in, out)::Pair,
   σ = tanh;
@@ -82,7 +84,10 @@ function RNNCell(
   return RNNCell(σ, Wi, Wh, b)
 end
 
-(m::RNNCell)(x::AbstractVecOrMat) = m(x, zeros_like(x, size(m.Wh, 1)))
+function (rnn::RNNCell)(x::AbstractVecOrMat)
+  state = initialstates(rnn)
+  rnn(x, state)
+end
 
 function (m::RNNCell)(x::AbstractVecOrMat, h::AbstractVecOrMat)
   _size_check(m, x, 1 => size(m.Wi, 2))
@@ -261,6 +266,10 @@ end
 
 @layer LSTMCell
 
+function initialstates(lstm:: LSTMCell) 
+  return zeros_like(lstm.Wh, size(lstm.Wh, 2)), zeros_like(lstm.Wh, size(lstm.Wh, 2))
+end
+
 function LSTMCell(
   (in, out)::Pair;
   init_kernel = glorot_uniform,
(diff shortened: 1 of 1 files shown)


src/layers/recurrent.jl
```diff
@@ -69,6 +69,8 @@ end
 
 @layer RNNCell
 
+initialstates(rnn::RNNCell) = zeros_like(rnn.Wh, size(rnn.Wh, 2))
```
CarloLucibello, line 72: should be documented, added to the `docs`, and added to the `public` symbols in `src/Flux.jl`.

src/layers/recurrent.jl
```diff
@@ -82,7 +84,10 @@ function RNNCell(
   return RNNCell(σ, Wi, Wh, b)
 end
 
-(m::RNNCell)(x::AbstractVecOrMat) = m(x, zeros_like(x, size(m.Wh, 1)))
+function (rnn::RNNCell)(x::AbstractVecOrMat)
+  state = initialstates(rnn)
+  rnn(x, state)
```
CarloLucibello, line 89:
```suggestion
  return rnn(x, state)
```

CarloLucibello:
It should be defined also for `RNN`, `LSTM`, ....

MartinuzziFrancesco:
> It should be defined also for `RNN`, `LSTM`, ....

Ah sorry I was meaning to ask, I'll go forward with that

src/layers/recurrent.jl
```diff
@@ -69,6 +69,41 @@ end
...
+
+# Arguments
+- `rnn`: The recurrent neural network cell or recurrent layer for
+  which the initial state vector is requested. It can be any of 
+  `RNNCell`, `RNN`, `LSTMCell`, `LSTM`, `GRUCell`, `GRU`, 
+  `GRUv3Cell`, and `GRUv3`
+
+# Returns
+An `AbstractVector` of zeros representing the initial hidden state, whose length 
+matches the output dimension of the cell or recurrent layer.
```
CarloLucibello, line 87:
```suggestion
Return the initial hidden state for the given recurrent cell or layer `rnn`.
```

CarloLucibello: We don't want to assume the state is a vector of zero, and we don't want to assume the method is limited to built-in layers.

src/layers/recurrent.jl
```diff
@@ -173,12 +211,17 @@ end
 
 @layer RNN
 
+initialstates(rnn::RNN) = zeros_like(rnn.cell.Wh, size(rnn.cell.Wh, 1))
```
CarloLucibello, line 214:
```suggestion
initialstates(rnn::RNN) = initialstates(rnn.cell)
```
and similarly for the other recurrent layers

CarloLucibello:
Should  test that the `Flux.initialstates(rnn)` returns what is supposed to returned for all layers.

src/layers/recurrent.jl
MartinuzziFrancesco: true but at the moment the initial states are hardcoded to be zeros, so this should still be conveyed to the user I think

CarloLucibello: it is conveyed by the layer's docstrings (where we could also add a reference to `initialstates`).

MartinuzziFrancesco: that is true, I'll add a reference!

docs/src/reference/models/layers.md
```diff
@@ -117,6 +117,7 @@ GRUCell
 GRU
 GRUv3Cell
 GRUv3
+initialstates
```
CarloLucibello, line 120:
```suggestion
Flux.initialstates
```
to make Docs CI happy

CarloLucibello (approved):

PR #2557, opened 2024-12-24
Adding return state option to recurrent layers

MartinuzziFrancesco:
Addressing the `return_state` as discussed in #2514.

### PR Checklist

- [x] Tests are added
- [ ] ~Entry in NEWS.md~

### DIFF
--- a/src/layers/recurrent.jl
+++ b/src/layers/recurrent.jl
@@ -239,7 +239,8 @@ The arguments of the forward pass are:
        If given, it is a vector of size `out` or a matrix of size `out x batch_size`.
        If not provided, it is assumed to be a vector of zeros, initialized by [`initialstates`](@ref).
 
-Returns all new hidden states `h_t` as an array of size `out x len x batch_size`.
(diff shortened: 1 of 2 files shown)


CarloLucibello:
tests are failing

MartinuzziFrancesco:
is this something that has to be fixed at the Functor level? the fail is at https://github.com/FluxML/Flux.jl/blob/master/test/test_utils.jl#L66 with

> MethodError: no method matching RNN(::RNNCell{typeof(tanh), Matrix{Float64}, Matrix{Float64}, Vector{Float64}})

so I guess it's trying to rebuild the struct but it can't since it doesn't have the return_state parametrization information. I can add a workaround like

```julia
RNN(rnn::RNNCell) = RNN{false, RNNCell}(rnn)
```

but this wouldn't solve it at the level that it's needed. The tests would still fail for `return_state = true`

CarloLucibello:
yes this needs a custom `functor` implementation. Maybe something like
```julia
Functors.functor(rnn::RNN{S}) = (; rnn.cell), ch -> RNN{S}(ch.cell)
```
(not tested)

CarloLucibello (approved):

PR #2563, opened 2024-12-31
fix test enzyme

CarloLucibello:
marks as broken an enzyme test that makes the CI fail. Also, moves the enzyme testing utilities to the `test_gradients` function.

### DIFF
--- a/test/ext_enzyme/enzyme.jl
+++ b/test/ext_enzyme/enzyme.jl
@@ -1,100 +1,7 @@
-using Test
-using Flux
-import Zygote
-
-using Enzyme: Enzyme, make_zero, Active, Duplicated, Const, ReverseWithPrimal
-
-using Functors
-using FiniteDifferences
-
-
-function gradient_fd(f, x...)
-    f = f |> f64
-    x = [cpu(x) for x in x]
-    ps_and_res = [x isa AbstractArray ? (x, identity) : Flux.destructure(x) for x in x]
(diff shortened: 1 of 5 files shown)


test/ext_enzyme/enzyme.jl
```diff
@@ -1,100 +1,8 @@
...
-        g = gradient_ez(loss, model, x)
-        @test g[1] isa SimpleDense
-        @test g[2] isa Array{Float32}
-        @test g[1].weight isa Array{Float32}
-        @test g[1].bias isa Array{Float32}
-        @test g[1].weight ≈ ones(Float32, 4, 1) .* x'
-        @test g[1].bias ≈ ones(Float32, 4)
-    end
-end
+using Enzyme: Enzyme, Duplicated, Const, Active
```
mcabbott, line 1: Maybe this PR is a good time to insert an `if VERSION < v"1.12-"` in front of `using Enzyme`? Then CI on nightly might start telling us things again.

mcabbott: Thanks! Want to rebase on #2569?

ToucheSir: Between Enzyme and CUDA, I wonder if we shouldn't move CI to test on prereleases instead of nightly. The signal-to-noise ratio of testing only nightly is quite low because of how often those two deps are broken on it.

mcabbott (approved):

PR #2576, opened 2025-01-05
Add reactant forward and reverse pass tests

wsmoses:
Note that Reactant will automatically select a GPU backend, if available, and TPU backend, if available. Otherwise it will fall back to CPU.

cc @CarloLucibello @mcabbott


Project.toml
```diff
@@ -18,6 +18,7 @@ Optimisers = "3bd65402-5787-11e9-1adc-39752487f4e2"
 Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 ProgressLogging = "33c8b6b6-d38a-422a-b730-caa89a2f386c"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
```
ToucheSir, line 21: Much like we do with Enzyme and the GPU libraries, installing Reactant on CI should be gated behind an env var (reusing `FLUX_TEST_ENZYME` is probably fine), and it should definitely not show up as a normal dep. As for why:

1. Any non-hard dep that tends to be sensitive to changes in Julia internals (i.e. anything which uses `Core.Compiler` or GPUCompiler a lot) needs to be cordoned off, lest it takes down CI runs for everything else.
2. Most of these libraries add significant compilation time, and our CI already struggles with that.
3. If for whatever reason there is not enough short- or long-term bandwidth available downstream to fix issues which pop up in these tests, it's much easier to disable them. We had this problem before where legacy integration tests keep CI red and make pushing through critical bug fixes all that more tedious.

mcabbott: More narrowly, this line is in the main Project.toml not the tests, so it's adding Reactant as a full-time dep, although it isn't loaded.

mcabbott: Agree about trying to decouple tests. In the test Project.toml, we do not declare dep on CUDA, but for some reason we do GPUCompiler? Which can't be installed on 1.12 right now, tests fail.

wsmoses: ah yeah sorry, I meant to open this as a draft PR to see folks appetite (and see if this is a reasonable place for adding inference and gradient tests). For some reason I don't have permission to change it to draft?

I'll fix up the dependency structure shortly, but I had done that locally for seeing what works or not offhand

wsmoses:
okay this is now ready, and all labeled tests pass on my machine (with the reactant I just released a few minutes ago).

Several tests here fail for some weird reasons. Some are clear and have issues already.

Others appear to be issues in flux itself (specifically the "all arguments must have at least the same length of the firs one" ones). 

We can go ahead and merge for now, but would you guys be able to figure out is going awry with the remaining ones?

CarloLucibello:
I can give it a try

test/runtests.jl
```diff
@@ -13,6 +13,8 @@ using Pkg
 using FiniteDifferences: FiniteDifferences
 using Functors: fmapstructure_with_path
 
+using Reactant
```
mcabbott, line 16: This loads it before the `if FLUX_TEST_REACTANT
  Pkg.add("Reactant")` below?

PR #2587, opened 2025-02-06
fix `cpu(dataloader)`

CarloLucibello:
Fix https://github.com/JuliaML/MLUtils.jl/issues/198


ToucheSir (approved):
I'm wondering if we should separate out many of the Enzyme and all of the Reactant tests to their own CI pipeline because they are so incredibly flaky. It really ruins the effectiveness of CI as a tool.

Issue #2506, opened 2024-10-25
Weights shape not validated against kernel, channels


BioTurboNick:
```
weights = Flux.kaiming_normal()(3, 3, 1)
Conv((3, 3), 1 => 1; pad = (1, 1), init = (_...) -> weights)
# Conv((3,), 3 => 1, pad=1)  # 10 parameters

weights = Flux.kaiming_normal()(3, 3, 1, 1)
Conv((3, 3), 1 => 1; pad = (1, 1), init = (_...) -> weights)
# Conv((3, 3), 1 => 1, pad=1)  # 10 parameters
```

I wanted to strictly specify the weight init for testing, but encountered this odd result. I think there should be validation to ensure that the weight shape matches the kernel size and input channels, and error if there is a mismatch.

CarloLucibello:
yes, definitely those sizes should be validated.

mcabbott:
You might be looking for `Conv(weights; pad=(1,1))`? I.e. there's a method which accepts `weights::Array`, for exactly this purpose. It does not take `(3, 3), 1 => 1` since, as you note, these are implied by the array size.

The methods which accept an `init` function certainly assume `size(init(s...)) == s`. Maybe they can all be made to check somehow but it does seem a somewhat strange path. 

> The initialisation of the weight matrix is `W = init(out, in)`, calling the function given to keyword init

BioTurboNick:
Ah, that's fair, thanks. Didn't think to look for a different method signature for this.

mcabbott:
The check if we do want one could look like this -- maybe not as messy as I pictured it being, at first:
```julia
function _sizecheck(f, sz::Integer...)
  W = f(sz...)
  size(W) == sz || error("bad size! (except more friendly)")
  W
end

function Dense((in, out)::Pair{<:Integer, <:Integer}, σ = identity;
               init = glorot_uniform, bias = true)
  # Dense(init(out, in), bias, σ)  # current code
  Dense(_sizecheck(init, out, in), bias, σ)  # with new check
end
```
(When you pass an array `bias = [1,2,3.]` to layer constructors, I think it is always checked for size.)

CarloLucibello:
This is already implemented — `Flux._sizecheck` validates that an `init` function returns an array of the requested size, and the example in the OP now errors:

```julia
julia> weights = Flux.kaiming_normal()(3, 3, 1);

julia> Conv((3, 3), 1 => 1; pad = (1, 1), init = (_...) -> weights)
[...]

PR #2590, opened 2025-02-13
Added shape validation for Conv weight tensor

mikymatt01:
This pull request addresses an issue that was raised in [Flux.jl Issue #2506](https://github.com/FluxML/Flux.jl/issues/2506).
### Problem
Initially, when we overwrote the `init` function with a tensor of a shape different from the one specified by the `k` and `ch` parameters, the `Conv` function would create a filter using the shape of the `init` tensor instead of using the shape defined by the input parameters. This led to unexpected behavior.

### Solution
In this modification, if the shape of the created weight tensor does not match the expected shape (k..., ch.first ÷ groups, ch.second), an error is now raised. This ensures that the filter shape is consistent with the input parameters, preventing potential mismatches in convolution operations.
```
weights = Flux.kaiming_normal()(3, 3, 1)
Conv((3, 3), 1 => 1; pad = (1, 1), init = (_...) -> weights)
# Expected shape: (3, 3, 1, 1), got (3, 3, 1), error

weights = Flux.kaiming_normal()(3, 3, 1, 1)
Conv((3, 3), 1 => 1; pad = (1, 1), init = (_...) -> weights)
# Expected shape: (3, 3, 1, 1), got (3, 3, 1, 1), ok
```

### DIFF
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -515,7 +515,6 @@ If the shape does not match, a `DimensionMismatch` error is thrown.
 
 This is internally used to validate weight initialization functions.
 """
-
 function _sizecheck(f, sz::Integer...)
   W = f(sz...)
   size(W) == sz || throw(DimensionMismatch(
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -518,10 +518,9 @@ This is internally used to validate weight initialization functions.
 
 function _sizecheck(f, sz::Integer...)
   W = f(sz...)
-  err = DimensionMismatch(
+  size(W) == sz || throw(DimensionMismatch(
       "Weight shape mismatch: expected $(sz), got $(size(W))",
-  )
-  size(W) == sz || throw(err)
+  ))
   W
 end
 
--- a/src/layers/conv.jl
+++ b/src/layers/conv.jl
@@ -176,12 +176,6 @@ distribution.
 
 This is internally used by the [`Conv`](@ref) layer.
 """
-function _sizecheck(f, sz::Integer...)
-    W = f(sz...)
-    err = DimensionMismatch("Weight shape mismatch: expected $(sz), got $(size(W))")
-    size(W) == sz || throw(err)
-    W
-end
 
 function convfilter(filter::NTuple{N,Integer}, ch::Pair{<:Integer,<:Integer};
           init = glorot_uniform, groups = 1) where N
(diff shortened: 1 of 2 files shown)


mcabbott:
I think the right place for such a check is `function convfilter`. That's where init is called, and the arguments with which it is called are the desired size. Then there is exactly one place this size is computed, and the error-code can't be out of sync.

I would suggest wrapping this error into a function [like this](https://github.com/FluxML/Flux.jl/issues/2506#issuecomment-2439068099). That function can then be called everywhere that an `init` kw is used.

The PR will also want to contain some test for this. Something like your example, or like `@test_throws Exception Dense(3 => 4; init = (_...) -> rand(2, 2))`.

ToucheSir:
Another suggestion is to throw a `DimensionMismatch` instead of a plain `error`. We already do this elsewhere in Flux.

src/layers/conv.jl
```diff
@@ -176,12 +176,21 @@ distribution.
 
 This is internally used by the [`Conv`](@ref) layer.
 """
+function _sizecheck(f, sz::Integer...)
```
mcabbott, line 179: Note that inserting this here attaches the docstring above for convfilter to this function instead. 

Maybe it should live next to `_conv_size_check`? It might be good to have a short docstring saying what it does.

src/layers/conv.jl
```diff
@@ -501,6 +504,27 @@ function _conv_size_check(layer, x::AbstractArray)
...
+
+This is internally used to validate weight initialization functions.
+"""
+
+function _sizecheck(f, sz::Integer...)
+  W = f(sz...)
+  err = DimensionMismatch(
+      "Weight shape mismatch: expected $(sz), got $(size(W))",
+  )
+  size(W) == sz || throw(err)
```
mcabbott, line 524: I think it's slightly better to only make the string if the sizes don't match:
```suggestion
  size(W) == sz || throw(DimensionMismatch(
      "Weight shape mismatch: expected $(sz), got $(size(W))",
  ))
```

src/layers/conv.jl
```diff
@@ -501,6 +504,27 @@ function _conv_size_check(layer, x::AbstractArray)
...
+
+Ensures that the output of `f(sz...)` has the expected shape `sz`.
+
+Constructs a tensor using the function `f` with the given size `sz` and verifies that its shape matches `sz`.  
+If the shape does not match, a `DimensionMismatch` error is thrown.
+
+This is internally used to validate weight initialization functions.
+"""
+
+function _sizecheck(f, sz::Integer...)
```
mcabbott, line 519: Maybe no space here?
```suggestion
"""
function _sizecheck(f, sz::Integer...)
```

mcabbott (approved):
Thanks! Sorry this took a while...

PR #2595, opened 2025-03-13
Disable broken Reactant tests for now

ToucheSir:
These have been broken for weeks, and no fix appears to be forthcoming. Meanwhile, these tests have rendered both Flux's CI and upstream reverse CI completely unreliable.

A better fix would be to add a way to mark broken tests in the utility function the Reactant tests use, much like has been done elsewhere in FluxML. But for now, stop the bleeding first.

### DIFF
--- a/test/ext_reactant/reactant.jl
+++ b/test/ext_reactant/reactant.jl
@@ -89,7 +89,7 @@ end
     for (model, x, name) in models_xs
         @testset "check grad $name" begin
             println("testing $name with Reactant")
-            test_gradients(model, x; loss, compare_finite_diff=false, test_reactant=true)
+            # test_gradients(model, x; loss, compare_finite_diff=false, test_reactant=true)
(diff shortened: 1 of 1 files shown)


mcabbott (approved):

wsmoses:
oh I didn't know these were failing at all (and thus didn't know there was a fix needed).

In the future, can you open an issue if any breakage is found?

ToucheSir:
As I recall, the ask was to have new issues come with an MWE. At present, I don't think there's enough capacity to do that for the current failures, let alone every time we spot a regression in CI. Since https://github.com/EnzymeAD/Reactant.jl/blob/0f2b278c67d36c198bc042ec13ff825039c84018/test/nn/flux.jl already exists, perhaps a better long-term solution would be to add a minimal test with a difficult model (e.g. an RNN) in there.

PR #2596, opened 2025-03-15
Test re enable reactant test to identify failures

wsmoses:


avik-pal:
https://github.com/FluxML/Flux.jl/actions/runs/13876141492/job/38828690662?pr=2596#step:6:1956 that's an ancient version of reactant

wsmoses:
@mcabbott @ToucheSir did anything change in your compat which led to https://github.com/FluxML/Flux.jl/pull/2595

indeed as @avik-pal that is an ancient version of Reactant.

To prevent this from happening perhaps we could actually add it as a test dep and version bound it?

wsmoses:
on my laptop, where it picks up a later reactant, the tests indeed pass

ToucheSir:
> @mcabbott @ToucheSir did anything change in your compat which led to #2595

Nothing, which is the worrying part. Did you change anything in `test/Project.toml` locally?

I suspect the problem is https://github.com/FluxML/Flux.jl/blob/fa108bb994a2cc839240d8497c9e1610818a49ab/test/Project.toml#L24-L27, which is locking in an old version of GPUCompiler. That whole compat section should be cleaned up, or at the very least GPUCompiler and GPUArraysCore should be removed as test deps.


> To prevent this from happening perhaps we could actually add it as a test dep and version bound it?

The problem is that we run CI on nightly, and packages which rely on GPUCompiler frequently break during precompilation on nightly. I see two paths here:
1. Add dependencies as enabled by env vars like we do now, while maintaining some form of compat check for them to prevent this type of downgrade issue.
2. Declare compat bounds and test deps for all optional test packages, but somehow remove them as test deps before precompilation happens on CI.

I think (2) is cleaner, but it's not clearly what exactly that'd look like.

wsmoses:
confirmed, removing that fixed it here!

Issue #2584, opened 2025-01-18
Some models saved under Flux v0.14 do not load on v0.15+


dorn-gerhard:
Using `u = Flux.loadmodel!(empty_model, model_state)`
leads to the following error

![Image](https://github.com/user-attachments/assets/f37a26e8-e617-448d-9003-b0396f08948a)
since in Flux v0.14 `pad_size` is not stored extra (`model_state2`) whereas in the new version it is (`model_state`)

![Image](https://github.com/user-attachments/assets/00f28f25-e6c8-4565-9b31-10725f86360c)

Is there a workaround to upgrade old model states?

ToucheSir:
I don't recall any built-in layers which have a `pad_size` field, can you provide a MWE? Depending on what `pad_size` represents, I think the easiest solution could be to use Accessors.jl + maybe Functors.jl to remove `pad_size` from `model_state`.

mcabbott:
A possible cause is Functors v0.5 recursing into arbitrary custom structs?

For built-in layers, Flux.state does typically save integers like padding, does not load them (since they are immutable), but does complain if they don't match:
```julia
julia> st = Flux.state(Conv((2,), 1=>1, pad=1, stride=2))
(σ = (), weight = Float32[-0.8863135; -0.70230323;;;], bias = Float32[0.0], stride = (2,), pad = (1, 1), dilation = (1,), groups = 1)

julia> Flux.loadmodel!(Conv((2,), 1=>1, pad=0, stride=0), st)
Conv((2,), 1 => 1, stride=0)  # 3 parameters

julia> Flux.loadmodel!(Conv((2,), 1=>1, pad=0, stride=0), st[propertynames(st)[1:end-1]])
ERROR: ArgumentError: Tried to load (:pad, :σ, :weight, :bias, :stride, :dilation) into (:pad, :σ, :weight, :bias, :groups, :stride, :dilation) but the structures do not match.
```

One possible remedy might be to define a method of `loadmodel!` for your struct, as was done when some built-in layers changed what fields they had:

https://github.com/FluxML/Flux.jl/blob/44695a0b4fe7ffbd04d10030cd70a1928e41f7f9/src/deprecations.jl#L24-L28

CarloLucibello:
Closing as not reproducible as filed — no MWE was provided despite a couple of requests, and `pad_size` is not a field of any built-in Flux layer, so this involves a custom struct.

[...]

PR #2598, opened 2025-03-31
Fix loading of pooling layers

adrhill:
As [discussed on Slack](https://julialang.slack.com/archives/C690QRAA3/p1741632284486259), I tried my hand at fixing #2584 and https://github.com/FluxML/Metalhead.jl/issues/287.
This line of code appears to be enough to make Metalhead's `VGG` models load.

Without it, the following error is thrown:
```Julia-repl
julia> using Metalhead

julia> VGG(19; pretrain=true)
ERROR: ArgumentError: Tried to load Base.OneTo(0) into (:pad, :k, :stride) but the structures do not match.
Stacktrace:
  [1] loadmodel!(dst::MaxPool{2, 4}, src::Tuple{}; filter::Function, cache::IdSet{Any})
    @ Flux ~/Developer/Flux.jl/src/loading.jl:104
  [2] loadmodel!(dst::Tuple{…}, src::Tuple{…}; filter::Function, cache::IdSet{…})
    @ Flux ~/Developer/Flux.jl/src/loading.jl:118
  [3] loadmodel!(dst::Chain{Tuple{…}}, src::@NamedTuple{layers::Tuple{…}}; filter::Function, cache::IdSet{Any})
    @ Flux ~/Developer/Flux.jl/src/loading.jl:118
```

I could use some guidance on whether this fix is adequate and how to test it.

### DIFF
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -168,3 +168,11 @@ end
 #   """
 # ))
 
+# In Flux <= v0.14, pooling layers contained no state.
+# These workarounds allow such pooling layers to be loaded in Flux v0.16. 
+# (https://github.com/FluxML/Flux.jl/pull/2598)
+loadmodel!(dst::MaxPool{N, M}, src::Tuple{}; kw...) where {N, M} = dst
+loadmodel!(dst::AdaptiveMaxPool{S, O}, src::Tuple{}; kw...) where {S, O} = dst
+
+loadmodel!(dst::MeanPool{N, M}, src::Tuple{}; kw...) where {N, M} = dst
+loadmodel!(dst::AdaptiveMeanPool{S, O}, src::Tuple{}; kw...) where {S, O} = dst
--- a/src/deprecations.jl
+++ b/src/deprecations.jl
@@ -169,10 +169,9 @@ end
 # ))
 
 # In Flux <= v0.14, pooling layers contained no state.
-# These workarounds allow such pooling layers to be loaded in Flux v0.16. 
+# These workarounds allow such pooling layers to be loaded in Flux >= v0.16. 
 # (https://github.com/FluxML/Flux.jl/pull/2598)
 loadmodel!(dst::MaxPool{N, M}, src::Tuple{}; kw...) where {N, M} = dst
-loadmodel!(dst::AdaptiveMaxPool{S, O}, src::Tuple{}; kw...) where {S, O} = dst
-
 loadmodel!(dst::MeanPool{N, M}, src::Tuple{}; kw...) where {N, M} = dst
+loadmodel!(dst::AdaptiveMaxPool{S, O}, src::Tuple{}; kw...) where {S, O} = dst
 loadmodel!(dst::AdaptiveMeanPool{S, O}, src::Tuple{}; kw...) where {S, O} = dst
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -109,8 +109,6 @@ function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
   return dst
 end
 
-loadmodel!(dst::MaxPool{N, M}, src::Tuple{}; kw...) where {N, M} = dst
-
 """
     state(x)
 
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -109,6 +109,8 @@ function loadmodel!(dst, src; filter = _ -> true, cache = Base.IdSet())
   return dst
 end
 
+loadmodel!(dst::MaxPool{N, M}, src::Tuple{}; kw...) where {N, M} = dst
+
 """
     state(x)
 


mcabbott:
To be clear the before-and-after states are
```julia
julia> Flux.state(MaxPool((2,3)))
()

(jl_gaQBh1) pkg> st Flux
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_gaQBh1/Project.toml`
⌃ [587475ba] Flux v0.14.25
```
and
```julia
julia> Flux.state(MaxPool((2,3)))
(k = (2, 3), pad = (0, 0, 0, 0), stride = (2, 3))

(@v1.11) pkg> st Flux
Status `~/.julia/environments/v1.11/Project.toml`
  [587475ba] Flux v0.16.3
```
So this looks fine. Might be worth getting a few more layers which will have the same problem, e.g. https://fluxml.ai/Flux.jl/dev/reference/models/layers/#Pooling

adrhill:
For the sake of documentation:

Flux `v0.14`:
```Julia-repl
julia> Flux.state(MaxPool((5, 5)))
()

julia> Flux.state(MeanPool((5, 5)))
()

julia> Flux.state(AdaptiveMaxPool((5, 5)))
()

julia> Flux.state(AdaptiveMeanPool((5, 5)))
()

julia> Flux.state(GlobalMaxPool())
()

julia> Flux.state(GlobalMeanPool())
()
```

Flux `v0.16`:

```Julia-repl
julia> Flux.state(MaxPool((5, 5)))
(k = (5, 5), pad = (0, 0, 0, 0), stride = (5, 5))

julia> Flux.state(MeanPool((5, 5)))
(k = (5, 5), pad = (0, 0, 0, 0), stride = (5, 5))

julia> Flux.state(AdaptiveMaxPool((5, 5)))
(out = (5, 5),)

julia> Flux.state(AdaptiveMeanPool((5, 5)))
(out = (5, 5),)

julia> Flux.state(GlobalMaxPool())
()

julia> Flux.state(GlobalMeanPool())
()
```

So I guess the global pooling layers require no modification?

mcabbott (approved):

adrhill:
Could we get a patch release with this fix? I've got some students that will be using Metalhead very soon.

mcabbott:
Can you post the exact error? And ideally what `Flux.state(the_layer)` returns on Flux 0.14 and 0.16, as above?

adrhill:
I think this is largely a Metalhead issue, due to the recent introduction of  `PartialFunctions` (see issues https://github.com/FluxML/Metalhead.jl/issues/286, https://github.com/FluxML/Metalhead.jl/issues/287). 

As I've mentioned in https://github.com/FluxML/Metalhead.jl/issues/286#issuecomment-2850866971, it would be great if all layer types for common pre-trained architectures lived in Flux. Since Julia is heavily dispatch based, my packages otherwise end up requiring a strict dependency on Metalhead. This would also allow for bugs like https://github.com/FluxML/Metalhead.jl/issues/287 to be caught in Flux' testing.

For this specific problem, it would be nice to have a simple `AddAct` layer type instead of having to dispatch on the awful looking `PartialFunctions.PartialFunction{nothing, nothing, typeof(Metalhead.addact), Tuple{typeof(relu)}, @NamedTuple{}}.` type.

mcabbott:
If I understand right, you want this to happen:
```julia
julia> using Metalhead: addact, relu

julia> addact(relu, [1,2,3], [4,-3,-2])
3-element Vector{Int64}:
 5
 0
 1
```
and the `PartialFunction` is very close to being `Base.Fix1(addact, relu)`, except that this only accepts one argument for now... Julia 1.12 will change this:
```julia
julia> _addact(activation, xs...) = activation.(sum(xs));  # my version broadcasts, Metalhead does not

julia> addtanh = Base.Fix1(_addact, tanh)
(::Base.Fix1{typeof(_addact), typeof(tanh)}) (generic function with 2 methods)

julia> addtanh([1,2,3], [4,-3,-2])
3-element Vector{Float64}:
  0.9999092042625951
 -0.7615941559557649
  0.7615941559557649

julia> VERSION
v"1.12.0-beta3"
```
May I ask why not just use Base's ComposedFunction for this?
```julia
julia> addrelu = relu ∘ (+);

julia> addrelu([1,2,3], [4,-3,-2])
3-element Vector{Int64}:
 5
 0
 1
```
Like `Metalhead.addact`, this relies on the auto-broadcasting `relu(::AbstractArray)` method, so won't work for `tanh`:
```julia
julia> Metalhead.addact(tanh, [1,2,3], [4,-3,-2])
ERROR: MethodError: no method matching tanh(::Vector{Int64})

julia> (tanh ∘ (+))([1,2,3], [4,-3,-2])
ERROR: MethodError: no method matching tanh(::Vector{Int64})
```
This also appears to have the same state on old & new Flux:
```julia
julia> Flux.state(relu ∘ (+))  # Flux v0.14.25, and also Flux v0.16.3
(outer = (), inner = ())
```

adrhill:
> May I ask why not just use Base's ComposedFunction for this?

This is code from Metalhead.jl's ResNet implementation: https://github.com/FluxML/Metalhead.jl/issues/287#issuecomment-2934086139

PR #2600, opened 2025-04-19
Enable other reactant tests

wsmoses:


mcabbott:
From reading the log... error is from this utility function:

https://github.com/FluxML/Flux.jl/blob/0e36af98f6fc5b7f3c95fe819a02172cfaaaf777/test/test_utils.jl#L38-L44

Seems to be called with 1st argument Zygote gradient, and 2nd argument Enzyme gradient:

https://github.com/FluxML/Flux.jl/blob/0e36af98f6fc5b7f3c95fe819a02172cfaaaf777/test/test_utils.jl#L109

And Zygote will often truncate branches which are like `(nothing, nothing)`. I guess that's what's happening?

Signature in another error is other order... but it looks like Enzyme has left Tuple{Int64, Int64} in the gradient, where Zygote has nothing. Then in the next recursion inwards, you could get `map(..., x::Tuple{Int64, Int64}, ys::Nothing)` as above, I think?

```
map(f::Function, 
   t1::NTuple{6, KeyPath{Tuple{Int64, Symbol}}}, # who knows why, ignore
   t2::Tuple{Vector{Float32}, Tuple{Int64, Int64}, NTuple{4, Int64}, Tuple{Int64, Int64}, Tuple{Int64, Int64}, Int64},  # presumably Enzyme
   ts::Tuple{Vector{Float32}, Vararg{Nothing, 5}},  # presumably Zygote
)
```

wsmoses:
yeah Enzyme keeps the primal and shadow structures the same, so it sounds like the test works -- but the test function itself is busted here?

mcabbott:
Yes, that's right. Sorry I stopped before the conclusion!

Somehow the function for doing approximate equality on nested structs needs to be made smart enough to handle the difference between Zygote & Enzyme conventions.

mcabbott:
Sorry I don't have bandwidth to think about this.

If anyone else does, trying to simplify gradient tests would be very welcome. Zygote may be on the way out, so far there appears to be nobody trying to make it run on 1.12. Possibly working on tests and on removing this would best be done together.

wsmoses:
@mcabbott @CarloLucibello this should now pass, if it can get a review/merge?

test/ext_reactant/reactant.jl
```diff
@@ -51,13 +41,9 @@ end
     end
 
     models_xs = [
-        # %23 = "stablehlo.gather"(%22, %0) <{dimension_numbers = #stablehlo.gather<offset_dims = [1], collapsed_slice_dims = [0], start_index_map = [0, 1], index_vector_dim = 1>, indices_are_sorted = false, slice_sizes = array<i64: 1, 1>}> : (tensor<2x10xf32>, tensor<1x2xi64>) -> tensor<1x1xf32>
-        # (first ∘ LayerNorm(2), randn(Float32, 2, 10), "LayerNorm"), # Zygote comparison test fails on the GPUArraysCore.@allowscalar in scalarfirst, so we globally allow scalar
+        (first ∘ LayerNorm(2), randn(Float32, 2, 10), "LayerNorm"), # Zygote comparison test fails on the GPUArraysCore.@allowscalar in scalarfirst, so we globally allow scalar
```
CarloLucibello:
```suggestion
        (LayerNorm(2), randn(Float32, 2, 10), "LayerNorm"), # Zygote comparison test fails on the GPUArraysCore.@allowscalar in scalarfirst, so we globally allow scalar
```

PR #2603, opened 2025-05-18
reinstate enzyme tests

CarloLucibello:


test/train.jl
```diff
@@ -140,6 +136,7 @@ for (trainfn!, name) in ((Flux.train!, "Zygote"), (train_enzyme!, "Enzyme"))
       l2 = sum(pen2, Flux.trainables(m))
       err + 0.33 * l2
     end
```
CarloLucibello, line 138: @wsmoses without the change ReverseWithPrimal -> set_runtime_activity(ReverseWithPrimal) made above this gradient fails

wsmoses: Hm when were these tests disabled (and do you know why).

my guess offhand is this is another consequence of Julia 1.11’s now triple pointer array causing performance and analysis to be worse (assuming it worked without on 1.10)

CarloLucibello: I don't know, they were disabled already prior to #2602

Is it ok to keep set_runtime_activity in all our gradient infrastructure?

test/ext_enzyme/enzyme.jl
```diff
@@ -23,7 +25,13 @@
     for (model, x, name) in models_xs
         @testset "Enzyme grad check $name" begin
             println("testing $name with Enzyme")
-            test_gradients(model, x; loss, compare_finite_diff=false, test_enzyme=true)
+            if name ∈ ["Bilinear", "MultiHeadAttention"]
+                # these are failing with error:
+                # Error handling recursive stores for Symbol which has a fieldcount of 0
+                @test_broken test_gradients(model, x; loss, compare_finite_diff=false, test_enzyme=true)
```
CarloLucibello, line 31: @wsmoses these also used to work

PR #2635, opened 2025-12-08
add opencl support

VarLad:
imports and re-exports OpenCL from MLDataDevices, for usage in Flux.jl
Depends on https://github.com/LuxDL/Lux.jl/pull/1590


VarLad:
@CarloLucibello Should I bump the minimum required version of MLDataDevices as well?

Issue #2640, opened 2025-12-22
Replacing Zygote with Mooncake


yebai:
[Mooncake](https://github.com/chalk-lab/Mooncake.jl) is intended as a drop-in replacement for Zygote, and the Mooncake team is happy to address GPU-related issues.

CarloLucibello:
It is nice to see that these tests for Mooncake are passing

```julia
using Functors, ADTypes, Mooncake, Zygote
import DifferentiationInterface as DI
using Test
using Flux
using Random, Statistics, LinearAlgebra

function mooncake_withgradient(loss, x...)
    y, gs = DI.value_and_gradient(x -> loss(x...), AutoMooncake(), x)
    return y, _unwrap.(gs)
end

_unwrap(x) = fmap(_unwrap, x, exclude= y -> y isa Mooncake.Tangent)
_unwrap(x::Mooncake.Tangent) = _unwrap(x.fields)
_unwrap(x::Mooncake.MutableTangent) = _unwrap(x.fields)

function check_equal_leaves(a, b; rtol=1e-4, atol=1e-4)
    fmapstructure_with_path(a, b) do kp, x, y
        # @show kp
        if x isa AbstractArray
            @test x ≈ y rtol=rtol atol=atol
        elseif x isa Number
            @test x ≈ y rtol=rtol atol=atol
        end
    end
end

function test_mooncake_gradients(
            f, 
            xs...;
            rtol=1e-4, atol=1e-4,
            test_gpu = false,
            test_grad_f = true,
            test_grad_x = true,
            loss = (f, xs...) -> mean(f(xs...)),
            )


    Flux.trainmode!(f) # for layers like BatchNorm

    ## Let's make sure first that the forward pass works.
    l = loss(f, xs...)
    @test l isa Number
    if test_gpu
        gpu_dev = gpu_device(force=true)
        cpu_dev = cpu_device()
        xs_gpu = xs |> gpu_dev
        f_gpu = f |> gpu_dev
        l_gpu = loss(f_gpu, xs_gpu...)
        @test l_gpu isa Number
    end

    if test_grad_x
        y, g = Zygote.withgradient((xs...) -> loss(f, xs...), xs...)
        
        y_ez, g_ez = mooncake_withgradient((xs...) -> loss(f, xs...), xs...)
        @test y ≈ y_ez rtol=rtol atol=atol
        check_equal_leaves(g, g_ez; rtol, atol)
        
        if test_gpu
[...]

PR #2645, opened 2025-12-26
extend `gradient` to take an ADType argument

CarloLucibello:
Related to #2640 
cc @gdalle

In this PR we do the following
- Add the new dependency `ADTypes`.
- Introduce the methods `Flux.gradient(f, adtype, x...)` and `Flux.withgradient(f, adtype, x...)`, where adtype should be one of `AutoZygote()`, `AutoMooncake()`, `AutoEnzyme()`, `AutoFiniteDifferences`.
- Revamp the `test_gradients` util to be AD backend agnostic. Also address the comment https://github.com/FluxML/Flux.jl/pull/2600#issuecomment-2855146386, making the test function more robust.
- Add a Mooncake extension and corresponding tests

TODO:
- [x] news
- [x] docs
- [x] more tests
- [x] reactant

### DIFF
--- a/test/ext_reactant/reactant.jl
+++ b/test/ext_reactant/reactant.jl
@@ -1,5 +1,5 @@
 @testset "Reactant Models" begin
-    broken_models = (,)
+    broken_models = ()
     for (model, x, name) in TEST_MODELS
         @testset "Reactant grad check $name" begin
             @test test_gradients(model, x; reference=AutoZygote(), test_reactant=true, test_cpu=false)  broken=(name ∈ broken_models)


ext/FluxMooncakeExt.jl
```diff
@@ -0,0 +1,33 @@
...
+end
+
+function Flux.withgradient(f, adtype::AutoMooncake, x...)
+    val, grad = DI.value_and_gradient(x -> f(x...), adtype, x)
+    return (val=val, grad=_unwrap(grad))
+end
+
+_unwrap(x) = fmap(_unwrap, x, exclude= y -> y isa Union{Tangent, MutableTangent})
+_unwrap(x::Tangent) = _unwrap(x.fields)
+_unwrap(x::MutableTangent) = _unwrap(x.fields)
```
gdalle: This is the only part that seems specific to Mooncake, the rest could just be a DIExt?

ext/FluxMooncakeExt.jl
```diff
@@ -0,0 +1,33 @@
...
+    g = DI.gradient(x -> f(x...), adtype, x)
+    return _unwrap(g)
+end
+
+function Flux.withgradient(f, adtype::AutoMooncake, x)
+    val, grad = DI.value_and_gradient(f, adtype, x)
+    return (val=val, grad=(_unwrap(grad),))
+end
+
+function Flux.withgradient(f, adtype::AutoMooncake, x...)
```
gdalle:
```suggestion
function Flux.withgradient(f::F, adtype::AutoMooncake, x::Vararg{Any,N}) where {F,N}
```

ext/FluxMooncakeExt.jl
```diff
@@ -0,0 +1,33 @@
...
+    val, grad = DI.value_and_gradient(x -> f(x...), adtype, x)
```
gdalle:
```suggestion
    val, grad = DI.value_and_gradient(Base.splat(f), adtype, x)
```

ext/FluxMooncakeExt.jl
gdalle: Also beware that these are strictly speaking Mooncake internals. The right unwrapping mechanism is https://github.com/chalk-lab/Mooncake.jl/pull/860

src/gradient.jl
```diff
@@ -233,3 +232,64 @@ withgradient(f, args::EnzymeCore.Const...; zero::Bool=true) = throw(ArgumentErro
...
+    val, grad = DI.value_and_gradient(f, adtype, x)
+    return (val=val, grad=(grad,))
+end
+
+function withgradient(f, adtype::ADTypes.AbstractADType, x...)
+    val, grad = DI.value_and_gradient(x -> f(x...), adtype, x)
+    return (val=val, grad=grad)
+end
+
+# use our Enzyme infrastructure instead of DI
```
gdalle: Is there anything concrete for me to fix inside DI for your use of Enzyme?

src/gradient.jl
```diff
@@ -233,3 +232,64 @@ withgradient(f, args::EnzymeCore.Const...; zero::Bool=true) = throw(ArgumentErro
...
+julia> Flux.gradient(f, AutoMooncake(), [1.0, 2.0, 3.0])
+([2.0, 2.0, 2.0],)
+```
+"""
+function gradient(f, adtype::ADTypes.AbstractADType, x)
+    g = DI.gradient(f, adtype, x)
+    return (g,)
+end
+
+function gradient(f, adtype::ADTypes.AbstractADType, x...)
```
gdalle: Same remarks as in the MooncakeExt file apply here (in fact, the code duplication is probably unnecessary if you define
```julia
_unwrap(::AbstractADType, result) = result
```
in the main package and only overload it for Mooncake

gdalle (changes requested):
This looks promising!
Note that without an amortized preparation step, Mooncake will be very slow. Hopefully that can be fixed by https://github.com/chalk-lab/Mooncake.jl/pull/900

CarloLucibello:
> If you’re thinking of extending the api to support more general ADTypes I think it would likely be better to have the core function be something like update_with_gradient!(model, optimizer, ADType) that fuses the update and gradient.

we can add a `training_step!(loss, adtype, model, data, opt_state)`

wsmoses (commented):
Can DI be an extension here instead of a proper dependency? I worry a bit that users of flux and non DI based code can end up in version resolution hell since DI needs to mark the version compat of all auto diff packages simultaneously. I recall this causing issues for diffractor in the past.

You can still have the DI ext loaded as a req of any of the other exts here

test/ext_reactant/reactant.jl
```diff
@@ -86,7 +86,7 @@ end
     for (model, x, name) in models_xs
         @testset "check grad $name" begin
             println("testing $name with Reactant")
-            test_gradients(model, x; loss, compare_finite_diff=false, test_reactant=true)
+            test_gradients(model, x; loss, reference=AutoZygote(), compare=AutoReactant())
```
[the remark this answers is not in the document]
wsmoses: For example, the corresponding flux integration test on our side which has remained consistently passing: https://github.com/EnzymeAD/Reactant.jl/blob/main/test/nn/flux.jl

src/gradient.jl
CarloLucibello: using DI also for Enzyme seems to work fine. The main problem in general is that DI only supports array inputs officially, so differentiation with respect to flux's model could break at any point.

CarloLucibello: Also, a limitation of DI compared to the `Flux.withgradient`, is the lack of support for auxiliary outputs, that is issue https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/720

src/gradient.jl
```diff
@@ -233,3 +236,82 @@ withgradient(f, args::EnzymeCore.Const...; zero::Bool=true) = throw(ArgumentErro
 _enzyme_withgradient(f, args...; zero) = throw(ArgumentError(
     "Methods like `withgradient(f, x::Duplicated)` are only available when Enzyme is loaded."
 ))
+
+
+"""
+    gradient(f, adtype::AbstractADType, args...)
+
+A version of `gradient` that allows specifying the automatic differentiation backend via `adtype`. 
+Any backend compatible with `DifferentiationInterface.jl` can be used.
```
wsmoses: The fact that Enzyme and Zygote don't go via DI is important and should be distinguished in the docs (and also see my comment about moving DI as an extension to avoid compatibility issues).

Could this be reworded as AutoEnzyme and AutoZygote are supported natively, as well as any backend supported by DI.

see  https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/265#issuecomment-3270544686 for an example of code supported natively by enzyme but breaks via DI (among others which also just are slower)

ext/FluxEnzymeExt.jl
```diff
@@ -6,40 +6,16 @@ import Flux.Train: _enzyme_train!
...
-#     if Enzyme.guess_activity(typeof(x), Reverse) <: Duplicated
-#         fill!(x, zero(eltype(x)))
-#     elseif Enzyme.guess_activity(typeof(x), Reverse) <: Const
-#         # that's OK
-#     else
-#         error("not sure what it should do for Active?")
-#     end
-# end
+# Ref is needed for working with immutable field, e.g. LayerNorm.ϵ
+_make_zero!(x) = EnzymeCore.make_zero!(Ref(x))
```
wsmoses: bump at my earlier comment, this should be remake_zero!(x)

src/gradient.jl
```diff
@@ -27,6 +35,17 @@ julia> Flux.gradient([7, 11], 0, 1) do x, y, d
        end
 ([14.0, 22.0], 2.0, nothing)
 ```
+Specifying other AD backends:
+
+```julia-repl
+julia> using Mooncake
+
```
wsmoses: Maybe use autozygote here for clarity (showing the different interface)

CarloLucibello:
@wsmoses do you know why the Enzyme tests are erroring on recurrent layers?
https://github.com/FluxML/Flux.jl/actions/runs/20655148072/job/59306597399?pr=2645#step:6:1264

ext/FluxEnzymeExt.jl
wsmoses: Hm on mobile so can't lookup, but I thought I made an earlier comment that the proper thing to do here is call remake_zero!(x) instead of make_zero!(x). However regardless of which you pick this will change behavior as this now no longer can assume the different parts of the model don't Alia's each other so may be slower

test/ext_reactant/reactant.jl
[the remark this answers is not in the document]
CarloLucibello: reactant's specific problems will be addressed in another PR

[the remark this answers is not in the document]
wsmoses: Can it get addressed before this PR is released? Otherwise as evidenced below this is likely to cause breakages.

CarloLucibello:
> yeah so that specific error indicates that type inference inside the method deduced a concrete type (Float64) whereas typeinference at the callside found an abstract type (Any). Likely what's happening from the DI wrapper is that its deducing Any for both (and Enzyme works perfectly well on). It's better to retain the type stable version, though we ought figure out where/why it's lost. In that particular section of the code (sorry this is a bit internals specific) atm we support the inner type being less known than the outer type (or equal ofc), but not the other way round. So something changed due to this PR via type inference, though the specifics are a bit unclear (I can try to take a look and debug this more deeply in a bit).

I'll stick with DI then. If you want to experiment with direct enzyme call just uncomment the AutoEnzyme specialization for `withgradient`.

wsmoses:
> > yeah so that specific error indicates that type inference inside the method deduced a concrete type (Float64) whereas typeinference at the callside found an abstract type (Any). Likely what's happening from the DI wrapper is that its deducing Any for both (and Enzyme works perfectly well on). It's better to retain the type stable version, though we ought figure out where/why it's lost. In that particular section of the code (sorry this is a bit internals specific) atm we support the inner type being less known than the outer type (or equal ofc), but not the other way round. So something changed due to this PR via type inference, though the specifics are a bit unclear (I can try to take a look and debug this more deeply in a bit).
> 
> 
> 
> I'll stick with DI then. If you want to experiment with direct enzyme call just uncomment the AutoEnzyme specialization for `withgradient`.

Use of DI will break reactant (specifically DI's internal preparation has been shown to cause issues in this sort of setup isn't considered supported from the Enzyme/Reactant side atm and is likely to break without warning).

wsmoses:
If you want to move forward here without causing breakage before I can sort out the cause of the type instability mismatch I think your options are:
* fix the type instability introduced in this PR as distinct from main (I'm not quite sure it's not inferred but I presume the standard code typed and friends can help)
* explicitly don't use DI on reactant compilation. I opened https://github.com/SciML/ADTypes.jl/pull/140 earlier to distinguish DI uses of an adtype from non DI versions, so once that merged then you can use autoenzyme from within reactant and autodi{autoenzyme} for the other tests in the meantime

gdalle:
> that PR is distinct. That PR has DI on the outside of a reactant compile. What doesn't work here is Reactant compile of DI

Sorry for the misunderstanding. Can you provide an MWE of Reactant failing to compile something with DI? I think I might know where a possible issue comes from, and if so it would be easy to fix

gdalle:
Might want to rename the PR if DI is no longer used

gdalle:
Out of curiosity, ignoring Enzyme, what's missing from DI for the purposes of Flux+Mooncake? Is the reasoning simply "there are only two backends you care about now so you might as well use them manually"? @CarloLucibello

CarloLucibello:
Yes that + the fact that a extension was needed (for different reasons) for each backend in any case.

PR #2675, opened 2026-04-28
Add Untrainable layer type

fps:
Sometimes you just need a way to keep a layer from being updated. I'l update this PR with unchecked things below if there is interest in it at all :)

- [ ] Tests are added
- [ ] Entry in NEWS.md


fps:
Well `Flux.freeze` seems just as good.

CarloLucibello:
> Well Flux.freeze seems just as good.

Yes, I wouldn't have a second way to do this. Also, it is easy to implement manually.

PR #2681, opened 2026-06-07
Add train_step! and TrainState

CarloLucibello:
## Summary

Adds a step-by-step training API to complement `train!`.

The goal is to have a method, `train_step!`, that combines the gradient computation and the update step, while also implementing under the hood some extra goodies like: 1) caching allocator for gpu; 2) ad framework cache; 3) reactant compilation. 

## Added types and methods

- **`TrainState(rule, model)`** — a mutable container bundling everything needed to take training steps:
  - `model` — stored inside the state and mutated in place by `train_step!`
  - `rule` — the `Optimisers.AbstractRule` (e.g. `Adam()`)
  - `opt_state` — the optimiser state from `Flux.setup`
  - `step::Int` — number of steps taken so far
  - `alloc_cache` — a `GPUArrays.AllocCache` reused across steps to reduce GPU memory pressure
  - `ad_cache` — an (untyped) slot for AD frameworks (e.g. Mooncake) that can reuse gradient-preparation state across steps; defaults to `nothing`

- **`train_step!(loss, [adtype,] data, state::TrainState)`** — performs a single optimisation step, updating `state.model` in place and returning the loss. Inspired by `Lux.Training.single_train_step!`, adapted to Flux's convention where the model carries its own parameters. Storing the model inside the state means the same object can be threaded through every iteration and compiled (e.g. with Reactant).
  - Evaluates `loss(model, data...)` for tuple data, else `loss(model, data)`, matching `train!`.
  - Defaults to Zygote, or Enzyme when the model is `Duplicated`.
  - Throws `DomainError` on non-finite loss.
  - **Auxiliary outputs**: like `withgradient`, the loss may return a `Tuple`/`NamedTuple` with the scalar loss first; the full value is then returned (currently Zygote-only, matching `withgradient`).

```julia
state = Flux.TrainState(Adam(1e-3), model)
for data in dataloader
    Flux.train_step!(loss, data, state)
end
```

## Status / open questions

- The `ad_cache` field is an integration point — nothing reads/writes it yet. A follow-up (e.g. a Mooncake extension) would populate and reuse it.
- Aux output is not yet supported with Enzyme (inherited from `withgradient`).
- No tests or docs-page entries added yet — happy to add `test/train.jl` cases and a manual reference if the API shape looks right.

Feedback welcome on the API: argument order is `(loss, [adtype,] data, state)` and the model living inside `TrainState` rather than being passed separately.

### DIFF
--- a/src/Flux.jl
+++ b/src/Flux.jl
@@ -63,7 +63,7 @@ export Chain, Dense, Embedding, EmbeddingBag,
   # from OneHotArrays.jl
   onehot, onehotbatch, onecold,
   # from Train
-  setup, train!,
+  setup, train!, train_step!, TrainState,
   # from Optimsers.jl
   freeze!, thaw!, adjust!, update!, trainable,
   # from Zygote.jl
--- a/src/train.jl
+++ b/src/train.jl
@@ -10,7 +10,7 @@ using ProgressLogging: @progress, @withprogress, @logprogress
 using EnzymeCore: Duplicated
 using ADTypes: AbstractADType, AutoEnzyme, AutoZygote
 
-export setup, train!
+export setup, train!, train_step!, TrainState
 
 """
     opt_state = setup(rule, model)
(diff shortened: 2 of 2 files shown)


CarloLucibello:
done in other PRs

PR #2685, opened 2026-06-12
Update Buildkite pipeline for the new JuliaGPU cluster

maleadt:
The JuliaGPU Buildkite agents have moved to a dedicated cluster, where the single `juliagpu` queue has been split into per-backend queues. Steps now select agents using `queue: "cuda"` or `queue: "rocm"` instead of `queue: "juliagpu"` combined with a `cuda`/`rocm` tag.

Since queues are now scoped to a cluster, a single pipeline can no longer mix JuliaGPU and `juliaecosystem` jobs: `pipeline.yml` now contains only the GPU jobs, while the `juliaecosystem` steps have moved unmodified to `pipeline-julia.yml`, to be uploaded by a separate pipeline in the ecosystem cluster once that is back online.

This change only affects agent selection; the steps themselves are unchanged.


maleadt:
Updated: the JuliaGPU cluster now has a Metal agent (`queue: "metal"`), so the Metal job is back in the main pipeline and no separate ecosystem-cluster pipeline is needed.

PR #2694, opened 2026-07-22
Fix three DDP correctness issues: PMI guardrails, data padding, unuse…

edopica:
# Fix three DDP deadlock/crash scenarios in DistributedUtils

This PR addresses three independent correctness issues in Flux's distributed
training path (MPI backend), originally introduced in the distributed data parallel PR (#2464). Each bug causes either a deadlock or a hard crash
when running DDP with non‑trivial models or launch configurations.

Additionally, this PR enables MPI distributed tests to run automatically on Linux CI runners.

## 1. PMI version mismatch guardrails

**Problem:** Julia's default `MPICH_jll` speaks PMI2.  When launched under a
PMIx launcher (e.g. `srun` without `--mpi=pmi2`) or under OpenMPI's `mpirun`,
`MPI.Init()` aborts with an uninformative message, making diagnosis difficult.
Even inside a Slurm allocation with `mpiexecjl`, the missing PMI2 variables
produce a confusing hang.

**Fix:** Before calling `MPI.Init()`, check for environment markers:

- `PMIX_RANK` → error with PMIx mismatch message  
- `OMPI_COMM_WORLD_RANK` → error with OpenMPI mismatch message  
- `SLURM_JOB_ID` but no `PMI2_*` → warn about missing `--mpi=pmi2`

A `force=true` keyword allows advanced users to bypass the check (e.g. when
using a custom MPI build).  The logic is entirely inside `FluxMPIExt` and
runs before any MPI symbols are touched.

## 2. Padding in DistributedDataContainer to prevent allreduce deadlocks

**Problem:** When `n_samples % n_workers ≠ 0`, `DistributedDataContainer`
assigned fewer batches to the last rank.  That rank would exit the training
loop early while the other ranks blocked inside an `allreduce` call, causing a
permanent deadlock.

**Fix:** Pad the index sequence so every rank receives the same number of
elements.  Surplus slots are filled by wrapping around from the start
(`[1, 2, …, n_samples, 1, 2, …]`).  The data‑loading logic is unchanged;
only the index array passed to `DistributedDataContainer` is padded.

## 3. `resolve_unused_parameters!!` for conditional computation graphs

**Problem:** A model with conditional branches (e.g. `if`/`else` selecting
different layers) can produce `nothing` gradients on ranks that did not execute
a particular parameter.  When `DistributedOptimizer` later calls `allreduce`
over all gradients, the mismatched set of buffers deadlocks — different ranks
participate in different collectives.

**Fix:** New public function `resolve_unused_parameters!!(backend, gs, model)`
replaces every `nothing` gradient with a zero‑filled array matching the
parameter's shape, element type, and device.  It walks nested structures
(NamedTuples, etc.) and works with arbitrary tree depths thanks to Functors'
`fmap`.

## Tests added / updated

| File | Tests |
|------|-------|
| `test/ext_distributed/data_distributedtest.jl` | Renamed from `data.jl` to run with the test suite. 3 test groups: evenly-divisible partition, non-divisible (padded) partition with correct padded sum, and the original 10‑element case with expected padded sum. |
| `test/ext_distributed/unused_parameters_distributedtest.jl` (new) | 4 test groups (20 tests): replaces‑nothing‑with‑zeros, nested NamedTuples, no‑op when all present, preserves exact shapes and element types (Float32/Float64). Runs on both MPI and NCCL backends. |
| `test/ext_distributed/reduce_distributedtest.jl` | Fixed pre-existing upstream breakage in `reduce_distributedtest.jl` (added missing imports required for the test to run at all). |
| Pre-existing Orphaned Tests | Renamed `common.jl`, `optimizer.jl`, and `synchronized.jl` to `_distributedtest.jl` suffix so they are picked up by the test runner. Also fixed pre-existing upstream breakage (missing imports). |

All tests pass with `mpiexecjl -n 2` on CPU/MPI.

## Backward compatibility

All changes are additive or replace a broken code path with a working one.
`resolve_unused_parameters!!` is a new, optional public API — existing
training loops that don't use conditional computation are unaffected.
The PMI check can be bypassed with `force=true`.


docs/src/guide/gpu.md
```diff
@@ -394,6 +394,24 @@ Epoch 3: Loss 0.012763695
...
+
+To prevent this, use `DistributedUtils.resolve_unused_parameters!!` to replace `nothing` gradients with zero-filled arrays before updating the optimizer:
+
+```julia-repl
+julia> for epoch in 1:100
+           global model, st_opt
+           l, grad = Zygote.withgradient(loss, model)
+           
+           # Resolve nothing gradients to avoid deadlocks
+           grad_resolved = DistributedUtils.resolve_unused_parameters!!(backend, grad[1], model)
```
CarloLucibello, line 409: Users shouldn't do this manually. 
We should call `resolve_unused_parameters!!` internally, as part of `Flux.withgradient` or ` Optimisers.update`. This could be part of a future PR, but for the time being I would avoid adding anything to the guide.

test/ext_distributed/optimizer_distributedtest.jl
```diff
@@ -1,6 +1,22 @@
...
+    import MPI
+    const backend_type = MPIBackend
+elseif backend_string == "nccl"
+    import MPI, NCCL, CUDA
+    const backend_type = NCCLBackend
+else
+    error("unsupported backend: $backend_string")
+end
+
+const dev = Flux.cpu
```
CarloLucibello, line 19: this boilerplate code could be added to
https://github.com/FluxML/Flux.jl/blob/master/test/test_module.jl

instead of being part of each test file

test/ext_distributed/synchronized_distributedtest.jl
```diff
@@ -1,6 +1,22 @@
-const input_args = length(ARGS) == 2 ? ARGS : ("CPU", "mpi")
-const backend_type = input_args[2] == "nccl" ? NCCLBackend : MPIBackend
-const dev = input_args[1] == "CPU" ? Flux.cpu : Flux.gpu
+using Test
+using Flux
+using Flux: DistributedUtils, MPIBackend, NCCLBackend
+using Optimisers
+
+
+backend_string = ARGS[1]
```
CarloLucibello, line 7: instead of reading ARGS this should be set by ENV variables in 
https://github.com/FluxML/Flux.jl/blob/master/test/runtests.jl
as anything else

CarloLucibello:
before merging this PR, CI should be activated for the distributed tests, currently it is off

CarloLucibello:
what is left to do here?

edopica:
I was just about to push the latest changes. I've addressed your comments, fixed a bug in the CI workflow and dropped the unused parameters fix for this PR as new problems emerged that made the work's scope too vast.

edopica:
The PR now contains only:

1. a narrow guard for the known unsafe MPICH and PMIx/OpenMPI-launcher mismatch;
2. equal-length data shards with valid cyclic padding;
3. dedicated two-rank and four-rank MPI CI.

I removed resolve_unused_parameters!!, its tests, its NEWS entry, and all related documentation. Automatic unused-gradient support needs a separate optimizer-level design, so I have deferred it to another PR.

I also addressed the test-infrastructure comments:

- the distributed tests share one setup file to remove boilerplate;
- child processes read the backend from `FLUX_TEST_DISTRIBUTED_BACKEND`;
- the parent runner sets that variable with `addenv`;
- the main test suite routes distributed tests through one dedicated runner;

the four-rank OpenMPI CI job uses `--use-hwthread-cpus` because github exposes only two cores for the CI vm, which was making the test crash due to the number of cores being less than the number of processes. Now the four-rank job assigns processes to threads instead of cores.

The data tests cover N = 0, 1, 2, 7, 10 with two and four ranks. The documentation now explains that cyclic duplicates affect both metrics and training gradients.

Native NCCL validation and automatic unused-gradient support remain outside this PR.

PR #2703, opened 2026-08-04
Autocast: mixed precision via layer wrappers

CarloLucibello:
PyTorch-`torch.autocast`-style mixed precision for Flux, implemented as a **type-domain model transform**. Companion to the scope-based prototype in #2702 — this is the version proposed for merge; the two can be compared (ergonomics, inference, performance).

## API

```julia
# keyword on gradient / withgradient / train!
loss, grad = Flux.withgradient(model; autocast=BFloat16) do m
    Flux.logitcrossentropy(m(x), y)
end
Flux.update!(opt_state, model, grad[1])          # opt_state is for the plain model

# or wrap a model directly (e.g. for inference)
autocast(model, BFloat16)(x)
```

`autocast(model, T)` returns a model sharing the original parameter arrays, with matmul/convolution-heavy layers wrapped in `AutocastDown{T}` (cast params + inputs to `T`, run the original layer), `BatchNorm` in `AutocastKeep{T}` (activation stays half, statistics folded in `Float32`) and the other norms in `AutocastUp` (`Float32`). Parameters stay `Float32` ("master weights"); the keyword wraps inside the differentiated closure, so gradients come back `Float32` and **shaped like the unwrapped model** — the optimiser setup is unchanged.

## Benchmark findings (ResNet-18, RTX 5090)

Full harness in [`perf/autocast/`](perf/autocast/) (Flux `benchmark.jl` + a matching PyTorch `benchmark_pytorch.py`). Forward+backward step, batch 128:

| variant | time | vs Float32 |
|---|---|---|
| Float32 baseline | 39.2 ms | — |
| **autocast BFloat16** | **30.9 ms** | **1.27×** |
| **autocast Float16** | **28.3 ms** | **1.39×** |
| static `bf16mix` (also this PR) | 29.7 ms | 1.32× |
| PyTorch `torch.autocast` bf16 | 22.7 ms | 1.67× |

A naive op-level autocast actually *regressed* here (bf16 was 0.86×, slower than Float32) — profiling in the perf dir showed the bf16 conv speedup (~9 ms) was swamped by ~13 ms of `BatchNorm` fp32 round-trips and ~3 ms of differentiating the wrapper tree. Two fixes close the gap to static casting:

- **`BatchNorm` `:keep` mode**: keep the activation in the half type through the norm (statistics folded in `Float32` by the kernel — cuDNN on the GPU) instead of the up/down round-trip. Device-adaptive: on the CPU it casts up to `Float32` (half precision isn't faster there, and BFloat16 miscompiles on Julia ≥1.11/x86, [BFloat16s.jl#107](https://github.com/JuliaMath/BFloat16s.jl/issues/107)).
- **An `rrule` for `autocast`**: its adjoint is just unwrapping the cotangent, so Zygote stops differentiating through the wrapper-building `fmap` (bit-identical gradient, wrap-once cost).

Autocast is now within ~4% of the static-casting ceiling. Time is a genuine win; **peak memory** is still above PyTorch — a separate issue (Flux/Zygote keep all activations live within a step; even the Float32 baseline uses ~3× PyTorch's memory), independent of autocast and noted in the perf dir.

## Why the wrapper (vs a runtime scope)

Precision is a *type parameter* of the wrapper, so both plain and wrapped forward passes keep their **exact inferred element type** (`autocast(Dense(3=>4), Float16)(x) :: Matrix{Float16}`), and the layer source files are untouched from master. There is no inference penalty for defining autocast and no cost when it is unused.

Custom layers opt in with one line: `Flux.autocast_mode(::MyLayer) = :down` (or `:up`/`:keep`).

## Also in this PR
- `f16`/`bf16` are full parameter casts (PyTorch `model.half()`/`.bfloat16()`); new `f16mix`/`bf16mix` keep `BatchNorm`/`InstanceNorm`/`GroupNorm` statistics + affine in `Float32` (required by the GPU norm kernels for half-precision inputs).
- **Breaking**: `Flux.Losses` now always accumulate in `Float32` for `Float16`/`BFloat16` inputs (numerically safer under mixed precision); `Float32`/`Float64` unaffected.

## Notes / limitations
- `Float16` may need loss scaling (a `GradScaler` is planned as a follow-up); `BFloat16` trains without it.
- `MultiHeadAttention` runs attention (incl. softmax) in `T` via its wrapped projections; `Embedding` is left in `Float32` (matching PyTorch, avoids casting a large table each forward).
- `autocast=BFloat16` under `AutoEnzyme` is blocked upstream: [EnzymeAD/Enzyme.jl#3430](https://github.com/EnzymeAD/Enzyme.jl/issues/3430) (`Float16` + Enzyme works).
- Verified on CPU (Zygote/Mooncake/Enzyme) and on an RTX 5090 (CUDA).


CarloLucibello:
TODO factor out the following PRs:
- f16mix/bf16mix
- losses upcast

PR #2716, opened 2026-08-08
BatchNorm: error on a single value per channel in training (fixes #1992)

### LINKED ISSUE #1992: Batchnorm's variance becomes NaN when minibatch contains just one sample
Hi All,

I have found a bug in Batchnorm, when the minibatch contains just one sample.
MWE:
```julia
using Flux
bn = BatchNorm(1)
x = zeros(1,1)
gradient(() -> sum(bn(x)), Flux.params(bn))
bn.σ²
```

The problem is caused by using unbiassed estimate of the variance in function tracked stats
```julia

function _track_stats!(
  bn, x::AbstractArray{T, N}, μ, σ², reduce_dims,
) where {T, N}
  V = eltype(bn.σ²)
  mtm = bn.momentum
  res_mtm = one(V) - mtm
  m = prod(size(x, i) for i in reduce_dims)

  μnew = vec(N ∈ reduce_dims ? μ : mean(μ, dims=N))
  σ²new = vec(N ∈ reduce_dims ? σ² : mean(σ², dims=N))

  bn.μ = res_mtm .* bn.μ .+ mtm .* μnew
  bn.σ² = res_mtm .* bn.σ² .+ mtm .* (m / (m - one(V))) .* σ²newd
  return nothing
end
```
to fix it, it is sufficient to change the update the of variance to 
```julia
bn.σ² = res_mtm .* bn.σ² .+ mtm .* (m / (max(1,m - one(V)))) .* σ²newd
```
I am happy to prepare a proper PR, if it helps.

CarloLucibello:
Fixes #1992.

## Problem

`BatchNorm` in training mode silently produces `NaN` running statistics when a channel sees exactly one value (e.g. batch size 1 with no spatial dimensions):

```julia
bn = BatchNorm(1); trainmode!(bn)
bn(zeros(Float32, 1, 1))
bn.σ²   # was Float32[NaN]
```

The batch variance of a single value is `0`, so the Bessel-corrected running-variance update `m/(m-1)·σ² = 0/0` is `NaN`. I checked the other implementations for this exact case:

| Implementation | Result |
|---|---|
| Flux CPU (NNlib) | `σ² = NaN` |
| Flux GPU (cuDNN) | `σ² = NaN` |
| PyTorch | **errors**: *"Expected more than 1 value per channel when training"* |

So `NaN` is silent corruption, and the reporter's proposed `m/max(1, m-1)` would produce a finite-but-meaningless value that diverges from *both* cuDNN and PyTorch.

## Fix

Match PyTorch: throw an informative `ArgumentError` when training and there is a single value per channel. The guard sits in the layer forward pass before dispatch, so CPU and GPU behave identically. `testmode!` inference with batch size 1 is unaffected, and batches with spatial extent (`m > 1`, e.g. `N=1` but `W=4`) work as before.

## Tests

Added coverage in `test/layers/normalisation.jl`: errors in trainmode (forward and gradient), works with spatial extent, works in testmode. Full `layers/normalisation` suite passes locally.

### DIFF
--- a/src/layers/normalise.jl
+++ b/src/layers/normalise.jl
@@ -291,8 +291,19 @@ end
 
 function (BN::BatchNorm)(x::AbstractArray{T,N}) where {T,N}
   _size_check(BN, x, N-1 => BN.chs)
+  training = _isactive(BN, x)
+  if training === true
+    # Values seen per channel = all dims except the channel dim (N-1). With a single
+    # value the batch variance is 0 and the (Bessel-corrected) running-variance update
+    # `m/(m-1)·σ²` is `0/0 = NaN`, so we reject it as PyTorch does (issue #1992). cuDNN
+    # would also silently emit NaN here.
+    m = prod(size(x)) ÷ size(x, N-1)
+    m == 1 && throw(ArgumentError(
+      "BatchNorm expected more than 1 value per channel when training, got input of size $(size(x)). " *
+      "The batch variance of a single value is undefined; use a larger batch or call `testmode!` for inference."))
+  end
   y = NNlib.batchnorm(BN.γ, BN.β, x, BN.μ, BN.σ², BN.momentum;
-                      eps=BN.ϵ, training=_isactive(BN, x), track_stats=BN.track_stats)
+                      eps=BN.ϵ, training, track_stats=BN.track_stats)
   return BN.λ.(y)
 end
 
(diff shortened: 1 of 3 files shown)


src/layers/normalise.jl
```diff
@@ -291,8 +291,19 @@ end
 
 function (BN::BatchNorm)(x::AbstractArray{T,N}) where {T,N}
   _size_check(BN, x, N-1 => BN.chs)
+  training = _isactive(BN, x)
+  if training === true
+    # Values seen per channel = all dims except the channel dim (N-1). With a single
+    # value the batch variance is 0 and the (Bessel-corrected) running-variance update
+    # `m/(m-1)·σ²` is `0/0 = NaN`, so we reject it as PyTorch does (issue #1992). cuDNN
+    # would also silently emit NaN here.
```
CarloLucibello:
```suggestion
```

Issue #2722, opened 2026-09-11
Failure with distributed data parallel training (DDP) with non-CUDA/ROCM aware MPI


Alexander-Barth:
The following example code fails for on CUDA and ROCM (with non-GPU aware MPI) when using PARALLEL=true
with Flux v0.16.11:

```julia
parallel = get(ENV,"PARALLEL","false") == "true"

if !isnothing(Sys.which("nvidia-smi"))
    import CUDA, cuDNN
else
    import AMDGPU
end

using Flux
using Optimisers
using Zygote
using Statistics
using Random
if parallel
    import MPI
end

Random.seed!(42)

function pprintln(backend,args...)
    total_workers = DistributedUtils.total_workers(backend)
    rank = DistributedUtils.local_rank(backend)
    for i in 0:(total_workers-1)
        if i == rank
            print("[$rank/$total_workers] ")
            println(args...)
            flush(stdout)
        end
        MPI.Barrier(backend.comm)
    end
end
pprintln(::Nothing,args...) = println(args...)

if parallel
    const backend_type = MPIBackend
    DistributedUtils.initialize(backend_type)
    backend = DistributedUtils.get_distributed_backend(backend_type)
else
    backend = nothing
end

pprintln(backend,"Flux.MPI_ROCM_AWARE ",Flux.MPI_ROCM_AWARE)
pprintln(backend,"Flux.MPI_CUDA_AWARE ",Flux.MPI_CUDA_AWARE)

T = Float32
device = gpu

x = randn(T,256,256,32,16*2)

channels = 2 .^ vcat(5:7,6:-1:5)

model = Chain(
    [Conv((3,3),channels[i] => channels[i+1],selu,pad=SamePad()) for i in 1:length(channels)-1]...
)

losses = T[]
model = model |> device

loss(x,y) = mean((x-y).^2)

opt_s = Optimisers.Adam(1f-4)

if parallel
    data = DistributedUtils.DistributedDataContainer(backend, x)
    model = DistributedUtils.synchronize!!(backend, DistributedUtils.FluxDistributedModel(model); root=0)
    opt = DistributedUtils.DistributedOptimizer(backend, opt_s)
else
    data = x
    opt = opt_s
end

opt_state = Optimisers.setup(opt, model)

opt_state = opt_state |> device

if parallel
    opt_state = DistributedUtils.synchronize!!(backend, opt_state; root=0)
end

dl = Flux.DataLoader(data,batchsize=16)

nepochs = 4


[...]

PR #2725, opened 2026-09-14
return send/receive buffers in __reduce! and __reduceall! for non-GPU aware MPI implementations

Alexander-Barth:
This fixes issue https://github.com/FluxML/Flux.jl/issues/2722.

I tested this on AMDGPU with DDP and I get the equivalent value of the loss function using 1 GPU and batch size of 64 and 2 GPUs with each a batch size of 32 using DDP. 

This PR is based on https://github.com/FluxML/Flux.jl/pull/2725/changes/5d901eb0b0bd7b69dfae40c4cc6ddb8c07d55aea ( adding the missing pad_mode argument on AMD GPU) .


Alexander-Barth:
Thank you Carlo for merging the PR!

CarloLucibello:
thank you! I just tagged a new release

